PR #8990, opened 2020-09-09
Show webpack error on live reload of frontend resources

### LINKED ISSUE #8960: Refactor webpack to server communication channel
`DevModeHandler` and the `webpack-dev-server` process running under Node communications over a socket connection. It is used for two distinct purposes:
1. To terminate `webpack-dev-server` server when the Java server dies, in order to avoid keeping Webpack running accidentally.
2. To send a message from `webpack-dev-server` to the Java server when webpack recompiles the bundle, in order to trigger live reload of connected browsers.

These features have developed at different times and [the implementing class](https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/server/DevServerWatchDog.java) should be refactored to isolate these two functionalities from each other. The pure watchdog and live reload message functionalities are absolutely orthogonal and there should be two different interfaces implemented by same class to reflect this.

### LINKED ISSUE #8965: Livereload of frontend resources shows error when it should not
# Description of the bug
When making frontend changes and you have an error, you need to fix the error, save and the save again for the change to be taken into use, or at least for the error popup to disappear

1.
```
git clone https://github.com/vaadin/cookbook.git
cd cookbook
mvn
```
2. Wait for it to start
3. open http://localhost:8080/
4. Add `a` in the beginning of `frontend/views/recipes-list-view.ts`
The browser reloads and shows an empty view, as expected

5. Remove `a` from the beginning of the file

### Expected behavior
The browser reloads and shows the original view

### Actual behavior
The browser reloads, shows the original view and a large popup about a webpack error that did not occur.
It includes among other things errors about `aimport` so it seems a bit like the old webpack error is shown.

### Versions:
    - Vaadin version: 17

joheriks:
Instead of sending using a separate async message for identifying frontend reloads, use the webpack output parser in `DevModeHandler`. This ensures that the webpack error shown is up-to-date when live reload triggers (avoiding the race #8965). It also simplifies the implementation.

Fixes #8965. Fixes #8960.


flow-server/src/main/java/com/vaadin/flow/server/DevServerWatchDog.java
```diff
@@ -84,21 +78,6 @@ void stop() {
         private Logger getLogger() {
             return LoggerFactory.getLogger(WatchDogServer.class);
         }
-
-        private void enterReloadMessageReadLoop(Socket accept) throws IOException{
```
denis-anisimov, line 88: Should we close the ticket about refactoring watchdog + live reload now ?

I would open another one (or modify the existing one) to refactor dev mode handler to avoid refer directly to live reload because of exactly the same reason: dev mode handler should not know anything about extra functionality.
Live reload should be just one of derefenced SPI which is triggered on every reload.

flow-tests/test-live-reload/src/main/java/com/vaadin/flow/uitest/ui/FrontendTestUtil.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+
+import com.vaadin.flow.server.VaadinService;
+import com.vaadin.flow.server.frontend.FrontendUtils;
+
+public class FrontendTestUtil {
```
denis-anisimov, line 12: Is it needed ?

I think Apache io library should contain something which just writes the file.

Or Java `Files.write`

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/DevServerWatchDog.java
joheriks: Let's close that one and open a new one (otherwise would be a total rewrite which makes history confusing as it is already linked from elsewhere).

flow-tests/test-live-reload/src/main/java/com/vaadin/flow/uitest/ui/FrontendTestUtil.java
joheriks: Thanks, that's much shorter.

joheriks: Also removed the util class.

denis-anisimov (approved):

flow-server/src/main/java/com/vaadin/flow/server/DevServerWatchDog.java
denis-anisimov: Sure

PR #8993, opened 2020-09-09
feat(offline): read @PWA to configure webpack build

### LINKED ISSUE #8983: Use @PWA annotation for configuring offline resources in frontend build
Expected: the generated webpack config lists offline resources found in `@PWA` annotation.

Actual: the generated webpack config lists offline resources from the default PWA configuration.

Part of epic #8978

platosha:
Fixes #8983

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
@@ -226,7 +226,7 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
     }
 
     /**
-     * Class visitor for collecting default annotation values
+     * Class visitor for collecting default annotation values.
      */
     private static class DefaultsAnnotationClassVisitor extends ClassVisitor {
         private final Map<String, Object> defaults;
@@ -244,7 +244,7 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
     }
 
     /**
-     * Method visitor for collecting default annotation values
+     * Method visitor for collecting default annotation values.
      */
     private static class DefaultsAnnotationMethodVisitor extends MethodVisitor {
         private final String methodName;
@@ -265,7 +265,7 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
     }
 
     /**
-     * Collects default annotation values
+     * Collects default annotation values.
      */
     private static class DefaultsAnnotationVisitor extends AnnotationVisitor {
         private final String methodName;
@@ -293,7 +293,7 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
     }
 
     /**
-     * Collects data from possibly repeated annotations
+     * Collects data from possibly repeated annotations.
      */
     private static class DataAnnotationVisitor extends RepeatedAnnotationVisitor {
         private final List<HashMap<String, Object>> data;
@@ -340,7 +340,7 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
     }
 
     /**
-     * Collects a list of annotation array values
+     * Collects a list of annotation array values.
      */
     private static class ArrayAnnotationVisitor extends AnnotationVisitor {
         private final List values;
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
@@ -42,10 +42,10 @@ import org.slf4j.LoggerFactory;
  * @since 2.0
  */
 final class FrontendAnnotatedClassVisitor extends ClassVisitor {
+    private static Map<String, Map<String, Object>> annotationDefaults = new HashMap<>();
     private final String annotationName;
     private final List<HashMap<String, Object>> data = new ArrayList<>();
     private final ClassFinder finder;
-    private static Map<String, Map<String, Object>> annotationDefaults = new HashMap<>();
 
     /**
      * Create a new {@link ClassVisitor} that will be used for visiting a
@@ -112,7 +112,8 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
     }
 
     @Override
-    public AnnotationVisitor visitAnnotation(String descriptor, boolean visible) {
+    public AnnotationVisitor visitAnnotation(String descriptor,
+            boolean visible) {
         String cname = descriptor.replace("/", ".");
         if (cname.contains(annotationName)) {
             return new RepeatedAnnotationVisitor() {
@@ -146,7 +147,8 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
 
                 // Only visited when annotation is repeated
                 @Override
-                public AnnotationVisitor visitAnnotation(String name, String descriptor) {
+                public AnnotationVisitor visitAnnotation(String name,
+                        String descriptor) {
                     // initialize in each repeated annotation occurrence
                     info = new HashMap<>();
                     return this;
@@ -177,10 +179,8 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
      */
     @SuppressWarnings("unchecked")
     public <T> Set<T> getValues(String parameter) {
-        return (Set<T>)data.stream()
-                .filter(h -> h.containsKey(parameter))
-                .map(h -> h.get(parameter))
-                .collect(Collectors.toSet());
+        return (Set<T>) data.stream().filter(h -> h.containsKey(parameter))
+                .map(h -> h.get(parameter)).collect(Collectors.toSet());
     }
 
     /**
@@ -206,11 +206,11 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
      * @return a set of all values found
      */
     @SuppressWarnings("unchecked")
-    public <T> Set<T> getValuesForKey(String key, String value, String property) {
-        return (Set<T>)data.stream()
+    public <T> Set<T> getValuesForKey(String key, String value,
+            String property) {
+        return (Set<T>) data.stream()
                 .filter(h -> h.containsKey(key) && h.get(key).equals(value))
-                .map(h -> h.get(property))
-                .collect(Collectors.toSet());
+                .map(h -> h.get(property)).collect(Collectors.toSet());
     }
 
     /**
@@ -230,9 +230,9 @@ final class FrontendAnnotatedClassVisitor extends ClassVisitor {
         }
         Set<T> values = getValues(parameter);
         if (values.isEmpty()) {
-            getLogger().debug("No value for " + parameter + " using default: "
-                    + getDefault(parameter));
-            return (T) getDefault(parameter);
+            getLogger().debug("No value for {} using default: {}", parameter,
+                    getDefault(parameter));
+            return getDefault(parameter);
         }
         return values.iterator().next();
     }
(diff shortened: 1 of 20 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
```diff
@@ -37,6 +41,7 @@
  * @since 2.0
  */
 final class FrontendAnnotatedClassVisitor extends ClassVisitor {
+    private static Map<String, Map<String, Object>> annotationDefaults = new HashMap<>();
```
fluorumlabs, line 44: This is not thread-safe collection.
Also field could be made `final`

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
```diff
@@ -66,34 +75,65 @@
...
 
+                @Override
+                public AnnotationVisitor visitArray(String name) {
+                    List values = new ArrayList<>();
+                    info.put(name, values);
+
+                    return new AnnotationVisitor(api, this) {
+                        @Override
+                        public void visit(String dummy, Object value) {
+                            if (data.indexOf(info) < 0) {
```
fluorumlabs, line 129: Can be replaced with `!data.contains(info)`

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
```diff
@@ -66,34 +75,65 @@
...
+    }
+
+    /**
+     * Visit recursively a class to find annotations.
+     *
+     * @param name
+     *            the class name
+     * @param visitor
+     *            the visitor to use
+     * @throws IOException
```
fluorumlabs, line 88:
```suggestion
     * @throws UncheckedIOException
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
```diff
@@ -66,34 +75,65 @@
...
+    public AnnotationVisitor visitAnnotation(String descriptor,
+            boolean visible) {
         String cname = descriptor.replace("/", ".");
         if (cname.contains(annotationName)) {
             return new RepeatedAnnotationVisitor() {
                 // initialize for non repeated annotations
                 HashMap<String, Object> info = new HashMap<>();
 
+                @Override
+                public AnnotationVisitor visitArray(String name) {
```
fluorumlabs, line 122: All those anonymous visitors should be extracted into static inner classes

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
```diff
@@ -165,10 +204,106 @@ public AnnotationVisitor visitAnnotation(String name, String descriptor) {
...
+        visitClass(annotationName, new AnnotationClassVisitor(defaults));
+
+        getLogger().debug("Default values for {}: {}", annotationName,
+                defaults);
+
+        return defaults;
+    }
+
+    private Logger getLogger() {
+        return LoggerFactory.getLogger(this.getClass());
```
fluorumlabs, line 256:
```suggestion
        return LoggerFactory.getLogger(FrontendAnnotatedClassVisitor.class);
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -369,6 +385,59 @@ private void computePackages() throws ClassNotFoundException, IOException {
...
+        String manifestPath = pwaVisitor.getValue("manifestPath");
+        String offlinePath = pwaVisitor.getValue("offlinePath");
+        String display = pwaVisitor.getValue("display");
+        String startPath = pwaVisitor.getValue("startPath");
+        List<String> offlineResources = pwaVisitor.getValue("offlineResources");
+
+        this.pwaConfiguration = new PwaConfiguration(true, name, shortName,
+                description, backgroundColor, themeColor, iconPath,
+                manifestPath, offlinePath, display, startPath,
+                offlineResources.toArray(new String[offlineResources.size()]));
```
fluorumlabs, line 438: Replace with an empty String[] array constant: preallocation is not needed since Java 6

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
```diff
@@ -407,6 +426,62 @@ private Object invokeAnnotationMethod(Annotation target,
...
+            String manifestPath = getAnnotationValueAsString(pwa,
+                    "manifestPath");
+            String offlinePath = getAnnotationValueAsString(pwa,
+                    "offlinePath");
+            String display = getAnnotationValueAsString(pwa, "display");
+            String startPath = getAnnotationValueAsString(pwa, "startPath");
+            String[] offlineResources = (String[]) getAnnotationValue(pwa,
+                    "offlineResources");
+
+            return new PwaConfiguration(true, name, shortName, description,
```
fluorumlabs, line 476: `shortName` might be `null` here, `PwaConfiguration` constructor will throw NPE in that case

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
fluorumlabs: I have doubts about this static collection: SpringBoot dev tools can do a hot restart of Spring Application in certain cases (method signature changes, for example); in this case the collection contents will be retained.

platosha: Changed to `final` instance member. Thus it re-reads defaults upon every parent build task execution.

AFAIK this is a build step rather then runtime code. I wonder if thread-safety critical here. At least this change does not make it worse, since this is only used in `FrontendDependencies.java`, which has a bunch of HashMaps anyway.

Also, this only concerns framework-provided `@PWA` annotation default values. Would not expect those to change between hot reloads. Or am I wrong?

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
platosha: I don’t see how that’s possible in practice. `@PWA` annotation is found at this point, and seems like it must have `shortName` by definition. I’ve tried declaring `@PWA` with `shortName` null and absent, it does not compile.

Added an assert just in case to make dev tools happy though.

fluorumlabs: You are correct, `com.vaadin.flow.server.frontend.scanner.FullDependenciesScanner#getAnnotationValueAsString` confused me

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
fluorumlabs: In `FrontendDependencies` hashMaps are instance variables, so that's ok. 

I had a second look at the code, and realized that you can just use https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/Method.html#getDefaultValue() intead of invoking ASM :)

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
```diff
@@ -165,10 +170,187 @@ public AnnotationVisitor visitAnnotation(String name, String descriptor) {
...
+            return getDefault(parameter);
+        }
+        return values.iterator().next();
+    }
+
+    private <T> T getDefault(String parameter) {
+        return (T) annotationDefaults.get(annotationName).get(parameter);
+    }
+
+    private Map<String, Object> readAnnotationDefaultValues(
```
fluorumlabs, line 208: Consider using `java.lang.reflect.Method#getDefaultValue` instead of ASM to collect default annotation values.

fluorumlabs (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -370,6 +386,56 @@ private void computePackages() throws ClassNotFoundException, IOException {
...
+                getFinder(), PWA.class.getName());
+        Class<?> appShellConfiguratorClass = getFinder().loadClass(
+                AppShellConfigurator.class.getName());
+
+        for (Class<?> hopefullyAppShellClass :
+                getFinder().getAnnotatedClasses(PWA.class.getName())) {
+                    if (!Arrays.asList(hopefullyAppShellClass.getInterfaces())
+                            .contains(appShellConfiguratorClass)) {
+                        throw new IllegalStateException(
+                                ERROR_CAN_ONLY_HAVE_ONE_PWA_ANNOTATION);
```
Haprog, line 406: The name of the error message constant `ERROR_CAN_ONLY_HAVE_ONE_PWA_ANNOTATION` seems counter intuitive in this case, though the actual value does mention both cases (`"There can only be one @PWA annotation and it must be set on the " + AppShellConfigurator.class.getSimpleName() + " implementor."`).

I'd consider renaming the constant as now it seems like it's only about the other case below.

My suggestion for a name: `ERROR_INVALID_PWA_ANNOTATION`.

(Not a blocker though)

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksTest.java
```diff
@@ -17,8 +19,11 @@
 import com.vaadin.flow.server.ExecutionFailedException;
 import com.vaadin.flow.server.connect.Endpoint;
 import com.vaadin.flow.server.frontend.NodeTasks.Builder;
+import com.vaadin.flow.server.frontend.scanner.ClassFinder;
 import com.vaadin.flow.server.frontend.scanner.ClassFinder.DefaultClassFinder;
+import com.vaadin.flow.server.frontend.scanner.samples.pwa.AppShellWithPwa;
 
+import static com.vaadin.flow.server.frontend.FrontendUtils.DEAULT_FLOW_RESOURCES_FOLDER;
```
Haprog, line 26: `DEAULT_FLOW_RESOURCES_FOLDER` (with typo) is now deprecated (since last rebase). Please use the corrected one:  
`DEFAULT_FLOW_RESOURCES_FOLDER` instead.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksTest.java
```diff
@@ -132,19 +140,46 @@ public void should_Generate_Connect_Files() throws Exception {
...
+    }
+
+    @Test
+    public void should_Set_PWA_Offline_Resources() throws Exception {
+        ClassFinder classFinder = new DefaultClassFinder(new HashSet<>(
+                new ArrayList<>(Arrays.asList(AppShellWithPwa.class))));
+
+        Builder builder = new Builder(classFinder, new File(userDir))
+                .withFlowResourcesFolder(
+                        new File(userDir, DEAULT_FLOW_RESOURCES_FOLDER))
```
Haprog, line 160: `DEAULT_FLOW_RESOURCES_FOLDER` (with typo) is now deprecated. Please use the corrected one:  
`DEFAULT_FLOW_RESOURCES_FOLDER` instead.

Haprog (changes requested):
Looks mostly good to me, but found few minor issues.

haijian-vaadin (approved):

Haprog (approved):

PR #9002, opened 2020-09-11
Fixed: flow-server now closes resources properly when parsing java classes

mvysny:
The problem is that the `ClassReader(InputStream)` constructor doesn't close the input stream itself, which is then left open hanging. This is especially troublesome for Gradle Plugin since the opened input stream is left hanging in the Gradle Daemon, which can stay around for hours. This could be a possible cause for https://github.com/vaadin/vaadin-gradle-plugin/issues/81

This PR needs to be cherry-picked to 4.0 and also 2.3 branches.


mvysny:
This PR possibly fixes failing builds on Windows with the Gradle plugin, and therefore could be a blocker.

caalador (approved):

PR #9003, opened 2020-09-11
Use JS rather than CSS file for testing Webpack live reload

joheriks:


flow-tests/test-live-reload/src/main/java/com/vaadin/flow/uitest/ui/LiveReloadView.java
```diff
@@ -87,16 +87,14 @@ private void handleClickJavaLiveReload(ClickEvent event) {
...
     }
 
     // Touch a frontend file to trigger webpack reload
     private void handleClickWebpackLiveReload(ClickEvent event) {
         addPageReloadingSpan();
-        String css = String.format(
-                "\nbody { background-color: rgb(%d,%d,%d); }",
-                random.nextInt(256), random.nextInt(256), random.nextInt(256));
-        writeToFrontendCSSFile(VaadinService.getCurrent(), css);
+        String css = "\nconsole.log('Hello world');";
```
mshabarov, line 96: Don't know if that may raise issues on Windows, but maybe better to replace `\n` with `%n`?

mshabarov (changes requested):

flow-tests/test-live-reload/src/main/java/com/vaadin/flow/uitest/ui/LiveReloadView.java
joheriks: Good point, fixed.

mshabarov (approved):

PR #9008, opened 2020-09-15
Read text for template element

### LINKED ISSUE #8975: Handle element initial text content from template mapped elements
Given a template with contents
```html
<a routerlink id="link">This is a router link</a>
```
It should be at least possible to change the text value from server side Java code, by expecting to replace the contents completely, including any children it might have.
For parsing the initial text value, it is only supported when the element has no child elements, but just text content.
```java
@Id("link") RouterLink link;
link.getText(); // returns "This is a router link"
link.setText("Another text value"); // would replace any children it has
```
If replacing children is not desired, it is recommended to workaround that by using another wrapper element for the text
```html
<a routerlink id="link"><iron-icon icon="vaadin:vaadin-h"></iron-icon><span id="link-text">Link with icon</span></a>
```
```java
@Id("link-text") Span linkText;

linkText.setText("At least this is possible to do");
```

denis-anisimov:
Fixes #8975

### DIFF
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/template/ChildOrderIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/template/ChildOrderIT.java
@@ -163,29 +163,28 @@ public class ChildOrderIT extends ChromeBrowserTest {
         assertNodeOrder(container, "Client text");
 
         clickAndWaitForContainerToChange(container, "prependChildToContainer2");
-        assertNodeOrder(container, "Client text", "Server text 1");
+        assertNodeOrder(container, "Server text 1Client text");
 
         clickAndWaitForContainerToChange(container, "prependChildToContainer2");
-        assertNodeOrder(container, "Client text", "Server text 2",
-                "Server text 1");
+        assertNodeOrder(container, "Server text 2", "Server text 1Client text");
 
         clickAndWaitForContainerToChange(container,
                 "addClientSideChildToContainer2");
-        assertNodeOrder(container, "Client text", "Server text 2",
-                "Server text 1", "Client text");
+        assertNodeOrder(container, "Server text 2", "Server text 1Client text",
+                "Client text");
 
         clickAndWaitForContainerToChange(container, "prependChildToContainer2");
-        assertNodeOrder(container, "Client text", "Server text 3",
-                "Server text 2", "Server text 1", "Client text");
+        assertNodeOrder(container, "Server text 3", "Server text 2",
+                "Server text 1Client text", "Client text");
 
         clickAndWaitForContainerToChange(container,
                 "removeChildFromContainer2");
-        assertNodeOrder(container, "Client text", "Server text 3",
-                "Server text 2", "Client text");
+        assertNodeOrder(container, "Server text 3", "Server text 2",
+                "Server text 1", "Client text");
 
         clickAndWaitForContainerToChange(container,
                 "removeChildFromContainer2");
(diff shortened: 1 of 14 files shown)


flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/template/ChildOrderIT.java
```diff
@@ -163,29 +163,28 @@ public void prependTextsFromServer_textsAreAddedBeforeExistingOnes() {
         assertNodeOrder(container, "Client text");
 
         clickAndWaitForContainerToChange(container, "prependChildToContainer2");
-        assertNodeOrder(container, "Client text", "Server text 1");
+        assertNodeOrder(container, "Server text 1Client text");
 
         clickAndWaitForContainerToChange(container, "prependChildToContainer2");
-        assertNodeOrder(container, "Client text", "Server text 2",
-                "Server text 1");
+        assertNodeOrder(container, "Server text 2", "Server text 1Client text");
```
joheriks, line 169: This seems correct, so the order was wrong before? I find the assertions messy after this change, as some of the calls to `assertNodeOrder` will be with the concated texts and some with separate parameters. Would be easier to understand and modify if the `\n` were removed from the inserted text pieces in the test views and the assertion changed to simply:
 
```java
    private void assertNodeOrder(WebElement container, String... nodes) {
        String texts = container.getText().replace("\n", "");
        String expected = Stream.of(nodes).collect(Collectors.joining(""));
        Assert.assertEquals(expected, texts);
    }
```

That would abstract away from the extra complication of whitespace between the added nodes. Unless for some reason that is actually part of the test?

joheriks (changes requested):

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/template/ChildOrderIT.java
denis-anisimov: No any bugs here previously or now.

The behavior has changed: "Client text" is now set to the `containerWithText` on the server side which means the element now has a text element (on the server side text is an element). As a result the server side `containerWithText` always has initial one element now.
And every test which touches this element has to consider index shift back or forward.
So this is just direct consequence of setting the text for the element in template.
(there is a genuine bug in another test failure)

I don't like this test since it does too many things and it's too easy to break other things correcting another.
This test is not a part of this PR. It just fails with the changes in this PR and that's why I'm modifying it.
I don't want to change anything else here accept the minimal changes to keep it passing.

You suggestion makes some sense, but it may introduce false positive result if strings in the nodes is modified.
Currently "\n" is used as a clear separator for text in nodes, concatenating all string in one big string removes any separation and the comparison becomes fragile.

I understand you concern about ugly string `"Server text 1Client text"`which I don't like either but this is a consequence of how DOM is done on the client side and absence the proper method of getting text recursively in Selenium.

I see call `assertNodeOrder(container, "Server text 3", "Server text 2","Server text 1Client text", "Client text")` with the current impl is more reliable and in fact quite expected within the current limitations (even though it's quite not-obvious why it works in this way, so you need to go deeply).

Even though I completely agree that from "internal API" point of view `assertNodeOrder(container, "Server text 3", "Server text 2","Server text 1","Client text", "Client text")` is just much better: but simple impl of this is unclear for me.

So I would prefer to keep this as is : this issue  can be considered separately.

joheriks: Ok, thanks for the clarifications.

joheriks (approved):

PR #9017, opened 2020-09-16
Send clear event if no other changes done on attach

### LINKED ISSUE #9015: Set text to an empty string for @Id-mapped element has impact on following actions with the element
Here is java code for a template: 
```
@Route(value = "com.vaadin.flow.uitest.ui.template.SetInitialTextView", layout = ViewTestLayout.class)
@Tag("set-initial-text")
@JsModule("SetInitialText.js")
public class SetInitialTextView extends PolymerTemplate<TemplateModel>
        implements HasComponents {

    @Id("child")
    private Div child;

    public SetInitialTextView() {
        // this is no-op since the text is an empty string by default but it
        // removes all children
        child.setText("");

        NativeButton button = new NativeButton("Add a new child",
                event -> addChild());
        button.setId("add-child");
        add(button);
    }

    private void addChild() {
        Div div = new Div();

        div.setId("new-child");
        div.setText("New child");
        child.add(div);
    }

}
```

The template :
```
import {PolymerElement} from '@polymer/polymer/polymer-element.js';
import {html} from '@polymer/polymer/lib/utils/html-tag.js';

class SetInitialText extends PolymerElement {
  static get template() {
    return html`
       <div id='child'></div>
       <button id="addClientSideChild" on-click="_addClientSideChild">Add client side child</button>
       <slot></slot>
    `;
   }
    
   _addClientSideChild() {
     let element = document.createElement("div");
     element.innerHTML = "Client child";
     element.id='client-side';
     this.$.child.appendChild(element);
  }
  static get is() {
    return 'set-initial-text'
  }
    
}
  
customElements.define(SetInitialText.is, SetInitialText);

```

Steps to reproduce :
- open a view
- Click the first button `Add client side child`.
- Click the second button `Add a new child`.
- The result is: the added client side element (via the first button) `"Client child"` is removed  and replaced by the `"New child"`.

The expected result: both `"Client child"` and  `"New child"` are in the DOM.

It happens because `setText("")` removes all children from the element before it's even attached.
The text is not set since the value is already an empty value but it removes all children and no other modifications happen.
As a result for some reason `ListClearChange` event is not sent initially but it's sent on the first change after attach.
The client side element is added before any such change so it's removed by the late `ListClearChange`.

The `ListClearChange` event has to be sent along with attach and may not be postponed. 
In this case everything works fine,

denis-anisimov:
Fixes #9015

### DIFF
--- a/flow-tests/test-root-context/frontend/lit/SetInitialText.js
+++ b/flow-tests/test-root-context/frontend/lit/SetInitialText.js
@@ -5,7 +5,7 @@ export class SetInitialTextLit extends LitElement {
   render() {
     return html`
       <div id='child'></div>
-       <button id="addClientSideChild" on-click="_addClientSideChild">Add client side child</button>
+       <button id="addClientSideChild" @click="${e => this._addClientSideChild()}">Add client side child</button>
        <slot></slot>
     `;
   }
@@ -13,7 +13,7 @@ export class SetInitialTextLit extends LitElement {
      let element = document.createElement("div");
      element.innerHTML = "Client child";
      element.id='client-side';
-     this.$.child.appendChild(element);
+     this.shadowRoot.querySelector('#child').appendChild(element);
   }
 }
   
--- a/flow-tests/test-root-context/frontend/lit/SetInitialText.js
+++ b/flow-tests/test-root-context/frontend/lit/SetInitialText.js
@@ -1,0 +1,20 @@
+
+import { LitElement, html } from "lit-element";
+
+export class SetInitialTextLit extends LitElement {
+  render() {
+    return html`
+      <div id='child'></div>
+       <button id="addClientSideChild" on-click="_addClientSideChild">Add client side child</button>
+       <slot></slot>
+    `;
+  }
+  _addClientSideChild() {
+     let element = document.createElement("div");
+     element.innerHTML = "Client child";
+     element.id='client-side';
+     this.$.child.appendChild(element);
+  }
+}
+  
+customElements.define("set-initial-text-lit", SetInitialTextLit);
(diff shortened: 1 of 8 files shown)


joheriks (dismissed):
Not blocking but: could new IT template be written in lit? Less maintenance if polymer is deprecated in the future.

denis-anisimov:
> Not blocking but: could new IT template be written in lit? Less maintenance if polymer is deprecated in the future.

Done.

flow-tests/test-root-context/frontend/lit/SetInitialText.js
```diff
@@ -0,0 +1,20 @@
+
+import { LitElement, html } from "lit-element";
+
+export class SetInitialTextLit extends LitElement {
+  render() {
+    return html`
+      <div id='child'></div>
+       <button id="addClientSideChild" on-click="_addClientSideChild">Add client side child</button>
```
joheriks, line 8: Does `on-click` registration work in lit? When running the view nothing happened clicking on the client button.  should it be something like ` @click=${() => this._addClientSideChild()}`

flow-tests/test-root-context/frontend/lit/SetInitialText.js
```diff
@@ -0,0 +1,20 @@
...
+       <slot></slot>
+    `;
+  }
+  _addClientSideChild() {
+     let element = document.createElement("div");
+     element.innerHTML = "Client child";
+     element.id='client-side';
+     this.$.child.appendChild(element);
```
joheriks, line 16: I could access it in Lit with `this.shadowRoot.getElementById`

joheriks (approved):

PR #9021, opened 2020-09-17
Keep ui reference to avoid its GC during test

denis-anisimov:


flow-server/src/test/java/com/vaadin/flow/component/ComponentTest.java
```diff
@@ -55,11 +55,13 @@
 import com.vaadin.tests.util.TestUtil;
 
 import elemental.json.Json;
-import org.mockito.Mockito;
+import net.jcip.annotations.NotThreadSafe;
 
 @NotThreadSafe
 public class ComponentTest {
 
+    private UI ui;
```
joheriks, line 63: Does it have any effect on GC to store UI instance here since JUnit anyway creates a new instance of the class for each test and during a single test run a reference is stored in `CurrentInstance`, preventing garbage collection?

joheriks (approved):

PR #9026, opened 2020-09-17
Set `display:none` in addition to `hidden` attribute when hiding elements inside a shadow root

### LINKED ISSUE #8256: setVisible(false) on PolymerTemplate does not work if the PolymerTemplate is within a component with a shadow root.
### Description of the bug
When building apps using `PolymerTemplate`, `polymerTemplate.setVisible(false)` does not work. Normally `setVisible(false)` applies the `hidden="true"` attribute, and Vaadin has the generic CSS rule available `:host([hidden]) { display: none !important; }` which hides the component. Things break if the component you try to hide is within another component that uses a shadow root. The attribute is added correctly, but the CSS is not applied. If you have for example a VerticalLayout within a component that has a shadow root, then setVisible() works correctly, so the bug has so far only appeared with PolymerTemplate

### Expected behavior
VisibilityBug.java
```java
package com.vaadin.tutorial.crm.ui;

import com.vaadin.flow.component.button.Button;
import com.vaadin.flow.component.html.Paragraph;
import com.vaadin.flow.component.html.Span;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;

@Route("visibility-bug")
public class VisibilityBug extends VerticalLayout {

    boolean isVisible = true;

    public VisibilityBug() {
        VerticalLayout verticalLayout1 = new VerticalLayout(new Span("Hello VerticalLayout"));
        MyTemplateOuter template1 = new MyTemplateOuter();

        Button button1 = new Button("Toggle visibility when no shadow doms are present", e -> {
            isVisible = !isVisible;
            verticalLayout1.setVisible(isVisible);
            template1.setVisible(isVisible);
        });
        add(button1, verticalLayout1, template1);

        add(new Paragraph());

        MyTemplateOuter template2 = new MyTemplateOuter();

        Button button2 = new Button("Toggle visibility when a shadow dom is present", e -> {
            isVisible = !isVisible;
            template2.setInnerVisibilities(isVisible);
        });
        add(button2, template2);
    }
}
```
MyTempalteOuter.java
```java
package com.vaadin.tutorial.crm.ui;

import com.vaadin.flow.component.dependency.JsModule;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.component.polymertemplate.Id;
import com.vaadin.flow.component.polymertemplate.PolymerTemplate;
import com.vaadin.flow.component.textfield.TextField;
import com.vaadin.flow.templatemodel.TemplateModel;

@Tag("my-template-outer")
@JsModule("./src/views/my-template-outer.js")
public class MyTemplateOuter extends PolymerTemplate<MyTemplateOuter.MyTemplateOuterModel> {

    @Id("vl") private VerticalLayout vl;

    @Id("inner") private MyTemplateInner innerTemplate;

    public MyTemplateOuter() { }
    public interface MyTemplateOuterModel extends TemplateModel {}
    public void setInnerVisibilities(boolean visible){
        vl.setVisible(visible);
        innerTemplate.setVisible(visible);
    }
}
```
MyTemplateInner.java
```java
package com.vaadin.tutorial.crm.ui;

import com.vaadin.flow.component.dependency.JsModule;
import com.vaadin.flow.component.polymertemplate.PolymerTemplate;
import com.vaadin.flow.templatemodel.TemplateModel;

@Tag("my-template-inner")
@JsModule("./src/views/my-template-inner.js")
public class MyTemplateInner extends PolymerTemplate<MyTemplateInner.MyTemplateInnerModel> {
    public MyTemplateInner() {}
    public interface MyTemplateInnerModel extends TemplateModel {}
}
```

my-template-outer.js
```js
import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
import './my-template-inner.js';
import '@vaadin/vaadin-ordered-layout/src/vaadin-vertical-layout.js';

class MyTemplateOuter extends PolymerElement {

    static get template() {
        return html`
<style include="shared-styles">
                :host {
                    display: block;
                    height: 100%;
                    border: 1px solid black;
                    padding: 10px;
                }
            </style>
            <vaadin-vertical-layout id="vl">
              <span>Hello inner VerticalLayout</span>
            </vaadin-vertical-layout>
            <my-template-inner id="inner"></my-template-inner>
`;
    }

    static get is() {
        return 'my-template-outer';
    }

    static get properties() {
        return {
            // Declare your properties here.
        };
    }
}
customElements.define(MyTemplateOuter.is, MyTemplateOuter);
```

my-template-inner.js
```js
import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';

class MyTemplateInner extends PolymerElement {

    static get template() {
        return html`
<style include="shared-styles">
                :host {
                    display: block;
                    height: 100%;
                }
            </style>
<div>Hello template inner</div>
`;
    }

    static get is() {
        return 'my-template-inner';
    }

    static get properties() {
        return {
            // Declare your properties here.
        };
    }
}

customElements.define(MyTemplateInner.is, MyTemplateInner);
```

### Actual behavior
A PolymerTemplate stays visible after `setVisible(false)` if that PolymerTemplate is within a shadow root. `hidden="true"` is applied but not the CSS

vaadin-template-visibility

### Expected behavior
A component that you call `setVisible(false)` on should be hidden

### Versions:
    - Vaadin / Flow version: 14.1.21
    - Java version: Corretto 11
    - OS version: Windows 10

joheriks:
Fixes #8256.


flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeProperties.java
```diff
@@ -87,6 +87,14 @@
      */
     public static final String VISIBILITY_HIDDEN_PROPERTY = "hidden";
 
+    /**
+     * The property used on the client side only in addition to
+     * {@link #VISIBLE}. It stores the client side value of the CSS "display"
+     * property to be able to restore when making a hidden element visible
```
denis-anisimov, line 93: Since there is a code which checks for shadow root would be good to mention it also here.

flow-tests/test-root-context/frontend/lit/lit-template-inner.js
```diff
@@ -0,0 +1,12 @@
...
+class TemplateInner extends LitElement {
+
+  render() {
+    return html`
+        <div>Hello template inner</div>
+    `;
+  }
+}
+
+customElements.define('lit-template-inner', TemplateInner);
```
denis-anisimov, line 12: new line

denis-anisimov (changes requested):
There are some GWT unit tests for visibility.

Please review them whether they need to be extended.
Some tests are for the visibility feature (bind/unbind) mostly.

But other tests are for hidden attribute behavior.

E.g. 
`testBindHiddenElement_stateNodeIsVisible_elementStaysHidden`
`testBindHiddenElement_stateNodeChangesVisibility_elementStaysHidden`,
`testBindNotHiddenElement_stateNodeChangesVisibility_elementIsNotHidden`, ....

flow-client/src/main/java/com/vaadin/client/PolymerUtils.java
```diff
@@ -646,4 +646,20 @@ public static native void setProperty(Element element, String path,
          element.set(path, value);
      }-*/;
 
+    /**
+     * Returns true iff the element has a shadow root ancestor.
```
bogdanudrescu, line 650: double `f` type

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeProperties.java
joheriks: Done.

flow-tests/test-root-context/frontend/lit/lit-template-inner.js
joheriks: Done.

flow-client/src/main/java/com/vaadin/client/PolymerUtils.java
joheriks: Spelled out.

joheriks:
> There are some GWT unit tests for visibility.
> 
> Please review them whether they need to be extended.
> Some tests are for the visibility feature (bind/unbind) mostly.
> 
> But other tests are for hidden attribute behavior.
> 
> E.g.
> `testBindHiddenElement_stateNodeIsVisible_elementStaysHidden`
> `testBindHiddenElement_stateNodeChangesVisibility_elementStaysHidden`,
> `testBindNotHiddenElement_stateNodeChangesVisibility_elementIsNotHidden`, ....

I added an additional assertion that `style={display:none}` is not added in existing tests checking attribute `hidden`. This  protects against regressions where it would be added unncessarily (for elements not in shadow root). Not sure how to set up a GWT test for the shadow root case.

flow-client/src/main/java/com/vaadin/client/PolymerUtils.java
```diff
@@ -646,4 +646,20 @@ public static native void setProperty(Element element, String path,
...
      }-*/;
 
+    /**
+     * Returns true if and only if the element has a shadow root ancestor.
+     * 
+     * @param element
+     *            the element to test
+     * @return whether the element is in a shadow root
+     */
+    public static native boolean isInShadowRoot(Element element)
```
denis-anisimov, line 656: >I added an additional assertion that style={display:none} is not added in existing tests checking attribute hidden. This protects against regressions where it would be added unncessarily (for elements not in shadow root). Not sure how to set up a GWT test for the shadow root case.

I think we need more GWT tests for this.
It's possible to emulate code with shadow root with this code or change one.
You may create Elements hierarchy so that one element in the hierarchy returns '[object ShadowRoot]'.
Just fake it: create a fake `toString`  method and return this string from it.

Another way could be reimplement this method a bit so that it doesn't use `toString` but check whether an element has a parent whose `shadowRoot`  property is set to the element.
Then it's also possible to fake `shadowRoot`.

We have quite many tests like that which fakes elements.
See e.g. `GwtBasicElementBinderTest`: `initPolymer` , `addShadowRootElement`, etc.

With mocking/faking it should be possible to write the same tests for `display` style as we already have for `hidden`.

denis-anisimov (changes requested):

flow-client/src/main/java/com/vaadin/client/PolymerUtils.java
joheriks: Ok, thanks for the hints. Added two unit tests.

denis-anisimov (approved):

Issue #9012, opened 2020-09-15
Initial value of the value attribute is not handled properly from templates


pleku:
Considering template with contents
```html
<vaadin-text-field id="textField" value="foobar"></vaadin-text-field>
<input id="inputField" value="foobar"></input>
```
and the corresponding Java template
```java
@Id("textField")
TextField textField;

@Id("inputField")
Input inputField;

inputField.getValue(); // returns ""
textField.getValue(); // returns ""
```
Probably related to how setting the `value` property directly is not taken into account in `AbstractFieldSupport` or `AbstractSinglePropertyField` . Whatever it is, we need to handle this properly even though giving a "default value for value" in the template seems like something that is not commonly done... it was discovered quickly when testing though.

denis-anisimov:
This is  the bug:

```
Element element = new Element("vaadin-text-field");
        element.setProperty("value", "foo");
        UI ui = new UI();
        UI.setCurrent(ui);
        VaadinSession session = Mockito.mock(VaadinSession.class);
        ui.getInternals().setSession(session);
        VaadinService service = Mockito.mock(VaadinService.class);
        Mockito.when(session.getService()).thenReturn(service);

        Instantiator instantiator = Mockito.mock(Instantiator.class);

        Mockito.when(service.getInstantiator()).thenReturn(instantiator);

        Mockito.when(instantiator.createComponent(TextField.class))
                .thenAnswer(a -> new TextField());
        TextField field = Component.from(element, TextField.class);
        Assert.assertEquals("foo", field.getValue());
```

denis-anisimov:
`AbstractSinglePropertyField` uses the default value (via `AbstractFieldSupport` ) as a value for the field in its CTOR
disregarding a fact that the underlying `Element`  already may have a value for the property
(there can be an `Element`  which is not a new but already existing one for which component is created).

So first of all `AbstractSinglePropertyField` CTOR should set the `value` based on underlying `Element`  property value if it exists overriding the default value.

This should be done in Flow.

But this is far away from the end of the story.
`TextField`  uses `GeneratedVaadinTextField` CTOR with an empty string as initial value.
And here is the code in the generated CTOR:
```
if (initialValue != null) {
            setModelValue(initialValue, false);
            setPresentationValue(initialValue);
        }
```

So text field is initially designed so that it doesn't care about any value in underlying Element and always sets the empty string as an initial value.
[...]

PR #9031, opened 2020-09-18
Be aware of property value in the Element

denis-anisimov:
Part of #9012

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/AbstractSinglePropertyField.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/AbstractSinglePropertyField.java
@@ -22,7 +22,6 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 
 import com.googlecode.gentyref.GenericTypeReflector;
-
 import com.vaadin.flow.dom.DomListenerRegistration;
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.dom.PropertyChangeEvent;
@@ -242,6 +241,12 @@ public class AbstractSinglePropertyField<C extends AbstractField<C, T>, T>
         propertyReader = typeHandler.createReader(element, propertyName,
                 presentationToModel);
 
+        // if underlying element already has a property value then it should be
+        // set for the component
+        if (getElement().getPropertyRaw(propertyName) != null) {
+            doSetModelValue(false);
+        }
+
         doSetSynchronizedEvent(
                 SharedUtil.camelCaseToDashSeparated(propertyName) + "-changed");
     }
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/component/AbstractSinglePropertyField.java
```diff
@@ -242,6 +241,12 @@ public AbstractSinglePropertyField(String propertyName, T defaultValue,
         propertyReader = typeHandler.createReader(element, propertyName,
                 presentationToModel);
 
+        // if underlying element already has a property value then it should be
+        // set for the component
+        if (getElement().getPropertyRaw(propertyName) != null) {
+            doSetModelValue(false);
```
pleku, line 247: Is this really the right fix as the `AbstractField` creates the `AbstractFieldSupport` which sets the given default value to `defaultValue` and `bufferedValue`. I would expect it to be fixed there by setting something else as the initial value.
Otherwise any component that extends `AbstractField` but not `AbstractSinglePropertyField` would be broken.

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/AbstractSinglePropertyField.java
denis-anisimov: Every component which extends `AbstractField` should be aware of this thing in the same way as `AbstractSinglePropertyField`  aware in this PR.
In this sense any component which  extends `AbstractField` directly is not better anyhow than `AbstractSinglePropertyField` .

The problem with `AbstractField` is : it doesn't now anything about property.
Only `AbstractSinglePropertyField` has knowledge about the property so it's possible to get `getElement().getProperty(propertyName)`.
As I understand `AbstractField` is on another layer of abstraction : it cares only about model nature of the field.
I don't have deep knowledge of the overall design `AbstractField`/`AbstractSinglePropertyField` but it seems `AbstractField` is intentionally designed in this way so that it's simply impossible to fix this issue anyhow in `AbstractField`: the abstraction layer doesn't allow to do this.
But as I said: don't have deep knowledge in this area.

I don't see any way how this issue can be fixed in `AbstractField`: if you see this way please let me know.

About the right fix: I think it's good enough: `doSetModelValue` sets the model value regardless of default value based on the property value.
This is proven by the unit test.

pleku: >The problem with AbstractField is : it doesn't know anything about property.

Right. AF is mostly for server side state and it is a layer that doesn't know anything about the value that goes to client side. It might contribute to a similar issue in the field components that extend it, when there is maybe multiple elements or components that compose the value to the client side, but it is the up for those classes to fix it so that the value is jointly read from the template as an attribute.

pleku (approved):

PR #9032, opened 2020-09-18
chore: fix typo

haijian-vaadin:


Haprog (approved):
Looks good assuming it compiles and tests pass (which should prove that rename was complete)

PR #9033, opened 2020-09-18
Force size re-estimate when requested range end equals assumed size

mshabarov:
Since the `Range` is exclusive at the end, it may happen that the range end is the same as the assumed size. It that case the `flush()` would not be invoked, although it should. Requested range strategy generally speaking depends on web component and there may be the case when the component requests, for example [5700...5800] and the assumed size is already 5800. This situation reflects the case when the user scroll to the end and then he jumps to undefined size. Then it should force the size re-estimation. And thus this code change occurred.

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
@@ -541,7 +541,7 @@ public class DataCommunicator<T> implements Serializable {
              * estimated size. If there was a previous defined size used, then
              * that is kept until a reset occurs.
              */
-            if (requestedRange.contains(assumedSize)) {
+            if (requestedRange.contains(assumedSize - 1)) {
                 requestFlush();
             }
         }
--- a/flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
+++ b/flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
@@ -33,6 +33,7 @@ import com.vaadin.flow.component.UI;
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.function.SerializableConsumer;
 import com.vaadin.flow.internal.Range;
+import com.vaadin.flow.internal.StateNode;
 import com.vaadin.flow.server.VaadinRequest;
 import com.vaadin.flow.server.VaadinService;
 import com.vaadin.flow.server.VaadinSession;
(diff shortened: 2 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -541,7 +541,7 @@ public void setDefinedSize(boolean definedSize) {
              * estimated size. If there was a previous defined size used, then
              * that is kept until a reset occurs.
              */
-            if (requestedRange.contains(assumedSize)) {
+            if (requestedRange.contains(assumedSize - 1)) {
```
pleku, line 544: Good catch but where is the unit test ?

pleku (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: Unit test added

flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
```diff
@@ -1072,6 +1073,30 @@ public void itemCountChangeEvent_estimatedCount_estimateUsedUntilEndReached() {
...
+    public void setDefinedSize_rangeEndEqualsAssumedSize_flushRequested() {
+        // trigger client communication in order to initialise it and avoid
+        // infinite loop inside 'requestFlush()'
+        fakeClientCommunication();
+
+        StateNode stateNode = Mockito.spy(element.getNode());
+        DataCommunicator<Item> dataCommunicator = new DataCommunicator<>(
+                dataGenerator, arrayUpdater, data -> {}, stateNode);
+        dataCommunicator.setPageSize(pageSize);
+        dataCommunicator.setDataProvider(createDataProvider(), null);
```
pleku, line 1086: Could have been nice to add a comment here that states that `// the size will be 100` so it makes it very clear how this test works

flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
```diff
@@ -1072,6 +1073,30 @@ public void itemCountChangeEvent_estimatedCount_estimateUsedUntilEndReached() {
...
+        fakeClientCommunication();
+
+        dataCommunicator.setRequestedRange(0, 100);
+        // clean flushRequest
+        fakeClientCommunication();
+
+        Mockito.reset(stateNode);
+        dataCommunicator.setDefinedSize(false);
+        // Verify that requestFlush has been invoked
+        Mockito.verify(stateNode).runWhenAttached(Mockito.anyObject());
```
pleku, line 1097: Like (most?) other tests here, shouldn't this very that the (estimated) count is increased properly too (which is the desired outcome in terms of internal DC logic) instead of just relying that call to flush has been done ?

pleku (changes requested):

flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
mshabarov: Done

flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
mshabarov: Done

pleku (approved):

PR #9035, opened 2020-09-18
Send changes to web component after size reset

mshabarov:
When setting a new count callback or new size estimation to a DataCommunicator, it should send the updates to a web component (client) even if the requested range not changed. For instance, if we change the count callback for ComboBox, it would be great to see the scroller size change immediately after open the dropdown. This one-liner code change provide this feature.

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
@@ -1001,7 +1001,7 @@ public class DataCommunicator<T> implements Serializable {
     private boolean collectChangesToSend(final Range previousActive,
             final Range effectiveRequested, Update update) {
         boolean updated = false;
-        if (assumeEmptyClient || resendEntireRange) {
+        if (assumeEmptyClient || resendEntireRange || sizeReset) {
             if (!assumeEmptyClient) {
                 /*
                  * TODO: Not necessary to clear something that would be set back
--- a/flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
+++ b/flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
@@ -118,7 +118,7 @@ public class DataCommunicatorTest {
         lastSet = null;
         lastUpdateId = -1;
 
-        update = new ArrayUpdater.Update() {
+        update = Mockito.spy(new ArrayUpdater.Update() {
 
             @Override
             public void clear(int start, int length) {
@@ -134,7 +134,7 @@ public class DataCommunicatorTest {
             public void commit(int updateId) {
                 lastUpdateId = updateId;
             }
-        };
+        });
 
         Mockito.when(arrayUpdater.startUpdate(Mockito.anyInt()))
                 .thenReturn(update);
@@ -369,7 +369,7 @@ public class DataCommunicatorTest {
     }
 
     @Test
-    public void setSizeCallback_usedForDataSize() {
+    public void setCountCallback_usedForDataSize() {
         AbstractDataProvider<Item, Object> dataProvider = createDataProvider();
         dataProvider = Mockito.spy(dataProvider);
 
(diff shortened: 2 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1001,7 +1001,7 @@ private void passivateInactiveKeys(Set<String> oldActive, Update update,
     private boolean collectChangesToSend(final Range previousActive,
             final Range effectiveRequested, Update update) {
         boolean updated = false;
-        if (assumeEmptyClient || resendEntireRange) {
+        if (assumeEmptyClient || resendEntireRange || sizeReset) {
```
pleku, line 1004: Unit test ?

pleku (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: Done

pleku, line 1004: From commit message:
>  it should send the updates to a web component (client) even if the requested range not changed

What does this mean ? What do you refer with the updates here ?
Changing the row count (to defined, to new estimate) should cause new size to be reflected on the client side immediately after response. I believe there are tests (unit, ITs) for this at least for grid.

Is it so that without this change the combo box with dropdown closed keeps some old cached size (or ?) on the client side until it is opened ? Or what is the behavior and how this fixes that ? Please clarify with comments on the code

flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
```diff
@@ -452,6 +452,27 @@ public void setCountCallback_itemCountEstimatesWereSet_overridesItemCountEstimat
                 dataCommunicator.getItemCount());
     }
 
+    @Test
+    public void setCountCallback_changesCommitted() {
```
pleku, line 456: What changes would this be ? Really hard to figure that out from looking at the code what it means

pleku (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: The scenario I intended to improve is the following:
1. Initially set the estimated size of combo box to 1000
2. Verify the size is actually 1000
3. Set estimated size (`dataView.setItemCountEstimate()`) to 2000
4. Verify the size (via checking `filteredItems` property of combo box). On this step verification is failed because updated `assumedSize` hasn't been sent to the client, because when we open the combobox then the same requested range [0..50] is being sent to server and `previousActive` and `effectiveRequested` are also [0..50], so the server side does not trigger `commit` action.

BUT, I've double check the behaviour for grid and it looks the same, however the size is updated on client. Then I noticed this difference in `UpdateQueue` implementation between Grid and ComboBox:

`getElement().setProperty("size", size);`

ComboBox doesn't have this line of code, and that's probably the reason why the size update does not go to client.
I don't completely understand why we need it though, but tests are passed except 1 unit test `loadItems_addItemsToCrossPageSize_switchToServerSideFiltering`. However, I believe this is something I need to investigate.

So this PR is just a 'crutch' and apparently should be rejected.

pleku: I agree, as it sounds to me that the data communicator should not "do extra" for this as the requirement is specific to one component only. But it is another question if  `Update` interface should be somehow aware of the item count and its changes ? The API does not have anything related to that though.

I can't say which is better way forward. a) Provide a real API for the size changes via DataCommunicator that the components can utilize or b) just fix this for ComboBox.

pleku: Also, very interesting way to put it that this is PR is a "crutch" 😄

mshabarov: > But it is another question if Update interface should be somehow aware of the item count and its changes ? The API does not have anything related to that though.

I believe it's an issue of ComboBox and should be fixed there. Responsibility of `Update` is to collect the component's state (in terms of items) and send it to client via JS calls, I think it doesn't need to be somewhere else than inside `flush()` method.

> Also, very interesting way to put it that this is PR is a "crutch" 😄

It's a direct interpretation of Russian word that means a workaround, or something temporal or fixing the symptoms, not the root cause :)

PR #9036, opened 2020-09-18
feat: submit cached endpoint requests in window online listener

### LINKED ISSUE #8952: Vaadin for TypeScript: cached endpoint requests should be submitted when coming back online
Part of #8905

When offline, if an endpoint has the `Deferrable` annotation, the endpoint request will be cached locally (e.g. IndexedDB). When coming back online, the cached requests should be submitted by calling `client.call()`;

If a submission succeeds, the cached request should be removed.
If a submission fails, the cached request should stay in the cache.

haijian-vaadin:
Fixes #8952
- removed uuid and idb-keyval, use idb instead
- use indexeddb transaction to avoid duplicated submissions
- Add tests


platosha (dismissed):

Haprog (approved):
Just some minor code style inconsistencies but didn't see any blockers.

PR #9037, opened 2020-09-19
Init element based on template data after bidning to a component

### LINKED ISSUE #9012: Initial value of the value attribute is not handled properly from templates

denis-anisimov:
This is not really a fix for the main bug but it's rather a workaround
which fixes exactly #9012 but not it's main reason

Fixes #9012

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/component/littemplate/LitTemplateTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/littemplate/LitTemplateTest.java
@@ -22,9 +22,12 @@ import org.junit.Test;
 import org.mockito.Mockito;
 
 import com.vaadin.flow.component.Tag;
+import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.polymertemplate.Id;
+import com.vaadin.flow.di.Instantiator;
 import com.vaadin.flow.function.DeploymentConfiguration;
 import com.vaadin.flow.server.VaadinService;
+import com.vaadin.flow.server.VaadinSession;
 
 public class LitTemplateTest {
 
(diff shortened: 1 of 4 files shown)


flow-server/src/test/java/com/vaadin/flow/component/littemplate/LitTemplateTest.java
```diff
@@ -157,4 +189,27 @@ public void attachExistingElementWithChidlrenWithText_elementHasNoText() {
...
+        Mockito.when(service.getInstantiator()).thenReturn(instantiator);
+
+        Mockito.when(instantiator.createComponent(ComponentSetProperty.class))
+                .thenAnswer(invocation -> new ComponentSetProperty(service));
+
+        InjectComponentWhichSetProperty parent = new InjectComponentWhichSetProperty(
+                service);
+
+        Assert.assertEquals("bar",
+                parent.child.getElement().getProperty("foo"));
```
pleku, line 212: Should there be another test for something that uses an attribute (on the server side).
For instance, id attribute comes to mind if. Another one is style. I would expect it to be consistent with this (value in template wins over constructor). I would appreciate a test for these so that it is verified and we can say it is intentional later on.

pleku (changes requested):

flow-server/src/test/java/com/vaadin/flow/component/littemplate/LitTemplateTest.java
denis-anisimov: Failing IT test now revealed the unacceptable approach with this PR.
In fact this is shown by unit test already.

```
@Tag("component-set-property")
    public static class ComponentSetProperty
            extends PolymerTemplate<ModelClass> {
        public ComponentSetProperty() {
            super(new SimpleTemplateParser());
            getElement().setProperty("foo", "baz");
            getElement().setText("foo");
        }
    }
```

When I have this Java code I expect that the property and the text of the element will be set programmatically to the values which are used in the CTOR.
In fact with this PR template values (attribute `foo` and the template text) will be set.

This is extremely confusing (even though in fact this is what we are trying to achieve applying workaround for our components): this is a chicken and egg problem. Either we initialize `Element` first  but then `Component`  has to be aware of that in its CTOR or we create a `Component` first and then either `Element`  should be aware of already set properties (but this is exactly what we want to overwrite) or overwrite all values initialized by the `Component`.

This is not a solution at all: the `Component` CTOR should be able to do anything without any risk to be overwritten.
So the issue has to be fixed on the component side and the order of initialization can't be changed.

I'm closing this PR.

Issue #8433, opened 2020-05-27
No available routes after live-reloading with JRebel in Flow 3.2


joheriks:
To reproduce:

* Create a Vaadin 17 project (starter will do) and annotate a view with a `@Route("abc")`.
* Start the applicaton using JRebel agent (version 2020.2.1 or later) in dev mode.
* Go to some illegal route to see the error page with list of routes and verify that `abc` is among them.
* Modify any code (not `@Route` annotation) in the view to trigger JRebel hotswap.
* Wait until browser refresh automatically.
* Route `abc` is no longer in the list of available routes. However, accessing `abc` URL still works. 

Versions: 
* Vaadin 17-SNAPSHOT (Flow 3.2-SNAPSHOT)
* Java 11

Note: This bug does not occur in Vaadin 14.3 (Flow 2.3). There the route survives the hotswap cycle (and us also correctly updated if the `@Route` annotation value is modified.)

PR #9050, opened 2020-09-22
Clear routes.

bogdanudrescu:
Fixes https://github.com/vaadin/flow/issues/8433

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
@@ -62,12 +62,19 @@ public class ConfigureRoutesTest {
                 "", mutable.getTemplate(BaseTarget.class));
     }
 
-    @Test
+    @Test // #8433
     public void mutableConfiguration_clearRoutes() {
         ConfigureRoutes mutable = new ConfigureRoutes();
 
         assertSetRoutes(mutable);
 
+        // Bug 8433 was caused by an InvalidRouteConfigurationException raised
+        // as a result of the RouteModel not being cleared on
+        // ConfigureRoutes.clear(), thus perceiving reloaded routes as duplicate.
+        //
+        // Here we make sure that for a duplicate route the exception is raised,
+        // while after invoking clear(), it shouldn't be raised again when
+        // adding the same routes.
         try {
             mutable.setRoute("", BaseTarget.class);
             Assert.fail("Route is already registered");
--- a/flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
@@ -1,8 +1,5 @@
 package com.vaadin.flow.router.internal;
 
-import org.junit.Assert;
-import org.junit.Test;
-
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.router.BeforeEnterEvent;
@@ -10,6 +7,9 @@ import com.vaadin.flow.router.BeforeEvent;
 import com.vaadin.flow.router.ErrorParameter;
 import com.vaadin.flow.router.HasErrorParameter;
 import com.vaadin.flow.router.HasUrlParameter;
+import com.vaadin.flow.server.InvalidRouteConfigurationException;
+import org.junit.Assert;
+import org.junit.Test;
 
 
 public class ConfigureRoutesTest {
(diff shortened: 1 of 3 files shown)


flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
```diff
@@ -62,6 +62,56 @@ public void mutableConfiguration_canSetTargetRoute() {
...
 
+    @Test
+    public void mutableConfiguration_clearRoutes() {
+        ConfigureRoutes mutable = new ConfigureRoutes();
+
+        assertSetRoutes(mutable);
+
+        try {
+            mutable.setRoute("", BaseTarget.class);
+            Assert.fail("Route is already registered");
```
caalador, line 73: Failing should be due to route not registered yet as the exception is thrown for non unique route registration.

Also what does this actually test as the assert routes check that the route has been registered?
I would remove this part as it doesn't bring anything that is expected.

flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
```diff
@@ -62,6 +62,56 @@ public void mutableConfiguration_canSetTargetRoute() {
...
+        } catch (InvalidRouteConfigurationException e) {
+        }
+
+        try {
+            mutable.setRoute(":param", ParamTarget.class);
+            Assert.fail("Route is already registered");
```
caalador, line 79: Also remove this set as it doesn't bring anything that is not tested in assertSetRoutes.

caalador (changes requested):

flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
bogdanudrescu: We can remove this of course. I'd like though to keep it because that exception was somehow causing the bug itself, and the purpose of having it in the test is to make sure that the exception is raised for a duplicate route while after `clear` it doesn't. I can put this details in the code.

flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
bogdanudrescu: Same as above.

flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
bogdanudrescu: I've added the details in the code. Let me know please if you agree with this.

caalador: I would rather have these as their own tests e.g. duplicateRootPathRegistration_throwsException and duplicateParameterPathRegistration_throwsException. as then it's clear what is the expected outcome and having a test fail makes it clear. now it would seem that clearing routed doesn't work even if this wouldn't be the case.

caalador (changes requested):

flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
bogdanudrescu: Done. Thank you!

flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
```diff
@@ -62,6 +62,70 @@ public void mutableConfiguration_canSetTargetRoute() {
...
+    public void duplicateParameterPathRegistration_throwsException() {
+        ConfigureRoutes mutable = new ConfigureRoutes();
+
+        mutable.setRoute(":param", ParamTarget.class);
+        Assert.assertTrue(mutable.hasRouteTarget(ParamTarget.class));
+        Assert.assertEquals(ParamTarget.class, mutable
+                .getNavigationRouteTarget("123").getRouteTarget().getTarget());
+
+        try {
+            mutable.setRoute(":param", ParamTarget.class);
```
caalador, line 123: Instead of the try catch you a better choice would be to add as a common parameter
```
    @Rule
    public ExpectedException exception = ExpectedException.none();
```
and in the tests expecting exception have:
```
        exceptionRule.expect(InvalidRouteConfigurationException.class);
        exceptionRule.reportMissingExceptionWithMessage("Duplicate routes shouldn't be accepted.");
```

caalador (changes requested):

flow-server/src/test/java/com/vaadin/flow/router/internal/ConfigureRoutesTest.java
bogdanudrescu: But if doing this, how can we test on what line it was thrown from? Isn't this an expectation over the whole method just like `@Test(expected = InvalidRouteConfigurationException.class)`?

caalador: exceptionRule gives more possibilities than the expected, for instance the message that should be found
 `exceptionRule.expectMessage("Navigation targets must have unique routes, found navigation targets 'ParamTarget' and 'ParamTarget' with the same route.");` which would be good to check perhaps to see it fails for the correct version.


Also the cause can be checked with `exceptionRule.expectCause(CoreMatchers.isA(TheCauseClassWeWouldExpect.class))`

Also shouldn't the correct exception be `AmbiguousRouteConfigurationException` and not `InvalidRouteConfiguratioException`?

bogdanudrescu: Thank you. Please see if it's OK.

caalador (approved):

joheriks (approved):

PR #9052, opened 2020-09-23
Skip data provider verification in AbstractDataView if default one is used

### LINKED ISSUE #8583: AbstractDataView verification of data provider type is problematic
Currently the data provider type is verified in `verifyDataProviderType` which is called by the constructor of the `AbstractDataView`. This is problematic for `AbstractLazyDataView extends AbstractDataView` since the default data provider in `DataCommunicator` is an empty `ListDataProvider` and thus when the lazy data view is created before the data provider has been set, the verification fails.

One use case for this is at least when a someone wants to set the `SizeChangeListener` before the data source has been set.

While the verification could be delayed somehow, it probably would be even better to just make `DataCommunicator` have a `null` or a special type of data provider first, and then update the `AbstractDataView` to handle this properly (don't fail if no data provider is set).

Now the data views are not stored but created on demand, the verification will anyway happen quite often. But also it is a good idea to do the verification whenever some action is done in the data view.

mshabarov:
Fixes #8583

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
@@ -60,12 +60,15 @@ public abstract class AbstractDataView<T> implements DataView<T> {
         /*
          * Skip verification if the verified data provider has not been
          * initialized yet.
-         * This mainly refers to the following cases:
-         * 1. Component uses data communicator which initialises data provider
-         * lazily and meanwhile has a default empty one. Good example is a
-         * ComboBox.
-         * 2. Developer wants to set the ItemCountChangeListener before the data
-         * provider has been set.
+         *
+         * This mainly refers to the following cases: 1. Component uses data
+         * communicator which initialises data provider lazily and meanwhile has
+         * a default empty one. Good example is a ComboBox. 2. Developer wants
+         * to set the ItemCountChangeListener before the data provider has been
+         * set.
+         *
+         * NOTE: In-memory data view API is supported without explicitly setting
+         * the data provider.
          */
         if (isDataProviderInitialized(dataProviderType)) {
             verifyDataProviderType(dataProviderType);
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
@@ -56,13 +56,27 @@ public abstract class AbstractDataView<T> implements DataView<T> {
                 "DataProvider supplier cannot be null");
         this.dataProviderSupplier = dataProviderSupplier;
         this.component = component;
-        verifyDataProviderType(dataProviderSupplier.get().getClass());
+        final Class<?> dataProviderType = dataProviderSupplier.get().getClass();
+        /*
+         * Skip verification if the verified data provider has not been
+         * initialized yet.
+         * This mainly refers to the following cases:
+         * 1. Component uses data communicator which initialises data provider
+         * lazily and meanwhile has a default empty one. Good example is a
+         * ComboBox.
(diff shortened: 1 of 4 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -648,6 +648,30 @@ public void setPagingEnabled(boolean pagingEnabled) {
...
+     * {@link #setDataProvider(DataProvider, Object)}.
+     * <p>
+     * This type of data provider signals to data view to skip the data
+     * provider verification on data view instantiation phase, so as to
+     * support generic API for lazy loading case (add item count listener,
+     * for instance), even if no backend data provider set by user.
+     *
+     * @param <T1> item type
+     */
+    static final class EmptyDataProvider<T1> extends ListDataProvider<T1> {
```
pleku, line 665: This should be before instance variables, and any static variables too

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -648,6 +648,30 @@ public void setPagingEnabled(boolean pagingEnabled) {
...
+     * In-memory data provider with no items.
+     * <p>
+     * Data Communicator is initialised with this data provider by default
+     * until a new data provider is assigned with
+     * {@link #setDataProvider(DataProvider, Object)}.
+     * <p>
+     * This type of data provider signals to data view to skip the data
+     * provider verification on data view instantiation phase, so as to
+     * support generic API for lazy loading case (add item count listener,
+     * for instance), even if no backend data provider set by user.
```
pleku, line 661: NIT: it feels like this is documented now in two places, abstract data view and here, IMO this is a bit excessive and one would have sufficed with a comment `// see xyz` for the other place

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
```diff
@@ -56,13 +56,27 @@ public AbstractDataView(
...
+        final Class<?> dataProviderType = dataProviderSupplier.get().getClass();
+        /*
+         * Skip verification if the verified data provider has not been
+         * initialized yet.
+         * This mainly refers to the following cases:
+         * 1. Component uses data communicator which initialises data provider
+         * lazily and meanwhile has a default empty one. Good example is a
+         * ComboBox.
+         * 2. Developer wants to set the ItemCountChangeListener before the data
+         * provider has been set.
```
pleku, line 68: This documentation (and the other duplicate in the empty data provider) is missing the part that in-memory data view API is supported without explicitly setting the data provider. Since you're documenting the usage, might add that too

pleku (changes requested):
Small nits

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: I would better place it just before `SizeVerifier` declaration and after the instance variables, because IMO inner class declaration on the top of outer class decrease the readability, and I couldn't find the recommendations to place inner classes on the top (before variables declaration).

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
mshabarov: Done

mshabarov:
I've changed the EmptyDataProvider access modifier to public so that it can be used in components.

pleku (approved):

PR #9054, opened 2020-09-23
Make ElementInitializationStrategy serializable

mshabarov:


denis-anisimov (commented):
Why this PR is needed ?

The classes don't have to be serializable because they are short living and never stored anywhere.
The only benefit from this is : avoid hardcoding them in the serilalizable test. But this is questionable .

mshabarov:
Failed `ClassesSerializableTest` in combo box is the only driver for me to create this PR, however, you're right, it's better to add them into exclude list in `ClassesSerializableTest`.

PR #9057, opened 2020-09-24
Add default constructor to DefaultInstantiator

### LINKED ISSUE #8427: Make DefaultInstantiator extendable
Currently, if you extend `DefaultInstantiator` and register your new custom instantiator through SPI, the custom `Instantiator` must be instantiated through a no-arg constructor. This constructor will not have access to a `VaadinService` instance and thus the `private VaadinService service` member variable in the parent `DefaultInstantiator` can never be set. This, in turn, means the `init` method will return false and the custom Instantiator won't be picked up. If the `init` method is overridden to return true, the methods requiring the `VaadinService` instance will fail.

mvysny:
Fixes #8427

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/di/DefaultInstantiator.java
+++ b/flow-server/src/main/java/com/vaadin/flow/di/DefaultInstantiator.java
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.di;
 
+import java.util.Objects;
 import java.util.ServiceLoader;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Stream;
@@ -46,14 +47,25 @@ public class DefaultInstantiator implements Instantiator {
      *
      * @param service
      *            the service to use
+     * @deprecated the service is assigned in {@link #init(VaadinService)}. Use
+     * the zero-arg constructor.
      */
+    @Deprecated
     public DefaultInstantiator(VaadinService service) {
         this.service = service;
     }
 
+    /**
+     * Creates a new instantiator. The service is filled in later on via
+     * {@link #init(VaadinService)}.
+     */
+    public DefaultInstantiator() {
+    }
+
     @Override
     public boolean init(VaadinService service) {
-        return service == this.service;
+        this.service = Objects.requireNonNull(service);
+        return true;
     }
 
     @Override
--- a/flow-server/src/main/java/com/vaadin/flow/di/DefaultInstantiator.java
+++ b/flow-server/src/main/java/com/vaadin/flow/di/DefaultInstantiator.java
@@ -148,4 +148,13 @@ public class DefaultInstantiator implements Instantiator {
     private <T> T create(Class<T> type) {
         return ReflectTools.createInstance(type);
     }
+
+    /**
+     * Returns the Vaadin service set via {@link #init(VaadinService)}.
+     * @return the Vaadin service or null if {@link #init(VaadinService)}
+     * has not yet been called.
+     */
+    protected VaadinService getService() {
+        return service;
+    }
 }
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
@@ -381,8 +381,7 @@ public abstract class VaadinService implements Serializable {
      */
     protected Instantiator createInstantiator() throws ServiceException {
         return loadInstantiators().orElseGet(() -> {
-            DefaultInstantiator defaultInstantiator = new DefaultInstantiator(
-                    this);
+            DefaultInstantiator defaultInstantiator = new DefaultInstantiator();
             defaultInstantiator.init(this);
             return defaultInstantiator;
         });


denis-anisimov (changes requested):
This is extremely strait-forward fix which doesn't confirm the contract and I'm strongly against of it.

There are several issue with this PR:
- overall `init` method is evil which is used often in Vaadin code. The reason is extremely simple : anyone may call this method with any argument at any time. You may not force it to be called once via any contract. You may do this only in runtime and this is not really good but even this is not done here. CTOR is always almost the only correct way to go because you can't call it many times. If you need to initialize something then it has to be done via CTOR and there are extremely small amount of cases when this can't be done.
- previous `init`  method at least hadn't had the issue above: initialization has been done via CTOR and `init` had been used for checking whether the service is the expected one or not. Based on  that it returned a value. Now this contract is broken.
- The intention of having `init`  with the its contract was to filter out `Instantiator` instances based on their `service`  state (done in `VaadinService::loadInstantiators`). Now every found instantiator is just initialized with the instance unconditionally. Yes, this code is questionable because it's not clear how the  `Instantiator` will get the `service`  instance. But the contract is different than in this PR.

Overall : `init` should not be used for initialization ever (see explanation above).
May be `init`  should be just deprecated.
It seems that it plays a role of `isApplicable`  : I'm not sure whether it's needed or not in fact.

What I suggest at least: make a factory class which produces an `Instantiator` based on provided `VaadinService`. So there will be no need to call additional `init` method.
Factories should be instantiated instead of  `Instantiator`  directly via the default CTOR in `loadInstantiators`. Then factories should be used to create instances of  `Instantiator`  for the `VaadinService`.
Factory may have `isApplicable` method (though I'm not sure whether it has sense).
Factory doesn't have to produce anything : if it returns `null`  then skip it and go next.

mvysny:
Rework of the `init()` method or introducing `InstantiatorFactory` is outside of the scope of this PR. If need be, those things can be reworked in another PR. Until that's done, this PR can be used to make things less horrible.

denis-anisimov:
> Rework of the `init()` method or introducing `InstantiatorFactory` is outside of the scope of this PR. If need be, those things can be reworked in another PR. Until that's done, this PR can be used to make things less horrible.
> 
> But yes I agree, the current `init()` approach is horrible since it's not documented whether there's 1-1 relationship between the VaadinService and the Instantiator or not; also the threading safety is not discussed at all.

I'm totally fine to close this PR: I don't see how to fix the original issue properly without a factory.
But if you see a way please go on.

mvysny:
The original issue is solved by merging this PR. It's not a proper solution, but the underlying foundation is not proper either. Now we have two options: either to fix the foundation, or to accept the foundation as it is, and merge this PR.

denis-anisimov:
> The original issue is solved by merging this PR. It may not be a proper solution, but the underlying foundation is not proper either. Now we have two options: either to fix the foundation, or to accept the foundation as it is, and merge this PR.

This PR can't be merged because it breaks the contract. It fixes one thing via introducing another issues.
So it can't be accepted.

mvysny:
> The intention of having init with the its contract was to filter out Instantiator instances based on their service state (done in VaadinService::loadInstantiators). 

I disagree. The 'filter out Instantiator instances based on their service state' is nowhere to be found in the Instantiator javadoc. Moreover, the current Javadoc of `Instantiator.init()` says that

```
     * @param service
     *            the Vaadin service for which this instance is initialized
     * @return <code>true</code> if this instance should be considered as a
     *         candidate for usage for the provided service; <code>false</code>
     *         to opt-out from the selection process
```

which implies that the instantiator should either return false, or take the `service` parameter provided and work within the scope of that service.

denis-anisimov:
> > The intention of having init with the its contract was to filter out Instantiator instances based on their service state (done in VaadinService::loadInstantiators).
> 
> I disagree. The 'filter out Instantiator instances based on their service state' is nowhere to be found in the Instantiator javadoc. Moreover, the current Javadoc of `Instantiator.init()` says that
> 
> ```
>      * @param service
>      *            the Vaadin service for which this instance is initialized
>      * @return <code>true</code> if this instance should be considered as a
>      *         candidate for usage for the provided service; <code>false</code>
>      *         to opt-out from the selection process
> ```
> 
> which implies that the instantiator should either return false, or take the `service` parameter provided and work within the scope of that service.

This parameter description looks for me exactly as filtering behavior. 
> <code>true</code> if this instance should be considered as a
>      *         candidate for usage for the provided service; <code>false</code>
>      *         to opt-out from the selection process

What's more important : I see the code in the `loadInstantiators` which  exactly filters out found `Instantiator` instances. So if there are several instances and
only one (somehow) is able to handle the service object then the current code (without this PR) will choose the one and use it.
With this PR the code will throw an exception since two (and every) instances are able to handle the service.

I can't provide more arguments for you.
I'm as a reviewer blocks this PR.
You may find someone else who agree with you and who will disregard my review.

denis-anisimov:
I'm closing the PR since I've already made the proper design of instantiating `Instantiator`s: there is an `InstantiatorFactory` interface which should be implemented . There is no even need to register it anyhow : just have in in the app classpath.
Factory should create an `Instantiator` via provided `VaadinService`  parameter.
Sop there is no anymore need in default CTOR for `DefaultInstantiator` : the existing one may be used, just implement the factory,.

mvysny:
There is no `InstantiatorFactory` in Vaadin 14.5.1. @denis-anisimov is it planned to be back-ported to Vaadin 14.5.1?

denis-anisimov:
No.
It's in the `2.6` and won't be in `2.5` .

mvysny:
Good, that means that it will appear in Vaadin 14.6?

denis-anisimov:
yes

PR #9061, opened 2020-09-24
Restructure dev mode handler logic to avoid starting webpack twice

### LINKED ISSUE #8981: Two webpack-dev-servers started if the application is restarted during webpack startup
```
vaadin init double-webpack --latest
cd double-webpack
mvn install # Only to generate webpack.config.js
```

Add the snippet to the start of webpack.config.js
```
const fs = require("fs");
if (!fs.existsSync("target/classes/foo.txt")) {
  fs.writeFileSync("target/classes/foo.txt","Hello");
}
```

Run 
```
mvn
```

### Expected behavior
One `webpack-dev-server` process is running

### Actual behavior
Two `webpack-dev-server` processes are running

### Versions:
    - Vaadin version: 17
    - Java version: 11
    - OS version: macOS

denis-anisimov:
Fixes #8981

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
@@ -577,7 +577,7 @@ public final class DevModeHandler implements RequestHandler {
         boolean success = false;
 
         try {
-            success = doStartWebPack(config, webPackFiles, start);
+            success = doStartWebpack(config, webPackFiles, start);
         } finally {
             if (!success) {
                 removeRunningDevServerPort();
@@ -585,7 +585,7 @@ public final class DevModeHandler implements RequestHandler {
         }
     }
 
-    private boolean doStartWebPack(DeploymentConfiguration config,
+    private boolean doStartWebpack(DeploymentConfiguration config,
             Pair<File, File> webPackFiles, long start) {
         ProcessBuilder processBuilder = new ProcessBuilder()
                 .directory(npmFolder);
--- a/flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
@@ -15,9 +15,18 @@
  */
 package com.vaadin.flow.server;
 
-import javax.servlet.ServletOutputStream;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
+import static com.vaadin.flow.server.Constants.VAADIN_MAPPING;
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
```diff
@@ -188,8 +191,10 @@ public static DevModeHandler start(int runningPort,
                 || !configuration.enableDevServer()) {
             return null;
         }
-        atomicHandler.compareAndSet(null,
-                createInstance(runningPort, configuration, npmFolder, waitFor));
+        if (atomicHandler.get() == null) {
+            atomicHandler.compareAndSet(null, createInstance(runningPort,
```
joheriks, line 195: Minor: Because of the added null check (I assume to avoid calling `createInstance` in not null case as there is no lazy `compareAndSet` available), could the inner code be simplified to `atomicHandler.set(...)`

joheriks (approved):

flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
denis-anisimov: No, it can't be.
That's the point of thread safety.
The whole idea of using AtomicReference is to update the value safely and avoid extra action.
If two threads comes here then both threads may pass `atomicHandler.get() == null`  check.
Then one thread may set it first and another thread will also create a new instance and set it but it should be rejected sine there is already one.
The added check will work in many cases to avoid even `createInstance`   call . But in the worst case it will be called anyway twice (or even more).
The changes which I made inside `DevModeHandler`  logic allows to avoid side effects in this case.
But `atomicHandler.get() == null`  check already will work in most cases so there will be no even need to create another instance.
So in short : the added check is not necessary (and should not be necessary) . But it allows to avoid extra actions in many cases even though it's not a booletproof solution.

joheriks: Ok. I was under the impression that the `AtomicReference` is there to have thread safe access between calls to `DevModeHandler::start` and `DevModeHandler::stop`. Is there at all a possibility to have two simulatenous threads entering `DevModeHandler::start`?

Anyway, this is fine and yes probably safer.

denis-anisimov: Theoretically this method can be invoked from two different threads.
Practically: I don't know. This is not my code. So I don't know all the details behind decision to make it thread safe.
But this way it's thread safe without side effects.

joheriks (approved):

taefi:
Cherry-picked to `4.0` by @joheriks on 5.10.2020

PR #9070, opened 2020-09-28
Add data communicator API for disabling data provider interactions

mshabarov:


flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -530,8 +571,8 @@ public int getItemCountEstimateIncrease() {
     public void setDefinedSize(boolean definedSize) {
         if (this.definedSize != definedSize) {
             this.definedSize = definedSize;
-            countCallback = null;
-            skipCountIncreaseUntilReset = false;
+            this.countCallback = null;
+            this.skipCountIncreaseUntilReset = false;
```
pleku, line 575: Unrelated unnecessary changes

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -496,7 +537,7 @@ public void setItemCountEstimateIncrease(int itemCountEstimateIncrease) {
         }
         this.itemCountEstimateIncrease = itemCountEstimateIncrease;
         this.countCallback = null;
-        definedSize = false;
+        this.definedSize = false;
```
pleku, line 540: Unrelated unnecessary change (yes the `this.` for countCallback is unnecessary too)

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -341,36 +375,43 @@ public T getItem(int index) {
...
+                    throw new IndexOutOfBoundsException(String.format(
+                            "Requested index %d on empty data.", index));
                 } else if (index >= itemCount) {
                     throw new IndexOutOfBoundsException(String.format(
                             "Given index %d is outside of the accepted range '0 - %d'",
                             index, itemCount - 1));
                 }
             }
+
+            if (fetchDisabled) {
```
pleku, line 400: Why should this be blocked ??? Since this API is not directly exposed to users, like it should not, it feels weird to me that for some components I can do
```java
Component listing = createListing();
listing.setItems(query -> getStuff(query));
listing.getItem(0);
```
but not for all. So for Grid this would work and not for combobox (I guess) ??? That is confusing.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -648,15 +689,56 @@ public void setPagingEnabled(boolean pagingEnabled) {
...
      * query.
      *
+     * @throws IllegalStateException
+     *             if the calls to data provider are disabled.
+     *
      * @return the size of data provider with current filter
      */
     @SuppressWarnings({ "unchecked", "rawtypes" })
     protected int getDataProviderSize() {
+        if (fetchDisabled) {
```
pleku, line 734: Again, it would be weird for a data view API user that the same things work for another component, but not for another.

pleku (changes requested):
I think the implementation is done in a way that it will leak to the users (when using e.g. data view API) and causes errors that can confuse them. It should be basically about the component to be able to "delay" the first fetch to a certain point, but not cause errors for users.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: Reverted

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: Reverted

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: Excluded

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: Excluded

mshabarov:
> I think the implementation is done in a way that it will leak to the users (when using e.g. data view API) and causes errors that can confuse them. It should be basically about the component to be able to "delay" the first fetch to a certain point, but not cause errors for users.

I was thinking whether it's better to use the data communicator as a listener for 'switch to normal mode', but I think it is a over-complication and basically limited to 1 component so far, so I simply left it as it is (CTOR arg + getter and setter). Curious users could of course try them randomly, but I have no elegant protection for that.

pleku (changes requested):
It seems to me this PR has more unnecessary white space changes than the actual changes. I will not review it as such as it wastes my time

pleku:
https://www.jetbrains.com/help/idea/reformat-file-dialog.html

pleku (approved):

PR #9072, opened 2020-09-28
fix: fail access check when no csrf token in session

haijian-vaadin:
- use `request.getSesssion(false)` instead of `request.getSesssion()` so that it won't create a new session for access check.
- fix a bug that when there is no csrf token in the session (e.g. when session is expired), it's considered valid.
- allow empty csrf token in the request header when there is no session (e.g. using curl).

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
@@ -139,14 +139,14 @@ public class VaadinConnectAccessChecker {
         if (!xsrfProtectionEnabled) {
             return false;
         }
+
         HttpSession session = request.getSession(false);
-        String csrfTokenInHeader = request.getHeader("X-CSRF-Token");
         if (session == null) {
-            return csrfTokenInHeader != null;
-        } else {
-            String csrfTokenInSession = (String) session.getAttribute(VaadinService.getCsrfTokenAttributeName());
-            return csrfTokenInSession == null || !csrfTokenInSession.equals(csrfTokenInHeader);
+            return false;
         }
+
+        String csrfTokenInSession = (String) session.getAttribute(VaadinService.getCsrfTokenAttributeName());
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
```diff
@@ -138,9 +139,14 @@ private boolean requestForbidden(HttpServletRequest request) {
         if (!xsrfProtectionEnabled) {
             return false;
         }
-        String csrfToken = (String) request.getSession()
-                .getAttribute(VaadinService.getCsrfTokenAttributeName());
-        return csrfToken != null && !csrfToken.equals(request.getHeader("X-CSRF-Token"));
+        HttpSession session = request.getSession(false);
+        String csrfTokenInHeader = request.getHeader("X-CSRF-Token");
+        if (session == null) {
+            return csrfTokenInHeader != null;
```
platosha, line 145: This fails requests with a token header when there is no session. Any reasons behind it?

IMO it would read more logical if in case of empty session we return early before checking the header:

```java
HttpSession session = request.getSession(false);
if (session == null) {
  return false;
}

String csrfToken = (String) request.getSession();
return csrfToken == null || !csrfToken.equals(request.getHeader("X-CSRF-Token"));
```

platosha (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
haijian-vaadin: > This fails requests with a token header when there is no session. Any reasons behind it?
Isn't this exactly the case that the session has expired, and a cached request is being submitted (with an outdated csrf token)?

it could be fine when there is no csrf token in the request header, e.g. using curl.

platosha: If the session has expired, it should fail elsewhere in `isLoggedIn` kind of check.

platosha (approved):

Issue #3697, opened 2018-03-13
Flow doesn’t register a click for the <a> element when it happens on top of the nested <iron-icon> elements.


ZheSun88:
Not a certain issue, but according to feedback, it seems like Flow doesn’t register a click for the <a> element when it happens on top of the nested <iron-icon> elements.

This issue is originally from `Beverage-Starter-flow` with the navigate items' 'DeadZone' problem. (https://github.com/vaadin/beverage-starter-flow/issues/175) <- according to the comments, it reproduces every time

at the moment, there is a workaround for the starter to make it work https://github.com/vaadin/beverage-starter-flow/pull/185.

pleku:
This is a relevant case when using routerlink that has icons inside.

Legioth:
Could it be that the logic that detects `router-link` clicks doesn't take into account that the actual click target is inside a shadow root?

heruan:
I'm experience the same issue when `RouterLink` is inside a `ComponentRenderer`, e.g. in `IronList`. Could it be related to this?

pleku:
>I'm experience the same issue when RouterLink is inside a ComponentRenderer, e.g. in IronList. Could it be related to this?

Most likely the same case. Issue probably occurs either in https://github.com/vaadin/flow/blob/master/flow-client/src/main/java/com/vaadin/client/flow/RouterLinkHandler.java#L67 or https://github.com/vaadin/flow/blob/master/flow-client/src/main/java/com/vaadin/client/flow/RouterLinkHandler.java#L98.

heruan:
This project replicates the issue: https://github.com/heruan/flow-test/tree/router-link

There is a component with a red `div` in the shadow DOM and a green `div` in the light DOM: when clicking on the green one the link behaves as it should (i.e. in-app navigation), while clicking on the red one it triggers a full navigation instead.

pleku:
Since this is not reproducing with e.g. https://labs.vaadin.com/business/payments and there have not been any further reports about this, I don't think this issue is anymore valid. In case anyone runs into this, please create a new issue with the code/steps to reproduce the problem.

PR #9080, opened 2020-09-30
Fix router link with shadow dom.

bogdanudrescu:
Fixes https://github.com/vaadin/flow/issues/3697

### DIFF
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/RouterLinkIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/RouterLinkIT.java
@@ -1,10 +1,12 @@
 package com.vaadin.flow.uitest.ui;
 
+import com.vaadin.flow.testutil.TestBenchHelpers;
 import org.junit.Assert;
 import org.junit.Test;
 import org.openqa.selenium.By;
 
 import com.vaadin.flow.testutil.ChromeBrowserTest;
+import org.openqa.selenium.WebElement;
 
 public class RouterLinkIT extends ChromeBrowserTest {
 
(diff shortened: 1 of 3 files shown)


bogdanudrescu:
@pleku I'm not sure about `target/1.0` label, but I think that's where the bug was reported.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/RouterLinkIT.java
```diff
@@ -69,6 +71,35 @@ public void testImageInsideRouterLink() {
...
+
+        final WebElement routerLinkShadowComponent = routerLinkWithShadow
+                .findElement(By.tagName("shadow-component"));
+
+        testInsideServlet(
+                findInShadowRoot(routerLinkShadowComponent,
+                        By.id("router-link-with-shadow-red")).get(0),
+                "shadow", "shadow");
+
+        testInsideServlet(findElement(By.id("router-link-with-shadow-green")),
```
caalador, line 91: How does this not fail? At the step before we navigate away from the test view and the view should now be Shadow where we do not have other elements apart from the `<h1>shadow</h1>` .

This should be in it's own test case `customElementWithShadowRoot_clickLightDomSlottedElement_navigatedToShadowRoute` or the like.

caalador (changes requested):

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/RouterLinkIT.java
bogdanudrescu: I don't know how it's not failing, I didn't investigated that, but it doesn't, it behaves like all the other router links in the test. The text, without `<h1>` is embedded in the `<div id="location">`.

When testing without the change, a browser navigation will be triggered when clicking on the red div inside the shadow and it does navigate to another page, but still not from the green one.

caalador: Really? At least not when actually running in the browser and clicking the links. 
That would explain why the test works with the second validation, but it could still be its own test case.

This is again one of the stranger tests.

caalador (dismissed):

pleku:
Apparently this is not even a problem in e.g. https://labs.vaadin.com/business/payments and as the changeset is cannot be accepted as is, closing the PR.

PR #9091, opened 2020-10-01
fix: Rename method name to avoid double negation

mshabarov:
Part of https://github.com/vaadin/vaadin-combo-box-flow/issues/355

### DIFF
--- a/flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
+++ b/flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
@@ -1277,8 +1277,8 @@ public class DataCommunicatorTest {
     }
 
     @Test
-    public void fetchDisabled_getItemCount_stillReturnsItemsCount() {
-        dataCommunicator.setFetchDisabled(true);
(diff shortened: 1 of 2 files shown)


flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
```diff
@@ -1277,8 +1277,8 @@ public void fetchFromProvider_backendRunsOutOfItems_secondPageRequestSkipped() {
     }
 
     @Test
-    public void fetchDisabled_getItemCount_stillReturnsItemsCount() {
-        dataCommunicator.setFetchDisabled(true);
+    public void fetchEnabled_getItemCount_stillReturnsItemsCount() {
```
pleku, line 1280: The test name semantic is not maybe correct but it doesn't matter

pleku (approved):

pleku:
Technically this is not a `fix:` but a `refactor:` since this is fixing something that has not been released yet or doesn't even have a ticket open, so it should not be listed in release notes (`fix:` will be, `refactor:` not).

pleku (approved):

Issue #7799, opened 2020-03-11
Dialog doesn't prevent the execution of clickshortcuts behind it


knoobie:
When I create a view with a Shortcut Listener (Enter) on a button and the view opens a "Add XYZ Dialog" for some additional content - I would expect that I can create a second Shortcut Listener **within** the Dialog and the Dialog **prevents** that both Enter Shortcuts are executed. 

Only the Shortcut Listener within the Dialog should be executed. Currently **both** Shortcuts are executed.

This creates weird side effects. Additionally any Dialog should be modal (WCAG) - that mean's that interacting with the content behind it SHOULD BE impossible.

Vaadin: 14.1.19
Browser: FF 74

How to test:

1) Click Open Dialog
2) Press Enter
3) 2 Notification are displayed. It should only be one visible

```
@Route("shortcut")
public class ShortcutBug extends VerticalLayout {

  public ShortcutBug() {
    add(new Button("Open Dialog", e -> {
      Button myDialogButton = new Button("Hit Enter", event -> {
        Notification.show("Enter was called", 2000, Position.TOP_START).addThemeVariants(NotificationVariant.LUMO_SUCCESS);
      });
      myDialogButton.addClickShortcut(Key.ENTER);
      Dialog dialog = new Dialog(myDialogButton);
      dialog.open();
    }));

    Button myNonDialogButton = new Button("Should not be called by the other Other Enter Shortcut", e -> {
      Notification.show("ERROR: Enter was called", 5000, Position.TOP_START).addThemeVariants(NotificationVariant.LUMO_ERROR);
    });
    myNonDialogButton.addClickShortcut(Key.ENTER);

    add(new Hr(), myNonDialogButton);
  }
}
```

denis-anisimov:
I don't think this is a bug.
This works as designed even though I agree that the behavior could be better.
Here is the explanation:

- Flow allows to add the same shortcut to several components : that how it works (this can be an issue itself though).
- So they are global.
- `Dialog` is just a component for Flow the same like `Div` .
- If I have two buttons : one is on the page, another is inside `Div` on the page then it's expected (see above) that both of then will be triggered .
- As I said: for Flow there is no any difference between `Div` and `Dialog` .
- So it works as expected.

There can be a discussion whether we should trigger actions for every component with a shortcut (then cases when some components has assigned shortcuts but not attached should be considered).
But this discussion doesn't help here anyway since in this case both buttons presents on the page (the only difference that the one is in the dialog).

[...]

PR #9093, opened 2020-10-01
fix: Fix JavaScriptBootstrapUI::getChildren

### LINKED ISSUE #9069: JavaScriptBootstrapUI methods add, remove, getElement are not in sync with getChildren
This has been broken since Flow 3.0 due to introducing a wrapper element which is the "router outlet" with client side routing.
Thus any calls to `getUI().get().add(myComponent);` do not work when combined with `ui.getChildren();` or `ui.remove(myComponent);`

pleku:
It should return all components added to the UI, like Dialog and Notification.
Related to #7799

Fixes #9069


caalador (approved):

taefi:
cherry-picked to `4.0`

PR #9122, opened 2020-10-06
fix: Set disabled status for component mapped to an Element

### LINKED ISSUE #8973: Parse initial disabled attribute value from templates
Follow up on #8957, given a template with
```html
<div id="foo" disabled></div>
``` 
The initial attribute value should reflect the `enabled` state on server side for an `@Id` mapped component, like
```java
@Id("foo") Div fooDiv;

fooDiv.isEnabled(); // returns false initially
```
Toggling the disabled state initially for a component in a template can be considered a reasonably common thing to do via designer. The component is then later on enabled via server side Java logic.

denis-anisimov:
Fixes #8973

### DIFF
--- a/flow-tests/test-root-context/frontend/lit/DisabledLitTemplateElement.js
+++ b/flow-tests/test-root-context/frontend/lit/DisabledLitTemplateElement.js
@@ -1,5 +1,4 @@
 import {html, LitElement} from 'lit-element';
-import './lit-template-inner.js';
 
 class DisabledLitTemplateElement extends LitElement {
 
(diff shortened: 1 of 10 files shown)


flow-tests/test-root-context/frontend/lit/DisabledLitTemplateElement.js
```diff
@@ -0,0 +1,14 @@
+import {html, LitElement} from 'lit-element';
+import './lit-template-inner.js';
```
joheriks, line 2: Unnecessary import

joheriks (approved):

PR #9125, opened 2020-10-06
Re-implement the EventHandlerRpcHandler

### LINKED ISSUE #9116: Flow fails to bootstrap without the Polymer module
After Polymer was moved to it's own module Flow will fail to bootstrap with the client side bootstrap mode.

This can be replicated with the skeleton-starter by setting the project version to `18.0-SNAPSHOT` and excluding polymer from the vaadin dependency.
```
         <dependency>
            <groupId>com.vaadin</groupId>
            <!-- Replace artifactId with vaadin-core to use only free components -->
            <artifactId>vaadin</artifactId>
            <exclusions>
                <exclusion>
                    <groupId>com.vaadin</groupId>
                    <artifactId>flow-polymer-template</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

This will lead to the page staying blank after load.

In debug mode the message `Polymer handler is called even though the polymer module is not in the classpath` should be logged.

The Problem is that `PublishedServerEventHandlerRpcHandler` was seen as a polymer feature even though it is an element feature with some polymer template handling.

caalador:
EventHandlerRpcHandler handles the ClientCallable
server method and is thus not purely for polymer.
The hanlder was returned and tests were parted
between polymer (EventHandler) and server (ClientCallable)

Fixed tests to be more robust and also to work on Windows.

Closes #9116


flow-polymer-template/src/main/java/com/vaadin/flow/component/polymertemplate/rpc/PublishedEventRpcHandlerUtil.java
```diff
@@ -0,0 +1,79 @@
...
+     *         polymer template to get model from
+     * @param argValue
+     *         argument value
+     * @param convertedType
+     *         value type
+     * @return the provided model value
+     */
+    public static Object getTemplateItem(Component template,
+            JsonObject argValue, Type convertedType) {
+        StateNode node = template.getUI().get().getInternals().getStateTree()
```
denis-anisimov, line 70: This impl assumes that template is attached.
This should be explicitly checked via assertion e.g.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/PublishedServerEventHandlerRpcHandler.java
```diff
@@ -58,28 +68,307 @@ public String getRpcType() {
...
+            return findMethod(instance, clazz.getSuperclass(), methodName);
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    private static boolean hasMethodAnnotation(Method method) {
+        // Check for polymer event handler annotation.
+        final Optional<Annotation> eventHandler = ReflectTools
+                .getAnnotation(method,
```
denis-anisimov, line 170: There is a method `hasAnnotation` , no need to use `getAnnotation`.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/PublishedServerEventHandlerRpcHandler.java
```diff
@@ -58,28 +68,307 @@ public String getRpcType() {
...
+                                JsonValue.class, Class.class)
+                        .invoke(polymerUtil, instance, argValue,
+                                convertedType)) {
+                    return polymerUtil
+                            .getMethod("getTemplateItem", Component.class,
+                                    JsonObject.class, Type.class)
+                            .invoke(polymerUtil, instance, argValue,
+                                    method.getGenericParameterTypes()[index]);
+                }
+            } catch (ClassNotFoundException | NoSuchMethodException e) {
```
denis-anisimov, line 313: `ClassNotFoundException`  means there is no Polymer template in the classpath

`NoSuchMethodException`  means there is a mistake in the impl since polymer IS in the classpath but the method name used here doesn't match any method name in the module. It means that the method is either renamed or removed and this code has to be rewritten.

The `catch` should distinguish this two cases and ignore one case silently (log is enough) and do something more serious in the second case. May be just throw ISE since this is a mistake in the impl.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/PublishedServerEventHandlerRpcHandler.java
```diff
@@ -58,28 +68,307 @@ public String getRpcType() {
...
+                // NO-OP we just try to check if the polymer library is added
+                LoggerFactory
+                        .getLogger(PublishedServerEventHandlerRpcHandler.class)
+                        .trace("Tried finding polymer dependency classes.", e);
+            } catch (IllegalAccessException | InvocationTargetException e) {
```
denis-anisimov, line 318: `IllegalAccessException` : can't happen. If it happens then this is an impl mistake and it should throw probably.

`InvocationTargetException`  means there was some exception in calling the method (which is wrapped to this one because of reflection call).
This exception may not be silently ignored (or just logged). It should be thrown somehow.

denis-anisimov (changes requested):
As a note: this PR solves to unrelated issues at the same time (and has title for only one of them). It would be better to have one issue per PR. Though I don't insist here.

Another question: PublishedServerEventHandlerTest . Github shows that it has been moved but at the same time it shows some removals... Am I missing something ?
Is there any part which remained in polymer-template module  ?
There are only util methods in polymer-template module so the test would have been for the util static methods.
In `flow-server`  on the other hand it's possible to keep the tests as is.

caalador:
I moved the test fixes to its own PR, but it will need this to actually pass.

PublishedServerEventHandlerTest was actually split into 2 where the ClientCallable tests are in flow-server and EventHandler in polymer-template.

I'll add some tests for the util methods.

flow-polymer-template/src/main/java/com/vaadin/flow/component/polymertemplate/rpc/PublishedEventRpcHandlerUtil.java
```diff
@@ -0,0 +1,88 @@
...
+            if (ui.isPresent()) {
+                StateNode node = ui.get().getInternals().getStateTree()
+                        .getNodeById((int) argValue.getNumber("nodeId"));
+
+                ModelType propertyType = ((PolymerTemplate<?>) template)
+                        .getModelType(convertedType);
+
+                return propertyType.modelToApplication(node);
+            }
+            throw new IllegalStateException(
```
denis-anisimov, line 84: It might be that `IllegalArgumentException` is better here.
But I think it doesn't really matter.

flow-polymer-template/src/test/java/com/vaadin/flow/component/polymertemplate/rpc/PublishedEventRpcHandlerUtilTest.java
```diff
@@ -0,0 +1,222 @@
...
+        final Type messageType = ModelClass.class
+                .getMethod("setMessage", String.class)
+                .getGenericParameterTypes()[0];
+
+        final Object templateItem = PublishedEventRpcHandlerUtil
+                .getTemplateItem(instance, json, messageType);
+
+        // We test against the null node as it is impractical to get the
+        // full model setup in junit test.
+        Assert.assertNull("Null node should get a null template item", templateItem);
```
denis-anisimov, line 204: What should happen to fail the test ?

flow-polymer-template/src/test/java/com/vaadin/flow/component/polymertemplate/rpc/PublishedEventRpcHandlerUtilTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+        public List<PendingJavaScriptInvocation> dumpPendingJsInvocations() {
+            // Ensure element invocations are also flushed
+            getInternals().getStateTree().runExecutionsBeforeClientResponse();
+
+            return getInternals().dumpPendingJavaScriptInvocations();
+        }
+    }
+
+}
```
denis-anisimov, line 222: new line

denis-anisimov (changes requested):

flow-polymer-template/src/test/java/com/vaadin/flow/component/polymertemplate/rpc/PublishedEventRpcHandlerUtilTest.java
caalador: Removed the test as it makes no sense and I doen't see how it should work as a junit test

flow-polymer-template/src/main/java/com/vaadin/flow/component/polymertemplate/rpc/PublishedEventRpcHandlerUtil.java
caalador: changed

denis-anisimov (approved):

denis-anisimov (approved):

PR #9129, opened 2020-10-07
chore: Fix tests for windows execution

caalador:
Fix tests so they work also on windows.
Remove Polymer dependency form the test
root and only have it in packages actually
utilizing Polymer.


flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -217,10 +215,10 @@ private static Document getIndexHtmlDocument(VaadinService service)
         String frontendDir = FrontendUtils.getProjectFrontendDir(
                 service.getDeploymentConfiguration());
         String indexHtmlFilePath;
-        if(frontendDir.endsWith(File.separator)) {
+        if(frontendDir.endsWith("/")) {
```
pleku, line 218: To be honest, I don't even get how this helps - is it not going through the file system ? Maybe a clarifying comment so someone doesn't "fix this" again

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
caalador: I actually fixed this now in the test expectation.
The problem is ofc that for instance the deafult frontendDir is `./frontend/`
which will not match the windows file separator and thus is will become `./frontend/\index.html`
This looks strange, but as it's only an exception the work to handle the path correctly would be wasted.

pleku (approved):

PR #9133, opened 2020-10-07
fix: store app shell title into UI instead of AppShellRegistry

haijian-vaadin:
The app shell title is currently stored in `AppShellRegistry` which makes it a singleton. The fix is to store it into `UI` instead.

- Normally the app shell title is passed from client to server via`flowRoot.$server.connectClient(this.container.localName, this.container.id, this.getFlowRoute(ctx), document.title);` in `Flow.ts`
- When `eagerServerLoad` is set, the app shell title is passed via `IndexHtmlHandler`, since the UI is already available then and `document.title` is empty.


platosha (approved):

flow-server/src/main/java/com/vaadin/flow/server/AppShellRegistry.java
```diff
@@ -162,15 +160,6 @@ public boolean isShell(Class<?> clz) {
         }
     }
 
-    /**
-     * Return the text content of the title tag in the application shell.
-     *
-     * @return title;
-     */
-    public String getTitle() {
```
platosha, line 170: Note: this is technically a breaking change now that `AppShellRegistry#getTitle()` is removed.

Should we deprecate it instead for now?

platosha (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/AppShellRegistry.java
Haprog: I'm not sure. If we deprecate it, do we also need to keep/add logic so it returns what it used to?

I guess it would be good to have deprecation before removal, but I'm not sure if it's ok to just remove in this case.

Haprog: For backporting the fix we probably should not remove it though.

Artur-:
Isn't UI the wrong scope for this, just as AppShellRegistry? I page title is tied to a browser tab or a view. A UI instance is shared among many browser tabs. Also if you have a TS only app, you should not have any UI instance at all

haijian-vaadin:
> If you have a TS only app, you should not have any UI instance at all
The purpose of storing this app title into UI is that when navigating to a server-side view that doesn't have a title, then the app title could be used as a fallback one.
If you have a TS only app, then the title is not stored/used.

Artur-:
What is an "app title"? What is the "app title" if you navigate from a server side view to a server side view without a title? What is the "app title" if you directly navigate to a server side view without a title?

pleku:
> What is an "app title"?

"App title" is what is referred to here as the title set via the `AppShell` or `AppShellConfigurator`.

> What is the "app title" if you navigate from a server side view to a server side view without a title?

The title will be either the "app title" or if there is none set, the title is cleared to leave the browser to decide what is shown.
So if the previous server side view had a specific title like "Login", that would always change when the user navigates away from it. Either to 1) new specific view title 2) app title 3) cleared completely.

That is how it is intended to work, though there is no design for this (?) in any older issue. Not at least in the one that introduced the current issue.

haijian-vaadin:
> What is an "app title"? What is the "app title" if you navigate from a server side view to a server side view without a title? What is the "app title" if you directly navigate to a server side view without a title?
"app title" means the value of the <title> tag in the bootstrap page. However you navigate to a server-side view without a title, the <title> tag in the bootstrap page would be used.

flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts
```diff
@@ -170,7 +170,7 @@ export class Flow {
 
       // Call server side to navigate to the given route
       flowRoot.$server
-          .connectClient(this.container.localName, this.container.id, this.getFlowRoute(ctx));
+          .connectClient(this.container.localName, this.container.id, this.getFlowRoute(ctx), document.title);
```
pleku, line 173: Isn't this too late ? If a client side view sets another title instead of the one provided by the `AppShell` from Java, then that title would be set as the "app title" instead ? So basically the `document.title` should be captured somewhere after the initial index page with the appshell provided title comes (like I said yesterday in the call). Or what am I missing ?

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/AppShellRegistry.java
haijian-vaadin: Yep, good point, marked it as `@Deprecated` now

flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts
haijian-vaadin: very good catch, sorry that I forgot about it while implementing.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -126,6 +130,16 @@ public boolean synchronizedHandleRequest(VaadinSession session,
         return true;
     }
 
+    private void storeAppShellTitleToUI(Document indexDocument) {
+        if(UI.getCurrent()!=null){
```
pleku, line 134: a) Isn't this always `null` unless the legacy bootstrapping is used ? If the code calling this method cannot be moved inside the if-else that checks that checks the bootstrapping mode, a comment could point here that "this is only for the legacy bootstrapping"
b) IMO `UI.getCurrent().ifPresent(ui -> ...` would be cleaner code
c) this code is not formatted

flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts
```diff
@@ -87,7 +88,7 @@ export class Flow {
         // IE11 does not support document.baseURI
         (document.baseURI || elm && elm.href || '/')
             .replace(/^https?:\/\/[^\/]+/i, ''));
-
+    this.appShellTitle = document.title;
```
pleku, line 91: This code / feature is not tested anyway, so it can be accidentally removed and nobody will notice the regression before a user reports it. So an IT should be IMO added

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/AppShellRegistry.java
haijian-vaadin: This PR should be cherry-picked to 4.0, so I think maybe we keep it as Deprecated in this PR, so we can cherry-pick as is to 4.0, then make another PR in master to remove it.
I will update the PR to throw an exception instead.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
haijian-vaadin: It's also won't be null when [eagerServerLoad](https://github.com/vaadin/flow/blob/2fa75ddc88cdb719daaceb8ef21478b9acb43c37/flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java#L372) is configured.

I will update the code for other suggestions.

flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts
haijian-vaadin: Good point, will add one

pleku (approved):

flow-tests/test-ccdm/src/test/java/com/vaadin/flow/ccdmtest/NavigateToServerSideViewWithoutTitleIT.java
```diff
@@ -0,0 +1,52 @@
...
+        Assert.assertEquals("Shoul use app shell title", "app-shell-title",
+                getDriver().getTitle());
+    }
+
+    @Test
+    public void should_showAppShellTitle_after_titleHasBeenChangedOnTheClientSide_when_navigateToAServerSideViewWithoutTitle() {
+        openVaadinRouter();
+
+        // update document.title on the client-side
+        findElement(By.id("updatePageTitle")).click();
```
pleku, line 41: Slightly nit picking and thus not blocking: could verify that the title has been changed here in the middle, since if it for some reason would not be changed from the initial, then this test would not add anything extra to the previous test

pleku (approved):

platosha (approved):

Issue #9140, opened 2020-10-08
Redesign/reimplement/refactor element<->component mapping


denis-anisimov:
`Component::from` and it's delegate `ComponentUtil::componentFromElement`  provides a way to create a component
to an existing `Element` with a "transparent" way: so if there is a default CTOR for the `Component` then it will be instantiated with an existing  given `Element` instead of a new one.

That feature doesn't introduce any contract : the `Component`  implementor is not aware of the fact that the CTOR may be called with already initialized `Element`  and it's necessary to care about this fact.

We have many issues with that. See detailed explanation here: https://github.com/vaadin/flow/issues/9136#issuecomment-705401784.

This feature should be reimplemented somehow so that the contract is absolutely clear and there is no any magic/hidden things that implementor should know to be able to make `ComponentUtil::componentFromElement` works properly.

The feature is used every time for `@Id` injected elements in templates and every time when a component is injected via `@Id`  it's mapped via an existing `Element`. As a result at the moment the implementor have to care about two cases: normal workflow and the case when the component is mapped (via `@Id` ) e.g.
But this requirement is totally hidden.

- In the best case the implementor should not care about several cases (this is done already now but it's done in a wrong way: contract is the same but in fact it's broken because there are cases when you have to care about unknown special cases).
- If it's not possible then the contract has to be absolutely clear that there are special cases.

Though I think the second case will be quite (or even extremely) inconvenient.
But I'm not sure whether it's possible to solve the issue via the first case.

PR #9141, opened 2020-10-08
refactor: revert Set disabled status for component mapped to an Element (#9122)

### LINKED ISSUE #9136: Regression:  #9122 breaks Grid init
vaadin-flow-components master build is failing after flow-5.0-SNAPSHOT was updated.
git bisect reported that the culprit is a fix for disabled attribute (28115e1392131866eb53323959677d96c53055bb)

How to reproduce:
```
$ git clone git@github.com:vaadin/vaadin-flow-components.git
$ cd vaadin-flow-components
$ mvn -am -pl vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests verify -Dit.test=GridWithTemplateIT 
```

Test fails with the following output
```
Caused by: java.lang.NullPointerException
	at com.vaadin.flow.component.grid.Grid.onEnabledStateChanged(Grid.java:3305)
	at com.vaadin.flow.dom.Element.informEnabledStateChange(Element.java:1587)
	at com.vaadin.flow.dom.Element.setEnabled(Element.java:1561)
	at com.vaadin.flow.component.Component.mapToElement(Component.java:197)
	at com.vaadin.flow.component.Component.<init>(Component.java:118)
	at com.vaadin.flow.component.grid.Grid.<init>(Grid.java:1322)
	at com.vaadin.flow.component.grid.Grid.<init>(Grid.java:1196)
	at com.vaadin.flow.component.grid.Grid.<init>(Grid.java:1181)
```

pleku:
This reverts commit 28115e1392131866eb53323959677d96c53055bb.

Decided to fix this by reverting the implementation for mapping disabled
attribute for mapped components, as it is so little added user value for so
big risk of regressions (like this one), confusion and bad code introduced
into the framework, increasing maintenance costs. Will change docs too.
Further details in #9140

Fixes #9136


denis-anisimov (commented):
This revert is OK.
But the main purpose of the initial PR was to fix the behavior with `setEnabled(true)` for the element with `disabled`  attribute. This is confusing behavior for sure.
So : should the whole `"disabled"` reading attribute feature be reverted ?
Related question: should we set "disabled" property at all for "disabled" template attribute?
(without having a special handler in the code "disabled" will be treated as a usual attribute and `"disabled"` property will be set). I see this behavior as a consistent behavior so I would just remove the special handler and that's it.

pleku:
>  I would just remove the special handler and that's it.

Yeah right, I think that it probably needs to be reverted too if it would mean that `mappedComponent.setEnabled(true);` does not remove the `disabled` attribute in DOM for the component.

denis-anisimov:
> > I would just remove the special handler and that's it.
> 
> Yeah right, I think that it probably needs to be reverted too if it would mean that `mappedComponent.setEnabled(true);` does not remove the `disabled` attribute in DOM for the component. I would maybe revert/handle that separately so we can get the component build green with just this revert.

Sure

denis-anisimov (approved):

PR #9151, opened 2020-10-09
chore: remove mentions of “Connect”

jouni:
Didn’t touch the package names, as I have no idea of all the places where those should be changed. Just updated the strings that I believe are more visible for users.

I expect someone from the dev team to pick this up and finalize it. Or close it, if this is causing more trouble than what it’s worth.


flow-server/src/main/java/com/vaadin/flow/server/connect/generator/OpenApiSpecGenerator.java
```diff
@@ -43,8 +43,8 @@
     public static final String SERVER_DESCRIPTION = "vaadin.connect.server.description";
     public static final String PREFIX = "vaadin.endpoint.prefix";
     public static final String DEFAULT_SERVER = "http://localhost:8080";
-    public static final String DEFAULT_SERVER_DESCRIPTION = "Vaadin Connect backend";
-    public static final String DEFAULT_APPLICATION_TITLE = "Vaadin Connect Application";
+    public static final String DEFAULT_SERVER_DESCRIPTION = "Vaadin Fusion backend";
+    public static final String DEFAULT_APPLICATION_TITLE = "Vaadin Fusion Application";
```
Artur-, line 47: This should be "Vaadin Application"

flow-server/src/main/java/com/vaadin/flow/server/connect/generator/OpenApiSpecGenerator.java
```diff
@@ -43,8 +43,8 @@
     public static final String SERVER_DESCRIPTION = "vaadin.connect.server.description";
     public static final String PREFIX = "vaadin.endpoint.prefix";
     public static final String DEFAULT_SERVER = "http://localhost:8080";
-    public static final String DEFAULT_SERVER_DESCRIPTION = "Vaadin Connect backend";
-    public static final String DEFAULT_APPLICATION_TITLE = "Vaadin Connect Application";
+    public static final String DEFAULT_SERVER_DESCRIPTION = "Vaadin Fusion backend";
```
Artur-, line 46: Not sure where this is used but why point out "Fusion" at all?

flow-server/src/main/java/com/vaadin/flow/server/connect/generator/OpenApiSpecGenerator.java
```diff
@@ -84,7 +84,7 @@ public void generateOpenApiSpec(Collection<Path> sourcesPaths,
                 FileUtils.writeStringToFile(specOutputFile.toFile(),
                         Json.pretty(openAPI), StandardCharsets.UTF_8);
             } else {
-                log.info("There are no connect endpoints to generate.");
+                log.info("There are no Fusion endpoints to generate.");
```
Artur-, line 87: Same here, why use "Fusion endpoints" instead of "endpoints"?

flow-server/src/main/java/com/vaadin/flow/server/connect/generator/OpenApiSpecGenerator.java
jouni: Sure, no argument from me. I hope someone else makes that decision. I mainly opened this PR to trigger this type of discussion, and so that we don’t forget to update these.

joheriks (changes requested):
These mentions of "Vaadin Connect" could also be changed:
```
$ grep -r -w "Connect" flow-server/src/main/java 
<irrelevant hits removed>
flow-server/src/main/java/com/vaadin/flow/server/connect/generator/VaadinConnectClientGenerator.java: * Generates the Vaadin Connect Client file, based on the application
flow-server/src/main/java/com/vaadin/flow/server/connect/VaadinConnectControllerConfiguration.java:     * Prepends the Connect prefix URL from the Vaadin Connect properties to
flow-server/src/main/java/com/vaadin/flow/server/connect/VaadinConnectControllerConfiguration.java:     * @return a new mapping with the Connect prefix URL prepended to the
```

haijian-vaadin:
> These mentions of "Connect" could also be changed:
> 
> ```
> $ grep -r -w "Connect" flow-server/src/main/java 
> <irrelevant hits removed>
> flow-server/src/main/java/com/vaadin/flow/server/connect/generator/VaadinConnectClientGenerator.java: * Generates the Vaadin Connect Client file, based on the application
> flow-server/src/main/java/com/vaadin/flow/server/connect/VaadinConnectControllerConfiguration.java:     * Prepends the Connect prefix URL from the Vaadin Connect properties to
> flow-server/src/main/java/com/vaadin/flow/server/connect/VaadinConnectControllerConfiguration.java:     * @return a new mapping with the Connect prefix URL prepended to the
> ```

Good catch, done. Have to use connect-client for the first one, since the file name is still like that.

joheriks (approved):

PR #9161, opened 2020-10-12
fix(TypeScript): client-side form binder validation errors because of value having unused deep optional fields

### LINKED ISSUE #9114: Vaadin for TypeScript: @vaadin/form Binder throws exception when trying to validate nullable object property
To reproduce
1. Clone the [CRM TS](https://github.com/vaadin-learning-center/crm-tutorial-typescript) project
2. Change the Vaadin version to 17.0.4 and flow version to 5.0-SNAPSHOT
3. Open the list view, Click the **Add Contact** button. 
4. Select a value for the Company combobox, the attached exception is thrown
Screenshot 2020-10-05 at 16 09 07

platosha:
Fixes #9114

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/form/BinderNode.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/form/BinderNode.ts
@@ -339,8 +339,11 @@ export class BinderNode<T, M extends AbstractModel<T>> {
         }
       }
     } else if (this.model instanceof ArrayModel) {
-      for (const childModel of this.model) {
-        yield childModel;
+      // Only iterate if optional array is initialised
+      if (this.value !== undefined) {
+        for (const childModel of this.model) {
+          yield childModel;
+        }
       }
     }
   }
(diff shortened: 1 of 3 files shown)


haijian-vaadin (changes requested):
I tried it out on our CRM TS tutorial. It works now when I bind the company object to the Combobox, however, if I try to bind it to the company id property, as we did before
```
<vaadin-combo-box
        label="Company"
        item-label-path="name"
        item-value-path="id"
        .items=${this.companies}
        ...="${field(this.binder.model.company.id)}"
></vaadin-combo-box>
```
Then I got the error as attached screenshot
Screenshot 2020-10-14 at 15 06 47

platosha:
@haijian-vaadin good catch, thanks. It was another related issue, now fixed.

Note that `item-value-path="id"` has no effect for form binding, because the default ComboBoxStrategy is using `.selectedItem` instead of `.value` in `<vaadin-combo-box>`. With your example code, this means that the type of `items`, and, hence, `selectedItem` (which is a `Company` object) does not match the bound form property (which is an Number). This probably would result in more runtime errors.

I tried with using number array in `items`, works as expected after the fix.

haijian-vaadin (approved):

PR #9168, opened 2020-10-13
feat!: don't support disabled for lit id elements

### LINKED ISSUE #9152: Throw when detecting disabled attribute for lit templates
Follow up for #8973 which was reverted by #9136 abandoned.

Since there is no good way to support mapping `disabled` attribute for `setEnabled(false)` for `@Id`-mapped components in a lit template and the expected low added value for the high cost to make it work right, we should instead properly discourage users from using `disabled` attribute for components that are Id-mapped in lit templates.

Thus when parsing the disabled attribute value, we should instead throw an exception with clear message stating that "Instead of using the disabled attribute in the template xyz.ts on row 123, you should instead disable the component from Java logic with setEnabled(false)" or similar. The docs should be updated too.

For polymer templates, it can be an attribute that is read to server side memory (starting from v18).

caalador:
LitTemplate will throw an exception for an id-mapped
component with the disabled attribute informing to use
the server side setEnabled(false) to disable mapped element.

Fixes #9152


denis-anisimov (changes requested):
This solution puts the logic specific for `disabled`  attribute inside `LitTemplateInitializer` whose essence is more generic and I would say it should not know anything about any specific attribute.

I suggest to reuse the ready-made "local framework" which allows to apply any logic for any template attribute generally:
please see `InjectableLitElementInitializer`  and its superclass `AbstractInjectableElementInitializer`.
The latter class has mechanism which applies different strategies per attrtibute.
The strategies are registered via `createStategies`.

I suggest: 
- remove `DisabledInitializationStrategy` from the superclass.
- either  override `accept`  in subclass 
- or make `initialize` protected in the superclass and override it in the subclasses (looks a better solution)

then just implement the method so that if handles `disabled`  properly.

Then one more question: how this attribute should be handled in Polymer templates ?
The question needs to get an answer because Polymer `InjectablePolymerElementInitializer` class needs to have its own impl.
I'm fine to solve polymer related question in a separate PR but the current behavior should be reverted (and will be if  `DisabledInitializationStrategy`  is dropped).

caalador:
The problem with moving it into the injectable initializer is that then we only see the single element instead of being able to catch all for this template. Knowing coders they will only fix that one instance that is thrown for and not check for others in the same file.
Thus I would rather move the validation to `LitTemplateDataAnalyzer` where we do have all the data available.

For Polymer I read `For polymer templates, it can be an attribute that is read to server side memory (starting from v18).`  as just having it as is now. But I can drop the strategy ofc and have it be handled as just another attribute.

denis-anisimov:
> The problem with moving it into the injectable initializer is that then we only see the single element instead of being able to catch all for this template. Knowing coders they will only fix that one instance that is thrown for and not check for others in the same file.
> Thus I would rather move the validation to `LitTemplateDataAnalyzer` where we do have all the data available.
Hmm....
The point is correct.
But. But.
Not sure whether this worth it.....

I know that reading the long exception message is extremely hard. The message should be very short. Then I immediately see the problem.
When exception message is long then I'm simply is not able to catch everything what's in the message and will miss something anyway.......
That happens for me e.g. in our `ClassesSerializableTest` : the message is so long that it's easier to rerun test every time after fixing on issue.

On the other hand code-wise the isolation is better for sure than storing the code inside `LitTemplateDataAnalyzer`.

I don't know. May be good to ask a third opinion.... I would like to decouple `disabled` attribute logic from `LitTemplateDataAnalyzer`  for sure. Most likely `disabled`   attribute will be used so rarely that it's worth to keep code isolated vs better UX with the exception....
> 
> For Polymer I read `For polymer templates, it can be an attribute that is read to server side memory (starting from v18).` as just having it as is now. But I can drop the strategy ofc and have it be handled as just another attribute.

Do we have tests for that ? I haven't checked this in details.
I'm pretty sure that if you haven't removed `DisabledInitializationStrategy` and haven't done any changes in Polymer template related classes then `disabled`  will call `setEnabled(false)` for Polymer template which should not be done at least.
`DisabledInitializationStrategy` removal should fix this but then we need a specific test for `disabled` in polymer template

caalador:
Lets validate per mapped component and look into multi handling if it is felt like it would be needed later.

denis-anisimov (approved):

denis-anisimov:
> Lets validate per mapped component and look into multi handling if it is felt like it would be needed later.

Agreed.

PR #9171, opened 2020-10-14
Make LitTemplateParserImpl public. Fixes #9169

### LINKED ISSUE #9169: Please make LitTemplateParserImpl public
Vaadin 18.0.0.alpha1. Please make LitTemplateParserImpl public otherwise it's really hard to hook into the template loading mechanism.

mvysny:
Fixes #9169

### DIFF
--- a/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/LitTemplateParser.java
+++ b/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/LitTemplateParser.java
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.component.littemplate;
 
+import com.vaadin.flow.component.littemplate.internal.LitTemplateParserImpl;
 import org.jsoup.nodes.Element;
 
 import com.vaadin.flow.di.Instantiator;
--- a/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/LitTemplateParserImpl.java
+++ b/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/LitTemplateParserImpl.java
@@ -57,7 +57,7 @@ import elemental.json.JsonObject;
  *
  * @see BundleLitParser
  */
-class LitTemplateParserImpl implements LitTemplateParser {
+public class LitTemplateParserImpl implements LitTemplateParser {
 
     private static final LitTemplateParser INSTANCE = new LitTemplateParserImpl();
 
--- a/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/internal/LitTemplateParserImpl.java
+++ b/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/internal/LitTemplateParserImpl.java
@@ -13,7 +13,7 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-package com.vaadin.flow.component.littemplate;
+package com.vaadin.flow.component.littemplate.internal;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -24,6 +24,9 @@ import java.util.Locale;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
 
+import com.vaadin.flow.component.littemplate.BundleLitParser;
+import com.vaadin.flow.component.littemplate.LitTemplate;
+import com.vaadin.flow.component.littemplate.LitTemplateParser;
 import org.jsoup.UncheckedIOException;
 import org.jsoup.nodes.Element;
 import org.slf4j.Logger;
@@ -73,7 +76,7 @@ public class LitTemplateParserImpl implements LitTemplateParser {
     protected LitTemplateParserImpl() {
     }
 
-    static LitTemplateParser getInstance() {
+    public static LitTemplateParser getInstance() {
         return INSTANCE;
     }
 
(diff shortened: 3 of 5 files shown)


denis-anisimov (commented):
What's the purpose of having `LitTemplateParserImpl` `public`?

It has been done package local in purpose: `public` opens API and allows it to use by everyone. It's not API and may be needed only in special cases.
One such special case is testing. For that reason I made a proper way to instantiate the parser via factory and restricted visibility to package local.
In case of testing the package structure is not really important and you may just put your subclass in your own module into the same package and extend this class without any problems.

So in case there are no other cases except testing there is no need to make this `public`.
If there are other cases, then:
  - please let me know such cases
  - the class should be moved into some `internal`  package which indicates that this class is not supposed to be used externally (there was no such need if it's package local).

mvysny:
Karibu-Testing needs to patch into the `LitTemplateParserImpl` loading mechanism, to load the js/ts files from the local filesystem. That requires to extend the `LitTemplateParserImpl` class and override the `LitTemplateParserImpl.getSourcesFromTemplate()` function.

> In case of testing the package structure is not really important and you may just put your subclass in your own module into the same package and extend this class without any problems.

Yes, this hack can be used to make `LitTemplateParserImpl` effectively public. This hack can indeed be used. Unfortunately, with Karibu-Testing the situation is a bit more complex.

The reason is that Karibu-Testing compiles against Vaadin 14 in order to be always compatible with the latest LTS. In order to extend a Vaadin 18 class, a ByteBuddy dynamic extension mechanism needs to be used, which then fails with `IllegalAccessError... cannot access its superclass`.

Possible solutions:
1. Make `LitTemplateParserImpl` public
2. Make ByteBuddy generate the overriding class into the `com.vaadin.flow.component.littemplate` package and use `ClassLoadingStrategy.Default.INJECTION` - under testing.
3. Make Karibu-Testing compile against Vaadin 18 but test against Vaadin 14. Tricky, but will get rid of ByteBuddy hacks.

> So in case there are no other cases except testing there is no need to make this public.

Yeah, probably no other use-case. I can't picture anyone wishing to tinker with this loading mechanism. However, if someone does, then it's quite hard: `LitTemplateParserImpl` contains lots of code that needs to either be copied, or extended. The former is fragile (what if the class changes), the latter is hard (the `LitTemplateParserImpl` was not designed to be extensible, also it's package-private).

mvysny:
Fair enough, yeah the pure interface approach would make more sense. Thank you for explanation.

That being said, LitTemplateParserImpl contains a lot of logic which every implementor of LitTemplateParser will need (the annotation parsing+iteration). To be extra clean, we should extract some of the code into a helper classes or some kind of SPI layer. Alternatively, another factory for just locating the template file would be handy to have - then both `LitTemplateParserImpl.getSourcesFromTemplate(String tag, String url)` and NpmParser would delegate to that factory.

> Is the option 2 really that hard ? I would go with it unless it's extremely cumbersome.

I have found a way to configure ByteBuddy to be able to work, so option 2 is viable, code-wise. However, what about security? Java 9 Module mechanism prevents exactly this kind of 'split package' scenario: https://www.logicbig.com/tutorials/core-java-tutorial/modules/split-and-private-package.html .

denis-anisimov:
I didn't know about Java 9 module security (but I'm glad that they implemented this now : I never liked that package visibility could be broken so easily).

OK, then please :
- keep the class `public`
- move it to some `internal` package.

The question how to make this impl eligible for reusing is more complicated and it's not clear whether this is really necessary.
So let's not solve it now.

mvysny:
> The question how to make this impl eligible for reusing is more complicated and it's not clear whether this is really necessary.
So let's not solve it now.

I agree. For my purposes it's enough if I can extend the `LitTemplateParserImpl` class - no need for additional SPI. And you're right - nobody else wanted this feature so far, so no need to bother with SPIs and stuff.

@denis-anisimov just that I understand correctly: should I move the `LitTemplateParserImpl` class into the `com.vaadin.flow.component.littemplate.internal` package and keep it in the `flow-lit-template` Maven module?

denis-anisimov:
Yes, `com.vaadin.flow.component.littemplate.internal` I think is the proper package name.
Sure, keep it inside the same module. This functionality is lit related only.

denis-anisimov (approved):

mvysny:
@denis-anisimov one test failed:

```
java.lang.AssertionError: com.vaadin.flow.component.littemplate.internal.LitTemplateParserImpl
	at com.vaadin.flow.testutil.ClassesSerializableTest.serializeAndDeserialize(ClassesSerializableTest.java:336)
	at com.vaadin.flow.testutil.ClassesSerializableTest.classesSerializable(ClassesSerializableTest.java:308)
Caused by: java.io.NotSerializableException: com.vaadin.flow.component.littemplate.internal.LitTemplateParserImpl
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
	at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:348)
	at com.vaadin.flow.testutil.ClassesSerializableTest.serializeAndDeserialize(ClassesSerializableTest.java:232)
	at com.vaadin.flow.testutil.ClassesSerializableTest.serializeAndDeserialize(ClassesSerializableTest.java:334)
	... 26 more
```

However, I don't think `LitTemplateParserImpl` should be made serializable... Could you let me know how to exclude LitTemplateParserImpl from the serialization test please?

caalador:
@mvysny See `LitClassesSerializableTest` in the module

denis-anisimov (approved):

Issue #9005, opened 2020-09-11
Including flow server as a dependency breaks custom spring webmvc configuration


masc3d:
### Description of the bug

Merely including `flow-server` as a  dependency breaks custom spring config via `@EnableWebMvc`

### Minimal reproducible example

Here's a custom webmvc example for serving static resources

```
@Configuration
@Lazy(false)
@EnableWebMvc
class WebConfiguration : WebMvcConfigurer {
    override fun addResourceHandlers(registry: ResourceHandlerRegistry) {
        super.addResourceHandlers(registry.apply {
            // resource handler for static swagger-ui
            addResourceHandler("/rs/**")
                .addResourceLocations("classpath:/webapp/rs/")

            // resource handler for images
            addResourceHandler("/img/**")
                .addResourceLocations("classpath:/webapp/img/")

            // resource handler for icons
            addResourceHandler("/icon/**")
                .addResourceLocations("classpath:/webapp/icon/")
        });
    }

    override fun addViewControllers(registry: ViewControllerRegistry) {
        super.addViewControllers(
            registry.apply {
                // redirect(s) to index-html
                addRedirectViewController("/rs", "/rs/index.html")
                addRedirectViewController("/rs/", "/rs/index.html")
                addRedirectViewController("/rs/internal", "/rs/internal/index.html")
                addRedirectViewController("/rs/internal/", "/rs/internal/index.html")
            }
        )
    }
}
```

### Expected behavior

Custom webmvc config should not be adversely affected.
Also, when not including `vaadin-spring` or excluding its auto configuration, vaadin should not interfere with spring at all.

### Actual behavior

Just by including `flow-server` as a dependency custom webmvc setup stops working and fetching a static resource will yield:

```
Resolved [org.springframework.web.HttpRequestMethodNotSupportedException: Request method 'GET' not supported]
15:07:18.006 [mt-6] DEBUG o.s.web.servlet.DispatcherServlet - Completed 405 METHOD_NOT_ALLOWED
```

### Versions:
    - Vaadin / Flow version: 17.0.1
    - Java version: 14.0.2
    - OS version: macos-10.15.6
    - Browser version (if applicable): -
    - Application Server (if applicable): -
    - IDE (if applicable): IJ

### Sidenotes
* This issue did not occur with vaadin-14.2.0
* Spring configurations / controllers in non-spring artifact `flow-server` seem to be misplaced

haijian-vaadin:
[...]

PR #9172, opened 2020-10-14
fix: only register endpoint request handler when endpoint presents

haijian-vaadin:
Fixes https://github.com/vaadin/flow/issues/9005

For a non-boot Spring application or a Spring Boot application with auto-configuration disabled (e.g. when using `@EnableWebMvc`), Vaadin registers a request handler with mapping `/{endpoint}/{method}`. Note the connect prefix is not added in this case, since it relies on Spring Boot auto-configuraiton.

So it's better for users who are using endpoints, we don't register the endpoint request handler.

Test is in https://github.com/vaadin/spring/pull/672

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/Constants.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/Constants.java
@@ -412,6 +412,8 @@ public final class Constants implements Serializable {
     // Non-default port currently not supported (#7970)
     public static final int SPRING_BOOT_DEFAULT_LIVE_RELOAD_PORT = 35729;
 
+    public static final String VAADIN_ENDPINT_ENABLED = "endpoint.enabled";
+
     private Constants() {
         // prevent instantiation constants class only
     }
--- a/flow-server/src/main/java/com/vaadin/flow/server/Constants.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/Constants.java
@@ -412,8 +412,6 @@ public final class Constants implements Serializable {
     // Non-default port currently not supported (#7970)
     public static final int SPRING_BOOT_DEFAULT_LIVE_RELOAD_PORT = 35729;
 
-    public static final String VAADIN_ENDPINT_ENABLED = "endpoint.enabled";
-
     private Constants() {
         // prevent instantiation constants class only
     }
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/connect/VaadinConnectController.java
```diff
@@ -94,6 +97,7 @@
 @RestController
 @Import({ VaadinConnectControllerConfiguration.class,
         VaadinEndpointProperties.class })
+@ConditionalOnProperty(value = VAADIN_ENDPINT_ENABLED, havingValue = "true")
```
knoobie, line 100: Just a question; would it be better to use Custom Conditional to only activate the controller if the `@Endpoint` annotation is used? Or is this over-engineered because you guys think about removing spring? https://reflectoring.io/spring-boot-conditionals/#custom-conditions

haijian-vaadin: Hi @knoobie, thanks for the comment. I was also thinking about using a Custom Condition in the beginning, but then realized that we are already scanning the Endpoints at build time, so we can just reuse that instead of having another check in a custom condition.

Also for an extreme case, if there is only one class that has `@Endpoint` annotation, but without any public methods, then the annotation scanning-based custom condition would still active the Vaadin controller while the solution in this PR won't.

Another big reason is what you mentioned, we might remove the spring dependency in the future, so the check won't be necessary, thus like to have a minimal effort solution.

flow-server/src/main/java/com/vaadin/flow/server/Constants.java
```diff
@@ -412,6 +412,8 @@
     // Non-default port currently not supported (#7970)
     public static final int SPRING_BOOT_DEFAULT_LIVE_RELOAD_PORT = 35729;
 
+    public static final String VAADIN_ENDPINT_ENABLED = "endpoint.enabled";
```
joheriks, line 415:
```suggestion
    public static final String VAADIN_ENDPOINT_ENABLED = "endpoint.enabled";
```
There's a typo in the constant. BTW, should it be a public constant (from API maintenance point of view this means it can't easily be removed or renamed later) or could it be put inside `com.vaadin.flow.server.connect` (e.g., within an `internal` package)?

joheriks (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/Constants.java
haijian-vaadin: Very good point, instead of create a new internal package inside `connect`, I reused a flow internal package and added a new InternalConstants, please help to take another look

joheriks: 👍. Looks like the imports may need updating.

haijian-vaadin: Good spot, thanks. Thought VS Code would have done that automatically for me.

joheriks (approved):

joheriks (approved):

PR #9177, opened 2020-10-14
Merge master to offline

haijian-vaadin:
DO NOT squash


manolo (approved):
Tested in my local, and works fine to me

PR #9180, opened 2020-10-14
Simplify check for recursive getContent() in Composite

### LINKED ISSUE #9160: Questionable use of instance ThreadLocal in Composite
Fix #8529 introduced non-static `ThreadLocal` which is used to prevent stack overflow due to `Composite.getContent()` reentry. However, If `getContent()` can be called only when `UI` is locked, `ThreadLocal` is not needed. And if `getContent()` can be called by several threads, then `initContent()` might be called twice and different threads will receive different instances of the underlying component (because of lack of synchronization)

Legioth:
Using a regular instance field instead of a non-static ThreadLocal has
no functional difference except in the hypothetical case when
getContent() would be called concurrently on a different thread. That
case is not supposed to happen since the session lock should be used to
prevent concurrent access. Furthermore, concurrent invocation would
still not work properly since use of the `content` field is not
synchronized.

A simple boolean communicates the intent more clearly, thus making the
purpose of the implementation easier to understand. The simpler approach
also has a minimal positive impact on CPU usage and memory allocation.

Fixes #9160

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Composite.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Composite.java
@@ -15,8 +15,6 @@
  */
 package com.vaadin.flow.component;
 
-import java.io.IOException;
-import java.io.ObjectInputStream;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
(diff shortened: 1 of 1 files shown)


pleku (approved):

taefi:
Cherry-picked to `4.0`

Issue #8996, opened 2020-08-27
Vaadin for TypeScript: static files location handling for @pwa 


haijian-vaadin:
The `offlineResources` property in `@pwa` annotation should work the same as in V17. See this [documentation](https://vaadin.com/docs/v17/flow/pwa/tutorial-pwa-service-worker.html)

haijian-vaadin:
Close as this is implemented in https://github.com/vaadin/flow/pull/8940

vlukashov:
The way how the `offlineResources` property of the `@PWA` annotation is handled in #8940 is **incompatibly different** from how it's handled in V17:

With #8940 the files listed in the `offlineResources` property are resolved at the build time by webpack, relative to the `src/main/webapp` folder in the project file system. With V17 these files are resolved at the run time by the service worker, through the running web app by URL.

Since the V17 approach relies on the running web app's URL-to-file resolution logic, it means that the actual location of these files on the file system may vary a lot.

With the default setup Spring Boot apps serve static files from the `/static`, `/public`, `/resources`, and `/META-INF/resources` locations in the class path (in the project itself or in any .jar it loads), as well as from the `src/main/webapp` folder in the project (which is the .war-specific convention for static files for Java web apps). These defaults can be changed by configuration. See more details e.g. in https://vaadin.com/docs/v17/flow/importing-dependencies/tutorial-ways-of-importing.html or https://www.baeldung.com/spring-mvc-static-resources

Picking only the `src/main/webapp` folder as the supported location of files referenced in the `offlineResources` property is an incompatible breaking change from V17. It excludes any files that come from .jars as well as any files that are located in e.g. `src/main/resources/META-INF/resources`.

**Here are the missing bits before this task can be completed:**
 - either change the V18 approach so that the files referenced in the `offlineResources` property are resolved at the run time through the web app by URL rather than at the build time by a file name
 - or keep the resolution at the build-time and update the https://vaadin.com/docs/v17/flow/pwa/tutorial-pwa-service-worker.html#defining-custom-cache-resources documentation page to make it explicit that the files should be placed in the `src/main/webapp` folder

Haprog:
@vlukashov Thanks for the clarification. Now I understand what this is about. Should probably be marked as a bug instead of enhancement?

vlukashov:
[...]

PR #9181, opened 2020-10-15
fix(offline): support offlinePath from PWA annotation

### LINKED ISSUE #8856: Keep V18 compatible with the `offlinePath` property of `@PWA`
When upgrading a Vaadin 12+ app that uses the `@PWA` annotation to Vaadin 18, I want the `offlinePath` property of the `@PWA` annotation to remain working as before, so that my migration is easier. 

The `offlinePath` property in V12+ is a path to a static 'offline page' HTML file that's served by the service worker instead of the bootstrap page when the app is offline. In V18 doing exactly the same does not make sense because the same app shell remains active also when offline. The expectation is that the 'offline page' HTML is displayed in place of the server-side views, not in place of the entire application shell.

One solution to this would be creating a 'compatibility offline view' web component that renders the contents of the custom offline page inside an iframe.

DoD:
 - Having a V17 app with a `@PWA` annotation and a custom `offlinePath`, after an upgrade to V18, when the app goes offline I want to see the content of the custom `offline.html` rendered instead the app shell (i.e. the behavior in V18 remains exactly the same as it is in V17 and no new offline features are available to the app until the developer removes the `offlinePath` property from the `@PWA` annotation).
 - same as above, but instead of using a custom `offlinePath` property, the app has customized content in the default `offline.html` file path -> the presence of the `offline.html` resource is detected and when the app is offline, the content of `offline.html` is rendered instead the app shell
 - when the `offlinePath` property of the `@PWA` annotation is not set, when offline the app renders the app shell. See https://github.com/vaadin/flow/issues/9132.

platosha:
Fixes #8856


haijian-vaadin (changes requested):
Few findings:
1. the second point of DoD in the ticket seems is not taken into consideration. i.e. 

> same as above, but instead of using a custom offlinePath property, the app has customized content in the default offline.html file path -> the presence of the offline.html resource is detected and when the app is offline, the content of offline.html is rendered instead the app shell

2. in a v17 Spring app, the `offlinePath` is relative to `src/main/resources/META-INF/resources`, in our offline branch, it's relative to `src/main/webapp`, but maybe this is in the scope of #8996.

haijian-vaadin (dismissed):
Let's merge this one, have a new ticket regarding `customized content in the default offline.html file path`

flow-server/src/main/java/com/vaadin/flow/server/PWA.java
```diff
@@ -66,14 +66,15 @@
...
      * Path to the static offline html file.
      *
-     * Defaults to (relative) {@literal offline.html} with default configuration
-     * that is {@literal webapp/offline.html}
+     * Defaults to empty value, meaning that the application shell (`index
+     * .html` file) is used offline.
      *
-     * If offline file is not found, falls back to default offline page.
+     * If offline file is not found, falls back to using the application shell
+     * offline.
```
joheriks, line 73: Is that actually correct? I tested `@PWA(offlinePath="notexisting.html)` where `notexisting.html` is not served by the app and the browser replied `This site can’t be reached` rather than falling back to the appshell.

Haprog (approved):
Looks good.

PR #9183, opened 2020-10-15
fix: implement navigation fallback

### LINKED ISSUE #9137: Implement app shell as navigation fallback in sw.ts
In `feature/offline` branch this was already working/implemented as an option of `GenerateSW` but the PR https://github.com/vaadin/flow/pull/9109 changed to use `InjectManifest` so on top of that we need to implement the same behaviour in the new custom service worker template (`sw.ts`). So this is a regression caused by the above PR.

The user should still be able to reload any view (URL) while offline without seeing a network error like "No internet". Related to #8878 (but this issue is mainly about non-deep URLs)

Haprog:
Fixes #9137


haijian-vaadin (changes requested):
Works as expected, awesome! 
Meanwhile, wondering if we can have an IT test, for example, in flow-tests/test-ccdm for this feature? Some [quick reference ](https://stackoverflow.com/questions/40531771/selenium-set-browser-chrome-in-offline-mode) on how to mimic the offline mode with selenium driver.

haijian-vaadin (approved):
Agreed that we can add IT tests later

PR #9195, opened 2020-10-18
fix: Support .ts files in parsers

### LINKED ISSUE #9165: LitTemplate parser prints: LitTemplateParserImpl - Couldn't find the definition of the element with tag 'my-form'
Having the following Java class
```java
@Tag("my-form")
@JsModule("./src/my-form.ts")
public class MyForm extends LitTemplate {
    @Id
    public TextField nameField;
}
```

accompanied by the `my-form.ts` file:

```ts
import {html, LitElement} from 'lit-element';

// @customElement("my-form")
export class MyFormElement extends LitElement {
    render() {
        return html`
      <h2>Hello</h2>
      <vaadin-text-field id="nameField"></vaadin-text-field>
    `;
    }
}

customElements.define("my-form", MyFormElement);
```

Upon browsing to the route containing this component, the following is logged into the console:
```
[ForkJoinPool.commonPool-worker-1] INFO com.vaadin.flow.server.DevModeHandler - Started webpack-dev-server. Time: 10342ms
[http-nio-8080-exec-7] INFO com.vaadin.flow.component.littemplate.LitTemplateParserImpl - Couldn't find the definition of the element with tag 'my-form' in any lit template file declared using '@JsModule' annotations. Check the availability of the template files in your WAR file or provide alternative implementation of the method LitTemplateParser.getTemplateContent() which should return an element representing the content of the template file
[http-nio-8080-exec-7] INFO com.vaadin.flow.component.littemplate.LitTemplateDataAnalyzer - Couldn't parse template for class org.vaadin.example.MyForm class. Only specific Lit template format is supported. Please check that your template definition directly contains 'render' method which returns html`_template_content_`.
[http-nio-8080-exec-5] INFO com.vaadin.flow.component.littemplate.LitTemplateParserImpl - Couldn't find the definition of the element with tag 'my-form' in any lit template file declared using '@JsModule' annotations. Check the availability of the template files in your WAR file or provide alternative implementation of the method LitTemplateParser.getTemplateContent() which should return an element representing the content of the template file
[http-nio-8080-exec-5] INFO com.vaadin.flow.component.littemplate.LitTemplateDataAnalyzer - Couldn't parse template for class org.vaadin.example.MyForm class. Only specific Lit template format is supported. Please check that your template definition directly contains 'render' method which returns html`_template_content_`.
```

However, the form itself is displayed, and the `nameField` is populated correctly (the following code works properly and doesn't fail with NPE):

```java
@Route("")

    public MainView() {
        final MyForm form = new MyForm();
        add(form);
        form.nameField.setValue("Link");
    }
}
```

The INFO messages should either be an exception (can't load template, can't continue) or shouldn't be logged at all. Current behavior is confusing for the developer - is it working correctly or not? Do I perhaps have some kind of mistake in my ts file?

### Versions:
    - Vaadin / Flow version: 18.0.0.alpha1 / 5.0.0.alpha1
    - Java version: 1.8
    - OS version: Ubuntu 20.10 x86-64
    - Browser version (if applicable): Firefox
    - Application Server (if applicable): Tomcat 9.0.31
    - IDE (if applicable): Intellij

caalador:
Add support for parsing sources for
TypeScript modules.

Fixes #9165

### DIFF
--- a/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
+++ b/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
@@ -24,7 +24,6 @@ import static elemental.json.JsonType.STRING;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import org.apache.commons.io.FilenameUtils;
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Element;
@@ -181,12 +180,6 @@ public final class BundleLitParser {
         if (source == null && validKey(module, NAME, STRING)
                 && validKey(module, SOURCE, STRING)) {
             String name = module.getString(NAME);
-
-            // don't append extension if one already exists
-            if(FilenameUtils.getExtension(fileName).isEmpty()) {
-                // append `.js` extension if not yet as webpack does
-                fileName = fileName.replaceFirst("(\\.js|)$", ".js");
-            }
 
             String alternativeFileName = fileName
                     // Replace frontend part since webpack entry-point is
--- a/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
+++ b/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
@@ -74,14 +74,14 @@ public final class BundleLitParser {
      * <p>
      * <code>render\(\)[\s]*\{</code> finds the template getter method
      * <p>
-     * <code>[\s]*return[\s]*html`</code> finds the return statement
+     * <code>[\s]*return[\s]*html[\s]*`</code> finds the return statement
      * <p>
      * </p>
      * <code>(([^`]|\\\\.)*)</code> captures all text until we encounter the end
      * character with <code>;}</code> e.g. <code>';}</code>
      */
     private static final Pattern LIT_TEMPLATE_PATTERN = Pattern.compile(
-            "render\\(\\)[\\s]*\\{[\\s]*return[\\s]*html`(([^`]|\\\\.)*)`;[\\s]*\\}");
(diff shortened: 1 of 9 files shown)


denis-anisimov (commented):
Should we care at all about extension ?

caalador:
In the template parser the dependencyHasTagName seems like extra and I could just drop it.

denis-anisimov:
> In the template parser the dependencyHasTagName seems like extra and I could just drop it.

Do you see any potential issue with this dropping ?
Or do you have any idea behind the original reason of this check ?

I would just drop it if there is no any side effects: don't see why we should restrict extensions.

caalador:
I removed the check on the extension. the template parser one is for seeing that we are matching against the correct file when there are multiple module annotations. The bundle parser I guess is for cases where webpack has auto added the .js extension for some reason. I'll try to dig up if I can see why that would have happened

denis-anisimov:
> I removed the check on the extension. the template parser one is for seeing that we are matching against the correct file when there are multiple module annotations. The bundle parser I guess is for cases where webpack has auto added the .js extension for some reason. I'll try to dig up if I can see why that would have happened

So won't there will be any issues in multiple module annotations matching with this dropping ?

I guess the check for ".js" extension has been added exactly because webpack auto-addition.
But now ".js" suffix will be added to a file without extension anyway . And check against ".ts" has no sense at least.

I suppose there is no any real test which allows to reveal the reason of ".js" check?

caalador:
I dropped the extension check and we still require the tag to match the name. The bundle parser shouldn't need the .ts addition as webpack only adds .js to chunkFilename with the chunk id which doesn't matter with the stats which is based on the actual file name and not a chunkFilename.

denis-anisimov:
> I dropped the extension check and we still require the tag to match the name. The bundle parser shouldn't need the .ts addition as webpack only adds .js to chunkFilename with the chunk id which doesn't matter with the stats which is based on the actual file name and not a chunkFilename.

OK, cool.

denis-anisimov (approved):

Issue #8761, opened 2020-07-23
ListDataView: use ArrayList as the default option


haijian-vaadin:
When calling `grid.setItem(personA, personB)`, it's using `Array.asList()` as the default option. As a result, many of the API provided by the `ListDataView` won't work, since they are index-based. such as `ListDataView.addItemAfter()`.

mstahv:
Getters should work fine, but mutators don't work if the underlaying collection is immutable. Also when assigning a collection instead of an array/vararg. This is mentioned in the javadocs and documentation, but I agree it may appear as a bug. In-fact, I reported this same issue to the team as well, but it "works as coded". 

A potential way to improve the experience would be to create an ArrayList instead of using Array.asList when using varargs, but I guess we don't want to replace the collection.

Could you @haijian-vaadin verify if you had an issue with getNextItem?

haijian-vaadin:
oh, yes, actually `getNextItem` works.

mstahv:
Let's keep this open and see if more people expect this to work this way. An easy and safe enhancement if seen necessary, but I'd guess quite obsolete for real world use cases. But I might be wrong...

pleku:
To be honest, this feels to me like an existing bug that we have been carrying out since Vaadin 8. It has been introduced there as an accident and I don't know why we would not fix this.

PR #9197, opened 2020-10-19
fix: Use ArrayList while creating ListDataProvider from items array

taefi:
Wrapped the result of `Arrays.asList(T... items)` by `java.util.ArrayList` to support further add/remove operations through the `DataView` API without getting an `UnsupportedOperationException`. 

Fix: #8761

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataProvider.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataProvider.java
@@ -16,6 +16,7 @@
 package com.vaadin.flow.data.provider;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/DataProvider.java
```diff
@@ -291,7 +292,7 @@ protected F combineFilters(Q queryFilter, C configuredFilter) {
      */
     @SafeVarargs
     static <T> ListDataProvider<T> ofItems(T... items) {
-        return new ListDataProvider<>(Arrays.asList(items));
+        return new ListDataProvider<>(new ArrayList<>(Arrays.asList(items)));
```
mshabarov, line 295: Could you please also check the components whether they use the immutable collections to set the items to data provider. At least I could find this case in combo box: https://github.com/vaadin/vaadin-flow-components/blob/ecc6d468f11ba7dc33d4ae4cab86bdbeee2b3766/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L506

mshabarov (approved):

PR #9198, opened 2020-10-19
feat: disposable filter setup API in data communicator

mshabarov:
Extends the `DataCommunicator` API to set the permanent/disposable flag to the component's filter object and discards this filter after the next request to data provider, if disposable. This is needed to ignore the client filter of `ComboBox` in data view retrieve methods (`getItems` for instance) and to not fire the item count change event in case of client filtering.

Overloaded version of `setRequestedRange` has also added in order to avoid a doubled `flush` invocations in `ComboBox`, when the client filter has changed.

Part of https://github.com/vaadin/vaadin-flow-components/issues/210

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
@@ -101,8 +101,7 @@ public class DataCommunicator<T> implements Serializable {
 
     private DataProvider<T, ?> dataProvider = new EmptyDataProvider<>();
 
-    // Serializability of filter is up to the application
-    private Object filter;
+    private Filter<?> filter;
     private SerializableComparator<T> inMemorySorting;
 
     private Registration dataProviderUpdateRegistration;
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -225,17 +304,36 @@ public DataCommunicator(DataGenerator<T> dataGenerator,
...
+     * 
+     * @param start
+     *            the start of the requested range
+     * @param length
+     *            the end of the requested range
+     * @param requestNow
+     *            if {@code true}, the data will be requested immediately,
+     *            otherwise it will just set the range.
+     */
+    public void setRequestedRange(int start, int length, boolean requestNow) {
```
pleku, line 331: Maybe `requestFlush` is better name

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -306,16 +404,26 @@ public void confirmUpdate(int updateId) {
      * @param initialFilter
      *            the initial filter value to use, or <code>null</code> to not
      *            use any initial filter value
+     * @param permanentFilter
+     *            if {@code true}, then the initial filter is considered as a
+     *            permanent and won't be discarded after the next call to data
+     *            provider, otherwise it will be treated as a disposable filter
+     *            and will be discarded
```
pleku, line 411: What is a disposable filter ?
The semantics should be explained in this method

pleku (commented):
I would want to see a bit on the usage for CB

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -145,6 +144,86 @@ public EmptyDataProvider() {
...
+     * information whether this filter is set permanently or should it be
+     * discarded after the next request to data provider.
+     *
+     * @param <F>
+     *            filter's type
+     */
+    public static final class Filter<F> implements Serializable {
+
+        // Serializability of filter is up to the application
+        private Object filter;
```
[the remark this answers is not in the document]
knoobie: Just wondering; why is this field of type `Object` instead of `F`? Same as the return value of `getFilter`

mshabarov:
> I would want to see a bit on the usage for CB

The CB PR is here https://github.com/vaadin/vaadin-flow-components/pull/278

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mshabarov: Made a clarifications in javadocs

mshabarov:
Still in draft, because the tests should be added.

pleku (approved):

PR #9209, opened 2020-10-21
test: add methods to simulate offline mode with ChromeDriver

Haprog:


Haprog:
Closing in favor of ChromeDeviceTest that was merged in #9181.

PR #9227, opened 2020-10-23
fix: Update changed value to field with formatting

TatuLund:
Binder: Convert value back to representation and write back to field when changed. This ensures formatting of the value if converter is used.

Pick from https://github.com/vaadin/framework/pull/12132


pleku (approved):
While I agree that this is the behavior that should happen, I want to make couple of comments:
- The unit test is a bit shallow and just validates the happy path
- There is no issue for this (in /flow or /framework) so it is mayhaps not even a _fix_ but a new _feature_ as the behavior will change for existing apps.

I'm going to allow this as is, but as a precaution, I would treat this as a feature that affects existing UI behavior, though making it better.

PR #9228, opened 2020-10-23
Cherry picks for 2.5

mshabarov:
Merge not squash


caalador:
I think we should rebase/merge the whole 2.4 and not only these 2.

caalador (changes requested):
Should contain everything from 2.4 branch that os not in 2.5

mshabarov:
Force pushed to the branch, so I think it's better to create a new PR, because the commit history was rewrited.

Issue #9232, opened 2020-10-26
Investigate AmbiguousRouteConfigurationException in the root context tests for 2.4


caalador:
It seems like the TestingServiceInitListener may be called so that it gets a race condition making it try to register the DynamicallyRegisteredRoute twice:
```
Caused by: com.vaadin.flow.server.AmbiguousRouteConfigurationException: Navigation targets must have unique routes, found navigation targets 'com.vaadin.flow.uitest.ui.DynamicallyRegisteredRoute' and 'com.vaadin.flow.uitest.ui.DynamicallyRegisteredRoute' with the same route.
     at com.vaadin.flow.server.startup.RouteTarget.validateNormalTarget (RouteTarget.java:163)
     at com.vaadin.flow.server.startup.RouteTarget.addTargetByType (RouteTarget.java:107)
     at com.vaadin.flow.server.startup.RouteTarget.addRoute (RouteTarget.java:100)
     at com.vaadin.flow.router.internal.ConfigureRoutes.setRoute (ConfigureRoutes.java:133)
     at com.vaadin.flow.router.internal.AbstractRouteRegistry.addRouteToConfiguration (AbstractRouteRegistry.java:365)
     at com.vaadin.flow.router.internal.AbstractRouteRegistry.lambda$setRoute$83b15968$1 (AbstractRouteRegistry.java:307)
     at com.vaadin.flow.router.internal.AbstractRouteRegistry.configure (AbstractRouteRegistry.java:95)
     at com.vaadin.flow.router.internal.AbstractRouteRegistry.setRoute (AbstractRouteRegistry.java:306)
[09:10:50] :	 [Step 4/6]     at com.vaadin.flow.server.startup.ApplicationRouteRegistry.setRoute (ApplicationRouteRegistry.java:279)
     at com.vaadin.flow.router.RouteConfiguration.setRoute (RouteConfiguration.java:292)
     at com.vaadin.flow.router.RouteConfiguration.setRoute (RouteConfiguration.java:270)
     at com.vaadin.flow.uitest.ui.TestingServiceInitListener.serviceInit (TestingServiceInitListener.java:48)
```

PR #9233, opened 2020-10-26
test: Fix duplicate registration

### LINKED ISSUE #9223: PatientPortal for 2.4 and 2.5 fails with a `session deleted because of page crash`
For some reason the patient portal builds for 2.4 and 2.5 fail with page crash.

caalador:
Lock the registry when registering
dynamic route in serviceInit to not have
a race where 2 inits note that path not
registered and then register the route.

Fixes #9223


pleku (approved):
Anything to update in the error message thrown for the ambiguous modification that would have helped to pinpoint the problem this time ?

caalador:
> Anything to update in the error message thrown for the ambiguous modification that would have helped to pinpoint the problem this time ?

Not that I can come up with. The main issue is perhaps that the window size is too small in the screenshot and we get almost 20 megs of log.

caalador:
This fixed #9232 not 9223 as stated in the comment.

PR #9244, opened 2020-10-27
feat: defer deferrable endpoint calls on network failure

### LINKED ISSUE #9231: Fusion: should defer deferrable calls when server is unreachable
As a user, I want to submit a form not only when the browser is offline (by checking `navigator.online`), but also when the server is offline, e.g. the server is down, or when I'm behind a wifi gate, so that even though `navigator.online` is true but I cannot reach the server.

DoD:
1. When navigator.online is false, a deferrable endpoint call should be deferred.
2. When the server cannot be reached, i.e. error with message `Failed to fetch`  a deferrable endpoint call should be deferred.
3. When the server fail for other errors, e.g. `Endpoint 'ServiceEndpoint' method 'saveContact' execution failure`, a deferrable endpoint call should NOT be deferred.

A quick prototype in `Connect.ts`
```
async deferrableCall(
    endpoint: string,
    method: string,
    params?: any,
  ): Promise<DeferrableResult<any>> {
    if (this.checkOnline()) {
      try{
        const result = await this.call(endpoint, method, params);
        return { isDeferred: false, result };
      }catch(error){
        if(error.message === 'Failed to fetch'){
          let endpointRequest:EndpointRequest = { endpoint, method, params };
          endpointRequest = await this.cacheEndpointRequest(endpointRequest);
          return { isDeferred: true, endpointRequest };
        } else{
          throw error;
        }
      }
    } else {
      let endpointRequest:EndpointRequest = { endpoint, method, params };
      endpointRequest = await this.cacheEndpointRequest(endpointRequest);
      return { isDeferred: true, endpointRequest };
    }
  }
```

haijian-vaadin:
Fixes #9231


flow-client/src/main/resources/META-INF/resources/frontend/Connect.ts
```diff
@@ -339,12 +339,18 @@ export class ConnectClient {
     params?: any,
   ): Promise<DeferrableResult<any>> {
     if (this.checkOnline()) {
-      const result = await this.call(endpoint, method, params);
-      return { isDeferred: false, result };
+      try {
+        const result = await this.call(endpoint, method, params);
+        return { isDeferred: false, result };
+      } catch (error) {
+        if (error.message === 'Failed to fetch') {
```
platosha, line 346: In Firefox the error message is different: `NetworkError when attempting to fetch resource.`

This tells me that checking a message string is unreliable.

I’d suggest to check against the errors we throw ourselves instead:
```
if ((error instanceof EndpointResponseError) || (error instanceof EndpointError)) {
  // Error thrown in the client, carry to the user
  throw error;
} else {
  // Network error
  return this.cacheEndpointRequest({endpoint, methods, params});
}
```

This is safer: if it’s not the error we throw ourselves, then it’s a `fetch` error, which is only thrown for the network reasons.

platosha (changes requested):

flow-client/src/main/resources/META-INF/resources/frontend/Connect.ts
haijian-vaadin: very good point, done

platosha (approved):

PR #9253, opened 2020-10-28
fix: make field with @Id nullable in TypeScript

### LINKED ISSUE #9222: Fusion: field with @Id annotation in Java should be nullable in the generated TypeScript code
Given a Java entity
```
@MappedSuperclass
public abstract class AbstractEntity {
    @Id
    @GeneratedValue(strategy = GenerationType.SEQUENCE)
    private Long id;
    ...
}
```
The generated TypeScript code is
```
export default interface AbstractEntity {
  id: number;
}
```
This leads to the problem that an object that extends from AbstractEntity will have a default value 0 for id. This could result the problem that a record of id 0 in the DB would be overridden.

Note that in the example above, it's not a problem, since it's using `GenerationType.SEQUENCE` and by default 0 is considered an invalid id. But the default behavior can be overridden.

DoD:
For a Java field that has an `@Id` annotation (with simple name check), the generated typescript field should be nullable. e.g.
```
export default interface AbstractEntity {
  id: number | undefined;
}
```

haijian-vaadin:
Fixes #9222


flow-server/src/test/java/com/vaadin/flow/server/connect/ExplicitNullableTypeCheckerTest.java
```diff
@@ -423,6 +442,15 @@ public String stringNullable() {
...
     }
 
+    /**
+     * Method for testing
+     */
+    @Id
+    @Nullable
+    public Long methodWithIdAnnotation(){
+        return null;
+    }
```
platosha, line 452: Let’s remove `@Nullable` from here. Otherwise, even if `@Id` is not implemented properly in the checker, the test might still pass because of `@Nullable` here.
```suggestion
    public Long methodWithIdAnnotation() {
        return 1L;
    }
```

platosha (changes requested):

platosha (approved):

PR #9258, opened 2020-10-28
fix(offline): resolve offlineResources at runtime

### LINKED ISSUE #8996: Vaadin for TypeScript: static files location handling for @pwa

joheriks:
Depending on servlet container type and configuration, offline resources
may be served from multiple different file system locations, which cannot
be predicted for the resources to be handled by webpack. Revert to runtime 
resolution by the applicatiton.

Fixes #8996.


flow-tests/test-pwa/src/test/java/com/vaadin/flow/pwatest/ui/PwaTestIT.java
```diff
@@ -202,22 +222,22 @@ private void checkIcons(List<WebElement> icons, int expected) {
...
 
+    private boolean exists(String url) {
+        final String script = "const resolve = arguments[0];" + "fetch('" + url
+                + "', {method: 'HEAD'})"
+                + ".then(response => resolve(response.status===200 && !response.redirected))"
+                + ".catch(err => resolve(false));";
+        return (boolean) ((JavascriptExecutor) getDriver())
+                .executeAsyncScript(script);
+    }
+
```
joheriks, line 240: Note. I have replaced the previous version of `exists` which checked for existence by connecting directly to the server with one that goes via the browser. This makes the test more end-to-end, and allows testing the caching of resources in offline mode.

flow-tests/test-pwa/src/test/java/com/vaadin/flow/pwatest/ui/PwaTestIT.java
```diff
@@ -38,12 +43,24 @@
...
+    @Before
+    public void start() {
+        lock.lock();
+    }
+
+    @After
+    public void done() {
+        lock.unlock();
+    }
```
Haprog, line 59: What does this (and the above `@NotThreadSafe`) actually do?

Haprog (changes requested):

flow-tests/test-pwa/src/test/java/com/vaadin/flow/pwatest/ui/PwaTestIT.java
Haprog: Or is this about making the tests run in sequence instead of parallel?

flow-server/src/main/java/com/vaadin/flow/server/PwaRegistry.java
```diff
@@ -226,17 +227,31 @@ private JsonObject initializeManifest() {
...
         // List of icons for precache
         List<String> filesToCache = getIcons().stream()
                 .filter(PwaIcon::shouldBeCached).map(PwaIcon::getCacheFormat)
                 .collect(Collectors.toList());
 
+        // When offlinePath is in use, it is also an offline resource to
+        // precache
+        if (pwaConfiguration.isEnabled()) {
+            filesToCache.add(offlinePageCache());
+        }
```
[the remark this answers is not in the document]
joheriks: Done, this was a mistake, should use `isOfflinePathEnabled` instead.

flow-tests/test-pwa/src/test/java/com/vaadin/flow/pwatest/ui/PwaTestIT.java
joheriks: Forcing the tests to run in sequence. I will remove these once it's clear whether the tests can run in parallel or not. If they need to be sequenced, `@NotThreadSafe` should be enough anyway (the annotation is recognized by `maven-failsafe-plugin`, https://maven.apache.org/surefire/maven-failsafe-plugin/examples/fork-options-and-parallel-execution.html)

flow-tests/test-pwa/src/test/java/com/vaadin/flow/pwatest/ui/PwaTestIT.java
```diff
@@ -202,22 +239,22 @@ private void checkIcons(List<WebElement> icons, int expected) {
...
-            HttpURLConnection.setFollowRedirects(false);
-            HttpURLConnection con = (HttpURLConnection) new URL(URLName)
-                    .openConnection();
-            con.setInstanceFollowRedirects(false);
-            con.setRequestMethod("HEAD");
-            return (con.getResponseCode() == HttpURLConnection.HTTP_OK);
-        } catch (Exception e) {
-            e.printStackTrace();
-            return false;
+    private void checkResources(String... resources) {
```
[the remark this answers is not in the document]
joheriks: That sounds like a good idea. ´Content-Length` is a bit iffy because resources can be served in `chunked` mode in which case the header is missing. What do you think about checking `Content-Type` and correlating it with the expected file (e.g. icon should be `image/png`). I guess a typical bug that would now go unnoticed is where we incorrectly serve a fallback `text/html` for a resource).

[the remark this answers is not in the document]
Haprog: `Content-Type` sounds like a good enough generic check when we're expecting something else than HTML. It would be a good improvement to these tests here I think.

flow-tests/test-pwa/src/test/java/com/vaadin/flow/pwatest/ui/PwaTestIT.java
Haprog: Ok, thanks for the explanation.

Haprog:
Code looks ok so far. Just need to make the new tests work reliably.

flow-tests/test-pwa/src/test/java/com/vaadin/flow/pwatest/ui/PwaTestIT.java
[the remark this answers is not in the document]
joheriks: Done. Now checks `Content-Type` of the atual served file if `URLConnection#guessContentTypeFromName` returns non-`null` for the resource (currently `null` only for `manifest.webmanifest`).

flow-tests/test-pwa/src/test/java/com/vaadin/flow/pwatest/ui/PwaTestIT.java
joheriks: Seems parallelism may have been a red herring, as the test was failing because of the service worker not installing due to the (now corrected) error I had in `PWARegistry` (when checking whether to precache `offlinePath`). Removed the forced sequencing to see if the tests pass. The race between SW installation and going offline (addressed in #9261) is also likely to be an issue here.

Haprog (approved):

Haprog (approved):

PR #9268, opened 2020-10-30
refactoring: move OSGi resource provider impl to flow-server

denis-anisimov:
OSGi `ResourceProvider`  impl should be available as a service at the moment when Vaadin WAB register a servlet.
It's not possible without extra unclear config to make sure that the service is registered if it's in the `flow-osgi` : the bundle may be activated at any moment regardless of servlet registration.
The issue doesn't appear if the service is registered at the `flow-bundle`  start phase.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/osgi/OSGiResourceProvider.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/osgi/OSGiResourceProvider.java
@@ -13,7 +13,7 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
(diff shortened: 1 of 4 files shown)


pleku (changes requested):
Probably would be good to leave some mark in either code or even in the commit message body on _why is it moved_

denis-anisimov:
> Probably would be good to leave some mark in either code or even in the commit message body on _why is it moved_

Not sure which commit message but I've updated the comment in the PR.
Is it enough or should I do something else ?

pleku (approved):

PR #9277, opened 2020-11-02
refactoring: get rid of OSGiClientStaticResource

denis-anisimov:


pleku:
@denis-anisimov there are conflicts

pleku (dismissed):
Just curious that what is the reason that we'll register a new resource servlet for each purpose instead of using just one for both client & push resources ?

denis-anisimov:
> Just curious that what is the reason that we'll register a new resource servlet for each purpose instead of using just one for both client & push resources ?

By the servlet you mean the servlet instance ?
The instance depends on bundle.
It may not be the same instance.

pleku (approved):

PR #9278, opened 2020-11-02
refactoring: App classpath resources should be access via ResourceProvider

### LINKED ISSUE #9269: OSGi: go through all the code, find getResource methods and decide whether they should be replaced with ResourceProvider
`Class::getResource`, `Class::getResourceAsStream`, `Classloader::getResource`, etc. methods in many cases are used to get a resource from the classpath which _ARE NOT_ in the same Jar/bundle (but may be in some other Jar/bundle which assumed to be in the Web Application dependencies).

This is not valid way to access the resources. It works only for WAR where the whole app is loaded by one classloader and it's possible to access the resources inside WAR via this classloader.

It doesn't work in OSGi. To be able to load resources properly `ResourceProvider`  abstraction is created .
Every resource access should be done via `ResourceProvider` unless the resource is located inside the same Jar/Bundle.

The ticket's goal is to find all usages of methods mentioned above (`Class::getResource`, etc.)  and check whether they access the resource in the same Jar or this is an access to  an app resource which should be accessed via `ResourceProvider`

denis-anisimov:
Fixes #9269

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
@@ -188,8 +188,8 @@ public class NpmTemplateParser implements TemplateParser {
                     url);
             content = appResource == null ? null : appResource.openStream();
         } catch (IOException exception) {
-            getLogger().warn("Coudln't get resource for the template '{}'",
-                    url);
+            getLogger().warn("Coudln't get resource for the template '{}'", url,
+                    exception);
         }
         if (content != null) {
             getLogger().debug(
--- a/flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
@@ -166,6 +166,8 @@ public class NpmTemplateParser implements TemplateParser {
     /**
      * Finds the JavaScript sources for given tag.
      *
+     * @param service
+     *            the related Vaadin service
      * @param tag
      *            the value of the {@link com.vaadin.flow.component.Tag}
      *            annotation, e.g. `my-component`
--- a/flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
@@ -17,6 +17,7 @@ package com.vaadin.flow.component.polymertemplate;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -30,6 +31,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.component.dependency.JsModule;
+import com.vaadin.flow.di.Lookup;
+import com.vaadin.flow.di.ResourceProvider;
 import com.vaadin.flow.function.DeploymentConfiguration;
 import com.vaadin.flow.internal.AnnotationReader;
 import com.vaadin.flow.internal.Pair;
@@ -66,8 +69,9 @@ public class NpmTemplateParser implements TemplateParser {
     private JsonObject jsonStats;
 
     /**
-     * The default constructor. Protected in order to prevent direct instantiation,
-     * but not private in order to allow mocking/overrides for testing purposes.
+     * The default constructor. Protected in order to prevent direct
+     * instantiation, but not private in order to allow mocking/overrides for
+     * testing purposes.
      */
     protected NpmTemplateParser() {
     }
@@ -104,7 +108,7 @@ public class NpmTemplateParser implements TemplateParser {
             }
 
             String url = dependency.getUrl();
-            String source = getSourcesFromTemplate(tag, url);
+            String source = getSourcesFromTemplate(service, tag, url);
             if (source == null) {
                 try {
                     source = getSourcesFromStats(service, url);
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
```diff
@@ -162,16 +166,31 @@ private boolean dependencyHasTagName(Dependency dependency, String tag) {
     /**
      * Finds the JavaScript sources for given tag.
      *
-     * @param tag the value of the {@link com.vaadin.flow.component.Tag} annotation,
-     *            e.g. `my-component`
-     * @param url the URL resolved according to the {@link com.vaadin.flow.component.dependency.JsModule}
-     *            spec, for example {@code ./view/my-view.js} or {@code @vaadin/vaadin-button.js}.
+     * @param tag
```
caalador, line 169: Missing service parameter from the javaDoc

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
```diff
@@ -162,16 +166,31 @@ private boolean dependencyHasTagName(Dependency dependency, String tag) {
...
      *         such source can be found.
      */
-    protected String getSourcesFromTemplate(String tag, String url) {
-        InputStream content = getClass().getClassLoader()
-                .getResourceAsStream(url);
+    protected String getSourcesFromTemplate(VaadinService service, String tag,
+            String url) {
+        Lookup lookup = service.getContext().getAttribute(Lookup.class);
+        ResourceProvider resourceProvider = lookup
+                .lookup(ResourceProvider.class);
```
caalador, line 184: Should we just have a method getResourceProvider in service so you wouldn't need to do this dance in every place as now it's basically the same code in every place and then we could have some null checks in place also.

Perhaps even have the shorthand of getApplicationResource in  service

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
```diff
@@ -162,16 +166,31 @@ private boolean dependencyHasTagName(Dependency dependency, String tag) {
...
+        InputStream content = null;
```
caalador, line 185: the inputStream should be closed after the streamToString call.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapUtils.java
```diff
@@ -301,13 +305,24 @@ static String getDependencyContents(VaadinRequest request, String file) {
...
+        URL appResource = resourceProvider.getApplicationResource(service,
+                file);
+
+        InputStream stream = null;
+        try {
+            stream = appResource == null ? null : appResource.openStream();
+        } catch (IOException e) {
+            throw new IllegalStateException(String.format(
+                    "Couldn't open application resource '%s' for inline resource",
+                    file, e));
```
caalador, line 320:
```suggestion
                    file), e);
```

flow-server/src/main/java/com/vaadin/flow/server/BootstrapUtils.java
```diff
@@ -462,26 +477,26 @@ private static JsonObject createInlineDependencyObject(String content) {
...
+                             */
+                            List<Class<? extends RouterLayout>> layouts = RouteUtil
+                                    .getParentLayoutsForNonRouteTarget(
+                                            errorNavigationTarget);
+                            if (layouts.isEmpty()) {
+                                return errorNavigationTarget;
+                            } else {
+                                return layouts.get(layouts.size() - 1);
+                            }
+                        });
```
caalador, line 494: Unrelated formatting change

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
denis-anisimov: `getApplicaitonResource` shouldn't be in the service for sure.
Just because this method may be called when there is no any service yet at all (and this is not the only method in `ResourceProvider`).

For the same reason I would avoid having `getResourceProvider`  in the service.
The only reason why `getResourceProvider`  could have been added into the service is just a shorthand.
But this is not a good reason. The shorthand can be a static method like `SomeUtilClass.getResourceProvider(VaadinSevrice)`. Not sure whether it's worth or not: I don't like static methods.

Overall : it's antipattern vs pattern and I'm not sure whether this is an antipattern or a pattern.
Technically there is no any logical reason to have `ResourceProvider` in the service (except a convenience method).
And declaring such method in the service litters API: there can be other services which are used in various places in the code like `ResourceProvider`. Should we add a getter for every such service ? And should we do this every time when such service appears? (the opposite is also true: we may decide to remove any service at any time).
Such (I would say antipattern) allows to dynamically get service in the same way as OSGi allows to do this without modifying API everytime when new service appears (or disappears).
It has advantage (described above: keep API the same) and disadvantage: one has to know that there is a service which can be accessed this way.
For me the advantage at the moment outbalances a bit disadvantages.

caalador: Perhaps we could have in ResourceProvider the method as `getXYZ(service)` the main concern is the moment you get a null on lookup which should be handled nicer than the current NPE that it will throw which in some cases is hard to know what was null for instance if the line is `service.getContext().getAttribute(Lookup.class).lookup(ResourceProvider.class)`

denis-anisimov: `ResourceProvider` is an interface.
It may not have any static method.

`service.getContext().getAttribute(Lookup.class)`  may not be `null` 
`lookup.lookup(ResourceProvider.class)` also may not be `null`.
(at least in runtime)

The code is done in the way that the instances always there because they are set in the servlet init phase.

But I don't really see a benefit having this in a separate method: you may not prevent NPE (NPE means everything is broken, it may just not happen, it's not possible to do a fallback). And if NPE happens (e.g. in the test) then how you may properly handle this ?
It just means that the test is not properly configured and I don't see any good message which may say something meaningful here.

denis-anisimov: The only benefit which I see for now is replacing NPE with some ISE or RuntimeException.
And I don't think replacing one exception with another is really a benefit.

May be it's worth to add a check in `getAttribute()`  method specifically for `Lookup.class`  parameter and log a warning but this is a separate ticket.

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/NpmTemplateParser.java
denis-anisimov: `streamToString`  should do this.
But not even this call : `IOUtils.toString`  should do this.
Quite unexpected that it's no done there.

denis-anisimov: Actually , 
```
public static String streamToString(InputStream inputStream) {
        String ret = "";
        try {
            return IOUtils.toString(inputStream, StandardCharsets.UTF_8)
                    .replaceAll("\\R", System.lineSeparator());
        } catch (IOException exception) {
            // ignore exception on close()
            LoggerFactory.getLogger(FrontendUtils.class)
                    .warn("Couldn't close template input stream", exception);
        }
        return ret;
    }
```

The code is wrong : it expects that `IOEcception`  happens on `close` only which is wrong.
And also it expects that `IOUtils.toString`  is closed based on the log message (I haven't found the code which closes the stream in apache utils).

So this is a separate issue. I'm relying here that the stream is closed inside this util method.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapUtils.java
denis-anisimov: There was changes in this file, formatting is done by Eclipse based on the standard rules.

This has been discussed without any decision again in the slack.

caalador (approved):

PR #9280, opened 2020-11-02
Fix navigation fallback

### LINKED ISSUE #9218: [service worker] Navigation fallback feature doesn't work in all cases
Navigation fallback feature which was implemented in https://github.com/vaadin/flow/pull/9183 doesn't seem to work in all cases.

It seems to only work if you first reload a view (other than `/`) at least once (after initial load that registered the service worker) while online and then go offline. Otherwise it still fails to load other than root URL when reloading offline.
Problem seems to be that it doesn't use the precached app shell (which is listed in Cache Storage like `/?__WB_REVISION__=d29685c11fdd4caf3f5ba2e313e484b8`) but instead the fallback strategy creates a runtime cache with an entry like `/` which is used only after it has been runtime cached.

Haprog:
Fixes #9218.

This makes navigation fallback (offline start) use the same service worker logic as we use for `offlinePath`, just using the app shell route `/` instead of the route to offline page for the cache entry to be returned. I did some manual testing and this does seem to fix the issues described in #9218.

The added IT is still a work in progress as it seems to be a false negative (always passes) and doesn't fail even without the fix. If I try to do the same manually when running the test view, it fails as it should. So the test still needs some work.


flow-server/src/main/resources/sw.ts
```diff
@@ -5,43 +5,33 @@ import {skipWaiting, clientsClaim} from 'workbox-core';
...
-  navigationFallback = new NavigationRoute(async (params: any) => {
-    // Use offlinePath fallback if offline was detected
-    if (!self.navigator.onLine) {
-      const offlinePathPrecachedResponse = await matchPrecache(offlinePath);
-      if (offlinePathPrecachedResponse) {
-        return offlinePathPrecachedResponse;
-      }
+const appShellPath = '/';
+// @ts-ignore: OFFLINE_PATH_ENABLED and OFFLINE_PATH are defined by webpack.generated.js
+const offlinePath = OFFLINE_PATH_ENABLED ? '/' + OFFLINE_PATH : appShellPath;
```
Haprog, line 15: This is the main change here.

The `else` block (which used to separately handle offline start situation) from previous `if (OFFLINE_PATH_ENABLED) {` is removed. The main condition is now on this line. Otherwise the contents that were inside the `if` is kept as is.

Haprog:
Should we keep the current test, ignore it or add some comment before merging?

haijian-vaadin (approved):
Agreed that we can merge this one first since it works, the test can be updated in
https://github.com/vaadin/flow/issues/9314

haijian-vaadin:
@Haprog, since the test passes, I think we don't need to ignore it. We can reference this PR in #9314 that test should be updated.

PR #9297, opened 2020-11-03
Cherry Picks to 2.4

taefi:


flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchyMapper.java
```diff
@@ -625,6 +626,16 @@ public void destroyAllData() {
...
+        return !expandedItems.isEmpty();
+    }
+
+    /**
+     * Returns the expanded items in form of an unmodifiable collection.
+     *
+     * @return an unmodifiable {@code Collection<T>} containing the expanded
+     * items.
+     */
+    public Collection<T> getExpandedItems() {
```
mshabarov, line 638: This is a breaking change `0.1.0` and is not appropriate for `2.4.X`. Can we replace this new method with a reflection, or by using `isExpanded` to loop through the items and find all the expanded ones?

mshabarov (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchyMapper.java
taefi: Good point. Both suggestions have pros. and cons. `2.4.x` is a LTS, reflection is nastier and actually we kinda did this to avoid reflection workarounds, but it is more performant than looping over items and call the `isExpanded`. Personally, I choose performance for the LTS, but at the cost of reflection?! I'm not sure.

PR #9298, opened 2020-11-03
refactor: make web browser immutable

fluorumlabs:
Even though `WebBrowser` is updated only once via `com.vaadin.flow.server.VaadinService#createAndRegisterSession()`, it still has `public void updateRequestDetails()` method which may be called by the third party. `WebBrowser` instance is returned without defensive copy from `VaadinSession.getBrowser()`. This PR makes `WebBrowser` effectively immutable and removes `updateRequestDetails()` method.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
@@ -925,7 +925,7 @@ public abstract class VaadinService implements Serializable {
         storeSession(session, request.getWrappedSession());
 
         // Initial WebBrowser data comes from the request
-        session.getBrowser().updateRequestDetails(request);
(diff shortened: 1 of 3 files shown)


fluorumlabs:
TC build failed with some unrelated JavaDoc error. No idea how to proceed

joheriks (approved):
Making `WebBrowser` immutable makes a lot of sense (that it currently isn't is probably a relic from Framework when a second request updated additional properties (screen size, `window.name` for `@PreserveOnRefresh`,...) which is since moved to `ExtendedClientDetails`).

joheriks:
When cherry-picking to LTS, retain the method `updateRequestDetails()` as `@Deprecated`.

mshabarov:
Cherry picked to `5.0`

PR #9309, opened 2020-11-04
generate sw.ts to frontend folder when pwa enabled

### LINKED ISSUE #9143: sw.ts improvements
**Move sw.ts template to frontend directory to align with the existing frontend entry points**

Expected: `sw.ts` is located in the `frontend/` directory, which is the webpack context directory.

Actual: `sw.ts` is added to the project root.

Per discussion: We align the sw.ts the same as index.ts and index.html. Generate in the target folder unless the user has a sw.ts in the frontend folder. There should be a log message, the same as index.ts. index.html

**Generate sw.ts only when `@PWA` annotation is present**
Currently, the sw.ts file is generated even for a non-PWA application, which is wrong. We should only generate the `sw.ts` file when the `@PWA` annotation is present.

haijian-vaadin:
Fixes #9143

- generate `sw.ts` to target folder instead of the project root folder
- generate `sw.ts` only when PWA is enabled


joheriks (changes requested):
From the ticket (#9143) I understood the intention is to generate `sw.ts` to the `target` folder:

> Per discussion: We align the sw.ts the same as index.ts and index.html. Generate in the target folder unless the user has a sw.ts in the frontend folder. There should be a log message, the same as index.ts. index.html

and only use `frontend/sw.ts` if they user has created it themselves.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskUpdateWebpackTest.java
```diff
@@ -263,6 +263,37 @@ public void should_setPwaEnabledTrue_when_Pwa()
...
+        webpackUpdater.execute();
+
+        File serviceWorkerFile = new File(frontendFolder,
+                SERVICE_WORKER_SRC);
+        Assert.assertFalse(serviceWorkerFile.exists());
+    }
+
+    @Test
+    public void should_generateServiceworkerFile_when_pwa(){
+        createWebpackUpdater();
```
joheriks, line 279: `createWebpackUpdater` is run by `setup` so not needed to run explicitly unless creating a new `PwaConfiguration` in the test.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskUpdateWebpackTest.java
```diff
@@ -263,6 +263,37 @@ public void should_setPwaEnabledTrue_when_Pwa()
...
+        webpackUpdater.execute();
+
+        File serviceWorkerFile = new File(frontendFolder,
+                SERVICE_WORKER_SRC);
+        Assert.assertTrue(serviceWorkerFile.exists());
+    }
+
+    @Test
+    public void should_Not_generateServiceworkerFile_inProjectRootFoler_when_pwa(){
+        createWebpackUpdater();
```
joheriks, line 289: `createWebpackUpdater` is run by `setup` so not needed to run explicitly unless creating a new `PwaConfiguration` in the test.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -619,6 +623,13 @@ private void addBootstrapTasks(Builder builder) {
         TaskGenerateTsDefinitions taskGenerateTsDefinitions = new TaskGenerateTsDefinitions(
                 builder.npmFolder);
         commands.add(taskGenerateTsDefinitions);
+
+        if (frontendDependencies != null) {
+            PwaConfiguration pwaConfiguration = frontendDependencies.getPwaConfiguration();
+            if (pwaConfiguration.isEnabled()) {
+                commands.add(new TaskGenerateServiceWorker(builder.frontendDirectory, outputDirectory));
+            }
+        }
```
joheriks, line 632: Would perhaps make sense to extract this duplicate logic into a private method, e.g. `addServiceWorkerTask(Builder bilder, FrontendDependencies frontendDependencies)`.

joheriks (changes requested):
´TaskGenerateServiceWorker` needs an exclusion in `com.vaadin.flow.testutil.ClassesSerializableTest` to pass the validation.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
haijian-vaadin: Removed the other code block which should've been deleted.

joheriks (approved):

PR #9311, opened 2020-11-04
chore: create deepClean profile

caalador:
Using the deepClean profile one can
clean up all the generated files and
node_modules folders from all test modules.

### DIFF
--- a/flow-tests/pom.xml
+++ b/flow-tests/pom.xml
@@ -453,12 +453,11 @@
                                     <fileset>
                                         <directory>${project.basedir}</directory>
                                         <includes>
-                                            <include>frontend/main.js</include>
                                             <include>package*.json</include>
                                             <include>webpack.*.js</include>
                                             <include>types.d.ts</include>
                                             <include>tsconfig.json</include>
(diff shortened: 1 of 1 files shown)


flow-tests/pom.xml
```diff
@@ -412,6 +412,33 @@
...
+                                <filesets>
+                                    <fileset>
+                                        <directory>${project.basedir}</directory>
+                                        <includes>
+                                            <include>frontend/main.js</include>
+                                            <include>package*.json</include>
+                                            <include>webpack.*.js</include>
+                                            <include>types.d.ts</include>
+                                            <include>tsconfig.json</include>
+                                            <include>pmpm-lock.yaml</include>
```
taefi, line 432: Shouldn't this be `pnpm-...` instead of `pmpm-...`?

taefi (changes requested):

flow-tests/pom.xml
caalador: yes. updated.
Also I don't see a main.js anymore anywhere.

taefi (approved):

PR #9313, opened 2020-11-04
feat: Theme annotation should be only on AppShellConfigurator

### LINKED ISSUE #9092: Disallow @Theme on views
### Description of the bug
Starting from Vaadin 15, you are not allowed to configure `@PWA` on a main view but you need to put it on your `AppShellConfigurator` class. For some reason, you are still allowed to add `@Theme` on a main view even though it cannot work if you specify different `@Theme` annotations on different views.

We should disallow `@Theme` on a view, in the same way as `@PWA`

caalador:
If an AppShellConfigurator is available the `@Theme` annotation 
should be only defined there.

Fixes #9092

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinAppShellInitializer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinAppShellInitializer.java
@@ -22,7 +22,6 @@ import javax.servlet.ServletException;
 import javax.servlet.ServletRegistration;
 import javax.servlet.annotation.HandlesTypes;
 import javax.servlet.annotation.WebListener;
-
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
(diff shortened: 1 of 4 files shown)


pleku (changes requested):
1) Theme annotation javadocs need updating.

2) I think we should throw a meaningful error for people who migrate to flow 6 / v19 and have the theme annotation in the wrong place, even after they change their code after we've broken it elsewise. The error should state that the Theme annotation should go to the appshellconfigurator.

caalador:
I felt there isn't much more information to add than the default:
```
Found app shell configuration annotations in non `AppShellConfigurator` classes.
Please create a custom class implementing `AppShellConfigurator` and move the following annotations to it:
 - Theme from [nonAppShellClass]
```

pleku (approved):

mshabarov:
Cherry picked to `5.0`

Issue #8655, opened 2020-06-30
Filtering and sorting for in-memory data should not delegate to data provider


pleku:
This is problematic behavior when looking at what a data view is and what is a data provider. When filtering and sorting is done through the data view, it should not modify the contents of another component that might be using the same data provider.
This is different for adding, removing and updating which are more data provider specific but still have API in data view.
### Acceptance Criteria
- all filtering and sorting methods in ListDataView should not effect the (list) data provider directly but apply those through the query instead
- when a new data provider is set to the component, the old filters & sorting is still discarded

#### Implementation notes
For a component with `DataCommunicator` this is more straight forward as it accepts both filter and sorting.
For other components they need to be made aware of this for when creating the query, probably by just using suitable getter(s) from the `AbstractListDataView` to fetch the sorting and filtering to use. But this would imply that the data view instance should be stored in the component memory. Another alternative is to use `ComponentUtil.setData` with the component instance to store the filtering & sorting.

pleku:
Thoughts and guesses after initial investigation:
- `AbstractListDataView` should store (and retrieve) the filter(s) and sorting(s) using `ComponentUtil.setData` to the component instance
- there needs to be a way to only notify only one component, by somehow firing `DataChangeEvent` through the data provider
- the filters and sorting should be set to either the `DataCommunicator` (it has a filter slot already) and it takes care of combining it with the component provided filter (CB) or sorting (Grid)
- for any other component that works always as in-memory, those need to be changed to apply the filtering&sorting when creating the query as (the data provider does not have any dependency to the component)

PR #9315, opened 2020-11-04
fix!: Store in-memory filtering and sorting in component

mshabarov:
In-memory filtering and sorting are now stored directly in component, which gives an opportunity to change it through the data view API for a certain component separately from other components bound to the same data provider.

Related-to: https://github.com/vaadin/flow/issues/8655


flow-data/src/main/java/com/vaadin/flow/data/provider/ComponentInMemoryFilter.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.vaadin.flow.function.SerializablePredicate;
+
+/**
+ * Stores the component's specific in-memory filter.
+ * 
+ * @param <T>
+ *            component's items type
+ */
+public class ComponentInMemoryFilter<T> implements Serializable {
```
pleku, line 29: Why is this class needed ? For storing the filter with certain key to component data ?
Is it necessary that it is public ?

flow-data/src/main/java/com/vaadin/flow/data/provider/ComponentSorting.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.vaadin.flow.function.SerializableComparator;
+
+/**
+ * Stores the component's specific sorting.
+ * 
+ * @param <T>
+ *            component's items type
+ */
+public class ComponentSorting<T> implements Serializable {
```
pleku, line 29: Why is this class needed ? For storing the comparator in component data ?
Is it necessary that it is public ?

flow-data/src/test/java/com/vaadin/flow/data/provider/AbstractListDataViewTest.java
```diff
@@ -948,6 +952,174 @@ public void createListDataProviderFromArrayOfItems_removingOneItem_itemCountShou
...
+                new Item[] { new Item(1L, "baz"), new Item(2L, "foo"),
+                        new Item(1L, "bar") },
+                listDataView1.getItems().toArray());
+
+        Assert.assertArrayEquals("Unexpected items sorting for component #2",
+                new Item[] { new Item(1L, "baz"), new Item(2L, "foo"),
+                        new Item(1L, "bar") },
+                listDataView2.getItems().toArray());
+    }
+
```
pleku, line 1122: Should there be a test that checks case that setting a list data provider filter is applied correctly (doesn't discard data view filter) and that it is combined with the data view one ? Probably for sorting too ?

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractListDataView.java
```diff
@@ -380,4 +411,34 @@ private void addItemCollectionOnTarget(Collection<T> items, T target,
...
+
+    @SuppressWarnings("unchecked")
+    private Optional<ComponentInMemoryFilter<T>> getComponentInMemoryFilter() {
+        return Optional.ofNullable(ComponentUtil.getData(component,
+                ComponentInMemoryFilter.class));
+    }
+
+    private void fireDataChangeEvent() {
+        ComponentUtil.fireEvent(component,
+                new ComponentDataChangeEvent<>(component));
```
pleku, line 442: So how is this working when none of the components react to this event ? Why not fire the `DataChangeEvent` instead ?

pleku (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractListDataView.java
```diff
@@ -380,4 +411,34 @@ private void addItemCollectionOnTarget(Collection<T> items, T target,
         itemList.addAll(indexToInsertItems, items);
         dataProvider.refreshAll();
     }
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    private Query getQuery(boolean filteringOnly) {
```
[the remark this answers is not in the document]
mshabarov: Yes, `withSorting ` much more clear.

flow-data/src/main/java/com/vaadin/flow/data/provider/ComponentInMemoryFilter.java
mshabarov: Yes, this is for storing in the component. I was considering to use a string key like this 

`ComponentUtil.setData(component, ListDataView.IN_MEMORY_FILTER_KEY,  filter);`

but adding a public string to `ListDataView` seems even worse maybe, because it is still public and it modifies the public interface. This key should be visible in both `AbstractListDataView` and in the `Component` (Component needs to restore this object internally in some places), so I am not able to find out the better approach than to make it just public.

flow-data/src/main/java/com/vaadin/flow/data/provider/ComponentSorting.java
mshabarov: The same comment as for filter.

flow-data/src/test/java/com/vaadin/flow/data/provider/AbstractListDataViewTest.java
mshabarov: Yeah, good point! I will add such a test!

flow-data/src/main/java/com/vaadin/flow/data/provider/ComponentInMemoryFilter.java
pleku: If it were hidden to a protected method for the abstract data view, then it would be possible to access it in a component that creates the data view. If it was package protected, then the data communicator could access it. So package-protected static method in the AbstractListDataView ? I would not expose two new public classes just as markers for storing data.

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractListDataView.java
mshabarov: > So how is this working when none of the components react to this event ?
I assumed that the components have to add a listener by themselves and to restore the filter/sorting by themselves. I avoided using a callback for notifying the components because it would be a breaking change for list data view abstract class. 

> Why not fire the DataChangeEvent instead ?
There are tow obstacles to do that:
1. `DataChangeEvent ` does not have a 'component' field, nor any other information to distinguish the target component. Adding such a filed means the breaking change in this class.
2. If we fire this event via data provider listener handler, then it would require the certain component mention in the public method -> breaking change. We're also not able to fire it via `ComponentUtil.fireEvent` because this event class does not extend `ComponentEvent`.

pleku: > Adding such a filed means the breaking change in this class.

I see, this doesn't matter, but I don't see how it is necessarily a breaking change with another constructor and exposing a new getter which might return `null`. There is new API added here anyway which makes it impossible to backport anyway. I don't know if it is really worth it to introduce dubious new API just for avoiding breaking something that we introduced in version 4.0 and can now change for 5.0.

But as to an alternative, I don't know either how the data provider could even distinguish the correct component to fire the data change event on. But it feels somehow bad API design to me as a component developer that if I have to _know and remember to do something extra_ when using `AbstractListDataView` instead of being required to do it implicitly in the constructor. Basically I could easily miss adding the listener to this event ^ and then my component would not work when filters & sorting changes.

So either
1) make the data provider fire the event for all components with the optional meta data of `Component getDataChangeComponent()` (better name would be nice) and then our components & data communicator (?) can ignore the events where the component is something different than itself.
2) force components using abstractlistdataview to always pass a callback for reacting to filter/sorting change

flow-data/src/main/java/com/vaadin/flow/data/provider/ComponentInMemoryFilter.java
mshabarov: Yeah, good idea! So I will add `protected static final String ...KEY = "...key";` to `AbstractListDataView` and then override it in component-specific data views, for instance, in `ComboBoxListDataView`, with a `public` modifier, and then just use in in `ComboBox` - `ComboBoxListDataView.KEY` and in data communicator - `AbstractListDataView.KEY`.

Sound good!

pleku: > Yeah, good idea! So I will add `protected static final String ...KEY = "...key";` to `AbstractListDataView` and then override it in component-specific data views, for instance, in `ComboBoxListDataView`, with a `public` modifier, and then just use in in `ComboBox` - `ComboBoxListDataView.KEY` and in data communicator - `AbstractListDataView.KEY`.
> 
> Sound good!

Sorry, I'm not sure I understand this. Why do you need to expose or override keys ? If you just create package protected static getter & setter method in  `AbstractListDataView` that takes in component and optionally the filter OR sorting. Why would a component specific key be needed ????

mshabarov: Right, the component and data communicator will just use those static setter and getter. And the string key will be hidden in the `AbstractListDataView`. Sorry for confusing.

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractListDataView.java
mshabarov: > But it feels somehow bad API design to me as a component developer that if I have to know and remember to do something extra when using AbstractListDataView instead of being required to do it implicitly in the constructor.

Agree! I would go with option 2 then, because it guarantee that the custom component won't forget to handle these filtering/sorting events.

pleku: >it guarantee that the custom component won't forget to handle these filtering/sorting events.

I don't think it guarantees it per se, as existing components often can forget to add those listener into their data providers.. but is maybe a better pattern. I wonder now why we didn't just expose an API in the `HasItems` or `HasDataProvider`  that would have been called by the data provider and even optionally the users to just tell the component to _refresh the data for itself_... I don't know what sense it made to require acquiring the data provider to refresh things. Probably just wrong design principles with too much emphasis on the data providers.

flow-data/src/main/java/com/vaadin/flow/data/provider/ListDataView.java
```diff
@@ -266,11 +266,10 @@
      * that has been set or added previously. {@code null} will clear all
      * filters.
      * <p>
-     * A filter bound to data set, not to the component. That means this filter
-     * won't be retained when a new data or {@link DataProvider} is set to the
-     * component. Any other component using the same {@link DataProvider} object
-     * would be affected by setting a filter through data view of another
-     * component.
+     * A filter bound to the component. Any other component using the same
```
pleku, line 269: > A filter bound to the component.

This is a bit odd way of continuing the javadocs at this point... maybe

> This filter is bound to the component.

would make it more natural.

flow-data/src/main/java/com/vaadin/flow/data/provider/ListDataView.java
```diff
@@ -286,11 +285,10 @@
      * Adds a filter to be applied to all queries. The filter will be used in
      * addition to any filter that has been set or added previously.
      * <p>
-     * A filter bound to data set, not to the component. That means this filter
-     * and previously added filters won't be retained when a new data or
-     * {@link DataProvider} is set to the component. Any other component using
-     * the same {@link DataProvider} object would be affected by adding a filter
-     * through data view of another component.
+     * A filter bound to the component. Any other component using the same
```
pleku, line 288: The same as above

flow-data/src/main/java/com/vaadin/flow/data/provider/ListDataView.java
```diff
@@ -315,11 +313,11 @@
      * Sets the comparator to use as the default sorting. This overrides the
      * sorting set by any other method that manipulates the default sorting.
      * <p>
-     * A comparator bound to data set, not to the component. That means the
-     * default sorting won't be retained when a new data or {@link DataProvider}
-     * is set to the component. Any other component using the same
-     * {@link DataProvider} object would be affected by setting a sort
-     * comparator through data view of another component.
+     * A comparator bound to the component. Any other component using the same
```
pleku, line 316: The same as above, seem off. Maybe

> This comparator is bound to the component. Thus any other...

would be better

flow-data/src/main/java/com/vaadin/flow/data/provider/ListDataView.java
```diff
@@ -337,11 +335,11 @@
      * comparator will be used to determine the ordering of items that are
      * considered equal by the previously defined default sorting.
      * <p>
-     * A comparator added to data set, not to the component. That means the
-     * default sorting won't be retained when a new data or {@link DataProvider}
-     * is set to the component. Any other component using the same
-     * {@link DataProvider} object would be affected by adding a sort comparator
-     * through data view of another component.
+     * A comparator bound to the component. Any other component using the same
```
pleku, line 338: Same as above

pleku (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractListDataView.java
```diff
@@ -380,4 +488,22 @@ private void addItemCollectionOnTarget(Collection<T> items, T target,
...
+
+        final Optional<SerializableComparator<T>> sorting = withSorting
+                ? getComponentSortComparator(component)
+                : Optional.empty();
+
+        return new Query(0, Integer.MAX_VALUE, null,
+                sorting.orElse(null), filter.orElse(null));
+    }
+
+    private void fireDataChangeEvent(SerializablePredicate<T> filter,
```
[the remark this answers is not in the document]
mshabarov: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/ListDataView.java
mshabarov: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/ListDataView.java
mshabarov: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/ListDataView.java
mshabarov: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/ListDataView.java
mshabarov: Done

pleku (approved):

PR #9319, opened 2020-11-04
Cherry Picks to 2.4

taefi:


flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -545,4 +560,27 @@ private JsonValue generateJsonForExpandedOrCollapsedItem(T item) {
...
 
+    private Collection<T> getHierarchyMapperExpandedItems() {
+        HierarchyMapper<T, ?> hierarchyMapper = getHierarchyMapper();
+        if (!hierarchyMapper.hasExpandedItems()) {
+            return Collections.emptySet();
+        }
+
+        Collection<T> expandedItems = new HashSet<>();
+        getHierarchyMapper().fetchRootItems(null)
+                .forEach(root -> expandedItems.addAll(getExpandedItems(root)));
```
pleku, line 571: This goes through all the expanded items that `HierarchyMapper` says are expanded.
Does the hierarchy mapper have items only related to the active range (on client side) or ? Just wondering what is the cost of this operation. I bet that it throws out items that are not in the active range (done by `DataCommunicator`), but just want to verify as this look suspicious and there is no comment in the code.

mshabarov: `HierarchyMapper` does not currently know anything about whether the expanded item is active or not. We could force the `HierarchyMapper` to hold only the active items somehow, but it looks like another fix/improvement.

mshabarov: And another point is that the scroll position is not preserved (after attach/detach/refresh page). So, if we were in the middle of `TreeGrid` and the active items list contains the items in the middle, and then we refresh the page and come back to 0 position...I would expect the items in the 0 position also preserve the expanding state, though they could be not in active range before page refreshing. Does it make sense?

taefi: `HierarchyMapper#fetchRootItems(null)` ignores the range passing null, and fetches all the root nodes. Since it has `fetchXXXItems` set of methods accepting a `Range` object, I believe it stores the whole hierarchy rather than the active range.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -545,4 +560,27 @@ private JsonValue generateJsonForExpandedOrCollapsedItem(T item) {
         return json;
     }
 
+    private Collection<T> getHierarchyMapperExpandedItems() {
+        HierarchyMapper<T, ?> hierarchyMapper = getHierarchyMapper();
+        if (!hierarchyMapper.hasExpandedItems()) {
+            return Collections.unmodifiableCollection(Collections.emptySet());
```
mshabarov, line 566: `Collections.emptySet()` already produces empty unmodifiable set, as I recall, why do you wrap it again?

taefi: Correct, my bad. Undid it.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
taefi: I think the cost of the operation is not good for big data obviously, and @mshabarov's discussion also make sense. I think the cost of traversing the hierarchical data can be omitted by having that public method in `HierarchyMapper` for the next minor version (2.5).

mshabarov (dismissed):

mshabarov (approved):

mshabarov (approved):

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
pleku: So is it either done already for 2.5 and master changes or another issue opened so that it is not forgotten ?

Probably would be also good to acknowledge the resolution to this discussion in the code as a comment to make it easier for maintainers. Like `// FIXME This will cause xyz, to be fixed by #1234`

pleku: Or well, this is already 2.4 branch so it makes little impact here to add the comment. Would have been different for master but didn't notice it then

taefi: The public method is already added in `master` and `5.0` branches, but not in `2.5`. So is it fine if I cherry pick the master changes directly to `2.5` instead of the `// FIXME` comment?

pleku: Ok

taefi: [This issue](https://github.com/vaadin/flow/issues/9351) should address this for the next minor version branches `2.5` and `4.1`.

Issue #9248, opened 2020-10-28
Forwardport OSGi support from feature/osgi to be based on Flow 5.0/6.0


pleku:
- it should be based on flow 6 (v19) but until that snapshot is available, go with 5
- new repository osgi (can be renamed later on if necessary)
- flow-osgi module should be there
- the ITs and validation (PR & snapshot) builds should be running on top of that and snapshot should a build trigger for each night
- the license is CVAL 4.0 
- for now there is no need to add a license checker, will be handled separately

PR #9321, opened 2020-11-05
Forwardporting OSGi feature branch to the master

denis-anisimov:
Part of #9248


denis-anisimov (commented):
This should not be merged yet. There are still things that needs to be done: e.g. move `flow-osgi` along with OSGi ITs into a separate repo.

But I would like to see validation errors now since there are a lot changes in various places: the OSGi feature branch adaptation to the `master` requires significant efforts.

denis-anisimov:
> This should not be merged yet. There are still things that needs to be done: e.g. move `flow-osgi` along with OSGi ITs into a separate repo.
> 
> But I would like to see validation errors now since there are a lot changes in various places: the OSGi feature branch adaptation to the `master` requires significant efforts.

Ready for review.

flow-lit-template/bnd.bnd
```diff
@@ -3,5 +3,6 @@ Bundle-Name: Vaadin Flow Lit Templates Support
 Bundle-Version: ${osgi.bundle.version}
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-License: http://www.apache.org/licenses/LICENSE-2.0
+Import-Package: *
```
stbischof, line 6: I know that you use this in most `bnd.bnd` files. But not sure why -  its the default

> The Import-Package header lists the packages that are required by the contained packages. The default for this header is *, resulting in importing all referred packages. 

[from bnd docs](https://bnd.bndtools.org/heads/import_package.html)

flow-polymer-template/pom.xml
```diff
@@ -90,6 +90,13 @@
             <version>${osgi.core.version}</version>
             <scope>provided</scope>
         </dependency>
+        
+        <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>osgi.cmpn</artifactId>
+            <version>${osgi.compendium.version}</version>
+            <scope>provided</scope>
```
stbischof, line 98: i think the scope is compile.

flow-polymer-template/src/main/java/com/vaadin/flow/component/polymertemplate/rpc/OSGiPolymerPublishedEventHandlerRegistration.java
```diff
@@ -0,0 +1,56 @@
...
+ * @since
+ *
+ */
+@Component(immediate = true)
+public class OSGiPolymerPublishedEventHandlerRegistration {
+
+    private ServiceRegistration<DeprecatedPolymerPublishedEventHandler> registration;
+
+    @Activate
+    void activate() {
```
stbischof, line 41: You could a BundleContext as Parameter in the method. The ServiceComponentRuntime will inject it.

flow-lit-template/bnd.bnd
denis-anisimov: Thank you.
Didn't know that this is the default.

flow-polymer-template/pom.xml
denis-anisimov: Not really.
Or may be it would be better to set this dependency as optional....

The idea here is : this Jar should be used without any OSGi.
And that's exactly the reason why I've introduced `OSGiPolymerPublishedEventHandlerRegistration` class instead of declaring `PolymerPublishedEventRpcHandler` directly as a service impl.
It allows to use `PolymerPublishedEventRpcHandler` as is so that it doesn't depend on OSGi at all (no any references to OSGi annotations).
And `OSGiPolymerPublishedEventHandlerRegistration`  will be used only in OSGi, otherwise this class will be completely ignored.

I think I prefer `provided`  vs `optional`.

flow-polymer-template/src/main/java/com/vaadin/flow/component/polymertemplate/rpc/OSGiPolymerPublishedEventHandlerRegistration.java
denis-anisimov: Thank you.
I assumed that it should work this way but was too lazy to check that in the spec.

flow-server/bnd.bnd
```diff
@@ -4,8 +4,9 @@ Bundle-Version: ${osgi.bundle.version}
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-License: http://www.apache.org/licenses/LICENSE-2.0
 Import-Package: org.atmosphere*;resolution:=optional;version='${atmosphere.runtime.version}',\
-    org.apache.http*;resolution:=optional;,\
-  *
+    org.apache.http*;resolution:=optional;,com.github.javaparser*;resolution:=optional;,\
+    io.swagger*;resolution:=optional;,javax.annotation;version=1.3.2,\
```
stbischof, line 8: maybe just only  io.swagger.codegen an optional. oll the other are afaik theother io.swagger jars are osgi bundles

stbischof:
would be good to update the bnd version.

flow-server/bnd.bnd
denis-anisimov: Well, there are many packages.
And the reason is semantical here: all `io.swagger` packages are used by `vaadin-connect` functionality which is 
executed either via Maven goal or via `DevModeInitializer`  which is explicitly excluded in OSGi.

So in the end : the whole Flow functionality which requires `io.swagger`  doesn't work in OSGi and there is just no sense to keep some dependencies non optional.

At some point all these deps will be excluded from `flow-server`  along with this functionality which is not supposed to work in OSGi at all.

denis-anisimov:
> would be good to update the bnd version.

done

stbischof:
The bnd update needs this lines in the tests pom


https://github.com/vaadin/flow/blob/014de584c5f44dae8b91fca80a5e072419426e6c/flow-tests/pom.xml#L75

denis-anisimov:
> The bnd update needs this lines in the tests pom
> 
> https://github.com/vaadin/flow/blob/014de584c5f44dae8b91fca80a5e072419426e6c/flow-tests/pom.xml#L75

OK, thanks.

flow-polymer-template/src/main/java/com/vaadin/flow/component/polymertemplate/rpc/OSGiPolymerPublishedEventHandlerRegistration.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+@Component(immediate = true)
+public class OSGiPolymerPublishedEventHandlerRegistration {
+
+    private ServiceRegistration<DeprecatedPolymerPublishedEventHandler> registration;
+
+    @Activate
+    void activate(BundleContext context) {
+        context.registerService(DeprecatedPolymerPublishedEventHandler.class,
+                new PolymerPublishedEventRpcHandler(), null);
```
stbischof, line 43: you could remove this class if you declare PolymerPublishedEventRpcHandler als component 
```
@Component(immidiate=true,service=PolymerPublishedEventRpcHandler.class)
public class PolymerPublishedEventRpcHandler
```

denis-anisimov:
> > The bnd update needs this lines in the tests pom
> > https://github.com/vaadin/flow/blob/014de584c5f44dae8b91fca80a5e072419426e6c/flow-tests/pom.xml#L75
> 
> OK, thanks.

Reverted since the latter changes breaks everything without a clear reason.
Don't know which change exactly but these changes are extra here and can be done later on.

denis-anisimov:
> > would be good to update the bnd version.
> 
> done


Reverted since the latter changes breaks everything without a clear reason.
Don't know which change exactly but these changes are extra here and can be done later on.

flow-polymer-template/src/main/java/com/vaadin/flow/component/polymertemplate/rpc/OSGiPolymerPublishedEventHandlerRegistration.java
denis-anisimov: Already explained here: 
https://github.com/vaadin/flow/pull/9321/files#r519569258

denis-anisimov: Already explained here: https://github.com/vaadin/flow/pull/9321/files#r519569258

stbischof:
i will add the commits on top of this and have a look what breaks it

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
```diff
@@ -464,6 +474,69 @@ public void createInitParameters_readDevModeProperties() throws Exception {
...
+            throws VaadinConfigurationException, IOException {
+        VaadinContext context = Mockito.mock(VaadinContext.class);
+        VaadinConfig config = Mockito.mock(VaadinConfig.class);
+
+        ResourceProvider resourceProvider = mockResourceProvider(config,
+                context);
+
+        String path = VAADIN_SERVLET_RESOURCES + TOKEN_FILE;
+
+        File tmpFile = temporaryFolder.newFile();
```
caalador, line 507: Is the temp file necesary? We could just mock the URL and return an expected path.

flow-server/src/test/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandlerTest.java
```diff
@@ -41,13 +51,14 @@
...
+import com.vaadin.flow.shared.ApplicationConstants;
 import com.vaadin.tests.util.MockDeploymentConfiguration;
 
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.eq;
-
 public class WebComponentBootstrapHandlerTest {
 
+    @Rule
+    public final TemporaryFolder tmpDir = new TemporaryFolder();
```
caalador, line 60: Drop un-used rule and imports

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendUtilsTest.java
```diff
@@ -264,31 +273,89 @@ public void commandToString_commandContainsBaseDir_baseDirIsReplaced() {
...
+        Lookup lookup = Mockito.mock(Lookup.class);
+        context.setAttribute(Lookup.class, lookup);
+
+        ResourceProvider provider = Mockito.mock(ResourceProvider.class);
+
+        Mockito.when(lookup.lookup(ResourceProvider.class))
+                .thenReturn(provider);
+
+        if (content != null) {
+            File tmpFile = tmpDir.newFile();
```
caalador, line 349: This should also be mockable to not need to have the temporary file by returning a mock url that for `openStream()` returns the stream.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/PublishedServerEventHandlerRpcHandler.java
```diff
@@ -280,50 +284,36 @@ private static Object decodeArg(Component instance, Method method,
...
-                                convertedType)) {
-                    return polymerUtil
-                            .getMethod("getTemplateItem", Component.class,
-                                    JsonObject.class, Type.class)
-                            .invoke(polymerUtil, instance, argValue,
-                                    method.getGenericParameterTypes()[index]);
+            if (instance != null) {
+                Optional<UI> ui = instance.getUI();
+                assert ui
+                        .isPresent() : "Rpc handler may not be called for a component that is not attached";
```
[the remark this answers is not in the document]
denis-anisimov: This code is designed to work only with `StateNode`  which are attached to the `StateTree`: RPC may not be executed for a Node which is not in the tree.

So I would say `assertion` should be enough since it just checks that the assumption correct.
Exception should be thrown on a user mistake. Our internal mistake (or incorrect assumption) should be checked with assertions.
But I don't see a reason to have a strong position here on my side.
Let's make it throwing.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -832,20 +827,47 @@ private String getClientEngineUrl(BootstrapContext context) {
             final boolean productionMode = context.getSession()
                     .getConfiguration().isProductionMode();
 
-            boolean resolveNow = !productionMode || getClientEngine() == null;
+            ResourceProvider resourceProvider = context.getSession()
+                    .getService().getContext().getAttribute(Lookup.class)
+                    .lookup(ResourceProvider.class);
```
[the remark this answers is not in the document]
denis-anisimov: `BootstrapContext` is quite hidden internal class.
And `ResourceProvider` is used only once in the code.
So any specific reason to declare a shorthand ?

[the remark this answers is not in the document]
caalador: Readability is the main issue.

flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
```diff
@@ -326,26 +337,45 @@ private static String getResourceFromFile(Properties initParameters)
...
+                    classResources.size() + contextResources.size());
+            resources.addAll(classResources);
+            resources.addAll(contextResources);
+        }
+
         // Accept resource that doesn't contain
         // 'jar!/META-INF/Vaadin/config/flow-build-info.json'
         URL resource = resources.stream()
                 .filter(url -> !url.getPath().endsWith("jar!/" + tokenResource))
                 .findFirst().orElse(null);
```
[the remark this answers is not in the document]
denis-anisimov: Doesn't work.
Because it fixes only `URL resource =` line.

But then below there is a check for `resources.isEmpty()` which may not be done with the same stream.
It might be this can be simplified but it's out of scope of this PR.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -429,8 +435,13 @@ private boolean ensureCookiesEnabled(VaadinServletRequest request,
      *             if the application is denied access to the persistent data
      *             store represented by the given URL.
      *
+     *             <<<<<<< HEAD =======
+     * @deprecated As of 1.0. Will be removed in 3.0.
+     *
+     *             >>>>>>> 81a7e2d15a... Fix OSGi Lookup related issues
```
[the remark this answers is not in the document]
denis-anisimov: Good point.

flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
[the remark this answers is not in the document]
caalador: But the stream of an empty list is a "empty stream" (e.g. no-op) so it makes streams of the 2 lists (that are never null) and gets the same end result as the if with the isEmpty checks and new list creation.

flow-server/src/main/java/com/vaadin/flow/server/startup/LookupInitializer.java
```diff
@@ -0,0 +1,304 @@
...
+        implements ClassLoaderAwareServletContainerInitializer {
+
+    private static final String SPI = " SPI: ";
+
+    private static final String ONE_IMPL_REQUIRED = ". Only one implementation should be registered. "
+            + "Use lookupAll to get all instances of the given type.";
+
+    private static final String SEVERAL_IMPLS = "Found several implementations in the classpath for ";
+
+    private static class LookupImpl implements Lookup {
```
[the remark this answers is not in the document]
denis-anisimov: Similar question has been already answered in the original review of this code for `2.5` branch.

`LookupImpl` : I may extract this class to its own file. But why ?
This class is instantiated here directly and there is no any single reason to override it because overridden class won't be ever instantiated.

Implementation is done private explicitly. And this obviously says that this may not be overriden /changed.
Just because it's impossible.
And that's the reason why it's done `private` and nested.
The code is self-explanatory . 

Lookup is an entry point for "everything". It exactly allows to override any service (if it's done properly).
And this is the only thing which may not be overridden: because it give the way to override other things.

Other classes are made private for same reasons and e.g. `ResourceProviderImpl` may not be done `public` because otherwise it will be collected by the `LookupInitializer` and there will be no any way to override it.

Shortly : this functionality is explicitly made non-overridable because it gives the way to override everything.

denis-anisimov (commented):
-

flow-server/src/main/java/com/vaadin/flow/server/startup/LookupInitializer.java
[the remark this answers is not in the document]
caalador: So please add a javadoc with the information for both internal impl classes.

flow-server/src/test/java/com/vaadin/flow/server/BootstrapHandlerTest.java
```diff
@@ -1379,21 +1386,26 @@ public void getBootstrapPage_removesExportScript() throws ServiceException {
...
-
         String statsJson = "{\n" + " \"errors\": [],\n" + " \"warnings\": [],\n"
                 + " \"assetsByChunkName\": {\n" + "  \"bundle\": [\n"
                 + "    \"build/vaadin-bundle-e77008557c8d410bf0dc.cache.js\",\n"
                 + "    \"build/vaadin-bundle-e77008557c8d410bf0dc.cache.js.map\"\n"
                 + "  ],\n" + " }" + "}";
 
-        Mockito.when(classLoader.getResourceAsStream(Mockito.anyString()))
-                .thenReturn(new ByteArrayInputStream(statsJson.getBytes()));
+        File tmpFile = tmpDir.newFile();
```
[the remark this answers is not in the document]
denis-anisimov: Can we mock `final` classes?
I thought we can't.

I will check.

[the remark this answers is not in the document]
denis-anisimov: `final` class may not be mocked.

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
denis-anisimov: final class may not be mocked.

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendUtilsTest.java
denis-anisimov: final class may not be mocked.

flow-server/src/test/java/com/vaadin/flow/server/BootstrapHandlerTest.java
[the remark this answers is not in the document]
caalador: Right so the other possibility would be to either have a WrapperUrl that is mocked or   mock(URLConnection.class);
e.g. something like:
```
final URLConnection mockUrlCon = mock(URLConnection.class);

ByteArrayInputStream is = new ByteArrayInputStream(
        "<myList></myList>".getBytes("UTF-8"));
doReturn(is).when(mockUrlCon).getInputStream();

URLStreamHandler stubUrlHandler = new URLStreamHandler() {
    @Override
     protected URLConnection openConnection(URL u) throws IOException {
        return mockUrlCon;
     }            
};
URL url = new URL("foo", "bar", 99, "/foobar", stubUrlHandler);
```

Which is in many ways worse... Lets go with the temp file.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -48,12 +47,18 @@
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
+import javax.servlet.Servlet;
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletResponse;
```
pleku, line 52: This is also broken, javax. should be at the top

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -16,9 +16,7 @@
 
 package com.vaadin.flow.server;
 
-import javax.servlet.Servlet;
-import javax.servlet.ServletContext;
-import javax.servlet.http.HttpServletResponse;
+import static java.nio.charset.StandardCharsets.UTF_8;
```
pleku, line 19: Your import ordering is broken ?
Static imports should be last https://sites.google.com/a/vaadin.com/fw-flow-team-s-wiki/contributing-code/code-style/e

pleku (changes requested):
I noticed that your changes are touching imports and breaking the agreed ordering. Please fix this for everything

flow-server/src/test/java/com/vaadin/flow/server/MockVaadinServletService.java
```diff
@@ -39,24 +37,41 @@
...
-            DeploymentConfiguration deploymentConfiguration) {
-        super(servlet, deploymentConfiguration);
-
-        try {
-            servlet.init(new MockServletConfig());
-        } catch (ServletException e) {
-            throw new RuntimeException(e);
-        }
+        super(new MockVaadinServlet(deploymentConfiguration),
+                deploymentConfiguration);
```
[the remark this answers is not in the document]
denis-anisimov: `servlet.init` is almost not needed in the tests.
That's why I've removed it.

But `VaadinService::init` is needed everywhere.
So yes, good point.

flow-polymer-template/src/test/java/com/vaadin/flow/component/PolymerClassesSerializableTest.java
```diff
@@ -18,7 +18,8 @@
                 "com\\.vaadin\\.flow\\.component\\.polymertemplate\\.TemplateParser(\\$.*)?",
                 "com\\.vaadin\\.flow\\.component\\.polymertemplate\\.InjectablePolymerElementInitializer",
                 "com\\.vaadin\\.flow\\.component\\.polymertemplate\\.TemplateDataAnalyzer",
-                "com\\.vaadin\\.flow\\.component\\.polymertemplate\\.IdCollector"),
+                "com\\.vaadin\\.flow\\.component\\.polymertemplate\\.IdCollector",
+                "com\\.vaadin\\.flow\\.component\\.polymertemplate\\.rpc\\.OSGiPolymerPublishedEventHandlerRegistration"),
```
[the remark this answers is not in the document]
denis-anisimov: This class has been removed at some point.
So now it's not needed.
Just a leftover, thanks.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
denis-anisimov: I had no `flow.importorder`  imported in my Eclipse apparently.
Done.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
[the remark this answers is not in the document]
denis-anisimov: Ah, you mean just private method.
I thought `public` or `protected`  method for reuse.

OK.

flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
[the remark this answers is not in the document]
denis-anisimov: I think I don't understand you.

One stream instance may be used once. Stream is not interchangeable with List anyhow.
`resourceStream.filter....` closes the stream in the end.
Which means that `isEmpty`  or in fact `stream.count()` may not be called anymore: stream is closed.
That's why the list is used : it's possible to call `isEmpty` regardless of any other list operations.

Please provide the whole code snippet for the method instead of change for the snippet above if you still see how this can be imprioved.

flow-server/src/main/java/com/vaadin/flow/server/startup/LookupInitializer.java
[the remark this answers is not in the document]
denis-anisimov: Which kind of information?

I don't understand . These are impl classes.
The contract (interfaces) are described in details in javadocs of interface.
Impl classes just impl classes (one of impl).

flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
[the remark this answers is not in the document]
caalador: Right didn't see the reuse of the thing later on in the class. then I would just have it as
`Collection<URL> resources = Stream.of(classResources, contextResources).flatMap(Collection::stream).collect(Collectors.toList());`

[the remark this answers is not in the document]
denis-anisimov: Done without `flatMap`.

flow-server/src/test/java/com/vaadin/flow/server/VaadinServiceTest.java
```diff
@@ -15,42 +15,47 @@
  */
 package com.vaadin.flow.server;
```
caalador, line 16: This still doesn't have the fixed imports after Pekkas comment

flow-server/src/test/java/com/vaadin/flow/server/VaadinSessionTest.java
```diff
@@ -93,18 +88,15 @@ public Router getRouter() {
...
-            service.init();
             return service;
         }
     }
 
     @Before
     public void setup() throws Exception {
         httpSessionLock = new ReentrantLock();
-        mockServletConfig = new MockServletConfig();
-        mockServlet = new TestServlet();
```
caalador, line 105: The private static class TestServlet can be removed as it's not used anymore.

flow-server/src/test/java/com/vaadin/flow/server/communication/StreamReceiverHandlerTest.java
```diff
@@ -1,11 +1,9 @@
 package com.vaadin.flow.server.communication;
```
caalador, line 2: Imports

flow-server/src/test/java/com/vaadin/flow/server/communication/StreamRequestHandlerTest.java
```diff
@@ -1,13 +1,14 @@
 package com.vaadin.flow.server.communication;
 
-import javax.servlet.ServletConfig;
+import static com.vaadin.flow.server.communication.StreamRequestHandler.DYN_RES_PREFIX;
```
caalador, line 3: import position

flow-server/src/test/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandlerTest.java
```diff
@@ -15,6 +15,12 @@
  */
 package com.vaadin.flow.server.communication;
 
+import static com.vaadin.flow.server.Constants.STATISTICS_JSON_DEFAULT;
+import static com.vaadin.flow.server.Constants.VAADIN_SERVLET_RESOURCES;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.eq;
```
caalador, line 21: import position

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendUtilsTest.java
```diff
@@ -15,11 +15,16 @@
  */
 package com.vaadin.flow.server.frontend;
 
-import java.io.ByteArrayInputStream;
+import static com.vaadin.flow.server.Constants.STATISTICS_JSON_DEFAULT;
+import static com.vaadin.flow.server.Constants.VAADIN_SERVLET_RESOURCES;
+import static com.vaadin.flow.server.InitParameters.SERVLET_PARAMETER_STATISTICS_JSON;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
caalador, line 22: import position

caalador (changes requested):

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/startup/ApplicationRouteRegistry.java
```diff
@@ -67,6 +231,25 @@ protected ApplicationRouteRegistry() {
...
+     *            the vaadin context for which to get a route registry, not
+     *            <code>null</code>
+     * @return a registry instance for the given servlet context, not
+     *         <code>null</code>
+     * @deprecated this is deprecated in favor of
+     *             {@code getInstance(VaadinContext)} and will be removed in a
+     *             future release
+     */
+    @Deprecated
+    public static ApplicationRouteRegistry getInstance(ServletContext context) {
```
pleku, line 249: Again, adding back deprecated code is not that nice ... but OK for now if we address it soon anyway

flow-server/src/main/java/com/vaadin/flow/server/startup/ClassLoaderAwareServletContainerInitializer.java
```diff
@@ -15,15 +15,18 @@
...
-import javax.servlet.ServletException;
-
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Set;
 import java.util.stream.Stream;
 
+import javax.servlet.ServletContainerInitializer;
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
```
pleku, line 25: Wrong import order

flow-server/src/main/java/com/vaadin/flow/server/startup/LookupInitializer.java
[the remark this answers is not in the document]
pleku: > Which kind of information?

Answer review questions in code.

> Shortly : this functionality is explicitly made non-overridable because it gives the way to override everything.

Yes this is obvious to you as you've done this. It was not obvious for me or Mikael so we asked to document it.
Unless you do that, it will not be obvious to anyone else outside this conversation either. So spending a little time on documenting the code with comments to save time in the future.

flow-tests/test-root-context/pom.xml
```diff
@@ -11,9 +11,6 @@
     <artifactId>flow-test-root-context-npm</artifactId>
     <name>Flow root context tests in NPM dev mode</name>
     <packaging>war</packaging>
-    <properties>
-        <maven.deploy.skip>true</maven.deploy.skip>
-    </properties>
```
pleku, line 16: Uh, shouldn't this still be applied for the module not to be released/deployed with Flow ?
I understand that it is needed for snapshots due to being reused in tests, but probably not for named releases ? And thus this configuration should be still done for a release profile ?

flow-tests/pom.xml
```diff
@@ -280,6 +310,7 @@
             <id>release</id>
             <modules>
                 <module>test-mixed/pom-pnpm-production.xml</module>
+                <module>test-root-context</module>
```
pleku, line 313: I don't think we should release these modules other than for snapshots. Is it so ?

flow-server/src/test/java/com/vaadin/flow/server/startup/VaadinAppShellInitializerTest.java
```diff
@@ -15,6 +17,11 @@
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 
+import javax.annotation.concurrent.NotThreadSafe;
+import javax.servlet.ServletContext;
+import javax.servlet.ServletRegistration;
+import javax.servlet.http.HttpServletRequest;
```
pleku, line 23: Wrong import ordering

flow-server/src/test/java/com/vaadin/flow/server/startup/ServletDeployerTest.java
```diff
@@ -1,13 +1,17 @@
...
+import static org.easymock.EasyMock.anyBoolean;
+import static org.easymock.EasyMock.anyObject;
+import static org.easymock.EasyMock.anyString;
+import static org.easymock.EasyMock.capture;
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.mock;
+import static org.easymock.EasyMock.newCapture;
+import static org.easymock.EasyMock.replay;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
```
pleku, line 14: Wrong import ordering

flow-server/src/test/java/com/vaadin/flow/component/UITest.java
```diff
@@ -1,7 +1,10 @@
 package com.vaadin.flow.component;
 
-import javax.servlet.ServletConfig;
-import javax.servlet.ServletException;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+
```
pleku, line 7: Wrong import ordering

flow-server/src/test/java/com/vaadin/flow/server/BootstrapHandlerDependenciesTest.java
```diff
@@ -1,16 +1,28 @@
...
+import static org.hamcrest.CoreMatchers.either;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.Matchers.lessThan;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
```
pleku, line 14: Wrong import ordering

flow-server/src/test/java/com/vaadin/flow/server/BootstrapHandlerPushConfigurationTest.java
```diff
@@ -1,13 +1,19 @@
 package com.vaadin.flow.server;
 
-import javax.servlet.http.HttpServletRequest;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.mock;
```
pleku, line 7: Wrong import ordering

flow-server/src/test/java/com/vaadin/flow/server/BootstrapHandlerTest.java
```diff
@@ -1,8 +1,12 @@
 package com.vaadin.flow.server;
 
-import javax.servlet.http.HttpServletRequest;
+import static com.vaadin.flow.server.Constants.VAADIN_MAPPING;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
pleku, line 6: Wrong import ordering

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
```diff
@@ -1,14 +1,29 @@
...
+import static com.vaadin.flow.server.frontend.FrontendUtils.TOKEN_FILE;
+import static java.util.Collections.emptyMap;
+import static org.easymock.EasyMock.anyObject;
+import static org.easymock.EasyMock.capture;
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.mock;
+import static org.easymock.EasyMock.replay;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
pleku, line 16: Wrong import ordering

flow-server/src/test/java/com/vaadin/flow/server/I18NProviderTest.java
```diff
@@ -15,32 +15,36 @@
  */
 package com.vaadin.flow.server;
 
-import javax.servlet.ServletConfig;
+import java.util.Locale;
+import java.util.concurrent.locks.ReentrantLock;
+
```
pleku, line 20: Wrong import ordering

pleku (changes requested):
Please fix the introduced unnecessary import ordering change for all test files, not just the ones that I had the strength to check and comment on

flow-server/src/main/java/com/vaadin/flow/server/startup/ApplicationRouteRegistry.java
denis-anisimov: This is actually done my accident.
Code is reintroduced from `2.5`  but now this method is not used internally and indeed should be removed.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -407,11 +413,31 @@ protected Instantiator createInstantiator() throws ServiceException {
...
+        Lookup lookup = getContext().getAttribute(Lookup.class);
+        List<Instantiator> instantiators = null;
+        if (lookup != null) {
+            // lookup may be null in tests
+            Collection<InstantiatorFactory> factories = lookup
+                    .lookupAll(InstantiatorFactory.class);
+            instantiators = new ArrayList<>(factories.size());
+            for (InstantiatorFactory factory : factories) {
+                Instantiator instantiator = factory.createInstantitor(this);
+                if (instantiator != null && instantiator.init(this)) {
```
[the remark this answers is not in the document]
denis-anisimov: I hope you understand that it means my intention is failed: it should have been done for `2.5` and now this change should be back-ported after forwardporting. This is exactly what I've been trying to avoid.

This should be done by factory.
But the method exists. And if it's implemented we should use it I suppose.
Though I'm not sure: and as result I consider the comment absolutely useless since both approaches makes sense and the code is written using one or another way.
But sure, let's add more comments which nobody reads.

flow-server/src/main/java/com/vaadin/flow/server/osgi/VaadinBundleTracker.java
```diff
@@ -0,0 +1,314 @@
...
+
+        @Override
+        protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+                throws ServletException, IOException {
+            String pathInfo = req.getPathInfo();
+            if (pathInfo == null) {
+                resp.setStatus(HttpURLConnection.HTTP_NOT_FOUND);
+                return;
+            }
+            URL resource = bundle.getResource(resourceDirPath + pathInfo);
```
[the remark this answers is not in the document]
denis-anisimov: And again this code has been already reviewed in `2.5`.

This is done only for two bundles: `flow-client`  and `flow-push`.
`flow-push` contains only two resources (and nothing else) and both of them should be exposed.
`flow-client` contains client engine compiled files + `ts` files and `map` files. Do you see any reason why we should not allow to access any resource in this bundle ?

Do you suggest add a check for the `pathInfo` to avoid navigation to parent folders ?

I would not touch this for now because this should be rewritten via HTTP whiteboard resource registration which should take care about this in a separate PR.

flow-server/src/main/java/com/vaadin/flow/server/osgi/VaadinBundleTracker.java
```diff
@@ -0,0 +1,314 @@
...
+
+    private boolean isVaadinExtender(Bundle bundle) {
+        return !flowServerBundle.equals(bundle) && Boolean.TRUE.toString()
+                .equals(bundle.getHeaders().get("Vaadin-OSGi-Extender"));
+    }
+
+    private void handleFlowServerClassError(String className,
+            Throwable throwable) {
+        LoggerFactory.getLogger(VaadinBundleTracker.class)
+                .trace("Couldn't load class '{}'", className, throwable);
```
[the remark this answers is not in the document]
denis-anisimov: Perhaps, you mean by total duplication  that `trace` method is totally the same as `warn`  method.
But I think I disagree.
And this code here is for ages : https://github.com/vaadin/flow/blob/2.4/flow-server/src/main/java/com/vaadin/flow/server/osgi/VaadinBundleTracker.java#L137
https://github.com/vaadin/flow/blob/2.4/flow-server/src/main/java/com/vaadin/flow/server/osgi/VaadinBundleTracker.java#L143
This is reintroduced code.
So you are again reviewing the code in Nth time.

flow-server/src/main/java/com/vaadin/flow/server/startup/ApplicationRouteRegistry.java
```diff
@@ -52,6 +65,157 @@
...
+                    event -> update(() -> applyChange(event))));
+        }
+
+        private void applyChange(RoutesChangedEvent event) {
+            final RouteConfiguration routeConfiguration = RouteConfiguration
+                    .forRegistry(this);
+            Exception caught = null;
+            for (RouteBaseData<?> data : event.getRemovedRoutes()) {
+                caught = modifyRoute(() -> routeConfiguration
+                        .removeRoute(data.getUrl(), data.getNavigationTarget()),
```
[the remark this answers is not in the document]
denis-anisimov: This is reintroduced code.
I would suggest to avoid (re-)reviewing the code which is not new but has been removed and then added back overall.

This is exactly the reason why I've not done any change : to keep the code the same as in `2.5`.

But one more reason do not touch this: it will be completely removed once OSGi code moved to `osgi` repo.

flow-server/src/main/java/com/vaadin/flow/server/startup/ApplicationRouteRegistry.java
denis-anisimov: This is actually done by accident.
Code is reintroduced from 2.5 but now this method is not used internally and indeed should be removed.

flow-server/src/main/java/com/vaadin/flow/server/startup/LookupInitializer.java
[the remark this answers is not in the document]
denis-anisimov: There is a reason why we are writing the code using Java.
It allows to write the code in typesafe way, clear contract and hide things which should not be exposed.
All these things may be done so that compiler may check this.
Without this JS may be used to write the code: it doesn't provide any safety.

It's possible to use 4 visibilities : `public`, `protected` , package local and `private`.
If something is defined as `private` then this is intentionally made non-extendable and intentionally hidden from everyone. And this is enforced by the compiler.
Looks quite obvious. And the same obvious things are requested to be added as comment. That's why 
I really don't understand why.
We don't add any comment for a private field why it's `private` : it's obvious why it's private.
Here the internal impl class implements interface which defines a contract and since it's private: it's an implementation details which explicitly is not exposed.
So it's absolutely obvious for me why the nested class is private and what it means.
But apparently I'm the only one for whom this is obvious.
So OK, let's add comment which nobody reads.

flow-tests/test-root-context/pom.xml
denis-anisimov: It's hard to answer to the questions which are not related to the code at all without testing this.
And I'm not able to test this because this PR is under review for several weeks.

Yes, the module should be deployed as a snapshot. And yes, it should not be deployed for a regular release.
I'be been trying to prevent deploying it for the regular release via config options in the build jobs.
But looks like it was wrong. At least it turns out the `SNAPSHOT` doesn't even use `release` profile.

I made one more profile `snapshot` and the module `test-root-context` is not anymore in `release` profile which should prevent releasing it.

flow-tests/pom.xml
denis-anisimov: Moved to a dedicated profile.

flow-server/src/test/java/com/vaadin/flow/server/VaadinServiceTest.java
```diff
@@ -111,8 +117,8 @@ public void should_reported_routing_hybrid() {
                 e -> Constants.STATISTIC_ROUTING_SERVER.equals(e.getName())));
     }
 
-    @Test
-    public void testFireSessionDestroy() throws ServletException {
+    public void testFireSessionDestroy()
```
pleku, line 120: Why was this test removed ?

pleku (changes requested):
Just one odd change

flow-server/src/test/java/com/vaadin/flow/server/VaadinServiceTest.java
denis-anisimov: Good question.
No idea. Perhaps merge conflict consequence.
Returned back.

flow-server/src/test/java/com/vaadin/flow/server/VaadinServiceTest.java
```diff
@@ -47,10 +51,6 @@
 import com.vaadin.flow.server.communication.StreamRequestHandler;
 import com.vaadin.tests.util.MockDeploymentConfiguration;
 
-import static org.hamcrest.CoreMatchers.containsString;
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.junit.Assert.assertEquals;
-
```
pleku, line 47: How is it that your imports are again going to the wrong location and causing unnecessary changes ?

pleku (approved):

PR #9329, opened 2020-11-05
test(TypeScript): ensure CSRF token is updated when session is invalidated

### LINKED ISSUE #9164: Fusion: Random 'Anonymous access is not allowed' error for @AnonymousAllowed endpoint calls.
Some people have reported that they randomly got the `Anonymous access is not allowed` error for @AnonymousAllowed endpoint calls. The error looks like in the attached screenshot.

The difficult part is that it's not reproducible. From the code in `VaadinConnectAccessChecker`, it might come from the CSRF token check.
```
    private String verifyAnonymousUser(Method method,
            HttpServletRequest request) {
        if (!getSecurityTarget(method)
                .isAnnotationPresent(AnonymousAllowed.class)
                || cannotAccessMethod(method, request)) {
            return "Anonymous access is not allowed";
        }
        return null;
    
    private boolean cannotAccessMethod(Method method,
            HttpServletRequest request) {
        return requestForbidden(request)
                || entityForbidden(getSecurityTarget(method), request);
    }

    private boolean requestForbidden(HttpServletRequest request) {
        if (!xsrfProtectionEnabled) {
            return false;
        }

        HttpSession session = request.getSession(false);
        if (session == null) {
            return false;
        }

        String csrfTokenInSession = (String) session.getAttribute(VaadinService.getCsrfTokenAttributeName());
        return csrfTokenInSession == null || !csrfTokenInSession.equals(request.getHeader("X-CSRF-Token"));
    }
```

One observation: it happened today after the server has started and waited for about 6 minutes, then when trying to submit a request, I got the `Anonymous access is not allowed` error.

DoD:
1. Separate the error messages, 1 error message with 2 different conditions confuses the user.
When `!getSecurityTarget(method)`, return "Anonymous access is not allowed".
When `cannotAccessMethod(method, request)`, return "CSRF token has expired" or sth similar.
2. The root cause is identified and fixed.

platosha:
Fixes #9164

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/form/index.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/form/index.ts
@@ -9,5 +9,5 @@ $wnd.Vaadin = $wnd.Vaadin || {};
 $wnd.Vaadin.registrations = $wnd.Vaadin.registrations || [];
 $wnd.Vaadin.registrations.push({
   is: '@vaadin/form',
-  version: /* updated-by-script */ '5.0-SNAPSHOT'
+  version: /* updated-by-script */ '6.0-SNAPSHOT'
 });
--- a/flow-server/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
@@ -16,17 +16,19 @@
 
 package com.vaadin.flow.server.connect.auth;
 
-import com.vaadin.flow.server.VaadinService;
-
 import javax.annotation.security.DenyAll;
 import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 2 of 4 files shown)


haijian-vaadin (changes requested):
Can we also separate the error message as described in the ticket?

> Separate the error messages, 1 error message with 2 different conditions confuses the user.
> When !getSecurityTarget(method), return "Anonymous access is not allowed".
> When cannotAccessMethod(method, request), return "CSRF token has expired" or sth similar.

knoobie:
@haijian-vaadin I'm not sure if it's good that different error messages are returned to the client side. Having both return 403 should be more secure in the long run. This would reduce the footprinting and attacking vector (see OWASP for more details). Nevertheless, having different error message on the server side (like debug or info) logging would be good :)

haijian-vaadin:
Good point @knoobie, thanks. @platosha, let's add some server-side log then.

haijian-vaadin (approved):

mshabarov:
Cherry picked to `5.0`

mshabarov:
Cherry Picked to `4.0`

Issue #9203, opened 2020-10-20
design: a visual offline indicator


vlukashov:
When interacting with the app offline I want to see a visual 'offline' indication rather than an infinite loading indicator, so that I know that the app is offline and some app features may be limited.

This is a design task and has space for defining the exact user experience. The use cases are:
 - The browser looses the network connection when the app is already open and the user is in the middle of doing something, but no network request is in progress.
   - For a **client-side view** my UX expectation is that the app detects the online / offline status change and displays a permanent visual indication of the 'offline' state, but lets the user to continue working with the view. The developer may disable some UI elements while the app is offline.
   - For a **server-side view** my UX expectation is that the app detects the online / offline status change and displays a permanent visual indication of the 'offline' state, and blocks any user input until the connection is restored. Once the connection is restored the user input is unblocked.
 - The browser looses the network connection when the app is already open and the user has pressed a button, and is waiting for the response. The loading indicator is visible.
   - For a **client-side view** my UX expectation is that the loading indicator disappears (and gets replaced with the offline indicator). The network operation fails with a network error and the developer is expected to handle it. The user is free to interact with the app further, though the developer may disable some UI elements while the app is offline.
   - For a **server-side view** my UX expectation remains the same as with V17 and before: the app shows a 'reconnecting...' message and blocks any user input until the connection is restored. Once the connection is restored the pending request is re-sent and the app continues from where it was interrupted.
- The browser looses the network connection when the app is already open. The offline indicator is visible. Then the user navigates to another view.
   - For a **client-side view** my UX expectation is that the URL is updated and the new view is rendered. If it makes any network requests on the first render or in response to user actions, they are handled as described above.
   - For a **server-side view** my UX expectation is that the URL is updated, but instead of rendering the actual view content, the app shell renders a 'you are offline' stub view (as described in https://github.com/vaadin/flow/issues/9132). Once the connection is restored the view is rendered (unless the user has already moved to another view).

Current behavior with Vaadin 18.0.0.alpha1 / Flow 5.0.offline-SNAPSHOT:
![Screen Recording 2020-10-20 at 11 01 AM](https://user-images.githubusercontent.com/22416150/96561858-c4d53b80-12c8-11eb-8b98-bff63b2a6d36.gif)

jouni:
A quick prototype for one idea: https://output.jsbin.com/hurulub

Not really usable on touch devices, so those would need additional care.

Some thoughts:

[...]

PR #9325, opened 2020-11-05
chore: consolidated offline navigation tests to single class

joheriks:
Moved offline stub test to module `test-ccdm-flow-navigation` and unified test naming (`given_when_then`-scheme).

Fixes #9121

PR #9335, opened 2020-11-06
fix: do not attempt to call `$server.leaveNavigation` in offline case

### LINKED ISSUE #9326: Server-to-client navigation results in infinite loading indicator
Steps to reproduce:
1. Open a server-side view in a Vaadin PWA that has both server- and client-side views.
1. Go offline.
1. Try to navigate to a client-side view.

Observed behavior:
- Blinking loading indicator because `Flow.ts` attempts to call `flowRoot.$server.leaveNavigation`.

Expected behavior:
- The clients-side view is loaded (from SW cache). 

Note: IT implemented in #9325 as part of #9121, need to remove the `@Ignore`d test in #9325

joheriks:
Check `navigator.onLine` value before calling server. Note that when he have the
new "single source of truth" (part of #9203) for the connection state that should 
be used instead.

Fixes #9326.

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts
@@ -148,8 +148,8 @@ export class Flow {
       ctx: NavigationParameters,
       cmd?: PreventCommands): Promise<any> {
 
-    // server -> server or at the offline page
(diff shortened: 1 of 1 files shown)


joheriks:
Note: IT in #9325.

haijian-vaadin (approved):

PR #9377, opened 2020-11-11
feat: application connection state store

### LINKED ISSUE #9343: Fusion: Application Connection State store
As a developer, I want to have a single source of truth about the current application state, so that I can e.g. show a connection state indicator correspondingly. 

In Flow, there are different causes for a connection state change, xhr request, heartbeat, reconnecting attempts, push. 
In Fusion, it can be caused by an endpoint call, navigator.online, navigator.offline.

DoD:
Have a `ConnectionStateStore` class in `ConnectionSate.ts` file, which holds the state and should update the state based on the statechart described in #9203.
```
class ConnectionStateStore {
   // connection state could be but not limited to: connected, loading, loaded, reconnecting, or connection lost
   state: ConnectionState;
   
   getState():  ConnectionState{
     return sate
   }

   addStateChangeHandler()
   
   goOnline()
   goOffline()
   // methods in ConnectionStateHandler
}
export const connectionStateStore = new ConnectionStateStore();
```

Note, take a look at [ConnectionStateHandler](https://github.com/vaadin/flow/blob/master/flow-client/src/main/java/com/vaadin/client/communication/ConnectionStateHandler.java), see how Flow handles state transitions.

Part of #9203

joheriks:
Adds a global `ConnectionStateStore` instance as `window.Vaadin.Flow.connectionState`.
This instance should be used by both Flow (GWT client) and Fusion (`Flow.ts` and
related TS modules) for querying and transitioning between connection states (`CONNECTED`,
`LOADING`,  `RECONNECTING` and `CONNECTION_LOST`). The state store will be used for the 
implementation of the up-and-coming loading indicator. 

Fixes #9343.


flow-client/src/main/resources/META-INF/resources/frontend/ConnectionState.ts
```diff
@@ -0,0 +1,72 @@
...
+  CONNECTION_LOST = 'connection-lost'
+}
+
+type ConnectionStateChangeListener = (previous: ConnectionState, current: ConnectionState) => void;
+
+export class ConnectionStateStore {
+
+  private state: ConnectionState;
+
+  private stateChangeListeners: ConnectionStateChangeListener[] = [];
```
platosha, line 35: BTW: Using `Set` instead of array for storage works better for this use case, in particular, `addStateChangeListener` and `removeStateChangeListener` methods would map directly to `stateChangeListeners.add` and `stateChangeListeners.remove`.

Iterating a `Set` could be done with `for...of`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

flow-client/src/main/resources/META-INF/resources/frontend/ConnectionState.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  constructor(initialState: ConnectionState) {
+      this.state = initialState;
+  }
+
+  getState(): ConnectionState {
```
platosha, line 41: Should we do `get state()` `set state()` instead of `getState()`/`setState()` maybe?

flow-client/src/main/resources/META-INF/resources/frontend/ConnectionState.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  setState(newState: ConnectionState) {
+    if (newState !== this.state) {
+      const prevState = this.state;
+      this.state = newState;
+      this.stateChangeListeners.forEach(listener => listener(prevState, this.state));
+    }
+  }
+
+  isOnline(): boolean {
```
platosha, line 64: Maybe `get online(): boolean` should be preferred

flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts
```diff
@@ -404,19 +394,34 @@ export class Flow {
...
       } else {
         loading.setAttribute('style', 'none');
+        // Make Testbench know that server request is no longer active
+        this.isActive = false;
       }
-    };
+    });
+
+    $wnd.addEventListener('online', () => {
+      connectionState.setState(ConnectionState.CONNECTED);
```
platosha, line 419: Placing these handlers in `Flow.ts` makes them inactive before the first server-side view is requested, right? Was that intentional?

joheriks: This code is invoked from the constructor of `Flow`, and this object is always created (in `index.ts`) even when starting from a client-side view AFAIK.

flow-client/src/main/resources/META-INF/resources/frontend/ConnectionState.ts
joheriks: Done.

flow-client/src/main/resources/META-INF/resources/frontend/ConnectionState.ts
joheriks: Agreed, done.

flow-client/src/main/resources/META-INF/resources/frontend/ConnectionState.ts
joheriks: Done.

joheriks:
Currently, the global connection state store (`window.Vaadin.Flow.connectionState`) is imported and instantiated via `generated-flow-imports.js` rather than from `Flow.ts`. The sole reason for this is to load it also in V14 bootstrapping mode, where `Flow.ts` is not loaded.

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -92,6 +93,7 @@
  *
  * @since 1.0
  */
+@JsModule("@vaadin/flow-frontend/LoadingIndicator.js")
```
haijian-vaadin, line 96: Do we really need this? Isn't the `LoadingIndicator` imported in `Flow.ts`?

haijian-vaadin (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/UI.java
joheriks: `Flow.ts` is not loaded in legacy bootstrapping mode. Agree it's not very nice to have a `JsModule` annotation here, open to suggestions for other ways of solving the backwards compatibility issue.

haijian-vaadin (approved):

PR #9391, opened 2020-11-12
fix!: Store in-memory filtering and sorting in component

mshabarov:
In-memory filtering and sorting are now stored directly in component, which gives an opportunity to change it through the data view API for a certain component separately from other components bound to the same data provider.

Related-to: #8655


flow-data/src/main/java/com/vaadin/flow/data/provider/DataViewUtils.java
```diff
@@ -0,0 +1,129 @@
...
+                filter);
+    }
+
+    /**
+     * Sets the in-memory sort comparator to a given component instance.
+     *
+     * @param component
+     *            component instance the sort comparator is bound to
+     * @param sortComparator
+     *            component's in-memory sort comparator
```
pleku, line 96: apparently this could be null to remove it ?

flow-data/src/main/java/com/vaadin/flow/data/provider/DataViewUtils.java
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @param component
+     *            component instance the filter is bound to
+     * @param filter
+     *            component's in-memory filter
+     * @param <T>
+     *            items type
+     */
+    public static <T> void setComponentFilter(Component component,
+            SerializablePredicate<T> filter) {
```
pleku, line 85: apparently this could be null to remove it ?

pleku (changes requested):
Two copy paste errors (?) as pointed out by sonarqube. I think we could manage even without the methods.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataViewUtils.java
mshabarov: Yeah, added the explanations about it

flow-data/src/main/java/com/vaadin/flow/data/provider/DataViewUtils.java
mshabarov: Yeah, added the explanations about it

mshabarov:
Methods for removing both filter and sorting, and query getter have been extracted to utils as well, so as to remove more boilerplate code from components.

pleku (approved):

pleku (approved):

PR #9392, opened 2020-11-12
fix: check if Url path is encoded twice in Dev Mode

mshabarov:
Checks double encoding in Dev Mode requests paths, and if the given path contains unsafe double encoding, it then doesn't serve the request.


flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
```diff
@@ -305,6 +305,13 @@ public boolean serveDevModeRequest(HttpServletRequest request,
         // a valid request for webpack-dev-server should start with '/VAADIN/'
         String requestFilename = request.getPathInfo();
 
+        if (HandlerHelper.isPathUnsafe(requestFilename)) {
```
denis-anisimov, line 308: Two questions here:

- `serveDevModeRequest`  looks too late for checking.
I think this may be preserved _BUT_ this could be just double check.
- why it sets `HttpServletResponse.SC_NOT_MODIFIED` ? 

Clarification: 
`isDevModeRequest` should be called before `serveDevModeRequest` (if I understand correctly).
The method should return just `false`  and then some other handler will be used to handle a request or some status (404 most likely , but this doesn't matter) will be set.
Double check here can be done but then the code should just throw: since `isDevModeRequest` should be checked first by design and `serveDevModeRequest` shouldn't be ever called.
If it's called then our imp is broken and we should just throw.

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -176,4 +189,25 @@ public static String getCancelingRelativePath(String pathToCancel) {
...
+    /**
+     * Checks if the given URL path contains the symbols encoded twice in
+     * hexadecimal format, which can be injected maliciously.
+     *
+     * @param path
+     *            the URL path to be verified.
+     * @return {@code true}, if the given path has a double encoding,
+     *         {@code false} otherwise.
+     */
+    public static boolean isPathUnsafe(String path) {
```
denis-anisimov, line 201: If I remember correctly @caalador mentioned that we already has similar code somewhere.
If it's true then the code should be reused (may be extracted as an utility method but then some existing code should be modified as well since it already contains it).

denis-anisimov: Ah, sorry, I just realized that this is exactly done.

Just ignore this.

denis-anisimov (changes requested):

denis-anisimov:
Please add a reference to the ticket in the commit message: I can't see which ticket the PR fixes.

flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
mshabarov: >The method should return just false and then some other handler will be used to handle a request or some status

I agree to make a path check more early, in `isDevModeRequest`, but if we just return `false` and delegate to another handler, then the path check occurs twice. Is that correct and why not just terminate the request handling?

>why it sets HttpServletResponse.SC_NOT_MODIFIED ?

Good question and I also thinking about that. The response code should reflect the semantics of rejecting the request, some maybe `403 Forbidden` is better here.

>Double check here can be done but then the code should just throw: since isDevModeRequest should be checked first by design and serveDevModeRequest shouldn't be ever called.
If it's called then our imp is broken and we should just throw.

Agree!  `serveDevModeRequest` potentially could be called directly, because it's public, so let double check and throw if needed.

flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
```diff
@@ -575,6 +585,29 @@ public void devModeNotReady_handleRequest_returnsHtml() throws Exception {
         Mockito.verify(response).setContentType("text/html;charset=utf-8");
     }
 
+    @Parameterized.Parameters
+    public static Collection<String> pathsWithDirectoryChange() {
```
denis-anisimov, line 589: I would rename the method to have different name than `pathsWithDirectoryChange`: they have quite different nature (at least the types are different) event though they are connected.

Another thing : I'm pretty sure that `DevModeHandlerTest` should not be made parameterized .
As I understand when you parameterize the class its instance is created per every parameter _AND_ for every test method.
So if you have `N` parameters and `M` test methods then there will be `N*M` executed tests.
Since this class contains tests which don't need any parameters this is very excessive and will increase the time of execution dramatically.

So: either the method should accept parameter and there should be just one test method per parameter (hardcoded) or the class should be refactored to contain only test methods which really requires parameters.

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
```diff
@@ -537,34 +547,25 @@ private void staticBuildResourceWithDirectoryChange_nothingServed(
...
-                "/VAADIN/build/something\\..\\vaadin-bundle-1234.cache.js");
-    }
-
-    @Test
-    public void staticBuildResourceWithDirectoryChangeWithEncodedBackslashUpperCase_nothingServed()
-            throws IOException {
-        staticBuildResourceWithDirectoryChange_nothingServed(
-                "/VAADIN/build/something%5C..%5Cvaadin-bundle-1234.cache.js");
+    @Parameterized.Parameters
+    public static Collection<String> paths() {
```
denis-anisimov, line 555: Same thing here: check whether this class should be refactored

denis-anisimov (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
mshabarov: done

flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
mshabarov: done

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+import com.vaadin.flow.testutil.ChromeBrowserTest;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class UrlValidationIT extends ChromeBrowserTest {
```
denis-anisimov, line 27: `test-root-context`  doesn't use dev mode.
it compiles the resources and web pack dev server is not started.

It should not be changed because these tests are used for OSGi which is not able to run webpack dev server at all.
So the IT which relies on webpack dev server will fail there.

There are a bunch of modules which are executed in dev mode. E.g. `test-dev-mode`  which is explicitly made for dev mode. Please use another module for the test.

denis-anisimov (changes requested):

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
mshabarov: Moved to test-dev-mode, thanks!

denis-anisimov (approved):

PR #9406, opened 2020-11-16
feat: Only handle annotated theme

### LINKED ISSUE #9383: Send theme name to application-theme-plugin
Make it possible to give the name of the used theme to the application-theme-plugin so that it only handles the applicable themes and not all found themes from theme folders.

caalador:
Now we only handle the theme with
the name inside the Theme annotation.

Fixes #9383


flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -51,17 +49,30 @@ class ApplicationThemePlugin {
...
-      this.options.themeProjectFolders.forEach((themeProjectFolder) => {
-        if (fs.existsSync(themeProjectFolder)) {
-          logger.debug("Found themeFolder from ", themeProjectFolder);
-          handleThemes(themeProjectFolder, this.options.projectStaticAssetsOutputFolder);
+        if (fs.existsSync(this.options.themeResourceFolder)) {
+          logger.debug("Found themeFolder in jar file ", this.options.themeResourceFolder);
+          handleThemes(themeName, this.options.themeResourceFolder, this.options.projectStaticAssetsOutputFolder);
         }
-      });
+
```
pleku, line 66: So now when _theme-name_ is discovered from both inside a `.jar` and inside the `/theme/` folder in the project both will be loaded. While I cannot figure out at the moment why it would be a bad thing yet, the _spec_ says that when you have something inside a `.jar`, it should be used from there. If you have something inside  the `/theme/` folder in the project, it should be used there. But not both. If the in-project theme defines a parent theme, that should be loaded before the in-project theme.

So we should probably decide (and document) how it goes when you have this situation that the theme is present both in-project and inside a .jar.

pleku: Probably throwing and instructing the in-project theme to use the parent theme inside theme.json is the thing to do.

In context of this ticket, maybe it would be fine to just first check the in-project location and if it is not there, then the `.jar`. Or even now the `.jar` could be commented out until we re-enable it with parent theme support ...

pleku (changes requested):

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
caalador: Only handling frontend theme and trowing if found also from "jar"

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -51,17 +49,30 @@ class ApplicationThemePlugin {
...
+        this.options.themeProjectFolders.forEach((themeProjectFolder) => {
+          if (fs.existsSync(themeProjectFolder)) {
+            logger.debug("Found themeFolder from ", themeProjectFolder);
+            handleThemes(themeName, themeProjectFolder, this.options.projectStaticAssetsOutputFolder);
+          }
+        });
+      } else {
+        logger.log("No '", generatedThemeFile, "' found. Skipping application theme handling.");
```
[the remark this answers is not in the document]
caalador: Changed log level and added extra trace with more information.

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -53,25 +56,39 @@ class ApplicationThemePlugin {
...
         }
 
-        this.options.themeProjectFolders.forEach((themeProjectFolder) => {
-          if (fs.existsSync(themeProjectFolder)) {
-            logger.debug("Found themeFolder from ", themeProjectFolder);
-            handleThemes(themeName, themeProjectFolder, this.options.projectStaticAssetsOutputFolder);
+        if (fs.existsSync(this.options.themeResourceFolder)) {
+          if (themeFound && fs.existsSync(path.resolve(this.options.themeResourceFolder, themeName))) {
+            throw new Error("Theme '" + themeName + "'should not exist inside a jar and in the project at the same time\n" +
+              "If extending the jar theme the 'parent' theme feature should be used");
```
pleku, line 84: > "If extending the jar theme the 'parent' theme feature should be used")

I'm not sure if this is enough for anyone who doesn't know about the it already. Maybe
"Extending another theme is possible by adding `{ "parent": "my-parent-theme" }`  entry to the theme.json file inside your theme's folder." would be explicit

pleku (changes requested):

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
caalador: 👍

pleku (approved):

Issue #9397, opened 2020-11-13
Get null pointer exception when refresh multiple times


nloke:
### Description of the bug
I am using preserveOnRefresh and there are a lot of widgets in the form. When I refresh multiple times, the server throws a NPE also the UI console log will show:
Response didn't contain a server id. Please verify that the server is up-to-date and that the response data has not been modified in transmission.
### Minimal reproducible example
We don't have a minimal reproducible example. It happening in our production code. We talked to your expert via prime chat and he suggested to file a bug based on the stack trace.
### Expected behavior
No NPE after the refresh.
### Actual behavior
**Server error stack trace:**
java.lang.NullPointerException: null
	at com.vaadin.flow.component.internal.UIInternals.addJavaScriptInvocation(UIInternals.java:549)
	at com.vaadin.flow.component.page.Page.executeJs(Page.java:439)
	at com.vaadin.flow.component.page.Page.executeJavaScript(Page.java:394)
	at com.vaadin.flow.component.datepicker.DatePicker.lambda$initConnector$2f364bb9$1(DatePicker.java:320)
	at com.vaadin.flow.component.datepicker.DatePicker.lambda$null$fb4fab81$1(DatePicker.java:364)
	at com.vaadin.flow.internal.StateTree.lambda$runExecutionsBeforeClientResponse$1(StateTree.java:368)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at com.vaadin.flow.internal.StateTree.runExecutionsBeforeClientResponse(StateTree.java:365)
	at com.vaadin.flow.server.communication.UidlWriter.encodeChanges(UidlWriter.java:411)
	at com.vaadin.flow.server.communication.UidlWriter.createUidl(UidlWriter.java:187)
	at com.vaadin.flow.server.communication.UidlRequestHandler.writeUidl(UidlRequestHandler.java:121)
	at com.vaadin.flow.server.communication.UidlRequestHandler.synchronizedHandleRequest(UidlRequestHandler.java:90)
	at com.vaadin.flow.server.SynchronizedRequestHandler.handleRequest(SynchronizedRequestHandler.java:40)
	at com.vaadin.flow.server.VaadinService.handleRequest(VaadinService.java:1545)
	at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:247)
	at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:111)
[...]

PR #9415, opened 2020-11-17
Log warning instead of throwing NPE

TatuLund:
NPE thrown in addJavaScriptInvocation(..) if the session is null is not intuitive. I am proposing to log warning instead.

This is related to https://github.com/vaadin/flow/issues/9397

This PR is not fixing root cause, just making the behavior better.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -549,8 +549,13 @@ public class UIInternals implements Serializable {
      */
     public void addJavaScriptInvocation(
             PendingJavaScriptInvocation invocation) {
-        session.checkHasLock();
-        pendingJsInvocations.add(invocation);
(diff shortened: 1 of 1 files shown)


Legioth:
Seems weird to log "this should never happen" when there is indeed at least one known case when it will happen.

Legioth:
I'd strongly recommend fixing the root cause instead. Why is e.g. `runExecutionsBeforeClientResponse` even run in a situation when the session is `null`?

TatuLund:
> Why is e.g. runExecutionsBeforeClientResponse even run in a situation when the session is null?

This is a very valid question.

pleku:
I would not make this change for couple of reasons.
1) it does not help really, the actual issue should be fixed
2) this is done based on only one report: a slightly obscure bug report stack trace. 
3) this would not even "help" the user unless we backport this to certain versions the user can use. This is PITA since it is manual work

So simply not worth it to introduce this type of changes just based on one report.

PR #9418, opened 2020-11-17
feat: Theme component with app theme.

### LINKED ISSUE #9348: Add component theme feature for application theme.
Add importing component css files from the theme components folder into the generated `my-theme.js` file by adding component theme generation to the `theme-generator.js`.
See [componentsFiles](https://github.com/vaadin/flow/blob/068bca29dbb27e853d13b984325f71635c6e4c00/flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js#L109)

Extends test in test-npm-only-features/test-default-theme to check that component theme is set

caalador:
Added the feature that you can theme components
with the app theme by creating a css file in
'frontend/theme/myTheme/components' with the
component tag name e.g. 'my-field'  should have 'my-field.css'

Fixes #9348


flow-tests/test-themes/frontend/my-field.js
```diff
@@ -0,0 +1,210 @@
...
+      [part="input-field"] ::slotted(input)::-ms-clear {
+        display: none;
+      }
+    </style>
+  </template>
+</dom-module>`;
+
+document.head.appendChild($_documentContainer.content);
+
+import { PolymerElement } from '@polymer/polymer/polymer-element.js';
```
knoobie, line 181: Would it be good to add a test for lit template as well because it replaces polymer in the future?

caalador: It would, but I don't know how the theme handles in LIT.
@jouni would you have a simple component sample for Lumo on LIT?

jouni: Not sure what you mean by that. In any case, we should avoid using the `<dom-module>` element, and prefer using the official `registerStyles` method of injecting styles to component shadow roots: https://vaadin.com/docs/v14/themes/importing-style-sheets.html#component-styles

If the question is about ThemableMixin and LitElement, or how to the global Lumo style sheets, @web-padawan is the best to answer those questions.

flow-tests/test-themes/frontend/my-field.js
```diff
@@ -0,0 +1,210 @@
...
+import '@vaadin/vaadin-lumo-styles/sizing.js';
+import '@vaadin/vaadin-lumo-styles/spacing.js';
+import '@vaadin/vaadin-lumo-styles/style.js';
+import '@vaadin/vaadin-lumo-styles/typography.js';
+import '@vaadin/vaadin-lumo-styles/mixins/required-field.js';
+import '@vaadin/vaadin-lumo-styles/font-icons.js';
+import '@vaadin/vaadin-lumo-styles/mixins/field-button.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+
+const $_documentContainer = html`<dom-module id="lumo-text-field" theme-for="my-field">
```
jouni, line 27: Not sure if this is intentional, but it will conflict with the official Vaadin Text Field component style sheet, if both are imported to the same app. Prefer using unique IDs instead.

```suggestion
const $_documentContainer = html`<dom-module id="my-text-field" theme-for="my-field">
```

jouni: Note, that if we change to using `registerStyles` (see other thread), this is no longer relevant (no ID necessary).

caalador: This component is just here for testing purposes as we don't want Flow to depend on the components so its basically a ripped copy of the textField.

flow-tests/test-themes/frontend/my-field.js
caalador: The main question is how will the lit components and their theming work as compared to the polymer version.

jouni: Right. They will support `registerStyles`, but not `<dom-module>` (AFAIK).

flow-tests/test-themes/frontend/my-field.js
jouni: I would still change it to use `registerStyles`, to future-proof it if this test is ever converted to using LitElement instead of PolymerElement.

caalador: Do we already have a component implementation using registerStyles internally?
As this should test that the feature works with the components.

knoobie: @caalador https://github.com/vaadin/vaadin-app-layout uses registerStyles for example in the branch `next`

caalador: The big question is just that should we use here in the test something that we do not use in the actual version where the feature is coming out?

knoobie: I'm the wrong person to decide that ;) Just wanted to make sure, that theming is tested with both worlds because lit element is planned for V14.5 and V18.

jouni: registerStyles is an officially supported part of the API already in 14.x series. All components support it.

jouni: The implementation of our components for 14.x is still Polymer 2 based (because of the need to support Bower-mode), and the automatic conversion to Polymer 3 generates the uglier version of the style modules, which is used here as well (`$_documentContainer = ...`).

caalador:
Using actual component implementations from NPM for testing theming.
vaadin-text-field for polymer and vaadin-radio-button for LIT

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -77,6 +107,10 @@ function generateThemeFile(themeFolder, themeName) {
     ${globalCssCode.join('')}
     target['${globalCssFlag}'] = true;
   }
+  if (!document['${componentCssFlag}']) {
+    ${componentCssCode.join('')}
+    document['${componentCssFlag}'] = true;
+  }
 }
```
pleku, line 114: While not entirely connected to this PR, we should probably document this thing a bit, as it is quite unclear what does this applyTheme do and when it is used. At least when just reading the code again after I while, it makes me go _wat, what is this for and why_ ...

pleku (dismissed):

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
caalador: added some comment as there was need to merge and do conflict resolution.

pleku (approved):

pleku (approved):

PR #9427, opened 2020-11-17
feat: TypeScript API for connection indicator configuration

### LINKED ISSUE #9402: Fusion: TypeScript API for configuring the offline visual indicator
As a developer, I want to have some APIs for configuring the offline visual indicator, so that I can do customizations in a similar way as described in #9401.

Some example code, the API doesn't need to be the same

  ```ts
 import { setConnectionIndicatorConfiguration } from '@vaadin/flow-frontend/AppShell`;

    setConnectionIndicatorConfiguration({
        firstDelay: 300,
        secondDelay: 1500,
        thirdDelay: 5000
    });
```
    
NOTE: `@vaadin/flow-frontend` should eventually be changed to something like `@vaadin/vaadin` (out of scope of this task)

### PR DESCRIPTION
**Pull request by joheriks**
Use `ConnectionIndicator.ts: getConnectionIndicator` to access the component
and configure it using properties in `index.ts`. Also fixes broken connection state
transition when the app is entered via a client-side view.

Fixes #9402.

### CODE REVIEW

File: flow-client/src/main/resources/META-INF/resources/frontend/AppShell.ts [hunk 1]
```diff
@@ -0,0 +1,103 @@
...
+  if (delay !== undefined && delay! < 0) {
+    throw new ConfigurationError('delays must be positive');
+  }
+}
+
+/**
+ * Concrete values for each ConnectionIndicatorConfiguration to enable reflection over the field names.
+ */
+
+/* tslint:disable: max-classes-per-file */
```
haijian-vaadin, line 76: Maybe it's better to put this at the top of the file

flow-client/src/main/resources/META-INF/resources/frontend/index.ts
```diff
@@ -2,3 +2,4 @@ export * from './Flow';
 export * from './Connect';
 export * from './Authentication';
 export * from './Offline';
+export * from './ConnectionIndicator';
```
haijian-vaadin, line 5: Could also add Appshell.ts here?

haijian-vaadin (changes requested):

flow-client/src/main/resources/META-INF/resources/frontend/AppShell.ts
joheriks: Done.

flow-client/src/main/resources/META-INF/resources/frontend/index.ts
joheriks: Done.

flow-client/src/main/resources/META-INF/resources/frontend/AppShell.ts
```diff
@@ -0,0 +1,82 @@
...
+function validateDelay(delay?: number) {
+  if (delay !== undefined && delay! < 0) {
+    throw new ConfigurationError('delays must be positive');
+  }
+}
+
+/**
+ * Concrete values for each ConnectionIndicatorConfiguration to enable reflection over the field names.
+ */
+class ConnectionIndicatorConfigurationImpl implements ConnectionIndicatorConfiguration {
```
haijian-vaadin, line 56: Maybe DefaultConnectionIndicatorConfiguration would be better than ConnectionIndicatorConfigurationImpl?

flow-client/src/main/resources/META-INF/resources/frontend/AppShell.ts
```diff
@@ -0,0 +1,82 @@
+/* tslint:disable: max-classes-per-file */
+
+import {addConnectionIndicator} from "./ConnectionIndicator";
+
+export interface ConnectionIndicatorConfiguration {
+  onlineText?: string;
```
haijian-vaadin, line 6: Could we have some inline doc for the class and the properties?

haijian-vaadin (changes requested):

flow-client/src/main/resources/META-INF/resources/frontend/AppShell.ts
joheriks: Renamed to `ConcreteConnectionIndicatorConfiguration` instead (to avoid potentialconfusion with default configuration values, which are defined in the `ConnectionIndicator` component). There is now duplication of properties here---maybe we should consider whether we really need this intermediate configuration DTO, or if we instead could consider the `ConnectionIndicator` the public configuration interface instead.

flow-client/src/main/resources/META-INF/resources/frontend/AppShell.ts
joheriks: Done. However, there is now duplication of properties and documentation here. Maybe we should consider whether we really need this intermediate configuration DTO, or if we instead could consider the `ConnectionIndicator` the public configuration interface instead.

flow-client/src/main/resources/META-INF/resources/frontend/ConnectionIndicator.ts
```diff
@@ -124,9 +124,9 @@ export class ConnectionIndicator extends LitElement {
        style="${this.getLoadingBarStyle()}"></div>
 
       <div class="v-status-message ${classMap({
-        active: this.reconnecting,
-        modal: this.reconnectModal
-      })}">
+      active: this.reconnecting,
```
haijian-vaadin, line 127: indentations missing?

haijian-vaadin (dismissed):
Cool, really works nicely now

flow-client/src/main/resources/META-INF/resources/frontend/ConnectionIndicator.ts
joheriks: Done (apparently this was too difficult for my IDE to auto-indent properly).

haijian-vaadin (approved):

PR #9436, opened 2020-11-18
test: add test for setting global font

### LINKED ISSUE #9409: Add test for setting a font for global.css
Add a test for using a custom font in global.css into `flow-tests/test-themes`

Font should be in the theme folder and test should check that it has been correctly applied to the page.

caalador:
Fixes #9409


flow-tests/test-themes/frontend/theme/app-theme/global.css
```diff
@@ -1,3 +1,9 @@
+@font-face {
+    font-family: "Ostrich";
+    src: url("theme/app-theme/font/ostrich-sans-regular.ttf") format("TrueType");
+}
+
 body {
     background-image: url('theme/app-theme/img/bg.jpg');
+    font-family: "Ostrich";
```
pleku, line 8: > You should always include at least one generic family name in a font-family list, since there's no guarantee that any given font is available. This lets the browser select an acceptable fallback font when necessary.

Maybe add  `, monospace;` so that in case the font name loading doesn't work, it would use that instead and fail the test

pleku (changes requested):

pleku (approved):

flow-tests/test-themes/frontend/theme/app-theme/global.css
pleku: Nm, it doesn't work like that

PR #9439, opened 2020-11-18
Revert osgi

denis-anisimov:


pleku:
I think you misunderstood, and this doesn't even help for the history. It is now what it is.

PR #9442, opened 2020-11-19
fix: made VaadinRequest available in endpoint

haijian-vaadin:
Fixes https://github.com/vaadin/flow-and-components-documentation/issues/1391

The original fix in https://github.com/vaadin/flow/pull/7522 was incomplete since it requires a VaadinService being available.


flow-server/src/main/java/com/vaadin/flow/server/connect/VaadinConnectController.java
```diff
@@ -277,9 +279,7 @@ void validateEndpointBean(EndpointNameChecker endpointNameChecker,
             // Put a VaadinRequest in the instances object so as the request is
             // available in the end-point method
             VaadinServletService service = (VaadinServletService)VaadinService.getCurrent();
-            if (service != null) {
-                service.setCurrentInstances(new VaadinServletRequest(request, service), null);
-            }
+            CurrentInstance.set(VaadinRequest.class, new VaadinServletRequest(request, service));
```
joheriks, line 282: Could perhaps clear the instance in a `finally` block to avoid memory leaks?

joheriks (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/connect/VaadinConnectController.java
haijian-vaadin: That's very good point, thanks. Done

joheriks (approved):

Issue #9430, opened 2020-11-18
webpack should not touch external URLs in CSS


pleku:
We need to make sure & add tests that the users can have static resources like images from 
- the domain root (e.g. serving static files from apache/nginx with img.domain.com or /img/..)
- the context path (e.g. serving static files via spring from ./img/..)

and the URL in the CSS file is not modified.

PR #9451, opened 2020-11-20
feat: Serve theme static files from VAADIN/static

### LINKED ISSUE #9286: Investigate css and file-loader
`file-loader` loads `import`/`require()`  seemingly from .js files only.

This should be investigated that can the files be in css files as `@import` and how.

The `file-loader` could work fine with the confiuration:
```
      {
        // File-loader only copies files used as imports in .js files
        test: /\.(png|gif|jpg|svg|eot|woff|woff2|ttf)$/,
        use: [ {
          loader: 'file-loader',
          options: {
            outputPath: '../resources/VAADIN/static/',
            name: '[name].[ext]'
          }
        }],
      },
```

Where it copies all imported files in the `[themeName].js`  (e.g. `import './images/background.jpg';` )to `META-INF/resources/VAADIN/static` and can be used in css for instance as `background: url("VAADIN/static/background.jpg");` 

One question is can we have webpack css-loader influence the file-loader or do we need to parse the sources for static files to put into the `[themeName].js` as import or require()

Note! This actually only works in production mode as the dev server doesn't copy the static files with file-loader.

### LINKED ISSUE #9405: Jetty plugin not serving resources
The Jetty plugin when executing `mvn jetty:run` by default only serves `src/main/webapp` (also it seems `${project.output.directory}/webapp-tmp`) as static resources.

The app theme files are copied only `${project.build.outputDirectory}/META-INF/resources` which is served as per the Servlet 3.0 specification. 

A fix for running jetty should be found other than requiring adding the plugin configuration 
```
<webApp>
  <resourceBases> 
    <resourceBase>${project.build.outputDirectory}/META-INF/resources</resourceBase>
  </resourceBases>
</webApp>
```

caalador:
Static files in META-INF/VAADIN/static will now be served
on request to VAADIN/static.
Added new webpack loader that changes app theme
css urls targeting ./ to be VAADIN/static/ instead.

Fixes #9405
Closes #9286 
Touches #9430 which will still require the test.


flow-tests/test-themes/frontend/theme/app-theme/global.css
```diff
@@ -1,3 +1,3 @@
 body {
-    background-image: url('theme/app-theme/img/bg.jpg');
+    background-image: url("./img/bg.jpg");
```
knoobie, line 2: Isn't that exactly the opposite of https://github.com/vaadin/flow/issues/9430?  @pleku

caalador: hmm... but now it's a direct reference from the css file so IDEA can for instance auto complete and validate the reference file.

knoobie: That's a convenient for developer for "small" projects but can be a huge burden for dev or ops teams in bigger projects when you have requirements to load images from something outside of app. 

At least this "magic" should be configurable(opt-out) if that's something you need to have for vaadin's DX.

caalador: Updated so that we only change the url if the file actually exists relative to the css being handled.

knoobie: Interesting idea! 

Just a test case to verify: Project setup: Vaadin + Spring Boot 

`resources/public/img/first.png`
`resources/META-INF/VAADIN/static/img/second.jpg`
`resources/VAADIN/themes/my-theme/global.css`

global.css:
```css
body {
 background-image: url("./img/my-image.png");
}
main {
 background-image: url("./img/second.jpg");
}
```

Spring boot would serve the image `my-image.png` because it is inside the `public` folder. Now your change would not change the background-image of the css because `my-image.png` isn't reachable for webpack. And the background-image of main would be rewritten and of course served successfully. Correct?

caalador: The theme-loader will now actually check that the file exists relative to the css file handled so: if it can find the file resources/VAADIN/themes/my-theme/img/second.jpg it will be rewritten with VAADIN/static else it's seen as "external" and supplied by someone else.

knoobie: That was the missing piece, thanks!

flow-server/src/main/resources/plugins/application-theme-plugin/theme-copy.js
```diff
@@ -16,30 +16,23 @@
...
- * [staticResourcesFolder]/theme/[theme-name]
+ * [staticResourcesFolder]
  */
 
 const fs = require('fs');
 const path = require('path');
 
 /**
- * create theme/themeName folders and copy theme files there.
+ * copy theme files  to static assets folder.
```
pleku, line 26: maybe clarify what theme files will be copied

flow-server/src/main/resources/plugins/theme-loader/theme-loader.js
```diff
@@ -0,0 +1,26 @@
...
+
+const urlMatcher = /(url\()(\'|\")?(\.\/)(.*\2\))/g;
+
+module.exports = function (source, map) {
+  const handledResourceFolder = path.dirname(this._module.resource);
+  const logger = this.getLogger("theme-loader");
+
+  // Collect groups [url(] [ |'|"]optional './' and end of url
+  source = source.replace(urlMatcher, function (match, url, quoteMark, replace, endString) {
+    // for some reason this matcher can not be reused as a constant
```
pleku, line 12: I think to be reused you would have to make a `new Regexp(...)`

flow-tests/test-themes/frontend/theme/app-theme/global.css
```diff
@@ -1,9 +1,9 @@
 @font-face {
     font-family: "Ostrich";
-    src: url("theme/app-theme/font/ostrich-sans-regular.ttf") format("TrueType");
+    src: url("./font/ostrich-sans-regular.ttf") format("TrueType");
 }
 
 body {
-    background-image: url('theme/app-theme/img/bg.jpg');
+    background-image: url("./img/bg.jpg");
```
pleku, line 7: Lets add another test which verifies that `url("../icons/icon.png"); or similar works

pleku (changes requested):

flow-tests/test-themes/frontend/theme/app-theme/global.css
caalador: Here we have a small problem in that we won't accept static serving for anything that contains `../`
What we could do would be to accept navigation inside the theme folder and rewrite those to be absolute to the file target so we wouldn't have ../ e.g. `../icons/icon.png`  would fail from `/app-theme/global.css`, but would be rewritten to `icons/icon.png` in `app-theme/sub/extras.css`

pleku: Another option could be to use a magic URL thingy there like `theme://` but not sure if it should resolve to 
1) to the `theme/` folder
2) to the currently active theme
Usually it should resolve to 2) but in some cases when you're creating a reusable theme it should refer to the parent theme and not the currently active theme ... so probably this is just unnecessary complication and thus a bad idea.

I think using e.g. `../icons/icon.png` is common so if we can just automatically handle it, lets go with that.
For Java code, we'll expect people to anyway use the full path `"frontend/theme/my-theme/icons/icon.png"`.

One thing to note, even though it is not directly part of this issue, that the npm-based imports it is supposed to work with:
`"foobar/images/image.png"` when the package has been declared to be imported in `theme.json`. These URLs should not be touched even though they are not "external".

flow-server/src/main/resources/plugins/theme-loader/theme-loader.js
caalador: In the end the main matcher was updated so that we get the url part directly from there.

flow-tests/test-themes/frontend/theme/app-theme/global.css
caalador: in the format `"foobar/images/image.png"` they won't be touched. Only `./`is handled.

flow-tests/test-themes/frontend/theme/app-theme/sub-css/sub.css
```diff
@@ -0,0 +1,6 @@
+#sub-component {
+    background: url('./../icons/archive.png') left top;
```
pleku, line 2: As looking at `url(...)` examples from MDN, it has
```css
list-style-image: url('../images/bullet.jpg');
```
so I think we should be able to have the same, `../` is less things to type than `./../`.

flow-server/src/main/resources/plugins/theme-loader/theme-loader.js
```diff
@@ -1,23 +1,34 @@
+const loaderUtils = require("loader-utils");
 const fs = require('fs');
 const path = require('path');
 
-const urlMatcher = /(url\()(\'|\")?(\.\/)(.*\2\))/g;
+// Collect groups [url(] [ |'|"]optional './', file part and end of url
+const urlMatcher = /(url\()(\'|\")?(\.\/)(\S*)(\2\))/g;
```
pleku, line 6: So this IMO should handle `../` too and not just `./`

pleku (changes requested):

flow-server/src/main/resources/plugins/theme-loader/theme-loader.js
caalador: Added now with more \ and /

flow-tests/test-themes/frontend/theme/app-theme/sub-css/sub.css
caalador: done

pleku (approved):

PR #9456, opened 2020-11-20
fix: mark tests that should not be executed in OSGi

denis-anisimov:


pleku (approved):

pleku:
@denis-anisimov I don't know if there should be squashed or not on merge, probably if individual commits are not going to be picked anywhere ?

denis-anisimov:
> @denis-anisimov I don't know if there should be squashed or not on merge, probably if individual commits are not going to be picked anywhere ?

Squash should be OK.
don't see reasons to have separate commits in this PR.

PR #9472, opened 2020-11-23
refactor: Remove OSGi code from flow-server

### LINKED ISSUE #9219: OSGi: get rid of fake OSGiServletContext (and may be OSGiAccess).
`OSGiServletContext`  has been introduced because `ServletContextInitializer`s are not executed in OSGi container and `OSGiAccess` is used to collect `ServletContextInitializer` classes and run them.
To be able to run then a servlet context is required.
But since we are executing the initializers by ourselves via OSGi mechanisms there is no servlet context available at all.
That's why a fake context has been created: to pass it to the initializers (at the point when no any real context is available).

The suggestion is : use Servlet `ServletContextListener` (a real one) which will execute all the `ServletContextInitializer`s.
Two notes:
- it seems there is a way to register a `ServletContextListener`  in OSGi using HTTP whiteboard spec.
- the logic should be similar to the current one : collect all `ServletContextInitializer`s  , store then somewhere and then use their classes ( instantiate objects and run them) inside the `ServletContextListener` with a _REAL_ `ServletContext`.
- move all the logic to the `flow-osgi` module whenever it's possible : keep OSGi related code in `flow-server` at the minimum level.

denis-anisimov:
Fixes #9219


flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
```diff
@@ -364,6 +364,8 @@ private static String getResourceFromFile(Properties initParameters)
      */
     private static String getTokenFileFromClassloader(Class<?> contextClass,
             VaadinContext context) throws IOException {
+        System.out.println("SSSSSSSSSSSSSSSSSSSSSs "
+                + ((VaadinServletContext) context).getContext());
```
caalador, line 368: Debug system out should be removed.

flow-server/pom.xml
```diff
@@ -164,20 +164,6 @@
...
-            <scope>provided</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.osgi</groupId>
-            <artifactId>osgi.cmpn</artifactId>
-            <version>${osgi.compendium.version}</version>
-            <scope>provided</scope>
-        </dependency>
-
```
caalador, line 180: Compilation fails after removal
`package org.osgi.framework does not exist`
`package org.osgi.service.component.annotations does not exist`
`package org.osgi.util.tracker does not exist`
Multiple `cannot find symbol` all in  `com/vaadin/flow/server/osgi/ServletContainerInitializerExtender`

caalador (changes requested):

caalador (approved):

Issue #8848, opened 2019-04-04
Property update request for disabled element warning when calling setEnabled(false)


mlopezFC:
Tested with Vaadin 13.0.3

With this relative simple example:

```
@Route("")
public class MainView extends VerticalLayout {
    private TextField disabledTextField;
	public MainView() {
		disabledTextField = new TextField();
		disabledTextField.setEnabled(false);
		add(disabledTextField);
    }
}
```
Just by entering the application we get this warning in the log:

```
[qtp1326230282-42] WARN com.vaadin.flow.server.communication.rpc.MapSyncRpcHandler - Property update request for disabled element is received from the client side. The property is 'invalid'. Request is ignored.
```

Is that right?

pleku:
See comment https://github.com/vaadin/flow/pull/9473#issuecomment-739958715

Basically there are some updates happening for most of the Vaadin "user input" components that should not occur, and when the component has been set as disabled / readonly, this will give the warning. Working on this.

Frettman:
I still get this warning e.g. simply by adding a disabled DatePicker to my layout. (Vaadin Flow 20.0.6).

PR #9473, opened 2020-11-23
Dropping logging level from warn to debug

TatuLund:
We have encountered real-life scenario, where these warnings pollutes logs in excessive manner. As it is normal conditions harmless note, it does not warrant warn level, I am proposing to change to debug level.

"Currently in our server production log, 80% is filled with this and we will have hard time monitoring our application."

Fixes: https://github.com/vaadin/flow/issues/8848

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/ReturnChannelHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/ReturnChannelHandler.java
@@ -67,7 +67,7 @@ public class ReturnChannelHandler extends AbstractRpcInvocationHandler {
 
         if (!node.isEnabled() && channel
                 .getDisabledUpdateMode() != DisabledUpdateMode.ALWAYS) {
-            getLogger().debug("Ignoring update for disabled return channel: {}",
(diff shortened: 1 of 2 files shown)


Legioth:
If dropping the logging level, then maybe only one step to `info` instead? Those warnings can be highly relevant especially for someone debugging cases when their own component isn't updating the server as expected.

pleku (changes requested):
Agreed, `info` is better than debug.
If someone is looking to cut down the amount of logs, info level is already too noisy for that.
And as this is still kind of a sign of some "extra communication" between client->server that is not necessary, one should be able to take note about it without setting the level to debug.

pleku (approved):

pleku:
So after some investigation it would seem that the best thing to do would be to address any warnings that result from disabled Vaadin components individually - thus the warning level should stay there as it is not as intended. There is at least one case for TextField, NumberField, DataPicker, TimePicker BigDecimalField ... where it is just not working right and producing the warning because it is not working properly - and this should be fixed instead. Working on that at the moment and closing this PR.

PR #9475, opened 2020-11-24
fix: remove instance field from singleton IndexHtmlRequestHandler

### LINKED ISSUE #9287: IndexHtmlRequestHandler: com.vaadin.flow.server.VaadinService#modifyIndexHtmlResponse might be called for IndexHtmlResponse of other user
Method `IndexHtmlRequestHandler.synchronizedHandleRequest` stores created `indexHtmlResponse` into an instance field of this singleton-scoped class (see lines 81 and 87). It is retrieved later to call `VaadinService.modifyIndexHtmlResponse()` (line 117). However, since this is an instance field and no synchronization is performed, at the time of calling, the content of `indexHtmlResponse` might belong to another user. 

Depending on the implementation of particular `IndexHtmlRequestListener`, this may lead to sensitive data exposure or corruption (`IndexHtmlResponse` contains references of `VaadinRequest` and `VaadinResponse`, together with `index.html` document that can be modified).

https://github.com/vaadin/flow/blob/f617010cdff589bf8b8d714081d1a2d1488ca156/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java#L65-L130

haijian-vaadin:
Fixes #9287

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
@@ -63,7 +63,7 @@ public class IndexHtmlRequestHandler extends JavaScriptBootstrapHandler {
     public boolean synchronizedHandleRequest(VaadinSession session,
             VaadinRequest request, VaadinResponse response) throws IOException {
         DeploymentConfiguration config = session.getConfiguration();
-        IndexHtmlResponse indexHtmlResponse = null;
+        IndexHtmlResponse indexHtmlResponse;
 
         Document indexDocument = config.isProductionMode()
                 ? getCachedIndexHtmlDocument(request.getService())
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -59,12 +59,11 @@
  */
 public class IndexHtmlRequestHandler extends JavaScriptBootstrapHandler {
 
-    private transient IndexHtmlResponse indexHtmlResponse;
-
     @Override
     public boolean synchronizedHandleRequest(VaadinSession session,
             VaadinRequest request, VaadinResponse response) throws IOException {
         DeploymentConfiguration config = session.getConfiguration();
+        IndexHtmlResponse indexHtmlResponse = null;
```
fluorumlabs, line 66: Does it has to be initialized with `null`?

haijian-vaadin: No, but not initialized, i.e. `IndexHtmlResponse indexHtmlResponse;` is basically the same as `IndexHtmlResponse indexHtmlResponse = null;`? Or do you mean sth else?

fluorumlabs: Not really, there will be an `LDC` opcode to load variable with `null` value. Semantically, if `null` is not supposed to be there, it should be omitted, to reduce confusion during future refactoring

haijian-vaadin: ok, done

joheriks (approved):

PR #9482, opened 2020-11-24
test: add test that external urls are not touched

### LINKED ISSUE #9430: webpack should not touch external URLs in CSS

caalador:
Added test that we don't touch external
urls (relative or abosolute).
Moved flow to be under path so we can have
jetty serve "external" resources.

Fixes #9430


flow-tests/test-themes/frontend/theme/app-theme/global.css
```diff
@@ -9,3 +9,18 @@ body {
     background-image: url("./img/bg.jpg");
     font-family: "Ostrich";
 }
+
+#butterfly {
+    /* This shouldn't be translated */
+    background-image: url("./test/path/monarch-butterfly.jpg");
```
knoobie, line 15: +1 

Just one question: Would it be good to add a test for an external absolute URI (https..) as well? Or should flow not depend on external sources in the tests? (understandable)

caalador: I did think about it, but decided not to add external dependency that is dependent on another server somewhere.

pleku (approved):

PR #9484, opened 2020-11-24
fix: optimize handling of requests containing Range header

joheriks:
More efficient parsing of the Range: header value. Also, range count is capped at 16
(additional will be ignored) and overlapping ranges at 2 (request will be denied
if above).

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/ResponseWriter.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/ResponseWriter.java
@@ -29,8 +29,8 @@ import java.io.Serializable;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
-import java.util.ArrayList;
 import java.util.List;
+import java.util.Stack;
 import java.util.UUID;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -52,9 +52,9 @@ public class ResponseWriter implements Serializable {
     private static final int DEFAULT_BUFFER_SIZE = 32 * 1024;
 
     private static final Pattern RANGE_HEADER_PATTERN = Pattern.compile(
-            "^bytes=(([0-9]*-[0-9]*\\s*,\\s*)*[0-9]*-[0-9]*\\s*)$");
+            "^bytes=((\\d*-\\d*\\s*,\\s*)*\\d*-\\d*\\s*)$");
     private static final Pattern BYTE_RANGE_PATTERN = Pattern.compile(
-            "([0-9]*)-([0-9]*)");
+            "(\\d*)-(\\d*)");
 
     /**
      * Maximum number of ranges accepted in a single Range header. Remaining ranges will be ignored.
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/ResponseWriter.java
```diff
@@ -51,8 +51,20 @@
 public class ResponseWriter implements Serializable {
     private static final int DEFAULT_BUFFER_SIZE = 32 * 1024;
 
-    private static final Pattern RANGE_HEADER_PATTERN = Pattern.compile("^bytes=(([0-9]*-[0-9]*,?\\s*)+)$");
-    private static final Pattern BYTE_RANGE_PATTERN = Pattern.compile("([0-9]*)-([0-9]*)");
+    private static final Pattern RANGE_HEADER_PATTERN = Pattern.compile(
+            "^bytes=(([0-9]*-[0-9]*\\s*,\\s*)*[0-9]*-[0-9]*\\s*)$");
```
denis-anisimov, line 55: `[0-9]` is `\d`.
The expression will become shorter.

flow-server/src/main/java/com/vaadin/flow/internal/ResponseWriter.java
```diff
@@ -208,8 +221,14 @@ private void writeRangeContents(String range, HttpServletResponse response,
             long start = startGroup.isEmpty() ? 0L : Long.parseLong(startGroup);
             long end = endGroup.isEmpty() ? Long.MAX_VALUE
                     : Long.parseLong(endGroup);
+            for (Pair<Long, Long> accepted : ranges) {
```
denis-anisimov, line 224: unnecessary `for` inside `while` : please refactor

flow-server/src/main/java/com/vaadin/flow/internal/ResponseWriter.java
```diff
@@ -196,8 +208,9 @@ private void writeRangeContents(String range, HttpServletResponse response,
         long resourceLength = connection.getContentLengthLong();
         Matcher rangeMatcher = BYTE_RANGE_PATTERN.matcher(byteRanges);
 
+        int overlappingRangeCount = 0;
         List<Pair<Long, Long>> ranges = new ArrayList<>();
-        while (rangeMatcher.find()) {
+        while (rangeMatcher.find() && ranges.size() < MAX_RANGE_COUNT) {
```
denis-anisimov, line 213: What happens if `ranges.size() >= MAX_RANGE_COUNT`?
There is no mistake set. The processing goes in happy path.
This is not the same as check for `overlappingRangeCount` which fails the request processing.

denis-anisimov: Ah, is it covered by the case that `end < start`?

I would may be refactor this a bit and log an error message depending on what happened (instead of one big `if`).
That will also help to avoid comments: log messages will be self documented.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/internal/ResponseWriter.java
joheriks: According to the [RFC](https://tools.ietf.org/html/rfc7233#section-6.1), it is acceptable for the server to either drop some ranges or outright reject a request when ranges are out-of-order or overlapping. My implementation was inconsistent in the handling of too many ranges compared to too many overlapping range. In the case of a benevolent client (that happened to asks too many ranges or too many overlapping ranges), it is probably better to serve the maximum amount rather than outright deny the request. I have amended the implementation to serve only the ranges processed so far when hitting more than 2 overlapping ranges. Also added logging.

flow-server/src/main/java/com/vaadin/flow/internal/ResponseWriter.java
joheriks: Done. Moved the limit checks to a separate method.

flow-server/src/main/java/com/vaadin/flow/internal/ResponseWriter.java
joheriks: Done.

denis-anisimov (approved):

mshabarov:
Cherry picked to `2.5`

Issue #9480, opened 2020-11-24
OSGi: AppShellRegistry::isShell is broken 


denis-anisimov:
Weird code is used to check whether a class is a shell class or not : https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/server/AppShellRegistry.java#L155

```
clz.getClassLoader()
                    .loadClass(AppShellConfigurator.class.getName())
                    .isAssignableFrom(clz);
```
`clz.getClassLoader()` is classloader of web applicaiton.
`AppShellConfigurator` is a `flow-server` classloader.
Why web app classloader is used to load `AppShellConfigurator` ?

This throws an exception in OSGi and break it.

I would like to understand why the code `return AppShellConfigurator.class.isAssignableFrom(clz);` can't be used here ?

denis-anisimov:
@manolo , looks like this code is introduced by you . Could you please give me a hint why do you use a classloader here ?

denis-anisimov:
I made a workaround: #9486
I have no any idea why `clz.getClassLoader()` is used to load `AppShellConfigurator` class instead of using `AppShellConfigurator.class`  directly but apparently if `AppShellConfigurator.class` is a superinterface for `clz` via the check `AppShellConfigurator.class.isAssignableFrom(clz)` the method should return `true`.
So my correction should not break anything at least.

Another question why `clz.getClassLoader()` is used at all.
If there is a reason to use  `clz.getClassLoader()`  I would like to see a comment which explain this in the code.

fluorumlabs:
The only explanation I managed to find is somewhere in this PR comments: https://github.com/vaadin/flow/pull/7050

PR #9486, opened 2020-11-25
fix: check whether class implements an interface first directly

denis-anisimov:
Part of #9480


caalador (approved):
I would perhaps drop the whole classloader part as I can't come up with a usecase where isAssignable would return differently from the loaded class isAssignable.

denis-anisimov:
> I would perhaps drop the whole classloader part as I can't come up with a usecase where isAssignable would return differently from the loaded class isAssignable.

I totally agree with you.

But that's exactly the point why I don't want to drop it: it's so "obvious" that it should not be there but it's there.
So may be there is a reason which both of us are not aware. 
This is Fusion issue and I would like them to resolve it: let's put this on their table. If it's not needed then they will just remove it.
Otherwise I requested in the original ticket the comment which explains why.

PR #9489, opened 2020-11-25
fix: Add polyfill for constructable stylesheets

### LINKED ISSUE #9488: AppTheme doesn't work on FireFox
App theme fails to load on firefox as new CSSStyleSheet causes a TypeError: Illegal constructor
This is because at the moment constructable stylesheets are only implemented in chrome.

The feature is on the road map for FireFox, but until it's there we need to add a polyfill for the feature.

caalador:
Add the constructable stylesheets polyfill so
app theme also works with non chrome based browsers.

Fixes #9488


tanbt (approved):
Works! I was able to reproduce the problem in `master`, which is fixed by this PR.

PR #9490, opened 2020-11-25
5.0 cherry pick

haijian-vaadin:
DON'T SQUASH


pleku (approved):
The removed protected method should be be addressed in the release notes in case there is a theoretical chance that a user has been calling it

PR #9496, opened 2020-11-25
fix: Correct regexp to only match when digit first

### LINKED ISSUE #9494: com.vaadin.flow.internal.CustomElementNameValidator#checkHtmlTagRules doesn't work as expected
https://github.com/vaadin/flow/blob/907cbbd59612ac2e31daf5c1566fb48529d75200/flow-server/src/main/java/com/vaadin/flow/internal/CustomElementNameValidator.java#L58-L62

However, `STARTS_WITH_A_DIGITS` is defines as

https://github.com/vaadin/flow/blob/907cbbd59612ac2e31daf5c1566fb48529d75200/flow-server/src/main/java/com/vaadin/flow/internal/CustomElementNameValidator.java#L32

As a result the check will fail if custom element name contains digit anywhere, not just in the beginning

caalador:
use find instead of matches as matches never matched.

Fixes #9494

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/CustomElementNameValidator.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/CustomElementNameValidator.java
@@ -29,7 +29,7 @@ import java.util.stream.Stream;
  * @since 1.0
  */
 public final class CustomElementNameValidator {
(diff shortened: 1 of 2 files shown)


mshabarov (approved):

taefi (approved):

mshabarov:
Cherry picked to `2.4`

Issue #9354, opened 2020-11-10
Move fusion code to a separate module


haijian-vaadin:
Fusion has some dependencies that are not OSGi compatible (e.g. javaparser, swagger, spring), which is blocking the Flow OSGI support.

The first step could be extract the `com.vaadin.flow.server.connect` package from `flow-server` to a separate module like `fusion-server`. It's OK that `fusion-server` depends on `flow-server`, but `flow-server` should definitely not depend on `fusion-server`.

The benefit would be that for OSGi application, the `fusion-server` can be explicitly excluded.

Another thing that we could do is to renamve the `com.vaadin.flow.server.connect` to `com.vaadin.fusion.server.connect`, and mark the `com.vaadin.flow.server.connect` as deprecated. Could also drop the connect suffix.

There are also some classes that are outside the `com.vaadin.flow.server.connect` package, but use the connect package, for example, `TaskGenerateConnect`, for this case, we can create an interface for `TaskGenerateConnect` in flow-server, but have the implementation in `fusion-server`, use [`Lookup`](https://github.com/vaadin/flow/blob/osgi-master/flow-server/src/main/java/com/vaadin/flow/di/Lookup.java) to find the implementations if there are. Lookup is introduced in [this OSGi PR](https://github.com/vaadin/flow/pull/9321) which allows to define a service interface (e.g. in `flow-server`  or in another module ) and then use get the implementation of this service via : `lookup.lookup(TaskGenerateConnect.class)`. 

Note that there shouldn't be breaking change, the expected result is that for Fusion/Flow developer, the project works with the current pom file.

DoD:
1. OSGi incompatible dependencies are removed from [the bnd file](https://github.com/vaadin/flow/blob/master/flow-server/bnd.bnd) 
2. To verify

- https://github.com/vaadin/osgi : https://github.com/vaadin/osgi/tree/master/test-containers/felix-jetty You can use this project to verify that OSGi is not broken.
- You need to build Flow locally including `test-root-context` module
- install flow-osgi with `mvn install -DskipTests`
- Then you may run `felix-jetty`.
   - No need to run all the tests because if OSGi broken you won't be able even to start it.
   - Build `felix-jetty` via `mvn install -DskiptTests` , Then go to `target/felix-framework-6.3.0` and run OSGi container via `java -jar bin/felix.jar`.
   - Then open localhost:8888/view. If it works then OSGi is not broken.
Note, currently, it requires java 8 to run this application

PR #9499, opened 2020-11-25
chore: move fusion endpoint code to fusion-endpoint module

haijian-vaadin:
Fixes: #9354
move fusion code to a separate fusion-endpoint module, so that the fusion dependencies which are not OSGi compatible can be removed from the flow-server module.

- move the code in the connect package  to fusion-server
- removed the non OSGi compatible packages from the bnd file in flow-server
- change  `TaskGenerateOpenApi` and `TaskGenerateConnect` to interfaces in flow-server, have the implementations in fusion-endpoint since they are using code inside the connect package
- Move endpoint related tests:
   - from NodeTasksTest to  NodeTasksEndpointTest, 
   - from DevModeInitializerTest to DevModeInitializerEndpointTest 
- Add a `LookupImpl` class in the `flow-maven-plugin` module for finding endpoint services.
- Introduce a `isInstantiableService`method in `ReflectTools`


fusion-server/src/test/resources/com/vaadin/flow/server/connect/generator/endpoints/collectionendpoint/expected-model-collectionendpoint.CollectionEndpoint.Collection.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * This module is generated from com.vaadin.flow.server.connect.generator.endpoints.collectionendpoint.CollectionEndpoint.Collection.
+ * All changes to this file are overridden. Please consider to make changes in the corresponding Java file if necessary.
+ */
+export default interface Collection {
+  author: string;
+  collectionName: string;
+  type: string;
+}
```
denis-anisimov, line 9: new line OR is it generated ?

denis-anisimov (changes requested):

fusion-server/src/test/resources/com/vaadin/flow/server/connect/generator/endpoints/collectionendpoint/expected-model-collectionendpoint.CollectionEndpoint.Collection.ts
haijian-vaadin: generated, the same as in flow-server, I think one of them could be removed in the end

flow-server/pom.xml
```diff
@@ -164,51 +164,6 @@
...
-            <artifactId>jackson-databind</artifactId>
-            <version>${jackson.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.datatype</groupId>
-            <artifactId>jackson-datatype-jsr310</artifactId>
-            <version>${jackson.version}</version>
-        </dependency>
         <!-- Needed for security annotations -->
         <dependency>
```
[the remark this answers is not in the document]
haijian-vaadin: updated, does it look good now?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateEndpointBase.java
```diff
@@ -0,0 +1,41 @@
...
+ * the License.
+ */
+package com.vaadin.flow.server.frontend;
+
+import java.io.File;
+
+/**
+ * Base interface for endpoint generation related tasks.
+ */
+public interface TaskGenerateEndpointBase extends FallibleCommand {
```
denis-anisimov, line 23: Not sure the final usage of this interface but it seems the methods with the return type should return the class and not its superclass.

So may be it should be 
```
TaskGenerateEndpointBase<T extends TaskGenerateEndpointBase> extends FallibleCommand{

T withApplicationProperties(File applicationProperties);
}
```

denis-anisimov, line 23: Another thing which I would recommend: methods `withXXX` are good for builders.
These are commands, not builders.

I would  make factories instead which produces the command. 
Factory should accept paramer(s) and produces the command instance.

The only issue which I see here: there should be only one method. It means the fixed amount of parameters.
So factories extending has no sense.
Instead I would may be make builder (which may extends one each other) and the builder may produce an immutable instance of configuration which will be passed to a factory.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateOpenApi.java
```diff
@@ -16,53 +16,25 @@
...
 
 /**
  * Generate OpenAPI json file for Vaadin Endpoints.
  */
-public class TaskGenerateOpenApi extends AbstractTaskConnectGenerator {
-
-    private final File javaSourceFolder;
-    private final transient ClassLoader classLoader;
-    private final File output;
+public interface TaskGenerateOpenApi extends TaskGenerateEndpointBase {
```
denis-anisimov, line 23: same comments as above for `TaskGenerateEndpointBase`.

flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java
```diff
@@ -281,8 +282,9 @@ public static void initDevModeHandler(Set<Class<?>> classes,
         File flowResourcesFolder = new File(baseDir,
                 DEFAULT_FLOW_RESOURCES_FOLDER);
 
+        Lookup lookup = new VaadinServletContext(context).getAttribute(Lookup.class);
         Builder builder = new NodeTasks.Builder(new DevModeClassFinder(classes),
-                new File(baseDir), new File(generatedDir),
+                new File(baseDir), lookup, new File(generatedDir),
```
denis-anisimov, line 287: This code and code below assumes that connect will be available and configured.
(see `withConnectApplicationProperties`, etc...).

The proper logic should check whether a service is available and the configure it only if it's available.

denis-anisimov (changes requested):

flow-server/pom.xml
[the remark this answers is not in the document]
denis-anisimov: Looks good now

flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/BuildFrontendMojoTest.java
```diff
@@ -408,14 +408,6 @@ public void noTokenFile_noTokenFileShouldBeCreated()
         Assert.assertFalse(tokenFile.exists());
     }
 
-    @Test
-    public void mavenGoal_generateOpenApiJson_when_itIsInClientSideMode()
```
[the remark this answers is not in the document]
haijian-vaadin: It's moved to `BuildFrontendMojoEndpointTest.java`, the `BuildFrontendMojoEndpointTest.java` is currently ignored though, need some work to make it pass.

flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojo.java
```diff
@@ -123,8 +126,10 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         try {
             File flowResourcesFolder = new File(npmFolder,
                     DEFAULT_FLOW_RESOURCES_FOLDER);
+            ClassFinder classFinder = getClassFinder(project);
+            Lookup lookup= new LookupImpl(classFinder);
```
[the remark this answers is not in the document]
haijian-vaadin: Done, but couldn't make a non-args, since it depends on a private filed
(`private MavenProject project; `) to get the class finder.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -133,8 +136,8 @@
          * @param npmFolder
          *            folder with the `package.json` file
          */
-        public Builder(ClassFinder classFinder, File npmFolder) {
-            this(classFinder, npmFolder, new File(npmFolder, System
+        public Builder(ClassFinder classFinder, File npmFolder, Lookup lookup) {
```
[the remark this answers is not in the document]
haijian-vaadin: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateEndpointBase.java
haijian-vaadin: I removed this unnecessary interface.

haijian-vaadin: Good point, I removed those `withXXX` methods, use a `init()` method instead, in this way, we also don't need the factory.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateOpenApi.java
haijian-vaadin: done

flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java
haijian-vaadin: Add logic to check if endpoint services are available before running endpoint related logic

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateEndpointBase.java
denis-anisimov: `init` method may be better than `withXxx` but it's also an evil.

`init` method is a weak API: `public` method means that the method may be called by _anyone_ at _any time_.
So your intention with `init` method is to define a contract which says: this method should be called once.
But you don't define such contract via Java : as I said the method is `public` and can be called many times by anyone who has an access to the instance.

May be in this case since the instance is "private" it's not an issue.
But I'm pretty sure that you may replace easily `init` with a factory `create`(or whatever), it's almost the same with one significant difference: `init` may be called any number of times , CTOR may be called _only once_.
So it's always better to create some new instance via CTOR with arguments instead of `init` method with same arguments.
The only difference is : to be able to call CTOR you will need a factory (which can be created with a default CTOR) and then create a required instance via a factory method which instantiate an instance via CTOR instead of calling `init` on the required instance right away.

So it's better to use factories instead of ugly `init` method: it defines good clear contract via Java without any additional assumption.

haijian-vaadin: done

flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
```diff
@@ -161,8 +159,11 @@ private void runNodeUpdater() throws ExecutionFailedException, MojoExecutionExce
         } catch (URISyntaxException e) {
             throw new MojoExecutionException("Failed to parse " + nodeDownloadRoot, e);
         }
+        
+        ClassFinder classFinder = getClassFinder(project);
+        Lookup lookup= createLookup(classFinder);
         // @formatter:off
-        new NodeTasks.Builder(getClassFinder(project),
+        new NodeTasks.Builder(lookup, classFinder,
```
denis-anisimov, line 166: I would may be avoid passing `ClassFinder`  as a separate parameter here and everywhere.

Lookup may be created so that it may return the `ClassFinder` by `ClassFinder.class`.
You may implement this by yourself here or wait for https://github.com/vaadin/flow/pull/9572 where I introduced two static methods into Lookup which allows to extend it.

But this is not very important.

flow-maven-plugin/src/main/java/com/vaadin/flow/utils/LookupImpl.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param classFinder
+     *            for searching service implementations.
+     */
+    public LookupImpl(ClassFinder classFinder) {
+        this.classFinder = classFinder;
+    }
+
+    @Override
+    public <T> T lookup(Class<T> serviceClass) {
+        return lookupAll(serviceClass).stream().findFirst().orElse(null);
```
denis-anisimov, line 46: Here you may add directly a special case for `ClassFinder.class`.

flow-server/pom.xml
```diff
@@ -78,6 +78,16 @@
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
```
denis-anisimov, line 82: Have you checked that these deps don't break OSGi ?
Do they have proper OSGi manifest ?

denis-anisimov: By the way : why they are needed ?

flow-server/src/main/java/com/vaadin/flow/internal/ReflectTools.java
```diff
@@ -825,6 +825,37 @@ public static Object getAnnotationMethodValue(Annotation annotation,
         return Optional.empty();
     }
 
+    /**
+     * Check if a class can be instantiated as a service or not.
```
denis-anisimov, line 829: Would be good to improve javadoc.
It should explain what the service is: this is a static check whether a class may be instantiated via its default CTOR via reflection. static check means that the CTOR may throw an exception and this is not checked but only possibility to instantiate the class based on type data

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -125,39 +126,47 @@
...
+         * @param lookup
+         *            a {@link Lookup} to discover services used by Flow (SPI)
          * @param classFinder
          *            a class finder
          * @param npmFolder
          *            folder with the `package.json` file
          */
-        public Builder(ClassFinder classFinder, File npmFolder) {
-            this(classFinder, npmFolder, new File(npmFolder, System
+        public Builder(Lookup lookup, ClassFinder classFinder, File npmFolder) {
```
denis-anisimov, line 141: I suggest to skip `ClassFinder`  passing as a parameter here (and everywhere).

flow-server/src/main/java/com/vaadin/flow/server/frontend/EndpointGeneratorTaskFactory.java
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @param applicationProperties
+     *            application properties file.
+     * @param openApi
+     *            openApi json file.
+     * @param outputFolder
+     *            the output folder.
+     * @param frontendDirectory
+     *            the frontend folder.
+     * @return an endpoint tasks for generating TypeScript files for endpoints.
```
denis-anisimov, line 40: `create` implies that the return value won't be `null`  but anyway please add to the `@return` that it's never `null`

flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java
```diff
@@ -394,13 +397,19 @@ public static void initDevModeHandler(Set<Class<?>> classes,
         DevModeHandler.start(config, builder.npmFolder, runNodeTasks);
     }
 
+    private static boolean isEndpointServiceAvailable(Lookup lookup) {
+        if (lookup == null) {
```
denis-anisimov, line 401: This is technically extra check.
The lookup is never `null` here.

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
```diff
@@ -299,29 +295,6 @@ public void shouldUseFullPathScannerByDefault() throws Exception {
                 Mockito.any(FallbackChunk.class));
     }
 
-    @Test
-    public void should_generateOpenApi_when_EndpointPresents()
```
denis-anisimov, line 303: Should these test be really removed ?
Are they moved somewhere else ?
Should there be some replacement ?

E.g. now there is no need to check the functionality (integration: something is generated and so on).
But you may check that some methods of classes that are in the lookup are called: e.g. factory is called.
And also objects produced by the factory are called somehow (doesn't matter what they do but they are called).

fusion-endpoint/src/main/java/com/vaadin/flow/server/frontend/fusion/EndpointGeneratorTaskFactoryImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import com.vaadin.flow.server.frontend.TaskGenerateConnect;
+import com.vaadin.flow.server.frontend.TaskGenerateOpenApi;
+
+/**
+ * An implementation of the EndpointGeneratorTaskFactory, which creates endpoint generator tasks.
+ */
+public class EndpointGeneratorTaskFactoryImpl implements EndpointGeneratorTaskFactory {
+
+    @Override
+    public TaskGenerateConnect createTaskGenerateConnect(File applicationProperties, File openApi,
```
denis-anisimov, line 30: Up to you , but I would may be add checks that arguments are not `null` (if they should not be `null` ).
And javadocs should be extended so it mentions which arguments are allowed to be `null` and which not.

fusion-endpoint/src/main/java/com/vaadin/flow/server/frontend/fusion/TaskGenerateConnectImpl.java
```diff
@@ -0,0 +1,61 @@
...
+
+import static com.vaadin.flow.server.connect.generator.VaadinConnectClientGenerator.CONNECT_CLIENT_NAME;
+import static com.vaadin.flow.server.connect.generator.VaadinConnectClientGenerator.CUSTOM_CONNECT_CLIENT_NAME;
+
+/**
+ * Generate the Vaadin TS files for endpoints, and the Client API file.
+ */
+public class TaskGenerateConnectImpl extends AbstractTaskConnectGenerator implements TaskGenerateConnect {
+
+    private File outputFolder;
```
denis-anisimov, line 34: I think all (or most) fields may be set `final`.

fusion-endpoint/src/main/java/com/vaadin/flow/server/frontend/fusion/TaskGenerateOpenApiImpl.java
```diff
@@ -0,0 +1,68 @@
...
+import com.vaadin.flow.server.ExecutionFailedException;
+import com.vaadin.flow.server.frontend.TaskGenerateOpenApi;
+import com.vaadin.flow.server.connect.generator.OpenApiSpecGenerator;
+
+/**
+ * Generate OpenAPI json file for Vaadin Endpoints.
+ */
+public class TaskGenerateOpenApiImpl extends AbstractTaskConnectGenerator implements TaskGenerateOpenApi {
+
+    private File javaSourceFolder;
```
denis-anisimov, line 31: All fields are most likely `final`.

fusion-endpoint/src/main/java/com/vaadin/flow/server/frontend/fusion/TaskGenerateOpenApiImpl.java
```diff
@@ -0,0 +1,68 @@
...
+    private transient ClassLoader classLoader;
```
denis-anisimov, line 32: Do you need this class to be `Serializable`?
Does this class even implements `Serializable`.

I think the class should not be `Serializable` : it's not stored anywhere.

So if the class is serializable then you should care about `transient` fields: they won't be restored after deseriallization.
If class should not be serialzable then `transient` is not needed here.

denis-anisimov (changes requested):
There are still things which may be improved.
Though things are done properly considering the required architecture decision (and current limitations with ugly `Builder` class).

Only some things related to the implementation.

I've not reviewed unit tests since they are related to the Fusion fincitonality.
These should be reviewed by the Fusion team.
But some introduced API usage looks untested (in DevModeInitializer) : this is an architectural issue.

fusion-endpoint/src/main/java/com/vaadin/flow/server/frontend/fusion/EndpointGeneratorTaskFactoryImpl.java
denis-anisimov: I see you have some checks for parameters in the CTOR.
But : is it enough?
And anyway the javadocs doesn't say anything about that.

flow-server/pom.xml
haijian-vaadin: They are used in the [JsonUtils](https://github.com/vaadin/flow/blame/feature/offline/flow-server/src/main/java/com/vaadin/flow/internal/JsonUtils.java#L34), there will be compile errors without them.

How do I check the OSGi manifest?

haijian-vaadin: I tried to run the testing project, looks like it works fine

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
haijian-vaadin: The tests were moved to [DevModeInitializerEndpointTest](https://github.com/vaadin/flow/blob/haijian/fusion-server-module/fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java)

fusion-endpoint/src/main/java/com/vaadin/flow/server/frontend/fusion/TaskGenerateOpenApiImpl.java
haijian-vaadin: Done, it was inherited from the existing code, now removed.

fusion-endpoint/src/main/java/com/vaadin/flow/server/frontend/fusion/TaskGenerateOpenApiImpl.java
haijian-vaadin: done

fusion-endpoint/src/main/java/com/vaadin/flow/server/frontend/fusion/TaskGenerateConnectImpl.java
haijian-vaadin: done

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
haijian-vaadin: True, good point, I will add tests to verify that certain methods get called

haijian-vaadin: Done

platosha (changes requested):

flow-server/pom.xml
denis-anisimov: The manifest should have a number of entries like 
`Bundle-SymbolicName: com.vaadin.flow.server`
`Export-Package: com.vaadin.flow.component`,
....

But if everything works in OSGi container it means that manifest is correct.

flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojo.java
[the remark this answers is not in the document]
denis-anisimov: I thought the classes have( or could have) a common superclasess where the field may be used (may be as a protected field even though I don't like protected). And then it will be possible.
But that's minor.

fusion-endpoint/src/main/java/com/vaadin/flow/server/frontend/fusion/EndpointGeneratorTaskFactoryImpl.java
haijian-vaadin: done, added null check

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
haijian-vaadin: Removed the ClassFinder parameter, use Lookup to look for ClassFinder

flow-server/src/main/java/com/vaadin/flow/internal/ReflectTools.java
haijian-vaadin: updated the java doc

flow-maven-plugin/src/main/java/com/vaadin/flow/utils/LookupImpl.java
haijian-vaadin: didn't make it a special case, but instead using Lookup.compose() when creating the Lookup

flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
haijian-vaadin: done

denis-anisimov (commented):
I'm fine with the code now.

As mentioned : the business logic should be reviewed by the Fusion team (and the approval should go from there).

From the architectural point of view it's OK for me.

platosha (approved):

PR #9500, opened 2020-11-26
fix: Initialize VaadinServlet after Lookup is available

denis-anisimov:
* In WAR case init will be called as usual without any changes
* IN OSGi ServletContextListener::contextInitialized may be called after Servet::init is called and in this case info necessary for the servlet won't be available yet.
* ServletContextListener::contextInitialized  will call Servet::init one more time to complete initialization.

### DIFF
--- a/flow-polymer-template/src/test/java/com/vaadin/flow/server/MockVaadinServletService.java
+++ b/flow-polymer-template/src/test/java/com/vaadin/flow/server/MockVaadinServletService.java
@@ -15,12 +15,12 @@
  */
 package com.vaadin.flow.server;
 
-import java.util.Collections;
-import java.util.List;
-
 import javax.servlet.ServletConfig;
 import javax.servlet.ServletContext;
 import javax.servlet.ServletException;
+
+import java.util.Collections;
+import java.util.List;
 
 import org.mockito.Mockito;
 
@@ -95,8 +95,6 @@ public class MockVaadinServletService extends VaadinServletService {
 
     public void init(Instantiator instantiator) {
         this.instantiator = instantiator;
-
-        init();
     }
 
     @Override
--- a/flow-polymer-template/src/test/java/com/vaadin/flow/server/MockVaadinServletService.java
+++ b/flow-polymer-template/src/test/java/com/vaadin/flow/server/MockVaadinServletService.java
@@ -95,6 +95,8 @@ public class MockVaadinServletService extends VaadinServletService {
 
     public void init(Instantiator instantiator) {
         this.instantiator = instantiator;
+
+        init();
     }
 
     @Override
@@ -110,15 +112,18 @@ public class MockVaadinServletService extends VaadinServletService {
         try {
             MockVaadinServlet servlet = (MockVaadinServlet) getServlet();
             servlet.service = this;
-            ServletConfig config = Mockito.mock(ServletConfig.class);
-            ServletContext context = Mockito.mock(ServletContext.class);
-            Mockito.when(config.getServletContext()).thenReturn(context);
 
-            Mockito.when(lookup.lookup(ResourceProvider.class))
-                    .thenReturn(resourceProvider);
-            Mockito.when(context.getAttribute(Lookup.class.getName()))
-                    .thenReturn(lookup);
-            getServlet().init(config);
+            if (getServlet().getServletConfig() == null) {
+                ServletConfig config = Mockito.mock(ServletConfig.class);
+                ServletContext context = Mockito.mock(ServletContext.class);
+                Mockito.when(config.getServletContext()).thenReturn(context);
+
+                Mockito.when(lookup.lookup(ResourceProvider.class))
+                        .thenReturn(resourceProvider);
+                Mockito.when(context.getAttribute(Lookup.class.getName()))
+                        .thenReturn(lookup);
+                getServlet().init(config);
+            }
             super.init();
         } catch (ServiceException | ServletException e) {
             throw new RuntimeException(e);
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
@@ -67,17 +67,19 @@ public class VaadinServlet extends HttpServlet {
     public void init(ServletConfig servletConfig) throws ServletException {
         CurrentInstance.clearAll();
 
-        if (getServletConfig() == null) {
-            super.init(servletConfig);
-        }
+        try {
+            if (getServletConfig() == null) {
+                super.init(servletConfig);
+            }
 
-        ServletContext servletContext = getServletConfig().getServletContext();
-        if (new VaadinServletContext(servletContext)
-                .getAttribute(Lookup.class) == null) {
-            return;
-        }
+            ServletContext servletContext = getServletConfig()
+                    .getServletContext();
+            if (servletService != null
+                    || new VaadinServletContext(servletContext)
+                            .getAttribute(Lookup.class) == null) {
+                return;
+            }
 
-        if (servletService == null) {
             try {
                 servletService = createServletService();
             } catch (ServiceException e) {
@@ -92,9 +94,9 @@ public class VaadinServlet extends HttpServlet {
             staticFileHandler = createStaticFileHandler(servletService);
 
             servletInitialized();
+        } finally {
+            CurrentInstance.clearAll();
         }
-
-        CurrentInstance.clearAll();
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
@@ -68,6 +68,22 @@ public class VaadinServlet extends HttpServlet {
         CurrentInstance.clearAll();
 
         try {
+            /*
+             * There are plenty of reasons why the check should be done. The
+             * main reason is: init method is public which means that everyone
+             * may call this method at any time (including an app developer).
+             * But it's not supposed to be called any times any time.
+             * 
+             * This code protects weak API from being called several times so
+             * that config is reset after the very first initialization.
+             * 
+             * Normally "init" method is called only once by the servlet
+             * container. But in a specific OSGi case {@code
+             * ServletCointextListener} may be called after the servlet
+             * initialized. To be able to initialize the VaadinServlet properly
+             * its "init" method is called from the {@code
+             * ServletCointextListener} with the same ServletConfig instance.
+             */
             if (getServletConfig() == null) {
                 super.init(servletConfig);
             }
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
@@ -16,6 +16,7 @@
 package com.vaadin.flow.server;
 
 import javax.servlet.ServletConfig;
+import javax.servlet.ServletContext;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
@@ -27,6 +28,7 @@ import java.net.URL;
 import java.util.Properties;
 
 import com.vaadin.flow.component.UI;
+import com.vaadin.flow.di.Lookup;
 import com.vaadin.flow.function.DeploymentConfiguration;
 import com.vaadin.flow.internal.CurrentInstance;
 import com.vaadin.flow.server.HandlerHelper.RequestType;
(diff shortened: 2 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -64,23 +66,35 @@
     @Override
     public void init(ServletConfig servletConfig) throws ServletException {
         CurrentInstance.clearAll();
-        super.init(servletConfig);
 
-        try {
-            servletService = createServletService();
-        } catch (ServiceException e) {
-            throw new ServletException("Could not initialize VaadinServlet", e);
+        if (getServletConfig() == null) {
```
caalador, line 70: init is called by the container so why would this be called more than once. 
Also I don't know the reason why we are overriding `init(ServletConfig)` except to get the config when we could just override `init()` and get the config with `getServletConfig()` and not have to call `super.init`

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -64,23 +66,35 @@
...
+            super.init(servletConfig);
+        }
+
+        ServletContext servletContext = getServletConfig().getServletContext();
+        if (new VaadinServletContext(servletContext)
+                .getAttribute(Lookup.class) == null) {
+            return;
```
caalador, line 77: Who will call again for the `init` of a VaadinServlet?
The container should only inform once that you are being taken into use.

caalador (commented):
I guess the topic is `Initialize VaadinServlet after Lookup is available`, but who will recall the init method of VaadinServlet again?
Also some information what the problem is in what situation would be helpful.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
pleku: Yeah this check makes no sense according to the javadocs of the init methods

denis-anisimov:
> I guess the topic is `Initialize VaadinServlet after Lookup is available`, but who will recall the init method of VaadinServlet again?
> Also some information what the problem is in what situation would be helpful.

Done: in the title and description

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
denis-anisimov: Yeah, why `init` is not overriden instead is not clean.
But I think it allows to override `init` without caring to call `super.init()`.
It's not supposed that you are going to override `super.init(ServletConfig)`.

In "plain" WAR case nothing is changed: `init` will be called once by the container and it never will be called one more time.

This is done for OSGi case: even though if `ServletContextListener` (which is registered via HTTP Whiteboard spec) is instantiated _before_ the servlet is registered the `ServletContextListener::contextInitialized` method may be called after `Servlet::init` which is quite unexpected.
That doesn't allow to put all necessary info into the `ServletContext` before the servlet is initialized.
In the latter case `ServletContextListener::contextInitialized` will call `Servlet::init`  if the servlet is not yet initialized.

pleku (changes requested):
This fix lacks an issue it is addressing and the details on what is happening.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
denis-anisimov: See above.

`init` methods are evil: it's public method which may be called by anyone at any time.
It's supposed that it will be called by the container. But Java API doesn't prevent to call this method by anyone at any point.
So in fact this method should be normally implemented to be aware of the fact that it can be called several times.
`init` methods make API weak overall.

In WAR case nothing changes.
This is needed for OSGi only

denis-anisimov:
> This fix lacks an issue it is addressing and the details on what is happening.

Quite hard to address OSGi issue in non OSGi repo.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
denis-anisimov: >Yeah this check makes no sense according to the javadocs of the init methods

I have no any idea what you are talking about .
Which javadoc ?

This  check allows to check whether this is the first time of method call or not.

pleku:
>Quite hard to address OSGi issue in non OSGi repo.

You can refer to and close an issue in another repository with syntax e.g. `Fixes vaadin/osgi#123`.

Would it be possible to have a `OsgiVaadinServlet` (in flow-osgi) and use that so that the changes would go there ?

denis-anisimov:
> > Quite hard to address OSGi issue in non OSGi repo.
> 
> You can refer to and close an issue in another repository with syntax e.g. `Fixes vaadin/osgi#123`.
> 
> Would it be possible to have a `OsgiVaadinServlet` (in flow-osgi) and use that so that the changes would go there ?

The issue can't be fixed by one fix in Flow.
Extracting OSGi into osgi repo complicated things a lot.
To be able to fix one issue I have to make two PRs in different repos.

`OsgiVaadinServlet` will solve the issue. And that means that developer has to be aware of the this class and may not use `VaadinServlet` for OSGi . That will produce user errors.
I'm trying to reduce OSGi specific classes usage as much as possible.

`VaadinServlet` will work exactly in the same way in WAR case without any single behavioral change.
This change affects only OSGi. I don't see serious reasons for now to introduce a new servlet class for OSGi at the moment.

I can introduce one more method `doInit` (or whatever) in the `VaadinServlet` which will do this intialization.

But if API is designed so that there is a public method then this method may be called any times by anyone.
I just don't want to introduce one more extra method with almost the same semantic.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -64,23 +66,35 @@
...
         }
 
-        // Sets current service as it is needed in static file server even
-        // though there are no request and response.
-        servletService.setCurrentInstances(null, null);
+        if (servletService == null) {
```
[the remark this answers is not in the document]
denis-anisimov: Actually this comment  allowed me to find a bug: `return` may be called without calling `CurrentInstance.clearAll()`.

denis-anisimov:
>is the initial context more valuable than the secondary?
>Shouldn't they be the same so setting it shouldn't be a problem as super.init(context) only stores the reference.

by the `context`  you mean `config`  ?
This question is about weakness of `init` method API.
`init` is always called first time by the servlet container.
`init`  method should be protected to avoid changing the `config` : because there is no way to get info that the config is changed. All these things had to be solved via Java API. But they are not.
This is done by the code (in runtime) : `config` should be set only once. And this is exactly done.
Again: _anyone_ may call this method with anything passed there. And this is the problem of API.

This PR prevents changing the `config` after the very first call. And it's meaningful even if there is no any need to call this method in OSGi.

In OSGi code I call `init`  from the method which is called also by the Servlet container second time with exactly the same `config` parameter passed.

What can I do here in addition : check that if `init`  is called second time then `config` instance is exactly the same as the one  which is already stored.
It makes sense.

pleku (changes requested):

pleku:
>OsgiVaadinServlet will solve the issue. And that means that developer has to be aware of the this class and may not use VaadinServlet for OSGi . That will produce user errors.

This is the same as for Portlet ... and has been. And in the same way it is already the issue with `VaadinServlet` for anything.
So I don't know if it is really an issue - and if we we're to register the servlet automatically for vaadin-osgi then they would not need to know about it until they decide to register one themselves ?

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -64,23 +66,43 @@
...
     public void init(ServletConfig servletConfig) throws ServletException {
         CurrentInstance.clearAll();
-        super.init(servletConfig);
 
         try {
-            servletService = createServletService();
-        } catch (ServiceException e) {
-            throw new ServletException("Could not initialize VaadinServlet", e);
-        }
+            if (getServletConfig() == null) {
```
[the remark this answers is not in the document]
denis-anisimov: I will add it but it's useless.
OSGi is in the different repo. Without looking at the code and knowing the context the comment means anything to someone who will see this code first time.

Or may be the comment should be formulated in a totally different way. In that case it makes sense.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -64,23 +66,43 @@
...
-        CurrentInstance.clearAll();
+            // Sets current service as it is needed in static file server even
+            // though there are no request and response.
+            servletService.setCurrentInstances(null, null);
 
+            staticFileHandler = createStaticFileHandler(servletService);
+
+            servletInitialized();
+        } finally {
+            CurrentInstance.clearAll();
```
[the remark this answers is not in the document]
denis-anisimov: Sure.
https://github.com/vaadin/flow/issues/9503

This is not a new code and it had no unit tests so this is a separate issue.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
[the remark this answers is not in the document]
pleku: In case there is no comment, it will be impossible to know that the reason for this code is to support some code that is in another repository. The tests in this repository should prevent them from breaking things, but they will not still be able to understand why the code is like that when it looks like it should not be (based on the javadocs for servlet). Then they have to spend time to search history/blame for the changes that were done and check the review for help on understanding. That will take time that a simple comment could have spared them from.

denis-anisimov:
> > OsgiVaadinServlet will solve the issue. And that means that developer has to be aware of the this class and may not use VaadinServlet for OSGi . That will produce user errors.
> 
> This is the same as for Portlet, and Spring ... and has been pretty much always. And in the same way it is already the issue with `VaadinServlet` for anything.
> So I don't know if it is really an issue - and if we we're to register the servlet automatically for vaadin-osgi then they would not need to know about it until they decide to register one themselves ?

It is an issue. A proven one.

`Portlet` : nobody uses Portlets . The number of user is vanishingly small. So this issue just had no chance to appear.

Spring: exactly the case.
Spring has it's own servlet because of just one reason: it's not supposed to be used at all.
You don't register and never uses the servlet in Spring normally. It's registered for you automatically and you may pass to it all the configuration which you want.
And there are quite many cases when users wanted to use a custom servlet for Spring  every now and then.
And it's quite common mistake that they register a `VaadinServlet` in this case instead of `SpringServlet`.

OSGi: we don't have at the moment automatic servlet registration. But regardless of automatic servlet registrations I'm pretty sire there will be a lot of users who will want to register the servlet manually .
It's easier for users to have only one class instead of number of public classes for every framework.

I would just had made OSGiVaadinServlet without any doubt if it have been just an impl detail with no visibility.
But it's not the case.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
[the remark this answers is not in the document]
denis-anisimov: Yes and no.
The comment should be extremely simple and give the idea why it's done this way and not another without going into details.
I was against of "mentioning" OSGi here: because it's useless. If you don't even know what OSGi is the comment gives you 0 chances to make anything clearer.
But I've added the comment which makes OSGi just a detail. So that way it has sense.

flow-polymer-template/src/test/java/com/vaadin/flow/server/MockVaadinServletService.java
```diff
@@ -120,7 +120,10 @@ public void init() {
                     .thenReturn(resourceProvider);
             Mockito.when(context.getAttribute(Lookup.class.getName()))
                     .thenReturn(lookup);
-            getServlet().init(config);
+            if ( getServlet().getServletConfig() ==null) {
```
caalador, line 123: This could instead just be a check at the beginning of init that if 
`if(this.equals(getServlet())) return;`
Basically we have gone wrong at creation as the constructor init will create an instantiator that we then set later again. Perhaps we should have a constructor that takes in the instantiator so init is only called once by the tests.

caalador:
Note that the Mock service is in 3 places flow-server, polymer and lit

flow-polymer-template/src/test/java/com/vaadin/flow/server/MockVaadinServletService.java
denis-anisimov: I didn't get how `this` which is instance of `VaadinService` may equal a servlet instance but this doesn't matter.
I've redone this.

flow-polymer-template/src/test/java/com/vaadin/flow/server/MockVaadinServletService.java
```diff
@@ -95,8 +95,6 @@ protected RouteRegistry getRouteRegistry() {
 
     public void init(Instantiator instantiator) {
         this.instantiator = instantiator;
-
-        init();
```
caalador, line 99: This makes the init use the wrong instantiator for all the initial things making this init method basically useless.

denis-anisimov: Tests are passing........

Well, I will revert everything back since I don't understand your previous comment.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -68,6 +68,22 @@ public void init(ServletConfig servletConfig) throws ServletException {
...
+             * 
+             * This code protects weak API from being called several times so
+             * that config is reset after the very first initialization.
+             * 
+             * Normally "init" method is called only once by the servlet
+             * container. But in a specific OSGi case {@code
+             * ServletCointextListener} may be called after the servlet
+             * initialized. To be able to initialize the VaadinServlet properly
+             * its "init" method is called from the {@code
+             * ServletCointextListener} with the same ServletConfig instance.
```
pleku, line 85: Non-blocking: typoed servletcontextlistener. Also I don't think it is necessary to lecture about how "anyone call this method at any time" and "weak API". I would have just settled for the reason:
 "This init method is called twice in some scenarios, like with OSGi {@code ServletContextListener} because not all things are available when the container triggers the first init. Duplicate initialization needs to be avoided and creation of service needs lookup to be present."

pleku (approved):

flow-polymer-template/src/test/java/com/vaadin/flow/server/MockVaadinServletService.java
caalador: I guess the usage of the init(Instantiator) should be that we wouldn't execute `VaadinService.createInstantiator()`  but we now execute this as the first line in `VaadinService.init()` is `instantiator = createInstantiator();` 

ofc it blocks any other instantiations later in the tests.

caalador (approved):

PR #9516, opened 2020-11-27
fix: Add empty list handling to DataCommunicator#collectKeysToFlush

TatuLund:
Fixes: https://github.com/vaadin/vaadin-grid/issues/1837


mshabarov:
Added an IT (regression) test on components side https://github.com/vaadin/vaadin-flow-components/pull/570.
I believe it's not necessary to add any unit tests on the Flow side for this particular case, because almost everything in DC is private and finally the unit test will become a full-fledged IT test anyway.

mshabarov (approved):

mshabarov:
Cherry picked to `2.4`

PR #9524, opened 2020-11-28
Fixed flaky tests caused by gc:ed variable ui in ValidatorTestBase.java

KellyShao:
Fixed flaky test that caused by `com.vaadin.flow.component.UI` in class `ValidatorTestBase`. 

- **[Issue]** 
If only run `mvn -pl flow-data test -Dtest=BeanValidatorTest#testInvalidDecimalsFailsInFrench` command, the [`testInvalidDecimalsFailsInFrench`](https://github.com/vaadin/flow/blob/be4ddc785a22728b999e37551bb0d85bb3e4def6/flow-data/src/test/java/com/vaadin/flow/data/validator/BeanValidatorTest.java#L63) will failed because of the incorrect language setup 
    - **error msg**: 
    ```
        [ERROR] Failures: 
        [ERROR] BeanValidatorTest.testInvalidDecimalsFailsInFrench:66->ValidatorTestBase.assertFails:59 expected:<[valeur numérique hors limite (<3 chiffres>.<2 chiffres> attendu])> but was:<[numeric value out of bounds (<3 digits>.<2 digits> expected])>
    ```
    - **solution**: 
Added a `ui` variable in the `ValidatorTestBase` class to attach strong reference, in order to **avoid garbage collecting** it during the test run. 

The flaky test is found by running [iDFlakies](https://github.com/idflakies/iDFlakies), after fixing the issue, there's no flaky tests in `BeanValidatorTest.java`

### DIFF
--- a/flow-data/src/test/java/com/vaadin/flow/data/validator/ValidatorTestBase.java
+++ b/flow-data/src/test/java/com/vaadin/flow/data/validator/ValidatorTestBase.java
@@ -30,10 +30,11 @@ public class ValidatorTestBase {
 
     private TestLabel localeContext;
     private Locale locale = Locale.US;
+    private UI ui;
 
     @Before
     public void setUp() {
-        UI ui = new UI() {
+        ui = new UI() {
             @Override
             public Locale getLocale() {
                 return locale;


joheriks:
@KellyShao Thank you for your contribution. However, I am unable to provoke the test failure with the provided Maven command (`mvn -pl flow-data test -Dtest=BeanValidatorTest#testInvalidDecimalsFailsInFrench`) that runs the test in isolation. It would be useful to see the the iDFlakies command (`mvn testrunner:testplugin -Ddetector.detector_type=...`) that flagged this test as potentially flaky.

Legioth:
Based on the nature of the fix, I would guess the problem is caused by a background garbage collection causing `UI.getCurrent()` to suddenly become `null`. If that is the case, then you can further provoke it by adding a `System.gc()` or two to the beginning of the `@Test` method.

An alternative way of fixing would be to use `MockUI` which internally attaches itself to a `MockVaadinSession` instance that has an internal thread local specifically to prevent it (or it's associated UIs) from being garbage collected.

joheriks:
Right, that is certainly not obvious so a comment along the lines `reference kept to prevent being gc:ed before calling UI.getCurrent()`  on the `UI` as IDEs may suggest refactoring it back to a local variable.

Does `MockUI` actually fix this, as the references are `[ UI ] --> [ VaadinSession ]` so even holding a reference to a `VaadinSession` wouldn't prevent  `UI` from being gc:ed?

KellyShao:
Thanks for all your suggestions! 
I'm still working on this issue now. Meanwhile, in order to provide more info, I've attached some log files regarding this issue FYI.

- **Maven command** (`mvn -pl flow-data test -Dtest=BeanValidatorTest#testInvalidDecimalsFailsInFrench`) with diff. java environment:
    1. jdk8: 
        - env info:
        ```
        java version "1.8.0_261"
        Java(TM) SE Runtime Environment (build 1.8.0_261-b12)
        Java HotSpot(TM) 64-Bit Server VM (build 25.261-b12, mixed mode)
        ```
        - log file (failed): [single-test-testInvalidDecimalsFailsInFrench-jdk8.txt](https://github.com/vaadin/flow/files/5627313/single-test-testInvalidDecimalsFailsInFrench-jdk8.txt)
    2. jdk11: 
        - env info:
        ```
       openjdk version "11.0.1" 2018-10-16
       OpenJDK Runtime Environment 18.9 (build 11.0.1+13)
       OpenJDK 64-Bit Server VM 18.9 (build 11.0.1+13, mixed mode)
        ```
        - log file (failed):[single-test-testInvalidDecimalsFailsInFrench-jdk11.txt](https://github.com/vaadin/flow/files/5627337/single-test-testInvalidDecimalsFailsInFrench-jdk11.txt)
     3. jdk13: 
        - env info:
        ```
        java version "13.0.2" 2020-01-14
        Java(TM) SE Runtime Environment (build 13.0.2+8)
        Java HotSpot(TM) 64-Bit Server VM (build 13.0.2+8, mixed mode, sharing)
        ```
        - log file (failed):[single-test-testInvalidDecimalsFailsInFrench-jdk13.txt](https://github.com/vaadin/flow/files/5627348/single-test-testInvalidDecimalsFailsInFrench-jdk13.txt)

-  **iDFlakies command**(`mvn testrunner:testplugin -Ddetector.detector_type=random-class-method -Ddt.detector.original_order.all_must_pass=false -Ddt.randomize.rounds=50 -pl flow-data`):
    - log file (with failure in 1st round): [idflakies-flow-data-50.txt](https://github.com/vaadin/flow/files/5627366/idflakies-flow-data-50.txt)
    - `[INFO] Found 1 tests in round 1 of 50` means iDFlakies found flaky tests

KellyShao:
>  as the references are [ UI ] --> [ VaadinSession ] so even holding a reference to a VaadinSession (MockVaadinSession) wouldn't prevent UI from being gc:ed?

Yes, exactly! I tried to use `MockUI` and reference it to a `VaadinSession (MockVaadinSession)` as in https://github.com/vaadin/flow/blob/3bc2e60af3600735fc5b188ebb31b48059a13349/flow-data/src/test/java/com/vaadin/flow/data/provider/AbstractListDataViewListenerTest.java#L157

However, it still `gc:ed`, since `VaadinSession` is **stored using a weak reference** to avoid leaking memory in case it is not explicitly cleared per the comment in https://github.com/vaadin/flow/blob/9f7cfa76fd21d8d18338237e6e325c1eeaf5c21b/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java#L458

Legioth:
`MockVaadinSession` has it's own `ThreadLocal` to ensure a non-weak reference whenever the session is locked:

https://github.com/vaadin/flow/blob/ae47abc0ee03aafe3dbbe1f503804a9db8cf0a92/flow-server/src/test/java/com/vaadin/flow/server/MockVaadinSession.java#L39

Combine this with `AlwaysLockedVaadinSession` which isn't actually _always_ locked but rather locks itself when it's created, and you should end up with a session that won't be garbage collected during the lifetime of the thread that created the session.

When a `UI` is attached to a session, there are references both ways. The `UI` class has an `internals` field which in turn has a `session` field. The session on the other hand has a `uIs` map that references all UI instances attached to that session.

Legioth:
Now I see. `MockUI` is never attaching itself to its session, which means that there will actually not be any reference from the session to the UI. I wonder whether it should be updated to always do `session.addUI(this)`?

(That's definitely a separate issue to consider. Declaring an explicit instance field for the ui like it's done in this PR is also fine.)

KellyShao:
> I wonder whether it should be updated to always do session.addUI(this)?

Thanks for pointing it out! Based on my understanding, `session.addUI(ui)` requires `ui` to be initialized, otherwise there's an exception with msg `Can not add an UI that has not been initialized.`  per https://github.com/vaadin/flow/blob/079c7d260a42ee87c474844ef93bc2275ce44b3a/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java#L814

- So probably, we should add it within the `@Before` method after the `ui` object is created, then use `com.vaadin.flow.component.UI.doInit()`: https://github.com/vaadin/flow/blob/079c7d260a42ee87c474844ef93bc2275ce44b3a/flow-server/src/main/java/com/vaadin/flow/component/UI.java#L207
- Then, we further need to make more changes to use `MockUI(VaadinSession session)` instead of `MockUI()` to bind the session, as it implemented in here: https://github.com/vaadin/flow/blob/3bc2e60af3600735fc5b188ebb31b48059a13349/flow-data/src/test/java/com/vaadin/flow/data/provider/AbstractListDataViewListenerTest.java#L163

> That's definitely a separate issue to consider. Declaring an explicit instance field for the ui like it's done in this PR is also fine.

The current issue on my side is that initialize UI get involved in many `request` & `service` related knowledge that need more time to figure it out. In order not to explode this PR, an alternative way is to submit an issue regarding this problem.

joheriks (changes requested):
The proposed fix is already good and addresses the issue discovered. I'm be ready to merge this change, just missing note in the test base class that the `UI` reference is there to avoid garbage collecting it during test run (since this is not immediately obvious).

KellyShao:
> The proposed fix is already good and addresses the issue discovered. 

Sure! I've added a **solution** section in my topmost comment addressing the issue: 
> - solution:
> Added a ui variable in the ValidatorTestBase class to add a strong reference, in order to avoid garbage collecting it during the test run.

joheriks (approved):

PR #9543, opened 2020-11-30
feat: Enable importing CSS from node_modules

### LINKED ISSUE #9410: Importing css to application theme from npm packages
There is a `theme.json` file inside the theme folder that has the following property for importing things from npm packages.
```json
{ "importCss": ["@foo/bar/baz.css"] }
```
This does not install any packages, that is left for the user. In case the package is not there, we should fail with error message that tells the user to install the package with pnpm / npm.

The css is imported as global scoped css to the theme.

tanbt:
Allow importing CSS files as global styles from installed node_modules by adding the `importCss` to theme/my-theme/theme.json.

Fixes #9410


flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -46,7 +46,7 @@ export const injectGlobalCss = (css, target) => {
  * @param {string} themeName name of the handled theme
```
caalador, line 46: No added param documentation for themeProperties.

flow-tests/test-themes/src/test/java/com/vaadin/flow/uitest/ui/theme/ThemeIT.java
```diff
@@ -38,11 +39,11 @@
     public void secondTheme_staticFilesNotCopied() {
         getDriver().get(getRootURL() + "/path/VAADIN/static/img/bg.jpg");
         Assert.assertFalse("app-theme static files should be copied",
-            driver.getPageSource().contains("HTTP ERROR 404 Not Found"));
```
caalador, line 41: Lots of unrelated formatting changes in the test file.
Could you check the TeamWiki Code style section. indentation at least seems different

caalador:
webpack.generated.js might need to have in css-loader also the part:
```

              import: (url, media, resourcePath) => {
                // resourcePath - path to css file
                // If the import happens from within a node_modules file, we must resolve and inline. Otherwise resources will not be found at runtime as node_modules is not deployed
                return resourcePath.includes('/node_modules/');
              }
```

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
tanbt: Added.

flow-tests/test-themes/src/test/java/com/vaadin/flow/uitest/ui/theme/ThemeIT.java
tanbt: Reverted. My formatter setting was disappear somehow.

caalador:
Perhaps it needs the rule:
```
      {
        // File-loader only copies files used as imports in .js files
        test: /\.(png|gif|jpg|svg|eot|woff|woff2|ttf)$/,
        use: [ {
          loader: 'file-loader',
          options: {
            outputPath: '../resources/VAADIN/static/',
            name: '[name].[ext]'
          }
        }],
      },
```

caalador:
You are missing the file-loader from NodeUpdater getDefaultDevDependencies
Guess it can be "file-loader", "6.1.0"

caalador:
It would be good to run the tests locally to see that all pass as now there is " in the actual string gotten from the browser which was not expected by the test. 
For the second failure you should set the webpack logging to have debug and see that the subcss is handled by the theme-loader

tanbt:
Is it possible to test only the `test-themes` module?
I used
```
cd flow-tests/test-themes
mvn test
```
but looks like no test is executed.

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -173,7 +173,8 @@ module.exports = {
             options: {
               url: (url, resourcePath) => {
                 // Only translate files from node_modules
-                return resourcePath.includes('/node_modules/');
+                return resourcePath.includes('/node_modules/')
+                    || resourcePath.includes('/frontend/theme/');
```
caalador, line 177: This has the problem that "external" url's are not filtered out making it fail as the files can not be found.

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -189,6 +190,17 @@ module.exports = {
           }
         ],
       },
+      {
+        // File-loader only copies files used as imports in .js files
+        test: /\.(png|gif|jpg|svg|eot|woff|woff2|ttf)$/,
+        use: [{
+          loader: 'file-loader',
+          options: {
+            outputPath: '../resources/VAADIN/static/',
```
caalador, line 199: This was my mistake, but the output path should be `outputPath: 'static/',` also I can't get this to actually copy anything.

flow-server/src/main/resources/webpack.generated.js
tanbt: Added a check to skip external URL.

flow-server/src/main/resources/webpack.generated.js
tanbt: Fixed.
I can still see files are copied to `target\classes\META-INF\VAADIN\static` when using `outputPath: '../resources/VAADIN/static/'`.

caalador: that's because it's not file-loader that is doing the copying.

flow-server/src/main/resources/webpack.generated.js
caalador: Only `http://` doesn't mean external also `./served/by/server` or `/absolute` is external from the view of the theme see test module octopuss and monarch-butterfly

caalador:
As I look at this it might be that the feature was fine at commit d8474de and that it's the test that is wrong as the font icons style sheet probably needs to be in the head and not where it lands with this PR.

flow-server/src/main/resources/webpack.generated.js
tanbt: Will update with https://github.com/vaadin/flow/pull/9558.

tanbt:
> style sheet probably needs to be in the head

Do you mean the stylesheet should be in a `<link rel="stylesheet" ...>` and placed at `target.head`? 
I tried locally but a link like `<link rel="stylesheet" src="@fortawesome/fontawesome-free/css/all.css">` can't be resolved.

flow-server/src/main/resources/webpack.generated.js
caalador: Let's take in 9558 and you can rebase the fix on that.

caalador:
see #9559 for a working fontawesome test

flow-tests/test-themes/frontend/theme/app-theme/global.css
```diff
@@ -12,7 +12,7 @@ body {
 
 #butterfly {
     /* This shouldn't be translated */
-    background-image: url("./test/path/monarch-butterfly.jpg");
+    background-image: url("/path/test/path/monarch-butterfly.jpg");
```
caalador, line 15: As this should stay unchanged the path should get path from the application as it should be requested with the application context. There is a test that `./` should match `RootUrl+/path` when executed in browser context.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -303,6 +303,7 @@ static void addVaadinDefaultsToJson(JsonObject json) {
         defaults.put("compression-webpack-plugin", "4.0.1");
         defaults.put("webpack-merge", "4.2.2");
         defaults.put("css-loader", "4.2.1");
+        defaults.put("file-loader", "6.1.0");
```
caalador, line 306: Remove the extra file-loader as it was added from the merge in master. see 3 lines down

flow-tests/test-themes/src/test/java/com/vaadin/flow/uitest/ui/theme/ThemeIT.java
```diff
@@ -79,6 +80,16 @@ public void applicationTheme_GlobalCss_isUsed() {
...
     }
 
+    @Test
+    public void applicationTheme_importCSS_isUsed() {
+        open();
+        checkLogsForErrors();
+
+        Assert.assertEquals("Imported FontAwesome css file should be applied.",
+                "\"Font Awesome 5 Brands\"", $(SpanElement.class).id(FONTAWESOME_ID)
+                        .getCssValue("font-family"));
```
caalador, line 90: Should this check for something more. Are we certain the resource is available just because the css has been applied and no exceptions are on the console?

Should we also check the resource e.g.
```
        getDriver().get(getRootURL() + "/path/VAADIN/static/[THE_RESOURCE_THAT_IS_EXPECTED]");
        Assert.assertFalse("Font resource should be available",
            driver.getPageSource().contains("HTTP ERROR 404 Not Found"));
```

flow-tests/test-themes/src/main/java/com/vaadin/flow/uitest/ui/theme/ThemeView.java
```diff
@@ -45,12 +46,16 @@ public ThemeView() {
         Span octopuss = new Span();
         octopuss.setId(OCTOPUSS_ID);
 
+        Span faText = new Span("This test is FontAwesome.");
+        faText.setClassName("fab");
```
caalador, line 50: Does this give something visible or should we add a icon class name also?

caalador (changes requested):

caalador (changes requested):
Update the version number of the application-theme package.json with +0.0.1

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
taefi: Done.

flow-tests/test-themes/frontend/theme/app-theme/global.css
taefi: I'm not sure about this, since the `monarch-butterfly.jpg` is a non-theme resource under `src/webapp` and that leading `/path` is actually a folder under `webapp`.

caalador: But the point is that the url is relative to the application and /path/ is from the application servlet path.

caalador: Also this feature shouldn't touch the external url tests as this shouldn't change them in any way.

taefi: > /path/ is from the application servlet path
The project structure is like this:
```
project-root
|_ src
    |_ webapp
        |_ path
            |_ test
                |_ path
                    |_ monarch-butterfly.jpg
```
So I think I got confused because `@WebServlet("/path/*")` was the URL-pattern and not the context-path. So I removed the `/path` and changed it back to `./` as it was before.

caalador: I agree it's a bit confusing and should perhaps have a better comment on top.

flow-tests/test-themes/src/main/java/com/vaadin/flow/uitest/ui/theme/ThemeView.java
taefi: Added an Icon. Done.

flow-tests/test-themes/src/test/java/com/vaadin/flow/uitest/ui/theme/ThemeIT.java
taefi: Done.

caalador (approved):

Issue #9533, opened 2020-11-30
Add theme feature to inject css into the document


caalador:
There should be a feature to import css into the root element of the page.

There should be a possibility add to the `theme.json` :
```
"documentCss": ["@fortawesome/fontawesome-free/css/all.css"]
```

This would add the css into the `document.documentElement` which is the root element of the document (for example, the <html> element for HTML documents).

PR #9558, opened 2020-12-02
feat: Resource handling to not need plugin copying

caalador:
Fixed the url handling so that theme resources
get prepended with theme/[themeName] while
having the correct absolute path. With this we can handle
these url resources with the css-loader which in turn
leads to file-loader gettign the files for copying.

External url are still not touched in any way.

part of #9410 and #9533

### DIFF
--- a/flow-server/src/main/resources/webpack.generated.js
+++ b/flow-server/src/main/resources/webpack.generated.js
@@ -210,7 +210,7 @@ module.exports = {
             name(resourcePath, resourceQuery) {
               const urlResource = resourcePath.substring(frontendFolder.length);
               if(urlResource.match(themePartRegex)){
-                return /[\s\S]*(\\|\/)theme\1[\s\S]*?\1(.*)/.exec(resourcePath)[2];
+                return /^(\\|\/)theme\1[\s\S]*?\1(.*)/.exec(urlResource)[2];
               }
               return '[path][name].[ext]';
             }
--- a/flow-server/src/main/resources/webpack.generated.js
+++ b/flow-server/src/main/resources/webpack.generated.js
@@ -181,11 +181,6 @@ module.exports = {
                 // Only translate files from node_modules
                 const resolve = resourcePath.match(/(\\|\/)node_modules\1/);
                 const themResource = resourcePath.match(/(\\|\/)theme\1[\s\S]*?\1/) && url.match(/theme\/[\s\S]*?\//);
-                if(resolve) {
-                  console.debug("Inlining node_module resource: ", url);
-                } else if(themResource) {
-                  console.debug("Handling theme resource: ", url);
-                }
                 return resolve || themResource;
               },
               // use theme-loader to also handle any imports in css files
--- a/flow-server/src/main/resources/webpack.generated.js
+++ b/flow-server/src/main/resources/webpack.generated.js
@@ -179,7 +179,14 @@ module.exports = {
             options: {
               url: (url, resourcePath) => {
                 // Only translate files from node_modules
-                return resourcePath.includes('/node_modules/');
+                const resolve = resourcePath.match(/(\\|\/)node_modules\1/);
+                const themResource = resourcePath.match(/(\\|\/)theme\1[\s\S]*?\1/) && url.match(/theme\/[\s\S]*?\//);
+                if(resolve) {
+                  console.debug("Inlining node_module resource: ", url);
+                } else if(themResource) {
+                  console.debug("Handling theme resource: ", url);
+                }
+                return resolve || themResource;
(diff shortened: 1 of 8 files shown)


flow-server/src/main/resources/webpack.generated.js
```diff
@@ -179,7 +179,14 @@ module.exports = {
...
+                const themResource = resourcePath.match(/(\\|\/)theme\1[\s\S]*?\1/) && url.match(/theme\/[\s\S]*?\//);
+                if(resolve) {
+                  console.debug("Inlining node_module resource: ", url);
+                } else if(themResource) {
+                  console.debug("Handling theme resource: ", url);
+                }
+                return resolve || themResource;
               },
               // use theme-loader to also handle any imports in css files
               importLoaders: 1
```
pleku, line 192: If I understood correctly, the comments for the theme-loader below are not valid anymore as it is not aboud `VAADIN/static` but the ` theme/[themeFolder]`

pleku (changes requested):
So
>Only negative is now we loose any folder the resources
were located at.

means that while resources can be referenced with e.g. `theme/mytheme/img/myfunnymeme.png` via Java or `./img/myfunnymeme.png` in css, in the end the resources is not available inside an `img` folder but instead at the root level of `VAADIN/static` ?

caalador (commented):
Got file-loader updated to generate the correct folder structure also.

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -179,7 +179,9 @@ module.exports = {
             options: {
               url: (url, resourcePath) => {
                 // Only translate files from node_modules
-                return resourcePath.includes('/node_modules/');
+                const resolve = resourcePath.match(/(\\|\/)node_modules\1/);
+                const themResource = resourcePath.match(/(\\|\/)theme\1[\s\S]*?\1/) && url.match(/theme\/[\s\S]*?\//);
```
pleku, line 183: typo

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -195,6 +197,23 @@ module.exports = {
...
+      {
+        // File-loader only copies files used as imports in .js files or handled by css-loader
+        test: /\.(png|gif|jpg|jpeg|svg|eot|woff|woff2|ttf)$/,
+        use: [{
+          loader: 'file-loader',
+          options: {
+            outputPath: 'static/',
+            name(resourcePath, resourceQuery) {
+              const urlResource = resourcePath.substring(frontendFolder.length);
+              if(urlResource.match(/(\\|\/)theme\1[\s\S]*?\1/)){
```
pleku, line 209: Should this regex be somewhere so that it is named and shared for both usage (this & css-loader) ?

pleku, line 209: Just to make sure, the `frontendFolder` will never end with `/` or `\` ?

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -195,6 +197,23 @@ module.exports = {
...
+                return /[\s\S]*(\\|\/)theme\1[\s\S]*?\1(.*)/.exec(resourcePath)[2];
```
pleku, line 210: Would it make sense to either use the output from the match above so there is only one regex operation ?

pleku, line 210: Will this code be only executed for the currently active theme's resources as the theme name is not checked here ? I think so but just checking

flow-server/src/main/resources/webpack.generated.js
caalador: fixed

flow-server/src/main/resources/webpack.generated.js
caalador: path.resolve never ends in a path separator. From the documentation `The resulting path is normalized and trailing slashes are removed unless the path is resolved to the root directory.`

caalador: done.

flow-server/src/main/resources/webpack.generated.js
caalador: The match above doesn't collect the file name at all just checks that the theme part is there. 

In the first case we match for the css file not for the resource file so then they need to differ in usage.

caalador: No other themes are pointed to so that webpack would find them so this will only handle the active theme.

pleku (approved):

pleku (approved):

PR #9559, opened 2020-12-03
feat: document level npm css

### LINKED ISSUE #9533: Add theme feature to inject css into the document

caalador:
Now there is a possibility to use the
"documentCss" key in theme.json to
inject npm css into the document rool level
element.

depends on #9558 

Fixes #9533


flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -78,6 +79,15 @@ function generateThemeFile(themeFolder, themeName) {
     globalCssCode.push(`injectGlobalCss(${variable}.toString(), target);\n`);
   });
 
+  let i = 0;
+  if (themeProperties.documentCss) {
+    themeProperties.documentCss.forEach((cssImport) => {
+      const variable = 'module' + i++;
+      imports.push(`import ${variable} from '${cssImport}';\n`);
+      globalCssCode.push(`    injectGlobalCss(${variable}.toString(), target);\n`);
+      globalCssCode.push(`    injectGlobalCss(${variable}.toString(), document);\n`);
```
pleku, line 88: While this is the same as to the global files, maybe the reason _why_ this needs to be injected to both could be documented here

flow-tests/test-themes/frontend/theme/app-theme/theme.json
```diff
@@ -1,4 +1,5 @@
 {
+  "documentCss": ["@fortawesome/fontawesome-free/css/all.css"],
```
pleku, line 2: I thought this _fort_ was a typo first 🤦

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -78,6 +79,15 @@ function generateThemeFile(themeFolder, themeName) {
     globalCssCode.push(`injectGlobalCss(${variable}.toString(), target);\n`);
   });
 
+  let i = 0;
+  if (themeProperties.documentCss) {
+    themeProperties.documentCss.forEach((cssImport) => {
+      const variable = 'module' + i++;
+      imports.push(`import ${variable} from '${cssImport}';\n`);
+      globalCssCode.push(`    injectGlobalCss(${variable}.toString(), target);\n`);
```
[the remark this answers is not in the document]
caalador: The generated theme.js will contain these inside `export const applyTheme = (target) => {`
which is called from the generated-flow-imports.js with
```
import {applyTheme} from 'theme/theme-generated.js';
applyTheme(document);
```
=_= I can't compute today.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
caalador: it should perhaps not need to. will verify once we get the resource PR merged.

caalador: I guess the document level target is for webcomponents that need to have the css inside the shadowroot, but also on the document. 

Let's have this PR on hold as this is apparently a embedded feature as a normal application always injects to document.

pleku:
Is this still waiting for #9543 ? Embedded support for theme went in

caalador:
Yes I'll update this on embedded and change the tests a bit.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
caalador: I'm not certain about the document level css feature. On the Embedded component only adding it to the global scope makes the css not apply into the embedded component, but adding it only to the shadow target works fine. 

Another issue is that the shadow targeted css bleeds out from the shadow to the global space.

For non embedded the issue doesn't exist as they by default target document

flow-tests/test-themes/frontend/theme/app-theme/theme.json
mshabarov: Same for me ;) I asked Google and wondered it's indeed `fort`, not `font`

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
caalador: The documentCss needs to be injected both to the document root and to the shadow root as chrome has a bug in having font-face go into shadowRoot `Issue 336876: @font-face definitions in shadowRoot cannot be used within the shadowRoot`
So it needs to be defined on both levels.

pleku: Can you add the issue reference as a comment to the code ?

caalador: Added.

flow-tests/test-embedding/test-embedding-application-theme/src/test/java/com/vaadin/flow/webcomponent/ApplicationThemeComponentIT.java
```diff
@@ -73,11 +68,34 @@ public void applicationTheme_GlobalCss_isUsedOnlyInEmbeddeComponent() {
 
         Assert.assertNotEquals("Ostrich", body.getCssValue("font-family"));
 
-        getDriver().get(getRootURL() + "/VAADIN/static/themes/embedded-theme/img/bg.jpg");
+        Assert
+            .assertEquals("Embedded style should not match external component",
+                "rgba(0, 0, 255, 1)",
+                $(SpanElement.class).id("overflow").getCssValue("color"));
+                getDriver().get(getRootURL() + "/themes/embedded-theme/img/bg.jpg");
```
pleku, line 75: Broken indent

pleku (approved):

pleku (approved):

flow-tests/test-embedding/test-embedding-application-theme/src/test/java/com/vaadin/flow/webcomponent/ApplicationThemeComponentIT.java
caalador: I'll fix it if the build breaks.

PR #9575, opened 2020-12-04
fix: file-loader name for windows execution

caalador:
Windows file path separator \ should not
be used as it will result in %0B when it should
be /


pleku (approved):
Did our windows build catch this ?

caalador:
Windows builds don't get that far, but they would have caught that.

PR #9577, opened 2020-12-04
feat: add entry for @Deferrable to Vaadin usage statistics

### LINKED ISSUE #9540: add an entry for `@Deferred` endpoints in the Vaadin usage statistics
When an app has one or more @Deferred endpoints, it should add a separate entry about that into the usage statistics data like so:
Screenshot 2020-11-28 at 23 22 20

name: endpoint/@Deferred
version: [same as the Flow version]

joheriks:
Fixes #9540.

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/DeferrableEndpoint.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/DeferrableEndpoint.ts
@@ -1,0 +1,7 @@
+const $wnd = window as any;
+$wnd.Vaadin = $wnd.Vaadin || {};
+$wnd.Vaadin.registrations = $wnd.Vaadin.registrations || [];
+$wnd.Vaadin.registrations.push({
+    is: '@vaadin/@Deferrable',
+    version: /* updated-by-script */ '6.0-SNAPSHOT'
+});
--- a/flow-client/src/main/resources/META-INF/resources/frontend/DeferrableEndpoint.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/DeferrableEndpoint.ts
@@ -5,3 +5,4 @@ $wnd.Vaadin.registrations.push({
     is: '@vaadin/@Deferrable',
     version: /* updated-by-script */ '6.0-SNAPSHOT'
 });
+export {};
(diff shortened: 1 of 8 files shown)


flow-client/src/main/resources/META-INF/resources/frontend/DeferrableEndpoint.ts
```diff
@@ -0,0 +1,7 @@
+const $wnd = window as any;
+$wnd.Vaadin = $wnd.Vaadin || {};
+$wnd.Vaadin.registrations = $wnd.Vaadin.registrations || [];
+$wnd.Vaadin.registrations.push({
+    is: '@vaadin/@Deferrable',
+    version: /* updated-by-script */ '6.0-SNAPSHOT'
```
platosha, line 6: In addition, `scripts/version.js` should be updated to make it actually updated by script.

platosha (changes requested):

flow-client/src/main/resources/META-INF/resources/frontend/DeferrableEndpoint.ts
joheriks: Missed it. Now updates.

haijian-vaadin: Got TS compile erros when trying out the PR.
```
ERROR in [at-loader] ./node_modules/@vaadin/flow-frontend/DeferrableEndpoint.ts:1:7 
    TS2451: Cannot redeclare block-scoped variable '$wnd'.

ERROR in [at-loader] ./target/flow-frontend/DeferrableEndpoint.ts:1:7 
    TS2451: Cannot redeclare block-scoped variable '$wnd'.
```

haijian-vaadin (changes requested):

flow-client/src/main/resources/META-INF/resources/frontend/DeferrableEndpoint.ts
joheriks: Good catch, I had somehow dropped the empty export from the commit and it was working randomly when testing it. Now fixed.

haijian-vaadin (approved):

platosha (approved):

Issue #9574, opened 2020-12-04
Add 'refreshAll' method to the DataView API


mshabarov:
`DataView` API does not contain the `refreshAll` method, . It was supposed to be used internally in list data view as soon as the item/items are added to a backed collection, and there would be no need for developers to call it explicitly. For a lazy loading, the main considered use case was to only fetch the data, but it's not the case in a real life and application logic may require to add/remove the items to the backed directly and then refresh the component to display up-to-date state. `HasLazyDataView::setItems` can be used to refresh the data, but it completely removes all including the cached items.

So, these are at least two drivers to add `refreshAll` method:
1. Refresh lazy loading component after changes in backend:
```java
save.addClickListener(e -> {
    if (this.entity == null) {
        this.entity = new Entity();
    }
    binder.writeBean(this.entity);
    // Add or update the entity
    entityService.update(this.entity);
    grid.getDataProvider().refreshAll();
});
```
2. Refresh in-memory component changed items in one tweak:
```java
Item first = new Item(1L, "First");
Item second = new Item(2L, "Second");
Item third = new Item(3L, "Third");

RadioButtonGroup<Item> radioButtonGroup = new RadioButtonGroup<>();
RadioButtonGroupListDataView<Item> listDataView = radioButtonGroup.setItems(first, second, third);
...
updateButton.addClickListener(e -> {
   first.setName("Changed");
   listDataView.refreshItem(first);

   third.setName("Changed");
   listDataView.refreshItem(third);

   // Good to have one call to listDataView.refreshAll() instead
});
```
Solution is to add this method into `DataView` interface and delegate it to `DataProvider::refreshAll` in `AbstractDataView`.

PR #9580, opened 2020-12-07
fix: Add missing refreshAll method to data view

mshabarov:
Adds the missed refresh all method to data view API.

Fixes: https://github.com/vaadin/flow/issues/9574


pleku (approved):
The javadocs could have linked to `refreshItem` and vice-versa but lets just get this in.

PR #9583, opened 2020-12-07
fix: encode attribute value during transfer from embedded web app

denis-anisimov:


flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -378,16 +381,28 @@ private void transferAttribute(Writer writer, String elementRef,
...
+                    + encode(suffix.substring(queryIndex + 1));
+        }
+        return URI.create(basePath + suffix).toString();
+    }
 
+    private String encode(String str) throws UnsupportedEncodingException {
+        if (str == null) {
+            return null;
         }
+        return URLEncoder.encode(str, StandardCharsets.UTF_8.name());
```
fluorumlabs, line 405: Hm. `URLEncoder.encode` is for encoding parameters for using in `application/x-www-form-urlencoded` data, it changes forward slash (perfectly valid in path part of URL) to `%2f`.

flow-server/src/test/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandlerTest.java
```diff
@@ -102,6 +103,27 @@ public void writeBootstrapPage_skipMetaAndStyleHeaderElements()
...
+    public void writeBootstrapPage_escapeAttributeValue() throws IOException {
+        WebComponentBootstrapHandler handler = new WebComponentBootstrapHandler();
+
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+
+        Element head = new Document("").normalise().head();
+
+        Element script = head.ownerDocument().createElement("script");
+        head.appendChild(script);
+        script.attr("src", "foo'bar %27?baz%22");
```
fluorumlabs, line 116: Would be nice to see the test with a real URL (with protocol, path parameters etc.)

fluorumlabs (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
denis-anisimov: The forward slash is not a problem : encoded slash is recognized correctly in the URI.
Encoded `?` is an issue.

But if not `URLEncoder.encode`  then which way do you suggest to escape/encode here ?

fluorumlabs: One particular problem is whitespace character: in path it should be encoded as `%20` whereas `URLEncoder.encode` encodes it as `+`. The real solution is to either escape single quote to avoid injection, or, possibly, use double quotes in JS (double quote is not allowed in URL AFAIR)

denis-anisimov: Why it is a problem ?

Single quote is not always a single quote.
The problem is : you have to detect whether the symbol `'` is a single quote or not.
E.g. `\'` is not a single quote.
And `%27` is also a single quote.

Please provide a code which escape single quote.
>or, possibly, use double quotes in JS (double quote is not allowed in URL

double quote is not allowed. But I don't understand the sentence.

denis-anisimov: >or, possibly, use double quotes in JS (double quote is not allowed in URL

It's really hard to decrypt what you mean sometimes.

Do you suggest to use double quotes in the code which creates JS content inside `transferAttribute`?

Then yes, looks like it's a good suggestion.

flow-server/src/test/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandlerTest.java
denis-anisimov: Double quotes are use so there is no encoding done anymore.

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -390,19 +388,18 @@ protected String modifyPath(String basePath, String path)
...
-            suffix = encode(suffix.substring(0, queryIndex)) + '?'
-                    + encode(suffix.substring(queryIndex + 1));
-        }
-        return URI.create(basePath + suffix).toString();
+        return URI.create(checkURL(basePath + suffix)).toString();
     }
 
-    private String encode(String str) throws UnsupportedEncodingException {
-        if (str == null) {
+    private String checkURL(String url) {
```
fluorumlabs, line 394: Since this method doesn't returns argument as is, it should probably return `boolean` instead and called `isUrlUnsafe` to align with `com.vaadin.flow.server.HandlerHelper#isPathUnsafe`. The `IllegalArgumentException` should be thrown in `modifyPath` then.

fluorumlabs (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandlerTest.java
```diff
@@ -113,15 +114,31 @@ public void writeBootstrapPage_escapeAttributeValue() throws IOException {
...
         head.appendChild(script);
-        script.attr("src", "foo'bar %27?baz%22");
+        script.attr("src", "foo'bar%20%27?baz%22");
 
         VaadinResponse response = getMockResponse(stream);
         handler.writeBootstrapPage("", response, head, "");
 
         String resultingScript = stream.toString(StandardCharsets.UTF_8.name());
+        MatcherAssert.assertThat(resultingScript, CoreMatchers
+                .not(CoreMatchers.containsString("foo'bar %27?baz%22")));
```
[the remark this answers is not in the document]
denis-anisimov: Don't understand this .
Please decrypt.

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
denis-anisimov: This is an internal `private` method and it's name and how it's called is a  matter of taste .
The method returns the provided argument as is without any change and it throws the exception if something wrong with the argument.
It works quite similar to the `Objects::requireNonNull`.

flow-server/src/test/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandlerTest.java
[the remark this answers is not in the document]
fluorumlabs: IMO the assertion should be `MatchAssert.assertThat(resultScript, CoreMatchers.containsString(...))`. If you disagree, could you explain what exactly this test is testing?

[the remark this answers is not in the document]
denis-anisimov: Now I've got it: `CoreMatchers.not` was a mistake in copy paste.
The intention was to have only `CoreMatchers.containsString`.
Thanks,

fluorumlabs (approved):

mshabarov:
Cherry picked to `2.5`

PR #9584, opened 2020-12-07
fix: reorder manifest and compression plugins

### LINKED ISSUE #9457: Offline service worker is broken in production mode by webpack compression plugin working first
In production mode, generated webpack config has `CompressionPlugin` added to plugins. This produces two copies of bundle files in the output directory, one raw `.js` and one compressed with `.js.gz` filename extension.

While `InjectManifest` manifest is always adding manifest to the raw `.js` output file, the compressed `.js.gz` output could be missing the injected manifest if compression plugin works first.

Currently in the generated webpack config, `CompressionPlugin` is defined first, well before `InjectManifest`. If compressed assets are then served to the browser, this results in a broken service worker, as the manifest is entirely missing there.

Solution idea: make sure `InjectManifest` is defined before `CompressionPlugin` in the webpack config.

joheriks:
Also run offline test modules (`test-pwa` and `test-ccdm-flow-navigation`) in production mode.

Fixes #9457.

### DIFF
--- a/flow-server/src/main/resources/webpack.generated.js
+++ b/flow-server/src/main/resources/webpack.generated.js
@@ -256,10 +256,14 @@ module.exports = {
     maxAssetSize: 2097152 // 2MB
   },
   plugins: [
-    // Generate compressed bundles when not devMode
-    !devMode && new CompressionPlugin(),
+    // Service worker for offline
+    pwaEnabled && serviceWorkerPlugin,
+
     // Generate manifest.json file
     new ManifestPlugin(),
+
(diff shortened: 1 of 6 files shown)


Haprog (approved):

Haprog:
There seems to be a compilation error, but otherwise looks good.

joheriks:
> There seems to be a compilation error, but otherwise looks good.

Thanks for the heads-up, removed use of `IOUtils` in the IT.

platosha (approved):

PR #9588, opened 2020-12-07
feat: Embedded component appTheme support

### LINKED ISSUE #8564: Use application theme for exported web components
See the original issue :#8543.
The current behavior is by design but I don't even know how to style the embedded component since it's inside shadow root of the generated wrapper.
I've made already a [ticket in the docs about styling of embedded components](https://github.com/vaadin/flow-and-components-documentation/issues/1300).

May be we should provide API for `WebComponentExporter` which allows to include styles which works for its content and allow to style the embedded component in its shadow root.

The only way to do styling which I see at the moment is: make the embedded component as a web component ( e.g. a Polymer element) and "inline styling" there so that it's local and doesn't leak globally.
But it's silly because if I want to use only standard HTML components (like `div`) I have to make my own wrapper just to style it.
It's reasonable to add styling to already existing wrapper which we generate.

caalador:
Add support for application theme with
embedded components.

Fixes #8564


flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponentUI.java
```diff
@@ -286,7 +288,8 @@ private void assignThemeVariant() {
         WebComponentConfigurationRegistry registry = getConfigurationRegistry();
         Optional<Theme> theme = registry
                 .getEmbeddedApplicationAnnotation(Theme.class);
-        if (!theme.isPresent()) {
+        if (!theme.isPresent() || theme.get().themeClass()
+            .equals(AbstractTheme.class)) {
```
pleku, line 292: So if one has `@Theme(value="my-theme", variant=Lumo.DARK)` in their exported web component it will not apply the variant ?
Or does it fail already before this place ?

flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
```diff
@@ -158,4 +158,10 @@
      * component UI imports them as dependencies.
      */
     public static final String COMPILED_WEB_COMPONENTS_PATH = "compiled.web.components.path";
+
+    /**
+     * Configuration for determining if embeddable web components should
+     * be generated. By default they are generated.
+     */
+    public static final String GENERATE_WEB_COMPONENTS = "generateEmbeddableWebComponents";
```
pleku, line 166: Not sure why do we need to expose this even now ?

flow-server/src/main/resources/com/vaadin/flow/server/webcomponent/webcomponent-script-template.js
```diff
@@ -1,4 +1,4 @@
-class _TagCamel_ extends HTMLElement {
+_ThemeImport_class _TagCamel_ extends HTMLElement {
```
pleku, line 1: These look odd since there is no whitespace after the `_ThemeImport_` like with `_TagCamel_`, looks like a mistake but it still works ?

flow-server/src/main/resources/com/vaadin/flow/server/webcomponent/webcomponent-script-template.js
```diff
@@ -12,7 +12,7 @@ class _TagCamel_ extends HTMLElement {
         display: inline-block;
       }
     `;
-    shadow.appendChild(style);
+    _ApplyTheme_shadow.appendChild(style);
```
pleku, line 15: Same here, the missing whitespace looks like a mishap even though it doesn't fail due to the `;` ending

flow-server/src/main/resources/plugins/application-theme-plugin/package.json
```diff
@@ -6,7 +6,7 @@
   ],
   "repository": "vaadin/flow",
   "name": "@vaadin/application-theme-plugin",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
pleku, line 9: Isn't exported web component support a new feature ? :trollface:

pleku (changes requested):

flow-server/src/main/resources/plugins/application-theme-plugin/package.json
caalador: Not from the plugin point of view as the addition is only a globalCssCode.push

flow-server/src/main/resources/com/vaadin/flow/server/webcomponent/webcomponent-script-template.js
caalador: The reason is that it is replaced with `something ` + new line or with nothing
It's this way by design, but we can update tests to check for `\n\nStart String` also the matching target is _name_ which doesn't actually need spaces before or after, it's more on what replaces the placeholder.

flow-server/src/main/resources/com/vaadin/flow/server/webcomponent/webcomponent-script-template.js
caalador: this is also replaced by `something`+newLine or ""

flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponentUI.java
caalador: This will not do anything if you have a themeClass in theme. You can have `my-theme` and theme variant.

flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
caalador: I don't understand why this was not configurable from the beginning, but I'll revert this.

flow-server/src/main/resources/com/vaadin/flow/server/webcomponent/webcomponent-script-template.js
caalador: see `WebComponentGenerator::getReplacementsMap` l162

pleku (approved):

PR #9592, opened 2020-12-07
feat: allow endpoints to work on non-public classes

### LINKED ISSUE #8168: @Endpoint should work on non-public classes
When calling a backend endpoint from TypeScript, I want to be able to call it regardless of whether or not the `@Endpoint`-annotated Java class is public, so that my DX is more consistent with that of Spring's `RestController`s.

The `getRequest()` method in the example below should be callable from TypeScript.
```java
@Endpoint
@AnonymousAllowed
class RequestEndpoint {
    public Request getRequest() {
        return new Request();
    }
}
```

With 15.0.5 calling such method results in a 500 HTTP error and an unhelpful error message:
image

The server logs are not helpful either:
```
2020-04-27 11:43:42.550 ERROR 25421 --- [nio-8080-exec-3] c.v.f.s.connect.VaadinConnectController  : Endpoint 'RequestEndpoint' method 'getRequest' access failure

java.lang.IllegalAccessException: Class com.vaadin.flow.server.connect.VaadinConnectController can not access a member of class com.example.demo.backend.RequestEndpoint with modifiers "public"
        at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102) ~[na:1.8.0_232]
        at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296) ~[na:1.8.0_232]
        at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288) ~[na:1.8.0_232]
        at java.lang.reflect.Method.invoke(Method.java:491) ~[na:1.8.0_232]
        at com.vaadin.flow.server.connect.VaadinConnectController.invokeVaadinEndpointMethod(VaadinConnectController.java:330) [flow-server-3.0.6.jar:3.0.6]
```

NOTE: See how this is handled in the [RestController implementation](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/method/support/InvocableHandlerMethod.java#L188).

tdq:
Fix sets accessible all public methods of endpoint which allows to call them even if class is not public.
Checked with Oracle documentation that fix supports java 12 (there was a doubt that it's not)
https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/lang/reflect/AccessibleObject.html

Fixes #8168


flow-server/src/main/java/com/vaadin/flow/server/connect/VaadinConnectController.java
```diff
@@ -221,8 +222,11 @@ void validateEndpointBean(EndpointNameChecker endpointNameChecker,
                             endpointName, validationError));
         }
 
+        Method[] endpointPublicMethods = beanType.getMethods();
+        AccessibleObject.setAccessible(endpointPublicMethods, true);
```
joheriks, line 226: This does set the accessible override flag on all public methods in the class, including those inherited from `Object` (`wait`, `toString`, ...) even though these are from being exposed as endpoints later on. Seems it would be safer to flag only the methods that can actually be called as endpoints.

joheriks (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/connect/VaadinConnectController.java
tdq: setAccessible is not changing method itself, it's changing only instance of that Method (it's not singleton). The rest of security checks is done in com.vaadin.flow.server.connect.auth.VaadinConnectAccessChecker which is not changed in this PR.

joheriks: Right. I missed that `getMethods` returns copies so the override has only local impact. This looks fine then.

joheriks: Looks good to me after conflict resolution then.

joheriks (dismissed):

haijian-vaadin (dismissed):
Also looks good to me, ready to be merged once the conflicts are resolved.

haijian-vaadin (approved):

PR #9593, opened 2020-12-07
fix: loading indicator visible and isActive true while endpoint fetch

joheriks:


flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts
```diff
@@ -385,16 +388,21 @@ export class Flow {
...
-      if (action) {
-        loading.removeAttribute('style');
-        timeout2nd = setTimeout(() => loading.classList.add('second'), 1500);
-        timeout3rd = setTimeout(() => loading.classList.add('third'), 5000);
-      } else {
-        loading.setAttribute('style', 'none');
+      const show = action && this.isActiveCount === 0;
+      const hide = !action && this.isActiveCount === 1;
+      this.isActiveCount = Math.max(0, this.isActiveCount + (action ? 1 : -1));
+      if (show || hide) {
```
haijian-vaadin, line 394: The code here is not easy to understand.
`action && this.isActiveCount === 0;` 
`!action && this.isActiveCount === 1`
`Math.max(0, this.isActiveCount + (action ? 1 : -1));`
could these expression be extracted  as more meaningful functions?

`if (show || hide)`
Is it possible that it could be neither `show` nor `hide`?

haijian-vaadin (changes requested):

flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts
joheriks: I clarified the code a bit and added comments. `show` and `hide` would both be false when there are multiple requests being handled at the same time. I added a unit test.

haijian-vaadin (approved):

PR #9600, opened 2020-12-08
fix: Vaadin.Flow.clients.TypeScript.isActive for endpoint requests

joheriks:
Allows TestBench tests to assume that all initial endpoint requests have
completed before the test case is executed.

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/Connect.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/Connect.ts
@@ -420,23 +420,17 @@ export class ConnectClient {
     // chain item for our convenience. Always having an ending of the chain
     // this way makes the folding down below more concise.
     const fetchNext: MiddlewareNext =
-      async(context: MiddlewareContext): Promise<Response> => {
-        if ($wnd.Vaadin.connectionState) {
-          $wnd.Vaadin.connectionState.loadingStarted();
-        }
-       return fetch(context.request)
-            .then(response => {
-              if ($wnd.Vaadin.connectionState) {
-                $wnd.Vaadin.connectionState.loadingSucceeded();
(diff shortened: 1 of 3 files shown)


flow-client/src/main/resources/META-INF/resources/frontend/Connect.ts
```diff
@@ -420,23 +420,17 @@ export class ConnectClient {
...
+        $wnd.Vaadin.connectionState.loadingStarted();
+        return fetch(context.request)
+          .then(response => {
+            $wnd.Vaadin.connectionState.loadingSucceeded();
+            return response;
+          })
+          .catch(error => {
+            $wnd.Vaadin.connectionState.state = ConnectionState.CONNECTION_LOST;
+            return Promise.reject(error);
+          });
```
joheriks, line 433: We should be able to assume that `$wnd.Vaadin.connectionState` is defined here., the `if` guards were there previously only to make a test (`ConnectTests: should set header for preventing CSRF using Flow csrfToken`) pass. This is now corrected in the test instead.

flow-client/src/test/frontend/ConnectTests.ts
```diff
@@ -224,7 +225,10 @@ describe('ConnectClient', () => {
       // @ts-ignore
       const OriginalVaadin = window.Vaadin;
       // @ts-ignore
-      window.Vaadin = {TypeScript: {csrfToken: 'foo'}};
+      window.Vaadin = {
+        TypeScript: {csrfToken: 'foo'},
+        connectionState: (window as any).Vaadin.connectionState
+      };
```
joheriks, line 231: This is the test correction mentioned in the previous comment.

haijian-vaadin (approved):

PR #9604, opened 2020-12-08
fix(TypeScript): binder should allow typing decimal numbers

### LINKED ISSUE #9521: Binding to a double using a TS form does not work
1. Go to https://start.vaadin.com/?preset=fusion
2. Change view type to "master-detail" and download
3. Change the type of `phone` in `Person.java` to `double`
4. Remove the data generation code from `DateGenerator`
5. Run the project and try to set the value of phone to `12.2`.

### Expected behavior
You can type "12.2" in the field

### Actual behavior
When you type "12." the value changes to "12" and it is impossible to input a decimal number.

### Versions:
    - Vaadin / Flow version: 18rc1

### LINKED ISSUE #9522: Binding a field to a NumberModel should not change an invalid input to NaN
1. Go to https://start.vaadin.com/?preset=fusion
2. Change view type to "master-detail" and download
3. Change the type of `phone` in `Person.java` to `double`
4. Remove the data generation code from `DateGenerator`
5. Run the project and type e.g. "," in the phone field

### Expected behavior
The field value becomes "," and it turns red to indicate it is invalid

### Actual behavior
The field value becomes "NaN" and is shown as being valid

### Versions:
    - Vaadin / Flow version: 18rc1

platosha:
Fixes #9521
Fixes #9522

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/form/Validators.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/form/Validators.ts
@@ -46,7 +46,7 @@ abstract class AbstractValidator<T> implements Validator<T> {
 
 export class Required<T> extends AbstractValidator<T> {
   impliesRequired = true;
-  validate(value: T){
+  validate(value: T) {
     if (typeof value === 'string' || Array.isArray(value)) {
       return value.length > 0;
     } else if (typeof value === 'number') {
@@ -64,7 +64,22 @@ function _value(attrs: ValueNumberAttributes | number | string) {
   return typeof attrs === 'object' ? attrs.value : attrs;
 }
 
-abstract class ValueNumberValidator<T> extends AbstractValidator<T> {
+abstract class NumberValidator<T> extends AbstractValidator<T> {
+  constructor(attrs?: ValidatorAttributes) {
+    super(attrs);
+  }
+  validate(value: T) {
+    return isNumeric(String(value));
+  }
+}
+
+export class IsNumber extends NumberValidator<number> {
+  constructor(attrs?: ValidatorAttributes) {
+    super({message: 'must be a number', ...attrs});
+  }
+}
+
+abstract class ValueNumberValidator<T> extends NumberValidator<T> {
   value: number;
   constructor(attrs: ValueNumberAttributes | number | string) {
     super(_asValidatorAttributes(attrs));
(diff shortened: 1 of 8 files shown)


flow-client/src/main/resources/META-INF/resources/frontend/form/Validators.ts
```diff
@@ -56,15 +56,30 @@ export class Required<T> extends AbstractValidator<T> {
   }
 }
 
-function _asValidatorAttributes(attrs: ValueNumberAttributes | number | string | PatternAttributes | RegExp) {
+function _asValidatorAttributes(attrs: ValueNumberAttributes | number | string | PatternAttributes | string | RegExp) {
```
haijian-vaadin, line 59: redundant `string`?

haijian-vaadin (changes requested):
Also, could we have a test for the fix?

platosha:
@haijian-vaadin added tests, there is one ensuring that the value is not reset from the binder while typing.

haijian-vaadin:
It looks good now, just found another issue. If I type `12,5` or `12,kjsdf` into the field, I don't get a validation error, the data is saved as 12. Could this be fixed in this PR, so should it be in another ticket?

platosha:
@haijian-vaadin thanks for pointing out, fixed here also

haijian-vaadin:
There are some lint errors for Models.ts file

platosha:
@haijian-vaadin oh, I always forget those, fixed 😅

flow-client/src/test/frontend/form/FieldTests.ts
```diff
@@ -311,6 +394,76 @@ suite("form/Field", () => {
...
+          // Invalid separator
+          [',', NaN],
+          [',2', NaN],
+          ['1,', NaN],
+          ['1,2', NaN]
+        ];
+
+        for (const [inputValue, expectedNumber] of cases) {
+          for (const eventName of ['input', 'change']) {
+            debugger;
```
haijian-vaadin, line 440: remove debugger?

flow-client/src/test/frontend/form/ValidatorsTests.ts
```diff
@@ -120,6 +121,18 @@ suite("form/Validators", () => {
     assert.isFalse(validator.validate(true));
   });
 
+  test('IsNumber', () => {
+    const validator = new IsNumber();
+    assert.isNotTrue(validator.impliesRequired);
+    assert.isTrue(validator.validate(0));
+    assert.isTrue(validator.validate(1));
+    assert.isTrue(validator.validate(1.2));
```
haijian-vaadin, line 129: Maybe we could add validations for `1,2`, `1,abc`, `abc` etc?

flow-client/src/test/frontend/form/FieldTests.ts
haijian-vaadin, line 440: remove debugger?

haijian-vaadin (changes requested):
It works! Good job! Found some minors things that could be improved.

flow-client/src/test/frontend/form/ValidatorsTests.ts
platosha: This validator cannot check those cases — it only works on the `number` data type, which is the only type `NumberModel` works with.

flow-client/src/test/frontend/form/FieldTests.ts
platosha: done

haijian-vaadin (approved):

PR #9609, opened 2020-12-08
refactor: simplify ResourceProvider

### LINKED ISSUE #9601: Simplify ResourceProvider
`ResourceProvider` is an abstraction which allow to access _WEB APP_ resources in various containers.
There are a number of methods which are quite cumbersome because of their first argument:

`getApplicationResource(Class<?> contextClass, String path);` 
`getApplicationResources(Class<?> contextClass, String path)` 
`getApplicationResources(Object context, String path)``
`getApplicationResource(Object context, String path);`

In the current usage `context` in the last pair of the methods is `VaadinContext`.
But `VaadinContext` is available when the Lookup is created (which produces the `ResourceProvider` instance).
So `Lookup` should be implemented being based on `VaadinContext` (at the moment it's stateless especially in OSGi case).
And then methods with `Object context` can be removed.

The second step would be to remove the first argument in the remaining methods: the context class is required to detect the application bundle but this bundle may be detected via `ServletContext::getClassLoader`.
The only problem is : `ServletContext::getClassLoader` returns something wrong being called inside `ServletContextListener::contextInitialized` (it returns `flow-osgi` bundle classloader, may be depending on the order of Servlet:: init and context init calls).
If `ServletContextListener::contextInitialized` is called inside `Servlet::init` then the classloader is the correct classloader: it's an app bundle classloader.
This is the second step.

### LINKED ISSUE #9605: Subticket: get rid of ResourceProvider:: getApplicationResource(s) methods with Object as the first arg
See #9601.

Two methods 
`getApplicationResources(Object context, String path)`
and
`getApplicationResource(Object context, String path);`

should be removed since the `Object context`  should be already available for the `Lookup` which returns the `ResourceProvider` instance (and for the resulting `ResourceProvider` as a consequence).

denis-anisimov:
fixes #9601

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapUtils.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapUtils.java
@@ -257,8 +257,12 @@ class BootstrapUtils {
         VaadinService service = request.getService();
         ResourceProvider resourceProvider = service.getContext()
                 .getAttribute(Lookup.class).lookup(ResourceProvider.class);
-        URL appResource = resourceProvider.getApplicationResource(service,
-                file);
+        Class<?> clazz = null;
+        if (service instanceof VaadinServletService) {
+            clazz = ((VaadinServletService) service).getServlet().getClass();
+        }
+        URL appResource = clazz == null ? null
+                : resourceProvider.getApplicationResource(clazz, file);
 
         InputStream stream = null;
         try {
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapUtils.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapUtils.java
@@ -257,12 +257,7 @@ class BootstrapUtils {
         VaadinService service = request.getService();
         ResourceProvider resourceProvider = service.getContext()
                 .getAttribute(Lookup.class).lookup(ResourceProvider.class);
-        Class<?> clazz = null;
-        if (service instanceof VaadinServletService) {
-            clazz = ((VaadinServletService) service).getServlet().getClass();
-        }
-        URL appResource = clazz == null ? null
-                : resourceProvider.getApplicationResource(clazz, file);
+        URL appResource = resourceProvider.getApplicationResource(file);
 
         InputStream stream = null;
         try {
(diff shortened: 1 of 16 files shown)


flow-server/src/main/java/com/vaadin/flow/server/BootstrapUtils.java
```diff
@@ -257,8 +257,12 @@ private static InputStream getInlineResourceStream(VaadinRequest request,
         VaadinService service = request.getService();
         ResourceProvider resourceProvider = service.getContext()
                 .getAttribute(Lookup.class).lookup(ResourceProvider.class);
-        URL appResource = resourceProvider.getApplicationResource(service,
-                file);
+        Class<?> clazz = null;
+        if (service instanceof VaadinServletService) {
```
denis-anisimov, line 261: temp changes: this will be removed completely in the second step: resourceProvider.getApplicationResource won't need the first arg anymore.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -608,7 +609,13 @@ private static InputStream getStatsFromClassPath(VaadinService service) {
                 .replaceFirst("^/", "");
         ResourceProvider resourceProvider = service.getContext()
                 .getAttribute(Lookup.class).lookup(ResourceProvider.class);
-        URL statsUrl = resourceProvider.getApplicationResource(service, stats);
+        Class<?> appClass = null;
```
denis-anisimov, line 612: temp changes: this will be removed completely in the second step: resourceProvider.getApplicationResource won't need the first arg anymore.

caalador (approved):

caalador:
JavaDoc exception
```
/opt/agent/work/a69251da2b2bb9e9/flow-server/src/main/java/com/vaadin/flow/di/ResourceProvider.java:58: error: @param name not found
     * @param contextClass
```

caalador (approved):

caalador:
Seems the mock is still wrong as all failing tests fail with:
```
java.lang.ArrayIndexOutOfBoundsException: 1
	at org.mockito.internal.invocation.InvocationImpl.getArgumentAt(InvocationImpl.java:66)
	at com.vaadin.flow.server.MockServletServiceSessionSetup.lambda$new$2(MockServletServiceSessionSetup.java:317)
```

denis-anisimov:
> MockServletServiceSessionSetup

yep, thanks.

caalador (approved):

PR #9616, opened 2020-12-09
Fix: Generated theme.js is now named theme.generated.js

### LINKED ISSUE #9595: Rename the application theme generated file to theme-name.generated.js
The generated theme file should also have a header informing that this is a generated file.

Note that the application plugin resolves the theme name from the theme-generates.js file import statement and this should be updated at to correctly parse the name in `application-theme-plugin.js : nameRegex`

caalador:
my-theme.js is now generated as my-theme.generated.js

Fixes #9595

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,5 +54,5 @@ yarn.lock
 
 flow-client/src/main/resources/META-INF/resources/frontend/FlowClient.js
 flow-tests/**/types.d.ts
-flow-tests/test-themes/frontend/theme/app-theme/app-theme.js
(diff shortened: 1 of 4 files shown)


taefi (commented):
Is it reasonable to have a test that verifies a `foobar.generated.js` is created for a theme named "foobar"?

caalador:
The problem is that the file is generated by webpack and we don't have a js framework setup that would  be usable to test webpack or the plugins so it's testes in the 2 IT test modules as if it would go wrong they would fail.

taefi (approved):

PR #9617, opened 2020-12-09
Feat: Check that npm resources are available

### LINKED ISSUE #9615: Add checks that npm module is installed
When adding either `importCSS` or npm `assets` the application-theme-plugin should check that the resource is available and throw a descriptive Error if the resource is not installed or found that gives the user two alternatives to install the package with `@NpmPackage` (preferred) or manual `npm/pnpm i`.

caalador:
Check that we have named npm resources
available or throw an exception informing
the user to intall it using a NpmPackage
annotation or manually with (p)npm i

Fixes #9615

### DIFF
--- a/flow-server/src/main/resources/plugins/application-theme-plugin/theme-copy.js
+++ b/flow-server/src/main/resources/plugins/application-theme-plugin/theme-copy.js
@@ -61,15 +61,18 @@ function copyStaticAssets(themeName, themeProperties, projectStaticAssetsOutputF
   fs.mkdirSync(projectStaticAssetsOutputFolder, {
     recursive: true
   });
+  const missingModules = checkModules(Object.keys(assets));
+  if (missingModules.length > 0) {
+    throw Error("Missing npm modules '" + missingModules.join("', '")
+      + "' for assets marked in 'theme.json'.\n" +
+      "Install package(s) by adding a @NpmPackage annotation or install it using 'npm/pnpm i'");
+  }
   Object.keys(assets).forEach((module) => {
 
     const copyRules = assets[module];
     Object.keys(copyRules).forEach((copyRule) => {
       const nodeSources = path.resolve('node_modules/', module, copyRule);
-      if(!fs.existsSync(path.resolve('node_modules/', module))) {
-        throw Error("Missing '" + module + "'. Install package by adding a @NpmPackage annotation or install it using 'npm/pnpm i'");
-      }
-      const files = glob.sync(nodeSources, { nodir: true });
+      const files = glob.sync(nodeSources, {nodir: true});
       const targetFolder = path.resolve(projectStaticAssetsOutputFolder, "theme", themeName, copyRules[copyRule]);
 
       fs.mkdirSync(targetFolder, {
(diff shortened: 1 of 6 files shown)


flow-server/src/main/resources/plugins/application-theme-plugin/theme-copy.js
```diff
@@ -66,7 +66,10 @@ function copyStaticAssets(themeName, themeProperties, projectStaticAssetsOutputF
     const copyRules = assets[module];
     Object.keys(copyRules).forEach((copyRule) => {
       const nodeSources = path.resolve('node_modules/', module, copyRule);
-      const files = glob.sync(nodeSources, {nodir: true});
+      if(!fs.existsSync(path.resolve('node_modules/', module))) {
+        throw Error("Missing '" + module + "'. Install package by adding a @NpmPackage annotation or install it using 'npm/pnpm i'");
```
pleku, line 70: Would it be better if we would check for all the missing packages in a loop and output all of them at once ?
So if one is missing 5 of those (for example due to not reading our docs on how to package a reusable theme), they could know all of the missing things after one run instead of 5 ?

pleku, line 70: Should this clarify that the module is needed as an CSS asset in `theme.json` so they don't have to wonder where does this module come from ?

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -83,6 +84,11 @@ function generateThemeFile(themeFolder, themeName, themeProperties) {
   let i = 0;
   if (themeProperties.importCss) {
     themeProperties.importCss.forEach((cssPath) => {
+      const nodeSources = path.resolve('node_modules/', cssPath);
+      if(!fs.existsSync(nodeSources)) {
+        throw Error("Missing resource '" + cssPath + "'. Install package by adding a @NpmPackage annotation or install it using 'npm/pnpm i'");
```
pleku, line 89: Same comment here, should we first collect all of the missing packages and output them once ?

pleku, line 89: Same, should this mention what this resource is and where is it coming from, like "Missing imported css file <path> imported by theme.json. Install ..."

pleku (changes requested):

flow-server/src/main/resources/plugins/application-theme-plugin/theme-copy.js
caalador: done.

caalador: No the assets are of style:
```
  "assets": {
    "@fortawesome/fontawesome-free": {
      "svgs/regular/**": "fortawesome/icons"
    }
  }
```
so it's just the module that is missing. These are added when you want to use it from java.
But added clarification.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
caalador: done

caalador: added

pleku (approved):

PR #9618, opened 2020-12-09
chore: add internal classes to access classloader and init context

denis-anisimov:
related to #9601 : needed to restore OSGi resource provider

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
@@ -30,7 +30,9 @@ import java.util.Properties;
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.di.Lookup;
 import com.vaadin.flow.function.DeploymentConfiguration;
+import com.vaadin.flow.internal.ApplicationClassLoaderAccess;
 import com.vaadin.flow.internal.CurrentInstance;
+import com.vaadin.flow.internal.VaadinContextInitializer;
 import com.vaadin.flow.server.HandlerHelper.RequestType;
 import com.vaadin.flow.shared.JsonConstants;
 
@@ -84,8 +86,11 @@ public class VaadinServlet extends HttpServlet {
              * its "init" method is called from the {@code
              * ServletContextListener} with the same ServletConfig instance.
              */
+            VaadinServletContext vaadinServletContext = null;
             if (getServletConfig() == null) {
                 super.init(servletConfig);
+
+                vaadinServletContext = initializeContext();
             }
 
             if (getServletConfig() != servletConfig) {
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -513,4 +520,22 @@ public void destroy() {
...
 
+    private VaadinServletContext initializeContext() {
+        ServletContext servletContext = getServletConfig().getServletContext();
+        VaadinServletContext vaadinServletContext = new VaadinServletContext(
+                servletContext);
+        // ensure the web application classloader is available via context
+        ApplicationClassLoaderAccess access = () -> servletContext
+                .getClassLoader();
+        vaadinServletContext.getAttribute(ApplicationClassLoaderAccess.class,
+                () -> access);
```
caalador, line 531: Why do we have a runnable running a runnable. can't we just have `access` as the default instead of  () -> access` else drop access and put it directly in the getAttribute default value.

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
denis-anisimov: The first "runnable" is a factory which produces a value for the attribute.
The second "runnable" is the access which returns a classloader.

You suggestion means  the `ClassLoader`  instance will be set as the attribute  "ClassLoader.class" value.

There are two reasons why I don't want to use `ClassLoader` instance directly but instead wrap it:
- if I recall correctly you had problems with accessing to `ServletContext::getClassLoader` . Sometimes it may throw an exception if it's called "too early" (I have no idea what "too early" means). `ServletContext::getClassLoader` has bad API since there is a public method which you may call at any point but in fact it may not be called at any point. So I'm trying to avoid calling it right away. Instead "access"/factory is used to call it.
- I don't want to expose `ClassLoader`  as the attribute value for "ClassLoader.class"  attribute.
`ClassLoader` is the standard JVM class. So it's possible that someone (an app  developer) may use `ClassLoader.class`  attribute for his own purposes. This will break internal assumptions.

Ideally `ApplicationClassLoaderAccess` would be a private class so no-one is able to use `ApplicationClassLoaderAccess.class`  attribute to override its value. But that will require the servlet class in `osgi`  module (which can be done but then `OSGiXXXXServlet` should be used instead of `VaadinServlet` and I'm trying to avoid that).

caalador: A conversation is not resolved by you answering it. It should be up to the one making the comment to resolve it as agreed on slack a while ago.

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
denis-anisimov: Do you mean I should not click "resolved" ?
Does it mean that you should resolve it ?
But then why the button is shown to me ?
Or should I use the button only when I made some change ?

denis-anisimov: But you may click technically "unresolve conversation" if you disagree .....

I'm trying to make all the conversations resolved if I answer it so if I don't go into the same conversation if I forget that it has a complete answer.
And I expect that it will be unresolved if it's not resolved yet...

denis-anisimov: OK, I missed the agreement.
So then it's the responsibility of the author to resolve regardless of comment or real cahnge.
OK.

caalador: This is more of an issue with how this works on github. With reviewable it only hid the conversation for the one clicking the acc button, but here it fully closes the comment block  making finding the conversation really hard especially if there happens to be multiple review rounds with comments in different faces.

PR #9626, opened 2020-12-09
refactor: rename the app-theme root folder from theme to themes

### LINKED ISSUE #9611: Rename the App Theme root folder to themes
As the `frontend/theme` folder may contain a multiple themes and the current rules are to use the `components` (in plural) folder for components custom styles, it makes sense to rename `theme` to `themes` so as to make the things consistent.

taefi:
Fixes #9611


caalador (changes requested):
theme folder in `test-embedding/test-embedding-application-theme` also need updating.

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -231,7 +231,7 @@ module.exports = {
     !devMode && new CompressionPlugin(),
 
     new ApplicationThemePlugin({
-      themeResourceFolder: path.resolve(flowFrontendFolder, 'theme'),
+      themeResourceFolder: path.resolve(flowFrontendFolder, 'themes'),
```
caalador, line 234:
```suggestion
      themeResourceFolder: path.resolve(flowFrontendFolder, 'theme'),
```
This matches `target/flow-frontend/theme/theme-generated.js` not `frontend/themes`

caalador (changes requested):

caalador (approved):

PR #9629, opened 2020-12-10
fix(offline): do not cache CSRF token and clear on offline logout

### LINKED ISSUE #9537: logout() should clear the CSRF token even if the logout call to the backend fails
As a user I expect that after I logged out of the app offline, I would not be able to make authenticated requests to the backend when coming back online, so that I can trust the "Logout" button. With Vaadin 19.0-SNAPSHOT / Flow 6.0.pr-9427-SNAPSHOT this is not the case.

**Steps to reproduce:**
 - check out the `offline-demo` branch of the Vaadin CRM app from https://github.com/vaadin-learning-center/crm-tutorial-typescript/tree/offline-demo
 - remove the workaround in `frontend/auth.ts` (keep only `await logoutImpl();`):
    ```ts
    (window as any).Vaadin.TypeScript.csrfToken = 'logged-out/offline';
    try {
      await logoutImpl();
    } catch {
      // whatever
    }
    ```
 - remove the app's own auth check from `offlineSyncToastHandler` (in `frontend/utils/offline-sync-toast-handler.ts`)
    ```ts
    if (!isLoggedIn()) {
      submitter.keepDeferredCallInTheQueue();
      return;
    }
    ```
 - open the app and login
 - go offline
 - make a deferrable call
 - logout() (see that the `/logout` network request fails and `logout()` throws)
 - manually change the URL to `/`
 - go online (see that deferred requests are submitted even though you have not logged in)

**Expected behavior:**
 - logout() does not throw even if the `/logout` network request fails
 - logout() clears the CSRF token even if the `/logout` network request fails
 - deferred request submission fails (as unauthorized) when going online after having logged out offline

joheriks:
Fixes #9537.

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
@@ -81,12 +81,17 @@ export async function login(username: string, password: string, options?: LoginO
 export async function logout(options?: LogoutOptions) {
   // this assumes the default Spring Security logout configuration (handler URL)
   const logoutUrl = options && options.logoutUrl ? options.logoutUrl : '/logout';
-  const response = await fetch(logoutUrl);
 
-  // TODO: find a more efficient way to get a new CSRF token
-  // parsing the full response body just to get a token may be wasteful
-  const token = getCsrfTokenFromResponseBody(await response.text());
-  (window as any).Vaadin.TypeScript.csrfToken = token;
+  await fetch(logoutUrl).then(async response => {
+    // TODO: find a more efficient way to get a new CSRF token
+    // parsing the full response body just to get a token may be wasteful
+    const token = getCsrfTokenFromResponseBody(await response.text());
+    (window as any).Vaadin.TypeScript.csrfToken = token;
+  }).catch(error => {
+    // clear the token if the call fails
+    delete (window as any).Vaadin.TypeScript.csrfToken;
+    throw error;
+  });
 }
 
 const getCsrfTokenFromResponseBody = (body: string): string | undefined => {
--- a/flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
@@ -82,16 +82,17 @@ export async function logout(options?: LogoutOptions) {
   // this assumes the default Spring Security logout configuration (handler URL)
   const logoutUrl = options && options.logoutUrl ? options.logoutUrl : '/logout';
 
-  await fetch(logoutUrl).then(async response => {
+  try {
+    const response = await fetch(logoutUrl);
     // TODO: find a more efficient way to get a new CSRF token
     // parsing the full response body just to get a token may be wasteful
     const token = getCsrfTokenFromResponseBody(await response.text());
     (window as any).Vaadin.TypeScript.csrfToken = token;
-  }).catch(error => {
+  } catch (error) {
     // clear the token if the call fails
     delete (window as any).Vaadin.TypeScript.csrfToken;
     throw error;
-  });
+  }
 }
 
 const getCsrfTokenFromResponseBody = (body: string): string | undefined => {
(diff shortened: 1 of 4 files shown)


haijian-vaadin (approved):
Looks good to me, seems a valid fix. Interested to see what @platosha and @Haprog think.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -162,10 +162,15 @@ private void addDevmodeGizmo(Document indexDocument, VaadinSession session,
...
     private void addInitialFlow(JsonObject initialJson, Document indexDocument,
-                                VaadinSession session) {
-        String csrfToken = session.getCsrfToken();
-        if (csrfToken != null) {
-            initialJson.put(CSRF_TOKEN, csrfToken);
+                                VaadinSession session, VaadinRequest request) {
+        // Do not add the CSRF token if the request comes from the service
+        // worker, to not have the token cached locally (#9537)
+        String referer = request.getHeader("referer");
+        if (referer == null || !referer.endsWith("/sw.js")) {
```
Haprog, line 169: I wonder is it guaranteed that this way of detecting request from SW will keep working in all supported browsers? Have we tested this on Chrome, Firefox and Safari? Any idea if this referer header for SW initiated requests is part of some spec?

I also started wondering does the user need to download the index page twice now on first load (first normally with the token and then SW would load and precache it without the token)? If so it would be good if we can somehow avoid the double load but that could be a new issue.

Otherwise the code seems good to me. If it works like intended on all supported browsers now, I'm ok with merging this.

flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
```diff
@@ -81,12 +81,17 @@ export async function login(username: string, password: string, options?: LoginO
 export async function logout(options?: LogoutOptions) {
   // this assumes the default Spring Security logout configuration (handler URL)
   const logoutUrl = options && options.logoutUrl ? options.logoutUrl : '/logout';
-  const response = await fetch(logoutUrl);
 
-  // TODO: find a more efficient way to get a new CSRF token
-  // parsing the full response body just to get a token may be wasteful
-  const token = getCsrfTokenFromResponseBody(await response.text());
-  (window as any).Vaadin.TypeScript.csrfToken = token;
+  await fetch(logoutUrl).then(async response => {
```
platosha, line 85: BTW not blocking, but looks like `try { await .... } catch { }` would fit here, making for less code not requiring inline callbacks.

platosha (dismissed):

platosha:
Needs rebase, otherwise looks good.

flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
joheriks: Done.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
joheriks: Tested it on Chrome, Firefox and Safari, and the use case in the ticket work (i.e., enqueueing requests and logging out while offline does not cause the deferred request to complete when going back online). I am doubtful whether `referer` is a standard or a lucky coincidence, will do some more investigation regarding that.

Haprog (approved):

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
Haprog: I tested that this seems to work ok currently also in Brave, Vivaldi and Firefox with default settings at least.

But if I set `network.http.sendRefererHeader` to `0` in Firefox (`about:config`), then the detection breaks and CSRF token is stored in workbox cache. Maybe it's still ok for now, but if we get a bug ticket about this later, we might want to consider implementing this in some other way.

joheriks: Decided to leave this solution for now. Other possibilities if this causes problems are: always sending the token in `index.html` but filtering it in precaching, as above but with a custom header in the request, or using cookie-based token transfer.

PR #9631, opened 2020-12-10
chore: fix flaky tests that depends on mock count

haijian-vaadin:
Move mock initialization to the setup method so that the tests depend on verify count won't affect each other.

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
@@ -53,6 +53,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.never;;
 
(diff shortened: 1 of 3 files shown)


joheriks (dismissed):

joheriks:
In addition, to deflake the tests, I think the following should be done:
1) Update `should_generateTs_files` and `should_generateOpenApi_when_EndpointPresents` to create a temporary folder (using e.g. `Files.createTempDirectory`) instead of `getClass().getClassLoader().getResource("com")`
2) Wrap each test that sets a system property with:
```java
        String originalPropertyValue = null;
        try {
            originalPropertyValue = System.getProperty("tested.property");
            System.setProperty("tested.property", baseDir);
           
            // actual test code

        } finally {
            if (originalUserDirValue != null) {
                System.setProperty("tested.property", originalUserDirValue);
            } else {
                System.clearProperty("tested.property");
            }
        }
```

platosha (dismissed):

joheriks (changes requested):
`DevModeInitializerTest.should_notGenerateOpenApi_when_EndpointIsNotUsed` is still flaky, see second most recent validation build.

joheriks (approved):

Issue #9552, opened 2020-12-02
Add a test for checking that the document.css file is injected to the document scope


pleku:
As per spec
>A special document.css stylesheet for css that doesn’t work inside SD and need to be loaded into the document root, such as @font-face declarations, when these need to work for components embedded into a non-Vaadin-app such as the design system documentation tool (DocBuilder).
>Java components exported as WCs using Flow’s WebComponentExporter apply the theme’s styles automatically inside the exported WC and css defined in document.css (e.g. @font-face declarations) to the document itself.

This is already done but we need to add a test that validates that e.g. a font from `document.css` is applied to the `html` element (which is the highest root element) where as the global styles go to `body`.

PR #9639, opened 2020-12-14
test: add extra document level CSS tests

mshabarov:
Adds a tests for checking that the document.css file is injected to the document scope and style dependencies from theme.json are available in document.

Fixes: https://github.com/vaadin/flow/issues/9552


mshabarov:
These changes should contain only the `document.css` styles tests and go on top of `master` branch.

PR #9642, opened 2020-12-14
Extract app config from deployment config

### LINKED ISSUE #9417: Make StubServletConfig private and remove every usage this class in the code outside of ServletDeployer
`ServletDeployer` is total evil.
Its method `createDeploymentConfiguration`  is used from everywhere and its usage is just buggy.
The only purpose of this method is: create a deployment configuration for a _automatically deployed servlet_.
It may not be used for anything else.

But it's used since sometimes some code wants deployment configuration just "because".
The problem is : this deployment configuration doesn't belong to anything.
If I try to set initial parameters via Servlet (and e.g. in most cases in OSGi this is the only way to set the parameters) they are ignore by _MAGIC_ without any clear reason.
I've spent ages to understand what's broken with my servlet registration: in fact nothing is broken but servlet deployment config is never invoked.
Instead some fake config is used.
This is extremely inconvenient / unclear and unacceptable. 

Classes which invokes `createDeploymentConfiguration` at least are: `VaadinAppShellInitializer`, `DevModeInitializer`.

denis-anisimov:
fixes #9417


flow-server/src/test/java/com/vaadin/flow/server/CustomUIClassLoaderTest.java
```diff
@@ -59,8 +62,12 @@ public void testWithDefaultClassLoader() throws Exception {
...
         Properties properties = new Properties();
         properties.put(InitParameters.UI_PARAMETER, MyUI.class.getName());
-        return new DefaultDeploymentConfiguration(CustomUIClassLoaderTest.class,
-                properties);
+        ApplicationConfiguration config = Mockito
+                .mock(ApplicationConfiguration.class);
+        Mockito.when(config.getPropertyNames())
+                .thenReturn(Collections.emptyEnumeration());
+        return new DefaultDeploymentConfiguration(config,
+                CustomUIClassLoaderTest.class, properties);
```
pleku, line 70: non-blocking, can be ignored sincethis is not that common except the couple of first test classes in the review:
```
        ApplicationConfiguration appConfig = Mockito
                .mock(ApplicationConfiguration.class);
        Mockito.when(appConfig.getPropertyNames())
                .thenReturn(Collections.emptyEnumeration());
```
 Could this bit of code could have been added to be an `EmptyMockDeploymentConfiguration(...)` or something to have to less c/p things to create this. Or add an utility method for the tests to reuse for creating this with given class & properties

flow-server/src/test/java/com/vaadin/flow/server/communication/UidlRequestHandlerTest.java
```diff
@@ -227,53 +232,33 @@ public void should_not_modify_non_MPR_Uidl() throws Exception {
...
-            "  \"barMethod\", [{}, {}]" +
-            " ],[]" +
-            "]," +
-            "\"meta\": {}, \"resources\": {},\"typeMappings\": {},\"typeInheritanceMap\": {}, \"timings\": []";
-
-        String locationChange =
-            "\"change\", {\"pid\": \"0\"}, [\"0\", {\"id\": \"0\", \"location\": \"http://localhost:9998/#!away\"}]";
-
-        String hashRpc =
-             "window.location.hash = '!away';";
```
pleku, line 273: Can you please revert this formatting change and add the missing `//@formatter:on` and `//@formatter:off` around it to avoid unnecessary changes in the future ?

fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java
```diff
@@ -128,105 +132,61 @@ public void teardown() throws Exception, SecurityException {
...
-                    generatedOpenApiJson.exists());
-        } finally {
-            if (originalJavaSourceFolder != null) {
-                System.setProperty("vaadin." 
-                    + CONNECT_JAVA_SOURCE_FOLDER_TOKEN, originalJavaSourceFolder);
-            } else {
-                System.clearProperty("vaadin." 
-                    + CONNECT_JAVA_SOURCE_FOLDER_TOKEN);
-            }
-            generatedOpenApiJson.delete();
```
pleku, line 160: Why was this cleanup removed ?

fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java
```diff
@@ -128,105 +132,61 @@ public void teardown() throws Exception, SecurityException {
...
-        }
```
pleku, line 191: Why was this cleanup removed ?

fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java
```diff
@@ -128,105 +132,61 @@ public void teardown() throws Exception, SecurityException {
...
-            assertTrue(ts2.exists());
-        } finally {
-            if (originalJavaSourceFolder != null) {
-                System.setProperty("vaadin." 
-                    + CONNECT_JAVA_SOURCE_FOLDER_TOKEN, originalJavaSourceFolder);
-            } else {
-                System.clearProperty("vaadin." 
-                    + CONNECT_JAVA_SOURCE_FOLDER_TOKEN);
-            }
-        }
```
pleku, line 229: Why was this cleanup removed ?

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
```diff
@@ -359,34 +292,35 @@ public void shouldNotThrow_tokenFileFoldersExist() throws Exception {
     }
 
     @Test
-    public void externalStatsFileTrue_predefinedContext() throws Exception {
```
pleku, line 362: This and the next external stats tests are removed but there is no replacement ?

pleku (commented):
Mostly questions

flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
```diff
@@ -126,21 +108,10 @@ public static DeploymentConfiguration createPropertyDeploymentConfiguration(
      * @throws VaadinConfigurationException
      *             thrown if property construction fails
      */
-    protected static Properties createInitParameters(
-            Class<?> systemPropertyBaseClass, VaadinConfig vaadinConfig)
-            throws VaadinConfigurationException {
+    protected Properties createInitParameters(Class<?> systemPropertyBaseClass,
+            VaadinConfig vaadinConfig) throws VaadinConfigurationException {
```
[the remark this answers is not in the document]
denis-anisimov: The warning is valid.
The exception is not thrown anymore because the code related to the deprecated annotation handling is removed along with the annotation.
Exception should be removed.

flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
```diff
@@ -67,14 +50,9 @@
...
             + "'flow-build-info.json' via running 'prepare-frontend' Maven goal.";
 
-    public static final String DEV_FOLDER_MISSING_MESSAGE = "Running project in development mode with no access to folder '%s'.%n"
-            + "Build project in production mode instead, see https://vaadin.com/docs/v15/flow/production/tutorial-production-mode-basic.html";
     private static final Logger logger = LoggerFactory
             .getLogger(DeploymentConfigurationFactory.class);
 
-    private DeploymentConfigurationFactory() {
-    }
-
```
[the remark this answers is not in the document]
denis-anisimov: The class may not be called `Factory`  as it has static methods only.
Factory is a generic pattern which is used via an instance to create something.
`static` methods may be considered as a factory methods _but_ the methods only may be considered as factory methods (not a whole class).
The class which contains only static methods is a `Helper` but not a factory.
So the name was confusing from the very beginning.

Several reason apart of that:
- It's possible to extend `DeploymentConfigurationFactory` now . (it's not possible to use the custom subclass but it's possible to allow to use it in the future: `Lookup`  can be used to get a `DeploymentConfigurationFactory` and then it will be possible to customize it via providing own config).
- `DeploymentConfigurationFactory` extends `AbstractConfigurationFactory` . `AbstractConfigurationFactory` contains methods which are used in subclasses of `AbstractConfigurationFactory`: `DefaultApplicationConfigurationFactory` is a real factory whose instance is used to create an `ApplicationConfiguration` and the `AbstractConfigurationFactory::getConfigParametersUsingTokenData`  method has to be an instance method ( so this method may not be used in the `static`  method as is. I would have been make an instance of `DeploymentConfigurationFactory` to use it which is absolutely ugly).
- Technically  `DeploymentConfigurationFactory`  is done consistently now with `DefaultApplicationConfigurationFactory` .
- Testing became much easier with correctly used patterns.

flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
```diff
@@ -87,10 +65,12 @@ private DeploymentConfigurationFactory() {
      * @throws VaadinConfigurationException
      *             thrown if property construction fails
      */
-    public static DeploymentConfiguration createDeploymentConfiguration(
```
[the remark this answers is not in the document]
denis-anisimov: I would like to avoid keeping `static` methods because of several reasons:
- as mentioned factory should be an object (not a class with static methods)
- I don't see any reason why would you want to use the methods in your app: they allow to make a deployment configuration and _the only purpose of this configuration_ is to create a `VaadinService` which should be created out of the box for the end developer.

I'm pretty sure that if you need to create a DeploymentConfiguration via those methods then you are doing something wrong (like eliminated code with `StubConfig` in `ServletContainerInitializer`  impls).

I may return back the static deprecated methods but I really would like to avoid this.
Do we have to keep backward compatibility in `6.0`?

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
```diff
@@ -43,22 +45,31 @@
...
      * @param systemPropertyBaseClass
      *            the class that should be used as a basis when reading system
      *            properties
      * @param initParameters
      *            the init parameters that should make up the foundation for
      *            this configuration
      */
-    public PropertyDeploymentConfiguration(Class<?> systemPropertyBaseClass,
-            Properties initParameters) {
+    public PropertyDeploymentConfiguration(
```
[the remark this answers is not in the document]
denis-anisimov: I'm not sure whether I get your concern right.....

But indeed I don't like `isOwnProperty` call here. The implementation is fragile indeed.

But fragility is non removable. It comes from the essence of configuration design.
`DeploymentConfiguration`  is an interface but it has a mixture of API and implementation methods.

E.g. `isProductionMode` is API method.
`getStringProperty` is an implementation method which should have been defined in an abstract class to reuse it in subclasses.

But we have what we have: 
- `isProductionMode`  whose value technically is based on `SERVLET_PARAMETER_PRODUCTION_MODE` property and `getStringProperty` is used to get the property value. **BUT**  `getStringProperty` doesn't have to be respected in `isProductionMode`  implementation since it's a totally different method. `isProductionMode`  may be implemented in one way and `getStringProperty` for `SERVLET_PARAMETER_PRODUCTION_MODE`  in another way. That's the consequence of the "API" design.
- As a result there are 4 cases to consider for `DeploymentConfiguration`:  `isProductionMode`  impl, `SERVLET_PARAMETER_PRODUCTION_MODE`  property value , `isProductionMode`  impl for _parent_ `ApplicationConfiguration` and `SERVLET_PARAMETER_PRODUCTION_MODE`  property value for _parent_ `ApplicationConfiguration`.

`DeploymentConfiguration` and `ApplicationConfiguration` properties are merged because `DeploymentConfiguration` should contain properties from the parent config as well.
But then `getApplicationProperty` (and `getStringProperty`  which is based on this method) _MAY NOT_ be used to implement `isProduction`: because `getApplicationProperty` asks for `DeploymentConfiguration` first and then it asks for  `ApplicationConfiguration` property. 
As a result `ApplicationConfiguration::isProductionMode` is ignored in this case. 
So only "implementation level" of the configuration design based on properties works. And this is wrong.

That's why I had to add `isOwnProperty`.

I can use another method for `isProductionMode`  implementation instead `getBooleanProperty` : yes.
But then I will have to introduce:
- override for `getBooleanProperty()` based on override for `getStringProperty`
- override for `getStringProperty()` : since the current impl may not be used as is ( described above why)
- override for `getApplicationProperty` which doesn't take into account parent properties.

So instead of one  `isOwnProperty` there will be 3 new methods.
And what's worst here: is you add one more property to the config (a new API method) then you won't be able to get _which_ `getBooleanProperty`/ `getStringProperty`/ `getApplicationProperty` to use: there will be couple of methods for every name.

I consider this as even more fragile : 3 additional methods which is too easy to mix up with the method having the same name.

I don't have strong opinion on this. I've decided that is `isOwnProperty` introduces less problems comparing to any other approach which I had as optional.

If you see the good proper way to solve this issue then I will be happy to apply it here.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServletConfiguration.java
```diff
@@ -1,124 +0,0 @@
...
- *             servlet is automatically registered and this annotation serves
- *             very little purpose.
- *
- * @author Vaadin Ltd
- * @since 1.0
- */
-@Retention(RetentionPolicy.RUNTIME)
-@Target(ElementType.TYPE)
-@Deprecated
-public @interface VaadinServletConfiguration {
```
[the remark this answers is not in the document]
denis-anisimov: yes, at least I see mentions of VaadinServletConfiguration is docs.
https://github.com/vaadin/docs/issues/73

flow-server/src/main/java/com/vaadin/flow/server/startup/AbstractConfigurationFactory.java
```diff
@@ -151,16 +157,6 @@
...
-    /**
-     * Sets to the dev mode properties to the configuration parameters.
-     * 
-     * @see #getConfigParametersUsingTokenData(JsonObject)
-     * 
-     * @param params
-     *            the configuration parameters to set dev mode properties to
-     * @param buildInfo
-     *            the token file data
-     */
```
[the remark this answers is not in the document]
denis-anisimov: Because of merge.
More PRs produces more problems with merge.

flow-server/src/main/java/com/vaadin/flow/server/startup/DefaultApplicationConfigurationFactory.java
```diff
@@ -78,25 +86,55 @@ public VaadinContext getContext() {
...
+     * 
+     * @param context
+     *            the Vaadin context, not {@code null}
+     * @param chunk
+     *            the fallback chunk, may be {@code null}
+     * @param properties
+     *            the context parameters, not {@code null}
+     * @return a new application configuration instance
+     */
+    protected ApplicationConfigurationImpl doCreate(VaadinContext context,
```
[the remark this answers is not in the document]
denis-anisimov: `create` contains an impl which you will have to repeat in your subclass and you don't want to do this for sure.

The implementation of `ItApplicationConfigurationFactory` is the answer.

flow-server/src/test/java/com/vaadin/flow/server/CustomUIClassLoaderTest.java
denis-anisimov: In the perfect world the tests should have been written so that they reuse some common code, yes.
But in the real world I've been going though many many unrelated tests  and fixed them one by one.
It was quite hard to understand what common cases they have.

I would consider this as a separate task since tests fixing was extremely annoying.
It's not that much code which is repeated and it's quite likely that this repetition won't be a repetition during code evolution. Then it will be just double work: unify and then de-unify.

flow-server/src/test/java/com/vaadin/flow/server/communication/UidlRequestHandlerTest.java
denis-anisimov: :(
Done

fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java
denis-anisimov: Why would you need a cleanup ?

Should every unit test do this cleanup ?

Clean up was there because of ugly usage system properties to "mock" deployment configuration.
There is no any need to set system property because mock is done properly.
As a result : no need to do cleanup .

fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java
denis-anisimov: Cleanup is not needed if you didn't litter.

fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java
denis-anisimov: Cleanup is not needed if you didn't litter.

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
denis-anisimov: The intention was to move these tests to `DefaultApplicationConfigurationFactoryTest`  which is the main reader of the 
token file (from the classpath ) but then I forgot about this because realized that `DeploymentConfigurationFactory` also may read the token file based on parameter value.

So yes, the tests need to be restored here and in the `DefaultApplicationConfigurationFactoryTest`

flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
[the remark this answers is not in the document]
denis-anisimov: Actually I'm wrong.
`VaadinServlet` creates a `DeploymentConfiguration` via the `createDeploymentConfiguration`  and a custom 
`VaadinServlet` may want to override the config created via  `createDeploymentConfiguration` .
So there can be usages of the static methods.
I may restore and deprecate them with the proper implementation which instantiates the `DeploymentConfigurationFactory` even though I would like to avoid static methods.

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
denis-anisimov: Actually there are some tests in `DefaultApplicationConfigurationFactoryTest` which read token file

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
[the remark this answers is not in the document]
pleku: I agree with your comments and I don't have a flawless solution to offer either.

One option, that we have sometimes opted for, would be to have a test that checks whether certain types of / named methods in e.g. `interface DeploymentConfiguration` are overridden in the child class `PropertyDeploymentConfiguration`, to avoid bugs that someone adds a new method with default implementation to the interface / abstract child class but not the `PropertyDeploymentConfiguration`.

I don't know how likely it is that this test will catch a bug in reality.

pleku (approved):
Only one comment open about a test checking whether methods are overridden or not, your call @denis-anisimov if it makes sense or not

flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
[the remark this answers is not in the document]
pleku: As it is quite rare case and trivial fix to do, agreed not to add back the static methods as deprecated.

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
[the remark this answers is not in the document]
denis-anisimov: There are many other ways to make the implementation invalid because of fragility.
In fact the default impl in the `DeploymentConfiguration` is not an issue until there is no the same method in the parent `ApplicationConfiguration`.
So default impl should be checked in the `AbstractConfiguration` . But then again it's possible to add methods with the same name into  `DeploymentConfiguration` and `ApplicationConfiguration` apart of `AbstractConfiguration`.

I think it's just not possible to catch all possible cases to make incorrect implementation so it's not worth to

denis-anisimov:
> Only one comment open about a test checking whether methods are overridden or not, your call @denis-anisimov if it makes sense or not

I've added the test sine this may be the most popular mistake. It won't catch all the possible errors but at least some.

pleku (approved):

PR #9646, opened 2020-12-14
test: add document.css level styles tests

mshabarov:
Adds a tests for checking that the document.css file is injected to the document scope and not injected to embedded element shadow root.

Fixes: https://github.com/vaadin/flow/issues/9552


flow-tests/test-embedding/test-embedding-application-theme/frontend/themes/embedded-theme/document.css
```diff
@@ -1,4 +1,4 @@
 @font-face {
-    font-family: "Ostrich";
-    src: url("./font/ostrich-sans-regular.ttf") format("TrueType");
+    font-family: "Open Sans";
```
caalador, line 2: Is there a reason to change the font away form Ostrich? 
Also now that you didn't update the global.css that uses Ostrich the feature can't be validated visually by opening the application.

caalador (changes requested):

flow-tests/test-embedding/test-embedding-application-theme/frontend/themes/embedded-theme/document.css
mshabarov: I think I've just mixed up the semantics of `global.css` and `document.css`.
According to a feature design doc:
>A special document.css stylesheet for css that doesn’t work inside SD and need to be loaded into the document root, such as @font-face declarations, when these need to work for components embedded into a non-Vaadin-app such as the design system documentation tool (DocBuilder).
Java components exported as WCs using Flow’s WebComponentExporter apply the theme’s styles automatically inside the exported WC and css defined in document.css (e.g. @font-face declarations) to the document itself.

The `Ostrich` font defined/imported in `document.css` and assigned in `global.css` within the `:host` block, should be propagated to document root and applied to an embedded component.

So, I've reverted the changes to global css and font folder and changed the tests accordingly.

flow-tests/test-embedding/test-embedding-application-theme/src/test/java/com/vaadin/flow/webcomponent/ApplicationThemeComponentIT.java
```diff
@@ -88,4 +88,42 @@ public void componentThemeIsApplied_forPolymerAndLit() {
...
+        Long adoptedStyles = (Long) getCommandExecutor().executeScript(
+                "return document.adoptedStyleSheets !== undefined ? document"
+                        + ".adoptedStyleSheets.length : 0");
+
+        Assert.assertTrue(
+                "Expected the document root to have the adopted styles",
+                adoptedStyles > 0);
+
+        Object ostrichFontStylesFound = getCommandExecutor().executeScript(
+                "return document.adoptedStyleSheets.map(styleSheet => styleSheet"
```
caalador, line 105: I feel this doesn't check that the `@font-face` declaration has been inserted only that some arbitrary rule in position 0 contains ostrich as a string part.
Expecting it to always be in index 0 is also easily broken by some minor change.
I would propose to make it more robust by using for instance
```
let found = false;
for (let as=0; as<document.adoptedStyleSheets.length;++as) {
    let styleSheetRules = document.adoptedStyleSheets[as].rules;
    for (var ss = 0; ss < styleSheetRules.length; ++ss) {
        let cssRule = styleSheetRules[ss];
            if (cssRule instanceof CSSFontFaceRule && cssRule.cssText.startsWith("@font-face { font-family: Ostrich;")) {
                found = true;
            } 
    }
}
return found;
```

This would guarantee that it is the `@font-face` declaration.

flow-tests/test-embedding/test-embedding-application-theme/src/test/java/com/vaadin/flow/webcomponent/ApplicationThemeComponentIT.java
```diff
@@ -88,4 +88,42 @@ public void componentThemeIsApplied_forPolymerAndLit() {
...
+    @Test
+    public void documentCssFonts_fromLocalCssFile_fontAppliedToEmbeddedComponent() {
+        open();
+
+        Object ostrichFontStylesFoundForEmbedded = getCommandExecutor()
+                .executeScript(
+                        "return document.getElementsByTagName('themed-component')[0]"
+                                + ".shadowRoot.adoptedStyleSheets.map(styleSheet => "
+                                + "styleSheet.cssRules[0].cssText).filter(cssText => "
+                                + "cssText.indexOf('Ostrich') > 0).length > 0;");
```
caalador, line 123: Same script as above, but the result should be false as the `@font-face` declaration should not be added to the embedded component.

caalador (changes requested):

flow-tests/test-embedding/test-embedding-application-theme/src/test/java/com/vaadin/flow/webcomponent/ApplicationThemeComponentIT.java
mshabarov: Done

flow-tests/test-embedding/test-embedding-application-theme/src/test/java/com/vaadin/flow/webcomponent/ApplicationThemeComponentIT.java
mshabarov: Done

caalador (approved):

PR #9649, opened 2020-12-14
fix(offline): simple state transitions when only Connect client is loaded

### LINKED ISSUE #9506: the connection state indicator is not visible in pure Fusion apps
When building an app with Fusion I expect the connection state indicator to be visible even if I do not add any server-side views, so that the UX is consistent between Fusion / Flow / hybrid Vaadin apps. With Flow 6.0.pr-9427-SNAPSHOT this is not the case.

Steps to reproduce:
 - create a TS + HTML app with start.vaadin.com
 - override the versions of Vaadin and Flow to 19.0-SNAPSHOT and 6.0.pr-9427-SNAPSHOT
 - remove the `Flow` import and the use of server-side routes in `index.ts`:
    ```typescript
    import { Flow } from '@vaadin/flow-frontend/Flow';
    
    const { serverSideRoutes } = new Flow({
      imports: () => import('../target/frontend/generated-flow-imports'),
    });
    ```
 - start the app
 - turn on the offline more in the browser dev tools
 - see that there is no connection state indicator

joheriks:
In pure fusion apps (`Flow.ts` not loaded), the indicator will directly reflect the
`navigator.onLine` state. If `Flow.ts` is loaded, a simple server connectivity check 
is done instead. If additionally the GWT client is loaded, the V14 reconnection logic 
(same as after missed heartbeat) will be done instead.

Fixes #9506.


platosha (approved):

flow-client/src/main/resources/META-INF/resources/frontend/Connect.ts
```diff
@@ -285,6 +285,19 @@ export class ConnectClient {
 
     // ensure connection indicator is visible from this entry point
     getConnectionIndicator();
+
+    // Listen to browser online/offline events and update the loading indicator accordingly.
+    // Note: if Flow.ts is loaded, it instead handles the state transitions.
+    $wnd.addEventListener('online', () => {
+      if ($wnd.Vaadin.Flow === undefined) {
```
fluorumlabs, line 292: I know it's a bit of the edge case, but `Vaadin.Flow` could be also defined if there is an embedded Flow component (not the whole application, just a component, probably even coming from another server)

haijian-vaadin: Any suggestions on how it can be improved?

haijian-vaadin: I prefer to merge the PR now, as the fix is more important than covering the edge case atm. 
After the merge, we can start merging the offline branch to master, which can unblock the release. 
Further suggestions can be applied to the master branch later on.

fluorumlabs: Hard to say. To me it looks Flow client doesn't listen to `online`/`offline` events at all. Needs to be rechecked though.

fluorumlabs: > I prefer to merge the PR now, as the fix is more important than covering the edge case atm.

Sure thing. Let's move this to technical debt :)

joheriks: https://github.com/vaadin/flow/blob/feature/offline/flow-client/src/main/java/com/vaadin/client/communication/DefaultConnectionStateHandler.java#L541-L551

Issue #9585, opened 2020-12-07
Support making reusable ApplicationTheme


caalador:
The feature for supporting reusable application theme should be added. This would mean creating a jar that only contains theme resources and a java class with Npm resource annotations if needed.

For an example see https://github.com/Artur-/theme-example/tree/reusable

There should be a test module that applies the theme from the jar file.

pleku:
#### Acceptance Criteria
- [ ] Adding a module with a theme that has some css and resources in src/main/resources/META-INF/resources/themes/reusable-theme/ and it is used by another module

tanbt:
When packaging themes to JAR file, themes from (at least) two locations:
* `src/main/resources/META-INF/resources/themes/` 
* `frontend/themes/`
should be packaged, together with the related Java files which contains `@NpmPackage`.

PR #9650, opened 2020-12-14
Support reusable application theme

tanbt:
Resolves https://github.com/vaadin/flow/issues/9585


flow-server/src/main/resources/webpack.generated.js
```diff
@@ -228,7 +231,7 @@ module.exports = {
 
     new ApplicationThemePlugin({
       // The following matches target/flow-frontend/theme/theme-generated.js and not frontend/themes
```
caalador, line 233: Comment is wrong now. Should be updated to say it for the theme-generated and for the jar theme

caalador (changes requested):

flow-server/src/main/resources/webpack.generated.js
tanbt: Updated.

flow-tests/test-resources/src/main/java/com.vaadin.apptheme.reusabletheme/Dependency.java
```diff
@@ -0,0 +1,8 @@
+package org.vaadin.apptheme.reusabletheme;
```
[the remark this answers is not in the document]
tanbt: Moved to reusable module.

flow-tests/test-resources/src/main/resources/META-INF/resources/themes/reusable-theme/README.md
```diff
@@ -0,0 +1,2 @@
+This theme is cloned from frontend/themes/app-theme.
```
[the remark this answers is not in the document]
tanbt: Moved to reusable module.

pom.xml
```diff
@@ -36,6 +36,7 @@
         <module>flow-bom</module>
         <module>fusion-endpoint</module>
         <module>build-tools</module>
+        <module>flow-tests/test-application-theme/reusable-theme</module>
```
caalador, line 39: This is fine to build only in test-application-theme when it is built. it shouldn't be used outside of the module.

flow-tests/test-application-theme/reusable-theme/pom.xml
```diff
@@ -0,0 +1,20 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>reusable-theme</artifactId>
+    <name>Flow reusable application theme</name>
+    <packaging>jar</packaging>
+
+    <properties>
+        <maven.deploy.skip>true</maven.deploy.skip>
+    </properties>
+
```
caalador, line 19: I expect that you will need some flow dependencies to get the imports to compile as at least flow-tests only depend on lumo and not much else.

caalador (changes requested):

pom.xml
tanbt: Removed.

flow-tests/test-application-theme/reusable-theme/pom.xml
tanbt: Some flow dependencies are already in the parent (or grandparent) of this module already. Is that enough?

caalador: Should be fine if it compiles.

flow-tests/test-application-theme/reusable-theme/src/main/resources/META-INF/resources/themes/reusable-theme/global.css
```diff
@@ -0,0 +1,26 @@
...
+    height: 274px;
+    display: block;
+}
+
+#octopuss {
+    background-image: url('/octopuss.jpg');
+    width: 224px;
+    height: 224px;
+    display: block;
+}
```
caalador, line 26: Not certain we would want the external urls in the reusable theme as then the one reusing should somehow have them served.

caalador (changes requested):

flow-tests/test-application-theme/reusable-theme/src/main/resources/META-INF/resources/themes/reusable-theme/global.css
tanbt: Yeah, make sense.
It's not a good practice to to call an external resources inside an independent theme pack (JAR file).
Should we remove this test?

flow-server/src/main/java/com/vaadin/flow/server/Constants.java
```diff
@@ -240,6 +240,16 @@
      */
     public static final String RESOURCES_FRONTEND_DEFAULT = "META-INF/frontend";
 
+    /**
+     * The name of the application theme root folder.
+     */
+    public static final String APPLICATION_THEME_ROOT = "themes";
```
caalador, line 246: @taefi  is running for this to be in FrontendUtils.
> Of course, agree to define it once, but I have a feeling that it is better to define it here since many other 
> default/hardcoded path and folder constants defined in this file, e.g. NODE_MODULES, FRONTEND, 
> DEFAULT_NODE_DIR, DEFAULT_FRONTEND_DIR, etc. I propose to keep it here but rename it to 
> APPLICATION_THEME_ROOT as Tan defined.

tanbt: So should I move this constant to `FrontendUtils`? Or @taefi  will do that in his PR?

pleku: Please resolve this kind of trivial issues as fast as possible by e.g. just agreeing over Slack

pleku (approved):

caalador (approved):

tanbt:
I guess @taefi  can continue from `master`, i.e. moving `APPLICATION_THEME_ROOT` to FrontendUtils
and update the comment `// If we define a theme name we need to import theme/theme-generated.js` to `.... themes/theme-generated.js`

PR #9652, opened 2020-12-14
fix(offline): Rewrite <base href> in service worker using sw scope

### LINKED ISSUE #8878: Offline start should work for deep URLs as well
When starting the app offline, from a cached app shell, as a developer I want to Vaadin to rewrite the base URL in the cached app shell so that it consistently points to the application deployment root URL regardless of whether the app is offline-started from the root URL (`https://my.app.com/`) or from a deep URL (`https://my.app.com/some/path`) so that users can start the app from deep URLs as well.

The app shell may have relative URLs like `<a href="issues/new">`. When the app is loaded from the `/` URL such links would be resolved relative to `/` (i.e. `issues/new` becomes `/issues/new`). When the app is loaded from the `/issues/new` URL such links _by default_ would be resolved relative to `/issues/new` (i.e. `issues/new` becomes `/issues/new/issues/new`). That breaks the app because links no longer point to the correct resources.

In order to counter that, when Vaadin dynamically writes a `<base href="[base URL]">` tag when the app shell is generated on the server (see `IndexHtmlRequestHandler`). However, this does not happen when the app shell is retrieved from the cache in the Service Worker (as implemented in https://github.com/vaadin/flow/issues/8854).

DoD:
 - the app shell is loaded correctly and relative in-app links work in the same way both when the app is started offline from the root path (`/`), and from a deep path (e.g. `/users/new`).

Implementation notes:
Workbox allows custom processing of an HTTP response after it is retrieved from the cache and before it is returned to the browser. A prototype is available in https://github.com/vlukashov/v18-offline-app/commit/977d100138115adcc9e92d19243871b8445ac57f

platosha:
Fixes #8878


joheriks (approved):

haijian-vaadin (approved):
Tried with the PR  snapshot, it works, awesome!

Issue #9656, opened 2020-12-15
Grid ComponentRenderer may fail to render since CompositeDataGenerator is not deterministic


TatuLund:
Grid ComponentRenderer may fail to render since CompositeDataGenerator is not deterministic. See the sample code below.

As the item has two generators, the component renderer and class name generator, calling refreshItem will trigger them to occasionally be called in different order. Refresh page multiple times and you will sometimes see that Button is not rendered.

```
import java.util.Arrays;
import com.vaadin.flow.component.button.Button;
import com.vaadin.flow.component.grid.Grid;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.data.provider.ListDataProvider;
import com.vaadin.flow.router.Route;

@Route("componentbug")
public class ComponentRendererBug extends VerticalLayout {

    public ComponentRendererBug() {
        final Grid<Integer> grid = new Grid<>();
        grid.setDataProvider(new ListDataProvider<>(Arrays.asList(1,2,3)));
        grid.addColumn(integer -> integer);
        Grid.Column<Integer> componentColumn = grid.addComponentColumn(integer -> new Button("Button "+integer));
        componentColumn.setClassNameGenerator(integer -> {
            grid.getDataProvider().refreshItem(integer);
            return "custom-class-name";
        });
        add(grid);
    }
}
```

PR #9657, opened 2020-12-15
fix: use LinkedHashSet (not HashSet) for deterministic CompositeDataGenerator 

TatuLund:
Fixes https://github.com/vaadin/flow/issues/9656

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/CompositeDataGenerator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/CompositeDataGenerator.java
@@ -39,7 +39,7 @@ import elemental.json.JsonObject;
 public class CompositeDataGenerator<T>
         implements DataGenerator<T>, HasDataGenerators<T> {
 
-    private final Set<DataGenerator<T>> dataGenerators = new LinkedHashSet<>();
+    final Set<DataGenerator<T>> dataGenerators = new LinkedHashSet<>();
 
     @Override
     public void generateData(T item, JsonObject jsonObject) {
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/CompositeDataGenerator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/CompositeDataGenerator.java
@@ -15,8 +15,8 @@
  */
 package com.vaadin.flow.data.provider;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.util.ArrayList;
+import java.util.List;
 
 import com.vaadin.flow.shared.Registration;
 
@@ -39,7 +39,7 @@ import elemental.json.JsonObject;
 public class CompositeDataGenerator<T>
         implements DataGenerator<T>, HasDataGenerators<T> {
 
-    private final Set<DataGenerator<T>> dataGenerators = new HashSet<>();
+    private final List<DataGenerator<T>> dataGenerators = new ArrayList<>();
 
     @Override
     public void generateData(T item, JsonObject jsonObject) {
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/CompositeDataGenerator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/CompositeDataGenerator.java
@@ -15,8 +15,8 @@
  */
 package com.vaadin.flow.data.provider;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 import com.vaadin.flow.shared.Registration;
 
@@ -39,7 +39,7 @@ import elemental.json.JsonObject;
 public class CompositeDataGenerator<T>
         implements DataGenerator<T>, HasDataGenerators<T> {
 
-    private final List<DataGenerator<T>> dataGenerators = new ArrayList<>();
+    private final Set<DataGenerator<T>> dataGenerators = new LinkedHashSet<>();
 
     @Override
     public void generateData(T item, JsonObject jsonObject) {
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/CompositeDataGenerator.java
```diff
@@ -39,7 +39,7 @@
 public class CompositeDataGenerator<T>
         implements DataGenerator<T>, HasDataGenerators<T> {
 
-    private final Set<DataGenerator<T>> dataGenerators = new HashSet<>();
+    private final List<DataGenerator<T>> dataGenerators = new ArrayList<>();
```
joheriks, line 42: Does this change API slightly in that `addDataGenerator` is no longer idempotent? Would perhaps be safer to use `LinkedHashSet `, which has set semantics but preserves order.

TatuLund: True

Artur-:
I am not saying tests are mandatory but if you leave them out, it means you are fine if somebody reverts the patch tomorrow. It should not be difficult to add a test that the order is preserved

joheriks:
> Ok, did that, which required to make dataGenerators package protected.

It can be tested without the visibility modificiation, with just slightly blacker-box test. E.g. public method `destroyAllData` calls the `dataGenerators` in order. The test could register e.g. three `DataGenerator` instances which have side effect on a local mutable, e.g., appending `this` to a local list. Then the list can be asserted to be in order of registration.

flow-data/src/test/java/com/vaadin/flow/data/provider/CompositeDataGeneratorTest.java
```diff
@@ -196,4 +196,14 @@ public void dataGeneratorRegistration_remove_dataIsDestroyed() {
                 CoreMatchers.not(CoreMatchers.hasItem("item1")));
     }
 
+    @Test
+    public void orderIsPreserved() {
+        CompositeDataGenerator<String> cdg = new CompositeDataGenerator<>();
+        DataGenerator<String> dg1 = (String, JsonObject) -> {};
+        DataGenerator<String> dg2 = (String, JsonObject) -> {};
+        List<DataGenerator<String>> expected = Arrays.asList(dg1,dg2);
```
joheriks, line 204: `Arrays` may need import.

flow-data/src/test/java/com/vaadin/flow/data/provider/CompositeDataGeneratorTest.java
```diff
@@ -196,4 +196,14 @@ public void dataGeneratorRegistration_remove_dataIsDestroyed() {
...
+        cdg.addDataGenerator(dg1);
+        cdg.addDataGenerator(dg2);
+        assertEquals(expected,new ArrayList<>(cdg.dataGenerators));
```
joheriks, line 207: `Assert.assertEquals` or import.

joheriks (changes requested):

TatuLund:
> It can be tested without the visibility modificiation

The both approaches have their pros and cons, it is really hard to judge which is better.

joheriks (approved):

joheriks:
> The both approaches have their pros and cons, it is really hard to judge which is better.

True, I'm fine with package-private. Just wanted to highlight the alternative.

flow-data/src/test/java/com/vaadin/flow/data/provider/CompositeDataGeneratorTest.java
```diff
@@ -196,4 +200,14 @@ public void dataGeneratorRegistration_remove_dataIsDestroyed() {
                 CoreMatchers.not(CoreMatchers.hasItem("item1")));
     }
 
+    @Test
+    public void orderIsPreserved() {
```
mshabarov, line 204:
```suggestion
    public void addDataGenerator_orderIsPreserved() {
```

flow-data/src/test/java/com/vaadin/flow/data/provider/CompositeDataGeneratorTest.java
```diff
@@ -196,4 +200,14 @@ public void dataGeneratorRegistration_remove_dataIsDestroyed() {
...
+        CompositeDataGenerator<String> cdg = new CompositeDataGenerator<>();
+        DataGenerator<String> dg1 = (String, JsonObject) -> {};
+        DataGenerator<String> dg2 = (String, JsonObject) -> {};
+        List<DataGenerator<String>> expected = Arrays.asList(dg1,dg2);
+        cdg.addDataGenerator(dg1);
+        cdg.addDataGenerator(dg2);
+        assertEquals(expected,new ArrayList<>(cdg.dataGenerators));
```
mshabarov, line 211:
```suggestion
        assertEquals(expected, new ArrayList<>(cdg.dataGenerators));
```

flow-data/src/test/java/com/vaadin/flow/data/provider/CompositeDataGeneratorTest.java
```diff
@@ -196,4 +200,14 @@ public void dataGeneratorRegistration_remove_dataIsDestroyed() {
                 CoreMatchers.not(CoreMatchers.hasItem("item1")));
     }
 
+    @Test
+    public void orderIsPreserved() {
+        CompositeDataGenerator<String> cdg = new CompositeDataGenerator<>();
+        DataGenerator<String> dg1 = (String, JsonObject) -> {};
+        DataGenerator<String> dg2 = (String, JsonObject) -> {};
+        List<DataGenerator<String>> expected = Arrays.asList(dg1,dg2);
```
mshabarov, line 208:
```suggestion
        List<DataGenerator<String>> expected = Arrays.asList(dg1, dg2);
```

mshabarov (approved):
Small nits

mshabarov (changes requested):
> > It can be tested without the visibility modificiation

> The both approaches have their pros and cons, it is really hard to judge which is better.

I would go with Johannes's approach (use `callOrder`) and keep the modifier `private`, because the package consists of a lot of classes with heavy data handling logic, and the risk of accidentally modify the internal state is high.

mshabarov (approved):

mshabarov:
Cherry picked to `2.4`

PR #9663, opened 2020-12-15
merge master to offline

haijian-vaadin:
Merge, DO NOT SQUASH


flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
```diff
@@ -302,62 +301,95 @@ public void shouldUseFullPathScannerByDefault() throws Exception {
...
+            originalJavaSourceFolder = System.getProperty("vaadin." 
+                + CONNECT_JAVA_SOURCE_FOLDER_TOKEN);
+            System.clearProperty("vaadin." 
+                + CONNECT_JAVA_SOURCE_FOLDER_TOKEN);
+
+            
+            Assert.assertFalse(generatedOpenApiJson.exists());
+            devModeInitializer.onStartup(classes, servletContext);
+
+            Mockito.verify(taskGenerateConnect, never()).execute();
```
joheriks, line 352: Looks like there is still a race condition here. This test passes for me locally, but if I add a small delay (`            Thread.sleep(150)`) above this line (`Mockito.verify`) it fails.

joheriks: Is there potentially a
```
        Mockito.doReturn(null).when(lookup).lookup(EndpointGeneratorTaskFactory.class);
```
missing from this test? As by the default `DevModeInitializerTestBase.setup` returns a factory mock but this test seems to be about the case when endpoints are disabled.

haijian-vaadin: good catch, will do

haijian-vaadin:
Closing this one, as the commits from master were merged as a single commit. Will make another PR.

PR #9676, opened 2020-12-16
fix: verify theme folder existence based on app-theme name in @Theme

### LINKED ISSUE #9610: Throw a meaningful exception when no custom theme folder created
It's not clear enough what is happening and how to fix the error when webpack compilation fails on App startup due to absence of `my-theme` folder, whenever the `@Theme("my-theme")` annotation is present. 
Webpack shows:
```
ERROR in ../target/flow-frontend/theme/theme-generated.js
Module not found: Error: Can't resolve 'theme/my-theme/my-theme.js' in '/Users/mikhail/workspace/vaadin/application-theme-dx/appthemedx-master/target/flow-frontend/theme'
```
and the browser console shows: `Uncaught (in promise) Error: Cannot find module 'theme/my-theme/my-theme.js'`.

Although the App Theme feature experienced developers may quickly recognise the issue, let's explicitly tell the developer what to do by throwing the exception immediately with a meaningful message what happened and how to fix it.

1. Download app project https://gitlab.vaadin.com/mgrankvi/appthemedx/-/tree/master
2. Add `@Theme("my-theme")` on `Application` class
3. Add `theme` folder to `frontend` folder
4. Run the app `mvn spring-boot:run` and open the page localhost:8080.
5. See webpack error and error message in log.
### Expected behavior
Application logs the explanatory message what the developer did wrong
### Actual behavior
Webpack fails and shows low-level error message
### Versions:
    - Vaadin / Flow version: flow `6.0.0.alpha1`
    - Java version: 8
    - OS version: mac os
    - Browser version (if applicable): chrome 87
    - Application Server (if applicable): spring-boot app
    - IDE (if applicable): IDEA ultimate

taefi:
Fixes #9610


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -316,6 +316,13 @@
 
     public static final String BRIGHT_BLUE = "\u001b[94m%s\u001b[0m";
 
+    /**
+     * The directory name for the themes directory. It should be the direct
+     * child of frontend directory which is by default named as "frontend", @see
+     * {@link FrontendUtils#FRONTEND} and {@link FrontendUtils#DEFAULT_FRONTEND_DIR}
+     */
+    public static final String APP_THEMES_FOLDER_NAME = "themes";
```
caalador, line 324: Tan is addin to Constants 
```
    /**
     * The name of the application theme root folder.
     */
    public static final String APPLICATION_THEME_ROOT = "themes";
```
And I feel we should define it only in one place.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -67,4 +75,20 @@ public void execute() throws ExecutionFailedException {
                 "Unable to write theme import file", e);
         }
     }
+
+    private void verifyThemeDirectoryExistence() throws ExecutionFailedException {
```
caalador, line 79: Nit: you could just have the method take in `String themeName` to not have `theme.getName()` in so many places

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -67,4 +75,20 @@ public void execute() throws ExecutionFailedException {
...
+        File themesDir = new File (frontendDirectory, APP_THEMES_FOLDER_NAME);
```
caalador, line 80: Again this is not the only place we support so we should also check 
- `src/main/resources/META-INF/resources`
- `src/main/resources/static`
- `${flowFrontend}` (where the generated is written to, will also get the jar contents)

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -67,4 +75,20 @@ public void execute() throws ExecutionFailedException {
...
+        File customThemeDir = new File (themesDir, theme.getName());
+        if (!customThemeDir.exists()) {
+            String errorMessage = "Discovered @Theme(\"%s\") annotation but " +
+                    "no \"%s/%s/%s/\" directory present in the " +
+                    "project or available or inside a jar dependency for the " +
+                    "app. Check if you forgot to create the folder or have " +
+                    "typo in the theme or folder name \"%s\".";
```
caalador, line 87:
```suggestion
            String errorMessage = "Discovered @Theme annotation with theme name '%s', " +
                    "but could not find the theme directory in the " +
                    "project or available as a jar dependency." +
                    "Check if you forgot to create the folder in '%s' or have " +
                    "mistyped the theme or folder name for '%s'.";
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -67,4 +75,20 @@ public void execute() throws ExecutionFailedException {
...
+
+            throw new ExecutionFailedException(String.format(errorMessage,
+                    theme.getName(), frontendDirectory.getName(),
+                    APP_THEMES_FOLDER_NAME, theme.getName(), theme.getName()));
```
caalador, line 91:
```suggestion
            throw new ExecutionFailedException(String.format(errorMessage,
                    theme.getName(), themesDir, theme.getName()));
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
caalador: We can ofc suggest creating to the frontend folder, but we should still check all the possibilities. 
Also this doesn't check for jar contents at the moment. This is possible to check in the flowFrontend directory as the files will have been copied out from the jar before we process this step.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
taefi: Of course, agree to define it once, but I have a feeling that it is better to define it here since many other default/hardcoded path and folder constants defined in this file, e.g. NODE_MODULES, FRONTEND, DEFAULT_NODE_DIR, DEFAULT_FRONTEND_DIR, etc. I propose to keep it here but rename it to `APPLICATION_THEME_ROOT` as Tan defined.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -76,18 +80,63 @@ public void execute() throws ExecutionFailedException {
...
+
+    private Stream<File> getOtherFoldersToSearchForAppTheme(String themeName,
+                                                    File frontendDirectory) {
+        File projectRootDir = frontendDirectory.getParentFile();
+
+        File metaInfDir = new File(projectRootDir,
+                APPLICATION_META_INF_RESOURCES);
+        File metaInfThemesDir = new File(metaInfDir, APPLICATION_THEME_ROOT);
+        File metaInfCustomThemeDir = new File(metaInfThemesDir, themeName);
+
```
caalador, line 129: I would just have this method return a list (stream?) of strings concatenate as `String.join("/", APPLICATION_META_INF_RESOURCES, APPLICATION_THEME_ROOT, themeName)` Also I would not call this other folders as it should just be the full set of folders.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -76,18 +80,63 @@ public void execute() throws ExecutionFailedException {
...
+                        "folder under './%s/%s/' or have mistyped the theme " +
+                        "or folder name for '%s'.";
+
+                throw new ExecutionFailedException(String.format(errorMessage,
+                        themeName, frontendDirectory.getName(),
+                        APPLICATION_THEME_ROOT, themeName));
+            }
+        } else {
+            if (otherFoldersToSearch.anyMatch(File::exists)) {
+                String errorMessage = "Discovered Theme folder for theme " +
```
caalador, line 108: This exception is sort of wrong if you have the same themeName in frontend and a jar.
In that case the exception should be more like `"Theme '" + themeName + "'should not exist inside a jar and in the project at the same time\n Extending another theme is possible by adding { \"parent\": \"my-parent-theme\" } entry to the theme.json file inside your theme folder."`
But this case is checked in the webpack plugin.

flow-server/src/main/java/com/vaadin/flow/server/Constants.java
```diff
@@ -412,6 +412,25 @@
...
+    /**
+     * The constant "src/main/resources/META-INF/resources" path in project
+     * source files.
+     */
+    public static final String APPLICATION_META_INF_RESOURCES = "src/main/resources/META-INF/resources";
+
+    /**
+     * The constant "src/main/resources/static" path in project sources files.
+     */
+    public static final String APPLICATION_STATIC_RESOURCES = "src/main/resources/static";
```
caalador, line 432: Will anyone else have need of these paths? I would have them in the task only.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
caalador: And the stream/list would be used as `Stream.filter(path -> new File(npmFolder, path).exists()).count()`

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -76,18 +80,63 @@ public void execute() throws ExecutionFailedException {
...
-            String errorMessage = "Discovered @Theme annotation with theme name '%s', " +
-                    "but could not find the theme directory in the " +
-                    "project or available as a jar dependency." +
-                    "Check if you forgot to create the folder in '%s' or have " +
-                    "mistyped the theme or folder name for '%s'.";
-
-            throw new ExecutionFailedException(String.format(errorMessage,
-                    theme.getName(), themesDir, theme.getName()));
+    private void verifyThemeDirectoryExistence(String themeName,
+                                               File frontendDirectory)
```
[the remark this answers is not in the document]
taefi: I'm trying to maintain integrity for the input parameters, so either the method to be totally dependent on fields or totally act as independent as possible. In this case, I would remove `theme name` also from the input args and would create a local variable as `String themeName = theme.getName()`, since feels the same for me to pass `theme.getName()` to it while the `theme` is available as a field.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -76,18 +80,63 @@ public void execute() throws ExecutionFailedException {
...
+
+                throw new ExecutionFailedException(String.format(errorMessage,
+                        themeName, frontendDirectory.getName(),
+                        APPLICATION_THEME_ROOT));
+            }
         }
     }
+
+    private Stream<File> getOtherFoldersToSearchForAppTheme(String themeName,
+                                                    File frontendDirectory) {
```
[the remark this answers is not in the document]
taefi: Same as https://github.com/vaadin/flow/pull/9676/files#r546623022

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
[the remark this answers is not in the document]
caalador: We shouldn't pass instance variables around internal private methods. Even though theme is available the name is as such not, but it's fine to have getName everywhere it is needed..

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -76,18 +80,63 @@ public void execute() throws ExecutionFailedException {
...
+        File projectRootDir = frontendDirectory.getParentFile();
```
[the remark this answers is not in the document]
taefi: Yeah, right, totally understand this since I was struggling with the same issue in my mind while I was taking the look at available runtime parameters from `builder` in `NodeTask.java` and didn't find any proper parameter which holds the `project-root`. Should I assume that `npmFolder` is the project-root? Since I wasn't sure that it is not gonna be optimized by the users.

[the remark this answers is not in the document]
caalador: It's where everything else is calculated from also so it should be fine to assume it's the project root. it bsically needs to be the place where pom.xml exists.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
taefi: I'm validating this to cover such situations that user wrongly copies the same `/themes/some-theme` structure in more than one place, for instance having it in both `[project-root]/frontend/themes/some-theme` and `[project-root]/src/main/resources/static/themes/some-theme`. As far as I understood, webpack is only checking the duplicity in jar and project files, not between to places in project files. Please correct me if I'm wrong.

caalador: yes, but as this is done earlier and the check does even check for jar resource theme it should handle the fronten+jar scenario and inform that the jar should be marked as a parent theme and the other should get a new name.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -70,4 +84,77 @@ public void execute() throws ExecutionFailedException {
...
+
+        String themePathInStaticResources = String.join("/",
+                APPLICATION_STATIC_RESOURCES, themePath);
+
+        String themePathInClassPathResources = FrontendUtils.NODE_MODULES
+                + FrontendUtils.FLOW_NPM_PACKAGE_NAME + themePath;
+
+        return Arrays.asList(themePathInMetaInfResources,
+                themePathInStaticResources, themePathInClassPathResources);
+    }
```
caalador, line 159:
```suggestion
    private List<String> getAppThemePossiblePaths(String themePath) {
        String frontendTheme = String.join("/",
            npmFolder.toPath().relativize(frontendDirectory.toPath())
                .toString(), themePath);
                
        String themePathInMetaInfResources = String.join("/",
                APPLICATION_META_INF_RESOURCES, themePath);

        String themePathInStaticResources = String.join("/",
                APPLICATION_STATIC_RESOURCES, themePath);

        String themePathInClassPathResources = String
            .join("", FrontendUtils.NODE_MODULES,
                FrontendUtils.FLOW_NPM_PACKAGE_NAME, themePath);

        return Arrays.asList(frontendTheme, themePathInMetaInfResources,
            themePathInStaticResources, themePathInClassPathResources);
    }
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -70,4 +84,77 @@ public void execute() throws ExecutionFailedException {
...
+        if (!mainCustomThemeDir.exists()
+                && existingAppThemeDirectories.isEmpty()) {
+            String errorMessage = "Discovered @Theme annotation with theme "
+                    + "name '%s', but could not find the theme directory "
+                    + "in the project or available as a jar dependency. "
+                    + "Check if you forgot to create the folder under '%s' "
+                    + "or have mistyped the theme or folder name for '%s'.";
+            throw new ExecutionFailedException(String.format(errorMessage,
+                    themeName, mainCustomThemeDir.getPath(), themeName));
+        }
```
caalador, line 143:
```suggestion
        
        if (existingAppThemeDirectories.isEmpty()) {
            String errorMessage = "Discovered @Theme annotation with theme "
                + "name '%s', but could not find the theme directory "
                + "in the project or available as a jar dependency. "
                + "Check if you forgot to create the folder under '%s' "
                + "or have mistyped the theme or folder name for '%s'.";
            throw new ExecutionFailedException(String
                .format(errorMessage, themeName,
                    new File(frontendDirectory, APPLICATION_THEME_ROOT)
                        .getPath(), themeName));
        }
        if (existingAppThemeDirectories.size() >= 2) {

            boolean appThemeFoundInClassPath = existingAppThemeDirectories
                .stream().map(File::getPath)
                .anyMatch(path -> path.contains(FrontendUtils.FLOW_NPM_PACKAGE_NAME));

            if (appThemeFoundInClassPath) {
                String errorMessage = "Theme '%s' should not exist inside a "
                    + "jar and in the project at the same time.%n"
                    + "Extending another theme is possible by adding "
                    + "{ \"parent\": \"your-parent-theme\" } entry to the "
                    + "'theme.json' file inside your theme folder.";
                throw new ExecutionFailedException(
                    String.format(errorMessage, themeName));
            } else {
                String errorMessage = "Discovered Theme folder for theme '%s' "
                    + "in more than one place in the project. Please "
                    + "make sure there is only one theme folder with name '%s' "
                    + "exists in the your project. "
                    + "The recommended place to put the theme folder inside "
                    + "the project is '%s'";
                throw new ExecutionFailedException(String
                    .format(errorMessage, themeName, themeName,
                        new File(frontendDirectory, APPLICATION_THEME_ROOT)
                            .getPath()));
            }
        }
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -70,4 +84,77 @@ public void execute() throws ExecutionFailedException {
                 "Unable to write theme import file", e);
         }
     }
+
+    private void verifyThemeDirectoryExistence()
+            throws ExecutionFailedException {
+
+        String themeName = theme.getName();
+        String themePath = String.join("/", APPLICATION_THEME_ROOT, themeName);
+        File mainCustomThemeDir = new File(frontendDirectory, themePath);
```
caalador, line 93:
```suggestion
```

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -70,4 +84,77 @@ public void execute() throws ExecutionFailedException {
...
+                                    .getPath(), themeName));
+        }
+        if (existingAppThemeDirectories.size() >= 2) {
+
+            boolean appThemeFoundInClassPath = existingAppThemeDirectories
+                    .stream().map(File::getPath)
+                    .anyMatch(path -> path
+                            .contains(FrontendUtils.FLOW_NPM_PACKAGE_NAME));
+
+            if (appThemeFoundInClassPath) {
```
caalador, line 119: Minor nit. Now that I read this like this appThemeFoundInClasPath is not perhaps the most explanatory name. Could be `jarThemeFound` as that's what it checks, also it's basically not on the classpath what we check.

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
taefi: Agree, renamed it `themeFoundInJar` to be read as a natural sentence in `if`

caalador (approved):

PR #9686, opened 2020-12-17
Fix: add the missing otf font extension.

### LINKED ISSUE #9683: Fix and test including font as .otf
Currently we're only testing that `.tff` works but including `.otf` fails with:
```
 ERROR in ./themes/not-lumo/document.css
 Module build failed (from ../node_modules/.pnpm/registry.npmjs.org/extract-loader/5.1.0/node_modules/extract-loader/lib/extractLoader.js):
 ModuleParseError: Module parse failed: Unexpected character '' (1:4)
```
Reproducible by adding `IBMPlexMono-Regular.otf` & `@font-face 'IBM Plex Mono'` into `document.css`

[IBM_Plex_Sans.zip](https://github.com/vaadin/flow/files/5709234/IBM_Plex_Sans.zip)

caalador:
The file-loader didn't handle otf extensions so having a otf font-face
failed the build.
Fixes #9683


pleku (approved):

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -202,7 +202,7 @@ module.exports = {
       },
       {
         // File-loader only copies files used as imports in .js files or handled by css-loader
-        test: /\.(png|gif|jpg|jpeg|svg|eot|woff|woff2|ttf)$/,
+        test: /\.(png|gif|jpg|jpeg|svg|eot|woff|woff2|otf|ttf)$/,
         use: [{
```
knoobie, line 206: Isn't ico missing as well?

pleku: Is that even used ever for web ? At a quick glance it seems Microsoft specific

knoobie: Wasn't the vaadin favicon an .ico file in the old framework versions? I'm so used to that file ending kinda, that's why I'm asking :)

caalador: Favicon is still often an `ico` file, but it is set as a link with the `rel=icon"` attaribute so it doesn't go through css.
@rolfsmeds how common would it be to use a .ico in css?

rolfsmeds: Extremely rare. Safe to ignore I'd say..

knoobie: Sorry for the noise, ignoring is 👍 - I missed the obvious comment above the list of file endings.. 🤦‍♂️

PR #9694, opened 2020-12-18
Update StreamReceiverHandler.java

xdenser:
progress events in upload component are almost not called ( I get only 2 events in 500Mb upload), this change fixes the problem for me.


caalador (approved):
This is a good catch as we currently never send an event if the time taken for the loop execution doesn't take more than 500ms.

mshabarov:
Cherry picked to `2.4`

PR #9698, opened 2020-12-20
make IndexHtmlRequesthandler overrideable

stbischof:
This would help to extend IndexHtmlRequesthandler  and cleanup IndexHtmlHolder. Especially when you want to load the index file from an other Location then the default one it would help to override just the part where the Frontendutils is used. You also need easyly acces to the request which gives you more options then the service.

Especially if you have not an
One bundle,one servlet,one context one service scenario.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
@@ -19,6 +19,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.io.UncheckedIOException;
+import java.util.function.Supplier;
 
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.DataNode;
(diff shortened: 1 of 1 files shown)


Legioth:
Thank you for your contribution. Before we can consider taking this in, it would be useful to better understand the use case: what is the purpose of making `IndexHtmlRequesthandler` overrideable?

stbischof:
This would help to extend IndexHtmlRequesthandler and cleanup IndexHtmlHolder. Especially when you want to load the index file from an other Location then the default one it would help to override just the part where the Frontendutils is used. You also need easyly acces to the request which gives you more options then the service.

Especially if you have not an
One bundle,one servlet,one context one service scenario.

Legioth:
Any other use cases than allowing custom logic to load the index file?

If that's the only use case, then it should be considered to add a feature specifically for that instead of changing visibility to `protected` for methods that have never been designed to be used in that way (and that aren't documented or tested as such).

pleku:
I agree with @Legioth. Just changing private methods to be protected would not be enough. Protected methods that are extendable are considered public framework API that needs javadocs and should be tested too to avoid anyone breaking or even removing "the feature" accidentally.

I would suggest creating an issue first which describes the use case(s) and optionally proposes a suitable solution that can be discussed before jumping to the implementation part.

pleku:
As this PR has been waiting for a month for the author to reply and because it is not a priority for the team, I'm closing the PR.

For discussion on adding a feature for customizing the indexhtmlrequesthandler, please open an issue first so design may be considered before submitting a PR. Thank you.

PR #9699, opened 2020-12-20
feat: use superclass of VadinServlet/Request if possible

stbischof:
I cases you are not using the `VaadinServletService` because you ship your own own (Servlet-based) `VaadinService`.

Then you are, at the moment, not able to use fundamental classes like `VaadinServlet`. This is because they use your VaadinServletService. But you are not using any methods of VaadinServletService because VaadinService is enough. So I suggest to use superclasses if possible.

`VaadinServletRequest`needs a `VaadinServletService`

```
    @Override
    public VaadinServletService getService() {
        return vaadinService;
    }
```
If you have your own VaadinService implementation you can't rely on VaadinServletRequest
ant an the moment you cant use a lot of classes like `VaadinServlet`, `PushRequestHandler`, `StreamResourceHandler`, `WebComponentBootstrapHandler`

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
@@ -1706,9 +1706,10 @@ public abstract class VaadinService implements Serializable {
             SystemMessages systemMessages = getSystemMessages(
                     HandlerHelper.findLocale(null, request), request);
             String sessionExpiredURL = systemMessages.getSessionExpiredURL();
+
             if (sessionExpiredURL != null
-                    && (response instanceof VaadinServletResponse)) {
-                ((VaadinServletResponse) response)
+                    && (response instanceof HttpServletResponse)) {
+                ((HttpServletResponse) response)
                         .sendRedirect(sessionExpiredURL);
             } else {
                 /*
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
@@ -1707,7 +1707,12 @@ public abstract class VaadinService implements Serializable {
                     HandlerHelper.findLocale(null, request), request);
             String sessionExpiredURL = systemMessages.getSessionExpiredURL();
 
+            
             if (sessionExpiredURL != null
+                    && (response instanceof VaadinServletResponse)) {
+                ((VaadinServletResponse) response)
+                        .sendRedirect(sessionExpiredURL);
+            } else if (sessionExpiredURL != null
                     && (response instanceof HttpServletResponse)) {
                 ((HttpServletResponse) response)
                         .sendRedirect(sessionExpiredURL);
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
@@ -52,7 +52,7 @@ import com.vaadin.flow.shared.JsonConstants;
  * @since 1.0
  */
 public class VaadinServlet extends HttpServlet {
-    private VaadinServletService servletService;
+    private VaadinService vaadinService;
     private StaticFileHandler staticFileHandler;
 
     private volatile boolean isServletInitialized;
@@ -89,12 +89,12 @@ public class VaadinServlet extends HttpServlet {
              * its "init" method is called from the {@code
              * ServletContextListener} with the same ServletConfig instance.
              */
-            VaadinServletContext vaadinServletContext = null;
+            VaadinContext vaadinContext = null;
             if (getServletConfig() == null) {
                 isServletInitialized = true;
                 super.init(servletConfig);
 
-                vaadinServletContext = initializeContext();
+                vaadinContext = initializeContext();
             }
 
             if (getServletConfig() != servletConfig) {
@@ -103,18 +103,18 @@ public class VaadinServlet extends HttpServlet {
                                 + "instance which has been used for the initial method call");
             }
 
-            if (vaadinServletContext == null) {
-                vaadinServletContext = new VaadinServletContext(
+            if (vaadinContext == null) {
+                vaadinContext = new VaadinServletContext(
                         getServletConfig().getServletContext());
             }
 
-            if (servletService != null || vaadinServletContext
+            if (vaadinService != null || vaadinContext
                     .getAttribute(Lookup.class) == null) {
                 return;
             }
 
             try {
-                servletService = createServletService();
+                vaadinService = createServletService();
             } catch (ServiceException e) {
                 throw new ServletException("Could not initialize VaadinServlet",
                         e);
@@ -122,9 +122,9 @@ public class VaadinServlet extends HttpServlet {
 
             // Sets current service as it is needed in static file server even
             // though there are no request and response.
-            servletService.setCurrentInstances(null, null);
+            vaadinService.setCurrentInstances(null, null);
 
-            staticFileHandler = createStaticFileHandler(servletService);
+            staticFileHandler = createStaticFileHandler(vaadinService);
 
             servletInitialized();
         } finally {
@@ -145,13 +145,13 @@ public class VaadinServlet extends HttpServlet {
      * to find and serve static resources. By default it returns a
      * {@link StaticFileServer} instance.
      *
-     * @param servletService
-     *            the servlet service created at {@link #createServletService()}
+     * @param vaadinService
+     *            the vaadinService created at {@link #createServletService()}
      * @return the file server to be used by this servlet, not <code>null</code>
      */
     protected StaticFileHandler createStaticFileHandler(
-            VaadinServletService servletService) {
-        return new StaticFileServer(servletService);
+            VaadinService vaadinService) {
+        return new StaticFileServer(vaadinService);
     }
 
     protected void servletInitialized() throws ServletException {
@@ -287,8 +287,8 @@ public class VaadinServlet extends HttpServlet {
 
         CurrentInstance.clearAll();
 
-        VaadinServletRequest vaadinRequest = createVaadinRequest(request);
-        VaadinServletResponse vaadinResponse = createVaadinResponse(response);
+        VaadinRequest vaadinRequest = createVaadinRequest(request);
+        VaadinResponse vaadinResponse = createVaadinResponse(response);
         if (!ensureCookiesEnabled(vaadinRequest, vaadinResponse)) {
             return;
         }
@@ -419,7 +419,7 @@ public class VaadinServlet extends HttpServlet {
         }
     }
 
-    private VaadinServletResponse createVaadinResponse(
+    private VaadinResponse createVaadinResponse(
             HttpServletResponse response) {
         return new VaadinServletResponse(response, getService());
     }
@@ -442,8 +442,8 @@ public class VaadinServlet extends HttpServlet {
      *
      * @return the Vaadin service
      */
-    public VaadinServletService getService() {
-        return servletService;
+    public VaadinService getService() {
+        return vaadinService;
     }
 
     /**
@@ -457,13 +457,13 @@ public class VaadinServlet extends HttpServlet {
      * @return false if cookies are disabled, true otherwise
      * @throws IOException
      */
-    private boolean ensureCookiesEnabled(VaadinServletRequest request,
-            VaadinServletResponse response) throws IOException {
+    private boolean ensureCookiesEnabled(VaadinRequest request,
+            VaadinResponse response) throws IOException {
         if (HandlerHelper.isRequestType(request, RequestType.UIDL)) {
             // In all other but the first UIDL request a cookie should be
             // returned by the browser.
             // This can be removed if cookieless mode (#3228) is supported
-            if (request.getRequestedSessionId() == null) {
+            if (request instanceof HttpServletRequest && ((HttpServletRequest) request).getRequestedSessionId() == null) {
                 // User has cookies disabled
                 SystemMessages systemMessages = getService().getSystemMessages(
                         HandlerHelper.findLocale(null, request), request);
@@ -535,7 +535,7 @@ public class VaadinServlet extends HttpServlet {
         }
     }
 
-    private VaadinServletContext initializeContext() {
+    private VaadinContext initializeContext() {
         ServletContext servletContext = getServletConfig().getServletContext();
         VaadinServletContext vaadinServletContext = new VaadinServletContext(
                 servletContext);
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServletRequest.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServletRequest.java
@@ -32,7 +32,7 @@ import javax.servlet.http.HttpSession;
 public class VaadinServletRequest extends HttpServletRequestWrapper
         implements VaadinRequest {
 
-    private final VaadinServletService vaadinService;
+    private final VaadinService vaadinService;
 
     /**
      * Wraps a http servlet request and associates with a vaadin service.
@@ -43,7 +43,7 @@ public class VaadinServletRequest extends HttpServletRequestWrapper
      *            the associated vaadin service
      */
     public VaadinServletRequest(HttpServletRequest request,
-            VaadinServletService vaadinService) {
+            VaadinService vaadinService) {
         super(request);
         this.vaadinService = vaadinService;
     }
@@ -73,7 +73,7 @@ public class VaadinServletRequest extends HttpServletRequestWrapper
     }
 
     @Override
-    public VaadinServletService getService() {
+    public VaadinService getService() {
         return vaadinService;
     }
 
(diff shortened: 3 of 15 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -51,7 +51,7 @@
  * @since 1.0
  */
 public class VaadinServlet extends HttpServlet {
-    private VaadinServletService servletService;
+    private VaadinService servletService;
```
pleku, line 54: Since this is _VaadinServlet_ it means that there should be a servlet. I don't see a point in doing any changes for this class for "no servlet present". Why are there changes done here ?

flow-server/src/main/java/com/vaadin/flow/server/VaadinServletRequest.java
```diff
@@ -32,7 +32,7 @@
 public class VaadinServletRequest extends HttpServletRequestWrapper
         implements VaadinRequest {
 
-    private final VaadinServletService vaadinService;
+    private final VaadinService vaadinService;
```
pleku, line 35: This is a _servlet_ request, so why shouldn't it have access to servlet service ?

flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
```diff
@@ -285,14 +285,14 @@ private InputStream getResourceInServletContextAsStream(String path) {
...
-        assert request instanceof VaadinServletRequest;
+        assert request instanceof HttpServletRequest;
         // Generate location from the request by finding how many "../" should
         // be added to the servlet path before we get to the context root
 
         // Should not take pathinfo into account because the base URI refers to
         // the servlet path
 
-        String servletPath = ((VaadinServletRequest) request).getServletPath();
+        String servletPath = ((HttpServletRequest) request).getServletPath();
```
pleku, line 295: Again, servlet based class and these changes seem like a waste of time, or why would these be needed ?

flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
```diff
@@ -239,7 +255,7 @@ public boolean handleRequest(VaadinSession session, VaadinRequest request,
             return false;
         }
 
-        if (request instanceof VaadinServletRequest) {
+        if (request instanceof HttpServletRequest) {
```
pleku, line 258: When would it be a case here that `VaadinServletRequest` is not applicable here ??

flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
```diff
@@ -248,16 +264,16 @@ public boolean handleRequest(VaadinSession session, VaadinRequest request,
             try {
                 atmosphere.doCometSupport(
                         AtmosphereRequestImpl
-                                .wrap((VaadinServletRequest) request),
+                                .wrap((HttpServletRequest) request),
```
pleku, line 267: When would it be a case here that `VaadinServletRequest` is not applicable here ??

flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
```diff
@@ -248,16 +264,16 @@ public boolean handleRequest(VaadinSession session, VaadinRequest request,
...
                         AtmosphereResponseImpl
-                                .wrap((VaadinServletResponse) response));
+                                .wrap((HttpServletResponse) response));
```
pleku, line 269: When would it be a case here that `VaadinServletRequest` is not applicable here ??

flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
```diff
@@ -248,16 +264,16 @@ public boolean handleRequest(VaadinSession session, VaadinRequest request,
...
             } catch (ServletException e) {
                 // TODO PUSH decide how to handle
                 throw new RuntimeException(e);
             }
         } else {
             throw new IllegalArgumentException(
-                    "Only VaadinServletRequests are supported");
+                    "Only HttpServletRequest are supported");
```
pleku, line 276: When would it be a case here that `VaadinServletRequest` is not applicable here ??

flow-server/src/main/java/com/vaadin/flow/server/communication/StreamResourceHandler.java
```diff
@@ -60,7 +61,7 @@ public void handleRequest(VaadinSession session, VaadinRequest request,
         StreamResourceWriter writer;
         session.lock();
         try {
-            ServletContext context = ((VaadinServletRequest) request)
+            ServletContext context = ((ServletRequest) request)
```
pleku, line 64: When would it be a case here that `VaadinServletRequest` is not applicable here ??

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -136,7 +138,7 @@ protected boolean canHandleRequest(VaadinRequest request) {
      * @return Request's url.
      */
     protected String getRequestUrl(VaadinRequest request) {
-        return ((VaadinServletRequest) request).getRequestURL().toString();
+        return ((HttpServletRequest) request).getRequestURL().toString();
```
pleku, line 141: When would it be a case here that `VaadinServletRequest` is not applicable here ??

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -434,7 +436,7 @@ protected String getServiceUrl(VaadinRequest request,
         String url = request.getParameter(REQ_PARAM_URL);
         // if 'url' parameter was not available, use request url
         if (url == null) {
-            url = ((VaadinServletRequest) request).getRequestURL().toString();
+            url = ((HttpServletRequestWrapper) request).getRequestURL().toString();
```
pleku, line 439: When would it be a case here that `VaadinServletRequest` is not applicable here ??

pleku (changes requested):
Thanks for the contribution and sorry for the long delay with the review, but this PR basically shows how reluctant people are to comment on PRs that don't explain _why_ changes are proposed. There is no issue open that would explain why this is needed so you need to start with that.

Secondly many of the changes seem like invalid and, sorry for being direct but, a waste of time. Please help me understand why this would be valuable. If not, I think we'll just have to say no thanks this time - I hope you understand.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
stbischof: I do not use the `VaadinServletService` i am using my own `VaadinServic`.If I want to use your `VaadinServlet.class` i have no way to do this because it must be your `VaadinServletService`.

But you are not using any methods of VaadinServletService because VaadinService is enough.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServletRequest.java
stbischof: because then you cant use it with other VasdinService Implementations then yours

flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
stbischof: I fixed the description.

flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
stbischof: I fixed the description.

flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
stbischof: I fixed the description.

flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
stbischof: I fixed the description.

flow-server/src/main/java/com/vaadin/flow/server/communication/StreamResourceHandler.java
stbischof: I fixed the description.

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
stbischof: I fixed the description.

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
stbischof: here it is just a cleanup because you do not need a `VaadinServletRequest`. The `HttpServletRequestWrapper` is enough to `.getRequestURL()`. And others that do not use the `VaadinServletRequest`will not end with a ClassCastException

flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
stbischof: when it is a HttpServletRequest but not a VaadinServletRequest

pleku (commented):
If we we're to _allow having your own `VaadinService`_ then it would be good if there was some sort of test in the project that would make it so that this feature is not accidentally broken.

Now that the reasoning behind the PR is clear, the next question would be about the added value and use case - why do you want to use your own VaadinService ?

(edit: this type of discussions would be better to have in an issue before starting the implementation)

stbischof:
Because of your OSGi approach is not the way how I would expect to handle OSGi and Vaadin. It is kind of "Works with OSGi" and will fit most of people. But really working with `Services` is in a deep way would look other.

And your OSGi bundles also have the wrong license.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -520,7 +520,7 @@ public void destroy() {
         getService().destroy();
     }
 
-    private VaadinServletContext initializeContext() {
+    private VaadinContext initializeContext() {
```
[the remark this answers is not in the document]
stbischof: sure. will change this back.

pleku:
> Because of your OSGi approach is not the way how I would expect to handle OSGi and Vaadin. It is kind of "Works with OSGi" and will fit most of people. But really working with Services is in a deep way would look other.

I take it you mean with this that _there would be a better way for this to work by utilizing declarative services in OSGi_. I think this is a direction that we would want to take too and would gladly accept those contributions, but this direction should be taken by first discussing and agreeing on the steps in issues before jumping into implementation - it will be more smooth that way as then we know what is incoming and are more likely to help you get it in instead of need to start by questioning that "why is this code needed". From our perspective, we have to start that way as we're promising to maintain the code for 15 years to users who need that.

> And your OSGi bundles also have the wrong license.

Just to be sure - do you mean the commercial license in vaadin/osgi or something else ?

> If you allow others to having an own VaadinService i will add test that force and check ClassCastExceptions.

I don't have any objections to this now that I understand the goal. Though, I would rather work together to improve our OSGi support for all instead of enabling people to build their own support. We're gladly willing to give out licenses to those who contribute to it.

stbischof:
I broken all my contribution in the moment where i had seen that you moved the osgi part to commercial license in vaadin/osgi.
Also parts that had been under the normal license. 

Because this was just an private interest a commercial license would not be an option for me.

pleku:
To clarify to myself for later times, this is still waiting for a test to be added that would prevent breaking the "feature" that easily.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -1706,9 +1706,10 @@ protected void handleSessionExpired(VaadinRequest request,
             SystemMessages systemMessages = getSystemMessages(
                     HandlerHelper.findLocale(null, request), request);
             String sessionExpiredURL = systemMessages.getSessionExpiredURL();
+
             if (sessionExpiredURL != null
-                    && (response instanceof VaadinServletResponse)) {
-                ((VaadinServletResponse) response)
+                    && (response instanceof HttpServletResponse)) {
```
knoobie, line 1711: Overall I like the idea behind this PR, I just have a problem with the explicit downcast from `VaadinServletResponse` to `HttpServletResponse` with the upcoming jakarta namespace change in mind. Totally out of scope, just my 2 cents.

stbischof: New interfaces that extends response and provides the Methods may be a way to solve this

stbischof:
i added the wished tests: `OtherImplementationsTest.java`

pleku:
The build failed due to unrelated build agent issues, I'll retrigger it tomorrow and take a look at the test.

stbischof:
Could you please retrigger it?

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -457,13 +457,13 @@ public VaadinServletService getService() {
...
-    private boolean ensureCookiesEnabled(VaadinServletRequest request,
-            VaadinServletResponse response) throws IOException {
+    private boolean ensureCookiesEnabled(VaadinRequest request,
+            VaadinResponse response) throws IOException {
         if (HandlerHelper.isRequestType(request, RequestType.UIDL)) {
             // In all other but the first UIDL request a cookie should be
             // returned by the browser.
             // This can be removed if cookieless mode (#3228) is supported
-            if (request.getRequestedSessionId() == null) {
+            if (request instanceof HttpServletRequest&&((HttpServletRequest)request).getRequestedSessionId() == null) {
```
knoobie, line 466:
```suggestion
            if (request instanceof HttpServletRequest && ((HttpServletRequest) request).getRequestedSessionId() == null) {
```

@stbischof This could help with your rebuild as well ;) and makes the code a little bit easier to read.

stbischof:
forced push to trigger rerun of tests, (last one also did not finished

flow-server/src/test/java/com/vaadin/flow/server/OtherImplementationsTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import com.vaadin.flow.server.communication.PushRequestHandler;
+
+public class OtherImplementationsTest {
+    @Test
+    public void StaticFileServer_Constructor_uses_VadinService()
```
pleku, line 21: non-blocking nit: typoed Vaadin here and c/p the error to other method names too ...

flow-server/src/test/java/com/vaadin/flow/server/OtherImplementationsTest.java
```diff
@@ -0,0 +1,205 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import com.vaadin.flow.server.communication.PushRequestHandler;
+
+public class OtherImplementationsTest {
```
pleku, line 19: I get the test, but at a glance probably most developers will not. Please change the name to `CustomVaadinServiceImplementationTest` and maybe add a comment line "Makes sure that a custom vaadin service that is not vaadin servlet service can be used in when desired".

pleku (changes requested):
Sorry for the long delay, I was on winter vacation last week ⛷️ 

Only couple of small comments and then this would be 🟢

flow-server/src/test/java/com/vaadin/flow/server/OtherImplementationsTest.java
stbischof: Done, Thx

flow-server/src/test/java/com/vaadin/flow/server/OtherImplementationsTest.java
stbischof: changed into Vaadin

pleku (approved):

Legioth:
This change causes massive backwards binary compatibility issues with e.g. add-ons compiled against older versions of Vaadin. This happens because method invocations in bytecode are tied to the method return type and not only the method name and parameter types.

I would suggest reverting to avoid those issues.

Furthermore,  I don't even understand the whole point of making this change. If you want to use a custom service implementation together with `VaadinServlet`, then why wouldn't you extend `VaadinServletService` instead of the generic `VaadinService` class?

Legioth:
To further elaborate on my previous message, the point is that `VaadinServlet` and `VaadinServletService` should really be seen as a single unit that just happens to be split up into two separate classes because Java doesn't offer multiple inheritance.

They're not supposed to be separated from each other. A proper separation would require more thorough architectural considerations rather than only tweaking some return types that happened to get in the way when trying to do something that you're not supposed to even try.

pleku:
I don't know why the bot has added the label `Released with Platform 19.0.0` as it is not true - @ZheSun88 ?

ZheSun88:
@pleku somehow the bot caught flow 7.0.0.alpha1 release note for platform 19.0.0 release. i will check other tickets in that release.

PR #9706, opened 2020-12-21
refactor!: expose connectionIndicator as constant instead of via getter

### LINKED ISSUE #9671: Expose connectionIndicator as constant instead of getter function
Instead of accessing the connection via a getter, it should be exposed as a constant. So as a developer, I could configure it as follows in `index.ts`:

```
import {connectionIndicator} from '@vaadin/flow-frontend';
// ...
connectionIndicator.setPosition(TOP);
connectionIndicator.setPosition(BOTTOM);
```

joheriks:
Fixes #9671

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/Connect.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/Connect.ts
@@ -1,5 +1,5 @@
 /* tslint:disable:max-classes-per-file */
-import { getConnectionIndicator } from './ConnectionIndicator';
+import { ConnectionIndicator } from './ConnectionIndicator';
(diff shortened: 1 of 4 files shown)


haijian-vaadin (approved):
The code looks good to me. 
Just a comment regarding the commit message, would `refactor` be better than `fix` for the code change?

joheriks:
> The code looks good to me.
> Just a comment regarding the commit message, would `refactor` be better than `fix` for the code change?

Updated. Also fixed relying on accessor for side effect (adding indicator to DOM), as this could be optimized away (and so happened in `test-root-context` ITs).

haijian-vaadin (approved):

Issue #9594, opened 2020-12-08
One should not have to restart app when adding a new css file to app theme


pleku:
Currently this is needed when adding new files - we should just make webpack detect new files out-of-the-box so no restart is needed for that.

### Acceptance Criteria: 

To have test(s) in the live-reload module to cover the following scenarios:
- Adding a new CSS file
- Adding a static resource like a Font and address it in CSS files 
- Support for deleting an already existing theme file

Related issue: https://github.com/vaadin/flow/issues/9596

caalador:
To be noted for the adding of new files:
https://github.com/webpack/webpack-dev-server/issues/1006#issuecomment-320974577

PR #9708, opened 2020-12-21
fix: recompile Webpack when adding a new CSS file to app theme

mshabarov:
Adding a new file to `frontend/themes/my-theme` now triggers the `webpack` recompilation and updates the application.

Fixes: https://github.com/vaadin/flow/issues/9594


flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -90,6 +93,36 @@ class ApplicationThemePlugin {
         logger.debug("Skipping Vaadin application theme handling.");
         logger.trace("Most likely no @Theme annotation for application or only themeClass used.");
       }
+    };
+
+    compiler.hooks.afterEnvironment.tap("ApplicationThemePlugin", handleThemesHook);
+
+    // Adds the active application theme folder for webpack watching
+    compiler.plugin("after-compile", function (compilation, callback) {
```
caalador, line 101: This plugin should only be set in devMode

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -90,6 +93,36 @@ class ApplicationThemePlugin {
...
+      if (themeName) {
+        compilation.contextDependencies.add('frontend/themes/' + themeName);
+      }
+      callback();
+    });
+
+    // Adds a hook for theme files change event
+    compiler.plugin("watch-run", (compilation, callback) => {
```
caalador, line 109: Also this plugin should only run in dev mode.

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -127,5 +160,9 @@ function getThemeProperties(themeFolder) {
   if (!fs.existsSync(themePropertyFile)) {
     return {};
   }
-  return JSON.parse(fs.readFileSync(themePropertyFile));
+  const themePropertyFileAsString = fs.readFileSync(themePropertyFile);
+  if (themePropertyFileAsString.length === 0) {
+    return {};
+  }
+  return JSON.parse(themePropertyFileAsString);
```
caalador, line 167: Strange that it doesn't parse empty string.
A cleaner solution is `JSON.parse(fs.readFileSync(themePropertyFile)||'{}')`

caalador (changes requested):

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -51,12 +50,16 @@ class ApplicationThemePlugin {
   apply(compiler) {
     logger = compiler.getInfrastructureLogger("ApplicationThemePlugin");
 
-    compiler.hooks.afterEnvironment.tap("ApplicationThemePlugin", () => {
+    const handleThemesHook = () => {
       const generatedThemeFile = path.resolve(this.options.themeResourceFolder, "theme-generated.js");
       if (fs.existsSync(generatedThemeFile)) {
+        // matches theme folder name in 'themes/my-theme/my-theme.generated.js'
+        const nameRegex = /themes\/(.*)\/\1.generated.js/g;
```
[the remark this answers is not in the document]
mshabarov: Because `nameRegex.exec` return `null` on a second run, so I need to reset it. https://stackoverflow.com/questions/4724701/regexp-exec-returns-null-sporadically

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
mshabarov: Yep, this looks more clean, but the "truthy" value check of `fs.readFileSync(themePropertyFile)` does not work for some reason. It still gives: 
```javascript
return JSON.parse(fs.readFileSync(themePropertyFile)||'{}');
              ^
SyntaxError: Unexpected end of JSON input
```

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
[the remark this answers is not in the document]
caalador: That's because of the global status just remove the `g` from the end.

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
caalador: For this there also exists npmjs.com/package/extra-watch-webpack-plugin

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -51,12 +50,16 @@ class ApplicationThemePlugin {
   apply(compiler) {
     logger = compiler.getInfrastructureLogger("ApplicationThemePlugin");
 
-    compiler.hooks.afterEnvironment.tap("ApplicationThemePlugin", () => {
+    const handleThemesHook = () => {
```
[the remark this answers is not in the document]
caalador: Perhaps even move this out to its own file so it's easier to reuse.

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -51,53 +53,90 @@ class ApplicationThemePlugin {
...
 
-        // read theme name from the theme-generated.js as there we always mark the used theme for webpack to handle.
-        const themeName = nameRegex.exec(fs.readFileSync(generatedThemeFile, {encoding: 'utf8'}))[1];
-        if (!themeName) {
-          throw new Error("Couldn't parse theme name from '" + generatedThemeFile + "'.");
-        }
+    // Adds the active application theme folder for webpack watching
+    compiler.plugin("after-compile", function (compilation, callback) {
+      if (themeName) {
+        compilation.contextDependencies.add('frontend/themes/' + themeName);
```
caalador, line 61: This might be wrong as the folder might also be under `src/...` though editing anything there would restart the server due to changes in classpath.

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -48,127 +42,28 @@ class ApplicationThemePlugin {
     if (!this.options.themeProjectFolders) {
       throw new Error("Missing themeProjectFolders path array");
     }
+
+    const generatedThemeFile = getResolvedGeneratedThemeFile(this.options);
+    themeName = extractThemeName(generatedThemeFile);
+    generateThemeCallback = (logger) => generateTheme(this.options, logger);
```
caalador, line 48: These can and probably should be called on request so that `getThemeName` calls `extractThemeName` and `getThemeCallback` just creates the lambda when requested perhaps even using the callers logger instead of the appThemeLogger.

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
```diff
@@ -0,0 +1,62 @@
...
+          if (!themeGeneratedFileChanged) {
+            this.generateThemeCallbackGetter()(logger);
+          }
+        }
+        callback();
+      });
+    }
+}
+
+module.exports = ThemeLiveReloadPlugin;
```
caalador, line 62: Missing new line.

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
```diff
@@ -0,0 +1,62 @@
...
+        throw new Error("Missing theme name");
+      }
+      this.themeNameGetter = themeNameGetter;
+      this.generateThemeCallbackGetter = generateThemeCallbackGetter;
+    }
+
+    apply(compiler) {
+      // Adds a hook for theme files change event
+      compiler.plugin("watch-run", (compilation, callback) => {
+        logger = compiler.getInfrastructureLogger("ThemeLiveReloadPlugin");
```
caalador, line 42: insignificant: As the logger is not used outside of the watchRun hook then it doesn't need to be stored outside.

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
```diff
@@ -0,0 +1,62 @@
...
+      if (!themeNameGetter) {
+        throw new Error("Missing theme name");
+      }
+      this.themeNameGetter = themeNameGetter;
+      this.generateThemeCallbackGetter = generateThemeCallbackGetter;
+    }
+
+    apply(compiler) {
+      // Adds a hook for theme files change event
+      compiler.plugin("watch-run", (compilation, callback) => {
```
caalador, line 41:
```suggestion
      compiler.hooks.watchRun.tapAsync("ThemeLiveReloadPlugin", (compilation, callback) => {
```

caalador (changes requested):

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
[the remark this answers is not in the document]
mshabarov: Moved to it's own file

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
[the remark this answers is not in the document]
mshabarov: Done

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -90,6 +93,36 @@ class ApplicationThemePlugin {
...
+                themeGeneratedFileChanged = true;
+              }
+            });
+        logger.debug("Detected changes in the following files " + changedFiles);
+        if (!themeGeneratedFileChanged) {
+          handleThemesHook();
+        }
+      }
+      callback();
     });
```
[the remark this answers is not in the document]
mshabarov: Done, moved to its own plugin

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -81,7 +84,7 @@ class ApplicationThemePlugin {
         if (fs.existsSync(this.options.themeResourceFolder)) {
           if (themeFound && fs.existsSync(path.resolve(this.options.themeResourceFolder, themeName))) {
             throw new Error("Theme '" + themeName + "'should not exist inside a jar and in the project at the same time\n" +
-              "Extending another theme is possible by adding { \"parent\": \"my-parent-theme\" } entry to the theme.json file inside your theme folder.");
+                "Extending another theme is possible by adding { \"parent\": \"my-parent-theme\" } entry to the theme.json file inside your theme folder.");
```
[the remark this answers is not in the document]
mshabarov: Seems there are still some wrong indentation after moving to separate file. Will double check and fix

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
mshabarov: extra-watch-webpack-plugin is now used

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
mshabarov: done

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
mshabarov: Do you mean these folders `src/main/resources/static` and `src/main/resources/META-INF/resources/themes` whose can contains theme resources either? Well, if the live reload agent (or jetty, or webpack) has already them to be watched, then no need to add them. I will double check with starter project, if restart happens upon adding a new files there

caalador: Those yes, but I don't know if we need to support those with this.

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -238,6 +240,13 @@ module.exports = {
       projectStaticAssetsOutputFolder: projectStaticAssetsOutputFolder,
     }),
 
+    devMode && new ExtraWatchWebpackPlugin({
+      files: [],
+      dirs: [ 'frontend/themes/' + getThemeName() ]
+    }),
+
+    devMode && new ThemeLiveReloadPlugin(getThemeName, getGenerateThemeCallback),
```
caalador, line 248: One issue is that if we don't have a themeName then these shouldn't be added at all either.

mshabarov: Yep, assuming that the `getThemeName` should extract the theme name on request, I can create a `themeName` variable somewhere before these plugins array and just use it for extra watch and live reload plugins.

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -46,86 +42,28 @@ class ApplicationThemePlugin {
     if (!this.options.themeProjectFolders) {
       throw new Error("Missing themeProjectFolders path array");
     }
+
+    const generatedThemeFile = getResolvedGeneratedThemeFile(this.options);
+    themeName = extractThemeName(generatedThemeFile);
```
caalador, line 47: The build fails here as you do not check that the file exists in the first place.

```
[17:46:28]	[Step 5/11] Error: ENOENT: no such file or directory, open '/tmp/junit8056780026882223291/target/flow-frontend/themes/theme-generated.js'
[17:46:28]	[Step 5/11]     at Object.openSync (fs.js:448:3)
[17:46:28]	[Step 5/11]     at Object.readFileSync (fs.js:348:35)
[17:46:28]	[Step 5/11]     at extractThemeName (/tmp/junit8056780026882223291/node_modules/@vaadin/application-theme-plugin/theme-handle.js:130:28)
[17:46:28]	[Step 5/11]     at new ApplicationThemePlugin (/tmp/junit8056780026882223291/node_modules/@vaadin/application-theme-plugin/application-theme-plugin.js:47:17)
[17:46:28]	[Step 5/11]     at Object.<anonymous> (/tmp/junit8056780026882223291/webpack.generated.js:234:5)
```
```

File: flow-server/src/main/resources/webpack.generated.js (hunk 13 above)
Comment (caalador) [reply flow-server/src/main/resources/webpack.generated.js:248]: When we run webpack if the file will be created it will be available so it's fine to check and trust the result.

File: flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js (hunk 14 above)
Comment (mshabarov) [reply flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js:47]: Oh, thanks! I apparently didn't clean my `target` folder and it passes for my. So, you were 💯  right when requesting to make `getThemeName` to be on demand :)

Comment (mshabarov) [reply flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js:47]: Sorry, I've just got the meaning of this. Dev server (web pack) normally starts when the node tasks (including `theme-generated.js` creating) complete. Thus, the `theme-generated.js` should exist when the web pack starts (=> when the webpack.config.js is being processed), but not yet copied to `target/flow-frontend/themes`, right? Could I extract the theme name directly from `${project.root.dir}/node_modules/@vaadin/flow-frontend/themes`?

Comment (mshabarov) [reply flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js]: `theme-generated.js` already exists before webpack starts. Found the mistake in my code.

File: flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js [hunk 15]
```diff
@@ -112,31 +106,26 @@ function getThemeProperties(themeFolder) {
...
-/**
- * Generates theme resources meta data file name
- * @param themeName current theme name
- * @returns {string} theme resources meta data file name
- */
-function getThemeGeneratedFileName(themeName) {
-  return themeName + '.generated.js';
+function extractThemeName(themeResourceFolder) {
+  if (!themeResourceFolder) {
+    throw new Error('Theme resource folder cannot be empty');
```
pleku, line 151: So what happens if the user removes the `@Theme` annotation and deletes the `/themes/` folder ?

I'm not sure if this code is run in that case, but as I think that case will be probably for anyone adopting a existing project (any starter) they might want to delete the existing `/themes/` folder completely and go "vanilla style" instead without any theme.

pleku: And should the variable name be `themesFolder` or such ? Or actually I'm a bit confused that "what is theme resource folder" as the name is not IMO obvious

pleku: Or meh, maybe it is fine that we require them to restart the server at that point, but I think people will not know what does the error "Theme resource folder cannot be empty" mean and what to do after that ?

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
```diff
@@ -24,39 +22,42 @@ class ThemeLiveReloadPlugin {
...
-      if (!themeNameGetter) {
+    constructor(themeName, generateThemeCallback) {
+      if (!themeName) {
         throw new Error("Missing theme name");
       }
-      this.themeNameGetter = themeNameGetter;
-      this.generateThemeCallbackGetter = generateThemeCallbackGetter;
+      this.themeName = themeName;
+      if (!generateThemeCallback || typeof generateThemeCallback !== 'function') {
+        throw new Error("Missing theme generate callback or incorrect type");
```
pleku, line 35: Non-blocking: When could this error occur and is this error message supposed to be helpful ? If a user sees it, they have no idea what it means

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -13,7 +13,7 @@ const ExtraWatchWebpackPlugin = require('extra-watch-webpack-plugin');
 // Flow plugins
 const StatsPlugin = require('@vaadin/stats-plugin');
 const ThemeLiveReloadPlugin = require('@vaadin/theme-live-reload-plugin');
-const { ApplicationThemePlugin, getThemeName, getGenerateThemeCallback } = require('@vaadin/application-theme-plugin');
+const { ApplicationThemePlugin, generateTheme, extractThemeName } = require('@vaadin/application-theme-plugin');
```
pleku, line 16: Non-blocking: Seeing it here, I wonder if the `generateTheme` is a good name. How does it "generate the theme" ? The jsdocs say
```
 * Looks up for a theme resources in a current project and in jar dependencies,
 * copies the found resources and generates/updates meta data for webpack
 * compilation.
 ```
 so probably more suitable names would be `themeResourcesLookup` or `themeResourcesProcessor` or something like that, and same for the callback in the other file.

pleku (commented):
Only some comments. Probably @tanbt can finish the review once the build gets green as me and @caalador are away.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
mshabarov: > So what happens if the user removes the @Theme annotation and deletes the /themes/ folder ?
No, this code won't be executed in this case. 
Webpack won't trigger recompilation after delete of `/themes/`, no app reload, but the app keeps running with theme styles until you delete the `@Theme` annotation and triggers live reload. 
Then, the app theme will be deleted.
If you delete `/themes/` and then changes something else than `@Theme` and trigger live reload, then you get `Discovered @Theme annotation with theme name 'app-theme', but could not find the theme directory...` error message.

mshabarov: > Or meh, maybe it is fine that we require them to restart the server at that point, but I think people will not know what does the error "Theme resource folder cannot be empty" mean and what to do after that ?

`themeResourceFolder` refers to the folder where the `theme-generated.js` is generated. Basically, it's value is autogenerated in `webpack.generated.js` and if it's empty, then it's a bug in the flow. But since the `extractThemeName` can be potentially be reused, then it's good to change the message reflect the problem - could not extract theme name, because the path to folder with the autogenerated file `theme-generated.js` is not set, please set this path as a `application-theme-plugin` constructor parameter.

flow-server/src/main/resources/webpack.generated.js
mshabarov: Changed to `processThemeResources`

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
mshabarov: Replaced by more meaningful message.

tanbt (approved):

joheriks:
When/if adding more ITs in the module in the future, it should be remembered to add `@net.jcip.annotations.NotThreadSafe` or other sequencing mechanism since the tests modify server state and are run in parallel by default and on the CI.

mshabarov:
Thanks, @joheriks ! Here is the fix https://github.com/vaadin/flow/pull/9728

PR #9709, opened 2020-12-21
Process Iterable same way as Collection type

### LINKED ISSUE #9256: Fusion does not generate type definitions for return type Iterable
### Description of the bug

When a method in an `@Endpoint` annotated class returns an `Iterable<T>`, the corresponding TS method returns `Promise<any>`, and a definition for `T` is not created.

On the other hand, if the method returns `List<T>`, the TS method returns `Promise<Array<T>>` and a type definition is created for `T`.

```
@Endpoint
@AnonymousAllowed
public class VaadinEndpoint {

    public static class Foo {
        public String bar = "bar";
    }

    public Iterable<Foo> getFoos() {
        return Arrays.asList(new Foo(), new Foo());
    }
}
```

### Expected behavior

The generated `VaadinEndpoint.ts` should have the following definition

```
function _getFoos(): Promise<Array<Foo>> {
  return client.call('VaadinEndpoint', 'getFoos');
}
```

### Actual behavior

The generated `VaadinEndpoint.ts` has the following definition

```
function _getFoos(): Promise<any> {
  return client.call('VaadinEndpoint', 'getFoos');
}
```

### Versions:
    - Vaadin / Flow version: 17.0.7

tdq:
Extend SchemaResolver in the way to process Iterable type as Collection.
Fixes #9256


haijian-vaadin (changes requested):
Since TypeScript has type Iterable<>, would it better if can generate a TypeScript Iterable for Java Iterable?

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/generator/SchemaResolver.java
```diff
@@ -161,7 +161,7 @@ private boolean isNumberType(ResolvedType type) {
     }
 
     private boolean isCollectionType(ResolvedType type) {
-        return !type.isPrimitive() && isTypeOf(type, Collection.class);
+        return !type.isPrimitive() && (isTypeOf(type, Collection.class) || isTypeOf(type, Iterable.class));
```
fluorumlabs, line 164: `Collection` already extends `Iterable`

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/VaadinConnectControllerTest.java
```diff
@@ -50,6 +50,7 @@
 import com.vaadin.flow.server.connect.auth.VaadinConnectAccessChecker;
 import com.vaadin.flow.server.connect.exception.EndpointException;
 import com.vaadin.flow.server.connect.exception.EndpointValidationException;
+import com.vaadin.flow.server.connect.generator.endpoints.collectionendpoint.IterableEndpoint;
```
fluorumlabs, line 53: Looks like you forgot to commit `IterableEndpoint`

fluorumlabs (changes requested):

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/generator/SchemaResolver.java
tdq: Sure, so you think we can keep only Iterable here to allow such usage?
{code}
@Endpoint
@AnonymousAllowed
public class VaadinEndpoint {

    public static class Foo {
        public String bar = "bar";
    }

    public Iterable<Foo> getFoos() {
        return Arrays.asList(new Foo(), new Foo());
    }
}
{code}

platosha:
> Since TypeScript has type Iterable<>, would it better if can generate a TypeScript Iterable for Java Iterable?

For a regular (sync) Iterable means supporting sync `for...of` loop in TypeScript. That requires that the (async) result of the endpoint call would have all the items on the client side. In practice this means using an array under-the-hood anyway. Arrays are Iterable themselves, btw. I don’t think it’s worth to declare `Iterable` on what is in fact a regular array, it would make using that array harder for no benefit.

But, something potentially useful here is `AsyncIterable`. When used with `for await...of` iterators, as every iteration is async, this could enable requesting additional items from the endpoint on demand from the iterator, instead of everything at once. 

BTW Unlike `Collection<>`, `Iterable<>` does not contain a defined size, which means to me that for a given iterator there is no guarantee that it ever reaches an end. I wonder if it is fair that Fusion server tries to serialise all the items from a given iterable into a single JSON array response in Java. Would it rather be expected that items are requested from the iterator one-by-one upon iterating on the caller side, the client?

haijian-vaadin (approved):
Turns out using `Iterable` for TypeScript is not necessary.

fluorumlabs (approved):

PR #9718, opened 2020-12-23
Always install webpack plugins

caalador:
Plugin installation can be skipped
by defining `"update": false,` in installed
package.json of plugin under node_modules

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
@@ -120,14 +120,8 @@ public class TaskInstallWebpackPlugins implements FallibleCommand {
             String packageFile = FileUtils
                 .readFileToString(new File(pluginTargetFile, PACKAGE_JSON),
                     StandardCharsets.UTF_8);
-            final FrontendVersion packageVersion = new FrontendVersion(
-                Json.parse(packageFile).getString("version"));
-            FrontendVersion pluginVersion = new FrontendVersion(
-                packageJson.getString("version"));
-            if (packageVersion.isEqualTo(pluginVersion)) {
-                log().debug(
-                    "Skipping install of {} for version {} already installed",
-                    pluginName, pluginVersion.getFullVersion());
+            final JsonObject targetJson = Json.parse(packageFile);
+            if(targetJson.hasKey("update") && !targetJson.getBoolean("update")) {
                 return;
             }
         }


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
```diff
@@ -120,14 +120,8 @@ private void generatePluginFiles(String pluginName) throws IOException {
             String packageFile = FileUtils
                 .readFileToString(new File(pluginTargetFile, PACKAGE_JSON),
                     StandardCharsets.UTF_8);
-            final FrontendVersion packageVersion = new FrontendVersion(
-                Json.parse(packageFile).getString("version"));
-            FrontendVersion pluginVersion = new FrontendVersion(
-                packageJson.getString("version"));
-            if (packageVersion.isEqualTo(pluginVersion)) {
```
mshabarov, line 127: Why did you drop this check?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
```diff
@@ -120,14 +120,8 @@ private void generatePluginFiles(String pluginName) throws IOException {
...
-                log().debug(
-                    "Skipping install of {} for version {} already installed",
-                    pluginName, pluginVersion.getFullVersion());
+            final JsonObject targetJson = Json.parse(packageFile);
+            if(targetJson.hasKey("update") && !targetJson.getBoolean("update")) {
```
mshabarov, line 124: What is the use case when the developer wants to skip any webpack plugin?

caalador: Skipping getting the jar packaged plugin is really useful for when you develop the plugin in a project so you can make changes directly there without having the plugin overridden from the jar every time.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
caalador: it's dropped so that you don't get the situation that you had where the plugin was not the same as for the project.

mshabarov (dismissed):
Would be good to cover this feature by a unit test so that not to accidentally change or lost it.

pleku (commented):
So what this does is it leaves it up for the user to decide if they don't want to reinstall the plugins ?
And then it doesn't matter if we forget to update the version of the plugin when doing changes ?
A bit unsure why this is needed or better as there was no ticket or discussion on this.

A unit test would be nice to make sure nobody accidentally removes the "feature of skipping installing the plugins" but first we should probably agree that do we want to merge this or not.

caalador:
Seeing as the instance of not wanting to install the plugins is more for our development needs on the plugins and not for the end user. 
As more people have done development I've noted that due to the version check people have wasted time with wondering why the changes do not work or why they fail after they worked for a while when they have changed the flow version that writes an webpack js file that's uncompatible with the installed plugin (due to not updating plugin version on development time and having TC build a new snapshot that replaces the local snapshot).

Opting out of plugin installation makes it clearer that this is the version used and can be used for making changes before committing to build a new flow-server+maven-plugin instance.

pleku (approved):

Issue #9714, opened 2020-12-22
npm install fails because of npm 7


marcushellberg:
Vaadin 18.0.2, start.vaadin.com fusion project. 

The current pnpm setup doesn't support an easy way of adding npm dependencies (you have to install them with a very specific version of pnpm with `~/.vaadin/node_modules/pnpm/bin/pnpm.js i foo`), so I want to disable pnpm. 

I disabled pnpm in `application.properties` with `vaadin.pnpm.enable=false`

Running the app now fails on npm dependency resolution:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: no-name@undefined
npm ERR! Found: typescript@4.0.3
npm ERR! node_modules/typescript
npm ERR!   dev typescript@"4.0.3" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer typescript@"^2.7 || ^3" from awesome-typescript-loader@5.2.1
npm ERR! node_modules/awesome-typescript-loader
npm ERR!   dev awesome-typescript-loader@"5.2.1" from the root project
```

marcushellberg:
I suppose this has to do with `awesome-ts-loader` which hasn't been updated in 2 years, is archived in GitHub, and isn't aware of TS4

vlukashov:
There is indeed a problem with using an outdated `awesome-typescript-loader`, but apparently it's not critical as I am not able to reproduce this failure:

1. create a new Vaadin 18 / Fusion project with [https://start.vaadin.com/?preset=fusion](https://start.vaadin.com/?preset=fusion)
   (or with Vaadin CLI: `npx @vaadin/cli init --fusion npm-install-test`)
2. switch from `pnpm` to `npm`
   ```bash
   echo "vaadin.pnpm.enable = false" >> src/main/resources/application.properties
   ```
3. run `mvn` to start the project in the dev mode
   See that Vaadin runs npm to install frontend dependencies, and that the installation completes successfully.
   
   NOTE: npm does log a _warning_ related to `awesome-typescript-loader` requiring an old TS version, but that does not fail the installation:
   ```
   npm WARN awesome-typescript-loader@5.2.1 requires a peer of typescript@^2.7 || ^3 but none is installed. You must install peer dependencies yourself.
   ```
4. open the app in the browser at [http://localhost:8080/](http://localhost:8080/)

[...]

PR #9721, opened 2020-12-23
feat: replace awesome-typescript-loader with ts-loader

vlukashov:
`awesome-typescript-loader` has not been updated in the last 2 years and does not guarantee compatibility with TypeScript 4+. `ts-loader` on the other hand is well maintained and tested with TypeScript 4+, and is the most popular Webpack loader for TypeScript files.

Related to #9714.


web-padawan:
Here is an issue related to using `.js` extension in TS files which we need to keep in mind:

https://github.com/TypeStrong/ts-loader/issues/465

platosha (approved):

PR #9724, opened 2020-12-25
fix: use proper logging for installing pnpm and its directory

### LINKED ISSUE #9571: FrontendTools claims to install pnpm 4.4.0 if installed pnpm version is not supported
When installed pnpm version is not within range `[4.4.0, 4.6.0)`, `FrontendTools` installs `pnpm 4.5.0` locally. It logs, however, that version `4.4.0` is being installed:

https://github.com/vaadin/flow/blob/a635dc68edca91cb74af6dec4f4150e756fbf579/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java#L723-L724

denis-anisimov:
fixes #9571

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -150,7 +150,7 @@ public class FrontendTools {
 
     private final String nodeVersion;
     private final URI nodeDownloadRoot;
-    
+
     private final boolean ignoreVersionChecks;
 
     /**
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -639,13 +638,19 @@ private String getNonNull(String... valueArray) {
     private Supplier<List<String>> doEnsurePnpm() {
         List<String> path = getSuitablePnpm(baseDir);
         if (!path.isEmpty()) {
+            getLogger().trace(
```
mshabarov, line 641: Why is this message has level `trace`? Other log message within this method `Using pnpm command installed in the '{}' directory` has `info` level, thus, I just wondered what is the difference?

denis-anisimov: Technically I don't really care about log level here. It's hard to understand when `trace` should be used instead of `info`(and vice versa).
I may use `info` here  instead of `trace`.

But I've decided to use `trace`  here because one small observation : every place where `info` level (or more severe) is used is not standard situation: either new `pnpm`  is installed or some specific directory (home dir) is used for `pnpm` executable.
And I see this information as something that I may be interested in.

If global `pnpm` (or `pnpm` explicitly installed in the project dir) is used then this has been done explicitly by the user (global installation and `pnpm` installed inside the project dir has to be made manually). And then log info about this is something extra because there was an explicit action to install. So this is something known: I'm not interested in it.

mshabarov (approved):

mshabarov:
Cherry picked to `2.4`

PR #9729, opened 2020-12-30
fix: lookup for all available Executor services instead of only one

denis-anisimov:


flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
```diff
@@ -149,7 +149,8 @@ private DevModeHandler(Lookup lookup, int runningPort, File npmFolder,
         devServerPortFile = getDevServerPortFile(npmFolder);
 
         // Check whether executor is provided by the caller (framework)
-        Executor service = lookup.lookup(Executor.class);
+        Executor service = lookup.lookupAll(Executor.class).stream().findFirst()
```
mshabarov, line 152: Even if all executors are looked up, only any first one is used. So, what is the meaning of using `lookupAll` in place of `lookup`?
It's totally clear when it used as below:
```java
Collection<InstantiatorFactory> factories = lookup.lookupAll(InstantiatorFactory.class);
for (InstantiatorFactory factory : factories) { ... }
```
I propose to add some explanatory comment.

flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
```diff
@@ -149,7 +149,8 @@ private DevModeHandler(Lookup lookup, int runningPort, File npmFolder,
...
+                .orElse(null);
```
mshabarov, line 153: Is that change for further features/bugfixes? Does it require to be covered by unit test, or it anyway will be covered in other PRs?

denis-anisimov: Since there is no unit test for `lookup.lookup`  there is no "replacement" for it.
All `DevModeHandler`  functionality is too hard to test: the design is very bad.

I'm not going to add any unit tests here.

flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
denis-anisimov: done

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
mshabarov: Okay, as far as I understand the contract of `Lookup` is that `lookup` is supposed to return only one service impl. Thus, if we are expect there might be several executors here, then it makes sense.

PR #9735, opened 2021-01-02
feat: implement NativeDetails as html component

knoobie:
Reasoning: Currently `<details>` is missing from the list of html components. This PR fixes this. 

`NativeDetails`  is based on `vaadin-details` except that the html property is called `open` instead of `opened` and the event is called `Toggle` like the html dom event. 

To reduce the API noise `NativeDetails` extends `HtmlComponent` instead of `HtmlContainer`. `HasText` and `HasComponents` would interfere with the already set `<summary>` inside the `<details>`.

### DIFF
--- a/flow-html-components-testbench/src/main/java/com/vaadin/flow/component/html/testbench/NativeDetailsElement.java
+++ b/flow-html-components-testbench/src/main/java/com/vaadin/flow/component/html/testbench/NativeDetailsElement.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.html.testbench;
+
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.testbench.elementsbase.Element;
+
+/**
+ * A TestBench element representing a <code>&lt;details&gt;</code> element.
+ *
+ * @since
+ */
+@Element("details")
+public class NativeDetailsElement extends TestBenchElement {
+
+}
--- a/flow-html-components-testbench/src/main/java/com/vaadin/flow/component/html/testbench/NativeDetailsElement.java
+++ b/flow-html-components-testbench/src/main/java/com/vaadin/flow/component/html/testbench/NativeDetailsElement.java
@@ -29,8 +29,8 @@ import com.vaadin.testbench.elementsbase.Element;
 public class NativeDetailsElement extends TestBenchElement {
 
     /**
-     * Dispatches a {@code toggle} event by clicking the summary
-     * of the details. Toggles the details element open state.
+     * Dispatches a {@code toggle} event by clicking the summary of the details.
+     * Toggles the details element open state.
      */
     public void toggle() {
         findElement(By.tagName("summary")).click();
--- a/flow-html-components-testbench/src/main/java/com/vaadin/flow/component/html/testbench/NativeDetailsElement.java
+++ b/flow-html-components-testbench/src/main/java/com/vaadin/flow/component/html/testbench/NativeDetailsElement.java
@@ -15,6 +15,8 @@
  */
 package com.vaadin.flow.component.html.testbench;
 
+import org.openqa.selenium.By;
+
 import com.vaadin.testbench.TestBenchElement;
 import com.vaadin.testbench.elementsbase.Element;
 
@@ -26,4 +28,11 @@ import com.vaadin.testbench.elementsbase.Element;
 @Element("details")
 public class NativeDetailsElement extends TestBenchElement {
 
+    /**
+     * Dispatches a {@code toggle} event by clicking the summary
+     * of the details. Toggles the details element open state.
+     */
+    public void toggle() {
+        findElement(By.tagName("summary")).click();
+    }
 }
--- a/flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
+++ b/flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
@@ -1,0 +1,42 @@
+package com.vaadin.flow.component.html.testbench;
+
+import com.vaadin.flow.testutil.ChromeBrowserTest;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class NativeDetailsElementIT extends ChromeBrowserTest {
+
+    private NativeDetailsElement details;
+    private DivElement log;
+
+    @Before
+    public void open() {
+        getDriver().get("http://localhost:8888/Details");
+        details = $(NativeDetailsElement.class).id("details");
+        log = $(DivElement.class).id("log");
+    }
+
+    @Test
+    public void openDetails() {
+        details.setProperty("open", true);
+        Assert.assertEquals("Toggle event is 'true'", log.getText());
+    }
+
+    @Test
+    public void openAndCloseDetails() {
+        details.setProperty("open", true);
+        Assert.assertEquals("Toggle event is 'true'", log.getText());
+
+        details.setProperty("open", false);
+        Assert.assertEquals("Toggle event is 'false'", log.getText());
+    }
+
+    @Test
+    public void closingAlreadyClosedDetails() {
+        details.setProperty("open", false);
+        // Event should not be triggered, because details open property
+        // defaults to false
+        Assert.assertEquals("", log.getText());
+    }
+}
--- a/flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
+++ b/flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
@@ -1,42 +1,84 @@
 package com.vaadin.flow.component.html.testbench;
 
-import com.vaadin.flow.testutil.ChromeBrowserTest;
 import org.junit.Assert;
-import org.junit.Before;
 import org.junit.Test;
+
+import com.vaadin.flow.testutil.ChromeBrowserTest;
 
 public class NativeDetailsElementIT extends ChromeBrowserTest {
 
     private NativeDetailsElement details;
     private DivElement log;
-
-    @Before
-    public void open() {
-        getDriver().get("http://localhost:8888/Details");
-        details = $(NativeDetailsElement.class).id("details");
-        log = $(DivElement.class).id("log");
-    }
+    private NativeButtonElement button;
 
     @Test
     public void openDetails() {
+        prepareTest(false);
+
         details.setProperty("open", true);
-        Assert.assertEquals("Toggle event is 'true'", log.getText());
+        Assert.assertEquals("Toggle event number '1' is 'true'", log.getText());
     }
 
     @Test
     public void openAndCloseDetails() {
+        prepareTest(false);
+
         details.setProperty("open", true);
-        Assert.assertEquals("Toggle event is 'true'", log.getText());
+        Assert.assertEquals("Toggle event number '1' is 'true'", log.getText());
 
         details.setProperty("open", false);
-        Assert.assertEquals("Toggle event is 'false'", log.getText());
+        Assert.assertEquals("Toggle event number '2' is 'false'", log.getText());
     }
 
     @Test
     public void closingAlreadyClosedDetails() {
+        prepareTest(false);
+
         details.setProperty("open", false);
         // Event should not be triggered, because details open property
         // defaults to false
         Assert.assertEquals("", log.getText());
+    }
+
+    @Test
+    public void openAndCloseDetailsFromOtherServerSideComponent() {
+        prepareTest(false);
+
+        button.click();
+        Assert.assertEquals("Toggle event number '1' is 'true'", log.getText());
+
+        button.click();
+        Assert.assertEquals("Toggle event number '2' is 'false'", log.getText());
+    }
+
+    @Test
+    public void toggleNativeDetailsWithTestBenchElement() {
+        prepareTest(false);
+
+        details.toggle();
+        Assert.assertEquals("Toggle event number '1' is 'true'", log.getText());
+
+        details.toggle();
+        Assert.assertEquals("Toggle event number '2' is 'false'", log.getText());
+    }
+
+    @Test
+    public void openDetailsFromServerSideOnInitialRendering() {
+        prepareTest(true);
+
+        // Event should be triggered once, because details was already opened on server side
+        // triggering the toggle event.
+        Assert.assertEquals("Toggle event number '1' is 'true'", log.getText());
+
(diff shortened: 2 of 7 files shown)


flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Test
+    public void openAndCloseDetails() {
+        details.setProperty("open", true);
+        Assert.assertEquals("Toggle event is 'true'", log.getText());
+
+        details.setProperty("open", false);
+        Assert.assertEquals("Toggle event is 'false'", log.getText());
+    }
+
```
pleku, line 34: Missing an IT where the details is opened via server side (by executing code in a click listener)

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    public void closingAlreadyClosedDetails() {
+        details.setProperty("open", false);
+        // Event should not be triggered, because details open property
+        // defaults to false
+        Assert.assertEquals("", log.getText());
+    }
+}
```
pleku, line 42: Missing an IT where the details has been opened through server side state when the component is rendered

(I usually do it by using a query/view parameter and reading that in the view to determine it should be opened when present, e.g. `details?opened` or `details/opened`)

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * Adds a listener for {@code toggle} events fired by the
+     * details.
+     *
+     * @param listener
+     *            the listener
+     * @return a {@link Registration} for removing the event listener
+     */
+    public Registration addOpenChangedListener(
```
pleku, line 272: This is against our naming conventions - since the event name is `ToggleEvent`, the corresponding listener API should thus be `addToggleListener`.

I understand the intention to go with same event name as in browser, but then I think this API should match it.
This current API, `addOpenChangedListener` would IMO in this case have to match the `open`-property and it being changed.

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
```diff
@@ -0,0 +1,277 @@
...
+ * Component representing a <code>&lt;details&gt;</code> element.
+ *
+ * @author Vaadin Ltd
+ * @since
+ */
+@Tag(Tag.DETAILS)
+public class NativeDetails extends HtmlComponent implements ClickNotifier<NativeDetails> {
+
+    private static final PropertyDescriptor<Boolean, Boolean> openDescriptor = PropertyDescriptors
+      .propertyWithDefault("open", false);
```
pleku, line 44: One problem with the `open` property on server side is that it is not synchronized with the client side value. And I don't know if this is intentional or not because there is no javadocs mention on this - so I would expect it to be synchronized. Otherwise things will get unsync. So the options are:
a) fix the property synchronization to happen automatically on the `toggle` event. I don't know how much details you need to make this happen - the element API has `addPropertyChangeListener(String propertyName, String domEventName, PropertyChangeListener listener)` which can be used - but to do it properly you should have one property listener for each component and then have that one trigger the component level `ToggleEvent` on server side after updating the property value (in case there are listeners on server side). You can utilize the `ComponentEventBus` for firing the events on the component level.
b) Remove the `open` property and change the API to be `open()` & `close()` instead - then it doesn't match the DOM API though
c) Remove the `open` property and make it be `toggle(boolean open)` instead without getter ?

I would go with a), WDYT Christian ?

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
```diff
@@ -0,0 +1,42 @@
...
+        getDriver().get("http://localhost:8888/Details");
+        details = $(NativeDetailsElement.class).id("details");
+        log = $(DivElement.class).id("log");
+    }
+
+    @Test
+    public void openDetails() {
+        details.setProperty("open", true);
+        Assert.assertEquals("Toggle event is 'true'", log.getText());
+    }
```
pleku, line 24: Maybe: Missing an IT where the detais summary is focused and then opened/closed with space key ?

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
```diff
@@ -0,0 +1,277 @@
...
+import java.util.Objects;
+
+/**
+ * Component representing a <code>&lt;details&gt;</code> element.
+ *
+ * @author Vaadin Ltd
+ * @since
+ */
+@Tag(Tag.DETAILS)
+public class NativeDetails extends HtmlComponent implements ClickNotifier<NativeDetails> {
```
pleku, line 41: Non-blocking: As focus+space should open the details, I wonder if this should implement `Focusable` ...
maybe not needed for now (or ever)

flow-html-components-testbench/src/main/java/com/vaadin/flow/component/html/testbench/NativeDetailsElement.java
```diff
@@ -0,0 +1,29 @@
...
+import com.vaadin.testbench.elementsbase.Element;
+
+/**
+ * A TestBench element representing a <code>&lt;details&gt;</code> element.
+ *
+ * @since
+ */
+@Element("details")
+public class NativeDetailsElement extends TestBenchElement {
+
```
pleku, line 28: Should there be `toggle()` / `click()` API which simulates user clicking the element summary to open/close it

pleku (changes requested):

pleku:
Also one overall note - the import order does not match the conventions https://github.com/vaadin/flow/blob/master/CONTRIBUTING.md#project-setup

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
knoobie: While implementing, I asked myself the same and played with `Focusable`. The problem or let's say the technical detail is, that the `<detail>` isn't focused. Instead the `<summary>` gets focused and is used by the browser for keyboard handling to open/close the details. 

Instead of implementing `Focusable` in both components and delegating every call to the `<summary>`, I did not implement it to reduce complexity.

See WhatWG HTML Spec:

> This ability to request that additional information be shown or hidden may simply be the activation behavior of the appropriate summary element

Source: https://html.spec.whatwg.org/multipage/interactive-elements.html#htmldetailselement

This note was extremely interesting while testing the feature with "fast" clicking:

> When the open attribute is toggled several times in succession, these steps essentially get coalesced so that only one event is fired.

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
knoobie: Option A sounds like the most reasonable. I'm gonna take a look at it. 

I would probably try the `@Synchronize` approach first, which is used in the [flow component details](https://github.com/vaadin/vaadin-flow-components/blob/master/vaadin-details-flow-parent/vaadin-details-flow/src/main/java/com/vaadin/flow/component/details/Details.java). 

I'm wondering why I didn't used it, while creating the component.. I think I wanted to use the same API (PropertyDescriptor) as every other html component in the package.

knoobie:
I've read your other comments and try to fix them all on the next update. Probably in 1-2 days or at the weekend. Thanks for the thought out review!

flow-html-components-testbench/src/main/java/com/vaadin/flow/component/html/testbench/NativeDetailsElement.java
knoobie: Added a `toggle()` method and tested the method within `NativeDetailsElementIT#toggleNativeDetailsWithTestBenchElement`.

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
knoobie: Personally, I don't like to test browser default behavior on native elements, but I can create a test if you want ;)

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
knoobie: Done. Renamed the method to `addToogleListener`.

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
knoobie: Done  within `NativeDetailsElementIT#openAndCloseDetailsFromOtherServerSideComponent`.

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
knoobie: Done  within `NativeDetailsElementIT#openAndCloseDetailsFromOtherServerSideComponent`.

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsView.java
```diff
@@ -0,0 +1,22 @@
...
+public class NativeDetailsView extends Div {
+
+    public NativeDetailsView() {
+        Div log = new Div();
+        log.setId("log");
+
+        NativeDetails details = new NativeDetails("summary", new Paragraph("content"));
+        details.setId("details");
+        details.addOpenChangedListener(e -> {
+            log.setText("Toggle event is '" + e.isOpened() + "'");
```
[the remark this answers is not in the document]
knoobie: Done. Every test case was updated to include the current event count.

knoobie:
> Also one overall note - the import order does not match the conventions https://github.com/vaadin/flow/blob/master/CONTRIBUTING.md#project-setup

I've tried my luck with the intellij setup. Resulting in weird formatting behavior for me. 

image

I changed the obvious formatting for the imports by hand to match the contributing guideline and used `mvn formatter:format` for other stuff. 

FYI: `mvn formatter:format` resulted in this:

image

I think everything is done now :)

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
pleku: I agree, it is not necessary. Hence the _maybe_

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
```diff
@@ -257,21 +247,20 @@ public ToggleEvent(NativeDetails source, boolean fromClient,
          * @return whether details are expanded or collapsed
          */
         public boolean isOpened() {
-            return opened;
+            return getSource().isOpen();
```
pleku, line 249: My first impression was that I would have kept the "cached" value in the event and instead used that one for all the events. But after thinking it through, I thought about that since that is how it works (for good reasons) for `ValueChangeEvent`. Now this is not semantically exactly the same as that.

I can't say which one is more expected or better for usage, but I think the method javadocs need to be updated to explicitly mention that _is this the captured value_ or just delegating to the actual details component itself.

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
```diff
@@ -1,42 +1,84 @@
...
+        Assert.assertEquals("Toggle event number '2' is 'false'", log.getText());
+    }
+
+    @Test
+    public void openDetailsFromServerSideOnInitialRendering() {
+        prepareTest(true);
+
+        // Event should be triggered once, because details was already opened on server side
+        // triggering the toggle event.
+        Assert.assertEquals("Toggle event number '1' is 'true'", log.getText());
```
pleku, line 75: I'm wondering, is this event triggered by the server side or the client side ? (I understood it is latter)
When it happens on the client side after the component has been rendered, it will require another roundtrip to just fire the event (the property value has not changed). So basically for the `ToggleEvent` (always) reports that `isFromClient` is `true` event though the `setOpen` was actually called from the server side to trigger the toggle.

While this is a trivial component itself, it is not that trivial to integrate due to this event-property synchronization. I don't know how big (if any) problem it is that the programmatic toggle event would cause an extra roundtrip and not reveal if it actually occurred by server/client. But I do know that the intention should be documented (in event or adding listener method and `setOpen`) and preferably also confirmed in a test so that it is considered "intended behavior" instead of just an oversight.

Since the `Toggle` event semantics are not the same as would be with `OpenChangedEvent`, I would just document this behavior. Anyone who needs to know if the open changed by server/client, they would use `details.getElement.addPropertyChangeListener("open"....` to distinct that.

pleku (changes requested):

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
knoobie: While developing, I had the same problem in mind. Looking at e.g. other roundtrip / client based actions like `focus` I was like "okay, that's how it is." Same as you said here: https://github.com/vaadin/flow/issues/9768

I'm not sure how I can tackle this to force the "server side" change without any hacks. I would probably go with a javadoc notice like with your comment above.

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
knoobie: Done. Javadoc updated.

flow-html-components-testbench/src/test/java/com/vaadin/flow/component/html/testbench/NativeDetailsElementIT.java
knoobie: Done. Javadoc updated.

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
```diff
@@ -214,22 +213,27 @@ public void setOpen(boolean open) {
...
     @DomEvent("toggle")
     public static class ToggleEvent extends ComponentEvent<NativeDetails> {
 
+        private final boolean open;
+
         /**
          * ToggleEvent base constructor.
+         * <p>
+         * Note: This event is always triggered on client side.
+         * Resulting in {@code fromClient} to be always {@code true}.
```
pleku, line 236: Non-blocking: this note would have belonged to overridden `isFromClient()` to be discoverable (not blocking since it is in the add listener method)

pleku (approved):

flow-html-components/src/main/java/com/vaadin/flow/component/html/NativeDetails.java
knoobie: My first draft (non-commited) had this change. I decide against this, because I had to overwrite the method "just to clarify" the javadoc while additionally fighting with myself if I wanna return `true` or `super.isFromClient()`.

pleku:
Thanks for this contribution too @knoobie !

mshabarov:
Cherry picked to `2.5`

PR #9738, opened 2021-01-04
fix: made path separator based on os

taefi:
TaskUpdateThemeImportTest fails on windows agents. File Path separators are updated to use OS path separators.


caalador:
Basically `/` should work fine in windows as long as you don't mix and match the separator used.

`You can use slash character as file separator for both OS, in other words you can use C:/PDFMalwareDataAnalyser/Txt/ instead of C:\\PDFMalwareDataAnalyser\\Txt\\ it will still work on Windows OS.`

taefi:
Of course, java io/nio API supports `/` in paths instead of `\\`. This PR was just a test to be configured to be built on a Windows agent for testing purposes ;-)

caalador:
Is this still required or can it be closed?

taefi:
This PR is not needed anymore and was just a test.

PR #9740, opened 2021-01-04
Use CSS import in component theming

### LINKED ISSUE #9736: Test using css import inside a component theme
Test if it is possible to use CSS @import statements to theme the components.

If it can be done a IT test should be added for the case, else a note on this limitiation should be added to the documentation.

caalador:
Add test for using css `@import` for component
specific theme files.

Closes #9736


pleku (approved):
Should there be an addition to the component theming docs that says that importing another file works ?

caalador:
I guess `Component Styles` section could have a note or something that you can use import statements for component styles.

PR #9743, opened 2021-01-05
Fix: resource loading string

caalador:
Loading from classpath resource should
not start with '/'

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -831,7 +831,7 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
             String clientEngine = getClientEngine(resourceProvider);
             boolean resolveNow = !productionMode || clientEngine == null;
             if (resolveNow
-                    && resourceProvider.getClientResource("/META-INF/resources/"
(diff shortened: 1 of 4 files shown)


pleku (commented):
Should we instead make the resource provider methods say that the path should or should not start with "/" ? Or say that it does not care and make the implementation's responsibility to handle whether or not the path starts with "/" ?
Just wondering how we could avoid this type of issues in the future.

Also searched the sources and noticed that 
https://github.com/vaadin/flow/blob/master/flow-server/src/test/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandlerTest.java#L266 might also need changing (if the test fails)

pleku:
There are other failures too https://teamcity.vaadin.com/viewLog.html?buildId=70705&buildTypeId=Flow_FlowTests

caalador:
It's ok to use absolute names for resources, but in this instance the resource is not found if it's absolute. I think it's more of a Spring issue, but as it works fine in plain Java without the absolute then we should use this way as it's compatible with both.

pleku:
>I think it's more of a Spring issue, but as it works fine in plain Java without the absolute then we should use this way as it's compatible with both.

This is the tidbit that I think could be added to the javadocs by minimum if it cannot be enforced in code as then we might be able to catch these issues rather earlier than later next time.

pleku (approved):

PR #9744, opened 2021-01-05
feat!: Application theme only considers styles.css

### LINKED ISSUE #9741: Import only styles.css as global styles instead of all css files from theme root folder
Based on the DX testing feedback and internal review on how things worked in previous framework versions, we should only import one "magic css file" for the application theme shared styles instead of the original feature design of:
```
Any other stylesheets not in the components folder are loaded as global css.
```
The benefit of this is that the developer gets to have explicit control over what is loaded and in which order by having to import the css themselves to the file. The proposal is to name this file as `styles.css` like with legacy framework versions 6-8. Another alternative could be `shared-styles.css`, but at the time of writing of this the former should be used. This will mean that there is less magic and more control for the global css files.

### Acceptance Criteria
- [x] only load `styles.css` file from the active theme folder
- [x] when in development mode (devserver running), and there is an active in-project "application" theme in use without the `styles.css` file present in active theme the folder, automatically create the file with content that says something like `import your application global css files here or add the styles directly to this file`
- [x] when building the frontend for production and the active in-project theme folder does not contain the `styles.css` file, fail the build with error stating that "styles.css file is missing and is needed for the _theme-name_ in use" (even as empty as we want to avoid developer errors where the wrong theme is used or some files are not being added to the theme).
- [x] Update this requirement to documentation to mention that it is needed for getting explicit control over the loading order of the global css
- [x] Update https://github.com/vaadin/application-theme-demo to use this

caalador:
Now the application theme only takes
in 'styles.css' and 'document.css'. All other
css files need to be imported in one of these.
'styles.css' will be created if theme folder
exists in dev mode.

Fixes #9741


flow-server/src/main/resources/webpack.generated.js
```diff
@@ -171,6 +171,7 @@ const themeOptions = {
   // The following matches target/flow-frontend/themes/theme-generated.js
   // and for theme in JAR that is copied to target/flow-frontend/themes/
   // and not frontend/themes
+  devMode: devMode,
```
mshabarov, line 174: non-blocking: shift the comment lines one line down, because they correspond to `themeResourceFolder: flowFrontendThemesFolder,`

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -27,7 +27,9 @@ const { checkModules } = require('./theme-copy');
 const themeComponentsFolder = 'components';
 // The contents of a global CSS file with this name in a theme is always added to
 // the document. E.g. @font-face must be in this
-const themeFileAlwaysAddToDocument = 'document.css';
+const documentCssFile = 'document.css';
+// styles.css is the only entrypoint css file with document.css. Everything else should be imported using css @import
+const stylesCssFile = 'styles.css';
```
mshabarov, line 32: Should we add an extra test proving that the styles in other files (not `styles.css` and `document.css`) are ignored and not taken into account if they aren't explicitly imported in `styles.css`?

caalador: added

flow-server/src/main/resources/webpack.generated.js
caalador: done

mshabarov (approved):

PR #9753, opened 2021-01-07
fix: Generate TS definition for theme-generated.js

### LINKED ISSUE #9678: Change theme-generated.js to be TypeScript
When you try
```
import { applyTheme } from '@vaadin/flow-frontend/themes/theme-generated';
applyTheme(document)
```
in a TypeScript file, you get an error
```
Could not find a declaration file for module '@vaadin/flow-frontend/themes/theme-generated'.
```

Add the mentioned code to a V19 alpha 1 project

### Expected behavior
The import should just work

### Actual behavior
The declaration error is thrown

### Versions:
    - Vaadin / Flow version:19 alpha 1

tanbt:
Fixes #9678


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrap.java
```diff
@@ -91,9 +91,8 @@ private String getIndexTsEntryPath() {
     private Collection<String> getThemeLines() {
         Collection<String> lines = new ArrayList<>();
         if (shouldApplyAppTheme()) {
-            lines.add("//@ts-ignore");
             lines.add(
-                    "import {applyTheme} from '../../target/flow-frontend/themes/theme-generated.js';");
+                    "import { applyTheme } from '@vaadin/flow-frontend/themes/theme-generated';");
```
caalador, line 95: Shouldn't this work as `themes/theme-generated` as `target/flow-frontend` should be in the modules definition making it available to webpack.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrap.java
tanbt: Fixed.

caalador: Forgot to push?

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrapTest.java
```diff
@@ -86,8 +86,7 @@ public void should_load_AppTheme() throws MalformedURLException, ExecutionFailed
         String content = taskGenerateBootstrap.getFileContent();
         Assert.assertTrue(content.contains("import '../../target/index';\n" +
             "\n" +
-            "//@ts-ignore\n" +
-            "import {applyTheme} from '../../target/flow-frontend/themes/theme-generated.js';\n" +
+            "import { applyTheme } from '@vaadin/flow-frontend/themes/theme-generated';\n" +
```
caalador, line 89: this should then also be updated to expect `themes/theme-generated`

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrap.java
tanbt: Then I got it wrong. I thought `import { applyTheme } from '@vaadin/flow-frontend/themes/theme-generated';"` is the fixed.
I've tried `import { applyTheme } from 'themes/theme-generated';` but the module isn't found.

caalador: is it something to do with is being TS as it works fine on for instance embedded that only uses the shortened one.

caalador: I would perhaps also update `WebComponentGenerator` to use the same notation.

tanbt: I don't really get your ideas.

> is it something to do with is being TS

Having a TS definition file will allow importing by module name. i.e `@vaadin/flow-frontend/themes/theme-generated` which looks like a proper import in the TS file `vaadin.ts` (?).

>  instance embedded that only uses the shortened one

Has the shortened one `themes/theme-generated` (?) been used somewhere already? Is it correct when it doesn't work when I use `import { applyTheme } from 'themes/theme-generated';` ?

caalador: WebComponentGenerator uses the import as `import {applyTheme} from 'themes/theme-generated.js';` and that works fine, so I don't know why the TS file doesn't find the import on compilation except that it is searching somehow different as it's ts not js.
But we should keep the targets the same so that we don't confuse ourselves on those.

tanbt: I don't know where `import {applyTheme} from 'themes/theme-generated.js';` statement will be generated by `WebComponentGenerator`. If it is generated inside `target/some/folder/some-file`, then it can be resolved.

But from client-side, that statement should be resolved by path or `@vaadin/...` module name (?).
Is there any docs or example about `WebComponentGenerator`? Javadocs doesn't clarify its use-cases. There's also no IT test for it.

caalador: `flow-teests/test-embedding/test-embedding-application-theme` contains app theme tests for embedded components and `WebComponentGeneratorTest` contains junit test for generated embedded js

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrap.java
tanbt: So so the two-line applyTheme scripts in `WebComponentGenerator` are generated in `target\frontend\themed-component.js`.

I guess JS/TS scripts generated inside `target/....` can resolve `import { applyTheme } from 'themes/theme-generated';`  (because of some magical mechanisms that resolve the import path to `target/frontend` or `target/flow-frontend` folders) but not for scripts inside `frontend/...`. 
Scripts inside `project-root/frontend/...` can only resolve proper [JS Module Import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#syntax).

Did I got it correctly?

That's why the two import statements are different. So, `themed-component.js` works with both:
1.  `themes/theme-generated.js` which is short and magical, IDE can't resolve it (by ctrl + click), only flow-server can.
2.  `@vaadin/flow-frontend/themes/theme-generated` the proper import that can be used everywhere. I would keep this, even it's a bit longer.

caalador: webpack is the one that resolves 1. and that's because a the webpack definition contains `resolve{module{..., 'target/flow-frontend', ...} }`

caalador (approved):

PR #9760, opened 2021-01-07
chore: Bump Node.js supported version as it's getting EOL

mshabarov:
Bumps Node.js minimum supported major version from 10 to 12, as 10 is getting EOL in April 2021.


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -100,7 +100,7 @@
     static final String PNPM_INSTALLED_BY_NPM = PNPM_INSTALLED_BY_NPM_FOLDER
             + "bin/pnpm.js";
 
-    private static final int SUPPORTED_NODE_MAJOR_VERSION = 10;
+    private static final int SUPPORTED_NODE_MAJOR_VERSION = 12;
     private static final int SUPPORTED_NODE_MINOR_VERSION = 0;
```
pleku, line 104: Is 12.0 supported ?

pleku (changes requested):
The commit message should probably be `chore!` or such as this should be highlighted in release notes

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
mshabarov: Change to 12.20 as it's a latest minor version. Will double check if 12.20 is indeed supported by flow.

mshabarov: Tested with application theme demo. It works as expected. IT tests should also show if it's supported.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -57,11 +57,11 @@
  */
 public class FrontendTools {
 
-    public static final String DEFAULT_NODE_VERSION = "v12.18.3";
+    public static final String DEFAULT_NODE_VERSION = "v14.15.4";
 
     public static final String DEFAULT_PNPM_VERSION = "4.5.0";
 
-    public static final String INSTALL_NODE_LOCALLY = "%n  $ mvn com.github.eirslett:frontend-maven-plugin:1.10.0:install-node-and-npm -DnodeVersion=\"v12.18.3\" ";
+    public static final String INSTALL_NODE_LOCALLY = "%n  $ mvn com.github.eirslett:frontend-maven-plugin:1.10.0:install-node-and-npm -DnodeVersion=\"v14.15.4\" ";
```
manolo, line 64: I would use DEFAULT_NODE_VERSION here to avoid maintaining the  value twice in the code

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
```diff
@@ -87,15 +87,15 @@ public void setup() throws IOException {
     @Ignore("Ignored to lessen PRs hitting the server too often")
     public void installNode_NodeIsInstalledToTargetDirectory()
             throws FrontendUtils.UnknownVersionException {
-        String nodeExecutable = tools.installNode("v12.18.3", null);
+        String nodeExecutable = tools.installNode("v14.15.4", null);
```
manolo, line 90: Why not use the static DEFAULT_NODE_VERSION  in tests for the installation and assertions, having to change tests every time we bump node is cumbersome and error prone

manolo (dismissed):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
mshabarov: Done

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -100,8 +101,8 @@
     static final String PNPM_INSTALLED_BY_NPM = PNPM_INSTALLED_BY_NPM_FOLDER
             + "bin/pnpm.js";
 
-    private static final int SUPPORTED_NODE_MAJOR_VERSION = 10;
-    private static final int SUPPORTED_NODE_MINOR_VERSION = 0;
+    private static final int SUPPORTED_NODE_MAJOR_VERSION = 12;
+    private static final int SUPPORTED_NODE_MINOR_VERSION = 20;
```
pleku, line 105: Would there be some reason why we would not support node 12.13 where the LTS starts for node 12 ? 
This change would require a lot of users with Node versions 12.13-12.19 to update.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -100,8 +101,8 @@
...
     private static final int SUPPORTED_NPM_MAJOR_VERSION = 5;
     private static final int SUPPORTED_NPM_MINOR_VERSION = 6;
```
pleku, line 107: I think npm versions should be also updated to match the version that comes with the supported node version above, see https://nodejs.org/en/download/releases/

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
pleku: Or by minimum we could require the previous default version 12.18.3 but I would not require newer without a good reason

pleku (changes requested):
Up to @manolo to say if we would include this to v19 and what would be the minimum version. I think the previous default version of 12.18.3 could be a good choice, even though it will require some updates to be made.

Artur-:
Why are you mixing two things in the same PR? 
- Bump the default installed Node to latest LTS - non breaking, can and should be merged for V19
- Change the supported Node versions. Breaking change that cannot just be merged based on that "ITs pass"

pleku:
> Why are you mixing two things in the same PR?

True, bumping the default installed version could be merged immediately and the supported version can stay pending open for a decision on when to do it. Nobody has said anywhere that this (bumping the minimum supported version) would be merged "just because ITs pass".

On the _why_, there was no need to rush into this, but @mshabarov felt eager to get it going forward and opened this PR so here we are.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
mshabarov: Good point, I didn't take into account the fact 12.20 will require many updates on users side. Not certain about what is better - 12.13 or 12.18. I would go with 12.18 as it newer.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
mshabarov: Sure! Thanks!

mshabarov:
> Why are you mixing two things in the same PR?
> 
> * Bump the default installed Node to latest LTS - non breaking, can and should be merged for V19
> * Change the supported Node versions. Breaking change that cannot just be merged based on that "ITs pass"

Good comment! Will split these changes into two PRs

pleku:
So I think we won't merge this for 6.0 / v19. I don't know when we're going to update the versions and who decides it, so maybe @mshabarov you should convert this to a draft or just close it but leave the branch in tact.

Artur-:
Before doing this, we would need an auto installation of Node if the existing one is too old (#9864)

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
manolo: I think 12.18 is a good choice

manolo: regarding porting this to v19, I think it's not needed since we have v20 already

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -102,10 +102,10 @@
     private static final FrontendVersion WHITESPACE_ACCEPTING_NPM_VERSION = new FrontendVersion(
             7, 0);
 
-    private static final int SUPPORTED_NODE_MAJOR_VERSION = 10;
-    private static final int SUPPORTED_NODE_MINOR_VERSION = 0;
-    private static final int SUPPORTED_NPM_MAJOR_VERSION = 5;
-    private static final int SUPPORTED_NPM_MINOR_VERSION = 6;
+    private static final int SUPPORTED_NODE_MAJOR_VERSION = 12;
+    private static final int SUPPORTED_NODE_MINOR_VERSION = 18;
```
mshabarov, line 106:
```suggestion
    private static final int SUPPORTED_NODE_MINOR_VERSION = 22;
```

mshabarov:
#9864 is not done and merged, so we can revisit this Node version upgrade again.

mshabarov:
I've changed the title to not be a breaking change.

pleku (approved):

mshabarov:
@manolo could you please review this PR.

manolo (approved):
LGTM

Issue #4613, opened 2018-09-17
com.vaadin.flow.server.VaadinRequest requires servlet-spec


domdorn:
I'm currently implementing a Vaadin-Flow server on top of the Play!Framework

The VaadinRequest is supposed to be an platform agnostic abstraction over an http request.
However, the method
`com.vaadin.flow.server.VaadinRequest#getCookies`
directly references `javax.servlet.http.Cookie` which requires implementations to include the javax.servlet apis, although they are not used/required. 

Vaadin should introduce its own Cookie abstraction/interface which implementations can implement (like the WrappedSession)

pleku:
Thanks for the detailed issue report @domdorn !

In case someone is interested in contributing for this, please let us know (by writing here) so we can chat about possible solutions. Currently I cannot give an estimate when we might get to doing this ourselves - most like not during this year.

mcollovati:
This would be useful also for the Vert.x integration I'm working on (not so actively at the moment)

Issue #3722 and #2775 was also related to servlet API usage, but I didn't investigate recently to check if something has changed.

ovidiupopa91:
Hi @pleku ,

I would like to contribute and hopefully fix this enhancement.

pleku:
This issue is still up for grabs if someone wants to contribute - it is better to limit this to the `getCookies` & `addCookie` part.
It would be a breaking change to use something else, but as long as the API would be 1-1 it would not be so bad.
Another option to consider would be to move the current servlet based API to `VaadinServletRequest` & `VaadinServletResponse`
but that would still be problematic for backwards compatibility and being able to read&write cookies for other non-servlet based environments.

(sorry @ovidiupopa91 I've missed your ping completely due to being on vacation at the time)

PR #9777, opened 2021-01-08
Refactor startup initializers to use VaadinContext

### LINKED ISSUE #2775: When initializing RouteRegistry I want to use an abstraction to store the reference  a so that I'm not coupled to underlying technology
`RouteRegistry.getInstance()` keeps a `ServletContext` to store and retrieve the `RouteRegistry` instance; this way `RouteRegistry` is highly coupled to servlet technology.

As Flow already abstracts underlying technology (VaadinService, VaadinRequest, VaadinResponse, ...) also `RouteRegistry` may use an abstraction (name it RouteRegistryHolder) to handle the created instance.
This way it could be used also in non servlet environments (eg Vert.x) without stubbing the whole ServletContext interface.

An example in pseudo code

```java
interface RouteRegistryHolder {
  get(): RouteRegistry
  set(instance: RouteRegistry)
}
class ServletContextRouteRegistryHolder(ctx: ServletContext) : RouteRegistryHolder {
   get: RouteRegistry => ctx.getAttribute(RouteRegistry.class.getName())
   set(instance: RouteRegistry) => ctx.setAttribute(RouteRegistry.class.getName(), instance)
}
class VertxRouteRegistryHolder(ctx: Context) : RouteRegistryHolder {
   get: RouteRegistry => ctx.get(RouteRegistry.class.getName())
   set(instance: RouteRegistry) => ctx.put(RouteRegistry.class.getName(), instance)
}
```

or without extension

```java
class RouteRegistryHolder(getter: Supplier[RouteRegistry], setter: Consumer[RouteRegistry]) {
   get: RouteRegistry => getter.get()
   set(instance: RouteRegistry) => setter.accept(instance)
}
```

As an example `RouteRegistryInitializer.startup()` will do `RouteRegistry.getInstance(new ServletContextRouteRegistryHolder(servletContext))` whereas `VaadinVerticle.start()` will use `RouteRegistry.getInstance(new VertxRouteRegistryHolder(context))`.
Similar thing will be done in `VaadinServletService.getRouteRegistry` and `VertxVaadinService.getRouteRegistry`

### LINKED ISSUE #5735: Refactor servlet-based initialisers into service-based ones
Make all relevant initialisation triggered from `VaadinService` and/or its subclasses, rather than assume servlet-based initialisation.

This is related to #4760 and caused by #5565.

jhult:
This is the successor to [this pull request](https://github.com/vaadin/flow/pull/8469) (which ran long and had too many updates).

Most Vaadin startup initializers don't actually need `ServletContext`. Thus, their methods should use VaaddinContext instead. This pull request attempts to fix. This allows non-servlet implementing HTTP frameworks (such as [VertxVaadin](https://github.com/mcollovati/vertx-vaadin)) to directly call the main initialization logic via `VaadinContextStartupInitializer#initialize(Set, VaadinConntext)}` instead of needing to create fake `ServletContext`s.

I also standardized parameter names of overriden methods and also deprecated (where it made sense) public methods that used `ServletContext` in favor of a new method which uses `VaadinContext`.

I believe this fixes #5735 and also fixes #2775. In addition, this provides a good step towards allowing external libraries to reduce the dependency on servlet-api (#4613).

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/startup/DeferredServletContextInitializers.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/startup/DeferredServletContextInitializers.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 import com.vaadin.flow.di.Lookup;
+import com.vaadin.flow.di.LookupInitializer;
 import com.vaadin.flow.server.VaadinContext;
 
 /**
--- a/flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java
@@ -122,7 +122,7 @@ import static com.vaadin.flow.server.frontend.FrontendUtils.TARGET;
         HasErrorParameter.class, PWA.class, AppShellConfigurator.class })
 @WebListener
 public class DevModeInitializer
-        implements ClassLoaderAwareServletContainerInitializer, Serializable,
+        implements VaadinServletContextStartupInitializer, Serializable,
         ServletContextListener {
 
     static class DevModeClassFinder extends DefaultClassFinder {
@@ -196,11 +196,11 @@ public class DevModeInitializer
     // This attribute helps to avoid Dev Mode running twice.
     //
     // Addresses the issue https://github.com/vaadin/spring/issues/502
-    private static final String DEV_MODE_HANDLER_ALREADY_STARTED_ATTRIBUTE = "dev-mode-handler-already-started-attribute";
+    private static final class DevModeHandlerAlreadyStartedAttribute implements Serializable { }
 
     @Override
-    public void process(Set<Class<?>> classes, ServletContext context)
-            throws ServletException {
+    public void initialize(Set<Class<?>> classes, VaadinContext context)
+            throws VaadinInitializerException {
         initDevModeHandler(classes, context);
 
         setDevModeStarted(context);
@@ -218,13 +218,16 @@ public class DevModeInitializer
         }
     }
 
-    private void setDevModeStarted(ServletContext context) {
-        context.setAttribute(DEV_MODE_HANDLER_ALREADY_STARTED_ATTRIBUTE, true);
+    private void setDevModeStarted(VaadinContext context) {
+        context.setAttribute(DevModeHandlerAlreadyStartedAttribute.class, new DevModeHandlerAlreadyStartedAttribute());
     }
 
     /**
      * Initialize the devmode server if not in production mode or compatibility
      * mode.
+     *
+     * @deprecated Use {@link #initDevModeHandler(Set, VaadinContext)} instead
+     *            by wrapping {@link ServletContext} with {@link VaadinServletContext}.
      *
      * @param classes
      *            classes to check for npm- and js modules
@@ -234,12 +237,33 @@ public class DevModeInitializer
      * @throws ServletException
      *             if dev mode can't be initialized
      */
+    @Deprecated
     public static void initDevModeHandler(Set<Class<?>> classes,
             ServletContext context) throws ServletException {
-        VaadinContext vaadinContext = new VaadinServletContext(context);
+        try {
+            initDevModeHandler(classes, new VaadinServletContext(context));
+        } catch (VaadinInitializerException e) {
+            throw new ServletException(e);
+        }
+    }
+
+    /**
+     * Initialize the devmode server if not in production mode or compatibility
+     * mode.
+     *
+     * @param classes
+     *            classes to check for npm- and js modules
+     * @param context
+     *            VaadinContext we are running in
+     *
+     * @throws VaadinInitializerException
+     *            if dev mode can't be initialized
+     */
+    public static void initDevModeHandler(Set<Class<?>> classes,
+           VaadinContext context) throws VaadinInitializerException {
 
         ApplicationConfiguration config = ApplicationConfiguration
-                .get(vaadinContext);
+                .get(context);
         if (config.isProductionMode()) {
             log().debug("Skipping DEV MODE because PRODUCTION MODE is set.");
             return;
@@ -264,12 +288,11 @@ public class DevModeInitializer
         File flowResourcesFolder = new File(baseDir,
                 DEFAULT_FLOW_RESOURCES_FOLDER);
 
-        Lookup lookupFromServletConetext = new VaadinServletContext(context)
-                .getAttribute(Lookup.class);
+        Lookup lookupFromContext = context.getAttribute(Lookup.class);
         Lookup lookupForClassFinder = Lookup.of(new DevModeClassFinder(classes),
                 ClassFinder.class);
         Lookup lookup = Lookup.compose(lookupForClassFinder,
-                lookupFromServletConetext);
+                lookupFromContext);
         Builder builder = new NodeTasks.Builder(lookup, new File(baseDir),
                 new File(generatedDir), new File(frontendFolder));
 
@@ -371,7 +394,7 @@ public class DevModeInitializer
         Executor service = lookup.lookupAll(Executor.class).stream().findFirst()
                 .orElse(null);
 
-        Runnable runnable = () -> runNodeTasks(vaadinContext, tokenFileData,
+        Runnable runnable = () -> runNodeTasks(context, tokenFileData,
                 tasks);
 
         CompletableFuture<Void> nodeTasksFuture;
@@ -398,16 +421,32 @@ public class DevModeInitializer
     /**
      * Shows whether {@link DevModeHandler} has been already started or not.
      *
+     * @deprecated Use {@link #isDevModeAlreadyStarted(VaadinContext)} instead
+     *            by wrapping {@link ServletContext} with {@link VaadinServletContext}.
+     *
      * @param servletContext
      *            The servlet context, not <code>null</code>
      * @return <code>true</code> if {@link DevModeHandler} has already been
      *         started, <code>false</code> - otherwise
      */
+    @Deprecated
     public static boolean isDevModeAlreadyStarted(
             ServletContext servletContext) {
-        assert servletContext != null;
-        return servletContext.getAttribute(
-                DevModeInitializer.DEV_MODE_HANDLER_ALREADY_STARTED_ATTRIBUTE) != null;
+        return isDevModeAlreadyStarted(new VaadinServletContext(servletContext));
+    }
+
+    /**
+     * Shows whether {@link DevModeHandler} has been already started or not.
+     *
+     * @param context
+     *            The {@link VaadinContext}, not <code>null</code>
+     * @return <code>true</code> if {@link DevModeHandler} has already been
+     *         started, <code>false</code> - otherwise
+     */
+    public static boolean isDevModeAlreadyStarted(
+            VaadinContext context) {
+        assert context != null;
+        return context.getAttribute(DevModeHandlerAlreadyStartedAttribute.class) != null;
     }
 
     private static Logger log() {
@@ -457,7 +496,7 @@ public class DevModeInitializer
      * will fail in Java 9+
      */
     static Set<File> getFrontendLocationsFromClassloader(
-            ClassLoader classLoader) throws ServletException {
+            ClassLoader classLoader) throws VaadinInitializerException {
         Set<File> frontendFiles = new HashSet<>();
         frontendFiles.addAll(getFrontendLocationsFromClassloader(classLoader,
                 Constants.RESOURCES_FRONTEND_DEFAULT));
@@ -486,7 +525,7 @@ public class DevModeInitializer
 
     private static Set<File> getFrontendLocationsFromClassloader(
             ClassLoader classLoader, String resourcesFolder)
-            throws ServletException {
+            throws VaadinInitializerException {
         Set<File> frontendFiles = new HashSet<>();
         try {
             Enumeration<URL> en = classLoader.getResources(resourcesFolder);
@@ -542,7 +581,7 @@ public class DevModeInitializer
     }
 
     private static File getPhysicalFileOfJBossVfsDirectory(URL url)
-            throws IOException, ServletException {
+            throws IOException, VaadinInitializerException {
         try {
             Object virtualFile = url.openConnection().getContent();
             Class virtualFileClass = virtualFile.getClass();
@@ -568,12 +607,12 @@ public class DevModeInitializer
             return rootDirectory;
         } catch (NoSuchMethodException | IllegalAccessException
                 | InvocationTargetException exc) {
-            throw new ServletException("Failed to invoke JBoss VFS API.", exc);
+            throw new VaadinInitializerException("Failed to invoke JBoss VFS API.", exc);
         }
     }
 
     private static File getPhysicalFileOfJBossVfsJar(URL url)
-            throws IOException, ServletException {
+            throws IOException, VaadinInitializerException {
         try {
             Object jarVirtualFile = url.openConnection().getContent();
 
@@ -591,7 +630,7 @@ public class DevModeInitializer
             return tempJarFile;
         } catch (NoSuchMethodException | IllegalAccessException
                 | InvocationTargetException exc) {
-            throw new ServletException("Failed to invoke JBoss VFS API.", exc);
+            throw new VaadinInitializerException("Failed to invoke JBoss VFS API.", exc);
         }
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.server.startup;
+
+import com.vaadin.flow.server.VaadinContext;
+
+import javax.servlet.ServletContext;
+import java.util.Set;
+
+/**
+ * Applies this initializer to the given {@link VaadinContext}.
+ *
+ * It is intended to be called either:
+ * <ul>
+ * <li>directly by non-servlet implementing HTTP frameworks or</li>
+ * <li>indirectly on servlet container initialization (via
+ * {@link ClassLoaderAwareServletContainerInitializer#onStartup(Set, ServletContext)})</li>
+ * </ul>
+ *
+ * @since
+ *
+ * @see ClassLoaderAwareServletContainerInitializer
+ * @see VaadinServletContextStartupInitializer
+ */
+@FunctionalInterface
+public interface VaadinContextStartupInitializer {
+
+    /**
+     * Applies this initializer to the given context
+     *
+     * @param classSet
+     *            the Set of application classes which this initializer
+     *            needs to do its job
+     *
+     * @param context
+     *            the {@link VaadinContext} to use with this initializer
+     *
+     * @throws VaadinInitializerException
+     *            if an error has occurred
+     */
+    void initialize(Set<Class<?>> classSet, VaadinContext context) throws VaadinInitializerException;
+
+}
--- a/flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinInitializerException.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinInitializerException.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.server.startup;
+
+/**
+ * Indicates an issue during Vaadin initialization.
+ *
+ * @since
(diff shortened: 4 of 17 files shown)


denis-anisimov:
>This allows non-servlet implementing HTTP frameworks 

This is the main question which I would like to get the answer to.
I'm not familiar with ` VertxVaadin` .
Could you please describe shortly how and when VaadinContextStartupInitializer are (going to be ?) called.

My original question about initial PR was exactly about the need of this refactoring: `ServletContainerInitializer`  in this PR is used _as is_.
And Servlet 3.0 specification calls `ServletContainerInitializer` directly .
So when Vaadin app works inside a Servlet container the `VaadinContextStartupInitializer` abstraction is not used anyhow.
So I would like to understand the usecase for `VaadinContextStartupInitializer` .
In which way `VaadinContextStartupInitializer`s are supposed to be called : how they will be collected from the application and who will trigger their execution?

jhult:
@denis-anisimov 

External, non-servlet frameworks (such as VertxVaadin) should be able to manually call these classes:

`AnnotationValidator`
`DevModeInitializer`
`ErrorNavigationTargetInitializer`
`RouteRegistryInitializer`
`VaadinAppShellInitializer`
`WebComponentConfigurationRegistryInitializer`
`WebComponentExporterAwareValidator`

It is up to the external, non-servlet framework to manually find the above classes. One method would be to search for any class (not interface) which implements `VaadinContextStartupInitializer`. Then the `initialize` method could be called to do all the necessary startup logic.

The external, non-servlet framework would also need to manually locate the required `Set` of classes which the initializer needs.

However, right now (in the `master` branch), if an external, non-servlet framework tries to call `RouteRegistryInitializer`, it has to use this method `onStartup(Set<Class<?>> classSet, ServletContext servletContext)`. That is troublesome for frameworks which don't have/need a `ServletContext`.

[Here](https://github.com/mcollovati/vertx-vaadin/blob/78bcdd4f26feb0d131b04ead355b08198d18ad18/vertx-vaadin-flow-parent/vertx-vaadin-flow/src/main/java/com/github/mcollovati/vertx/vaadin/VaadinVerticle.java#L305) is how VertxVaadin does it. VertxVaadin has to create a fake `ServletContext` when in reality `ServletContext` is unnecessary unnecessary for the above classes.

The external, non-servlet framework would need to implement `VaadinContext` (e.g. `MyFrameworkCustomVaadinContext`). Implementing the required code for `VaadinContext` is much simpler than doing so for `ServletContext`. Then, with this PR, external, non-servlet frameworks could call this:
```
MyFrameworkCustomVaadinContext context = new MyFrameworkCustomVaadinContext();
RouteRegistryInitializer routeRegistryInitializer = new RouteRegistryInitializer();
routeRegistryInitializer.initialize(classSet, context);
```

denis-anisimov:
OK, thank you. Now it starts to look reasonable.

flow-server/src/main/java/com/vaadin/flow/server/startup/DeferredServletContextInitializers.java
```diff
@@ -23,6 +23,7 @@
 import java.util.concurrent.CopyOnWriteArrayList;
 
 import com.vaadin.flow.di.Lookup;
+import com.vaadin.flow.di.LookupInitializer;
```
denis-anisimov, line 26: This is the only change in this class.
Is it necessary.
I think it's just extra.

flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java
```diff
@@ -196,11 +196,11 @@ private void ensureImplementation(Class<?> clazz) {
     // This attribute helps to avoid Dev Mode running twice.
     //
     // Addresses the issue https://github.com/vaadin/spring/issues/502
-    private static final String DEV_MODE_HANDLER_ALREADY_STARTED_ATTRIBUTE = "dev-mode-handler-already-started-attribute";
+    private static final class DEV_MODE_HANDLER_ALREADY_STARTED_ATTRIBUTE implements Serializable { }
```
denis-anisimov, line 199: The name of the class should not be in upper case.
It should be in camel case.
The constant (an object of the type) should be upper case.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package com.vaadin.flow.server.startup;
+
+import com.vaadin.flow.server.VaadinContext;
+
+import javax.servlet.ServletContext;
+import java.util.Set;
+
+/**
+ * A functional interface which performs the required startup logic for a
```
denis-anisimov, line 24: `A functional interface` doesn't add anything to the javadocs.
Java type allows to see that this is an interface and it's functional. 
This sentence start just repeats the fact which is forced by Java.

We cleaned up (I hope everywhere) such extra words in javadocs. Could you please reformulate the sentence so that it doesn't contain extra words?

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+ * Vaadin initializer (those classes implemented by
```
denis-anisimov, line 25: What's the Vaadin initializer ?

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+ * {@link VaadinServletContextStartupInitializer}).
```
denis-anisimov, line 26: `VaadinServletContextStartupInitializer` should not be mentioned here in that way I think.
Because it's just a subinterface and the whole point of this PR to get rid of Servlet dependency.

It can be mentioned but only as a possible subinterface/example.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * This functional interface is intended to be called either:
```
denis-anisimov, line 28: `functional interface` is extra.

s/`This functional interface is intended`/`It is intended`/

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+ * <ul>
+ * <li>directly by non-servlet implementing HTTP frameworks or</li>
+ * <li>indirectly on servlet container initialization (via
+ * {@link ClassLoaderAwareServletContainerInitializer#onStartup(Set, ServletContext)})</li>
+ * </ul>
+ *
```
denis-anisimov, line 34: A couple of `@see`  tags could be added here: to 
- `ClassLoaderAwareServletContainerInitializer` (or `ServletContainerInitializer` instead or in addition to)
- `VaadinServletContextStartupInitializer`.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+ * @since
+ */
+@FunctionalInterface
+public interface VaadinContextStartupInitializer {
+
+    /**
+     * Performs the required startup logic for this initializer
```
denis-anisimov, line 41: I would may be use something like `applies this initializer to the given context`.
The current sentence doesn't say anything at all to me.
The suggested sentence is not that meaningful as well but at least it doesn't contain "startup logic" (because it's confusing: which start up logic ?)

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param classSet
```
denis-anisimov, line 43: This javadoc should be completely rewritten somehow.
That's the main challenge when you make an abstraction layer.
Since `VaadinContextStartupInitializer` doesn't know anything (and not based anymore) about `ServletContainerInitializer`  then it may not mention `HandlesTypes`. The latter annotation is an implementation detail of `ServletContainerInitializer`  and it's a servlet based.
So this mention should be removed completely.
The set of classes explanation should be rewritten as well: you are referring here to the implementation/realization detail of the  `ServletContainerInitializer` .
It should be just a set of classes which initializer needs to do its job.
There should not be any mention about specifics of this classes : extend, implement, etc...

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinInitializerException.java
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.server.startup;
+
+/**
+ * Exception indicating an issue during Vaadin initialization.
```
denis-anisimov, line 19: `Exception` is an extra word which repeats the Java declaration

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinServletContextStartupInitializer.java
```diff
@@ -0,0 +1,77 @@
...
+
+import com.vaadin.flow.server.VaadinContext;
+import com.vaadin.flow.server.VaadinServletContext;
+
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import java.util.Set;
+
+/**
+ * Interface which allows a library/runtime to be notified of a web
```
denis-anisimov, line 26: `Interface` is an extra word.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinServletContextStartupInitializer.java
```diff
@@ -0,0 +1,77 @@
...
+ * application's startup phase and perform any required programmatic
+ * registration of servlets, filters, and listeners in response to it.
+ *
+ * Known implementing classes include the following Vaadin initializers:
+ *
+ * <ul>
+ * <li>{@link AnnotationValidator}</li>
```
denis-anisimov, line 33: I would avoid mentioning all the implemented classes here.
This list has to be updated on every change in such classes :if some new class is added that will require change this list.
It's quite easy to find the list via introspecting hierarchy.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinServletContextStartupInitializer.java
```diff
@@ -0,0 +1,77 @@
...
+ * <li>{@link RouteRegistryInitializer}</li>
+ * <li>{@link ServletVerifier}</li>
+ * <li>{@link VaadinAppShellInitializer}</li>
+ * <li>{@link WebComponentConfigurationRegistryInitializer}</li>
+ * <li>{@link WebComponentExporterAwareValidator}</li>
+ * </ul>
+ *
+ * @since
+ *
+ * @see ClassLoaderAwareServletContainerInitializer
```
denis-anisimov, line 45: This is the only interface which I would keep here.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinServletContextStartupInitializer.java
```diff
@@ -0,0 +1,77 @@
...
+ * @see WebComponentConfigurationRegistryInitializer
+ * @see WebComponentExporterAwareValidator
+ */
+@FunctionalInterface
+public interface VaadinServletContextStartupInitializer extends
+        ClassLoaderAwareServletContainerInitializer,
+        VaadinContextStartupInitializer {
+
+    /**
+     * By default, this simply delegates to {@link #initialize(Set, VaadinContext)}
```
denis-anisimov, line 62: Are you sure that this javadoc is needed ?

Several issues here:
- you have completely overridden the javdocs from the original `process`
- the original javadocs are not mentioned here anyhow (via `inheritdoc`)
- the delegation note is extremely easy to see from the implementation since the impl is quite easy.
- it's quite obvious that any initializer may override this method and it's up to the initializer to do it: superclass should not mentioned subclass specifics.

flow-test-generic/src/main/java/com/vaadin/flow/testutil/ClassesSerializableTest.java
```diff
@@ -112,6 +112,8 @@
                 "com\\.vaadin\\.flow\\.server\\.startup\\.ApplicationRouteRegistry\\$OSGiRouteRegistry",
                 "com\\.vaadin\\.flow\\.server\\.startup\\.ApplicationRouteRegistry\\$OSGiDataCollector",
                 "com\\.vaadin\\.flow\\.server\\.startup\\.ClassLoaderAwareServletContainerInitializer",
+                "com\\.vaadin\\.flow\\.server\\.startup\\.VaadinServletContextStartupInitializer",
```
denis-anisimov, line 115: I have doubts for this change: our `ServletContainerInitiaizer`s  were `Serializable`.
Should the new abstraction layer also be `Serializable` ?

I'm not sure in fact because quite many implementations of `ServletContainerInitiaizer` weren't `Serializable` in fact.....

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/startup/DeferredServletContextInitializers.java
jhult: This import was added by my IDE because `LookupInitializer` is mentioned in the Javadoc. `VaadinContext` is also imported strictly for a Javadoc (not used anywhere in the class). I can either leave the import or make both fully qualified in the Javadoc. Which would you prefer?

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinServletContextStartupInitializer.java
jhult: I will remove this. However, I did pattern it off [ServletContainerInitializer]( https://github.com/eclipse-ee4j/servlet-api/blob/master/api/src/main/java/jakarta/servlet/ServletContainerInitializer.java#L23) which starts the Javadoc with `Interface`.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinServletContextStartupInitializer.java
jhult: This has been removed.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinServletContextStartupInitializer.java
jhult: The others have been removed.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinServletContextStartupInitializer.java
jhult: I removed this Javadoc.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinInitializerException.java
jhult: Fixed.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
jhult: Removed `A functional interface which`.

flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java
jhult: Fixed.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
jhult: Updated.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
jhult: I added these:

- `ClassLoaderAwareServletContainerInitializer`
- `VaadinServletContextStartupInitializer`

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
jhult: I will rework this.

FYI, the verbiage of `classes that extend, implement, or have been annotated` is patterned off [ServletContainerInitializer](https://github.com/eclipse-ee4j/servlet-api/blob/master/api/src/main/java/jakarta/servlet/ServletContainerInitializer.java#L68).

flow-test-generic/src/main/java/com/vaadin/flow/testutil/ClassesSerializableTest.java
jhult: I'm not sure whether this should be `Serializable` or not. If I don't include it, this error is thrown for the test:

```
[INFO] Running com.vaadin.tests.server.component.FlowClassesSerializableTest
[ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.017 s <<< FAILURE! - in com.vaadin.tests.server.component.FlowClassesSerializableTest
[ERROR] classesSerializable(com.vaadin.tests.server.component.FlowClassesSerializableTest)  Time elapsed: 0.735 s  <<< FAILURE!
java.lang.AssertionError: 
Serializable not implemented by the following classes and interfaces: ,
com.vaadin.flow.server.startup.VaadinContextStartupInitializer,
com.vaadin.flow.server.startup.VaadinServletContextStartupInitializer
	at com.vaadin.flow.testutil.ClassesSerializableTest.failSerializableClasses(ClassesSerializableTest.java:433)
	at com.vaadin.flow.testutil.ClassesSerializableTest.classesSerializable(ClassesSerializableTest.java:337)
```

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
jhult: Removed the mention of `VaadinServletContextStartupInitializer`.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
jhult: Updated Javadoc to remove reference to "Vaadin initializer".

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
jhult: Updated Javadoc.

flow-server/src/main/java/com/vaadin/flow/server/startup/DeferredServletContextInitializers.java
denis-anisimov: OK, no problems.
The only thing which is weird: there are no other changes. It means that javadocs were broken before this change.

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
denis-anisimov: >FYI, the verbiage of classes that extend, implement, or have been annotated is patterned off

It's totally fine for `ServletContainerInitializer`  to use the words and everything because this is how it works: it's part of Servlet 3.0 specification.

`VaadinContextStartupInitializer` is an abstraction layer and it's not specified by anyone what classes are passed to the method.
`HandlesTypes` is not used anymore in this abstraction layer (or do you still expect that it's used ?) and as a consequence its semantic with `classes that extend, implement, or have been annotated` can't be used here.
You may write whatever you want/require here but this should be something which anyone may understand.
It's not clear anymore where those classes comes from now (strongly related to  my point with the whole refactoring) and what's their meaning.

If you still want to rely the `VaadinContextStartupInitializer` implementation on `HandlesTypes` then please describe it properly here and then you may use `classes that extend, implement, or have been annotated`.

But the question : is `HandlesTypes` still applicable for `VaadinContextStartupInitializer`  ? ( I think no, but it's not my spedcification ).

denis-anisimov: I'm kind of fine with the current javadocs.
So this can be kept as is.
If you think you may improve it then please go ahead.

The javadocs in fact are quite confusing since they don't define the exact contract: any class may be passed to the `VaadinContextStartupInitializer`  in the set. 
As I mentioned already `ServletContainerInitializer` works with `HandlesTypes` : so the Servlet container passes the classes based on `HandlesTypes`  annotation.
It's completely unclear  which classes are passed to `VaadinContextStartupInitializer` : as I mentioned `HandlesTypes` is not anymore related to this abstraction (and in fact it can't be related since this class in a servlet specific).
So in fact the NON Servlet container has to define the semantic based on every `VaadinContextStartupInitializer` implementation instead of defining the clear contract for the abstraction layer (this is exactly done in Vertx : every initializer impl class is used directly and classes are passed to it per its nature).

Technically you should invent something like  `HandlesTypes`  for `VaadinContextStartupInitializer` here and this is strongly connected to the my initial question about this refactoring.

flow-test-generic/src/main/java/com/vaadin/flow/testutil/ClassesSerializableTest.java
denis-anisimov: The test fails because the interfaces don't extend `Serializable`.
To avoid the test failure you should either make the classes `Serializable`  or add them to the exclusion list.
The latter is not always the correct solution.

But it's up to the every implementation class to decide whether it should be `Serializable` or not : `ServletContainerInitializer` doesn't extend `Serializable`  so no need to require that from `VaadinContextStartupInitializer`.

denis-anisimov (approved):

flow-server/src/main/java/com/vaadin/flow/server/startup/VaadinContextStartupInitializer.java
jhult: I'm fine with the current Javadocs.

To answer your question, for [VertxVaadin](https://github.com/mcollovati/vertx-vaadin), `HandlesTypes` is still used as a means to find the necessary classes (since most classes come from core Vaadin which uses `HandlesTypes` or addons which also use `HandlesTypes`.

jhult:
Thanks @denis-anisimov for all your comments and effort reviewing. I'm glad we were able to get this merged.

Issue #9759, opened 2021-01-07
add configurable limit on number of heartbeat retries


HermanBovens:
Currently, heartbeats can be enabled (with a configurable interval) or disabled completely.

As described in https://stackoverflow.com/questions/65602379/how-to-limit-the-number-of-times-vaadin-14-tries-to-send-heartbeat-requests, there are cases where you'd like to have heartbeats (to ensure detection of UIs that should be closed) but don't want them to retry endlessly, such as [due to this mod_auth_openidc issue](https://github.com/zmartzone/mod_auth_openidc/issues/331).  We currently had to disable heartbeats completely to work around this, but this is not ideal.

So I suggest to add a config option `maxHeartbeatRetries` to be able to limit the number of retries.  After the limit is reached, the client should stop sending any requests.

Legioth:
There's also the `closeIdleSessions` which makes heartbeats stop after a while.

HermanBovens:
@Legioth  Correct me if I'm wrong, but I think `closeIdleSessions` only works if the heartbeats actually reach the server.  The problem here is that when the max session time (OpenID) is reached, the heartbeats won't reach the server anymore.  The server will detect that no heartbeats or other requests are coming in and will clean up, but the client will keep trying as far as I know.

Legioth:
I believe the `Vaadin-Refresh` token is also considered for the heartbeat responses. Making the authentication proxy include that token in the intercepted response might thus be one way of making the client reset itself to e.g. a login form after the session timeout.

knoobie:
@HermanBovens your problem description reminds me of a  problem I was facing with V8 and Spring Boot. I leveraged the internal implementation of `DefaultConnectionHandler` and created a custom ErrorHandler for clients that send HEARTBEAT calls when the server session was already expired. 


![image](https://user-images.githubusercontent.com/3968629/103889221-5a24e680-50e6-11eb-8dd5-9f194d38973f.png)


@Legioth  the PR that changes the DefaultConnectionHandler doesn't look like it was ported to flow. 


Framework: https://github.com/vaadin/framework/blob/master/client/src/main/java/com/vaadin/client/communication/DefaultConnectionStateHandler.java#L149

Flow: https://github.com/vaadin/flow/blob/74e62ebb32cbf6a1aa986ec9aa547ce00b19ee24/flow-client/src/main/java/com/vaadin/client/communication/DefaultConnectionStateHandler.java#L137

HermanBovens:
[...]

PR #9784, opened 2021-01-11
chore: framework#11859 Expired session: use 403 Forbidden instead of 410

knoobie:
relates to #9759 

forward port of https://github.com/vaadin/framework/pull/11859

Not sure about the `chore` label. 

Edit: I updated the issues to the full path for easier navigation.


pleku (approved):
Thanks. I think this should be referred to as a fix, not a chore in-it-self. I'll handle that for the merge.

Issue #9773, opened 2021-01-08
Windows: TaskGeneratedBootstrapTest fails with faulty expectation


caalador:
TaskGeneratedBootstrapTest expects strings with `\n`, but this fails on windows as the actual generated newline is `\r\n`

PR #9790, opened 2021-01-11
test: Use portable line separators in TaskGeneratedBootstrapTest

mshabarov:
Fixes: https://github.com/vaadin/flow/issues/9773


flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrapTest.java
```diff
@@ -67,7 +66,8 @@ public void setUp() throws Exception {
     public void should_importTargetIndexTS() throws ExecutionFailedException {
         taskGenerateBootstrap.execute();
         String content = taskGenerateBootstrap.getFileContent();
-        Assert.assertTrue(content.contains("import '../../target/index';\n"));
+        Assert.assertTrue(content.contains(
+                "import '../../target/index';" + System.lineSeparator()));
```
caalador, line 70: In this test the line separator is not all that important. It would be enough to check that it has the code.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrap.java
```diff
@@ -54,10 +54,11 @@
     @Override
     protected String getFileContent() {
         List<String> lines = new ArrayList<>();
-        lines.add(String.format("import '%s';%n", getIndexTsEntryPath()));
+        lines.add(String.format("import '%s';%s", getIndexTsEntryPath(),
+                System.lineSeparator()));
```
caalador, line 58: This is not the same as it was except more complicated to read.
```
By using %n in your format string, you tell Java to use the value returned by System. getProperty("line. separator")
```

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrapTest.java
```diff
@@ -76,18 +76,24 @@ public void should_importFrontendIndexTS()
         new File(frontendFolder, INDEX_TS).createNewFile();
         taskGenerateBootstrap.execute();
         String content = taskGenerateBootstrap.getFileContent();
-        Assert.assertTrue(content.contains("import '../index';\n"));
+        Assert.assertTrue(content
+                .contains("import '../index';" + System.lineSeparator()));
```
caalador, line 80: Also here the line separator is extra.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrapTest.java
```diff
@@ -76,18 +76,24 @@ public void should_importFrontendIndexTS()
...
-            "applyTheme(document);\n"));
+        //@formatter:off
+        Assert.assertTrue(content.contains("import '../../target/index';"
+                + System.lineSeparator()
+                + System.lineSeparator()
+                + "import { applyTheme } from '@vaadin/flow-frontend/themes/theme-generated';"
+                + System.lineSeparator()
+                + "applyTheme(document);"
+                + System.lineSeparator()));
+        //@formatter:on
```
caalador, line 96: This could also be update to check just that the expected lines exist see `TsFormTest::should_generate_FormModels` for inspiration.
This will also help in figuring out what is wrong instead of just getting a expected true was false.
Also we don't need to care about lineSeparators.

caalador (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrapTest.java
mshabarov: Yep, not needed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrap.java
mshabarov: Right! Reverted

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrapTest.java
mshabarov: Removed

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrapTest.java
mshabarov: Right! Replaced by the expected lines check.
I was thinking about extracting the file content check code from `should_generate_FormModels` to a common place, but haven't found an appropriate way for it.

caalador (approved):

PR #9812, opened 2021-01-14
fix: Only create needed folders

caalador:
When copying files for custom theme
only create directories that contain file
to be copied.

Generate directories with mkdirp instead
of fs.mkdir as it is more reliable.

### DIFF
--- a/flow-server/src/main/resources/plugins/application-theme-plugin/theme-copy.js
+++ b/flow-server/src/main/resources/plugins/application-theme-plugin/theme-copy.js
@@ -21,6 +21,7 @@
 const fs = require('fs');
 const path = require('path');
 const glob = require('glob');
+const mkdirp = require('mkdirp');
 
 const ignoredFileExtensions = [".css", ".js", ".json"];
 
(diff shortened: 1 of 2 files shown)


denis-anisimov (approved):

mshabarov (approved):

flow-server/src/main/resources/plugins/application-theme-plugin/theme-copy.js
```diff
@@ -35,38 +36,59 @@ const ignoredFileExtensions = [".css", ".js", ".json"];
...
-
     if (fs.statSync(fileToCopy).isDirectory()) {
-      if (!fs.existsSync(targetFile)) {
-        fs.mkdirSync(targetFile, {recursive: true});
+      logger.debug("Going through directory", fileToCopy);
+      const result = collectFolders(fileToCopy, logger);
+      if (result.files.length > 0) {
+        collection.directories.push(fileToCopy);
+        logger.debug("Adding directory", fileToCopy);
+        collection.directories.push.apply(collection.directories, result.directories);
```
mshabarov, line 81: Is this necessary? I thought that the directories listed in `result.directories` should have been already added in the inner calls of `collectFolders`

caalador: We don't send the object for the loop we only collect the recursive results.

mshabarov: Ah, okay, `collection` is a local object, so we need to add inner result there. Thanks!

PR #9814, opened 2021-01-14
Use existing getRequestUrl method

jhult:
`((VaadinServletRequest) request).getRequestURL().toString()` is currently used in the `getServiceUrl` method. However, this code already exists as method `getRequestUrl`. Thus, we should use `getRequestUrl`.


caalador (approved):

mshabarov:
Cherry picked to `2.4`

Issue #9687, opened 2020-12-17
Move theme-generated.js from the target/flow-frontend/themes folder to frontend/generated


pleku:
Follow up to #9678, this should be done so that the active theme can be imported in a Fusion similarly regardless of the origin of the theme: in-project `frontend/themes/my-theme` would be imported the same way when it is inside a `.jar`.

caalador:
Should it be better to have it in `target/flow-frontend/themes/` as then it's definable as `@vaadin/flow-frontend/themes/my-theme/my-theme.generated.js`

pleku:
Yes most likely the desired option for now would be to just have to do 
```ts
import { theme } from './generated/theme';

@customElement('main-layout')
export class MainLayout extends LitElement {
  static styles = [theme];
}
```
and then there is no need to worry about what the currently active theme actually is.

pleku:
Related discussion in `https://vaadin.slack.com/archives/CDNTYMS9H/p1608641893104400?thread_ts=1608219489.062900&cid=CDNTYMS9H` and this issues should match bullet 1) meaning the file is moved to `generated/theme.js`. 2) and 3) are done in #9684

caalador:
This has a problem when we have a parent theme also. the parent theme can not also be generated/theme.js are we talking about the same generated file as the main entrypoint for theme is target/flow-frontend/themes/theme-generated.js and the XYZ.generated.js file in the respective theme folders define the theme imports which could be n´moved to under target somewhere so they wouldn't be seen.

Artur-:
`generated/theme.js` is just an import for the active theme

PR #9820, opened 2021-01-15
refactor!: Move theme-generated.js to frontend/generated

mshabarov:
Renames `theme-generated.js` to `theme.js` and moves it to `frontend/generated` folder.

Fixes: https://github.com/vaadin/flow/issues/9687

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -591,7 +591,8 @@ public class NodeTasks implements FallibleCommand {
                     builder.webpackGeneratedTemplate,
                     new File(builder.generatedFolder, IMPORTS_NAME),
                     builder.useDeprecatedV14Bootstrapping,
-                    builder.flowResourcesFolder, pwaConfiguration));
+                    builder.flowResourcesFolder, pwaConfiguration,
+                    builder.connectClientTsApiFolder));
         }
 
         if (builder.enableImportsUpdate) {
@@ -604,7 +605,8 @@ public class NodeTasks implements FallibleCommand {
 
             commands.add(new TaskUpdateThemeImport(builder.npmFolder,
                 frontendDependencies.getThemeDefinition(),
-                    builder.frontendDirectory));
+                    builder.frontendDirectory,
+                    builder.connectClientTsApiFolder));
         }
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -604,7 +604,7 @@ public class NodeTasks implements FallibleCommand {
                             builder.tokenFileData, builder.enablePnpm));
 
             commands.add(new TaskUpdateThemeImport(builder.npmFolder,
-                frontendDependencies.getThemeDefinition(),
+                frontendDependencies.getThemeDefinition(), // NOSONAR
                     builder.frontendDirectory,
                     builder.connectClientTsApiFolder));
         }
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
@@ -51,15 +51,12 @@ public class TaskUpdateThemeImport implements FallibleCommand {
     private final File npmFolder;
 
     TaskUpdateThemeImport(File npmFolder, ThemeDefinition theme,
-            File frontendDirectory) {
+            File frontendDirectory, File frontendGeneratedFolder) {
         this.theme = theme;
         this.frontendDirectory = frontendDirectory;
         this.npmFolder = npmFolder;
-        themeImportFile = new File(
-                new File(frontendDirectory, FrontendUtils.GENERATED),
-                THEME_IMPORTS_NAME);
-        themeImportFileDefinition = new File(
-                new File(frontendDirectory, FrontendUtils.GENERATED),
+        themeImportFile = new File(frontendGeneratedFolder, THEME_IMPORTS_NAME);
+        themeImportFileDefinition = new File(frontendGeneratedFolder,
                 THEME_IMPORTS_D_TS_NAME);
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
@@ -33,8 +33,8 @@ import static com.vaadin.flow.server.frontend.FrontendUtils.THEME_IMPORTS_D_TS_N
 import static com.vaadin.flow.server.frontend.FrontendUtils.THEME_IMPORTS_NAME;
 
 /**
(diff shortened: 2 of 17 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
```diff
@@ -52,18 +52,15 @@
...
-                new File(flowFrontend, APPLICATION_THEME_ROOT),
-                THEME_IMPORTS_NAME);
-        themeImportFileDefinition = new File(
-            new File(flowFrontend, APPLICATION_THEME_ROOT),
-            THEME_IMPORTS_D_TS_NAME);
         this.theme = theme;
         this.frontendDirectory = frontendDirectory;
         this.npmFolder = npmFolder;
+        themeImportFile = new File(
+                new File(frontendDirectory, FrontendUtils.GENERATED),
```
caalador, line 59: These should probably target `generatedTsFolder` (or `connect.generated` property) to be in the correct folder.

Asking fusion team about why it's there but only used for one instance.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -140,22 +140,24 @@ function getThemeProperties(themeFolder) {
...
- * @param themeFolder theme folder where flow generates 'theme-generated.js'
+ * @param frontendFolder folder in project containing frontend resources and
+ * theme auto-generated files
  * file and copies local and jar resource frontend files
  * @returns {string} current theme name
  */
-function extractThemeName(themeFolder) {
-  if (!themeFolder) {
-    throw new Error("Couldn't extract theme name from 'theme-generated.js'," +
+function extractThemeName(frontendFolder) {
```
caalador, line 150: This should be `generatedTsFolder` that is actually definable so we shouldn't make expectations and only check the given folder for `theme.js`
```
    /**
     * The folder where flow will put TS API files for client projects.
     */
    @Parameter(defaultValue = "${project.basedir}/" + FRONTEND + "/generated")
    protected File generatedTsFolder;
```

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -169,12 +169,12 @@ const flowFrontendThemesFolder = path.resolve(flowFrontendFolder, 'themes');
...
   devMode: devMode,
-  // The following matches target/flow-frontend/themes/theme-generated.js
-  // and for theme in JAR that is copied to target/flow-frontend/themes/
-  // and not frontend/themes
+  // The following matches folder 'target/flow-frontend/themes/'
+  // (not 'frontend/themes') for theme in JAR that is copied there
   themeResourceFolder: flowFrontendThemesFolder,
   themeProjectFolders: themeProjectFolders,
   projectStaticAssetsOutputFolder: projectStaticAssetsOutputFolder,
+  projectFrontendFolder: frontendFolder
```
caalador, line 177: This should be the generated folder and probably should be the configurable target folder.

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
```diff
@@ -24,6 +24,9 @@ const { processThemeResources, extractThemeName } = require('./theme-handle');
  *  themeResourceFolder             - theme folder where flow copies local and jar resource frontend files
  *  themeProjectFolders             - array of possible locations for theme folders inside the project
  *  projectStaticAssetsOutputFolder - path to where static assets should be put
+ *  projectFrontendFolder           - folder in project where frontend (flow
+ *  templates, fusion views, styles etc.) files and theme auto-generated
+ *  files are placed
```
caalador, line 29: Should be only the generated folder.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateThemeImport.java
mshabarov: Yeah, It does make sense. And then, `connectTsFolder` property should be available always (not only if TS endpoints are available).

mshabarov: Done. 
`connect.generated` is now used as a target folder for `theme.js`.
`connect.generated` is proposed to be renamed `frontend.generated` as it used not only for connect.
Will create a separate PR for it.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
mshabarov: Done. Definable generated path is used.

flow-server/src/main/resources/webpack.generated.js
mshabarov: Done. Definable generated path is used.

flow-server/src/main/resources/plugins/application-theme-plugin/application-theme-plugin.js
mshabarov: Done. Definable generated path is used.

caalador (dismissed):
This will be a breaking change until #9832 is fixed as the theme is not removed with `mvn clean` and might break the project unexpectedly when the file is left, but the theme is removed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -604,7 +604,7 @@ private NodeTasks(Builder builder) {
                             builder.tokenFileData, builder.enablePnpm));
 
             commands.add(new TaskUpdateThemeImport(builder.npmFolder,
-                frontendDependencies.getThemeDefinition(),
+                frontendDependencies.getThemeDefinition(), // NOSONAR
```
caalador, line 607: This is actually valid that it might not be initialized. I have made a note into the parking lot about this issue with NodeTasks

mshabarov: Why? if `builder.enableImportsUpdate` is `true` then the `frontendDependencies ` is being initialised via `new FrontendDependenciesScanner.FrontendDependenciesScannerFactory().createScanner`

caalador: Beacuse you can (well should be able) to run without webpackTemplate and then if you don't have packages or imports enabled then you will npe.

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
mshabarov: Yeah, if `builder.enablePackagesUpdate || builder.enableImportsUpdate || enableWebpackConfigUpdate` is false meaning that no webpack template, no packages and no import then `frontendDependencies` will be `null` ofc.

But for this particular case the sonar complains we have `if (builder.enableImportsUpdate) {...}`, so the code execution should have gone inside `builder.enablePackagesUpdate || builder.enableImportsUpdate || enableWebpackConfigUpdate` block => initialise `frontendDependencies`.

For other cases/code blocks where nothing among those three flags are checked, there are already a NPE checks inside methods/tasks, like this `protected boolean shouldGenerate() { return frontDeps != null; }`.

So, okay let's look into it more deeply in another ticket.

Issue #9802, opened 2021-01-13
Running from SpringBoot's main is not detecting existing theme jar dependency


taefi:
If the theme jar dependency is added to a SpringBoot application, by running it from the main method, the theme jar dependency is not being detected/loaded, so it fails to start with the following error log:
```
Caused by: com.vaadin.flow.server.ExecutionFailedException: Discovered @Theme annotation with theme name 'custom-theme', but could not find the theme directory in the project or available as a jar dependency. Check if you forgot to create the folder under '/Users/soroosh/workspaces/vaadin/application-theme-demo/reusing-theme-jar-demo/./frontend/themes' or have mistyped the theme or folder name for 'custom-theme'.
	at com.vaadin.flow.server.frontend.TaskUpdateThemeImport.verifyThemeDirectoryExistence(TaskUpdateThemeImport.java:108) ~[flow-server-6.0.0.alpha4.jar:6.0.0.alpha4]
	at com.vaadin.flow.server.frontend.TaskUpdateThemeImport.execute(TaskUpdateThemeImport.java:68) ~[flow-server-6.0.0.alpha4.jar:6.0.0.alpha4]
	at com.vaadin.flow.server.frontend.NodeTasks.execute(NodeTasks.java:672) ~[flow-server-6.0.0.alpha4.jar:6.0.0.alpha4]
	at com.vaadin.flow.server.startup.DevModeInitializer.runNodeTasks(DevModeInitializer.java:472) ~[flow-server-6.0.0.alpha4.jar:6.0.0.alpha4]
	at com.vaadin.flow.server.startup.DevModeInitializer.lambda$initDevModeHandler$0(DevModeInitializer.java:374) ~[flow-server-6.0.0.alpha4.jar:6.0.0.alpha4]
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1736) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) ~[na:na]
	... 1 common frames omitted
```
But running the application using `mvn spring-boot:run` is working completely fine.

Tested against: 
Flow: `6.0.0.alpha4` and `6.0-SNAPSHOT`
Vaadin: `19.0.0.alpha2 and `19.0-SNAPSHOT`
[...]

PR #9823, opened 2021-01-16
fix: Also load default JAR resources for DevModeInitializer

tanbt:
Fixes https://github.com/vaadin/flow/issues/9802


flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
```diff
@@ -37,13 +36,15 @@
 import com.vaadin.flow.router.Route;
 import com.vaadin.flow.server.DevModeHandler;
 import com.vaadin.flow.server.InitParameters;
+import com.vaadin.flow.server.VaadinServletContext;
```
tanbt, line 39: I think my IntelliJ reordering import works correctly in this case.

flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java
```diff
@@ -502,6 +505,8 @@ private static String getBaseDirectoryFallback() {
                 Constants.RESOURCES_FRONTEND_DEFAULT));
         frontendFiles.addAll(getFrontendLocationsFromClassloader(classLoader,
                 Constants.COMPATIBILITY_RESOURCES_FRONTEND_DEFAULT));
+        frontendFiles.addAll(getFrontendLocationsFromClassloader(classLoader,
+            Constants.RESOURCES_JAR_DEFAULT));
```
tanbt, line 509: It could be more specific to load `META-INF/resources/themes` instead of just `META-INF/resources/`.

But `META-INF/resources/` works just fine and could help covering more cases in the future.

caalador: I would target themes as the compatibility resources is actually at `META-INF/resources/frontend` so now they overlap and we would want to do as little as possible.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java
tanbt: Fixed.

caalador (commented):
What do the new tests with loadingJars_allFilesExist actually test?
I couldn't figure it out by looking at the full test class.

tanbt:
Fixed.

caalador (approved):

PR #8533, opened 2020-06-09
Add support for Unit in similar fashion to Vaadin 8

TatuLund:
When we have been doing migration assessments of various customer projects, we have learned that usage of methods like setWidth(float width, Unit unit) in Vaadin 8 are rather popular. In order to make migration of such applications easier and support legacy style of setting sizes of components it is useful to have this feature in flow HasSize as well.

Issue #5854, opened 2019-06-06
Clarify usage of set(Min/Max)(Height/Width) in HasSize


edler-san:
The above mentioned methods only work on block level elements. Improve the documentation for the respective methods.

pleku:
As discussed in the PR #5855, we should:
- [ ] add a new interface for the set&get-min&max-width&height-methods, with a suitable name
- [ ] deprecate the API in `HasSize` with instructions that it might not work for certain components (inline), and instruct for component developers to use the API in the new interface instead
- [ ] update all Vaadin components that actually can support the min&max API to use the new interface instead to make sure the deprecated API is not visible

Anything else ?

pleku:
It probably would be nice to get this resolved in Flow 2.2 (adding deprecations) so that we could remove the methods in 3.0.

pleku:
For now it would be good to:
- remove the methods in `master` branch of Flow
- for next 2.x branch deprecate the methods

PR #9824, opened 2021-01-17
feat: add getters for used Unit to HasSize

knoobie:
Improves the addition to `HasSize` for `Unit`. 

* adds `getWidthUnit` and `getHeightUnit` for better discoverability
  * Open Questions:
    * Optional<Unit> vs. null vs. Unit.PIXELS as default?
    *  Adding those APIs for Min/Max as well? This would increase the amount of methods quite a lot with less gain
* improve performance of `Unit`
  * change streaming to simple for-each 
  * change ordering of enum values, most used (% > px > others) 
  * Open Questions:
    * Should I replace the streaming API in `getSize` as well? 
* adds missing units `vw`, `vh`, `vmin` and `vmax`

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
@@ -15,6 +15,8 @@
  */
 package com.vaadin.flow.component;
 
+import java.util.Optional;
+
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.dom.ElementConstants;
 
@@ -165,6 +167,15 @@ public interface HasSize extends HasElement {
     }
 
     /**
+     * Gets the width unit defined for the component.
+     *
+     * @return the width unit which has been set for the component
+     */
+    default Optional<Unit> getWidthUnit() {
+        return Unit.getUnit(getWidth());
+    }
+
+    /**
      * Sets the height of the component.
      * <p>
      * The height should be in a format understood by the browser, e.g. "100px"
@@ -294,6 +305,15 @@ public interface HasSize extends HasElement {
      */
     default String getMaxHeight() {
         return getElement().getStyle().get(ElementConstants.STYLE_MAX_HEIGHT);
+    }
+
+    /**
+     * Gets the height unit defined for the component.
+     *
+     * @return the height unit which has been set for the component
+     */
+    default Optional<Unit> getHeightUnit() {
+        return Unit.getUnit(getHeight());
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
@@ -167,9 +167,10 @@ public interface HasSize extends HasElement {
     }
 
     /**
-     * Gets the width unit defined for the component.
+     * Gets the width unit of the component, if defined.
      *
-     * @return the width unit which has been set for the component
+     * @return an optional width unit for the component, or an empty
+     *         optional if no width unit has been set
      */
     default Optional<Unit> getWidthUnit() {
         return Unit.getUnit(getWidth());
@@ -308,9 +309,10 @@ public interface HasSize extends HasElement {
     }
 
     /**
-     * Gets the height unit defined for the component.
+     * Gets the height unit of the component, if defined.
      *
-     * @return the height unit which has been set for the component
+     * @return an optional height unit for the component, or an empty
+     *         optional if no height unit has been set
      */
     default Optional<Unit> getHeightUnit() {
         return Unit.getUnit(getHeight());
--- a/flow-server/src/main/java/com/vaadin/flow/component/Unit.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Unit.java
@@ -23,17 +23,14 @@ import java.util.stream.Stream;
  */
 public enum Unit {
     /**
+     * Unit code representing in percentage of the containing element
+     * defined by terminal.
+     */
+    PERCENTAGE("%"),
+    /**
      * Unit code representing pixels.
      */
     PIXELS("px"),
-    /**
-     * Unit code representing points (1/72nd of an inch).
-     */
-    POINTS("pt"),
-    /**
-     * Unit code representing picas (12 points).
-     */
-    PICAS("pc"),
     /**
      * Unit code representing the font-size of the root font.
      */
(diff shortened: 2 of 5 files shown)


flow-server/src/test/java/com/vaadin/flow/component/HasSizeTest.java
```diff
@@ -168,4 +168,27 @@ public void setSizeUndefined() {
...
+        Assert.assertFalse(component.getWidthUnit().isPresent());
+    }
+
+    @Test
+    public void getHeightUnit() {
+        HasSizeComponent component = new HasSizeComponent();
+        Assert.assertFalse(component.getHeightUnit().isPresent());
+
+        component.setHeight("10%");
+        Assert.assertEquals(Unit.PERCENTAGE, component.getHeightUnit().get());
```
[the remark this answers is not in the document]
knoobie: Same for both comments: I can understand your concern and I would never do this in production code. I only used it in tests to make the test short and readable. Depending on the question above, this has to be changed anyway. Gonna wait for that before I tackle this.

flow-server/src/main/java/com/vaadin/flow/component/Unit.java
```diff
@@ -91,10 +107,16 @@ public String toString() {
      * @return A Optional unit.
      */
     public static Optional<Unit> getUnit(String cssSize) {
-        if (cssSize == null) {
-             throw new IllegalArgumentException("The parameter can't be null");
+        if (cssSize == null || cssSize.isEmpty()) {
```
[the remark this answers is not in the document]
knoobie: `trim` is rarely used within the whole code base - I can easily add it, if that's something to improve the code and serve a purpose, but I can't really see a use case where this method is called with a blank string or possibility.

flow-server/src/test/java/com/vaadin/flow/component/UnitTest.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
```
knoobie, line 2: Thanks for the header! I copied another test class in the package.. and got the only one without a copyright header. Damn it :)  

See: https://github.com/vaadin/flow/blob/master/flow-server/src/test/java/com/vaadin/flow/component/HasThemeTest.java

taefi: Thanks for the note about the missing license header in that Test class. I'll fix it.

taefi:
Hi and thanks for the contribution. 

My opinion about some of the questions:

> Optional vs. null vs. Unit.PIXELS as default?
- I would avoid `Unit.PIXELS` as default since IMO it creates more confusion. Between, `null` and `Optional.empty` I vote for the current implementation. 

> Adding those APIs for Min/Max as well? This would increase the amount of methods quite a lot with less gain
- To be honest, the already added APIs fills like double edge sword! It feels right to get `Unit` of the `size` from the same API that we can get the size itself. One the other hand, these methods are also seems like on-liner utility methods that we can define a lot of them. Since this is a base interface, we need to be cautious about adding so many methods to it.

knoobie:
@taefi Thanks for your opinion! 

> I would avoid Unit.PIXELS...

The idea was based on the default implementation of V8 where PIXELS was the default instead of null. Personally I'm not a huge fan of Optional (until java 9) but I used it because it's often used in flow for optional values. 

> To be honest, the already added APIs fills like double edge sword...

The whole addition was based on the comment of Ben in https://github.com/vaadin/flow/pull/8533#issuecomment-755279268 - meaning having a "official" API inside the interface is easier to find for the developer and probably working ;) 

Additionally, the HasSize interface is probably going to be cleaned up in the future once #5854 is done.

flow-server/src/main/java/com/vaadin/flow/component/Unit.java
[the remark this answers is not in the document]
taefi: Fair point!

flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
```diff
@@ -296,6 +307,15 @@ default String getMaxHeight() {
         return getElement().getStyle().get(ElementConstants.STYLE_MAX_HEIGHT);
     }
 
+    /**
+     * Gets the height unit defined for the component.
```
pleku, line 311: Or what ? This javadoc doesn't tell me when the optional might be empty

flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
```diff
@@ -164,6 +166,15 @@ default String getMaxWidth() {
         return getElement().getStyle().get(ElementConstants.STYLE_MAX_WIDTH);
     }
 
+    /**
+     * Gets the width unit defined for the component.
```
pleku, line 170: Or what ? This javadoc doesn't tell me when the optional might be empty

pleku (changes requested):
Sorry that it has taken a while. Only two clarifications for the javadocs would be required

flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
knoobie: Updated both javadocs based on other `Optional<?>` methods like `Element::getShadowRoot`

pleku (dismissed):

taefi (approved):

pleku:
Sigh, test failure should not be related in any way, restarted the build.

knoobie:
@pleku  I'm not sure about the target/2.5 and target/2.6 label because the original PR for this isn't backported or? (https://github.com/vaadin/flow/pull/9824)

pleku:
> @pleku I'm not sure about the target/2.5 and target/2.6 label because the original PR for this isn't backported or? (#8533)

It is now as the manual cherry-pick PRs will be merged before this, then the bot will backport this too.

But @knoobie can you please rebase this on top of the `master` as it now failed for the silly Chrome version test.

knoobie:
@pleku thanks! Rebased and waiting for the build process to do it's job :)

fluorumlabs:
Unit `ch` is missing from the unit list. While rarely used, I believe it's more popular than, let's say, centimeters :)

knoobie:
@fluorumlabs Thanks! Found a reason to trigger a build restart :)

pleku (approved):

pleku (approved):

PR #9836, opened 2021-01-18
refactor!: Rename connect.generated as it now used for flow either

mshabarov:
Frontend/generated folder is now used by flow to generate a theme files there, so It's not limited to a Fusion's connect functionality. 
Thus, this PR proposes to rename it to something more common. 
As `vaadin.frontend.generated.folder` parameter name already exists, the new name is proposed to be `project.frontend.generated` to not mix them up with each other.

Related-to: https://github.com/vaadin/flow/issues/9687


haijian-vaadin:
The build is failing, also I tried to run a fusion application downloaded from start.vaadin.com with this PR snapshot build (6.0.pr-9836-SNAPSHOT). the page cannot be loaded and there is js error in the console as attached screenshot
Screenshot 2021-01-19 at 21 44 34

mshabarov:
Thanks, @haijian-vaadin! I'll double check why it doesn't work.

mshabarov:
@haijian-vaadin is that issue with lumo version you mentioned above looks like this issue https://vaadin.slack.com/archives/C3TGRP4HY/p1610972324035100 ? If so, you need to clean up your .m2 folder to fetch updated flow snapshots.

mshabarov:
Checked with fresh fusion app from SVC and it works as expected with no errors in console.

haijian-vaadin (approved):

PR #9842, opened 2021-01-19
Inject Lumo CSS into shadow roots

### LINKED ISSUE #9765: The default Lumo theme should work also in Fusion apps
When using the default Lumo theme by adding `@Theme(themeClass = Lumo.class)` annotation or by omitting the `@Theme` annotation altogether, I want the global Lumo styles to be applied to the light DOM of the app automatically, so that the way to use the default theme is consistent with that for [custom themes](https://github.com/vaadin/flow/issues/9684).

This is not the case with Vaadin 19.0-SNAPSHOT.

**Steps to reproduce**:
1. create a new Vaadin 19 / Fusion project with [https://start.vaadin.com/?preset=fusion](https://start.vaadin.com/?preset=fusion)
   (or with Vaadin CLI: `npx @vaadin/cli init --fusion --pre fusion-theme`)
2. change the Vaadin version to `19.0-SNAPSHOT` in `pom.xml`
3. stop using the custom `fusion-theme` theme (comment out `@Theme(value = "fusion-theme")` in `Application.java`)
4. remove the workarounds that include the Lumo theme manually:
   - in `index.html` remove:
     ```html
     <custom-style>
       <style include="lumo-color lumo-typography"></style>
     </custom-style>
     ```
   - in `index.ts` remove:
     ```ts
     import { applyTheme } from '@vaadin/flow-frontend/themes/theme-generated';
     applyTheme(document);
     ```
 - run the app with `mvn` and see that the Lumo theme is NOT applied to the page light DOM: in the browser console evaluate `window.getComputedStyle(document.body)['font-family']` and see that it has the deafult value (`"serif"`)

**Expected behavior**
The Lumo theme should be applied to the body of the document, including the font family. That is, `window.getComputedStyle(document.body)['font-family']` should evaluate to `"-apple-system, BlinkMacSystemFont, \"Roboto\", \"Segoe UI\", Helvetica, Arial, sans-serif, \"Apple Color Emoji\", \"Segoe UI Emoji\", \"Segoe UI Symbol\""`.

### LINKED ISSUE #9767: The theming API should be usable inside shadow roots
When using a Vaadin application theme from TypeScript, I want to have an API to include the application theme styles inside the shadow root of a custom component, so that there is a convenient way to use the theme styles in my custom views and I do not have to resort to workarounds like those in the projects generated by start.vaadin.com:
```ts
import { CSSModule } from '@vaadin/flow-frontend/css-utils';

@customElement('main-view')
export class MainView extends LitElement {
  static get styles() {
    return [
      CSSModule('lumo-typography'),
      CSSModule('lumo-color'),
      CSSModule('app-layout'),
      css`...`
    ]
  }
}
```

**Expected behavior:**
1. use the default Lumo theme (by omitting the `@Theme` annotation) OR use a custom application theme by adding a `@Theme("my-theme")` annotation
2. add a view component with a shadow DOM (e.g. `<main-view>`)
   ```ts
   @customElement('main-view')
   export class MainView extends LitElement {
     static styles = css`...`;
   }
   ```
3. see that inside the shadow root of the `main-view` component the Lumo theme is available only through inherited CSS properties, but not as explicity CSS rules. That is,
   - all the `--lumo-xxx` custom properties cascade through the light DOM from the `html` element and are available
   - all the built-in CSS properties with the `inherit` default value inherit the values from the shadow host element, e.g. `font-size: var(--lumo-font-size, var(--lumo-font-size-m));`
   - CSS rules of the Lumo theme that apply in the light DOM do not exist in the shadow DOM (e.g. `[theme~="dark"] { --lumo-base-color: hsl(214, 35%, 21%); }` or `h6 { color: var(--lumo-header-text-color); }`)

   Example: add a `<h1>Header</h1>` tag into the shadow root and check its effective `color` and `font-size` style properties (`window.getComputedStyle($0)['color']`). Without Lumo they have the default values: `rgb(0, 0, 0)` and `32px`.
4. import the application theme into the shadow root of the `main-view` component:
   ```ts
   import { theme } from './generated/theme';

   @customElement('main-view')
   export class MainView extends LitElement {
     static styles = [
       theme,
       css`...`
     ];
   }
   ```
5. see that inside the shadow root of the `main-view` component every aspect of the Lumo theme is available. That is, all Lumo custom properties are reset to their default values, all CSS rules defined by Lumo are available.
   
   Example: add a `<h1>Header</h1>` tag into the shadow root and check its effective `color` and `font-size` style properties (`window.getComputedStyle($0)['color']`). With Lumo they have the Lumo-defined default values: `rgb(25, 36, 52)` and `40px`.

Artur-:
Adds support for
"lumoImports": ["color", "typography", "badge", ... ]
to theme.json to override the default ["color", "typography"]

package.json is added to the test to be able to get only one version of @polymer/polymer

If you do not have an application theme specified, Lumo imports are injected by generated-flow-imports.js like before.
If you have an application theme, Lumo imports are injected as part of applying the theme.

In both cases, Lumo imports are added to the beginning of <head> so they are considered before any other styles, such as @CssImport:ed stylesheets. This allows overriding Lumo CSS rules no matter how you add the CSS.

Fixes #9765

### DIFF
--- a/flow-tests/test-application-theme/test-theme-reusable/src/main/java/com/vaadin/flow/uitest/ui/theme/TSView.java
+++ b/flow-tests/test-application-theme/test-theme-reusable/src/main/java/com/vaadin/flow/uitest/ui/theme/TSView.java
@@ -1,0 +1,13 @@
+package com.vaadin.flow.uitest.ui.theme;
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.Tag;
+import com.vaadin.flow.component.dependency.JsModule;
+import com.vaadin.flow.router.Route;
+
+@Route("com.vaadin.flow.uitest.ui.theme.TSView")
+@Tag("ts-component")
+@JsModule("./ts-component.ts")
+public class TSView extends Component {
+
+}
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
@@ -59,7 +59,7 @@ import static com.vaadin.flow.server.frontend.FrontendUtils.IMPORTS_NAME;
  */
 public class TaskUpdateImports extends NodeUpdater {
 
-    private static final String THEME_LINE_TPL = "addCssBlock('%s', true);";
+    private static final String THEME_LINE_TPL = "%saddCssBlock('%s', true);";
     private static final String THEME_VARIANT_TPL = "document.documentElement.setAttribute('%s', '%s');";
     // Trim and remove new lines.
     private static final Pattern NEW_LINE_TRIM = Pattern
@@ -133,12 +133,17 @@ public class TaskUpdateImports extends NodeUpdater {
             ThemeDefinition themeDef = getThemeDefinition();
 
             if (theme != null) {
+                boolean hasApplicationTheme = themeDef != null && !"".equals(themeDef.getName());
+                // There is no application theme in use, write theme includes here. Otherwise they are written by the theme
                 if (!theme.getHeaderInlineContents().isEmpty()) {
                     lines.add("");
+                    if (hasApplicationTheme) {
+                        lines.add("// Handled in the application theme");
+                    }
                     theme.getHeaderInlineContents()
-                            .forEach(html -> addLines(lines,
-                                    String.format(THEME_LINE_TPL, NEW_LINE_TRIM
-                                            .matcher(html).replaceAll(""))));
+                    .forEach(html -> addLines(lines,
+                    String.format(THEME_LINE_TPL, hasApplicationTheme ? "// ":"" ,NEW_LINE_TRIM
+                    .matcher(html).replaceAll(""))));
                 }
                 if (themeDef != null) {
                     theme.getHtmlAttributes(themeDef.getVariant()).forEach(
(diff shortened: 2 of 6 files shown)


flow-tests/test-application-theme/test-theme-reusable/src/main/java/com/vaadin/flow/uitest/ui/theme/TSView.java
```diff
@@ -0,0 +1,13 @@
...
+import com.vaadin.flow.component.Tag;
+import com.vaadin.flow.component.dependency.JsModule;
+import com.vaadin.flow.router.Route;
+
+@Route("com.vaadin.flow.uitest.ui.theme.TSView")
+@Tag("ts-component")
+@JsModule("./ts-component.ts")
+public class TSView extends Component {
+
+}
```
caalador, line 13: Missing new line.

caalador (changes requested):
I noted that in the package.json you forced polymer to 3.4.1, but as things should work out of the box by building the project I made the fix #9859 
I guess the issue comes from vaadin-text-field that has the dependencies
    "@polymer/iron-component-page": "^4.0.0",
    "@polymer/iron-form": "^3.0.0",
    "@polymer/iron-test-helpers": "^3.0.0",
That bring in polymer 3.4.1.
On the other hand if there is no issues then we should probably update our polymer version to the latest.

package.json files should be removed

Artur-:
Rebased on top of #9859 and removed package.json files

caalador:
Looks good, I'll fix it in with the package fix asap.

Artur-:
Rebased on master

caalador (approved):

PR #9848, opened 2021-01-20
feat: support current pnpm versions

### LINKED ISSUE #9719: pnpm install should just work in Vaadin projects
After I read in the Vaadin docs that Vaadin uses `pnpm` as the client-side package manager (instead of `npm`), I want to add dependencies with `pnpm install`. That allows me to use any examples from the `pnpm` documentation in my Vaadin projects "as-is". At the moment that does not work because Vaadin requires a particular version of pnpm (4.5.0) with a particular configuration (`--shamefully-hoist=true`). Any attempt to use a different pnpm version or a different configuration leads to an incompatible `node_modules` folder structure (i.e. the app start / production build fails).

**Steps to reproduce**:
1. create a new Vaadin 18 / Fusion project with [https://start.vaadin.com/?preset=fusion](https://start.vaadin.com/?preset=fusion)
   (or with Vaadin CLI: `npx @vaadin/cli init --fusion pnpm-install-test`)
2. trigger generation of the 'special' Vaadin npm packages (like `@vaadin/flow-frontend` and `@vaadin/form`): `mvn vaadin:prepare-frontend`
3. install a dependency with the latest pnpm (5.13.7): `npx pnpm install mobx`
4. run `mvn` to start the app in the dev mode
5. see a run-time error:
   ```
   2020-12-23 15:24:40.679  INFO 79222 --- [         task-1] c.v.flow.server.frontend.FrontendTools   : using '/Users/viktor/.vaadin/node/node /Users/viktor/.vaadin/node_modules/pnpm/bin/pnpm.js' for frontend package installation
   2020-12-23 15:24:41.705 ERROR 79222 --- [         task-1] c.v.f.s.frontend.TaskUpdatePackages      : Command `/Users/viktor/.vaadin/node/node /Users/viktor/.vaadin/node_modules/pnpm/bin/pnpm.js --shamefully-hoist=true install` failed:
   Using hooks from: /Users/viktor/Downloads/pnpm-install-test/pnpmfile.jsreadPackage hook is declared. Manifests of dependencies might get overridden ERROR  This "node_modules" folder was created without the --shamefully-hoist option. You must remove that option, or else "pnpm install --force" to recreate the "node_modules" folder.
   2020-12-23 15:24:41.705 ERROR 79222 --- [         task-1] c.v.f.s.frontend.TaskUpdatePackages      : >>> Dependency ERROR. Check that all required dependencies are deployed in pnpm repositories.
   ```
6. delete `node_modules` and install a dependency with pnpm 4.5.0, but without `--shamefully-hoist=true`
   ```bash
   rm -rf node_modules
   npx -p pnpm@4.5.0 pnpm install mobx
   ```
7. See the same error

**Workaround**:
Use `npx -p pnpm@4.5.0 pnpm --shamefully-hoist=true install [dependency]` to add an npm dependency.

joheriks:
Remove pnpm-lock.yaml when package.json dependencies 
are updated to work around pnpm/pnpm#2587. Invoke pnpm
via `npx` instead of installing into `~/.vaadin/node_modules`.

Part of #9719

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -15,23 +15,18 @@
  */
 package com.vaadin.flow.server.frontend;
 
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
-import java.io.InputStreamReader;
 import java.io.UncheckedIOException;
 import java.net.URI;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Properties;
 import java.util.function.Supplier;
-import java.util.stream.Collectors;
 
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
@@ -42,9 +37,6 @@ import com.vaadin.flow.server.frontend.FrontendUtils.UnknownVersionException;
 import com.vaadin.flow.server.frontend.installer.InstallationException;
 import com.vaadin.flow.server.frontend.installer.NodeInstaller;
 import com.vaadin.flow.server.frontend.installer.ProxyConfig;
-
-import elemental.json.Json;
-import elemental.json.JsonObject;
 
 /**
  * Provides access to frontend tools (node.js and npm, pnpm) and optionally
@@ -96,11 +88,6 @@ public class FrontendTools {
                     new FrontendVersion("6.11.1"),
                     new FrontendVersion("6.11.2"));
 
-    static final String PNPM_INSTALLED_BY_NPM_FOLDER = "node_modules/pnpm/";
-
-    static final String PNPM_INSTALLED_BY_NPM = PNPM_INSTALLED_BY_NPM_FOLDER
-            + "bin/pnpm.js";
-
     private static final int SUPPORTED_NODE_MAJOR_VERSION = 10;
     private static final int SUPPORTED_NODE_MINOR_VERSION = 0;
     private static final int SUPPORTED_NPM_MAJOR_VERSION = 5;
@@ -109,9 +96,6 @@ public class FrontendTools {
     private static final int SHOULD_WORK_NODE_MINOR_VERSION = 9;
     private static final int SHOULD_WORK_NPM_MAJOR_VERSION = 5;
     private static final int SHOULD_WORK_NPM_MINOR_VERSION = 5;
-
-    public static final int SUPPORTED_PNPM_MAJOR_VERSION = 4;
-    public static final int SUPPORTED_PNPM_MINOR_VERSION = 4;
 
     private static final FrontendVersion SUPPORTED_NODE_VERSION = new FrontendVersion(
             SUPPORTED_NODE_MAJOR_VERSION, SUPPORTED_NODE_MINOR_VERSION);
@@ -132,6 +116,9 @@ public class FrontendTools {
     static final String SYSTEM_NOPROXY_PROPERTY_KEY = "NOPROXY";
     static final String SYSTEM_HTTPS_PROXY_PROPERTY_KEY = "HTTPS_PROXY";
     static final String SYSTEM_HTTP_PROXY_PROPERTY_KEY = "HTTP_PROXY";
+
+    public static final int SUPPORTED_PNPM_MAJOR_VERSION = 5;
+    public static final int SUPPORTED_PNPM_MINOR_VERSION = 0;
 
     private static final FrontendVersion SUPPORTED_PNPM_VERSION = new FrontendVersion(
             SUPPORTED_PNPM_MAJOR_VERSION, SUPPORTED_PNPM_MINOR_VERSION);
@@ -272,23 +259,14 @@ public class FrontendTools {
      *
      * @return the list of all commands in sequence that need to be executed to
      *         have pnpm running
-     * @see #getPnpmExecutable(String, boolean)
      */
     public List<String> getPnpmExecutable() {
-        Supplier<List<String>> result = doEnsurePnpm();
-        List<String> pnpmCommand = result.get();
+        List<String> pnpmCommand = getSuitablePnpm();
         if (!pnpmCommand.isEmpty()) {
+            pnpmCommand = new ArrayList<>(pnpmCommand);
             pnpmCommand.add("--shamefully-hoist=true");
         }
         return pnpmCommand;
-    }
-
-    /**
-     * Ensure that pnpm tool is available and install it if it's not.
-     *
-     */
-    public void ensurePnpm() {
-        doEnsurePnpm();
     }
 
     /**
@@ -324,47 +302,6 @@ public class FrontendTools {
             getLogger().warn("Error checking if npm is new enough", e);
         }
 
-    }
-
-    /**
-     * Locate <code>pnpm</code> executable if it's possible.
-     * <p>
-     * In case the tool is not found either {@link IllegalStateException} is
-     * thrown or an empty list is returned depending on {@code failOnAbsence}
-     * value.
-     *
-     * @param dir
-     *            the directory to search local pnpm script
-     *
-     * @param failOnAbsence
-     *            if {@code true} throws IllegalStateException if tool is not
-     *            found, if {@code false} return an empty list if tool is not
-     *            found
-     *
-     * @return the list of all commands in sequence that need to be executed to
-     *         have pnpm running
-     */
-    protected List<String> getPnpmExecutable(String dir,
-            boolean failOnAbsence) {
-        // First try local pnpm JS script if it exists
-        List<String> returnCommand = new ArrayList<>();
-        Optional<File> localPnpmScript = getLocalPnpmScript(dir);
-        if (localPnpmScript.isPresent()) {
-            returnCommand.add(getNodeExecutable());
-            returnCommand.add(localPnpmScript.get().getAbsolutePath());
-        } else {
-            // Otherwise look for regular `pnpm`
-            String command = FrontendUtils.isWindows() ? "pnpm.cmd" : "pnpm";
-            if (failOnAbsence) {
-                returnCommand.add(
-                        getExecutable(command, null, false).getAbsolutePath());
-            } else {
-                returnCommand.addAll(frontendToolsLocator.tryLocateTool(command)
-                        .map(File::getPath).map(Collections::singletonList)
-                        .orElse(Collections.emptyList()));
-            }
-        }
-        return returnCommand;
     }
 
     /**
@@ -425,64 +362,6 @@ public class FrontendTools {
                     npmVersion.getFullVersion(),
                     "by updating your global npm installation with `npm install -g npm@latest`");
             throw new IllegalStateException(badNpmVersion);
-        }
-    }
-
-    private List<String> ensurePnpm(String dir) {
-        List<String> pnpm = getSuitablePnpm(dir);
-        if (pnpm.isEmpty()) {
-            // copy the current content of package.json file to a temporary
-            // location
-            File packageJson = new File(dir, "package.json");
-            File tempFile = null;
-            boolean packageJsonExists = packageJson.canRead();
-            if (packageJsonExists) {
-                try {
-                    tempFile = File.createTempFile("package", "json");
-                    FileUtils.copyFile(packageJson, tempFile);
-                } catch (IOException exception) {
-                    throw new IllegalStateException(
-                            "Couldn't make a copy of package.json file",
-                            exception);
-                }
-                packageJson.delete();
-            }
-            try {
-                JsonObject pkgJson = Json.createObject();
-                pkgJson.put("name", "temp");
-                pkgJson.put("license", "UNLICENSED");
-                pkgJson.put("repository", "npm/npm");
-                pkgJson.put("description",
-                        "Temporary package for pnpm installation");
-                FileUtils.writeLines(packageJson,
-                        Collections.singletonList(pkgJson.toJson()));
-                JsonObject lockJson = Json.createObject();
-                lockJson.put("lockfileVersion", 1);
-                FileUtils.writeLines(new File(dir, "package-lock.json"),
-                        Collections.singletonList(lockJson.toJson()));
-            } catch (IOException e) {
-                getLogger().warn("Couldn't create temporary package.json");
-            }
-            // install pnpm locally using npm
-            installPnpm(dir, getNpmExecutable(false));
-
-            // remove package-lock.json which contains pnpm as a dependency.
-            new File(dir, "package-lock.json").delete();
-
-            if (packageJsonExists && tempFile != null) {
-                // return back the original package.json
-                try {
-                    FileUtils.copyFile(tempFile, packageJson);
-                } catch (IOException exception) {
-                    throw new IllegalStateException(
-                            "Couldn't restore package.json file back",
-                            exception);
-                }
-                tempFile.delete();
-            }
-            return Collections.emptyList();
-        } else {
-            return pnpm;
         }
     }
 
@@ -628,38 +507,9 @@ public class FrontendTools {
         return null;
     }
 
-    private Supplier<List<String>> doEnsurePnpm() {
-        List<String> path = getSuitablePnpm(baseDir);
-        if (!path.isEmpty()) {
-            getLogger().trace(
-                    "Found installed pnpm (globally or inside the '{}' project dir)",
-                    baseDir);
-            return () -> path;
-        }
-        String alternativeDir = getAlternativeDir();
-        List<String> pnpm = ensurePnpm(alternativeDir);
-        if (pnpm.isEmpty()) {
-            return () -> getPnpmExecutable(alternativeDir, true);
-        } else {
-            getLogger().info(
-                    "Using pnpm command installed in the '{}' directory",
-                    alternativeDir);
-            return () -> pnpm;
-        }
-    }
-
     private List<String> getNpmExecutable(boolean removePnpmLock) {
-        List<String> returnCommand = getNpmScriptCommand(baseDir);
-        if (returnCommand.isEmpty()) {
-            returnCommand = getNpmScriptCommand(getAlternativeDir());
-        }
-
-        if (returnCommand.isEmpty()) {
-            // Otherwise look for regular `npm`
-            String command = FrontendUtils.isWindows() ? "npm.cmd" : "npm";
-            returnCommand
-                    .add(getExecutable(command, null, true).getAbsolutePath());
-        }
+        List<String> returnCommand = new ArrayList<>(
+                getNpmCliToolExecutable("npm"));
         returnCommand.add("--no-update-notifier");
         returnCommand.add("--no-audit");
         returnCommand.add("--scripts-prepend-node-path=true");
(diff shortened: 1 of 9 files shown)


haijian-vaadin (dismissed):
The change looks good to me. 
I also tried to verify with the pr snapshot build, it works great!
Would be good to have some input from Flowers.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -683,11 +525,30 @@ private String getNonNull(String... valueArray) {
         return returnCommand;
     }
 
-    private List<String> getNpmScriptCommand(String dir) {
+    private List<String> getNpmCliToolExecutable(String toolName) {
+        assert "npm".equals(toolName) || "npx".equals(toolName);
```
caalador, line 529: As tool name can/should only be npm or npx could we just have an `enum Tool` from where we take the accepted tool. Then we wouldn't typo it in mistake either.

In any case the assert should have a message on what was wrong.

joheriks: Done.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -697,108 +575,55 @@ private String getNonNull(String... valueArray) {
...
+            // pass version specifier to npx
+            pnpmCommand.set(pnpmCommand.size() - 1,
+                    "pnpm@" + DEFAULT_PNPM_VERSION);
+        }
+        if (!ignoreVersionChecks && !checkPnpmVersion(pnpmCommand)) {
+            // This should not happen unless there is a too old pnpm installed
+            // locally into the project's node_modules folder overriding the
+            // package version specifier passed to npx.
+            throw new IllegalStateException(
+                    "unable to locate pnpm of supported version");
```
pleku, line 597: I don't think this comment is in any way helpful if I have installed a pnpm version that used to work in my project folder - _what should I do (ignore version or install new) and what version should I use_ ?

mshabarov, line 597: Can we suggest any actions the dev should do to fix that, when he see this message? Delete the too old version from node_modules, for instance. Or maybe restart the application. 
Also would be good to explicitly mention the supported version here in the message.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -697,108 +575,55 @@ private String getNonNull(String... valueArray) {
...
+                getNpmCliToolExecutable(NpmCliTool.NPX));
+        pnpmCommand.add("--yes");
+        pnpmCommand.add("--quiet");
+        pnpmCommand.add("pnpm");
+
+        // check version of default pnpm
+        if (!ignoreVersionChecks && !checkPnpmVersion(pnpmCommand)) {
+            // pass version specifier to npx
+            pnpmCommand.set(pnpmCommand.size() - 1,
+                    "pnpm@" + DEFAULT_PNPM_VERSION);
```
pleku, line 590: Why ? How does passing a default version to the _version check_ make a difference here ?
I've understood that this code is executed in case an old version was detected or no version was detected, but I can't figure out how this makes a difference, please elaborate (preferably in code)

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -697,108 +575,55 @@ private String getNonNull(String... valueArray) {
...
-
-        ProcessBuilder builder = FrontendUtils.createProcessBuilder(command);
-        builder.environment().put("ADBLOCK", "1");
-        builder.directory(new File(dir));
-
-        builder.redirectInput(ProcessBuilder.Redirect.INHERIT);
-        builder.redirectError(ProcessBuilder.Redirect.INHERIT);
-
-        Process process = null;
+    private boolean checkPnpmVersion(List<String> pnpmCommand) {
```
pleku, line 605: The method name could be clearer as my first thought was - _check pnpm version is what_? The return value has the answer `versionNewEnough`.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -697,108 +575,55 @@ private String getNonNull(String... valueArray) {
...
-                process.destroyForcibly();
+            List<String> versionCmd = new ArrayList<>(pnpmCommand);
+            versionCmd.add("--version"); // NOSONAR
+            FrontendVersion pnpmVersion = FrontendUtils
+                    .getVersion("pnpm", versionCmd);
+            boolean versionNewEnough = FrontendUtils
+                    .isVersionAtLeast(pnpmVersion, SUPPORTED_PNPM_VERSION);
+            if (!versionNewEnough) {
+                getLogger()
+                        .warn("pnpm at '{}' is version {} which is not supported (expected >={})",
```
pleku, line 615: I'm not sure if this warning is needed in case there is another check with the _default version_ and then everything will be OK ?

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -697,108 +575,55 @@ private String getNonNull(String... valueArray) {
...
+                                String.join(" ", pnpmCommand),
+                                pnpmVersion.getFullVersion(),
+                                SUPPORTED_PNPM_VERSION.getFullVersion());
             }
+            return versionNewEnough;
+        } catch (UnknownVersionException e) {
+            getLogger().warn("error checking pnpm version", e);
```
pleku, line 622: Not sure if we could give more details to the user on what to do next if running into this.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -92,9 +92,21 @@ public void execute() {
...
+                    // With pnpm dependency versions are pinned via pnpmfile.js
+                    // (instead of @vaadin/vaadin-shrinkwrap). When updating
+                    // a dependency in package.json, the old version may be
+                    // left in the pnpm-lock.yaml file, causing duplicate
+                    // dependencies. Work around this issue by deleting
+                    // pnpm-lock.yaml ("pnpm install" will re-generate).
+                    // For details, see:
+                    // https://github.com/pnpm/pnpm/issues/2587
+                    // https://github.com/vaadin/flow/issues/9719
+                    deletePnpmLockFile();
```
pleku, line 108: According to https://github.com/pnpm/benchmarks-of-javascript-package-managers#lots-of-files there is a significant difference in the benchmarks when the lockfile is still present. I presume that being able to use a newer pnpm version however makes things faster ?
Are there any other alternative ideas on what to do instead of deleting the lockfile or an issue open describing further improvements for this ? Or will 9719 be updated and kept open ?

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
```diff
@@ -496,40 +459,38 @@ public void should_useHomeNpmFirst() throws Exception {
...
-        Assume.assumeFalse(
-                tools.getNodeExecutable().isEmpty());
+    public void getSuitablePnpm_incompatibleDefaultVersionInstalled_rejected() throws Exception {
+        createStubNode(false, true, baseDir);
         createFakePnpm("4.5.0");
-        List<String> pnpmCommand = tools.getSuitablePnpm(baseDir);
-        Assert.assertNotEquals("expected pnpm version 4.5.0 accepted", 0,
-                pnpmCommand.size());
+        List<String> pnpmCommand = tools.getSuitablePnpm();
+        Assert.assertEquals("expected pnpm version 4.5.0 rejected",
```
pleku, line 466: So the previously installed default version will start failing starting from this version ?
Yet there is no mention of this in the linked issue or PR. Probably should be highlighted somehow.

My expectation would be that 
- if pnpm was installed by Vaadin, it would just update it for me
- if I have installed pnpm myself, things would fail and tell me to update to a new version or let Vaadin install it (and how to)

pleku (changes requested):
Comments ready and testing next with windows and an svc project.

Let me know if you want to discuss comment directly via voice/slack to speedup things

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -683,11 +545,27 @@ private String getNonNull(String... valueArray) {
...
-    private List<String> getNpmScriptCommand(String dir) {
+    private List<String> getNpmCliToolExecutable(NpmCliTool cliTool) {
+        List<String> returnCommand = getNpmScriptCommand(baseDir,
+                cliTool.getScript());
+        if (returnCommand.isEmpty()) {
+            returnCommand = getNpmScriptCommand(getAlternativeDir(),
+                    cliTool.getScript());
+        }
+        if (returnCommand.isEmpty()) {
+            // Otherwise look for regular `npm`/`npx`
```
mshabarov, line 556: So what will be returned in this case? 
I expect `getNpmCliToolExecutable` return `npm-cli` always, but this block of code looks for `npm` and return it.
I'm sure this does work, but could you briefly explain how? :)

mshabarov (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
```diff
@@ -496,40 +459,38 @@ public void should_useHomeNpmFirst() throws Exception {
...
-        tools = new FrontendTools(baseDir, () -> vaadinHomeDir,
-                "v12.10.0", new File(baseDir).toURI(), true);
-        Assume.assumeFalse(
-                tools.getNodeExecutable().isEmpty());
-
-        createFakePnpm("5.5.0");
-        List<String> pnpmCommand = tools.getSuitablePnpm(baseDir);
-        Assert.assertNotEquals("expected pnpm version 5.5.0 accepted", 0,
-                pnpmCommand.size());
+        tools = new FrontendTools(baseDir, () -> vaadinHomeDir, "v14.15.4",
```
mshabarov, line 482: Use DEFAULT_NODE_VERSION constant instead.

pleku:
When running a locally built version of this with a v19 SVC project, I get the following logs:
- no `target`, `pnpm-lock` & `package.json` & `node_modules`
- old pnpm 4.5.0 installation in `/.vaadin`
- windows 10
```
...
2021-01-26 13:59:18.079  INFO 10480 --- [         task-1] c.v.f.s.frontend.TaskUpdatePackages      : writing file C:\Users\Pekka\workspaces\flow2\svc-test\package.json.
2021-01-26 13:59:18.088  INFO 10480 --- [         task-1] c.v.f.s.frontend.TaskUpdatePackages      : Running `pnpm install` to resolve and optionally download frontend dependencies. This may take a moment, please stand by...
2021-01-26 13:59:18.097  INFO 10480 --- [         task-1] c.v.f.s.frontend.TaskUpdatePackages      : Generated pnpmfile hook file: 'C:\Users\Pekka\workspaces\flow2\svc-test\pnpmfile.js'
2
...
2021-01-26 13:59:28.900  INFO 10480 --- [         task-1] c.v.flow.server.frontend.FrontendTools   : using 'C:\Users\Pekka\.vaadin\node\node.exe C:\Users\Pekka\.vaadin\node\node_modules\npm\bin\npx-cli.js --yes --quiet pnpm' for frontend package installation
2021-01-26 14:00:25.862  INFO 10480 --- [  restartedMain] c.e.a.data.generator.DataGenerator       : Generated demo data
```
This seems a bit off to me - the first doesn't seem to do anything (?) and the latter is not descriptive as it takes quite long.

mshabarov:
Tested on `Mac` with a fresh `Java+Html` project downloaded from SVC.
`pnpm` version already installed in `~/.vaadin` is `4.5.0`.

I changed `chokidar: "^3.5.0"` to `3.5.1` in `package.json` and restarted the app.
`pnpm-lock.yaml` has been deleted and re-generated as expected.

Application works fine.

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
pleku: Right, so after testing and then reading further the linked issue comments it seems that:
When I have an existing pnpm 4.5.0 installed to `.vaadin`, it is logged that it would be used but it is actually not as it is determined to be too old, instead `npx` is installing and running the default version instead but I didn't see any logs for this. 

The side effect of this is that I don't really know which version of pnpm is used, but maybe that is not so important. But the false log of "using pnpm from .vaadin" could be removed ?

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -697,108 +575,55 @@ private String getNonNull(String... valueArray) {
...
+    List<String> getSuitablePnpm() {
+        // do NOT modify the order of flags, as it changes the behavior of npx
+        List<String> pnpmCommand = new ArrayList<>(
+                getNpmCliToolExecutable(NpmCliTool.NPX));
+        pnpmCommand.add("--yes");
+        pnpmCommand.add("--quiet");
+        pnpmCommand.add("pnpm");
+
+        // check version of default pnpm
+        if (!ignoreVersionChecks && !checkPnpmVersion(pnpmCommand)) {
```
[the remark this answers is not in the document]
joheriks: Rewrote the whole method in functional style, which also made it shorter. Slightly changed the behavior so that even if `ignoreVersionChecks` is true it runs `npx pnpm .... --version` (but ignores the reported version) so that it at least fails fast also in this case if the pnpm command can't be run at all (should only happen if something is seriously broken).

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
joheriks: The statement mutated `pnpmCommand` which was not only passed to the version check but returned from the method and to be used for the actual pnpm command. However, using the array like this was maybe bad style; now rewritten using only immutables.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
joheriks: Changed to `validatePnpmVersion`. Also the `ignoreVersionChecks` flag is now checked in this method instead.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
joheriks: `warn` -> `info`

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
joheriks: Added the full command to log so that user can test run it manually in console.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
joheriks: Deletion should happen only when `package.json` is updated due to platform updates, so quite rarely. Test both suggestions given here https://github.com/pnpm/pnpm/issues/2587#issuecomment-636169750 and deleting the lockfile followed by `install` was not perceived slower than the alternative `update --depth Infinity`.

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
joheriks: It shouldn't start failing. Node.JS and npm/npx scripts auto-installed by Vaadin go into `~/.vaadin/node` (scripts in `~/.vaadin/node/node_modules`), while `pnpm` installed by Vaadin currently goes into `~/.vaadin/node_modules/pnpm`. So the `npx pnpm` command (`~/.vaadin/node/node ~/.vaadin/node/node_modules/npm/bin/npx-cli.js pnpm` if the user has relied on Vaadin-installed Node.js) will run a `pnpm` version independent of old Vaadin-installed `pnpm`. It will fetch the latest version and cache it in `~/.npm`. So for the first bullet point, there is no need to do anything. For the second, the user (globally) installed version will be too old so it will try to get a supported version (5.15.1) instead.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -137,13 +117,12 @@
     static final String SYSTEM_HTTPS_PROXY_PROPERTY_KEY = "HTTPS_PROXY";
     static final String SYSTEM_HTTP_PROXY_PROPERTY_KEY = "HTTP_PROXY";
 
+    public static final int SUPPORTED_PNPM_MAJOR_VERSION = 5;
+    public static final int SUPPORTED_PNPM_MINOR_VERSION = 0;
```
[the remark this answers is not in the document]
joheriks: Done.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -280,25 +279,16 @@ public String forceAlternativeNodeExecutable() {
      *
      * @return the list of all commands in sequence that need to be executed to
      *         have pnpm running
-     * @see #getPnpmExecutable(String, boolean)
      */
     public List<String> getPnpmExecutable() {
-        Supplier<List<String>> result = doEnsurePnpm();
-        List<String> pnpmCommand = result.get();
+        List<String> pnpmCommand = getSuitablePnpm();
         if (!pnpmCommand.isEmpty()) {
```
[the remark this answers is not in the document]
joheriks: It would have propagated the empty command and higher in the call chain. However, this could not happen as `ensurePnpm` will now always either throw an exception or return a non-empty list. Added an `assert` here to clarify this.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
joheriks: Done.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
joheriks: Added some comments to describe the search logic. Note that it may need to be changed at some point (see #9863).

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
joheriks: Done.

mshabarov (approved):

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
joheriks: It is true that the `pnpm` version that it chooses to use is a bit non-deterministic due to how `npx pnpm` selects the version (on the other hand, this also held true to some degree when we defaulted to global `pnpm` in command search path; in most cases this would be the latest version of `pnpm` at the time the user installed it). 

The thing to perhaps decide is what should be the supported way of using a specific `pnpm` version if the user has this requirement, and document it. One way that will already work after this PR is to install global Node.js and then do e.g. `npm i -g pnpm@5.15.2`. Then this version will be invoked when doing `npx pnpm install`.

Followup question is whether it should also be supported to fiddle with project-local installations into `project/node_modules` and/or `project/node/node_modules` (with `frontend-maven-plugin`). This is where it gets hairy, with a package manager installing itself into the managed packages...

platosha (dismissed):

pleku (approved):

PR #9852, opened 2021-01-20
Fix concurrency issue

Artur-:
The cache is only protected from multiple accesses at the same time by the same user. Many users can use it simultaneously


mshabarov (approved):

mshabarov:
Cherry picked to `2.4`

PR #9859, opened 2021-01-21
fix: Lock versions to what is set in package.json

caalador:
Even without platform versions we should
manage the used versions by using exactly
what is defined in the package.json
as if we had a platform versions file.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
@@ -44,6 +44,8 @@ import elemental.json.Json;
 import elemental.json.JsonObject;
 import static com.vaadin.flow.server.frontend.FrontendUtils.FLOW_NPM_PACKAGE_NAME;
 import static com.vaadin.flow.server.frontend.FrontendUtils.commandToString;
+import static com.vaadin.flow.server.frontend.NodeUpdater.DEPENDENCIES;
+import static com.vaadin.flow.server.frontend.NodeUpdater.DEV_DEPENDENCIES;
 import static com.vaadin.flow.server.frontend.NodeUpdater.HASH_KEY;
 import static com.vaadin.flow.server.frontend.NodeUpdater.VAADIN_DEP_KEY;
 import static elemental.json.impl.JsonUtil.stringify;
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -202,6 +204,36 @@ protected String generateVersionsJson() throws IOException {
...
+    /**
+     * If we do not have the platform versions to lock we should lock any
+     * versions in the package.json so we do not get multiple versions
+     * for defined packages.
+     *
+     * @return versions Json based on package.json
+     * @throws IOException
+     *     If reading package.json fails
+     */
+    private JsonObject generateVersionsFormPackageJson() throws IOException {
```
mshabarov, line 216:
```suggestion
    private JsonObject generateVersionsFromPackageJson() throws IOException {
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -188,7 +190,7 @@ protected String generateVersionsJson() throws IOException {
 
             JsonObject versionsJson = getVersions(content);
             if (versionsJson == null) {
-                return null;
+                versionsJson = generateVersionsFormPackageJson();
```
mshabarov, line 193:
```suggestion
                versionsJson = generateVersionsFromPackageJson();
```

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
```diff
@@ -440,11 +444,17 @@ public void generateVersionsJson_userDefinedVersions_versionOnlyPinnedForNotAdde
...
         TaskRunNpmInstall task = createTask();
 
-        Assert.assertNull(task.generateVersionsJson());
+        final String versions = task.generateVersionsJson();
+        Assert.assertNotNull(versions);
+
+        File generatedVersionsFile = new File(versions);
+        final JsonObject versionsJson = Json.parse(FileUtils
+            .readFileToString(generatedVersionsFile, StandardCharsets.UTF_8));
+        Assert.assertEquals("{}", versionsJson.toJson());
```
mshabarov, line 457: It's also worth to test that some dependencies are copied from `package.json`.
Not only the empty file case.

mshabarov (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
mshabarov: Other examples contain that writings into package.json:

```
        // Write package json file: dialog doesn't pin its Overlay version which
        // is transitive dependency.
        FileUtils.write(packageJson,
                "{\"dependencies\": {"
                        + "\"@vaadin/vaadin-dialog\": \"2.2.1\"}}",
                StandardCharsets.UTF_8);
```

This test could write-test-clean.

caalador: Added test.

caalador: package json is written into a temporaryFolder which is created new for each test so there's no need to specifically clean the package.json file.

mshabarov (dismissed):

mshabarov (approved):

mshabarov:
Cherry picked to `2.4`

PR #9872, opened 2021-01-24
chore!: mark Abstract class with appropriate modifier

manolo:


pleku:
While this may feel like a chore, it is also a breaking change. We can merge it, but it has to wait for 7.0 now.

pleku (commented):
OKing for now but not approving so it is not merged at this point.

manolo:
> While this may feel like a chore, it is also a breaking change

Added the `!` mark to the type part in the commit message so as it's considered a breaking change

pleku (approved):

PR #9875, opened 2021-01-25
Use time-constant comparison for CSRF tokens

Legioth:
This hardens the framework against a theoretical timing attack based on
comparing how quickly a request with an invalid CSRF token is rejected.


Legioth:
No tests since this is functionality equivalent to the previous implementation aside from timing differences that would be very fragile to verify in an automated test.

taefi (approved):

taefi:
> No tests since this is functionality equivalent to the previous implementation aside from timing differences that would be very fragile to verify in an automated test.

I approved the changes. 

The only drawback is about sacrificing the performance just a little bit for preventing a timing attack, which I assume not be that huge due to the size of the Token.

mshabarov:
Cherry picked to `2.5`

PR #9878, opened 2021-01-26
Backport lit support.

### LINKED ISSUE #8862: Backport lit-html to 2.5
Backport the lit html feature to 2.4.
Also remember taking in the documentation.

caalador:
Fixes #8862


flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/BundleParser.java
```diff
@@ -81,9 +84,38 @@
...
+     * <code>render\(\)[\s]*\{</code> finds the template getter method
+     * <p>
+     * <code>[\s]*return[\s]*html([\`|\'|\"])</code> finds the return statement
+     * and captures the used string character
+     * <p>
+     * </p>
+     * <code>([\s\S]*)\1;[\s]*\}</code> captures all text until we encounter the
+     * end character with <code>;}</code> e.g. <code>';}</code>
+     */
+    private static final Pattern LIT_TEMPLATE_PATTERN = Pattern.compile(
```
denis-anisimov, line 116: Is it the same as in the `master` ?
Looks quite weird that polymer template module contains Lit template methods.

flow-polymer-template/src/main/java/com/vaadin/flow/component/polymertemplate/rpc/PublishedServerEventHandlerRpcHandler.java
```diff
@@ -0,0 +1,370 @@
...
+ * @author Vaadin Ltd
+ * @since 1.0
+ *
+ * @deprecated Polymer template support is deprecated - we recommend you to use
+ *             {@code LitTemplate} instead. Read more details from <a href=
+ *             "https://vaadin.com/blog/future-of-html-templates-in-vaadin">the
+ *             Vaadin blog.</a>
+ */
+@Deprecated
+public class PublishedServerEventHandlerRpcHandler implements Serializable {
```
denis-anisimov, line 72: This is a breaking change though the class is internal

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/IdCollector.java
```diff
@@ -24,6 +24,7 @@
 
 import org.jsoup.nodes.Element;
 
+import com.vaadin.flow.component.Component;
```
denis-anisimov, line 27: This class is not moved to the polymer template module though it's in the polymer package.

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/IdMapper.java
```diff
@@ -38,15 +38,15 @@
 
     private final HashMap<String, Element> registeredElementIdToInjected = new HashMap<>();
 
-    private AbstractTemplate<?> template;
+    private Component template;
```
denis-anisimov, line 41: The package of this class is polymer template but it's in the flow-server.

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/PolymerTemplate.java
```diff
@@ -22,6 +22,8 @@
 import com.vaadin.flow.component.HasComponents;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.component.dependency.HtmlImport;
+import com.vaadin.flow.component.dependency.JsModule;
```
denis-anisimov, line 25: This class is not moved to the polymer template module

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/TemplateParser.java
```diff
@@ -74,6 +74,37 @@ public Element getTemplateElement() {
         }
     }
 
+    /**
+     * Template parser factory.
```
denis-anisimov, line 78: Polymer template class is in the flow-server module

denis-anisimov (changes requested):
There is a polymer module created in the PR.
It's not clear whether this should be done at all or in this PR.
At the same time polymer related classes are in both modules: polymer and flow-server.

If polymer needs to be extracted then everything should go into polymer module.
If not then moved classes should stay in flow-server

flow-bom/pom.xml
```diff
@@ -18,6 +18,16 @@
                 <artifactId>flow-server</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>com.vaadin</groupId>
+                <artifactId>flow-polymer-template</artifactId>
```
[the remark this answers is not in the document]
caalador: I was expecting not to and keep Polymer in the server module.

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/BundleParser.java
caalador: It's not there in master, but it was changed in a deprecate polymer commit which is not applicable here.

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/Id.java
```diff
@@ -25,17 +25,17 @@
...
- * to as a value for a field inside your {@link PolymerTemplate} class.
+ * to as a value for a field inside your {@code LitTemplate} class.
  * <p>
  * Here is a Java sample:
  *
  * <pre>
  * <code>
  * &#64;Tag("details")
- * public class Details extends PolymerTemplate&lt;EditorModel&gt;{
+ * public class Details extends LitTemplate {
```
denis-anisimov, line 38: I think _this_ `Id`  annotation should be deprecated and not changed.
And additional `Id` in the package outside of `polymertemplate` should be introduced in the same was is it has been done in the `master`.
This is not a breaking change since you still may use the old `Id`  with polymer templates.
But in Lit template one has to use the new `Id` .
The open question should new `Id` also work in polymer templates ? 
I think yes to avoid confusions with imports.
That will require to add logic into the polymer template id collector to recognize both `Id`  (which already in the `master`).

But it's quite weird to use `Id` in Lit template with `Id`  being in the polymer template package.

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/InjectablePolymerElementInitializer.java
```diff
@@ -0,0 +1,96 @@
...
+import com.vaadin.flow.dom.Element;
+
+/**
+ * Initialize a polymer template element with data.
+ * 
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+public class InjectablePolymerElementInitializer
```
denis-anisimov, line 32: That functionality introduces attributes/properties initialization for PolymerTemplate (from the template content into Java).
Should we backport it at all?
Should we backport it in this PR ?

Also I believe the same functionality is for LitTemplate. 
Should it also be backported along with LitTemplates ?
LitTemplate is a new functionality overall so i think yes, we should backport it along with LitTemplates.
But for PolymerTemplate it may cause unexpected changes in the existing applications.

flow-server/src/test/java/com/vaadin/flow/component/polymertemplate/InjectablePolymerElementInitializerTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.Test;
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.HasStyle;
+import com.vaadin.flow.component.Tag;
+import com.vaadin.flow.component.template.internal.AbstractInjectableElementInitializer;
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.dom.ElementFactory;
+
+public class InjectablePolymerElementInitializerTest {
```
denis-anisimov, line 30: This test is for Polymer template functionality which sets attribute from the template file content to the Java object.
So should this bebackported
(and the same question is for all the related classes to this functionality)?

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/InjectablePolymerElementInitializer.java
pleku: It is considered a new improvement for the existing polymer templates. This has been communicated clearly last year and I just checked that the feedback has only been about "how this makes an existing add-on useless".

So this is intended.

flow-server/src/test/java/com/vaadin/flow/component/polymertemplate/InjectablePolymerElementInitializerTest.java
pleku: This is like intended

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/Id.java
pleku: The idea is that
1) Nothing related to polymer templates should be deprecated in 14 as it is not going away and we don't want to push people to get worried they have to do something. We have announced however the direction and are now giving them the possibility to migrate gradually within the 14-series and then have easier moving to nextLTS where polymer is deprecated but not removed.
2) NO changes should be needed when going from 14.4 to 14.5 when you have existing polymer templates
3) NO changes should be needed when going from 14.5 with lit templates to 19+, or with polymer templates from 14.5 to 19+

Is there something that cannot be done according to 1-3 with the backporting ?

caalador: Will keep polymer as is, but add the new Id for Lit usage so then there should be no changes needed for current or when updating to 18+

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/InjectablePolymerElementInitializer.java
denis-anisimov: It's not clear for me why this is a part of _lit templates_  backport ?

I don't have a strong objection but it's not really correct to put changes which are not related to the main topic of this backport regardless of confirmed improvement.

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/Id.java
denis-anisimov: It will be confusing If you can't use new `Id` with Polymer templates (didn't get the exact intention from the last comment).

caalador: Polymer template still uses the polymertemplate.Id where as list should use template.Id

denis-anisimov: OK I see from the impl that both Ids can be used in the polymer templates.

flow-server/src/main/java/com/vaadin/flow/component/polymertemplate/InjectablePolymerElementInitializer.java
caalador: 8988 introduced the feature to lit and polymer at the same time.

denis-anisimov: Should 8988 be backported separately then ?

Again : up to you.

denis-anisimov (dismissed):

caalador:
Rebased on branch to get browser fix.

pleku (approved):

Issue #3585, opened 2018-02-21
Use vanilla Atmosphre


Legioth:
Flow has so far been using the same Atmosphere build that is used by Vaadin Framework 8. This is a custom build to avoid a direct dependency on `slf4j` in FW8.

Since Flow already uses the `slf4j` API for logging, there is no point in using a custom build instead of the vanilla upstream version.

jreznot:
Besides that annoying message about commercial support in logs )

heruan:
Is this somehow related to #3251?

Legioth:
@heruan Not that I'm aware of, but on the other hand the root cause of #3251 doesn't seem to fully understood either, so there might still be some unexpected side effects.

Artur-:
One step towards this: #5818

jcgueriaud1:
There are a couple of issues with the push and we are still using an old version of Atmosphere (the java library) and an old version of atmosphere.js (inside VaadinPush.js) for Vaadin 14 and Vaadin 8.
I think one of the main ticket is here: https://github.com/vaadin/flow/issues/7553

I've merged our vaadinPush and the fixes from atmosphere-js. In 
[vaadinPush-patch.js.zip](https://github.com/vaadin/flow/files/5572762/vaadinPush-patch.js.zip)

I tried with Johannes example but it didn't help.

PR #9884, opened 2021-01-27
Use the latest atmosphere-runtime version

mstahv:
Just simple module and version upgrade to get tests run and discussions started on #3585. Quickly looked at our forks and there were some custom changes that should be manually checked that similar fixes are in in the latest official atmosphere as well.


pleku:
Since the build failed the push tests I guess this is not going in until someone starts working on figuring out why those fail - and that task should be put to the backlog so it happens. Closing the PR for now.

PR #9898, opened 2021-01-27
feat(TypeScript): add Frontend path alias

### LINKED ISSUE #9544: Consider adding an alias for the generated frontend assets
Refactoring a frontend project could be easier if we aliased the generated folder with webpack https://webpack.js.org/configuration/resolve/

That way, you could write
`import Company from 'Generated/com/vaadin/tutorial/crm/backend/entity/Company';`

instead of 
`import Company from '../../../../../generated/com/vaadin/tutorial/crm/backend/entity/Company';`

Which would also not break if you move the file around.

platosha:
Fixes #9544


haijian-vaadin (approved):
Wow, it really works nicely!

PR #9914, opened 2021-01-28
fix: Tasks are sorted for execution

### LINKED ISSUE #9893: Order of TS code generation and npm install changed
With Vaadin 19 alpha 3, the order of npm install and TS code generation is:
1. generate TS code
2. npm install

This enables e.g. using the generated TS code in a postinstall script if you so please

With the latest Flow snapshot, the order is suddenly the other way around:
1. npm install
2. generate TS code

Now using TS code in a postinstall script is impossible

If this is caused by #9839 there does not even seem to be a reason to change the order

### Expected behavior
This is a breaking change that should not be there

### Actual behavior
It should behave like before

caalador:
NodeTasks are now sorted before
execution so that they are always
executed in a predefined order.

Fixes #9893

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -20,9 +20,12 @@ import java.io.File;
 import java.io.Serializable;
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.Collection;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.IntStream;
 
 import com.vaadin.flow.di.Lookup;
 import com.vaadin.flow.server.ExecutionFailedException;
@@ -507,7 +510,29 @@ public class NodeTasks implements FallibleCommand {
         }
     }
 
-    private final Collection<FallibleCommand> commands = new ArrayList<>();
+    private final List<FallibleCommand> commands = new ArrayList<>();
+
+    // This list keeps the tasks in order so that they are executed
+    // without depending on when they are added.
+    private static final List<Class<? extends FallibleCommand>> commandOrder = Collections.unmodifiableList(Arrays.asList(
+        TaskGeneratePackageJson.class,
+        TaskGenerateIndexHtml.class,
+        TaskGenerateIndexTs.class,
+        TaskGenerateTsConfig.class,
+        TaskGenerateTsDefinitions.class,
+        TaskGenerateServiceWorker.class,
+        TaskGenerateOpenApi.class,
+        TaskGenerateConnect.class,
+        TaskGenerateBootstrap.class,
+        TaskUpdatePackages.class,
+        TaskRunNpmInstall.class,
+        TaskInstallWebpackPlugins.class,
+        TaskCopyFrontendFiles.class,
+        TaskCopyLocalFrontendFiles.class,
+        TaskUpdateWebpack.class,
+        TaskUpdateImports.class,
+        TaskUpdateThemeImport.class
+    ));
 
     private NodeTasks(Builder builder) {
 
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -507,7 +510,29 @@ public Builder withNodeDownloadRoot(URI nodeDownloadRoot) {
...
+        TaskGenerateConnect.class,
+        TaskGenerateBootstrap.class,
+        TaskUpdatePackages.class,
+        TaskRunNpmInstall.class,
+        TaskInstallWebpackPlugins.class,
+        TaskCopyFrontendFiles.class,
+        TaskCopyLocalFrontendFiles.class,
+        TaskUpdateWebpack.class,
+        TaskUpdateImports.class,
+        TaskUpdateThemeImport.class
```
mshabarov, line 534: Awesome to have this order here, because it does not only guard the application behaviour from reordering, but also gives a clarity to the devs how the tasks should go. Great!

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -557,6 +575,13 @@ private NodeTasks(Builder builder) {
 
         }
 
+        if (builder.createMissingPackageJson) {
```
mshabarov, line 578: Doesn't really matter now, but why is this block has been moved to another place?

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -667,9 +692,38 @@ private FrontendDependenciesScanner getFallbackScanner(Builder builder,
...
+    /**
+     * Find index of command for which it is assignable to.
+     * @param command command to find execution index for
+     * @return index of command or -1 if not available
+     */
+    private int getIndex(FallibleCommand command) {
+        return IntStream.range(0, commandOrder.size())
+                .filter(i -> commandOrder.get(i)
+                        .isAssignableFrom(command.getClass())).findFirst()
+                .orElse(-1);
```
mshabarov, line 727: If some command from resulted command list is not present in the expected ordering command list, then why not throw the exception?
That might happens when someone adds a new Task for execution, but forgets to add it into the ordering list, and the order of this Task matters.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksExecutionTest.java
```diff
@@ -0,0 +1,97 @@
...
+        commandsField.setAccessible(true);
+        final List<FallibleCommand> commands = (List<FallibleCommand>) commandsField
+                .get(nodeTasks);
+
+        Assert.assertEquals("No commands should be added initially, "
+                        + "update mock builder so that we don't automatically add any tasks!",
+                0, commands.size());
+
+        // Assemble the command list with random order
+        commands.addAll(commandsMock);
```
mshabarov, line 54: Hmm, does this really modify the field?

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksExecutionTest.java
```diff
@@ -0,0 +1,97 @@
...
+                executionOrder.add(command);
+                return null;
+            }).when(mock).execute();
+        }
+
+        Collections.shuffle(commands);
+
+        return commands;
+    }
+}
```
mshabarov, line 97:
```suggestion
}

```

mshabarov (commented):
Awesome to have this order here, because it does not only guard the application behaviour from reordering, but also gives a clarity to the devs how the tasks should go. Great!

One more test for catching the case, when some new Task is added for execution, but not added to ordering, should be added IMO (If I got this feature properly).

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksExecutionTest.java
caalador: It should return the actual instance of the List and it's that List value from the field that we are modifying and not the field itself.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksExecutionTest.java
caalador: done

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksExecutionTest.java
mshabarov: Yep, got it. Thanks!

mshabarov (approved):

mshabarov (approved):

mshabarov:
@caalador does this need to be CP'ed? I assume target/5.0 and target/2.5 ?

caalador:
As the cleanup is not cherry picked to any version this should not be picked either.

PR #9931, opened 2021-02-01
fix: Remove old devDependencies

### LINKED ISSUE #9720: unused npm dev dependencies, once added by Vaadin, are never removed
Vaadin has a list of 'required npm dependencies' that are automatically added into `package.json` during the `prepare-frontend` step (by the Vaadin Maven plugin [during the build](https://github.com/vaadin/flow/blob/bbfa1e7e655e35f5c2d280248899eab3e92ac0e7/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojo.java#L137) or by the Vaadin DevMode handler [during the starup](https://github.com/vaadin/flow/blob/bbfa1e7e655e35f5c2d280248899eab3e92ac0e7/flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java#L373-L388)). If a dependency is no longer needed in a later Vaadin version, it's removed from the list of 'required dependencies', so it's no longer automatically added to `package.json` by Vaadin. However, if a dependency was required in an earlier Vaadin version and is not needed anymore in a later Vaadin version, it remains in `package.json` making the `npm install` phase of the build / statup slower and potentially leading to version conflicts.

With Vaadin 19.0.0.alpha1 this is the case with the following dependencies:
 - webpack-babel-multi-target-plugin (removed in 65b43f3 / 3.0.0.beta2)
 - progress-webpack-plugin (added in edc43a1 / 3.1.0.alpha1, removed in 244f81c / 5.0.0.alpha1)
 - terser (added in 1b2b3ba / 3.1.0.alpha1, removed in 08862de / 3.2.0.alpha7)
 - raw-loader (removed in 04b1692 / 5.0.0.alpha1)

After manually removing obsolete dependencies the build time of the default start.vaadin.com V19 Fusion project went down by about 10%
 - [when pnpm downloads everything from the Internet] by ~7 seconds: from 50 to 43 sec
 - [when pnpm takes everything from the local cache] by ~2 seconds: from 17 to 15 sec

**Steps to reproduce**:
1. create a new Vaadin 19 / Fusion project with [https://start.vaadin.com/?preset=fusion](https://start.vaadin.com/?preset=fusion)
   (or with Vaadin CLI: `npx @vaadin/cli init --fusion --pre unused-deps-test`)
2. see that `package.json` contains the 4 obsolete dependencies mentioned above (since this is a project that was sequentially updated through every Vaadin version in history)
2. (optional) clean the local pnpm packages store to simulate a clean dev machine: `rm -rf ~/.pnpm-store`
3. run `mvn` to start the app in the dev mode
4. see in the logs how long the `using '/Users/viktor/.vaadin/node/node /Users/viktor/.vaadin/node_modules/pnpm/bin/pnpm.js' for frontend package installation` step takes
5. clean the `node_modules` folder, remove the 4 obsolete dependencies from `package.json`, run `mvn` again and see how long the same steps takes in this case

caalador:
Clean away also old devDependencies
on Package update.

Fixes #9720

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
@@ -154,8 +154,32 @@ public class TaskUpdatePackages extends NodeUpdater {
                     || !enablePnpm && !ensureReleaseVersion(dependencies);
         }
 
+
+        // Remove obsolete devDependencies
+        dependencyCollection = getDefaultDevDependencies().entrySet().stream()
+                .map(Entry::getKey).collect(Collectors.toList());
+
+        int removedDev = 0;
+        JsonObject devDependencies = packageJson.getObject(DEV_DEPENDENCIES);
+        JsonObject vaadinDevDependencies = packageJson.getObject(VAADIN_DEP_KEY)
+                .getObject(DEV_DEPENDENCIES);
+        if (devDependencies != null) {
+            for (String key : devDependencies.keys()) {
+                if (!dependencyCollection.contains(key)
+                        && vaadinDevDependencies.hasKey(key)) {
+                    devDependencies.remove(key);
+                    vaadinDevDependencies.remove(key);
+                    log().debug("Removed \"{}\".", key);
+                    removedDev++;
+                }
+            }
+        }
+
         if (removed > 0) {
             log().info("Removed {} dependencies", removed);
+        }
+        if(removedDev > 0) {
+            log().info("Removed {} devDependencies", removedDev);
         }
 
         if (doCleanUp) {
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -154,9 +154,33 @@ private boolean updatePackageJsonDependencies(JsonObject packageJson,
...
+                .getObject(DEV_DEPENDENCIES);
+        if (devDependencies != null) {
+            for (String key : devDependencies.keys()) {
+                if (!dependencyCollection.contains(key)
+                        && vaadinDevDependencies.hasKey(key)) {
+                    devDependencies.remove(key);
+                    vaadinDevDependencies.remove(key);
+                    log().debug("Removed \"{}\".", key);
+                    removedDev++;
+                }
```
pleku, line 174: There a bit of code duplication here from the above, not sure if it would help to have one method for removing the obsolete dependencies or if it would cause more bugs...

But one thing though - if dev dependencies are removed with npm, we will not delete package-lock similarly as with normal dependencies above (calling `cleanUp()`) ?

pleku:
Should this be targeting 2.4, 2.5 and 5.0 too ?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
caalador: Clean up is not checked for devDependencies as devDependencies never contain shrinkwrap.

pleku (approved):

PR #9935, opened 2021-02-01
Upgrade loader deps

Artur-:


pleku (approved):

pleku:
Very difficult to say if updating these would break something, but I think it is fine to realize that during beta instead of postponing merging this to a later point.

PR #9936, opened 2021-02-01
Upgrade lit-html + LitElement

Artur-:


pleku (approved):

pleku (approved):

pleku:
Should be fine to include even for 6.0 beta2.

PR #9942, opened 2021-02-02
test: Linux local build fixes

miguelatvaadin:
These commits fix some issues which arise while locally building the project on a linux desktop


flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
```diff
@@ -580,7 +580,9 @@ public void withConverter_writeBackValue() {
         binder.setBean(item);
         rentField.setValue("10");
 
-        assertEquals("€ 10.00", rentField.getValue());
+        //€ 10.00
+        assertEquals("€ " + new DecimalFormat("0.00").format(10)
+                , rentField.getValue());
```
mshabarov, line 585: Comma is on the new line. Format changes made by you via `shift+option+cmd+L` (on mac)

miguelatvaadin: I see. Thanks a lot, Mikhail.

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
```diff
@@ -16,35 +16,66 @@
...
         HttpURLConnection connection = (HttpURLConnection) url.openConnection();
         connection.setRequestMethod("GET");
         int responseCode = connection.getResponseCode();
         Assert.assertEquals("HTTP 403 Forbidden expected for urls with " +
                 "directory change", HttpURLConnection.HTTP_FORBIDDEN, responseCode);
     }
+
+    private void waitForFrontendCompilation(URL url) throws IOException, InterruptedException {
+        boolean frontendCompiled = false;
+        int attemptsRemaining = 100;
```
mshabarov, line 63: I'm not sure we need to wait so long (100 secs) for frontend compilation. Is it better to decrease the wait threshold to some standard value like in TestBench API (20 secs)?

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
```diff
@@ -16,35 +16,66 @@
...
+        while (!frontendCompiled && attemptsRemaining-- > 0) {
```
mshabarov, line 64: If for some reason the frontend compilation is not be completed in a given timeout, then the test execution would go further and fail because of unexpected response code.
Whether it's better to fail fast the test with an appropriate message when it's reach the timeout?
I would help analyse the root cause in the future.

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
miguelatvaadin: Sure. I will decrease it

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
```diff
@@ -16,35 +16,66 @@
...
+            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+            connection.setRequestMethod("GET");
+            int responseCode = connection.getResponseCode();
+            frontendCompiled =  responseCode != HttpServletResponse.SC_OK || !getResponseBody(connection)
```
mshabarov, line 68: Should it be && instead of || ?
I expect that the both conditions should met.

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
miguelatvaadin: It's true. I will change it to throw an exception

mshabarov:
Thanks for fixing the timing issue, @miguelatvaadin ! I was not taking it into consideration when made this test.

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
miguelatvaadin: mmm.... a 403 response code also means that the frontend has been compiled. The problem was that for this url, while the frontend is being compiled, a simple web page (with response code 200) is returned, so we must look for both a 403 response code or a response body without the "plase wait, we are building" message inside the body

mshabarov: Yep, got it! Make sense.

mshabarov (dismissed):

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
```diff
@@ -16,35 +16,71 @@
...
+                Thread.sleep(200);
+            }
+        }
+        if (!frontendCompiled)
+            throw new Exception(
+                    "Timeout while waiting for frontend compilation");
+    }
+
+    private String getResponseBody(HttpURLConnection connection) throws IOException {
+        String body = String.join("\n", IOUtils.readLines(connection.getInputStream(), StandardCharsets.UTF_8));
```
taefi, line 82: I think we had this policy of not using Apahce Commons and etc. because of versioning conflicts and the challenges it creates for the users and should be avoided. However, this is just a reminder here, since this is a test code.

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
```diff
@@ -16,35 +16,71 @@
...
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.stream.Collectors;
 
 import com.vaadin.flow.testutil.ChromeBrowserTest;
+
+import org.apache.commons.io.IOUtils;
 import org.junit.Assert;
 import org.junit.Test;
```
taefi, line 30: I believe `com.vaadin.*` imports should appear _after all other imports_ and before `elemental.*` and `import static` statements. Although the code was like that even before these changes, but it's a chance to fix them here ;)

mshabarov:
Needs a merge with latest `master` to not complain about chrome version.

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
miguelatvaadin: Yes, I saw it in the wiki. Anyway, IOUtils was already used in several points in the project, even in flow-server, so I understand it's not a problem.

flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/UrlValidationIT.java
miguelatvaadin: ok. I will change this. Thanks, Soroosh ;)

miguelatvaadin:
> Needs a merge with latest `master` to not complain about chrome version.

Ok. I go for it. Thanks!

mshabarov (approved):

Issue #9885, opened 2021-01-27
Enable reuse of forks during unit test execution


miguelatvaadin:
### Description of the bug / feature
There are 2 unit tests at the flow project which prevent us from reusing forks at the test phase. As a result a new jvm is started for each unit test class, so it heavily impacts the project build time.

### Minimal reproducible example
Steps to reproduce the error:
1. modify the pom.xml and comment the `<reuseForks>false</reuseForks>` fagment
2. run `mvn clean test -pl flow-server`

### Expected behavior
Build ends with SUCCESS, and we get a 5 minutes build time improvement by reusing forks.

### Actual behavior
Output extract which shows the failing tests:
```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   VaadinServiceTest.should_reported_routing_server:94
[ERROR] Errors: 
[ERROR] DevModeHandlerTest.serveDevModeRequest_prepareTasksThrows_serveDevModeReturnsFalseAndDoesNotThrow:533 » NullPointer
[INFO] 
[ERROR] Tests run: 2859, Failures: 1, Errors: 1, Skipped: 3
```

PR #9943, opened 2021-02-02
build: enable reuse of forks during unit tests

miguelatvaadin:
Several changes which enable reusing forks during the surefire (unit tests) phase. As a result, a new jvm is not started for each unit test class and a 5 minute build time gain is experienced for the full build.

Fixes: #9885

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
@@ -40,11 +40,13 @@ import java.util.concurrent.atomic.AtomicLong;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.mockito.Matchers;
 import org.mockito.Mockito;
 
 import com.vaadin.flow.function.DeploymentConfiguration;
+import com.vaadin.flow.internal.CurrentInstance;
 
 import static com.vaadin.flow.server.Constants.POLYFILLS_DEFAULT_VALUE;
 import static com.vaadin.flow.server.Constants.STATISTICS_JSON_DEFAULT;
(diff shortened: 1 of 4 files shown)


flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
```diff
@@ -91,6 +93,12 @@ protected URLConnection openConnection(URL u) throws IOException {
 
     }
 
+    @BeforeClass
+    public static void beforeClass() {
```
denis-anisimov, line 97: It's executed once for this test class.
What does this solve ?
I would expect this call in `@Before` instead `@BeforeClass` so it's not clear what does this do.

pom.xml
```diff
@@ -491,7 +491,9 @@
                     <version>${maven.surefire.plugin.version}</version>
                     <configuration>
                         <trimStackTrace>false</trimStackTrace>
+                        <!--
```
denis-anisimov, line 494: Comments should either have explanation or removed.

denis-anisimov (changes requested):

pom.xml
miguelatvaadin: no problem. I will definitely remove it. Thanks!

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
miguelatvaadin: This call solves an issue which arises when reusing forks. For some reason some trash is left there by a previous test from another class and, as a result, one of these tests fails so some clearance must be done. It was not a problem when a new jvm was instantiated for each test class, but now it is made visible.

`CurrentInstance.clearAll();` only needs to be called once for the whole class, before all these tests run. That's why I placed it in the `@BeforeClass`.

denis-anisimov: Alright, thanks for explanation.

denis-anisimov (approved):

PR #9952, opened 2021-02-02
feat: add all platform dependencies to package.json

### LINKED ISSUE #9834: `pnpm i` should not require the `--shamefully-hoist` option
When installing a front-end dependency with `pnpm i`, I do not want to have to provide the `--shamefully-hoist` option. Yet I still want that after `pnpm i` (no hoist) the `mvn` command still works normally. With Vaadin 19.0.0.alpha3 this is not the case.

Depends on https://github.com/vaadin/flow/issues/9719

DoD:
 - running `pnpm i [dependency]` works in a Vaadin 19 project downloaded from start.vaadin.com (before and after running `mvn` first)
 - all Flow integration tests pass, all starters work fine, start.vaadin.com projects work fine

**Context:** (more in the [feature design document](https://docs.google.com/document/d/1UQ_JpxRSGBKeK2V48mJmM3B8ZViF6O_5dhwu6bRMguw/edit#heading=h.u357i1tq3rkm))
Why does Vaadin use the --shamefully-hoist flag of pnpm?

It was added to solve issues with using transitive dependencies without explicitly declaring them. 
The `--shamefully-hoist` flag was added in #7059 about a year ago. A comment on the related issues explains the reasoning:

> - pnpm installs packages in a specific way. It doesn't use flat structure for node_modules as npm. It installs the declared dependencies into node_modules but all transitive dependencies are installed into node_modules/.pnpm/registry.npmjs.org/. As a result node_modules contains only small set of declared dependencies and all other transitive deps are installed in a different folder. This is correct structure which should be handled properly. BUT webpack is not able to handle this structure at all. When you run it it fails with error because it tries to find a transitive dependency (e.g. mkdirp ) and it's not able to find it.
>
> - to be able to solve this there is a config parameter shamefully-hoist=true which says pnpm to install deps into node_modules . The parameter works being added into .nmprc (another question : is there a way to avoid using this file?). But when I run webpack I still get an error like this Error: Can't resolve '@vaadin/flow-frontend/dndConnector-es6.js?babel-target=es5' in '/Users/denis/workspace/flow/flow-tests/test-mixed/target/frontend'. Will see what can I do about it.

**Implementation ideas:**
 - just drop the option, if possible
   Drop the option, list the `mkdirp` (and other) implicitly used dependencies explicitly in package.json, mention in the release notes this as a breaking change and recommend deleting the node_modules folder after the next Vaadin version update.
 - if dropping the option is not possible, move it to `.npmrc` so that specifying it explicitly is no longer necessary

joheriks:
- [x] support non-flat `node_modules` hierarchy (fixes #9834).
- [ ] verify that platform upgrades work without vaadin-shrinkwrap


joheriks:
Closing for now, may revisit in context of #9986.

PR #9968, opened 2021-02-04
chore: Recommend lit over polymer

### LINKED ISSUE #8862: Backport lit-html to 2.5

caalador:
Add recommendation to use LitTemplate
instead of PolymerTemplate into polymer
JavaDocs.

Closes #8862


pleku (approved):
Could nitpick that "polymer is not deprecated, polymer support is" but I think people should be able to get the semantics

Issue #9973, opened 2021-02-05
Compatibility mode OSGi tests are broken in 2.5


denis-anisimov:
https://bender.vaadin.com/viewLog.html?buildId=212394&buildTypeId=Flow_Flow25_Nightly_FlowFelixJettyIntegrationTests25&tab=buildLog

Artifacts show that application can't navigate to the route which is expected.
Dev mode page lists all available routes: there are only few routes.
There are plenty of IT views. It means that for some reason route registration is broken.
I'm pretty sure that this is not related to OSGi anyhow: everything worked yesterday.
Most likely this is related to compatibility mode and this is a regression introduced by some backport.

mshabarov:
This issue reproduces with `felix-jetty` configuration no matter whether the compatibility mode is enabled or not.
However, `test-root-context` ITs run successfully even if I enable the compatibility mode.

Another observation is that the test views with a route targets containing an URL parameter (`@OptionalParameter` or `WildcardParameter`) are resolved with a double parameter placeholder, like this:

`com.vaadin.flow.uitest.ui.BackButtonServerRoundTripView/:___url_parameter/:___url_parameter`,

which is wrong, should be:

`com.vaadin.flow.uitest.ui.BackButtonServerRoundTripView/:___url_parameter`.

If I hack the `HasUrlParameterFormat::getTemplate` to only generate a template, if no template has been generated previously, like this:

```java
if (hasUrlParameter(navigationTarget) && !urlBase.contains(PARAMETER)) {
// template generate code
}
```
then the tests are passed and routes registered as expected.

Apparently, the root cause is that the `AbstractRouteRegistry::setRoute` invoked twice for each route target.
Will check that and propose a fix.

PR #10001, opened 2021-02-09
fix: Skip adding url parameter template if already present

mshabarov:
Fixes https://github.com/vaadin/flow/issues/9973

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
@@ -349,13 +349,8 @@ public abstract class AbstractRouteRegistry implements RouteRegistry {
     private void configureWithFullTemplate(String path,
             Class<? extends Component> navigationTarget,
             SerializableBiConsumer<ConfigureRoutes, String> templateConfiguration) {
-        if (HasUrlParameterFormat.hasUrlParameter(navigationTarget)
-                && !HasUrlParameterFormat.hasUrlParameterTemplate(path)) {
-            path = HasUrlParameterFormat.getTemplate(path, navigationTarget);
-        }
-        String finalPath = path;
         configure(configuration -> templateConfiguration.accept(configuration,
-                finalPath));
+                HasUrlParameterFormat.getTemplate(path, navigationTarget)));
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
@@ -349,8 +349,13 @@ public abstract class AbstractRouteRegistry implements RouteRegistry {
     private void configureWithFullTemplate(String path,
             Class<? extends Component> navigationTarget,
             SerializableBiConsumer<ConfigureRoutes, String> templateConfiguration) {
+        if (HasUrlParameterFormat.hasUrlParameter(navigationTarget)
+                && !HasUrlParameterFormat.hasUrlParameterTemplate(path)) {
+            path = HasUrlParameterFormat.getTemplate(path, navigationTarget);
+        }
+        String finalPath = path;
         configure(configuration -> templateConfiguration.accept(configuration,
-                HasUrlParameterFormat.getTemplate(path, navigationTarget)));
+                finalPath));
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/HasUrlParameterFormat.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/HasUrlParameterFormat.java
@@ -99,6 +99,20 @@ public class HasUrlParameterFormat implements Serializable {
         return urlBase;
     }
 
+    /**
+     * Gets the url base from a given url containing the url parameter template
+     * placeholder {@link HasUrlParameterFormat#PARAMETER_NAME} if it's
+     * implementing {@link HasUrlParameter}.
+     * 
+     * @param urlTemplate
+     *            url with a parameter template
+     * @param navigationTarget
+     *            {@link HasUrlParameter} navigation target.
+     * @throws IllegalArgumentException
+     *             if the given url template doesn't contain url parameter
+     *             template.
+     * @return url excluding parameter template.
+     */
     public static String excludeTemplate(String urlTemplate,
             Class<? extends Component> navigationTarget) {
         if (hasUrlParameter(navigationTarget)) {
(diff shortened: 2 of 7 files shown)


flow-server/src/main/java/com/vaadin/flow/router/internal/HasUrlParameterFormat.java
```diff
@@ -69,7 +69,7 @@ private HasUrlParameterFormat() {
      */
     public static String getTemplate(String urlBase,
             Class<? extends Component> navigationTarget) {
-        if (hasUrlParameter(navigationTarget)) {
+        if (hasUrlParameter(navigationTarget) && !hasUrlTemplate(urlBase)) {
```
denis-anisimov, line 72: Why would I call `getTemplate`  method at all for `urlBase` with PARAMETER ?

It looks like the fix should be done in a different place: `urlBase` should not contain unexpected strings at all and there should be a check for that (or assertion).

And the caller should just not call the `getTemplate` method at all (under some conditions).
So the fix should be done in the caller (not sure which level) and not here.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/internal/HasUrlParameterFormat.java
mshabarov: Now the client (caller) code verifies, that the given navigation target is a `HasUrlParameter` instance and the path does not contain the template, and only then call `getTemplate`.

`getTemplate` now throws if the unexpected path comes, or the passed navigation target is not a `HasUrlParameter`.

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
```diff
@@ -349,8 +349,13 @@ public void clean() {
     private void configureWithFullTemplate(String path,
             Class<? extends Component> navigationTarget,
             SerializableBiConsumer<ConfigureRoutes, String> templateConfiguration) {
+        if (HasUrlParameterFormat.hasUrlParameter(navigationTarget)
+                && !HasUrlParameterFormat.hasUrlParameterTemplate(path)) {
+            path = HasUrlParameterFormat.getTemplate(path, navigationTarget);
```
denis-anisimov, line 354: This just transfers the check from the final place to the intermediate place.

The main question which I would like to understand and get fixed: _why_ the path which is passed as a parameter here may contain PARAMETER at all?
The _here_ means the first call in the stack trace of this method call: I will transfer this question along with traveling this  check to the caller.

I think the bug is: `path`  which is passed at the beginning of the call (may be somewhere in the route registry code) is wrong. It should never contains already PARAMETER.
Most likely the route is registered being registered once (may be being deregistered) and the second registration causes incorrect behavior.

This PR in its initial state and now just hides the symptoms but doesn't fix the original issue.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
mshabarov: > Most likely the route is registered being registered once (may be being deregistered) and the second registration causes incorrect behavior.

This is very close to true I believe. The initial route registration causes no errors. Every path, which has a url parameter, is being extended with a pattern, as expected.

The issue occurs when the `RoutesChangedEvent` is fired as a result of routes update (next step after setting a new routes), and then the existing routes are copied to a OSGi route configuration.

That means the path is being copied too, and those paths with url parameters are passing through the `setRoute` method, which tries to add (again!) a template to them.

Well, I can explicitly add the `getTemplate` method call to the points, where I know in advance that the route (path) is created from navigation target class (i.e. a brand new route), and skip templates when the routes are being copied or removed.
This has a drawback though - this mutation points will be distributed along the code in several point and this is error-prone solution IMO, but technically more correct.
Will do it this way.

denis-anisimov: >That means the path is being copied too, and those paths with url parameters are passing through the setRoute method, which tries to add (again!) a template to them.

Looks like this is the place which should be fixed.
There should be either API which returns the path without template or the path should be converted back (somehow) so that it becomes the same as it was in the very first registration.
As I understand the API is weak at the moment: it always return the modified path and doesn't return the original path.
That should be just fixed.

mshabarov: I've reverted the previous source code changes and changed only the part where the routes updates are being fired to another route registry.
So, basically, from a high level point of view, those changes impacts only the part which is in `2.4`, but not in `6.0+`.

I found nothing better than just to trim the urls from the original registry, to having no param template there -> and then this template is added via normal `setRoute` functionality.

This has a drawback that if someone will want to get the route from registry and add it again in another registry, then he will have to do the same hack. But this scenario is not that possible IMO.

Please give your opinion about the approach I've pushed. If it's fine, I would add the rest of the tests and javadoc.

flow-server/src/main/java/com/vaadin/flow/router/internal/HasUrlParameterFormat.java
```diff
@@ -66,45 +66,54 @@ private HasUrlParameterFormat() {
...
-            urlBase += "(" + RouteParameterRegex.getRegex(parameterType) + ")";
+            if (!String.class.equals(parameterType)) {
+                urlBase += "(" + RouteParameterRegex.getRegex(parameterType)
+                        + ")";
+            }
         }
         return urlBase;
     }
 
+    public static String excludeTemplate(String urlTemplate,
```
denis-anisimov, line 102: The method pretends to be an API but has no javadocs.

denis-anisimov (changes requested):
This can go with that solution.

I still don't like the solution since it makes assumptions: it assumes that the navigation route being added into registry contains a PARAMETER which may be safely removed.
These are quite fragile assumptions.

At the moment this is localized to OSGi which has quite hacky implementation from the very beginning so we may keep this working this way.

BUT OSGi case reveals this as a common problem: one may not use the registry to remove a route from there and add the same route either in the same registry or another.
I may imaging that I want to register Routes via `@Route`  annotation, then modify routes at runtime so that some routes are removed and then at some point I want to restore the routes back. And this is no such way to do it properly.
I have to know about some "template" (which is very internal concept) and I have to know that I have to modify it to restore the correct route and I have to know the method which should be used. All of these things are magic.

The resume is : the API is weak.
It should be amended  with the method which returns the original navigation target path without any PARAMETER.
Please make a ticket about this.

flow-server/src/main/java/com/vaadin/flow/router/internal/HasUrlParameterFormat.java
mshabarov: Added

mshabarov:
I've added a separate ticket https://github.com/vaadin/flow/issues/10033.

denis-anisimov (approved):

Issue #9905, opened 2021-01-28
Each TestBench command for a Fusion app takes 20s with Flow 6 alpha 5


Artur-:
### Description of the bug / feature
The waiting logic in TestBench goes partly like
```
if (window.Vaadin.Flow && !window.Vaadin.Flow.clients) {
  // Wait for the ApplicationConnection to show up
}
```

This no longer works at all in Fusion apps with Flow 6 alpha 5 because the application theme is set as
```
window.Vaadin.Flow._vaadintheme_something
```

apparently also the DevModeGizmo is set in dev mode as
```
window.Vaadin.Flow.devModeGizmo
```

The result is that each and every command in a test takes 20s, which is the timeout for waiting

Not sure if this should be fixed in Flow or TestBench

Legioth:
Both those features are active even if not using any Flow views, so it would be logical that they would be in the `window.Vaadin` namespace rather than `window.Vaadin.Flow`.

pleku:
Would be nice to have a smoketest in Flow that checks that _the wait_ the TB test execution start doesn't take too long for _any_ Vaadin app.

taefi:
The wait for Vaadin script in [TestBench](https://github.com/vaadin/testbench/blob/master/vaadin-testbench-core/src/main/java/com/vaadin/testbench/commands/TestBenchCommandExecutor.java#L53) is like this: 

`if (!window.Vaadin || !window.Vaadin.Flow) {` 

And after moving `_vaadintheme_something` and `devModeGizmo` to Vaadin namespace, Flow object is only initialized for `clients` object.

taefi:
The problem was in the TS only version, the rendered elements are being created in the shadow root and then statements like `$(TextFieldElement.class)` or `$(ButtonElement.class)` will fail since TB fails to find the associated elements. If these statements are used within a condition passed as a lambda to `waitUntil(...)`, they will not fail with `NoSuchElementException`, but will wait until a timeout happens. 

By downloading a simple application from `start.vaadin.com` containing the default **Hello World** view, it can be validated that the following test would pass without any timeout:

```java
@Test
public void testHelloWorldView_enterNameAndClickSayHello_notificationShouldBePresentedWithProperValue() {

    getDriver().get("http://localhost:8080/hello");

    waitUntil(driver -> $("hello-world-view").first() != null);

    TestBenchElement helloWorldView = $("hello-world-view").first();

[...]

PR #10014, opened 2021-02-10
fix: level up namespace of devModeGizmo and theming objects to Vaadin

taefi:
TestBench wait logic is based the existence of Flow object
and in Fusion Only apps Flow object should not be initialized.
moving _vaadintheme_something and devModeGizmo to Vaadin 
namespace fixes the TB wait logic in Fusion Only apps.

Fixes: #9905


pleku (approved):
Some type of regression test would be great or this might get broken again, but not blocking this fix.

Issue #9988, opened 2021-02-08
IllegalArgumentException: length must not be negative in DataCommunicator with undefined size


Springrbua:
### Description of the bug / feature
I have a `Grid` with a `DataCommunicator` with unknown size (`setDefinedSize(false)`).
I am using a `TextField` to search for (i.e. filter) the items.
When I scroll down until a new "block" of data is fetched, and I insert a filter string, which limits the result to a few items only, I am getting a `IllegalArgumentException: length must not be negative`

### Minimal reproducible example

1. Create a `TextField` for the searchTerm:
```
TextField tfSearchTerm = new TextField("Search");
tfSearchTerm.addValueChangeListener(e -> grid.getDataProvider().refreshAll());
```
2. Create a List of at least 200 items
3. Create an AbstractBackEndDataProvider for these items, use the `TextField`s value as a filter. Example:
```
new AbstractBackEndDataProvider<Person, Void>() {
  protected Stream<Person> fetchFromBackEnd(Query<Person, Void> query) {
    List<Person> filteredPersons = getFilteredItems();
      return filteredPersons.stream()
        .skip(Math.min(filteredPersons.size(), query.getOffset()))
        .limit(query.getLimit());
  }
  protected int sizeInBackEnd(Query<Person, Void> query) {
    return getFilteredItems().size();
  }
  private List<Person> getFilteredItems() {
    String searchTerm = tfSearchTerm.getValue();
    return items.stream()
        .filter(p -> p.getName().contains(searchTerm))
        .collect(Collectors.toList());
    }
}
```
4. Create a `Grid` with this `DataProvider` and call `grid.getDataCommunicator().setDefinedSize(false)`
5. Open the web page, scroll down so that a new block of data is fetched and then insert a searchTerm which only matches a few items.

### Expected behavior
The `Grid` should show the filtered items

### Actual behavior
A `java.lang.IllegalArgumentException: length must not be negative` is thrown instead

### Additional information:
The problem seems to be, that the previous fetch loaded items with an offset, which is larger then the size of the filtered list.
For example:
items 50 to 200 are loaded. After applying the filter there are only 2 items left.
[...]

PR #10020, opened 2021-02-12
fix: Resolve requested range properly after filtering the lazy backend

mshabarov:
Fixes https://github.com/vaadin/flow/issues/9988


mshabarov:
The regression test on Grid side comes soon.

pleku (approved):

pleku:
> The regression test on Grid side comes soon.

You've read my mind so I would not need to ask the Q  😄

mshabarov:
Regression test is here https://github.com/vaadin/vaadin-flow-components/pull/645

PR #10022, opened 2021-02-12
fix: remove obsoleted client engine pre-compression

mstahv:
Since 15 all goes through webpack tooling and gets compressed there.

Also fixes build on arm based macs, due the issues with abandoned jbrotly project

### DIFF
--- a/flow-client/pom.xml
+++ b/flow-client/pom.xml
@@ -199,23 +199,6 @@
                 </configuration>
             </plugin>
 
-            <plugin>
-                <groupId>com.github.ryanholdren</groupId>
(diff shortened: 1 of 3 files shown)


Legioth (commented):
LGTM based on my understanding, but I'm leaving ownership for the final review to the implementation team.

mstahv:
The ClientEngineSizeIT.testClientEngineSize in this repository now fails. I'll remove that, but do we have similar tests for the bundle size created by webpack? We probably should.

mstahv:
Removed the failing gz size tests, don't know that what kind of reporting it affects.

pleku (approved):

PR #10023, opened 2021-02-12
Use Lit 2.0.0-rc.1

### LINKED ISSUE #10040: Upgrade to Lit 2

To be able to implement rendering directives for Grid and other components in a good way, we need the features from Lit 2. 

Lit 2 allows us to enable inline renderers, like

https://github.com/web-padawan/lit-vaadin-helpers/blob/280c21e37c4f2f1dfe0e1565b379ba29c28f808b/test/grid.test.ts#L73-L80

or

https://github.com/web-padawan/lit-vaadin-helpers/blob/280c21e37c4f2f1dfe0e1565b379ba29c28f808b/test/grid.test.ts#L49

instead of having to manually bind a function to the element instance and then refer to the function in the template.

To be able to include the renderers in the platform, we need to move everything from Lit 1 to Lit 2. To avoid doing this as a last minute, rushed change before the next LTS, we should aim to get it done for Vaadin 21.

### Backwards incompatibilities in Lit 2

As Lit 2 is a new major version, it is not 100% compatible. The biggest breaking changes are for people who create their own directives but there is one noteworthy change for template users:
-`@internalProperty` is now named `@state`

Artur-:
Fixes #10040


web-padawan (commented):
Please note, Lit 2.0 contains breaking changes that require rewriting `field` directive.

knoobie:
Tbh. it is kinda scary that this change did not break any existing test and could be merged if accepted.

web-padawan:
The reason why this change didn't break is that `lit-element` 2.3.1 is still installed in `"dependencies"` block.
And the current PR only updates `"devDependencies"` so merging it as is would result in duplicate versions.

Artur-:
Right, I just updated the version without checking. Fixed so that it is not a devDep at all

Artur-:
Now there is a hacky port of the field directive and some test fixes and it actually makes sense that the tests pass. Field tests are still commented out as I did not spend time on figuring out how to mock Lit 2 parts.

The field directive works as shown by https://github.com/Artur-/form-lit2/commits/master but probably should be tweaked a bit to make more sense

pleku:
So why should this PR be accepted ? What value does it bring ? How does it affect existing apps using lit in fusion views / flow template views ? Are there breaking changes ? Is there documentation somewhere for users ? There is no issue for this so not clear if this is even a fix or an enhancement ...

> Tbh. it is kinda scary that this change did not break any existing test and could be merged if accepted.

Tbh. in the future we might just automatically close all PRs that fail to address an open issue unless the commit message starts with the `chore` or `test` keyword, but we need to work on the contributing guidelines and pull request template first.

Artur-:
Created #10040 to describe things a bit.

This PR should not be merged right now at least because the Field tests need to be enabled and Field probably refactored a bit.

pleku:
I think this PR should be converted to draft for now (or closed and left in a branch).

mstahv:
Would a feature branch (and automated feature snapshot) of this PR help components team to work on the field directive issue?

Artur-:
The field directive can be fixed as part of this PR directly

Haprog:
@Artur- I just noticed the PR now. I marked it as a draft since it seemed to be not ready yet. Are you still working on it to fix the tests and maybe something else? Or do you expect somebody else to work on it (in addition to just review)?

Artur-:
I expect somebody who knows more about the Field directive to fix it and re-enable the tests

pleku:
> I expect somebody who knows more about the Field directive to fix it and re-enable the tests

@haijian-vaadin so I take it this is on fusion team's table at some point but rather sooner than later. We're looking into any remaining flow related changes in #10818

haijian-vaadin:
> > I expect somebody who knows more about the Field directive to fix it and re-enable the tests
> 
> @haijian-vaadin so I take it this is on fusion team's table at some point but rather sooner than later. We're looking into any remaining flow related changes in #10818

Yes, @Haprog already started working on it.

flow-client/src/test/frontend/form/FieldTests.ts
```diff
@@ -1,632 +1,631 @@
-/* tslint:disable:max-classes-per-file */
```
denis-anisimov, line 1: All code here is commented out.
I'm not familiar with this functionality but I guess the test should be restored back.
As I understand this is fusion functionality and should be checked by the fusion team.

flow-client/src/test/frontend/form/ValidationTests.ts
```diff
@@ -1,586 +1,586 @@
-/* tslint:disable:max-classes-per-file */
```
denis-anisimov, line 1: All code here is also commented out.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -344,8 +344,6 @@ static void addVaadinDefaultsToJson(JsonObject json) {
         defaults.put("lit-css-loader", "0.0.4");
         defaults.put("file-loader", "6.2.0");
         defaults.put("loader-utils", "2.0.0");
-        defaults.put("lit-element", "2.4.0");
```
denis-anisimov, line 347: lit is excluded from the dev dependencies completely ?

flow-client/src/test/frontend/form/FieldTests.ts
Artur-: This is what @Haprog  is working on

flow-client/src/test/frontend/form/ValidationTests.ts
Artur-: This is what @Haprog  is working on

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
Artur-: Yes, it is not a dev dependency, it is a runtime dependency. It cannot be both

denis-anisimov: `lit-element` (with different versions) was in runtime dep and in dev dep.
Was it a mistake ?

denis-anisimov:
Question:
do we need to care about `it/polyfill-support.js`?

It's mentioned here: https://lit.dev/docs/releases/upgrade/#load-polyfill-support-when-using-web-components-polyfills

Artur-:
We do not need `webcomponentsjs` aka web components polyfill as all supported browsers have native support for all parts

Haprog:
> Question:
> do we need to care about `it/polyfill-support.js`?

I don't think so. We don't support such legacy browsers in any version of Vaadin that this is targeting. Also afaik any user who would need the polyfill support, can just load those scripts themselves and they can even be conditionally loaded only when needed.

denis-anisimov:
OK, thanks for the info.

denis-anisimov:
This change also should update all Lit views in ITs.
I think we don't have many of them but still there are some: `import` statements should be updated.

Haprog:
> This change also should update all Lit views in ITs.
> I think we don't have many of them but still there are some: `import` statements should be updated.

I already did that before you commented about it. I updated all occurrences of `from 'lit-` (covers importing from `lit-element` and `lit-html`) etc. Or did I miss something?

denis-anisimov:
> > This change also should update all Lit views in ITs.
> > I think we don't have many of them but still there are some: `import` statements should be updated.
> 
> I already did that before you commented about it (see the commit [7c8c768](https://github.com/vaadin/flow/commit/7c8c76850b2fb29aea502010bec3da144ec53141)). I updated all occurrences of `from 'lit-` (covers importing from `lit-element` and `lit-html`) etc. Or did I miss something?

No, I think it's fine. 
I've not checked a new version of the PR when I wrote the comment: just recalled that we have ITs with lit template and added the comment.

Haprog:
Rebased onto master.

flow-client/.eslintrc.js
```diff
@@ -43,7 +43,8 @@ module.exports = {
         "no-ex-assign": 1,
         "no-return-assign": 1,
         "no-use-before-define": 1,
-        "no-useless-constructor": 1,
+        "no-useless-constructor": 0,
+        "@typescript-eslint/no-useless-constructor": 1,
```
Haprog, line 47: Use the TS variant of this rule so it doesn't complain about a line like:
```ts
constructor(public element: Element & Field) {}
```
in an abstract class (this is used in `Field.ts`)

flow-client/src/main/frontend/form/Field.ts
```diff
@@ -98,98 +97,121 @@ export function getDefaultFieldStrategy(elm: any): FieldStrategy {
...
 
-    element.oninput = () => {
-      updateValueFromElement();
-    };
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    render(model: AbstractModel<any>, effect?: (element: Element) => void) {
+      return nothing;
+    }
```
Haprog, line 134: These ignore are a bit ugly but I think it's nice to have the param types here so we see what they are and below we can get these types automatically to params of `update` method via `DirectiveParameters<this>`.

I'm fine with changing this though since we don't really do anything in render. So that render would have no params and update would have the model and effect typed explicitly.

flow-client/src/test/frontend/form/FieldTests.ts
Haprog: Tests are re-enabled and fixed now.

flow-client/src/test/frontend/form/FieldTests.ts
```diff
@@ -466,113 +463,153 @@ suite("form/Field", () => {
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    class AnyVaadinElement extends LitElement {
+      static get version() {
+        return '1.0';
+      }
+
+      render() {
+        return html``;
+      }
+    }
```
Haprog, line 490: This replaces the weird mocking of constructor version for unknown element `any-vaadin-element-tag` below (`(element.constructor as any).version = '1.0';`). Apparently the existence of version property is used for deciding to use `VaadinFieldStrategy` so this is basically used to simulate that we would get this strategy with any Vaadin component.

flow-client/src/test/frontend/form/FieldTests.ts
```diff
@@ -466,113 +463,153 @@ suite("form/Field", () => {
...
+
+    beforeEach(() => {
+      render(nothing, div);
+    });
+
+    ['div', 'input', 'vaadin-rich-text-editor'].forEach((tag) => {
+      test(`GenericFieldStrategy ${tag}`, async () => {
+        /* eslint-disable lit/binding-positions, lit/no-invalid-html */
```
Haprog, line 498: The linter doesn't seem to like the new allowed syntax in Lit (static expressions used in tag name and element expressions for using directive directly on element without binding it to a property).
So this is needed so it doesn't complain about:
```ts
render(html`<${tagName} ${field(model)}></${tagName}>`, div);
```

flow-server/src/main/resources/com/vaadin/flow/server/frontend/types.d.ts
```diff
@@ -4,7 +4,7 @@
 // It is recommended to commit this file to the VCS.
 // You might want to change the configurations to fit your preferences
 declare module '*.css' {
-    import {CSSResult} from "lit-element";
-    const content: CSSResult;
+    import { CSSResultGroup } from 'lit';
+    const content: CSSResultGroup;
```
Haprog, line 8: This change is needed because in Lit 2 the return type of the `css` tagged template literal changed from `CSSResult` to `CSSResultGroup`.

flow-tests/test-root-context/frontend/lit/simple-lit-template-no-shadow-root.js
```diff
@@ -1,5 +1,4 @@
 
-import { LitElement, html } from "lit-element";
```
Haprog, line 2: Removed unused imports

Haprog:
The CI validation build failed on step "Flow root context tests in NPM dev mode" (flow-tests/test-root-context) during build-frontend with an error "java.io.UncheckedIOException: java.io.IOException: Is a directory" (when running NodeUpdater). 🤔

denis-anisimov (dismissed):

Haprog:
There seemed to be one test failure (`ThemeSwitchLiveReloadIT.switchThemeName_changeThemeNameAndRecompile_themeIsChangedOnFly[ANY_Chrome_]
`) with an error screenshot like this:
image

I tried locally running that test several times and it passed consistently. I restarted the validation build on CI.

Haprog:
Ok. Now the only issue seems to be a single ("critical") sonar complaint that isn't even introduced in this PR. I think we should add a `NOSONAR` comment on that line to ignore it in this case. Or maybe just temporarily disable sonar requirement to merge this PR.

Before merging, I'd still like to get a quick look by @platosha to check mainly `Field.ts` and `FieldTests.ts` if he has any comments about those.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -281,7 +281,7 @@ static JsonObject getJsonFileContent(File packageFile) throws IOException {
                     UTF_8.name());
             try {
                 jsonContent = Json.parse(fileContent);
-            } catch (JsonException e) {
+            } catch (JsonException e) { // NOSONAR
```
denis-anisimov, line 284: There is no need to care about sonar especially if you haven't changed anything.
Sonar reports are not mandatory. Sometimes they may be ignored. Sometimes it complains about things that have not been changed.

Sonar complains don't block the PR.

denis-anisimov (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
Haprog: I know but this is not the first time this line has blocked a PR because sonar thinks it's "critical". We talked about this in our daily today and seems like it makes sense to add this ignore in this specific case.

Haprog: At least I think it was blocking the merge, but actually I'm not sure now since there was also approval missing for a while after previous stale review got dismissed.

platosha (approved):

Issue #1897, opened 2017-06-26
Add API for accessing the current browser location


Legioth:
Add new API to `Page` for accessing the location as shown in the browser's address bar. Some things to take into account:
* This should reflect the actual location shown in the browser's address field, which might not be exactly the same as what's available through the servlet API in cases when e.g. a frontend proxy is used. One additional challenge is that this information isn't available during the initial request, so the API would either have to return e.g. `null` or only an approximation before the value has been verified from the browser.
* The corresponding API in FW8 uses `java.net.URI` for tracking the location, but this is not optimal since there are some weird cases that browsers accept but the `URI` class rejects.

Artur-:
`Page.reload()` is in #3299

gsedlacz:
To implement this feature Page.getLocation() could just use this:

```
public static URI getLocation(){
    VaadinServletRequest req = (VaadinServletRequest) VaadinService.getCurrentRequest();
    StringBuffer uriString = req.getRequestURL();
    return new URI(uriString.toString());
}
```

Legioth:
@gsedlacz. That approach works during initial rendering, but not from a background thread (where there is no current request) nor for incremental update requests since the URL of those are always the common XHR handler endpoint location and not the location that the browser is navigated to.

pleku:
In 2.x versions, there is still the internal API https://vaadin.com/api/platform/14.1.3/com/vaadin/flow/server/VaadinServlet.html#getApplicationUrl-HttpServletRequest- which constructs the URL based on what was in the given request.

pleku:
..but as described in the issue description, the URL might be different due to a frontend proxy, thus the JS `window.location.href` would be most reliable, and that could be now fetched using a `Page.executeJs` that provides a `PendingJavaScriptResult` since 2.0.

As for the use case described in #1894, there would be a need for keeping the "updated" value for saving it when a logout happens, and for that it could be sufficient to have a `getInitialPageLocation()` in `ExtendedClientResponse` and then another method in `Page` which would combine that to the latest routing changes, to be available immediately without a roundtrip for the logout usage.

samie:
Still relevant. Do we have a clear use case to base the implementation? Is this something Router solves?

Logically, getting the page title is similar case:  
https://vaadin.com/forum/thread/17979510/tip-how-to-get-page-title-and-view-of-current-page

knoobie:
I could be wrong, but this kinda looks like (just another way of approaching it / could be a part of)

[...]

PR #10024, opened 2021-02-12
feat: enable fetching browser URL

mshabarov:
Page::fetchCurrentUrl can be used to asynchronously get the current url from the browser.

Closes https://github.com/vaadin/flow/issues/1897

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
@@ -618,11 +618,18 @@ public class Page implements Serializable {
     }
 
     /**
-     * This method can be used to retrieve the current url from the browser.
+     * Retrieves the current url from the browser. The URL is fetched from the
+     * browser in another request asynchronously and passed to the callback. The
+     * URL is the full URL that the user sees in the browser (including hash #)
+     * and works even when client side routing is used or there is a reverse
+     * proxy between the client and the server.
      * <p>
      * In case you need more control over the execution you can use
      * {@link #executeJs(String, Serializable...)} by passing
      * {@code return window.location.href}.
+     * <p>
+     * <em>NOTE: </em> the URL is not escaped, use {@link URL#toURI()} to escape
+     * it.
      * 
      * @param callback
      *            to be notified when the url is resolved.
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
@@ -20,7 +20,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Serializable;
+import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.Objects;
 import java.util.function.Function;
@@ -615,4 +617,25 @@ public class Page implements Serializable {
                         getStringElseNull.apply("v-np")));
     }
 
+    /**
+     * This method can be used to retrieve the current url from the browser.
+     * <p>
+     * In case you need more control over the execution you can use
+     * {@link #executeJs(String, Serializable...)} by passing
+     * {@code return window.location.href}.
+     * 
+     * @param callback
+     *            to be notified when the url is resolved.
+     */
+    public void fetchCurrentURL(SerializableConsumer<URL> callback) {
+        Objects.requireNonNull(callback, "Url consumer callback should not be null.");
+        final String js = "return window.location.href";
+        executeJs(js).then(String.class, urlString -> {
+            try {
+                callback.accept(new URL(urlString));
+            } catch (MalformedURLException e) {
+                throw new IllegalStateException("Error while encoding the URL from client", e);
+            }
+        });
+    }
 }
--- a/flow-server/src/test/java/com/vaadin/flow/component/page/PageTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/page/PageTest.java
@@ -16,6 +16,7 @@
 package com.vaadin.flow.component.page;
 
 import java.io.Serializable;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -223,6 +224,64 @@ public class PageTest {
                 .dumpPendingJavaScriptInvocations().size();
         Assert.assertEquals(1, jsInvocations);
         Assert.assertEquals(2, callbackInvocations.get());
+    }
+
+    @Test
+    public void fetchCurrentUrl_consumerReceivesCorrectURL() {
+        // given
+        final UI mockUI = new MockUI();
+        final Page page = new Page(mockUI) {
+            @Override
+            public PendingJavaScriptResult executeJs(String expression,
+                    Serializable... params) {
+                super.executeJs(expression, params);
+                Assert.assertEquals(
+                        "Expected javascript for fetching location is wrong.",
+                        "return window.location.href", expression);
+
+                return new PendingJavaScriptResult() {
+
+                    @Override
+                    public boolean cancelExecution() {
+                        return false;
+                    }
+
+                    @Override
+                    public boolean isSentToBrowser() {
+                        return false;
+                    }
+
+                    @Override
+                    public void then(
+                            SerializableConsumer<JsonValue> resultHandler,
+                            SerializableConsumer<String> errorHandler) {
+                        resultHandler.accept(Json
+                                .create("http://localhost:8080/home"));
+                    }
+                };
+            }
+        };
+        final AtomicReference<URL> callbackInvocations = new AtomicReference<>();
+        final SerializableConsumer<URL> receiver = details -> {
+            callbackInvocations.compareAndSet(null, details);
+        };
+
+        // when
+        page.fetchCurrentURL(receiver);
+
+        // then
+        Assert.assertEquals("Returned URL was wrong",
+                "http://localhost:8080/home",
+                callbackInvocations.get().toString());
+    }
+
+    @Test
+    public void fetchCurrentUrl_passNullCallback_throwsNullPointerException() {
+        Assert.assertThrows(NullPointerException.class, () -> {
+            final UI mockUI = new MockUI();
+            Page page = new Page(mockUI);
+            page.fetchCurrentURL(null);
+        });
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
```diff
@@ -615,4 +617,25 @@ private void handleExtendedClientDetailsResponse(JsonValue json) {
...
+     * This method can be used to retrieve the current url from the browser.
+     * <p>
+     * In case you need more control over the execution you can use
+     * {@link #executeJs(String, Serializable...)} by passing
+     * {@code return window.location.href}.
+     * 
+     * @param callback
+     *            to be notified when the url is resolved.
+     */
+    public void fetchCurrentURL(SerializableConsumer<URL> callback) {
```
fluorumlabs, line 630: I know this is probably wrong place to ask that kind of questions... But why callbacks and not `CompletableFuture<>`? What is the expected usage pattern?

Another question: should it return `URL` or `URI`?

pleku: > Another question: should it return URL or URI?

Now with fresh eyes I would still go with URL as it is closer to `window.location.href` and to my understanding this will e.g. never be _opaque_ so using URI would not be correct that way. Choosing URI would be preferred for the fact that this is coming from the browser and would be then encoded, but we should IMO just add a note in the javadocs stating that _this is not encoded and for encoding, you should use `URL::toURI`_ should be enough. (I noticed that `Page::setLocation` has a variant with URI, but that is slightly different)

> But why callbacks and not Future<>? What is the expected usage pattern?

Yes as this is asynchronous it was considered, but it was chosen to be kept simple instead. As the javadocs mention, in case you need more control, you can just use the `executeJs` method to get the `PendingJavaScriptResult` which also provides a `CompletableFuture`. 

As the nature of this operation is very simple, it didn't seem like a good idea to make it always more complex to use, or is it expected that methods like cancel/isCancelled/isDone would be required all the time for this case ?

fluorumlabs: > As the nature of this operation is very simple, it didn't seem like a good idea to make it always more complex to use

Would be interesting to see expected usage patterns or use cases. 
- Is current URL something user might want to subscribe to? Then IMO it should be something that returns `Registration` and can be unregistered at any point.
- Is it one-time operation? Then perhaps using `Future` will make a code a bit cleaner (open question). Also, using callbacks may lead to situation where view class reference (I suppose that's the place where `fetchCurrentURL` is supposed to be used) will be captured as a part of the callback lambda. Proposed API doesn't provide means to cancel callback or otherwise clean up when the session is destroyed, so, theoretically, it may extend lifespan of `UI` to the lifespan of `VaadinSession`.

pleku: Only described use case in relevant issues is about redirecting back due to logout/login and thus storing it on the server side. It seems unlikely that it needs to be written it back to the UI as the browser shows the location to the user. As @knoobie and @jhult have commented on the issues, maybe you can elaborate on the usage you've had please ?

> Also, using callbacks may lead to situation where view class reference will be captured as a part of the callback lambda. Proposed API doesn't provide means to cancel callback or otherwise clean up when the session is destroyed, so, theoretically, it may extend lifespan of UI to the lifespan of VaadinSession.

No it doesn't provide canceling, but like said, there is an existing way how to handle that like mentioned in the javadocs in case necessary. Any dangling execution would always be bound to the UI anyway (regardless of callback or using `then(...)`, and if the user closes the window/tab at a suitable time when the execution is pending, I don't see how it would change anything as the UI and the pending execution would be have the same lifecycle. The UI lifecycle is not extended due to the pending invocation / callback. If someone will try to reference or store some UI components in the callback, I don't think we can prevent that and yet still it is only for the duration of the UI lifecycle. I don't know how they could increase the lifespan further to tie things into session.

I'm not sure how the API with `Future` would look (in order to be better), and since it is not used for `executeJS` either due to specific semantics in the operation, I would not use it. So this is how it is described in the docs PR:
(I don't know if it is necessary to even do this at logout click. It would be enough to do it once once the UI is initialized and then just use the URL combined to he currently opened view path, the following is just an example)
```java
// inside any component type class
logoutButton.addClickListener(event -> {
    UI.getCurrent().getPage().fetchCurrentURL(this::storeUrlAndLogout));
});
```
it would be instead (with specifically typed return type like, not `Future` as that won't work here as you cannot wait synchronously for the result)
```java
// inside any component type class
logoutButton.addClickListener(event -> {
    UI.getCurrent().getPage().fetchCurrentURL().then(this::storeUrlAndLogout));
    
    // or store and cancel it later, no idea of the naming but based on PendingJavaScriptResult
    PendingURLResult result = UI.getCurrent().getPage().fetchCurrentURL());
    result.then(this::storeUrlAndLogout);
    result.then(this::storeUrlAndLogout, this::onUrlFetchingError);
});
// in case the execution would need to be canceled later on, the API is there
atomicResult.get().cancelExecution();
```
But this seems like a major overkill since the chances of needing any of the above are thought to be very rare and if needed you could just get the same with
```java
PendingJavaScriptResult result = UI.getCurrent().getPage().executeJS("return window.location.href");
result.then(String.class, stringUrl -> this.storeUrl(new URL(stringUrl)));
...
```

fluorumlabs: The only use case that I was thinking of (apart from logout) was support for hash parameters in URL (for legacy reasons, for example)...

However, if we ignore hash parameters, what would be the difference compared to `com.vaadin.flow.component.internal.UIInternals#getActiveViewLocation`? Obviously, `Location` is missing protocol, host and port, but maybe bootstrapping can be adjusted to store that data?

Alternatively, the whole `location.href` can be passed in the `UIDL_NAVIGATION` event, but then, again, it won't support hash parameters (Flow doesn't support them anyway).

Either way it would not require extra roundtrip.

---

Another related question: what would value `fetchCurrentURL` will yield, if it will be followed by, let's say, `UI.navigate` or `Page.getHistory().replaceState()`?

knoobie: I had different use cases in mind once I saw the issue, like e.g. storing the current URL on logout for "go back where you left off" kind of thing or to create analytics about user behavior (time per page and so on) or you store the current URL once a user submits some kind of support message / question to your (?) on the bottom corner of e.g. Vaadin.com to help him with specific questions and you already know what transaction he has open and so on. 

Personally when I had this feature in mind, I never thought about the async way. I was expecting the API to be straight forward and everywhere available without additional round trips or (in my opinion) behind a kinda obscure API. I was expecting the vaadin-router to have all those information and they just have to be send to the server on every URL change and stored inside the Page or UI object to be easily accessible like "Page.getCurrent().getCurrentUrl()" - because the URL change event has to be send anyway to the server (except the # problematic)

jhult: I agree with @knoobie on this:
> I was expecting the API to be straight forward and everywhere available without additional round trips

pleku: > Either way it would not require extra roundtrip.

> Page or UI object to be easily accessible like "Page.getCurrent().getCurrentUrl()" - because the URL change event has to be send anyway to the server (except the # problematic)

Ok, thanks. Like mentioned in the issue: the URL from any request will not give the server side the same URL that the user sees on the browser when there is a proxy being used. Otherwise (if you don't have a proxy in the between) you can just fetch the current request and construct the URL from there (convenience API missing though).
Thus the location that the user sees has to be fetched using JS. So it is basically the same as with `Page::getExtendedClientDetails` as you can never have it already available for the initial request that bootstraps the Vaadin UI. And like with other similar cases, this has been taken into account in the API by forcing you to give a callback instead of returning `null` or `Optional<URL>` pending on whether the URL is available.

Another alternative could have been to expose this in the `ExtendedClientDetails` but for visibility and for the fact that this would not be cached (you can fetch it whenever you want) it was added separately to that.

It would seem to me that at least the javadocs could point out certain details and maybe the documentation PR also could be more detailed. I appreciate the feedback, please remember to read the issue too 😉 , but I'm not still convinced on what would be a better alternative to this due to the technical "limitations" ^.

One thing would be to provide an additional API in `Page` that gives you something like `getUrlBasedOnRequest()` which would work for those apps where no proxy is being used, then pointing out this other method for those who need it.

pleku: And also, the JS execution does bring the hash part visible too for those who need it as it is not given by the http request (like it shouldn't as hash is a browser-only feature, but please lets not go there).

knoobie: >  the URL from any request will not give the server side the same URL that the user sees on the browser when there is a proxy being used.

Understable! That's why I was hoping that the vaadin-router has those information and can send them on routing events. Because it runs on the client side and is involved in any routing anyway. If that's not the case, it's understandable why you wanna go the route. 

> ExtendedClientDetails

Now that you named it, I would probably move this method there and add Page.getLocationFromRequest() here instead, because it would probably work with 80% cases (I'm not really sure when you need the domain, I'm normally happy with the view chain and parameters) and document that "extended client detail" (domain/full URL) is available in there and vice versa.

fluorumlabs: > I'm not really sure when you need the domain

Application might be running behind proxy mapping, for example `app.server.com` (or `server.com/app`) to `int.server.com/appServlet`. Full URL is needed if you have an external authentication server, which accepts redirect URL.

Ideal solution would involve adding `ExternalClientDetails.getBaseUrl()` method that will return base URL based on the client-side path (it is highly unlikely that base will change) and `Page.getCurrentUrl()` or something like that (essentially, derived from `UI.getInternals().getActiveViewLocation()`).

pleku: > Ideal solution would involve adding ExternalClientDetails.getBaseUrl() method that will return base URL based on the client-side path (it is highly unlikely that base will change) and Page.getCurrentUrl() or something like that (essentially, derived from UI.getInternals().getActiveViewLocation()).

IMO it is not really ideal since that doesn't work when you want to 
a) redirect to existing url that had the hash - it should be possible to capture, store and redirect to a view's hash for logout/login (but not use it for server side routing purposes)
b) if there is client side routing used, the `Location` is useless

But discuss the changes with the team and adopt the changes following our principles of "keeping easy things easy, but complex cases possible".

pleku (approved):

caalador (approved):

PR #10026, opened 2021-02-14
fix: NumberValidator accepts undefined for optional fields

### LINKED ISSUE #10008: client-side validation fails for `@Id`-annotated properties
Client-side validation fails (incorrectly) when submitting a form that creates a _new_ entity using the TypeScript form binder API. `@Id`-annotated properties are optional in the generated TypeScript types (see https://github.com/vaadin/flow/issues/9222). The default value for optional properties is `undefined`. However, this default value does not pass the default client-side validation for optional numeric properties:

```
Uncaught (in promise) ValidationError: There are validation errors in the form.
 - id - IsNumber: must be a number
    ValidationError Validation.ts:25
    submitTo Binder.ts:162
    save master-detail-view.ts:137
```

**Steps to reproduce**:
1. create a new Vaadin 19 (Fusion) project with https://start.vaadin.com
2. add a 'master-detail' view
3. start the app and open the 'master-detail' view
4. fill-in the entity form to create a new 'person'
5. click save and see that the entity is not saved and there is an error in the browser console

**Workaround**:
Before calling `binder.submitTo(endpoint)` set `binder.value.id` to `0`:
```ts
this.binder.value.id = 0;
await this.binder.submitTo(PersonEndpoint.update);
```

joheriks:
Fixes #10008

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/form/Models.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/form/Models.ts
@@ -89,7 +89,7 @@ export class NumberModel extends PrimitiveModel<number> implements HasFromString
     ...validators: ReadonlyArray<Validator<number>>
   ) {
     // Prepend a built-in validator to indicate NaN input
-    super(parent, key, optional, new IsNumber(), ...validators);
+    super(parent, key, optional, new IsNumber(optional), ...validators);
   }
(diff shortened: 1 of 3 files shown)


platosha (approved):

haijian-vaadin (approved):
The code looks good, and I tried with the PR snapshot, it works like a charm!

PR #10075, opened 2021-02-18
Clarify warning comment for setEnabled

Legioth:


flow-server/src/main/java/com/vaadin/flow/component/HasEnabled.java
```diff
@@ -74,6 +74,9 @@ default void setEnabled(boolean enabled) {
          * XXX WARNING Do not override this method. Propagating the enabled
          * state to the element in this way is critical to fulfill generic
          * assumptions with regards to application security.
+         *
+         * Override Component::onEnabledStateChanged if you need to adjust how
+         * the disabled state is visually represented in the browser.
```
pleku, line 79: Just commenting that it is slightly weird that the warnings are here instead of in the method level javadocs 🤷

pleku (approved):

flow-server/src/main/java/com/vaadin/flow/component/HasEnabled.java
Legioth: These JavaDocs are directly inherited into each component that uses this interface. The added comments do not make any sense at all for e.g. `Button` or `ComboBo`. It's not optimal to have them only in the code either, but it's the best place I could think of that wouldn't pollute everything.

Issue #10065, opened 2021-02-17
ApplicationThemeComponentIT.componentThemeIsApplied_forPolymerAndLit is flaky


pleku:
Will ignore it for now but it should be enabled back for both master and 6.0
https://bender.vaadin.com/project.html?projectId=Flow&testNameId=-8031487230985450787&tab=testDetails

PR #10078, opened 2021-02-18
chore: Update how input field is gotten

caalador:
Update the way we get the text-field input
field as the original way has problems
when multiple fields exist on the page as the
id will depend on initialization order.
This was only apparent on the embedded test side.

Fxes #10065


pleku (approved):
(I wonder if there would be some TB element API that should have this so that it can be reused)

pleku:
Also this is a chore as it should not go to release notes 😉

caalador:
Yes, it's called TextFieldElement in the TextField component that we perhaps do not want to depend on in Flow.

PR #10081, opened 2021-02-18
chore: write package json for default changes

### LINKED ISSUE #10032: vaadin.devDependencies versions are not updated unless the installed versions match

If you have manually upgraded webpack to the latest version in a project and upgrade to Vaadin 19 beta2, the package.json will contain
```
{
  "devDependencies": {
    "webpack": "4.46.0",
  },
  "vaadin": {
    "devDependencies": {
      "webpack": "4.42.0",
    },
}
```

This makes little sense as Vaadin uses webpack 4.46.0 in this version. If Vaadin later on upgrade to e.g. webpack 5, I probably want that updated in my project also.

### Expected behavior
The `vaadin` section of `package.json` contains what the current Vaadin version would install.

### Actual behavior
The `vaadin` section of `package.json` contains versions which were installed by Vaadin earlier at some point.

### Versions:
    - Vaadin / Flow version: 19 beta2

caalador:
Always write the package json after
updateDefaultDependencies as even if
the dependencies are not updated the
defaults may be updated.

Fixes #10032

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGeneratePackageJson.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGeneratePackageJson.java
@@ -52,9 +52,7 @@ public class TaskGeneratePackageJson extends NodeUpdater {
             modified = false;
             JsonObject mainContent = getPackageJson();
             modified = updateDefaultDependencies(mainContent);
-            if (modified) {
-                writePackageFile(mainContent);
-            }
+            writePackageFile(mainContent);
 
             if(flowResourcesFolder == null) {
                 return;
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGeneratePackageJson.java
```diff
@@ -52,9 +52,7 @@ public void execute() {
             modified = false;
             JsonObject mainContent = getPackageJson();
             modified = updateDefaultDependencies(mainContent);
-            if (modified) {
-                writePackageFile(mainContent);
-            }
+            writePackageFile(mainContent);
```
pleku, line 55: So is the `modified` flag even necessary here ? Should it be removed not to confuse anyone and maybe comment that this will be always written here. As left as it, is just raises questions if the writing should happen or not

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGeneratePackageJson.java
caalador: For the generator the modified flag is for verifying that the updates are done as expected and it's part of the parent NodeUpdater.
UpdatePackages modified flag then again has to do with the npm execution check.

pleku: Ok. I think I get it but there is a slight code smell here

pleku (approved):

Issue #10080, opened 2021-02-18
Cannot use global CSS loader without manually adding 'style-loader' dep


manolo:
Flow updaters should add the dependency automatically as it does with other packages required by the bundling process.

How to reproduce:

- get a fusion project from start.vaadin.com
- add some `file.global.css` file to the app
- remove `package.json`

App refuses to start because the following error

```
Module not found: Error: Can't resolve 'style-loader' in '/Users/manolo/Github/PiT/restaurante/frontend'
```

PR #10085, opened 2021-02-18
fix: add style-loader dependency

haijian-vaadin:
Fixes https://github.com/vaadin/flow/issues/10080

Also removed the unnecessary `idb` dependency.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -325,7 +325,7 @@ public abstract class NodeUpdater implements FallibleCommand {
         defaults.put("chokidar", "^3.5.0");
 
         defaults.put("idb", "5.0.6");
-        defaults.put("style-loader", "^2.0.0");
+        defaults.put("style-loader", "2.0.0");
 
         return defaults;
     }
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -325,6 +325,7 @@ static void addVaadinDefaultsToJson(JsonObject json) {
         defaults.put("chokidar", "^3.5.0");
 
         defaults.put("idb", "5.0.6");
+        defaults.put("style-loader", "^2.0.0");
```
Haprog, line 328: Should probably use explicit version like in all the other versions here (except "chokidar", I'm not sure why that uses a caret range)?
```suggestion
        defaults.put("style-loader", "2.0.0");
```

Haprog (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
haijian-vaadin: done

joheriks: >  I'm not sure why that uses a caret range
This has been added as a special case to fix a broken transitive dependency (chokidar 2). As we do not directly depend on it, the version of the compatible version was left as a range (see #8451).

Haprog (approved):

PR #10090, opened 2021-02-18
Testing "manual rebase" of the plugin changes by @stbischof

mstahv:
Testing why git doesn't detect all file renames in a patch made by @stbischof . Not to be merged at this point, just for testing.


caalador:
How about just having a couple PRs for this. Start with a PR that makes the flow-plugins module and moves flow-maven-plugin there. And then in another PR create the flow-plugin-base. At least moving the whole plugin shows as moved and not deleted+created.

mstahv:
Like half of the lines, and most relevant lines are moved to a different file (not just relocated, but also rename). Then it is just too much for git. Not a big fan of putting broken commits to repository. The thing that @mcollovati suggested `git rebase --strategy-option=rename-threshold=10 develop` might be the only thing to try, but I think we are now wasting out time.

caalador:
It wouldn't be a broken commit as it would just be moving the files to the new package structure,

caalador:
Plugin refatoring was merged. Closing this PR.

PR #10099, opened 2021-02-19
fix: remember original state after LOADING

joheriks:
When transitioning back from `LOADING`, don't assume `CONNECTED`.

### DIFF
--- a/flow-client/src/test/frontend/ConnectTests.ts
+++ b/flow-client/src/test/frontend/ConnectTests.ts
@@ -202,7 +202,6 @@ describe('ConnectClient', () => {
         // expected
       } finally {
         (expect(stateChangeListener).to.be as any).calledWithExactly(ConnectionState.LOADING, ConnectionState.CONNECTION_LOST);
-        // expect((window as any).Vaadin.connectionState.state).to.equal(ConnectionState.CONNECTION_LOST);
       }
     });
 
--- a/flow-client/src/test/frontend/ConnectTests.ts
+++ b/flow-client/src/test/frontend/ConnectTests.ts
@@ -179,40 +179,19 @@ describe('ConnectClient', () => {
       expect(fetchMock.lastOptions()).to.include({method: 'POST'});
     });
 
-    it('should set CONNECTED followed by ConnectionState.loadingFinished when Flow is not loaded', async() => {
+    it('should set connection state to LOADING followed by CONNECTED on successful fetch', async() => {
       let $wnd = (window as any);
-      let states: Array<ConnectionState> = [];
-      $wnd.Vaadin.connectionState.addStateChangeListener(
-        (_, state) => states.push(state));
+      const stateChangeListener = sinon.fake();
+      $wnd.Vaadin.connectionState.addStateChangeListener(stateChangeListener);
+
       await client.call('FooEndpoint', 'fooMethod');
-      expect(states).to.deep.equal([ConnectionState.LOADING, ConnectionState.CONNECTED]);
+      (expect(stateChangeListener).to.be as any).calledWithExactly(ConnectionState.LOADING, ConnectionState.CONNECTED);
     });
 
-    it('should call Flow.loadingStarted followed by Flow.loadingFinished when Flow is loaded', async() => {
-      let calls: Array<boolean> = [];
-      (window as any).Vaadin.Flow = {
-        clients: {
-          TypeScript: {
-            loadingStarted: () => calls.push(true),
-            loadingFinished: () => calls.push(false)
-          }
-        }
-      };
-      await client.call('FooEndpoint', 'fooMethod');
-      expect(calls).to.deep.equal([true, false]);
-    });
-
-    it('should call Flow.loadingFinished and transition to CONNECTION_LOST upon network failure', async() => {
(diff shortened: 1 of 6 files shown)


flow-client/src/test/frontend/ConnectionStateTests.ts
```diff
@@ -64,38 +61,54 @@ describe('ConnectionStateStore', () => {
...
-        [ConnectionState.LOADING, ConnectionState.CONNECTED]]);
+    assert.equal(stateChangeListener.callCount, 4)
+    assert.isTrue(stateChangeListener.getCall(0).calledWithExactly(
+      ConnectionState.CONNECTED, ConnectionState.LOADING))
+    assert.isTrue(stateChangeListener.getCall(1).calledWithExactly(
+      ConnectionState.LOADING, ConnectionState.CONNECTION_LOST))
+    assert.isTrue(stateChangeListener.getCall(2).calledWithExactly(
+      ConnectionState.CONNECTION_LOST, ConnectionState.LOADING))
+    assert.isTrue(stateChangeListener.getCall(3).calledWithExactly(
+      ConnectionState.LOADING, ConnectionState.CONNECTION_LOST))
```
Haprog, line 97: Now here the logic changed so that the last state is `CONNECTION_LOST` instead of `CONNECTED`. This doesn't make sense to me.

Seems like the `store.loadingFinished();` call does not currently know or care if the loading was successful or not which in my opinion would be necessary information for these kind of tests and to be able to get the right behaviour.

See here https://github.com/vaadin/flow/blob/dd372e4d3c85352287151eaa3e05d09bb7448b59/flow-client/src/main/resources/META-INF/resources/frontend/Connect.ts#L374-L387
where `loadingFinished()` is called the same way in both cases if the call succeeds or fails. The only difference is that in the failure case the state is set to `CONNECTION_LOST` after calling `loadingFinished()`. So in this case I think the issue used to be that `loadingFinished()` was previously setting the state to `CONNECTED` and then it was immediately set to `CONNECTION_LOST` (in case the call failed).

So it looks like two different parts of the code are trying to both determine what should happen here but this state change should only be handled in one place imo.

If a call finishes successfully (meaning we got a result from the server without network failure) it should mean that the connection is not lost and if a call fails (with network error) it should mean that the connection is lost.

I think the result of the most recently finished call should determine do we transition to `CONNECTION_LOST` or `CONNECTED`. Not what the state was before starting to load something (that seems irrelevant to me). Think about if you had connection lost, then started to load something, and it finishes successfully. It means you have a working connection so it should not return to connection lost state.

Imo instead of `stateBeforeLoading` we should have something like `latestCallSuccessful` (true or false). This could be set in `Connect.ts` before calling `loadingFinished()` or it could be a parameter to `loadingFinished()` so that it knows if it finished without network errors or not. Or alternatively split it to two methods `loadingSucceeded()` and `loadingFailed()`.

Haprog: Imo we don't need to store `stateBeforeLoading` unless we would somehow support cancelling a call that has not finished yet.

Idea outside the scope of this issue: I think it might be useful to have another connection state that would only have 2 states (connected and connection lost, so without the loading states). This could be useful for users when they only care about is the current state considered to be connected or not without caring about possible ongoing/loading fetch calls.

Haprog (changes requested):
The logic seems slightly flawed to me and seems like it can still cause unexpected state in some cases. See my suggestion below.

flow-client/src/test/frontend/ConnectionStateTests.ts
joheriks: Good points. Added a method `loadingFailed` transitioning to `CONNECTION_LOST` when the counter reaches zero. So the connection state will remain in `LOADING` while there are ongoing backend calls, and the result of the last call (succeeed or failed) will determine the final state (`LOADING`/`CONNECTION_LOST`). So `ConnectionStateStore` will represent the result of the last server interaction.

(Also removed the extra complication of `Connect` calling methods `loadingStarted` and `loadingFinished` in Flow instead of directly accessing `ConnectionStateStore`, as the only thing this accomplished in addition was setting the `isActive` flag for TestBench which was not stacked anyway and hence meaningless).

Haprog (dismissed):
Looks good. Consider removing commented line of code.

flow-client/src/test/frontend/ConnectTests.ts
```diff
@@ -222,12 +201,13 @@ describe('ConnectClient', () => {
       } catch (error) {
         // expected
       } finally {
-        expect(calls).to.deep.equal([true, false]);
-        expect((window as any).Vaadin.connectionState.state).to.equal(ConnectionState.CONNECTION_LOST);
+        (expect(stateChangeListener).to.be as any).calledWithExactly(ConnectionState.LOADING, ConnectionState.CONNECTION_LOST);
+        // expect((window as any).Vaadin.connectionState.state).to.equal(ConnectionState.CONNECTION_LOST);
```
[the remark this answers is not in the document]
joheriks: Done. The assertion would still be valid but not necessary since the connection state is checked via the listener.

Haprog (approved):

PR #10112, opened 2021-02-23
Do not clear all children client side for empty template elements

### LINKED ISSUE #10106: All content of a @Id mapped element are removed from the DOM
If you have a template / component that modifies its child elements dynamically, you cannot use it with a `LitTemplate` as Flow will remove all the created children.

```java
@Tag("empty-view")
public class EmptyView extends LitTemplate {

    @Id
    private Div root;
}
```

```ts
import { customElement, html, LitElement, PropertyValues, query } from "lit-element";

@customElement("empty-view")
export class EmptyView extends LitElement {
  @query("#root")
  private root!: HTMLElement;

  createRenderRoot() {
    // Do not use a shadow root
    return this;
  }
  render() {
    return html`<div id="root"></div>`;
  }

  protected firstUpdated(_changedProperties: PropertyValues) {
    super.firstUpdated(_changedProperties);
    this.root.innerText = "Hello";
  }
}
```

### Expected behavior
You see "Hello" in the view

### Actual behavior
The view is empty. If you remove the `@Id` mapping, then you see "Hello"

### Versions:
    - Vaadin / Flow version: 18

Artur-:
The default value of Element.getText() is "" so setting "" should not be necessary

Fixes #10106

### DIFF
--- a/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/InjectableLitElementInitializerTest.java
+++ b/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/InjectableLitElementInitializerTest.java
@@ -134,14 +134,6 @@ public class InjectableLitElementInitializerTest {
 
         Assert.assertEquals("foo bar", element.getText());
     }
-    @Test
-    public void initializeElement_setText_emptyTextIsNotSet() {
-        element.appendChild(new Element("child"));
-        initializer.accept(Collections.singletonMap(
-                AbstractInjectableElementInitializer.TEXT_DATA, ""));
-
-        Assert.assertEquals("child", element.getChild(0).getTag());
-    }
 
     @Tag(Tag.DIV)
     public static class TestComponent extends Component implements HasStyle {
--- a/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/InjectableLitElementInitializerTest.java
+++ b/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/InjectableLitElementInitializerTest.java
@@ -18,6 +18,7 @@ package com.vaadin.flow.component.littemplate;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.hamcrest.Matchers;
 import org.junit.Assert;
@@ -129,13 +130,33 @@ public class InjectableLitElementInitializerTest {
     }
 
     @Test
-    public void initializeElement_setText_textIsSet() {
+    public void initializeElement_setText_textIsSetOnlyOnServer() {
         initializer.accept(Collections.singletonMap(AbstractInjectableElementInitializer.TEXT_DATA, "foo bar"));
 
         Assert.assertEquals("foo bar", element.getText());
+
         ElementChildrenList children = element.getNode().getFeature(ElementChildrenList.class);
         Assert.assertEquals(1, children.size());
-        Assert.assertEquals(0, children.getChangeTracker().size());
+        AtomicBoolean hasChange = new AtomicBoolean(false);
+        children.collectChanges(change -> {
+            hasChange.set(true);
+        });
+        Assert.assertFalse("There should be no pending changes for element children", hasChange.get());
+    }
+
+    @Test
+    public void initializeElement_setText_emptyTextIsSetOnlyOnServer() {
+        initializer.accept(Collections.singletonMap(AbstractInjectableElementInitializer.TEXT_DATA, ""));
+
+        Assert.assertEquals("", element.getText());
+
+        ElementChildrenList children = element.getNode().getFeature(ElementChildrenList.class);
+        Assert.assertEquals(0, children.size());
+        AtomicBoolean hasChange = new AtomicBoolean(false);
+        children.collectChanges(change -> {
+            hasChange.set(true);
+        });
+        Assert.assertFalse("There should be no pending changes for element children", hasChange.get());
     }
 
     @Tag(Tag.DIV)
--- a/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/InjectableLitElementInitializerTest.java
+++ b/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/InjectableLitElementInitializerTest.java
@@ -134,6 +134,14 @@ public class InjectableLitElementInitializerTest {
 
         Assert.assertEquals("foo bar", element.getText());
     }
+    @Test
+    public void initializeElement_setText_emptyTextIsNotSet() {
+        element.appendChild(new Element("child"));
+        initializer.accept(Collections.singletonMap(
+                AbstractInjectableElementInitializer.TEXT_DATA, ""));
+
+        Assert.assertEquals("child", element.getChild(0).getTag());
+    }
 
     @Tag(Tag.DIV)
     public static class TestComponent extends Component implements HasStyle {
--- a/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/InjectableLitElementInitializerTest.java
+++ b/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/InjectableLitElementInitializerTest.java
@@ -31,6 +31,7 @@ import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.component.template.internal.AbstractInjectableElementInitializer;
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.dom.ElementFactory;
+import com.vaadin.flow.internal.nodefeature.ElementChildrenList;
 
 public class InjectableLitElementInitializerTest {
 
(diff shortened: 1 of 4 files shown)


denis-anisimov (commented):
This PR doesn't fix #10106: #10106 is not a bug.
Changing the example  to
```
render() {
    return html`<div id="root">I'm here to keep the example broken</div>`;
  }
```
makes everything as described in the #10106  ticket.
The expectation in the ticket is not correct.
The behavior is a consequence of the introduced template feature which sets the attributes/properties and text values based on the template values.
The client side function is invoked apparently before initialization from the sever side happens. So the element is initialized with the empty (OR any text which is set in the template)  after the client side function is invoked.

Use either `afterServerUpdate`  function or set the text from the server side to avoid confusing behavior.

This PR is just a tiny improvement which I may be even against of because of inconsistent behavior: it works this way (as "expected" in the ticket) with an empty Text node value and it works in a different way with non-empty Text node value.
So I may imagine the same ticket which says that same scenario works when the text value is empty and doesn't work when it's not empty: even though this is expected behavior based on the implementation it looks inconsistent for others.
So I think this minor improvement just introduces a bug.

Artur-:
Why is it desired that the server side removes the client side DOM? The feature of reading text from the template is so that the server knows the initial value and the user can set it to something else. Not so that it can wipe out perfectly fine DOM created by a component in the browser.

A real use case for this is e.g. to integrate Leaflet as
```ts
  render() {
    return html`<leaflet-map id="theMap"></leaflet-map>`;
  }
```

```java
    @Id("theMap")
    private LeafletMap map;

...
        map.setView(55.0, 10.0, 4);
```

Where `<leaflet-map>` will add children to its DOM. There is no way to "do that from the server" nor is there any desire to.

denis-anisimov:
>Why is it desired that the server side removes the client side DOM?

Does it ?

The text is set _ONLY_ when element has no children.
Well, in any case I don't understand what you are talking about in the context of this PR.

The PR checks whether the text node value is empty or not.
If it's not empty then the text is set.
So the only thing which is changed by this PR: the empty text node value doesn't do anything but the non-empty text node still makes some action.

I think you are referring to potential use case when text is set even when a node has other children except text node.
This is handled properly in a different place https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/component/template/internal/IdCollector.java#L220

But anyway: this PR makes the behavior inconsistent regardless of the executed action.

denis-anisimov:
If your usecase contradicts to the feature which I mentioned : initializing the Java template with attribute/text node values from the template content then this should be discussed with Pekka.
I will be happy to revert this feature: I never liked it.

Artur-:
Yes, if you set the text to "" you wipe out all child elements. It would be fine if the server just set it on the server side but it also sends a "remove all children" to the client side. If the template has a text, fine. It is unlikely that the client side element then will add other child elements in addition to the text.

What does this PR make inconsistent?

denis-anisimov:
The inconsistency is :

I have a template with render:
```
render() {
    return html`<div id="root"></div>`;
  }
```

And it works fine : your expectation is fulfilled.
Then I decided to change my template for some reason to

```
render() {
    return html`<div id="root">Now I want to have a text here</div>`;
  }
```
And once I've changed the template content the expectation is failed.
Why ?
If I don't know the implementation details I consider this as a bug.

denis-anisimov:
>Yes, if you set the text to "" you wipe out all child elements. It would be fine if the server just set it on the server side but it also sends a "remove all children" to the client side. 

That contradicts to another old feature: when you set the text to the template then it has to remove all the chidlren.
Otherwise it will be inconsistent (again!).
This is a standard behavior implemented several years back because there was no other way to implement `setText()` properly for children inside templates.

Artur-:
If you have
```
render() {
    return html`<div id="root"></div>`;
  }
```
and map the div, then div.getText() returns ""

If I have
```
render() {
    return html`<div id="root">Now I want to have a text here</div>`;
  }
```
and map the div, then div.getText() returns "Now I want to have a text here"

Where is the inconsistency that I am missing?

Artur-:
> when you set the text to the template then it has to remove all the chidlren
Yes, you should not set the text for anything when READING the template. You should GET the values from the templates so they are correct. If I do `div.setText("Foo")` then it will wipe out all children

denis-anisimov:
In the first case the code 
```
if (value != null && !"".equals(value)) {
            element.setText(value);
        }
```

doesn't set the text at all and the client side won't be updated to the value.

In the second case the text is set and will be sent to the client side.

So if there is call (like in the original ticket description) on the client side which is executed right away after the element is created BUT _before_ the server side applies it changes then in the first case the call will be applied and in the second case the call will be no-op.

This is inconsistency.

Artur-:
In neither case, the value should be set on the client side. It should be read by the server.

denis-anisimov:
>In neither case, the value should be set on the client side. It should be read by the server.

May be.

But this is a completely new feature which has never been implemented.

In this PR we are talking about this PR.
The behavior in this PR is inconsistent.

I'm totally fine if you find a way to implement this `In neither case, the value should be set on the client side. It should be read by the server.` properly: this is far away from trivial.

Artur-:
Also the user has no expectations or does not care if the server happens to set a string "foo" where there is already a string "foo". But if the server wipes out everything a web component creates, then it is a problem

denis-anisimov:
>Also the user has no expectations or does not care if the server happens to set a string "foo" where there is already a string "foo". But if the server wipes out everything a web component creates, then it is a problem

The text is set only when there is no children (initially).
The problem arises only when there is intermediate client side call which modifies the DOM in between.

I agree that the current behavior can be considered confusing.
And as I said: I never liked this "feature" which sets the attibutes/text values based on the template content.
I felt from the very beginning that it will cause problems.

But this feature has been specified and implemented as requested.
I will be happy to get rid of this feature.

Artur-:
I still don't know what this would break or make inconsistent though because currently:
1. If the template element contains only a text, it is sent from the server to the client for no good reason (it already exists in the client)
2. In other cases, the server does not send anything

If the correct behavior is:
1. The server does not send anything about template element contents to the client

Then this fix moves things one step closer to that direction by changing it to:
1. If the template element contains only a **non-empty** text, it is sent from the server to the client for no good reason (it already exists in the client)
2. In other cases, the server does not send anything

So it moves one more case from 1. to 2, leaving only the case when the text is non-empty

denis-anisimov:
Take ticket  #10106.

This PR fixes exactly the description in the ticket.

change the `render` code in the example to
```
render() {
    return html`<div id="root">I'm here to keep the example broken</div>`;
  }
```

this PR doesn't fix this ticket.

Once this PR is  merged I can create exactly the same ticket with updated renderer.

We may argue about whether the text update should be sent or not but one behavior with empty text and another behavior with non-empty text is definitely a bug.
This PR introduces a bug.

I don't have anything to add here. May be someone else has another opinion about this: the PR is not closed.
My opinion is : the PR attempts to improve a little bit something but introduces a serious inconsistency which is just a bug from my point of view.

The workaround is extremely simple: use `return html'<div id="root"><some-child/></div>'`.

I also recall some discussion about optional possibility to prevent setting attributes/text values for the element (e.g. via a specific attribute whose presence says that feature should not be used). But this has not been agreed/designed and implemented. Can be an enhancement.

Artur-:
Changed implementation so that it never sends the text values to the client but sets them on the server as before

Artur-:
Ok this does not work either as for some reason there is this logic in `NodeList`:
```
        if (isRemoveAllCalled && !hasRemoveAll) {
            changes = Stream
                    .concat(Stream.of(new ListClearChange<>(this)),
                            allChanges.stream())
```

Artur-:
Fixed test to actually test the correct case and the implementation so it works...

Artur-:
Some more digging. I think there is a test missing that would update the text value later on and that test would probably be broken by this PR

denis-anisimov:
> Some more digging. I think there is a test missing that would update the text value later on and that test would probably be broken by this PR

Yes, it will.
This PR tries to address an issue which is fundamentally impossible in the current design: read text value on the server without modifying the client.
The text is not a property: it's something inside Text Node which is created as a child of the component where the text is set.
That's why `removeAllChildren` is necessary (which removes all client side) : it should remove all the children nodes which has no server side mapping (otherwise the text will be duplicated on the client at least).
Also it creates a (text) StateNode in the server and add it as a child.
As a result it makes a hierarchy which has to be in sync with client.
What you are trying to do is: avoid sending this StateNode to the client (whatever way). If it's not set then the hierarchy on the server side is not the same as on the client side.
So it produces inconsistency which lead to various bugs:
- setting a new  text value for the component
- the component on the server side now contains a child and any action with this child will be broken since it doesn't exist on the client side (as a client side `StateNode`).

This is the overall problem: it's fundamental issue with the current design.

This PR has its own issues: it totally breaks any encapsulation. Change tracker is an internal state of the feature and it may not be exposed outside. This is already broken in fact : the StateTree/StateNode has this access . But this is already existing issue: at least no one else should get access to it. Making the method public totally ruins encapsulation and makes the code unmaintainable: I already had issues with the changes tracker because of current design.

pleku:
Closing as the issue is on the _needs design_ phase still.

PR #10118, opened 2021-02-24
chore: detect illegal LookupServletContainerInitializer execution in OSGi

denis-anisimov:


flow-server/src/main/java/com/vaadin/flow/di/AbstractLookupInitializer.java
```diff
@@ -0,0 +1,106 @@
...
+ * one) may be available in the web application classpath and it's provided by
+ * the developers for the framework support (the main usecase here is Spring
+ * add-on).
+ * </ul>
+ * 
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+public interface AbstractLookupInitializer {
```
mshabarov, line 69: Why is this interface needed?
I suppose two reasons:
1. This interface will be implemented on OSGi side in some other way, rather that for Servlets 3.0
2. This is needed to properly catch the case when `classSet` does not contain the `LookupInitializer` instance, which may happen if OSGi container will implement servlet initializer somehow.

If the first option is true, then I'd propose to exclude all Servlet 3.0 mentions from javadoc of this class.

mshabarov, line 69: Is it better to rename those classes to `LookupInitializer` (interface), `DefaultLookupInitializer` (impl)?

denis-anisimov: Neither.

OSGi container should not run `ServletContainerInitializer` ever : it has no sense for OSGi.
PAX-web (Karaf) is a mixture of OSGi and Servlet container: it allows to run WARs inside it.
As a result it supports `ServletContainerInitializer`  and runs them. But it should not run `ServletContainerInitializer`  for non WAR case.
This is technically a bug in PAX-web but nobody cares about this. We need to support it and this is a workaround which is described here: https://github.com/vaadin/flow/pull/10118/files#diff-dee652b476265e2d89027e7d76283f2c1a8072aa0fcc5f3eeaf780b88a7a39d6R61

denis-anisimov: That will require changes in Spring add-on at least: this is a breaking change technically.
I don't want to introduce a breaking change though may be this is not a big deal: Flow 6.0 is not released yet.

The name is indeed not really suitable but I couldn't find better avoiding renaming.

mshabarov: Sure, got it! But why is a new interface `AbstractLookupInitializer ` needed? What happens if you just use `LookupInitializer` and check if it's being contained in `classSet`?

denis-anisimov: Depending on container.

Some container (like Jetty) doesn't pass the class itself into `ServletContainerInitializer` : only its subclasses.
Other containers passes also the classes from `HandTypes`  annotation.

In any case `LookupInitializer` has to be passed since it implements `AbstractLookupInitializer` and `AbstractLookupInitializer`  is in  `HandTypes`. Without `AbstractLookupInitializer` the behavior is non deterministic.

mshabarov: Yep, and assuming the flow 6.0 is in rc now...Let's not to change it.

mshabarov: Good, so it's a unified solution for different containers, thanks!

mshabarov (approved):

mshabarov:
The build failed for some strange reason with `5.0-SNAPSHOT` dependency:
```
[ERROR] Failed to execute goal com.vaadin:flow-maven-plugin:1111.5.0-SNAPSHOT:build-frontend (default) on project flow-test-no-theme:
```
I decided to rerun it again.

PR #10128, opened 2021-02-25
fix: treat "clear all" as permanent initial state

### LINKED ISSUE #10119: Button in a Dialog with LitTemplate: Button caption gets duplicated when dialog is closed and reopened

The caption of Buttons in a Dialog that contains a LitTemplate get duplicated if the dialog is opened more than once (if the Button is mapped server-side with  @Id).
Short gif highlighting the issue:
vaadin-dialog-okok

The code is based on vaadin/skeleton-starter-flow-spring - v18.
Client side template:
```typescript
@customElement('test-form')
export class TestForm extends LitElement {
    render() {
        return html`
        <vaadin-button id="submitButton" theme="primary">OK</vaadin-button>
        `;
    }
}
```
Server side template:
```java
@Tag("test-form")
@JsModule("./views/test-form.ts")
public class TestForm extends LitTemplate {
    
    @Id("submitButton")
    Button submitButton;
}

```
View:
```java

    public MainView() {
        TestForm testForm = new TestForm();
        Dialog dialog = new Dialog(testForm);

        Button button = new Button("Click");
        button.addClickListener(ce -> dialog.open());
        add( button);
    }
}
```

### Expected behavior

I would expect the caption to stay the same.

### Actual behavior

The caption is rendered twice.

### Versions
    - Vaadin / Flow version: 18.0.6
    - Java version: 8
    - OS version: Linux
    - Application Server (if applicable): Tomcat embedded
    - IDE (if applicable):

denis-anisimov:
fixes #10119

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
@@ -357,9 +357,8 @@ public abstract class NodeList<T extends Serializable> extends NodeFeature {
     @Override
     public void generateChangesFromEmpty() {
         if (isRemoveAllCalled) {
-            // if list has "clear" change and has no any other changes then
-            // this change should be stored in the tracker otherwise it will be
-            // incorrectly postponed
+            // if list ever had "clear" change then it
+            // should be stored in the tracker
             addChange(new ListClearChange<>(this));
         }
         if (values != null) {
--- a/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
@@ -28,7 +28,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import com.vaadin.flow.internal.StateNode;
 import com.vaadin.flow.internal.change.AbstractListChange;
@@ -288,8 +287,6 @@ public abstract class NodeList<T extends Serializable> extends NodeFeature {
 
     @Override
     public void collectChanges(Consumer<NodeChange> collector) {
-        boolean hasRemoveAll = false;
-
         // This map contains items wrapped by AbstractListChanges as keys and
         // index in the following allChanges list as a value (it allows to get
         // AbstractListChange by the index)
@@ -316,7 +313,6 @@ public abstract class NodeList<T extends Serializable> extends NodeFeature {
                 ((ListAddChange<T>) change).getNewItems()
                         .forEach(item -> indices.put(item, i));
             } else if (change instanceof ListClearChange<?>) {
-                hasRemoveAll = true;
                 allChanges.clear();
                 indices.clear();
                 index = 0;
@@ -329,17 +325,8 @@ public abstract class NodeList<T extends Serializable> extends NodeFeature {
 
         List<AbstractListChange<T>> changes;
 
-        if (isRemoveAllCalled && !hasRemoveAll) {
-            changes = Stream
-                    .concat(Stream.of(new ListClearChange<>(this)),
-                            allChanges.stream())
-                    .filter(this::acceptChange).collect(Collectors.toList());
-        } else {
-            changes = allChanges.stream().filter(this::acceptChange)
-                    .collect(Collectors.toList());
-        }
-
-        isRemoveAllCalled = false;
+        changes = allChanges.stream().filter(this::acceptChange)
+                .collect(Collectors.toList());
 
         if (isPopulated) {
             changes.forEach(collector);
@@ -369,16 +356,16 @@ public abstract class NodeList<T extends Serializable> extends NodeFeature {
 
     @Override
     public void generateChangesFromEmpty() {
+        if (isRemoveAllCalled) {
+            // if list has "clear" change and has no any other changes then
+            // this change should be stored in the tracker otherwise it will be
+            // incorrectly postponed
+            addChange(new ListClearChange<>(this));
+        }
         if (values != null) {
             assert !values.isEmpty();
             getChangeTracker().add(new ListAddChange<>(this, isNodeValues(), 0,
                     new ArrayList<>(values)));
-        } else if (isRemoveAllCalled) {
-            // if list has "clear" change and has no any other changes then
-            // this change should be stored in the tracker otherwise it will be
-            // incorrectly postponed
-            isRemoveAllCalled = false;
-            addChange(new ListClearChange<>(this));
         } else if (!isPopulated) {
             // make change tracker available so that an empty change can be
             // reported
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/littemplate/TestFormIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/littemplate/TestFormIT.java
@@ -26,7 +26,7 @@ import com.vaadin.testbench.TestBenchElement;
 public class TestFormIT extends ChromeBrowserTest {
 
     @Test
-    public void reattachedTemplateHasItsInitialText() {
+    public void reattachedTemplateHasExplicitlySetText() {
         open();
 
         WebElement button = findElement(By.id("click"));
@@ -38,7 +38,7 @@ public class TestFormIT extends ChromeBrowserTest {
                 .id("form-template");
         TestBenchElement div = template.$(TestBenchElement.class).id("div");
 
-        Assert.assertEquals("Template text", div.getText());
+        Assert.assertEquals("foo", div.getText());
 
         // detach template
         button.click();
@@ -48,7 +48,7 @@ public class TestFormIT extends ChromeBrowserTest {
 
         template = $(TestBenchElement.class).id("form-template");
         div = template.$(TestBenchElement.class).id("div");
-        Assert.assertEquals("Template text", div.getText());
+        Assert.assertEquals("foo", div.getText());
     }
 
 }
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/littemplate/TestFormIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/littemplate/TestFormIT.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.uitest.ui.littemplate;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+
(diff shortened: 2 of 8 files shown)


flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/littemplate/TestFormIT.java
```diff
@@ -0,0 +1,54 @@
...
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+
+import com.vaadin.flow.testutil.ChromeBrowserTest;
+import com.vaadin.testbench.TestBenchElement;
+
+public class TestFormIT extends ChromeBrowserTest {
+
+    @Test
+    public void reattachedTemplateHasItsInitialText() {
```
knoobie, line 29: I'm wondering by this test name.. please ignore if totally out of scope. just wondering. Does that mean the following:

```java
// already attached to the client
testForm.getDiv().setText("Bar");
// already attached client side is updated to "Bar"

// remove TestForm from client
// reattach TestForm to client
```
Does TestForm has "Template text" or "Bar" on the client? From the test name I would expect "Template text" but from my "state model understanding" I would think it should be "Bar"?

denis-anisimov: There is no "Bar" in the this scenario.

The original ticket is about duplicated text.
So "Template text" will be duplicated after reattach.

There is an implicit call `setText("Template text")`.

May be you are right that I should avoid implicit `setText` and call explicitely `setText("foo"` and check that this is shown in the result...

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
```diff
@@ -369,16 +356,16 @@ public void forEachChild(Consumer<StateNode> action) {
 
     @Override
     public void generateChangesFromEmpty() {
-        if (values != null) {
-            assert !values.isEmpty();
-            getChangeTracker().add(new ListAddChange<>(this, isNodeValues(), 0,
-                    new ArrayList<>(values)));
-        } else if (isRemoveAllCalled) {
+        if (isRemoveAllCalled) {
```
Legioth, line 359: Seems like this flag is never reset. This means that the semantics of `.clear()` and by extension also `Element::removeAllChildren` is significantly changed without any changes to the documentation of either method.

Isn't the actual problem here that `LitTemplate` doesn't do `removeAllChildren()` again on the id mapped element when it is reattached?

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
```diff
@@ -369,16 +356,16 @@ public void forEachChild(Consumer<StateNode> action) {
...
             // if list has "clear" change and has no any other changes then
```
Legioth, line 360: This comment no longer makes sense in this way since it implies that this would happen only if there are no other changes.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/littemplate/TestFormIT.java
denis-anisimov: Done.

knoobie: Thanks that makes the intention more clear and additionally checks that the state is stored in between attachments and not "overwritten" when reattached.

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
denis-anisimov: The flag is never reset.

The semantic is not changed: please provide a test otherwise.

>Isn't the actual problem here that LitTemplate doesn't do removeAllChildren() again on the id mapped element when it is reattached?

Why should it call anything ?

The problem is not related to LitTemplate anyhow at all.
The problem is : `removeAllChildren`  is not a state (and an element after reattaching restores only its state: `removeAllChildren`  disappear). And as a result it may not exist in API at all.

The problem may be reproduced with any Element which has client side children not known to the server.
This PR makes "clear all" /"removeAllChildren " a permanent initial state.

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
denis-anisimov: Done.

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
Legioth: The change in semantics can be observed with any regular element (i.e. not used through `@Id`) that adds client-side elements to itself when it's attached in the browser.

```
Element e = new Element("adds-children-on-client");
e.removeAllChildren();
attachedElement.appendChild(e);

// In another round trip
e.removeFromParent();

// In yet another round trip
attachedElement.appendChild(e);
```

Before this change, this would result in an element that has restored its original contents. After this change, the element would be empty after reattaching just like it was before it was detached. I suspect the new semantics are more in line with what a developer would expect, but it's regardless a change in semantics and should thus be documented.

From another point of view, the problem is that `LitElement` assumes that `removeAllChildren()` is persistent. `removeAllChildren()` happens the first time it's attached, but not the following times. Fixing it to ensure `removeAllChildren()` is called again might still not be the best way if we assume the new semantics of `removeAllChildren()` are indeed preferable.

denis-anisimov: I don't understand the test from this description.

>Before this change, this would result in an element that has restored its original contents. 

Which original content ?
The original content of ANY element is empty.

And regardless of this fix the original content will be restored: because it's empty.

"clear all" / "remove all"  _initial_ _EVENT_ is no-op ALWAYS for a server side element: because it's state is preserved as a state.
" removeAll" is never called in this fixed : so the state is never destroyed.
`ClearAll` event is added in the changes list once the element is attached : it has no any relation to the " removeAll" call.
As said: it's no-op for any sever side element.
And it removes the client side children for the element which has some client side mapping.

I don't see any changes in the semantic.

I spent enough time on this ticket and I'm convinced in its work.

There are tests which proves that it works as it should: they are enough for me.

Please write exact unit test which I can copy paste and see that there is a failure.

Legioth: > The original content of ANY element is empty.
Not if it's a custom element that adds something to itself in its client-side constructor (or `connectedCallback`), such as the one that is described in #10106. The semantic changes for such elements but not for native elements or custom elements that only modify their shadow root.

This kind of element might seem like a hypothetical case but they do actually exist in real-world cases, which is how Artur originally discovered #10106 (the case he describes in the ticket is just an artificial reduced test case).

mshabarov (dismissed):
I'm not certain whether we need to update the javadoc for `generateChangesFromEmpty()`.
From a modified implementation point of view, it now adds a clear change even if the values are empty.
But Denis's explanations looks reasonable.

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
Legioth: So the exact test case would be to use `/views/empty/empty-view.ts` from #10106 together with this Java class:
```
@Tag("empty-view")
public class EmptyView extends Component {
  public EmptyView() {
    getElement().removeAllChildren();
  }
}
```

When an instance of this component is initially attached, it will show as empty regardless of this PR. `firstUpdated` will be run but its effects will be reset by the clear event. (There is a small risk that the asynchronous nature of the client-side `LitElement` class actually means that `firstUpdated` would be run only after the clear event has been applied. In that case, the same effect can be applied manually using `connectedCallback` instead of relying on `firstUpdated`.)

When that instance is then detached in one round trip and attached again in another round trip, then without this PR it will show the "Hello" text because there will be a new client-side instance that runs `firstUpdated` again but there won't be any clear event from the server. With this PR, a clear event is sent which leads to an empty element.

denis-anisimov: True.

And this is aligned with the purpose of the `removeAllChildren`: it has been introduced to be able to set Text properly for the Templates. Without calling `removeAllChildren` it will _append_ the text to the existing one.
So if I set the text to the template Element the new text becomes _a new state_ of this element. And it should be preserved.

So for me personally there is no any doubt that `removeAllChildren` should behave as it behaves now in this PR.

Technically you just described the matter of the #10119 bug: there is an implicit call `setText("OK")`  when the template is created.
`setText("OK")` means:
- clear all children  via `removeAllChildren`  (so there is in fact this implicit call )
- append a text node "OK" as a child.

This is new state and it should be preserved now.
Replace "OK" with "foo" and it becomes absolutely clear that it should be preserved: since you set the text explicitely.

>When that instance is then detached in one round trip and attached again in another round trip, then without this PR it will show the "Hello" text because there will be a new client-side instance that runs firstUpdated again but there won't be any clear event from the server. With this PR, a clear event is sent which leads to an empty element.

True. But this is just a bug.
And this PR fixes this bug.
I don't see why this is a semantic change because I don't expect the buggy behavior. 
We have a generic issue with _ANY_ bug fix: someone may start to rely on the old buggy behavior and in that case the logic previously worked fine will be broken. I don't recall the working solution for this.

I think it should be just mentioned in the release notes.

One argument about state: we never preserve pure client side state (your scenario describes a preserved client side state).
The same will be true if Text node is replaced with attribute value: `firstUpdated` /`connectedCallback` or any other JS client side function may set an attribute value at any time.
If there is an attribute value on the server side for this attribute available in the state then the attribute value set from the client side via JS will be lost: it will be overriden by the server side state after reattach.

I don't see any description of this behavior for attribute/property anywhere (though I'm not familiar with the current state of docs: may be it somewhere mentioned): at least it's not documented in javadocs.

So I don't see a reason to make a specific explanation for  `removeAllChildren`: like don't expect the buggy behavior but expect the correct behavior.

But if we have already some description of old buggy behavior for  `removeAllChildren` then of course it needs to be clarified.
- I'm pretty sure that there is no need to update javadocs for this method because the reasons I've already described about attributes/properties. The correct semantic is implied by the fact that the method has been introduced exactly to fix the correct (stateful) behavior for `setText` .
- If we have somewhere in docs the description of `removeAllChildren` and it describes its buggy behavior then it should be fixed of course. But please let me know where such description is: I will update the docs.

I doubt that  `removeAllChildren` has any description in any docs since this method is quite specific (introduced specifically for `setText` ). So it's rather helper method than something which is worth by itself.

Legioth: I agree that the new behaviour seems to be more in line with what developers would expect.

I'm cautious because in this case a fix to one bug (related to `@Id` elements) is also having impact on an unrelated case (regular elements that are not `@Id` mapped). We need to be especially certain that we're not missing any details right now when we're so close to the RC.

This isn't only about the relatively internal `removeAllChildren` method, but also e.g. `HasComponents::removeAll` which by default delegates to `removeAllChildren`.

Still, considering that the semantics are changing only for a quite special edge case, it might indeed be appropriate to leave the documentation as is.

Mentioning the potential change in the release notes might still be a good idea.

mshabarov (approved):
> When that instance is then detached in one round trip and attached again in another round trip, then without this PR it will show the "Hello" text because there will be a new client-side instance that runs firstUpdated again but there won't be any clear event from the server. With this PR, a clear event is sent which leads to an empty element.

Apart from the knowledge on how the state nodes work under the hood, I would say that having empty text after first attach, and, then, having "Hello" text after subsequent attach/reattach, would be confusing and opens a gates for bugs, and, as Denis said, a faulty costumer-side code which relies on that behaviour. On the other hand, having the clear all state constantly looks more reasonable, because this behaviour at least expectable.

> I think it should be just mentioned in the release notes.

Yes, and I believe we shouldn't maybe release this fix in the flow 2.5 immediately, but, instead, have some test period in latest flow versions to be sure that this would not cause a new regressions.

Issue #9647, opened 2020-12-14
Enabling session expired notification prevents seamless logout button


mvysny:
Enabling the session-expired notification via `CustomizedSystemMessages.setSessionExpiredNotificationEnabled(true)` prevents me from having a seamless logout button since the "Session Expired" message flashes briefly when logging out.

The logout button is implemented as usual:

```java
        Button button = new Button("Logout",
                e -> {
                    UI.getCurrent().getPage().setLocation("/");
                    VaadinSession.getCurrent().close();
                    VaadinSession.getCurrent().getSession().invalidate();
                } );
```

### Expected behavior

Pressing the button should reload the page without showing the "session expired" notification.

### Actual behavior

Pressing the button will briefly show a red "session-expired" notification then reloads the page.

### Workarounds

1. Using `UI.getCurrent().getPage().open("/", "_self");` instead of `setLocation()` as suggested at https://github.com/vaadin/framework/issues/4255 doesn't help - the "session expired" notification still flashes.
2. Disable the session expired notification.

### Versions:
    - Vaadin / Flow version: 14.4.4 / 2.4.3
    - Java version: 11

mvysny:
Example project which reproduces the issue
[skeleton-starter-flow.zip](https://github.com/vaadin/flow/files/5689021/skeleton-starter-flow.zip)

denis-anisimov:
Hmmmm....
That should have been fixed.

At least we have a test for this case : `com.vaadin.flow.uitest.ui.LogoutView`.
Though if I recall correctly the main issue there was related to PUSH: it refused to redirect at all.
The reason was that session expired message is handled on the client without knowing that the app is closed at that point.
As a fix the application close has been added to the JS executed in `Page::open`: 
`"if ($1 == '_self') this.stopApplication(); window.open($0, $1)"`.

I think now it's a race condition because session is closed before redirect is done: the session expired message is handled on the client side _before_ the redirect JS is executed.
I think this is the case.
And I assume the steps to reproduce should sometimes work and sometimes doesn't.

mshabarov:
This issue has been fixed by making JavaScript commands execution happens before "session expire" and "app error" message handling in `MessageHandler` class:
Before:
```java
            ValueMap meta = valueMap.getValueMap("meta");

            if (meta != null) {
                // Message handling code
            }

            // Invokes all the flush listeners, including JS commands execution 
            Reactive.flush();
```

After:
```java
            // Invokes all the flush listeners, including JS commands execution 
[...]

Issue #9705, opened 2020-12-21
LogoutIT is flaky


denis-anisimov:
LogoutIT.setLocation_noErrorMessages

https://bender.vaadin.com/viewLog.html?buildId=205639&buildTypeId=Flow_Flow25_Nightly_FlowFelixJettyIntegrationTests25

I'm pretty sure this is the bug for the test:
https://github.com/vaadin/flow/issues/9647

joheriks:
Note that there was already an [existing exception in the IT](https://github.com/vaadin/flow/blob/master/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/LogoutIT.java#L34) to not fail for "Session expired" which could occur spuriously.
The test is now flaky because the logged error has more information:

```[SEVERE] http://172.16.1.226:8888/VAADIN/static/client/client-C4B53966888EAB6CBD0D20BD246409CB.cache.js 211:19 "Session Expired"```

So this could also be due to Chrome update. In the `offline` branch, the condition on the error log is relaxed as follows:
https://github.com/vaadin/flow/blob/feature/offline/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/LogoutIT.java#L35

If these log messages are due to an actual bug, that should of course be fixed and the relaxations in the ITs removed.

**Comment by denis-anisimov** [comment]
As I understand the `Session Expired` log message is a bug ( #9647 ) : it just should not be there.

**Comment by mshabarov** [comment]
No corresponding fix found so far, but most likely the issue is in client engine functionality in `flow-client`.
I temporary give it back to backlog and switch to more urgent issue.

**Comment by denis-anisimov** [comment]
```
Caused by: java.lang.AssertionError: Received error message in browser log console right after opening the page, message: [2020-12-21T00:26:05+0000] [SEVERE] http://172.16.1.226:8888/VAADIN/static/client/client-C4B53966888EAB6CBD0D20BD246409CB.cache.js 211:19 "Session Expired"
	at com.vaadin.flow.testutil.TestBenchHelpers.lambda$checkLogsForErrors$4(TestBenchHelpers.java:378)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at com.vaadin.flow.testutil.TestBenchHelpers.checkLogsForErrors(TestBenchHelpers.java:373)
	at com.vaadin.flow.uitest.ui.LogoutIT.setLocation_noErrorMessages(LogoutIT.java:34)
```

[...]

PR #10142, opened 2021-02-26
fix: Execute error handling as a last action in message handler

mshabarov:
Makes error handling the last action in the message handling queue, and ensures that the error message won't be shown if the page redirect has been requested on server side.

Fixes https://github.com/vaadin/flow/issues/9647
Related-to https://github.com/vaadin/flow/issues/9705

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/communication/MessageHandler.java
+++ b/flow-client/src/main/java/com/vaadin/client/communication/MessageHandler.java
@@ -421,31 +421,26 @@ public class MessageHandler {
 
             if (meta != null) {
                 Profiler.enter("Error handling");
+                final UIState uiState = registry.getUILifecycle().getState();
                 if (meta.containsKey(JsonConstants.META_SESSION_EXPIRED)) {
                     if (nextResponseSessionExpiredHandler != null) {
                         nextResponseSessionExpiredHandler.execute();
-                    } else {
-                        if (registry.getUILifecycle()
-                                    .getState() != UIState.TERMINATED) {
-                            registry.getSystemErrorHandler()
-                                    .handleSessionExpiredError(null);
-                            registry.getUILifecycle().setState(UIState.TERMINATED);
-                        }
-                    }
-                } else if (meta.containsKey("appError")) {
-                    if (registry.getUILifecycle()
-                                .getState() != UIState.TERMINATED) {
-                        ValueMap error = meta.getValueMap("appError");
-
-                        registry.getSystemErrorHandler().handleUnrecoverableError(
-                                error.getString("caption"),
-                                error.getString("message"),
-                                error.getString("details"),
-                                error.getString("url"),
-                                error.getString("querySelector"));
-
+                    } else if (uiState != UIState.TERMINATED) {
+                        registry.getSystemErrorHandler()
+                                .handleSessionExpiredError(null);
                         registry.getUILifecycle().setState(UIState.TERMINATED);
                     }
+                } else if (meta.containsKey("appError")
+                        && uiState != UIState.TERMINATED) {
+                    ValueMap error = meta.getValueMap("appError");
+
+                    registry.getSystemErrorHandler().handleUnrecoverableError(
+                            error.getString("caption"),
+                            error.getString("message"),
+                            error.getString("details"), error.getString("url"),
+                            error.getString("querySelector"));
+
+                    registry.getUILifecycle().setState(UIState.TERMINATED);
                 }
                 Profiler.leave("Error handling");
             }
--- a/flow-client/src/main/java/com/vaadin/client/communication/MessageHandler.java
+++ b/flow-client/src/main/java/com/vaadin/client/communication/MessageHandler.java
@@ -415,6 +415,8 @@ public class MessageHandler {
                     + (Duration.currentTimeMillis() - processUidlStart)
                     + " ms");
 
+            Reactive.flush();
+
             ValueMap meta = valueMap.getValueMap("meta");
 
             if (meta != null) {
@@ -423,26 +425,31 @@ public class MessageHandler {
                     if (nextResponseSessionExpiredHandler != null) {
                         nextResponseSessionExpiredHandler.execute();
                     } else {
-                        registry.getSystemErrorHandler()
-                                .handleSessionExpiredError(null);
-                        registry.getUILifecycle().setState(UIState.TERMINATED);
+                        if (registry.getUILifecycle()
+                                    .getState() != UIState.TERMINATED) {
+                            registry.getSystemErrorHandler()
+                                    .handleSessionExpiredError(null);
+                            registry.getUILifecycle().setState(UIState.TERMINATED);
+                        }
                     }
                 } else if (meta.containsKey("appError")) {
-                    ValueMap error = meta.getValueMap("appError");
+                    if (registry.getUILifecycle()
+                                .getState() != UIState.TERMINATED) {
+                        ValueMap error = meta.getValueMap("appError");
 
-                    registry.getSystemErrorHandler().handleUnrecoverableError(
-                            error.getString("caption"),
-                            error.getString("message"),
-                            error.getString("details"),
-                            error.getString("url"),
-                            error.getString("querySelector"));
+                        registry.getSystemErrorHandler().handleUnrecoverableError(
+                                error.getString("caption"),
+                                error.getString("message"),
+                                error.getString("details"),
+                                error.getString("url"),
+                                error.getString("querySelector"));
 
-                    registry.getUILifecycle().setState(UIState.TERMINATED);
+                        registry.getUILifecycle().setState(UIState.TERMINATED);
+                    }
                 }
                 Profiler.leave("Error handling");
             }
             nextResponseSessionExpiredHandler = null;
-            Reactive.flush();
 
             lastProcessingTime = (int) (Duration.currentTimeMillis() - start);
             totalProcessingTime += lastProcessingTime;
--- a/flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
+++ b/flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
@@ -25,7 +25,9 @@ import com.google.gwt.user.client.Timer;
 import com.vaadin.client.communication.MessageHandler;
 import com.vaadin.client.communication.MessageSender;
 import com.vaadin.client.communication.RequestResponseTracker;
+import com.vaadin.client.flow.ExecuteJavaScriptProcessor;
 import com.vaadin.client.flow.StateTree;
+import com.vaadin.flow.shared.JsonConstants;
 import com.vaadin.flow.shared.ui.Dependency;
 import com.vaadin.flow.shared.ui.LoadMode;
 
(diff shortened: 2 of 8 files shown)


denis-anisimov:
The solution looks good for me.
But I would like to know whether it's possible to write a GWT unit test for this ?

We have some such tests in `GwtMessageHandlerTest`  . Please check whether this functionality can be unit tested.

The ITs are not reliable for this scenario.

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/LogoutView.java
```diff
@@ -19,16 +19,38 @@
...
         });
-        add(button);
+        logoutButton.setId("logout-button");
+        add(logoutButton);
+
+        NativeButton notificationLogoutButton = new NativeButton(
+                "logout-with-notification", ev -> {
+                    VaadinService.getCurrent().setSystemMessagesProvider(
+                            (SystemMessagesProvider) systemMessagesInfo -> {
+                                final CustomizedSystemMessages systemMessages = new CustomizedSystemMessages();
```
taefi, line 41:
```suggestion
                                final CustomizedSystemMessages systemMessages = 
                                                new CustomizedSystemMessages();
```

mshabarov:
> But I would like to know whether it's possible to write a GWT unit test for this ?

Sure! I will check it.

flow-client/src/main/java/com/vaadin/client/communication/MessageHandler.java
```diff
@@ -415,34 +415,36 @@ assert getServerId(valueMap) == -1
...
                         nextResponseSessionExpiredHandler.execute();
-                    } else {
+                    } else if (uiState != UIState.TERMINATED) {
                         registry.getSystemErrorHandler()
                                 .handleSessionExpiredError(null);
                         registry.getUILifecycle().setState(UIState.TERMINATED);
                     }
-                } else if (meta.containsKey("appError")) {
+                } else if (meta.containsKey("appError")
+                        && uiState != UIState.TERMINATED) {
```
taefi, line 434: Not sure if applicable, but what happens if `meta.containsKey("appError")` resolves to true and also `uiState == UIState.TERMINATED` returns true?

taefi (changes requested):

flow-client/src/main/java/com/vaadin/client/communication/MessageHandler.java
mshabarov: If `uiState == UIState.TERMINATED` is `true`, then it means that the redirect has been occurred and `stopApplication` method has been invoked. In that case, it is pointless to show the message, because the page has been redirected, and this error message is only relevant to old page.

taefi (approved):

taefi:
Do not merge until a decision made for adding GWT tests to this PR.

flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
```diff
@@ -250,19 +283,166 @@ public void testForceHandleMessage_resyncIsRequested() {
...
+
+        JsonObject meta = Json.createObject();
+        meta.put(JsonConstants.META_SESSION_EXPIRED, true);
+        messagePayload.put("meta", meta);
+
+        JsonArray jsCommandsArray = Json.createArray();
+        messagePayload.put(JsonConstants.UIDL_KEY_EXECUTE, jsCommandsArray);
+
+        JsonArray redirectCommand = Json.createArray();
+        redirectCommand.set(0,
```
denis-anisimov, line 308: This looks overcomplicated and not a unit tests but rather IT.

The code should be unit tested: so there is no need to emulate various causes then application is terminated.
`this.stopApplication();` terminates the application.
You just need terminated application: terminate it directly via `UILifecycle::setState` .

flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
```diff
@@ -250,19 +283,166 @@ public void testForceHandleMessage_resyncIsRequested() {
...
+
+        JsonObject meta = Json.createObject();
+        meta.put("appError", true);
+        messagePayload.put("meta", meta);
+
+        JsonArray jsCommandsArray = Json.createArray();
+        messagePayload.put(JsonConstants.UIDL_KEY_EXECUTE, jsCommandsArray);
+
+        JsonArray redirectCommand = Json.createArray();
+        redirectCommand.set(0,
```
denis-anisimov, line 346: Same here `UILifecycle::setState `.

denis-anisimov (changes requested):

flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
mshabarov: Agree. Done

flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
mshabarov: Done

flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
```diff
@@ -250,19 +283,156 @@ public void testForceHandleMessage_resyncIsRequested() {
...
+        JavaScriptObject messagePayloadJS = JavaScriptObject.createObject();
+        JsonObject messagePayload = messagePayloadJS.cast();
+
+        JsonObject meta = Json.createObject();
+        meta.put("appError", true);
+        messagePayload.put("meta", meta);
+
+        // when: UI has been terminated (for instance, as a result of redirect
+        // JS caused by Page::setLocation)
+        getUILifecycle().setState(UILifecycle.UIState.RUNNING);
```
denis-anisimov, line 335: Is this needed ?

flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
```diff
@@ -250,19 +283,156 @@ public void testForceHandleMessage_resyncIsRequested() {
...
+        JavaScriptObject messagePayloadJS = JavaScriptObject.createObject();
+        JsonObject messagePayload = messagePayloadJS.cast();
+
+        JsonObject meta = Json.createObject();
+        meta.put(JsonConstants.META_SESSION_EXPIRED, true);
+        messagePayload.put("meta", meta);
+
+        // when: UI has been terminated (for instance, as a result of redirect
+        // JS caused by Page::setLocation)
+        getUILifecycle().setState(UILifecycle.UIState.RUNNING);
```
denis-anisimov, line 303: Is this needed ?

flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
```diff
@@ -250,19 +283,156 @@ public void testForceHandleMessage_resyncIsRequested() {
...
+                    getSystemErrorHandler().sessionExpiredMessageHandled);
+            assertFalse("Unrecoverable Error Message handling was not " +
+                        "expected when the page is being redirected",
+                    getSystemErrorHandler().unrecoverableErrorHandled);
+            assertEquals(UILifecycle.UIState.TERMINATED,
+                    getUILifecycle().getState());
+        });
+    }
+
+    public void testHandleJSON_sessionExpiredAndUIRunning_sessionExpiredMessageShown() {
```
denis-anisimov, line 354: the name of the test is confusing since it mentioned "session expires" is extra here.

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/LogoutView.java
```diff
@@ -19,16 +19,39 @@
...
             UI.getCurrent().getPage().setLocation(BaseHrefView.class.getName());
             UI.getCurrent().getSession().close();
         });
-        add(button);
+        logoutButton.setId("logout-button");
+        add(logoutButton);
+
+        NativeButton notificationLogoutButton = new NativeButton(
+                "logout-with-notification", ev -> {
+                    VaadinService.getCurrent().setSystemMessagesProvider(
```
denis-anisimov, line 39: Is this needed ?

denis-anisimov (changes requested):

flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
mshabarov: Yes, `UIState` check that the transition goes smoothly from initial state to terminated. It throws in case any jump init -> terminated.

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/LogoutView.java
mshabarov: Yes, it sets session notification message to be shown (it's not shown by default). Then, it actually should not be shown because of `Page::setLocation`. That is exactly I try to test: even if `setSessionExpiredNotificationEnabled(true)`, the page is being redirected with no blinking message.

flow-client/src/test-gwt/java/com/vaadin/client/GwtMessageHandlerTest.java
mshabarov: Why "session expires" is extra? I want this test to go through the code block which handles "session expire" flag and shown the corresponding message. It covers "session expire" case, so this wording is in the test name.

denis-anisimov: Right, my bad.

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/LogoutView.java
denis-anisimov: This action relies on the behavior: the system message provider and the expired notification are handled and set on the client side BEFORE the actions below are executed.

The is expected behavior indeed. But if there a bug on the client side and the commands are handled AFTER then this test will be false positive (in fact it's exactly the same as the previous one).
To avoid such fragile behavior the message provider and session expired settings has to executed before without any assumptions. It should not be one action which relies on the Java order of execution.

denis-anisimov (changes requested):

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/LogoutView.java
mshabarov: Thanks, it's fragile indeed! What if I call `setSessionExpiredNotificationEnabled(true)` separately by another button click, so that it will happens before page redirect and session expiration on the client side?
Moreover, this flag should be reverted at the end of test, so as to not impact another test.

denis-anisimov: Another button click : yes.

Reverting is almost impossible.
You may use `tearDown` in the test of course but the problem is : this needs to be implemented via UI actions which may fail. And as a result revert won't be done.
Moreover failing inside `tearDown`  prevents screenshots being generated. 
That is  already known from Portlets tests.

If there are such "permanent effects" which may influence another tests then it's better to create another servlet with a dedicated URL mapping so that any changes in any configuration will affects only this servlet and don't touch others.
That allows to make a sandbox via the dedicated servlet.

mshabarov: Alright, I've cut off the existing notification enabled part to the dedicated servlet, so now it should solve the problems with the ordering of configuration change and JS command execution, as well as configuration change impact on other tests.

denis-anisimov (approved):

PR #10145, opened 2021-02-27
fix: only inject scripts not already loaded when using an exported wc

Artur-:
This ensures that the same bundle is not loaded twice if injecting on a page where the app itself is already running


flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -287,6 +292,22 @@ protected void writeBootstrapPage(String contentType,
...
+                    // themselves and loading twice
+                    // will always cause custom element conflicts
+                    String src = element.attr("src");
+                    int index = src.indexOf("/VAADIN/");
+                    if (index != -1) {
+                        conditionalFilename = src.substring(index);
+                    }
+                }
+                if (conditionalFilename != null) {
+                    writer.append("if (!hasScript('" + conditionalFilename
```
fluorumlabs, line 308: Use `"` (double quote) character for JavaScript string: single quote (`'`) is a valid character for URL.

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -279,6 +279,11 @@ protected void writeBootstrapPage(String contentType,
         ArrayList<com.vaadin.flow.dom.Element> elementsForShadows = new ArrayList<>();
         try (BufferedWriter writer = new BufferedWriter(
                 new OutputStreamWriter(response.getOutputStream(), UTF_8))) {
+            writer.write("var hasScript = function(src) {\n"
+                    + "  var scriptTags = Array.from(document.head.querySelectorAll('script'));\n"
+                    + "  return !!scriptTags.find(script => script.src.endsWith(src))\n"
```
fluorumlabs, line 284: You probably wanted `return scriptTags.some(script => script.src.endsWith(src))`

fluorumlabs (changes requested):

platosha (commented):
Good idea. There should be tests for this feature though.

haijian-vaadin (changes requested):
Let's add some unit tests for this change before we merge it. The fusion team can make the tests.

joheriks:
Rebased and added unit test.

flow-server/src/test/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandlerTest.java
```diff
@@ -223,6 +223,13 @@ public void writeBootstrapPage_devmodeGizmoIsDisabled()
         handler.synchronizedHandleRequest(session, request, response);
 
         String result = stream.toString(StandardCharsets.UTF_8.name());
+
+        int guardIndex = result.indexOf("if (!hasScript(\"/VAADIN/build/vaadin-export-2222.cache.js\")) {");
```
haijian-vaadin, line 227: I think it would be better to check if the `hasScript` function is defined. Otherwise, if I remove the code for defining the function, the test still passes.

haijian-vaadin (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandlerTest.java
joheriks: Done.

haijian-vaadin (approved):

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -280,6 +280,11 @@ protected void writeBootstrapPage(String contentType,
         ArrayList<com.vaadin.flow.dom.Element> elementsForShadows = new ArrayList<>();
         try (BufferedWriter writer = new BufferedWriter(
                 new OutputStreamWriter(response.getOutputStream(), UTF_8))) {
+            writer.write("var hasScript = function(src) {\n"
```
fluorumlabs, line 283: Just noticed that bootstrap script runs in the global scope, which means, that `hasScript` (and `headElem`) end up in `window`. It might create a headache if some of user-supplied scripts rely on this. IIFE should solve this.

fluorumlabs (changes requested):
Couple of general questions:
1. What about (global) stylesheets? Loading the same stylesheet again will change the order how styles are overriding each other.
2. If we a planning to backport this to 14.x, we should verify that it works in compatibility mode as well (a bit contradictory here; `PATH_PATTERN` in the beginning implies that `WebComponentBootstrapHandler` can also serve `.html`, but content-type set at line 242 suggests otherwise).

fluorumlabs (approved):

haijian-vaadin (approved):

PR #10146, opened 2021-02-28
cache stats.json and asset chunks

### LINKED ISSUE #10116: loadtesting causes stats.json to fail to load sometimes
I ran some loadtests with apache jmeter. The application works fine during debugging and also during the loadtests until some threshold (around 1000 users). Above that threshold during UI Initialization the application fails to load the stats.json file (I guess due too many open files).
Haven't had the time yet to create a minimal example.
### Expected behavior
Opening a ui should cause the server to open the same file each time.
### Actual behavior
It fails to open stats.json
### Versions:
    - Vaadin / Flow version: Vaadin 14.4.7 (production mode, compatibility mode disabled)
    - Java version: 8
    - OS version: debian 9
    - Browser version (if applicable): apache jmeter
    - Application Server (if applicable): tomcat 8
    - IDE (if applicable):

I tried to capture some useful messages from the logs:
```
ERROR [http-nio-8080-exec-71] (FrontendUtils:505) - Cannot get the 'stats.json' from the classpath 'META-INF/VAADIN/config/stats.json'
WARN  [http-nio-8080-exec-71] (UIInitListener:67) - Unexpected exception occured: BootstrapException: Unable to read webpack stats file.
DEBUG [http-nio-8080-exec-71] (UIInitListener:70) - Unexpected exception occured: com.vaadin.flow.server.BootstrapException: Unable to read webpack stats file.
        at com.vaadin.flow.server.BootstrapHandler$BootstrapPageBuilder.setupFrameworkLibraries(BootstrapHandler.java:894)
        at com.vaadin.flow.server.BootstrapHandler$BootstrapPageBuilder.setupDocumentHead(BootstrapHandler.java:771)
        at com.vaadin.flow.server.BootstrapHandler$BootstrapPageBuilder.getBootstrapPage(BootstrapHandler.java:540)
        at com.vaadin.flow.server.BootstrapHandler.synchronizedHandleRequest(BootstrapHandler.java:481)
        at com.vaadin.flow.server.SynchronizedRequestHandler.handleRequest(SynchronizedRequestHandler.java:40)
        at com.vaadin.flow.server.VaadinService.handleRequest(VaadinService.java:1547)
        at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:247)
        at com.vaadin.cdi.CdiVaadinServlet.service(CdiVaadinServlet.java:67)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:741)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:231)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
        at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:52)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
        at org.apache.openejb.server.httpd.EEFilter.doFilter(EEFilter.java:65)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
        at org.apache.logging.log4j.web.Log4jServletFilter.doFilter(Log4jServletFilter.java:71)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
        at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:200)
        at org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:96)
        at org.apache.tomee.catalina.OpenEJBValve.invoke(OpenEJBValve.java:45)
        at org.apache.catalina.authenticator.AuthenticatorBase.invoke(AuthenticatorBase.java:543)
        at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:139)
        at org.apache.tomee.catalina.OpenEJBSecurityListener$RequestCapturer.invoke(OpenEJBSecurityListener.java:97)
        at org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:81)
        at org.apache.catalina.valves.AbstractAccessLogValve.invoke(AbstractAccessLogValve.java:690)
        at org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:87)
        at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:343)
        at org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:615)
        at org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:65)
        at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:818)
        at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1627)
        at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.io.IOException: The stats file from webpack (stats.json) was not found.
The application is running in production mode.Verify that build-frontend task has executed successfully and that stats.json is on the classpath.Or switch application to development mode.
        at com.vaadin.flow.server.BootstrapHandler$BootstrapPageBuilder.appendNpmBundle(BootstrapHandler.java:926)
        at com.vaadin.flow.server.BootstrapHandler$BootstrapPageBuilder.setupFrameworkLibraries(BootstrapHandler.java:892)
        ... 38 more


```

knoobie:
fixes #10116

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
@@ -46,6 +46,7 @@ import com.vaadin.flow.di.ResourceProvider;
 import com.vaadin.flow.function.DeploymentConfiguration;
 import com.vaadin.flow.server.Constants;
 import com.vaadin.flow.server.DevModeHandler;
+import com.vaadin.flow.server.VaadinContext;
 import com.vaadin.flow.server.VaadinRequest;
 import com.vaadin.flow.server.VaadinService;
 import com.vaadin.flow.server.frontend.FallbackChunk.CssImportData;
@@ -190,7 +191,8 @@ public class FrontendUtils {
 
     /**
      * File name of the bootstrap file that is generated in frontend
-     * {@link #GENERATED} folder. The bootstrap file is always executed in a Vaadin app.
+     * {@link #GENERATED} folder. The bootstrap file is always executed in a
+     * Vaadin app.
      */
     public static final String BOOTSTRAP_FILE_NAME = "vaadin.ts";
 
@@ -228,8 +230,8 @@ public class FrontendUtils {
     /**
      * Default generated path for generated frontend files.
      */
-    public static final String DEFAULT_PROJECT_FRONTEND_GENERATED_DIR =
-        DEFAULT_FRONTEND_DIR + GENERATED;
+    public static final String DEFAULT_PROJECT_FRONTEND_GENERATED_DIR = DEFAULT_FRONTEND_DIR
+            + GENERATED;
 
     /**
      * Name of the file that contains all application imports, javascript, theme
@@ -450,10 +452,11 @@ public class FrontendUtils {
      */
     public static String getStatsContent(VaadinService service)
             throws IOException {
+        VaadinContext context = service.getContext();
         DeploymentConfiguration config = service.getDeploymentConfiguration();
 
         if (config.isProductionMode()) {
-            return getOrCreateStatsCache(service).statsJson;
+            return getOrCreateStatsCache(context, config).statsJson;
         }
 
         InputStream content = null;
@@ -462,26 +465,26 @@ public class FrontendUtils {
         }
 
         if (config.isStatsExternal()) {
-            StatsCache externalCache = getStatsFromExternalUrl(
+            // This isn't cached because production mode isn't enabled
+            StatsCache externalStats = getStatsFromExternalUrl(
                     config.getExternalStatsUrl(), null);
-            if (externalCache != null) {
-                return externalCache.statsJson;
+            if (externalStats != null) {
+                return externalStats.statsJson;
             }
         }
 
         if (content == null) {
-            content = getStatsFromClassPath(service);
+            content = getStatsFromClassPath(context, config);
         }
         return content != null
                 ? IOUtils.toString(content, StandardCharsets.UTF_8)
                 : null;
     }
 
-    private static StatsCache getOrCreateStatsCache(VaadinService service)
-            throws IOException {
-        DeploymentConfiguration config = service.getDeploymentConfiguration();
-        StatsCache currentCache = service.getContext()
-                .getAttribute(StatsCache.class);
+    // package private for direct testing
+    static StatsCache getOrCreateStatsCache(VaadinContext context,
+            DeploymentConfiguration config) throws IOException {
+        StatsCache currentCache = context.getAttribute(StatsCache.class);
 
         // External stats file have higher priority because
         // we have to check the lastModified header to
@@ -491,7 +494,7 @@ public class FrontendUtils {
                     config.getExternalStatsUrl(), currentCache);
             // Refresh current cache if external cache could be created
             if (externalCache != null) {
-                service.getContext().setAttribute(externalCache);
+                context.setAttribute(externalCache);
                 currentCache = externalCache;
             }
         }
@@ -500,13 +503,13 @@ public class FrontendUtils {
             return currentCache;
         }
 
-        InputStream statsJson = getStatsFromClassPath(service);
+        InputStream statsJson = getStatsFromClassPath(context, config);
         if (statsJson == null) {
             return StatsCache.NOT_CACHEABLE;
         }
 
         currentCache = new StatsCache(statsJson);
-        service.getContext().setAttribute(currentCache);
+        context.setAttribute(currentCache);
         return currentCache;
     }
 
@@ -662,14 +665,15 @@ public class FrontendUtils {
         return host;
     }
 
-    private static InputStream getStatsFromClassPath(VaadinService service) {
-        String stats = service.getDeploymentConfiguration()
+    private static InputStream getStatsFromClassPath(VaadinContext context,
+            DeploymentConfiguration config) {
+        String stats = config
                 .getStringProperty(SERVLET_PARAMETER_STATISTICS_JSON,
                         VAADIN_SERVLET_RESOURCES + STATISTICS_JSON_DEFAULT)
                 // Remove absolute
                 .replaceFirst("^/", "");
-        ResourceProvider resourceProvider = service.getContext()
-                .getAttribute(Lookup.class).lookup(ResourceProvider.class);
+        ResourceProvider resourceProvider = context.getAttribute(Lookup.class)
+                .lookup(ResourceProvider.class);
         URL statsUrl = resourceProvider.getApplicationResource(stats);
         InputStream stream = null;
         try {
@@ -690,7 +694,8 @@ public class FrontendUtils {
     private static InputStream getFileFromWebpack(String filePath)
             throws IOException {
         DevModeHandler handler = DevModeHandler.getDevModeHandler();
-        return handler.prepareConnection("/" + filePath, "GET").getInputStream();
+        return handler.prepareConnection("/" + filePath, "GET")
+                .getInputStream();
     }
 
     /**
@@ -707,13 +712,14 @@ public class FrontendUtils {
      */
     public static String getStatsAssetsByChunkName(VaadinService service)
             throws IOException {
+        VaadinContext context = service.getContext();
         DeploymentConfiguration config = service.getDeploymentConfiguration();
 
         if (config.isProductionMode()) {
-            return getOrCreateStatsCache(service).assetChunks;
+            return getOrCreateStatsCache(context, config).assetChunks;
         }
 
-        if (!config.isProductionMode() && config.enableDevServer()) {
+        if (config.enableDevServer()) {
             return streamToString(getResourceFromWebpack("/assetsByChunkName",
                     "getting assets by chunk name."));
         }
@@ -727,7 +733,7 @@ public class FrontendUtils {
             }
         }
 
-        InputStream resourceAsStream = getStatsFromClassPath(service);
+        InputStream resourceAsStream = getStatsFromClassPath(context, config);
         if (resourceAsStream == null) {
             return null;
         }
@@ -1094,8 +1100,8 @@ public class FrontendUtils {
     }
 
     /**
-     * Parse "manifest.json" file contents obtained from webpack and extract
-     * the list of request paths to handle as static resources.
+     * Parse "manifest.json" file contents obtained from webpack and extract the
+     * list of request paths to handle as static resources.
      *
      * @param manifestJson
      *            "manifest.json" file contents
@@ -1141,18 +1147,25 @@ public class FrontendUtils {
         protected final String assetChunks;
         protected final LocalDateTime lastModified;
 
-        StatsCache() {
+        private StatsCache() {
             this.statsJson = null;
             this.assetChunks = null;
             this.lastModified = null;
         }
 
-        public StatsCache(InputStream statsJson) throws IOException {
+        /**
+         * Create a new container for internal stats.json caching.
+         *
+         * @param statsJson
+         *            the gotten stats.json as a {@link InputStream}, not
+         *            {@code null}.
+         */
+        StatsCache(InputStream statsJson) throws IOException {
             this(statsJson, null);
         }
 
         /**
-         * Create a new container for stats.json caching.
+         * Create a new container for external stats.json caching.
          *
          * @param statsJson
          *            the gotten stats.json as a {@link InputStream}, not
@@ -1162,7 +1175,7 @@ public class FrontendUtils {
          *            stats.json in RFC-1123 date-time format, such as 'Tue, 3
          *            Jun 2008 11:05:30 GMT'
          */
-        public StatsCache(InputStream statsJson, String lastModified)
+        StatsCache(InputStream statsJson, String lastModified)
                 throws IOException {
             Objects.requireNonNull(statsJson,
                     "stats.json stream can't be null");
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -338,6 +338,9 @@
 
     private static String operatingSystem = null;
 
+    private static String cachedStatsJson = null;
```
mshabarov, line 341: I'm not sure the util class (`FrontendUtils`) is a proper place to store the cached data.
There is already a solution for external stats.json caching here https://github.com/vaadin/flow/blob/eacee0f42b8b6b111d2858c0fe6998f9f996a755/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java#L463

What if we merge these two cases and store stats.json in context for both cases (external or classpath)?

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -338,6 +338,9 @@
...
+    private static String cachedAssetChunks = null;
```
mshabarov, line 342: Same goes for cached asset chunks. What if you add one more field to `Stats` container to store there a result of `getStatsAssetsByChunkName`?

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
mshabarov, line 341: There is one more place stated in the original ticket stack trace https://github.com/vaadin/flow/issues/10116, where the exception is thrown:

https://github.com/vaadin/flow/blob/eacee0f42b8b6b111d2858c0fe6998f9f996a755/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java#L892

I believe we should add a caching there as well. Context and configuration are available in this method, so we can check agains production mode and already cached object.

mshabarov (changes requested):
Thank you @knoobie for a contribution!
Your solution with caching looks good. I would propose refactor the code a bit in order to store the cached objects in a proper place.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
knoobie: Thanks @mshabarov! I didn't know that class :) Gonna refactor the code (not sure if today or in the next days) but it sounds totally reasonable to store it in a proper class/storage.

knoobie: I think this is already covered because `appendNpmBundle`:

https://github.com/vaadin/flow/blob/eacee0f42b8b6b111d2858c0fe6998f9f996a755/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java#L907-L910

already calls the cached chunk in the first line :)

knoobie: Okay.. I did some brainstorming and code digging. The biggest problem is `getStatsFromExternalUrl` because it populates the current cache / `Stats` within the context. Funny thing.. this is only done if the header `last-modified` of the `HttpConnection` has a value and it's older than the current stored value. 

I was thinking about removing the cache code from `getStatsFromExternalUrl` because it doesn't really make sense with the name but I couldn't satisfy the `Stats` constructor because of the missing `lastModified` information later on. (`getStatsFromExternalUrl` only returns an `InputStream`) .

Additionally.. do I have to check for the `lastModified` before I return the cache? with what value do I check it? Do I really have to create a `HttpConnection` everytime to check? That doesn't really sounds reasonable. This should only be done if `isExternalStatsFile=true` - meaning `lastModified` gets optional?

How about `assetChunks`? Does it inherit the `lastModified` from the `statsJson` or is can this be ignored? 

If you don't mind I would propose the following: 
 - Combined method to create and access the cache for `assetChunks` and `statsJson`. 
 - This method is only called in productionMode and returns the requested string 
 - If `isExternalStatsFile=true` I ALWAYS open a connection to get a `lastModified` timestamp to update the cache before return? or should I create a `LocalFileCache` and leave the old implementation alone? That would make a lot of things easier..

mshabarov: Sorry, I don't see how it's covered.
`appendNpmBundle` calls `getStatsAssetsByChunkName` in which the `stats.json` is being fetched either from local file or from external source.  
In case of a local file, the `stats.json` is never cached currently.

The only place where it is cached is here, for external source:
https://github.com/vaadin/flow/blob/eacee0f42b8b6b111d2858c0fe6998f9f996a755/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java#L463

I wanted to emphasise that improving the performance by caching the `stats.json`, when loading from a file, should be done for both cases: during initial bootstrap in `BootstrapHandler` and in other cases when `stats.json` is needed and `FrontendUtils::getStatsContent` is being invoked.

knoobie: In my current implementation is no call to `stats.json` if `productionMode = true && cachedAssetChunks` - yes that's faulty in other ways. See the comment from above with the wall of text / looking for ways around the whole external file cache situation.

mshabarov: > this is only done if the header last-modified of the HttpConnection has a value and it's older than the current stored value.

Why older? The current code looks fine: it updates the cached value if the object is not cached yet, or the source object has been modified (cached last modified timestamp is before the one obtained from external source). Another thing I have a doubts is that why is the timestamp is recalculated in getter method:
```java
        public LocalDateTime getLastModified() {
            return ZonedDateTime
                    .parse(lastModified, DateTimeFormatter.RFC_1123_DATE_TIME)
                    .toLocalDateTime();
        }
```
Why not store the `LocalDateTime` instead of `String`, but this is an another question.

> I was thinking about removing the cache code from getStatsFromExternalUrl because it doesn't really make sense with the name but I couldn't satisfy the Stats constructor because of the missing lastModified information later on. (getStatsFromExternalUrl only returns an InputStream) .

`getStatsFromExternalUrl` is a `private` methods, so we can change the return type so as to include the `lastModified` value there and extract the caching logic to a separate method and leave only the code related to fetching data via `HttpConnection`. The caching logic then will 'wrap' the actual fetch (from classpath or from external source) and populate the context if needed.

> Additionally.. do I have to check for the lastModified before I return the cache? with what value do I check it? Do I really have to create a HttpConnection everytime to check? That doesn't really sounds reasonable. This should only be done if isExternalStatsFile=true - meaning lastModified gets optional?

It's a good question! I would say, normally the `stats.json` should not be changed (?) if stored in classpath (in current WAR or another WAR on same server). But what happens if some hot redeploy mechanism is used (OSGi for instance), could it be changed then? Not sure about that. And even worse thing that if the `stats.json` is in another bundle, and it hasn't been loaded yet at the point when it's requested. Theoretically, it's possible. But that's would be an issues of a client code/environment. Not sure we should handle such a cases in the Flow. Long story short, I would ignore the `lastModified` for loading from classpath.

> How about assetChunks? Does it inherit the lastModified from the statsJson or is can this be ignored?

I believe it's just inherit the lastModified from stat.json. So if the stats.json is up-to-date, then no need to update the cached assetChunks.

> Combined method to create and access the cache for assetChunks and statsJson.
This method is only called in productionMode and returns the requested string
If isExternalStatsFile=true I ALWAYS open a connection to get a lastModified timestamp to update the cache before return? or should I create a LocalFileCache and leave the old implementation alone? That would make a lot of things easier..

Looks good, but I'm against to use `LocalFileCache` in Flow, as it would bring a new dependency, which may cause a serious unexpected conflicts/bug on the client project side.

knoobie: Your suggestions sound reasonable and align to my understanding. Gonna try to refactor the implementation (wanted to get approval for deeper digging/changes). 

About `LocalFileCache` - that was just a name ;) I wanted to create a inner private class `LocalFileCache` that works like `Stats` just for the internal files to reduce the gazillion creations of `InputStream` and file access for cacheable ressources (asset chunks and json) without touching the current `Stats` class for `external` calls - this would only be called if `productionMode = true && externalStatsFile == false`.

mshabarov: > About LocalFileCache - that was just a name ;) I wanted to create a inner private class LocalFileCache that works like Stats just for the internal files to reduce the gazillion creations of InputStream and file access for cacheable ressources (asset chunks and json) without touching the current Stats class for external calls - this would only be called if productionMode = true && externalStatsFile == false.

Sorry for confusing :) Okay, sounds good then.

> Edit: About getStatsFromExternalUrl - it gave me a headache because the Stats cache is only called and populated if the connection has a lastModified header.

Well, I don't see any contradictions - if the header hasn't been obtained, then no conclusions about cached resource 'version' can be made, thus, the code simply fetches it from connection. What if you use the cached resource timestamp, if not header obtained?

knoobie: Thanks for the clarification and your input! Gonna rework on this in the weekend by creating a separate `LocalFileCache` and if that seems good enough, we can try to integrate the whole `Stats` logic for external files there as well. Step by step.

mshabarov: Ah, sorry! You're right! Indeed, `getStatsAssetsByChunkName` contains changes. Ignore this comment then.

knoobie:
This is a complete rework based on the discussion with @mshabarov. The old `FrontendUtilsTest` are passing. Gonna see what the flow pipeline says to this change and create unit tests afterwards.

knoobie:
@mshabarov now it's ready to review. I've implemented the cache for `local stats.json` and `external stats.json` when `productionMode = true` otherwise there is no cache - removing the old `Stats` and cache for `external stats.json`. This change could result in a little overhead in development mode if somebody uses `external stats.json` and disables webpack.. (who does said? Nobody I think)

Additionally this implementation guards against your described problem with OSGI because it returns `StatsCache.NOT_CACHEABLE` (null) if the `stats.json` could not be found - resulting in retrying on the next execution. :) 

Open question: Now, I know why the `LocalDateTime` wasn't stored in the `Stats` - sonarqube complains. Do you want me to change it to `String` or add `transient`? I would go with `transient` because there isn't much information lost. The cache is gonna recreate itself when there is no `lastModified` and `externalStats=true`. 

And sorry for the unrelated code changes in the file.. I used `mvn formatter:format` in the end to comply with your rules and some older methods weren't aligned.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -587,31 +627,36 @@ private static InputStream getStatsFromExternalUrl(String externalStatsUrl,
...
+            return createCacheFromHttpURLConnection(currentCache, connection);
         } catch (IOException e) {
             getLogger().error("Failed to retrieve stats.json from the url {}.",
                     url, e);
         }
         return null;
     }
 
+    // package private for direct unit testing
+    static StatsCache createCacheFromHttpURLConnection(StatsCache currentCache,
```
mshabarov, line 640: This method is used only inside another `private` method.
If it was a public API, it would be fine to test it separately. But, as its effectively `private`, It looks a bit weird to test it like this.
I'd expect it is tested as a part of some public API test.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
knoobie: Sadly the other private method (caller) wasn't tested before and it seemed more straight forward to explicit test this part because of the additional logic. External stats were not tested at all (execution of `HttpUrlConnection` and so on) - testing this new logic seemed more important because the old logic didn't change.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -452,26 +453,66 @@ public static ProcessBuilder createProcessBuilder(List<String> command) {
...
         if (content == null) {
-            content = getStatsFromClassPath(service);
+            content = getStatsFromClassPath(context, config);
         }
         return content != null
                 ? IOUtils.toString(content, StandardCharsets.UTF_8)
                 : null;
     }
 
+    private static StatsCache getOrCreateStatsCache(VaadinContext context,
```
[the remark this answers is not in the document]
knoobie: I can try to create a version that only modifies `getStatsFromClassPath` to reduce the impact of the change. I would propose to do this in another branch, I don't want to lose all this with a force push. ;)

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -452,26 +453,66 @@ public static ProcessBuilder createProcessBuilder(List<String> command) {
...
+
+        InputStream content = null;
+        if (config.enableDevServer()) {
             content = getStatsFromWebpack();
         }
 
         if (config.isStatsExternal()) {
-            content = getStatsFromExternalUrl(config.getExternalStatsUrl(),
-                    service.getContext());
+            // This isn't cached because production mode isn't enabled
```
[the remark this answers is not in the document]
knoobie: This would be resolved by reducing the change to `getStatsFromClassPath` as well.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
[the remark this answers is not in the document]
mshabarov: Yes, let's proceed this way.

knoobie:
Closes this and create a new pr with less changes in the next 1-2 days.

PR #10151, opened 2021-03-01
Use the same bundle for exported wcs and the application

Artur-:
This allows embedding exported wcs inside the same application that exports them, e.g. a Fusion view embedding a Flow exported wc.

In case you have only exported wcs and no application, it should make no difference to the users.
In case you have only an application and no exported wcs, it should make no difference to the users.

In case you have an application and exported wcs, the users will load a slightly larger bundle. This can be fixed by putting the exported wcs in a separate project.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -84,7 +84,6 @@ import elemental.json.JsonValue;
 import elemental.json.impl.JsonUtil;
 
 import static com.vaadin.flow.server.Constants.VAADIN_MAPPING;
-import static com.vaadin.flow.server.frontend.FrontendUtils.EXPORT_CHUNK;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
@@ -802,7 +801,6 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
             // include all chunks but the one used for exported
             // components.
             return Arrays.stream(chunks.keys())
-                    .filter(s -> !EXPORT_CHUNK.equals(s))
                     .collect(Collectors.toList());
         }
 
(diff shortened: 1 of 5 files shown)


pleku:
Why is this needed @Artur- ?

Artur-:
It allows you to export a Flow component in a hybrid project and then use it in a Fusion view. For instance this way you can use collaboration engine (Flow only) in a Fusion project by only creating the collaboration engine parts in Flow and the rest in Fusion.

It might be that this should be opt-in or have an opt-out though

Artur-:
Based on tests by @eriklumme this might be broken when you export a component that uses a connector because the connector is loaded before the Flow client engine

pleku:
> this might be broken when you export a component that uses a connector because the connector is loaded before the Flow client engine

This PR might be broken or the test was broken ? 

There is no IT for this and I would rather not take this change in without one. Sure it can be tested manually by creating a project, but that takes a lot of effort, which is away from something else. And I'm not sure if this is still important or not - there is no ticket or feedback that this would be needed.

I'd suggest that either
a) a new IT module is added to flow-tests/test-embedding/ which tests the use case - though apparently it would need a dummy component which uses a connector (to get it to fail?)
b) create an issue on to track this use case and link the PR there but close it for now, to then work on this if/when someone actually needs it

pleku (changes requested):
> broken when you export a component that uses a connector because the connector is loaded before the Flow client engine

This PR is broken or the test app it broken now ?

I would not take this PR in without an IT in flow-tests/test-embedding/ (might need a new module). But then again - fusion apps need to use the spring add-on so it might be even impossible at the moment to create the IT in this project, which means that it would have to go to the spring add-on project instead. This is a lot of work to do, but so is validating the PR manually and this could be quite prone to regressions. So I suggest to either
a) create the IT, which apparently should also test with a component that uses a connector (?)
b) create an issue on this and wait until someone actually needs the support. The PR can be linked there but then closed to wait for becoming a priority

eriklumme:
Created an issue for my use case, https://github.com/vaadin/flow/issues/10913.

pleku:
As this is not a priority at this point - closing the PR for now. To be continued if this would be needed

PR #10162, opened 2021-03-02
10109 remove only added routes

denis-anisimov:


flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
```diff
@@ -68,9 +75,46 @@ public void initialize(Set<Class<?>> classSet, VaadinContext context)
...
+        routeConfiguration
+                .update(() -> setAnnotatedRoutes(routeConfiguration, routes));
+    }
+
     private void setAnnotatedRoutes(RouteConfiguration routeConfiguration,
             Set<Class<? extends Component>> routes) {
-        routeConfiguration.getHandledRegistry().clean();
         for (Class<? extends Component> navigationTarget : routes) {
             try {
                 routeConfiguration.setAnnotatedRoute(navigationTarget);
```
caalador, line 120: Here the cache/stored list could be populated after setting a route as the data can be gotten with `routeConfiguration.getTemplate(navigationTarget)`

flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
```diff
@@ -38,27 +41,31 @@
 public class RouteRegistryInitializer extends AbstractRouteRegistryInitializer
         implements VaadinServletContextStartupInitializer {
 
+    private static class StaticRoutesApplicationRouteRegistry
```
caalador, line 44: The class name indicates something that should always say the same, but it is used for removing some set of previously registered routes that excludes dynamic additions.
`PreviousRoutes` or `PreviousRouteInitialization` also it could be either a `List<String>` or `Set<String>` as only the template string is used.

flow-server/src/test/java/com/vaadin/flow/server/startup/RouteRegistryInitializerTest.java
```diff
@@ -1403,6 +1411,151 @@ public void registerClassesWithSameRoute_unrelatedClass_class_throws()
...
+
+        RouteData route1 = routes.get(0);
+        RouteData route2 = routes.get(1);
+
+        if (route1.getTemplate().equals("foo")) {
+            Assert.assertEquals("manual-route", route2.getTemplate());
+        } else {
+            Assert.assertEquals("manual-route", route1.getTemplate());
+            Assert.assertEquals("foo", route2.getTemplate());
+        }
```
caalador, line 1485: Could also instead check 
```
Assert.assertTrue(registry.getTemplate(BaseRouteTarget.class).isPresent());
Assert.assertTrue(registry.getNavigationTarget(PathUtil.getPath("manual-route", Collections.emptyList())).isPresent());
```
or
```
RouteConfiguration configuration = RouteConfiguration.forRegistry(registry);
Assert.assertTrue(configuration.isRouteRegistered(BaseRouteTarget.class));
Assert.assertTrue(configuration.isPathAvailable("manual-route"));
```

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/startup/ClassLoaderAwareServletContainerInitializer.java
```diff
@@ -136,23 +136,23 @@ default boolean requiresLookup() {
...
      *            {@link javax.servlet.annotation.HandlesTypes HandlesTypes}
      *            annotation, or <tt>null</tt> if there are no matches, or this
      *            <tt>ServletContainerInitializer</tt> has not been annotated
      *            with <tt>HandlesTypes</tt>
      *
-     * @param context
+     * @param ctx
      *            the <tt>ServletContext</tt> of the web application that is
-     *            being started and in which the classes contained in <tt>classSet</tt>
+     *            being started and in which the classes contained in <tt>c</tt>
```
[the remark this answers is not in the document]
denis-anisimov: reverted back the whole class since there is no need in these changes anymore.

flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
denis-anisimov: I don't like the class name as well.
But you are suggesting also quite bad name: your  name is based on the usage. It's quite weird to name class based on the single usage: doesn't matter how the class is used. It matters what the class represents: the same class can be used differently by many many pieces of code (even though this is quite specific case since the class is initially done private and intended to be used only in one place).

This class (ignoring its usage) contains static routes (unmutable). Comparing to the `ApplicationRouteRegistry` which is dynamic registry.
That was the initial idea for the name.
I will try to figure our some better name

flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
denis-anisimov: The API is confusing: `routeConfiguration.getTemplate(navigationTarget)` returns `Optional<String>`.
Perhaps it return the primary route path. Because I don't see how to get aliases. The registry API with `RouteData` gives a better API.

flow-server/src/test/java/com/vaadin/flow/server/startup/RouteRegistryInitializerTest.java
denis-anisimov: Done.

flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
caalador: I could see a clear use case for adding a `getRouteData(Class targetClass)` in the registry as it's anyway populated on demand and not reused. The build already exists in `AbstractRouteRegistry::populateRegisteredRoutes` would just need to create the public method.

denis-anisimov: Well, that's fine. But this is about `RouteConfiguration`  API which is out of scope of this PR.
Regardless of this change a registry copy is still preferable for me here since:
- it gives a copy of everything that has been stored
- the way of data collection is the same code which is used for the main registry
- the above observations allows to avoid any fragile code/assumptions which may be broken with some changes in the registry impl
- it's extremely easy to correct the code in case there will be need to remove more data stored in the registry (if there will be any need in this): comparing to any custom storage which will require changes
- I still need a class to store the data inside the context instance (and `List` class is not suitable: some wrapper is needed anyway): there is already a simple registry subclass with an empty impl.

So overall the current approach requires less efforts to maintain the code.

flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
```diff
@@ -38,27 +41,31 @@
 public class RouteRegistryInitializer extends AbstractRouteRegistryInitializer
         implements VaadinServletContextStartupInitializer {
 
+    private static class PreviouslyStoredRoutesImmutableRegistry
```
caalador, line 44: As long as it's a RouteRegistry it's not immutable. Else I'm fine with the naming.

denis-anisimov: It's effectively immutable: once created it's never changed.

But I'm going to the same direction of usage here.....
Let it be `PreviouslyStoredRoutesRegistry`

caalador (approved):

PR #10185, opened 2021-03-03
chore: add isInternalRequest for use in VaadinSpringSecurity helper

joheriks:


flow-server/src/main/java/com/vaadin/flow/internal/SecurityHelper.java
```diff
@@ -43,6 +47,24 @@
...
     };
 
+    /**
+     * Determines whether the request is internal to the Vaadin framework.
+     *
+     * @param request
+     *            a servlet request
+     * @return true iff the request is internal
+     */
+    public static boolean isInternalRequest(HttpServletRequest request) {
```
Artur-, line 57: What does "internal" mean? It is always safe? It is only for authenticated users?

flow-server/src/main/java/com/vaadin/flow/internal/SecurityHelper.java
```diff
@@ -43,6 +47,24 @@
...
+        String requestType = request
+                .getParameter(ApplicationConstants.REQUEST_TYPE_PARAMETER);
+        String referer = request.getHeader("referer");
```
Artur-, line 60: Is this a security related feature? You cannot rely on headers for that

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/internal/SecurityHelper.java
joheriks: Yes, realized this is not a viable solution.

flow-server/src/main/java/com/vaadin/flow/internal/SecurityHelper.java
joheriks: The `isInternalRequest` logic (from Bakery) is there used for two purposes: 1) in `CustomRequestCache` to check whether to save a request for redirect (which should be fine), but also 2) to grant `permitAll` to all internal requests (in which case the `Referer` check is a direct security hole):
https://github.com/vaadin/bakery-app-starter-flow-spring/blob/master/src/main/java/com/vaadin/starter/bakery/app/security/SecurityConfiguration.java#L83-L107
The helper would need to be specified to be used only for the first case.

joheriks: And as a corollary, it's not really a security feature and so should probably be moved to a more appropriate class.

joheriks:
Closing this as not clear whether `isInternalRequest` logic is useful enough as a helper method. Fix in bakery:
https://github.com/vaadin/bakery-app-starter-flow-spring/pull/1160

Issue #10182, opened 2021-03-03
[Vaadin 20] Move Lumo and Material theme integrations to Flow components


web-padawan:
### Description

We are going to move all the web components to the [monorepo](https://github.com/vaadin/vaadin-web-components) for Vaadin 20, including themes.

So we will have unified version for every web component as well as `vaadin-lumo-styles` and `vaadin-material-styles`.
This means every time we make a new platform release e.g. `20.0.1`, we bump version also for Lumo and Material.

### Problem

Currently, Flow theme integrations contains `@NpmPackage` annotation with fixed version ([Lumo](https://github.com/vaadin/flow/blob/fd3fe381e2807e97f669fdafbd2b44e9883d0cc3/flow-theme-integrations/lumo/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java#L34), [Material](https://github.com/vaadin/flow/blob/fd3fe381e2807e97f669fdafbd2b44e9883d0cc3/flow-theme-integrations/material/src/main/java/com/vaadin/flow/theme/material/Material.java#L33)).

When moving to a new versioning, we need to keep these npm package versions in sync across the platform.
Otherwise we will get recurring issues like #9762 (that was fixed by bumping version at #9763).

### Proposed solution

Move theme integrations to  [vaadin-flow-components](https://github.com/vaadin/vaadin-flow-components) monorepo.
This would allow us to bump `@NpmPackage` for all the web components, including themes, at the same time.

The releasing workflow would look as follows:

1. Release web components (+ themes)
2. Bump versions in Flow components
3. Release Flow components

This means that Flow release will no longer be coupled with web components releases (and can be done any time).

### Alternative solutions

- Keep theme integrations in Flow, create automatic PR to bump @NpmPackage every time we release web components.
  - pros: less effort
  - cons: makes Flow release part of the components releasing workflow, requires to have more releases
- Relax `@NpmPackage` in Flow to not use fixed version, e.g. use `~20.0.0` instead of `20.0.1`
  - pros: less effort
  - cons: sounds more riskier
- Align Flow version with the platform
  - pros: versions are consistent
  - cons: : makes Flow release part of the components releasing workflow, requires to have more releases

web-padawan:
As mentioned by @manolo there are two modules under flow-theme-integrations (lumo and material) but there are dependencies in flow-component-demo-helpers, flow-maven-plugin, flow-bom, and flow-tests.

[...]

PR #10193, opened 2021-03-04
chore: remove no longer used test class

web-padawan:
See https://github.com/vaadin/flow/issues/10182#issuecomment-790485350


caalador (approved):

web-padawan:
Closing. The file is needed so we have to go with a different approach.

Issue #10148, opened 2021-02-28
[Vaadin 19] After install PWA app I can't open this app via clicking on icon (from windows pulpit)


darmro:
### Description of the bug 
I installed Vaadin 19 PWA demo app by clicking in Chrome the install button (latest Chrome in Win 10). The app icon appeared on the Windows pulpit. **But after clicking this icon on Windows pulpit no app is started but I can only see this**:
![image](https://user-images.githubusercontent.com/56316289/109436942-914ab080-7a22-11eb-83f9-6945da6ebfdd.png)
**This is of course the chrome_proxy window but in its content there is only text (from manifest?)**.
Of course just after installing the application chrome opened a new proxy window and application was shown in this window. You have to close this and open installed app again by clicking on the pulpit icon.

**When I do the same with Vaadin 14 demo app everything is ok**. After clicking Vaadin 14 PWA application icon (on the Windows pulpit) the app is started and shown in the chrome_proxy window. Sic!  
The same app but different Vaadin version (and PWA mechanism).

**I must add that this is not the only Vaadin 19 application that has behaved in this way**. 
I'm creatin real production app (Java + Spring) which was on Vaadin 14. But I upgraded my app to Vaadin 19 version, I configured PWA (exactly as described in documentation- in the AppShellConfigurator), run the app and installed it as PWA application in Chrome. I was surprised that clickin on icon from pulpit doesn't work. That is why I did a test with yours demo app- the same problem. So it has to bee something with a new PWA mechanism in Vaadin (maybe :).

### Minimal reproducible example
I took from "Vaadin / Start building" page the demo/starting app named "Vaadin 18 – latest features" (in builder I chose latest Vaadin 19 version). I imported this project into my Eclipse workspace (maven import). I changed packaging to war in pom. Then run this app (tomcat 9, thats why war). Opened this app on Chrome (on Windows), installed it. Then closed Chrome and finaly clicked on the app icon on pulpit.
Source here: 
[my-app.zip](https://github.com/vaadin/flow/files/6058288/my-app.zip)

### Expected behavior
Why there's no app in chrome_proxy window but only text? What I'm doing wrong? 
I am in a difficult position because I am releasing the application soon and I've promised that it will be a PWA :|

### Versions:
    - Vaadin / Flow version: Vaadin 19.0.0.beta4
    - Java version: 8
    - OS version: Windows 10
    - Browser version (if applicable): Chrome 88.
    - Application Server (if applicable): Tomcat 9
[...]

PR #10194, opened 2021-03-04
fix: set default PWA start_url as "." instead of ""

haijian-vaadin:
Fixes https://github.com/vaadin/flow/issues/10148

No ITs, since the install workflow is to be tested manually. One can follow [this article](https://support.google.com/chrome/answer/9658361?co=GENIE.Platform%3DDesktop&hl=en) to install an PWA app  to Chrome. Note the icon is a bit different compare to the article, instead of a + icon, it's a Screenshot 2021-03-04 at 15 33 31

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/PwaConfiguration.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/PwaConfiguration.java
@@ -57,7 +57,7 @@ public class PwaConfiguration implements Serializable {
     public PwaConfiguration() {
         this(false, DEFAULT_NAME, "Flow PWA", "", DEFAULT_BACKGROUND_COLOR,
                 DEFAULT_THEME_COLOR, DEFAULT_ICON, DEFAULT_PATH,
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/PwaConfiguration.java
```diff
@@ -36,6 +36,7 @@
     public static final String DEFAULT_BACKGROUND_COLOR = "#f2f2f2";
     public static final String DEFAULT_DISPLAY = "standalone";
     public static final String DEFAULT_OFFLINE_PATH = "offline.html";
+    private static final String DEFAULT_START_URL = ".";
```
platosha, line 39: Should we align `com.vaadin.flow.server.PWA#startPath` also?

haijian-vaadin: good catch, done

platosha (approved):

PR #10220, opened 2021-03-08
refactor!: remove theme modules from Flow

### LINKED ISSUE #10182: [Vaadin 20] Move Lumo and Material theme integrations to Flow components

caalador:
Remove the Lumo and Material theme
modules from Flow project to have them
in the components project instead.

Closes #10182


web-padawan:
Looks like reusable theme tests need to be updated as well:

```
[Step 5/11] Failed to execute goal on project flow-test-application-theme-reusable: Could not resolve dependencies for project com.vaadin:flow-test-application-theme-reusable:war:1111.5.0-SNAPSHOT: Could not find artifact com.vaadin:vaadin-lumo-theme:jar:1111.5.0-SNAPSHOT in vaadin-prereleases (https://maven.vaadin.com/vaadin-prereleases/)
```

flow-tests/test-ccdm/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java
```diff
@@ -27,11 +27,11 @@
 import com.vaadin.flow.theme.AbstractTheme;
 
 /**
- * Lumo component theme class implementation.
+ * Lumo component theme class implementation. Used for testing theming features.
```
pleku, line 30: Should this be in some reusable testing module and shared from there ? Seems weird that it is duplicated here and in `test-common`

pleku: Or is this even needed for ccdm tests ?

pleku (commented):
Couple of questions, probably to @web-padawan :
1) are all the theme ITs (from here) moved to the components project too ?
2) is the plan to take the changes to 14.x version of the components too ? I was just wondering how that will work from versioning perspective ...

pleku:
The test failure in the build doesn't seem related to this issue nor really anything else either

web-padawan:
@pleku Thank you for taking a look.

1. All the theme ITs will be moved to the Flow components monorepo. There is now a PR for moving sources:  https://github.com/vaadin/vaadin-flow-components/pull/706, @manolo mentioned that ITs will be moved separately.
2. We are not planning to make this change for 14.x version of components. Vaadin 14 will not use the new monorepo.

flow-tests/test-ccdm/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java
caalador: The CCDM tests use the application theme and that requires Lumo.

Using the test-common in ccdm module makes some routes dupliacted and thus fails the module at startup.

pleku: Yeah then the `common` is more like `flow-common` and not anything that both f&f would share but that is out of scope for this

pleku:
> 2. We are not planning to make this change for 14.x version of components. Vaadin 14 will not use the new monorepo.

To clarify, @web-padawan you're referring to the web components ? As the flow integrations are in https://github.com/vaadin/vaadin-flow-components/tree/14.5 already but there would be a need to make changes to the web components too and that is not done ? 

Anyway, I think it is fine. Then there is no problem with the versions. And the pace of updates for v14 versions will probably wind down during the H2 of this year so there should not probably be any trouble for backporting things (nothing or very little to backport).

web-padawan:
Yes, I was talking about web components monorepo (we will not use that for Vaadin 14). 

The main reason for moving theme integrations to `vaadin-flow-components` is to make it easier for us to release things.
In theory nothing should prevent us from moving Flow theme integrations also for 14.x but I don't see a need for this.

pleku (commented):
One comment on existing discussion https://github.com/vaadin/flow/pull/10220#discussion_r592326995

flow-tests/test-common/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java
```diff
@@ -0,0 +1,83 @@
...
+import com.vaadin.flow.component.dependency.JsModule;
+import com.vaadin.flow.component.dependency.NpmPackage;
+import com.vaadin.flow.theme.AbstractTheme;
+
+/**
+ * Lumo component theme class implementation. Used for testing theming features.
+ *
+ * @since 1.0
+ */
+@NpmPackage(value = "@vaadin/vaadin-lumo-styles", version = "1.6.0")
```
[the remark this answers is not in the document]
pleku: While this is only for testing purposes and we should (?) detect any breaking anyway in components repo or at platform level, would it still make sense to use something like `^1.6.0` or `x.x` orwhatnot so that we might catch regressions early ? 

Prereleases cannot be used I think without specifying the version explicitly, so it has to be a released version anyway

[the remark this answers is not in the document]
caalador: Well it should match platform version that the build is targeting so should it then be only patch releases ~1.6.0 or ~20.0.0 if that's the versioning in the future, but it would then always need to be updated with the platform...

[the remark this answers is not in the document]
pleku: @manolo so it would be nice if this is updated once a new major version is bumped for the themes, but I don't think it is the end of the world if it does not as other tests should catch things if the theming gets broken.

pleku (approved):

Issue #9983, opened 2021-02-08
Custom theme feature backporting from master to 2.6


pleku:
- [x] Backport the custom theme feature in a backwards compatible (but forward incompatible) way, by not changing the theme annotation value - instead we add a new parameter to it:
```java
@Theme(themeFolder="foobar")
``` 
this means that we make the default `value` be `Lumo.class`.

What this implies is that:
```java
 @Theme(value=Lumo.class, themeFolder="foobar") // this would be the same as the default

// the following will need throw explaining that the custom theme folder only works with lumo theme
@Theme(value=SomeThingElseThanLumoOrExtendingLumo.class, themeFolder="foobar")
@Theme(value=Material.class, themeFolder="foobar")

// the following will work, given Savu extends Lumo
@Theme(value=Savu.class, themeFolder="foobar")

// because we make Lumo.class the default value, it will be possible to just add the annotation and that should fail because it is an user error
@Theme // this should throw, asking for specifying the custom theme folder
```

- [x] If someone tries using `@Theme(themeFolder="foobar")` when in compatibility mode, we should throw an error noting that custom themes don't' t work in compatibility mode
- [x] Backport the documentation, and note there that:
  - This feature only works for npm-mode, NOT in compatibility mode
  - This feature is available since 14.6 and 19
  - For version 19+ the theme annotation is changed so that the custom folder is the value parameter
- [ ] The SVC templates should be updated to use this feature for v14.
- [x] The theme demo app should get another branch with v14.6 in use
- [x] update [the project used for PiT](https://gitlab.vaadin.com/mgrankvi/appthemedx/-/tree/master) to use 14.6

If there are things that are only applicable for fusion projects, like maybe #9842, those don't have to backported to 14.
Adding the custom theme feature should not regress the frontend build time for v14. The plugins should not be applied when the custom theme folder is not in use.

caalador:
Having an empty `Theme` annotation will work the same way as not having a `Theme` annotation in the application and lead to usage of the default theme if it's available on the classpath

pleku:
Another thing to TODO, update [the project used for PiT](https://gitlab.vaadin.com/mgrankvi/appthemedx/-/tree/master) to use 14.6

caalador:
Created branch: https://gitlab.vaadin.com/mgrankvi/appthemedx/-/tree/14.6/
Still needs the AppShellConfig removed and theme annotation changed to work with 14 branch.

caalador:
[...]

PR #10226, opened 2021-03-08
Custom theme parent theme feature

caalador:
part of #9983


flow-tests/test-application-theme/test-reusable-as-parent/src/main/java/com/vaadin/flow/uitest/ui/servlets/WorkaroundServlet.java
```diff
@@ -0,0 +1,30 @@
...
+ * the License.
+ */
+package com.vaadin.flow.uitest.ui.servlets;
+
+import javax.servlet.annotation.WebServlet;
+
+import com.vaadin.flow.server.VaadinServlet;
+
+/**
+ * This is a temporary workaround until #5740 is fixed.
```
pleku, line 23: BTW This issue has been fixed a long time ago but not sure why/what is this comment referring to, what would not work

pleku (approved):

flow-tests/test-application-theme/test-reusable-as-parent/src/main/java/com/vaadin/flow/uitest/ui/servlets/WorkaroundServlet.java
caalador: Copy paste I would guess as apparently the workarounds have never been removed.
The others add `path` in the servlet, but the javadoc is still the same wrong one. Should probably be fixed in master also.

flow-tests/test-application-theme/test-reusable-as-parent/src/main/java/com/vaadin/flow/uitest/ui/theme/AppShell.java
```diff
@@ -0,0 +1,24 @@
...
+ * the License.
+ */
+
+package com.vaadin.flow.uitest.ui.theme;
+
+import com.vaadin.flow.component.page.AppShellConfigurator;
+import com.vaadin.flow.theme.Theme;
+
+@Theme(value = "child-theme")
+public class AppShell implements AppShellConfigurator {
```
caalador, line 23: =_= Should not be here at all.

flow-tests/test-application-theme/test-reusable-as-parent/src/main/java/com/vaadin/flow/uitest/ui/servlets/WorkaroundServlet.java
caalador: Seems like for some reason the test module fails if the workaround servlet is removed. Perhaps 5740 should not be closed?

pleku (approved):

flow-tests/test-application-theme/test-reusable-as-parent/src/main/java/com/vaadin/flow/uitest/ui/servlets/WorkaroundServlet.java
pleku: I don't even get what it is trying to do - where is the other non-root servlet mapped from ?

In case there is something broken with non root servlets, maybe a new issue could be opened.
Maybe we should refactor the whole testing pattern so that there is another IT profile or whatnot that just executes the tests in another servlet context and we run that, lets say nightly ? Just a tought

caalador: +1 for new issue. Don't really understand what the actual issue is or where it stems from here.

PR #10240, opened 2021-03-09
Folder verification and theme recompilation

caalador:
part of #9983


pleku (approved):
Just noting that I didn't test this intermediate stage as it is probably better to test all the things once everything is available in a prerelease of 14.6.

PR #10245, opened 2021-03-10
[WIP] Add rollup dependency

tanbt:
Rollup is used for bundling the generated custom theme, so the bundled file can be loaded in Designer (and only Designer for now, rollup isn't used anywhere else).


tanbt:
These dependencies will be added by Designer.

PR #10264, opened 2021-03-11
fix: fix shortcuts when listenOn is a dialog

pleku:
This makes it possible for a component to specify a client side only
element that should be used for listening keydown events on when the
component is setup as the listenOn target for a Shortcut. This is needed
for components like Vaadin Dialog which transport the dialog contents
inside an overlay that only exists on the client side and thus breaks
shortcut event handling.

Related to #7799, vaadin/vaadin-dialog#229

### DIFF
--- a/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
+++ b/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
@@ -1,0 +1,188 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.vaadin.flow.uitest.ui;
+
+import java.util.EventObject;
+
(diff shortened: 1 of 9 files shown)


flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
```diff
@@ -0,0 +1,188 @@
...
+
+    @Override
+    protected void onAttach(AttachEvent attachEvent) {
+        super.onAttach(attachEvent);
+        attachEvent.getUI().setId(UI_ID);
+    }
+
+    public void open(Dialog dialog) {
+        // final UI ui = DialogShortcutView.this.getUI().get();
+        // ui.add(this);
```
caalador, line 179: Commented out code that doesn't make sense as a comment.

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
```diff
@@ -0,0 +1,188 @@
...
+        add(dialog);
+    }
+
+    public void close(Dialog dialog) {
+        // final UI ui = DialogShortcutView.this.getUI().get();
+        // ui.remove(this);
```
caalador, line 185: Commented out code that doesn't make sense as a comment.

caalador (changes requested):

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
pleku: Done

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
pleku: Done

caalador (approved):

pleku:
Cherry-picked manually to 2.5 and 2.6.

PR #10280, opened 2021-03-12
fix: close InputStream in FrontendUtils

### LINKED ISSUE #9366: InputStream is not closed in com.vaadin.flow.server.frontend.FrontendUtils
Method `com.vaadin.flow.server.frontend.FrontendUtils#getFileContent` fails to close stream obtained via `getFileFromWebpack` or `getFileFromClassPath`

Method `com.vaadin.flow.server.frontend.FrontendUtils#getStatsContent` fails to close stream obtained via `getStatsFromClassPath`, `getStatsFromExternalUrl` or `getStatsFromWebpack`

knoobie:
fixes #9366

Instead of wrapping only the return, I wrapped the whole method to make sure random exception are closing the stream as well. 

Reasoning behind the change:

>  Wherever possible, the methods in this class do not flush or close the stream. This is to avoid making non-portable assumptions about the streams' origin and further use. Thus the caller is still responsible for closing streams after use. 

Source: http://commons.apache.org/proper/commons-io/javadocs/api-release/org/apache/commons/io/IOUtils.html

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
@@ -459,21 +459,25 @@ public class FrontendUtils {
         DeploymentConfiguration config = service.getDeploymentConfiguration();
         InputStream content = null;
 
-        if (!config.isProductionMode() && config.enableDevServer()) {
-            content = getStatsFromWebpack();
-        }
+        try {
+            if (!config.isProductionMode() && config.enableDevServer()) {
+                content = getStatsFromWebpack();
+            }
 
-        if (config.isStatsExternal()) {
-            content = getStatsFromExternalUrl(config.getExternalStatsUrl(),
-                    service.getContext());
-        }
+            if (config.isStatsExternal()) {
+                content = getStatsFromExternalUrl(config.getExternalStatsUrl(),
+                  service.getContext());
+            }
 
-        if (content == null) {
-            content = getStatsFromClassPath(service);
+            if (content == null) {
+                content = getStatsFromClassPath(service);
+            }
+            return content != null
+                   ? IOUtils.toString(content, StandardCharsets.UTF_8)
+                   : null;
+        } finally {
+            IOUtils.closeQuietly(content);
         }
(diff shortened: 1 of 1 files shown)


pleku (approved):

pleku:
Thanks for the contribution. It applies to 2.x versions but I don't think the automatic CP will pass because the code has forked.

pleku:
>Hi @knoobie , this commit cannot be picked to 5.0 by this bot, can you take a look and pick it manually?

No need to pick manually to 5.0 as it is EOL soon.
Let me know if you want to pick for 2.4 and then we use the bot to get it to other versions or if I'll do it.

knoobie:
@pleku should I wait for the other change of https://github.com/vaadin/flow/pull/10244 to be cherry picked to reduce merge conflicts?

pleku:
Right, it might be so but as I looked at the code in 2.4, I could only detect one place where the `InputStream content` is so it might be just easier to pick it by manually c/p the code there ? The code line is quite different to the one in master

knoobie:
@pleku I can create a commit which contains both changes (cache and closing input stream) with 2.4 as base line if you want. This can be picked to 2.5 / 2.6 afterwards. Probably tomorrow if that's okay and would reduce the need of two separate prs for the same code

pleku:
@knoobie that is fine, just good to let @mshabarov know that he doesn't need to do the backport of #10244 for 2.x even if he promised to do so 😉

mshabarov:
@pleku thanks for remind me :) I've stuck in back-porting, because it needs tests adaptation. And also I think we shouldn't port the new public method for cache clean up to `2.4`, `2.5`, but only to `2.6`, because only `2.6` is in alpha.
Back-port PR to `5.0` also needs some adaptation and new method exclusion.
@knoobie I would appreciate if you cherry pick them, otherwise I'll try to do it tomorrow, or another day this week.

knoobie:
@pleku @mshabarov Gonna create a PR tomorrow with both changes. I'm thinking about adding the cache deleting method as protected in 2.4 and change the modifier to public in 2.6 if that's okay for you to allow for easier testing and porting of the change.

PR #10285, opened 2021-03-15
chore: FrameworkUtil class presence doesn't mean OSGi

denis-anisimov:
Make OSGi check more correct otherwise the code works incorrectly if
FrameworkUtil is present in the classpath


caalador (approved):

pleku:
If this actually fixes a problem with CDI & Payara, maybe it should be highlighted as a fix 🤷

PR #10305, opened 2021-03-17
fix: avoid use servlet context classloader in ResourceProvider impl

denis-anisimov:


caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/startup/LookupInitializer.java
```diff
@@ -143,22 +143,14 @@ private CachedStreamData(byte[] data, IOException exception) {
         @Override
         public List<URL> getApplicationResources(VaadinContext context,
                 String path) throws IOException {
-            if (context instanceof VaadinServletContext) {
```
mshabarov, line 146: Sorry for commenting on closed PR, but wasn't this added for OSGi purposes? I'm just worrying this would break the OSGi add-on.

PR #10308, opened 2021-03-17
chore: update chromium drivers

caalador:


pleku (approved):

pleku:
Just in case needed, I added the pick labels to 2.4 and 5.0 too

caalador:
Left those out as they are going away in the near future, but doesn't ofc hurt to have it in them also.

Issue #10103, opened 2021-02-21
PreserveOnRefresh - enabled, PushMode=Automatic, PushTransport=LongPolling - Application fails when page is reloaded


silvan-lincan:
### Description of the bug / feature
Due to certain environment conditions / restrictions, for one of our clients, WebSockets cannot be enabled, therefore, the application falls back to LONG_POLLING.

PreserveOnRefresh=enabled
PushMode=AUTOMATIC
PushTransport=LONG_POLLING

Everything works nice, with the following exception:
if there is an Overlay opened (e.g. a ConfirmationDialog) and the user makes a page-reload (F5 or click from the browser), the application fails with the following exception and the result is an empty screen:

> Assertion error: No child found with id 3

<img width="1278" alt="Screenshot 2021-02-21 at 21 27 22" src="https://user-images.githubusercontent.com/47028355/108638544-203e5280-7490-11eb-93dd-e1947da1f40c.png">

The following error is displayed on the server-side:
```
[Atmosphere-Shared-2] ERROR com.vaadin.flow.server.communication.PushAtmosphereHandler - Exception in push connection
org.eclipse.jetty.io.EofException
	at org.eclipse.jetty.server.HttpConnection$SendCallback.reset(HttpConnection.java:708)
	at org.eclipse.jetty.server.HttpConnection$SendCallback.access$300(HttpConnection.java:667)
	at org.eclipse.jetty.server.HttpConnection.send(HttpConnection.java:526)
	at org.eclipse.jetty.server.HttpChannel.sendResponse(HttpChannel.java:910)
	at org.eclipse.jetty.server.HttpChannel.write(HttpChannel.java:987)
	at org.eclipse.jetty.server.HttpOutput.channelWrite(HttpOutput.java:284)
	at org.eclipse.jetty.server.HttpOutput.channelWrite(HttpOutput.java:268)
	at org.eclipse.jetty.server.HttpOutput.flush(HttpOutput.java:713)
	at org.eclipse.jetty.server.Response.flushBuffer(Response.java:1110)
	at javax.servlet.ServletResponseWrapper.flushBuffer(ServletResponseWrapper.java:215)
	at org.atmosphere.cpr.AtmosphereResponseImpl.flushBuffer(AtmosphereResponseImpl.java:506)
	at org.atmosphere.cpr.AtmosphereInterceptorWriter.flush(AtmosphereInterceptorWriter.java:102)
[...]

PR #10317, opened 2021-03-17
chore: add ignored test for long polling and @PreserveOnRefresh 

taefi:


mshabarov:
@taefi should these tests be added to https://github.com/vaadin/flow/pull/10582, which is going to fix the issue by upgrading the Atmosphere version?

taefi:
> @taefi should these tests be added to #10582, which is going to fix the issue by upgrading the Atmosphere version?

At the time I was working on #10103, I couldn't create a unit test to isolate the problem, so created the IT. I think if there cannot be a unit test, having this IT would be beneficial.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/PushLongPollingWithPreserveOnRefreshIT.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private void waitPageLoad() {
+        WebElement loadingIndicator = findElement(
+                By.className("v-loading-indicator"));
+        waitUntil(driver -> !loadingIndicator.isDisplayed());
+    }
+
+    private void ensureNoErrorIsDisplayed() {
+        Assert.assertThrows(NoSuchElementException.class,
```
mshabarov, line 66: Can `waitForElementNotPresent` be used here?

taefi: I believe that would behave a little different semantically, since I explicitly want to assert that the element with class name "v-system-error" is not shown at all, not to just wait for its disappearance and then continue with the rest of the process.

mshabarov (approved):

PR #10319, opened 2021-03-17
[plugin] extract flowResourcesFolder into AdapterBase

### LINKED ISSUE #10318: Give BuildUtil possability to  define FLOW_RESOURCES_FOLDER
Will do a PR

Extract  FrontendUtils.DEFAULT_FLOW_RESOURCES_FOLDER into Adapter to have more possibilitiea to control where the resource should be placed. For example, in not maven Layouts, the base path '/target' is not the best idea

https://github.com/vaadin/flow/blob/0be0cb985311bfe9a95eb5b264459a42d6b24ca9/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java#L133

stbischof:
Gives Plugins possability to define FLOW_RESOURCES_FOLDER. 

Closes #10318

### DIFF
--- a/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java
+++ b/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java
@@ -38,6 +38,7 @@ import com.vaadin.flow.plugin.base.BuildFrontendUtil;
 import com.vaadin.flow.plugin.base.PluginAdapterBase;
 import com.vaadin.flow.server.Constants;
 import com.vaadin.flow.server.frontend.FrontendTools;
+import com.vaadin.flow.server.frontend.FrontendUtils;
 import com.vaadin.flow.server.frontend.installer.NodeInstaller;
 import com.vaadin.flow.server.frontend.scanner.ClassFinder;
 
@@ -371,5 +372,11 @@ public abstract class FlowModeAbstractMojo extends AbstractMojo
     @Override
     public boolean isJarProject() {
         return "jar".equals(project.getPackaging());
+    }
+
+    @Override
+    public String flowResourcesFolder() {
+
+        return FrontendUtils.DEFAULT_FLOW_RESOURCES_FOLDER;
     }
 }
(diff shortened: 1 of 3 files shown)


flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java
```diff
@@ -372,4 +373,10 @@ public File webpackOutputDirectory() {
     public boolean isJarProject() {
         return "jar".equals(project.getPackaging());
     }
+
+    @Override
+    public String flowResourcesFolder() {
+
```
caalador, line 379:
```suggestion
```

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/PluginAdapterBase.java
```diff
@@ -274,5 +274,13 @@ default boolean isUseDeprecatedV14Bootstrapping() {
      */
     File webpackOutputDirectory();
 
+    /**
+     *  Default folder for copying front-end resources present in the classpath
+     * jars.
+     * 
+     * @return {@link String}
+     */
```
caalador, line 282:
```suggestion
    /**
     * Default target folder to copy front-end resources to from jars present
     * on the classpath and project local frontend files in
     * frontendResourcesDirectory.
     *
     * @return {@link String}
     */
```

caalador (changes requested):
`BuildFrontendUtil::runNodeUpdater` L241 still has the hardcoded folder  
```
        File flowResourcesFolder = new File(adapter.npmFolder(),
                DEFAULT_FLOW_RESOURCES_FOLDER);
```

Also as this is only set in BuildFrontend I guess running only prepare will generate a different result as it will still use the default target directory `File flowResourcesFolder = new File(baseDir, DEFAULT_FLOW_RESOURCES_FOLDER);` `DevModeInitialize:L278` The given folder should be also delegated here when only running PrepareFrontend

stbischof:
There is an other issue that will fix the static use of the target folder.

caalador:
But that is another case. Here we are adding the `flowResourcesFolder` through the plugin which should probably be written to the token file to get it into the configuration object in prepare-frontend and then removed from there in build-frontend step as then we shouldn't have any paths stored in the token file.

DevModeInitializer should then read the `PARAM_FLOW_RESOURCES` from the configuration as `File flowResourcesFolder = new File(baseDir, config.getStringProperty(PARAM_FLOW_RESOURCES, DEFAULT_FLOW_RESOURCES_FOLDER));`

caalador:
Is there still a need for this change as #10335 was fixed and the resources will go according to the defined build directory?

stbischof:
yes this should be enough!

PR #10343, opened 2021-03-18
fix: Add API to control whether Binder converts back to presentation…

### LINKED ISSUE #10342: fix: Add API to control whether Binder converts back to presentation (forward port from v8)
### Description

Since Vaadin 7 the binder API did not propagate conversion back to presentation. This changed in [8.12.1](https://github.com/vaadin/framework/pull/12132) and [19.0.0](https://github.com/vaadin/flow/pull/9227) to always update the field with the converted value as well. 

This fix / behaviour change introduced a major regression in most of our application that were pretty hard to spot and raised a lot of problems on our side that's why I opened VS-3752 after discussing this with Anna and Tatu to allow for configuration of the behaviour. 

### Use Case

We heavily rely on Binder and it's conversion to protect us from XSS attacks and other security concerns that come from user input. To reduce the need of checking every field on submit, we always trigger those conversion within the 

```java
    binder.forField(name)
      .withConverter(new CleanUpUserInputConverter())
      .bind(MyObject::getName, MyObject::setName);
```

One of those conversion uses `Jsoup` to escape anything `html` related like e.g. `Jsoup.clean(value, Whitelist.none())` - *this resulting in removal of trailing white spaces*.

This and other validations have to be done while the user is typing / filling out the field to allow for instant feedback (default 400ms from Textfield), that's why `onBlur` isn't really an option for us.

With the use case above, you get a okayish sanitized user input for further processing and the user gets instant feedback while typing without any magic which reformats his input. ;)  

### Problem Description

After we updated our applications to a version that has the fix, we got calls from customers that they space bar won't work anymore and we got messed up inputs from some users. We analyzed the problem and tried to reproduce it, which was pretty easy. While typing *slowly* you can literally see that the trailing space from an input is removed - resulting in a major problem for our older user. Our user base is kinda diverse (18-80 years old). 

To fix the problem for now, I updated our conversions to check if the input has a trailing space and adds it after the conversion. This fixes the problem for the user, but results in problems later for us because now we have to remove every trailing space on submit. 

### Example (V8)

https://gist.github.com/knoobie/74fe2bf2e4745be4bd52d5ccd48566d3

knoobie:
… (forward port from FW 8 (framework#12246))

fixes #10342

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -255,8 +255,14 @@ public class Binder<BEAN> implements Serializable {
          * Define whether the value should be converted back to the presentation
          * in the field when a converter is used in binding.
          * <p>
-         * As of version 6.0, when a converter is used on a binding and the user input value is modified by the converter,
-         * the value from the converter is applied back to the input. It is possible to control this behavior with this API.
+         * As of version 6.0, when a converter is used on a binding and the user
+         * input value is modified by the converter, the value from the converter
+         * is applied back to the input. It is possible to control this behavior
+         * with this API.
+         *
+         * @see BindingBuilder#withConverter(Converter)
+         * @see BindingBuilder#withConverter(SerializableFunction, SerializableFunction)
+         * @see BindingBuilder#withConverter(SerializableFunction, SerializableFunction, String)
          *
          * @param convertBackToPresentation
          *            A boolean value
@@ -528,6 +534,12 @@ public class Binder<BEAN> implements Serializable {
          * For instance, a {@code TextField} can be bound to an integer-typed
          * property using an appropriate converter such as a
          * {@link StringToIntegerConverter}.
+         * <p>
+         * The converted value is applied back to the field by default,
+         * this can be controlled with the method
+         * {@link Binding#setConvertBackToPresentation(boolean)}.
+         *
+         * @see Binding#setConvertBackToPresentation(boolean)
          *
          * @param <NEWTARGET>
          *            the type to convert to
@@ -554,6 +566,12 @@ public class Binder<BEAN> implements Serializable {
          * For instance, a {@code TextField} can be bound to an integer-typed
          * property using appropriate functions such as:
          * <code>withConverter(Integer::valueOf, String::valueOf);</code>
+         * <p>
+         * The converted value is applied back to the field by default,
+         * this can be controlled with the method
+         * {@link Binding#setConvertBackToPresentation(boolean)}.
+         *
+         * @see Binding#setConvertBackToPresentation(boolean)
          *
          * @param <NEWTARGET>
          *            the type to convert to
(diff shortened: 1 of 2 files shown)


knoobie:
@pleku If you want additional tests, like you mentioned in the pr that introduced this behavior (https://github.com/vaadin/flow/pull/9227) just let me know.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -250,6 +250,23 @@
          * @return A boolean value.
          */
         public boolean isValidatorsDisabled();
+
+        /**
+         * Define whether the value should be converted back to the presentation
+         * in the field when a converter is used in binding.
+         *
```
pleku, line 257:
```suggestion
         * <p>
         * As of version 6.0, when a converter is used on a binding and the user input value is modified by the converter,
         * the value from the converter is applied back to the input. It is possible to control this behavior with this API.
         *
```

pleku (changes requested):
Can you please add the `@see #...` links to the `withConverter` methods in `BindingBuilder` for the added API here. An mayhaps even clarify the javadocs for those that 
`The converter value is applied back to the field by default, but this can be controlled with the method {@link ...}.` 
I can do it myself too later on if needed. I should have done it for the original PR but lets fix things now.

Even as this adds API, I consider it to be a fix instead of feature since this is addressing a mistake in the original feature, thus I would land this as a fix to v19.

knoobie:
@pleku Sure, gonna update the other javadocs as well and improve your suggested message I just accepted to align with the character limit

pleku (approved):

pleku:
Thanks @knoobie.

pleku:
I don't know what happened but I apparently didn't use squash 🤦‍♂️

knoobie:
Want me to do something about it or can you fix it somehow with force push? I can create a pr for 6.0 as well

pleku:
Thanks for the manual CP @knoobie.

It would be now possible to cherry-pick both of the changes to 2.6 too but keep the default behavior (not write back changes).

knoobie:
@pleku The original change from Tatu did not land in 2.x (https://github.com/vaadin/flow/pull/9227) - should this be backported as well? Personally I would not backport Tatu and my change to 2.x but add a migration note to the next LTS (22) that the default behaviour is changed and can be configured with the new methods. 

My next step would have been to forward port https://github.com/vaadin/framework/issues/12207 with a proper test

pleku:
> should this be backported as well? 

Yes, that is what I meant to backport both the behavior change + the API to control it to get the value to the 14-series and it could allow upgrading towards next LTS behavior inside 14, if one wants to do so. But looking at vaadin/framework#12207 that would be more important.

knoobie:
Okay! Gonna forward port https://github.com/vaadin/framework/issues/12207 and see what I can do with cherry picking the changes from master to 2.x with a changed default value of `private boolean convertBackToPresentation = true;` to `false`.

PR #10350, opened 2021-03-19
Backport of Lookup initialization

denis-anisimov:


flow-server/src/main/java/com/vaadin/flow/server/osgi/OSGiAccess.java
```diff
@@ -92,6 +98,55 @@ private OSGiAccess() {
...
+    /**
+     * OSGi capable implementation of {@link ResourceProvider}.
+     *
+     */
+    private static class OSGiResourceProvider implements ResourceProvider {
+
+        @Override
+        public List<URL> getApplicationResources(VaadinContext context,
+                String path) throws IOException {
+            if (context instanceof VaadinService) {
```
mshabarov, line 110: Last time in `master` this way of getting the resource caused the exception in Spring add-on.
Don't know, maybe it's totally fine, but I though to stress it just in case.

mshabarov: Oh, this is an OSGi impl, sorry, never mind.

mshabarov: But another question is why is this impl differ from one in `OSGi` add-on.
The `Bundle` instance is used there to get the application resources.

mshabarov: Aaaand...going further the code, I saw `OSGiResourceProvider` class has been removed and transferred here. It explains all the things :)

flow-server/src/main/java/com/vaadin/flow/server/startup/LookupServletContainerInitializer.java
```diff
@@ -0,0 +1,213 @@
...
+    private Set<Class<?>> filterSubClasses(Class<?> clazz,
+            Set<Class<?>> classes) {
+        return classes == null ? Collections.emptySet()
+                : classes.stream().filter(clazz::isAssignableFrom)
+                        .filter(this::isInstantiableService)
+                        .filter(cls -> !clazz.equals(cls))
+                        .collect(Collectors.toSet());
+    }
+
+    private boolean isInstantiableService(Class<?> clazz) {
```
mshabarov, line 162: minor: this method is in `ReflectTools` in `master`, but since it is used only here, maybe it's fine.

mshabarov (approved):

mshabarov:
@denis-anisimov I'm about to merge this PR, but not certain about the prefix, should it be `refactor` or `feat` ?

PR #10358, opened 2021-03-22
fix: No deploy wihout lookup

denis-anisimov:


flow-server/src/main/java/com/vaadin/flow/server/startup/ServletDeployer.java
```diff
@@ -226,6 +226,14 @@ public static DeploymentConfiguration createDeploymentConfiguration(
     @Override
     public void contextInitialized(ServletContextEvent sce) {
         ServletContext context = sce.getServletContext();
+
+        VaadinContext vaadinContext = new VaadinServletContext(context);
+        if (vaadinContext.getAttribute(Lookup.class) == null) {
+            // if runtime has not yet initialized Lookup then it's up to runtime
+            // to deploy the servlet
```
mshabarov, line 233: Does this mean that if no `Lookup` instance provided at this point, then the custom `LookupInitializer` is supposed to be present, as well as custom servlet deployment configuration?

`ServletDeployer` javadoc needs to explain it.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/startup/ServletDeployer.java
denis-anisimov: No.
It means that deployer should not do anything if there is no Lookup.

mshabarov: > It means that deployer should not do anything if there is no Lookup.

Okay, when it would happen? When can it be absent?
As I know, `Lookup` should always be present in order to discover and use flow services.
Either a default lookup impl is added into context, or some other custom way is used to make it available through `VaadinContext.getAttribute(Class)`.
Doesn't that mean that something is actually broken, when no `Lookup` present on servlet deployment phase?

The idea to skip servlet deployment, when no lookup found, is not clear.
What does it mean from the client code point of view?

denis-anisimov: `Servlet 3.0` spec invokes `ServletContextInitialer` s first and then `ServletListener`s.
Lookup is initialized inside one of `ServletContainerInitializer`.
Other `ServletContainerInitializer`s should be invoked in the proper order which is forced if initializer extends `ClassLoaderAwareServletContainerInitializer`.
If it doesn't extend `ClassLoaderAwareServletContainerInitializer` then Lookup doesn't have to be in the context.

So : NO . `Lookup`  is not always present.
For Servlet 3.0 container `ServletListener` should have Lookup in the context.

>Doesn't that mean that something is actually broken, when no Lookup present on servlet deployment phase?

Yes, the container is broken. Because it's not Servlet 3.0 container. The example is Spring boot which works in a Spring way.
I don't know why Spring boot doesn't execute `ServletContainerInitializer` but it executes `ServletListener`  which may be executed when `ApplicationContext`  is not yet available. But since it happens it's up to non Servlet 3.0 capable container to execute ServletDeployer logic by itself in the same way as Spring boot executes `ServletContainerInitializer`s.
And this is exactly what's said here:
`it's up to runtime to deploy the servlet`

mshabarov (approved):

Issue #10302, opened 2021-03-17
DialogShortcutIT::dialogOpenedWithShortcutNoListenOn_sameShortcutListeningOnUi_bothExecuted is flaky


denis-anisimov:
`DialogShortcutIT::dialogOpenedWithShortcutNoListenOn_sameShortcutListeningOnUi_bothExecuted` failed in `2.6`:

https://bender.vaadin.com/viewLog.html?buildId=219206&tab=buildResultsDiv&buildTypeId=Flow_Flow26_Snapshot

Screenshot if link doesn't work:

![dialogOpenedWithShortcutNoListenOn_sameShortcutListeningOnUi_bothExecuted ANY_Chrome_ (com vaadin flow uitest ui DialogShortcutIT)](https://user-images.githubusercontent.com/11425472/111423248-0c89b300-8701-11eb-8cb2-1c4b0c9011af.png)

pleku:
It seems quite random that which test fails from `ComplexShortcutIT` or `DialogShortcutIT` so it might be just sense to merge this to https://github.com/vaadin/flow/issues/10284. I'm working on refactoring both tickets anyway.

mshabarov:
One more failure spotted in Flow 6.0 snapshot validation:

https://bender.vaadin.com/viewLog.html?buildId=221404&buildTypeId=Flow_Flow60_Flow60Snapshot

![image](https://user-images.githubusercontent.com/61410877/112873290-1b417400-90ca-11eb-8691-244df1b79703.png)


```
java.lang.RuntimeException: dialogOpenedWithShortcutNoListenOn_sameShortcutListeningOnUi_bothExecuted[ANY_Chrome_](com.vaadin.flow.uitest.ui.ComplexDialogShortcutIT): run failed 2 times
	at com.vaadin.testbench.RetryRule$1.evaluate(RetryRule.java:69)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
[...]

Issue #10284, opened 2021-03-15
ComplexDialogShortcutIT.dialogOpenedWithListenOnShortcut_sameShortcutListeningOnUi_focusDecidesWhichIsExecuted[ANY_Chrome_] is flaky


denis-anisimov:
`ComplexDialogShortcutIT::dialogWithShortcutListenOnAndPreserveOnRefresh_refreshWhenDialogOpened_shortcutScopingWorks` is flaky in all branches.

It at least failed in the `6.0` SNAPSHOT and `2.6`  SNAPSHOT:

https://bender.vaadin.com/viewLog.html?buildId=218806&tab=buildResultsDiv&buildTypeId=Flow_Flow60_Flow60Snapshot
https://bender.vaadin.com/viewLog.html?buildId=218817&tab=buildResultsDiv&buildTypeId=Flow_Flow26_Snapshot

Here is an error snapshot if it's not available by the link in the builds
![dialogWithShortcutListenOnAndPreserveOnRefresh_refreshWhenDialogOpened_shortcutScopingWorks ANY_Chrome_ (com vaadin flow uitest ui ComplexDialogShortcutIT)](https://user-images.githubusercontent.com/11425472/111132588-98c59a00-858a-11eb-8862-16b02f81d6ee.png)

denis-anisimov:
I've disabled the test since it fails too many builds: it's not possible to get latest SNAPSHOTS for many versions because of it.

pleku:
The wrong test was mentioned to be flaky, it is not the preserve-on-refresh test.

denis-anisimov:
Both of them are flaky.
Sometimes fails one sometimes both fail.

mshabarov:
Two more failures occur recently:

https://bender.vaadin.com/viewLog.html?buildId=221347&buildTypeId=Flow_Flow60_Flow60Snapshot

![image](https://user-images.githubusercontent.com/61410877/112874828-1087de80-90cc-11eb-9858-1c5b3a7257f0.png)

```
java.lang.RuntimeException: dialogOpenedWithListenOnShortcut_sameShortcutListeningOnUi_focusDecidesWhichIsExecuted[ANY_Chrome_](com.vaadin.flow.uitest.ui.DialogShortcutIT): run failed 2 times
	at com.vaadin.testbench.RetryRule$1.evaluate(RetryRule.java:69)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
[...]

PR #10359, opened 2021-03-22
chore: refactor flaky test

pleku:
Refactors the code a bit to make it easier to understand why are dialog
shortcut ITs flaky in the CI. Only for master branch now.

Related to #10302, #10284

### DIFF
--- a/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
+++ b/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
@@ -17,6 +17,7 @@
 package com.vaadin.flow.uitest.ui;
 
 import java.util.EventObject;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import com.vaadin.flow.component.AttachEvent;
 import com.vaadin.flow.component.ClickEvent;
@@ -48,7 +49,7 @@ public class DialogShortcutView extends Div {
     public static final Key SHORTCUT_KEY = Key.KEY_X;
     public static final String REUSABLE_DIALOG = "reusable-dialog";
 
-    private int dialogCounter = 0;
+    private AtomicInteger dialogCounter = new AtomicInteger(-1);
     private int eventCounter;
     private final Div eventLog;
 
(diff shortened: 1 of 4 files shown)


flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
```diff
@@ -48,7 +49,7 @@
     public static final Key SHORTCUT_KEY = Key.KEY_X;
     public static final String REUSABLE_DIALOG = "reusable-dialog";
 
-    private int dialogCounter = 0;
+    private AtomicInteger dialogCounter = new AtomicInteger(-1);
```
mshabarov, line 52: Previously, initial value was `0`, but now `-1`. Is it intentionally?

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/DialogShortcutIT.java
```diff
@@ -76,60 +77,62 @@ public void dialogOpenedWithShortcutNoListenOn_sameShortcutListeningOnUi_bothExe
     public void dialogOpenedWithListenOnShortcut_dialogReopened_oldShortcutStillWorks() {
         openReusedDialog();
 
-        pressShortcutKey(getFirstDialogInput());
+        pressShortcutKey(getDialogInput(reusedDialogIndex));
```
mshabarov, line 80: minor: Even though the `reusedDialogIndex` should be incremented after `openReusedDialog`,  would be more readable if you explicitly return the result like this:
`reusedDialogIndex = openReusedDialog();`

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/DialogShortcutIT.java
```diff
@@ -76,60 +77,62 @@ public void dialogOpenedWithShortcutNoListenOn_sameShortcutListeningOnUi_bothExe
     public void dialogOpenedWithListenOnShortcut_dialogReopened_oldShortcutStillWorks() {
         openReusedDialog();
```
mshabarov, line 78: minor: Even though the `reusedDialogIndex` should be incremented after `openReusedDialog`,  would be more readable if you explicitly return the result like this:
`reusedDialogIndex = openReusedDialog();`

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/DialogShortcutIT.java
```diff
@@ -76,60 +77,62 @@ public void dialogOpenedWithShortcutNoListenOn_sameShortcutListeningOnUi_bothExe
...
         pressShortcutKey(uiLevelButton);
         validateLatestShortcutEvent(3, DialogShortcutView.UI_BUTTON);
     }
-    protected void openReusedDialog() {
+
+    protected int openReusedDialog() {
         findElement(By.id(DialogShortcutView.REUSABLE_DIALOG)).click();
-        dialogCounter++;
+        if (reusedDialogIndex == -1) {
+            reusedDialogIndex = dialogCounter.incrementAndGet();
```
mshabarov, line 128: I'm not entirely sure, but `reusedDialogIndex` changes might be invisible in tests which run in parallel by more than one thread.
Even though the increment is done atomically, I'd recommend to make `reusedDialogIndex` volatile.

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DialogShortcutView.java
pleku: Yes

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/DialogShortcutIT.java
pleku: I'd rather keep the setting of the field in the `openReusedDialog()` method instead of replicate it in all tests. The value is only returned so the Complex test case can use it for waiting for the element transport to overlay.

pleku: Actually I refactored the code to make the index a shared constant for every single time since there will be a one element in the browser with the ID anyway.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/DialogShortcutIT.java
pleku: I refactored things to make the index of the reused dialog the same always.

mshabarov (approved):

pleku:
There were two consecutive snapshots builds green for 7.0 after the refactoring so I think the issue was addressed 😅

PR #10361, opened 2021-03-22
chore: skip service worker plugin for non PWAs

haijian-vaadin:
Reduce frontend build time for non-PWA applications by not running the service worker Webpack plugin.

### DIFF
--- a/flow-server/src/main/resources/webpack.generated.js
+++ b/flow-server/src/main/resources/webpack.generated.js
@@ -145,24 +145,26 @@ const swManifestTransform = (manifestEntries) => {
   return { manifest, warnings };
 };
 
-const serviceWorkerPlugin = new InjectManifest({
-  swSrc: clientServiceWorkerEntryPoint,
-  swDest: serviceWorkerPath,
-  manifestTransforms: [swManifestTransform],
-  maximumFileSizeToCacheInBytes: 100 * 1024 * 1024,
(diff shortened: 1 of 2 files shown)


flow-server/src/main/resources/webpack.generated.js
```diff
@@ -123,46 +123,50 @@ if (useClientSideIndexFileForBootstrapping) {
...
+    if (indexEntryIdx !== -1) {
+      manifest[indexEntryIdx].url = appShellUrl;
+      appShellManifestEntry = manifest[indexEntryIdx];
+    } else {
+      // Index entry is only emitted on first compilation. Make sure it is cached also for incremental builds
+      manifest.push(appShellManifestEntry);
+    }
+  
+    return { manifest, warnings };
+  };
```
joheriks, line 147: Does it actually provide any performance improvement to wrap `swManifestTransform` into a function, since the value is already a function (meaning that the body doesn't get executed until it is actually called).

joheriks (changes requested):

flow-server/src/main/resources/webpack.generated.js
haijian-vaadin: good catch, fixed.

joheriks (approved):

joheriks (approved):

PR #10363, opened 2021-03-22
chore: internal helper identifying framework internal requests

joheriks:
Part of vaadin/spring#735.


flow-server/src/main/java/com/vaadin/flow/internal/SecurityHelper.java
```diff
@@ -45,4 +49,20 @@
...
+     *
+     * @param request the servlet request
+     * @return {@code true} iff the request is Vaadin internal.
+     */
+    public static boolean isFrameworkInternalRequest(HttpServletRequest request) {
+        final String parameterValue = request
+                .getParameter(ApplicationConstants.REQUEST_TYPE_PARAMETER);
+        return parameterValue != null && Stream
+                .of(HandlerHelper.RequestType.values())
+                .anyMatch(r -> r.getIdentifier().equals(parameterValue));
```
pleku, line 66: Would it make sense to refactor this code to reuse `HandlerHelper` code
```java
    public static boolean isRequestType(VaadinRequest request,
            RequestType requestType) {
```
as basically this is just doing the same but looking at all request types ?

pleku (changes requested):
One comment in code and also I'm not sure how "secure" is this API as one can quite easily just fake such a request type. Should the code maybe be in `HandlerHelper` instead to downplay this as a _security_ thing and maybe be more discoverable.

flow-server/src/main/java/com/vaadin/flow/internal/SecurityHelper.java
```diff
@@ -45,4 +49,20 @@
...
+     * presence of the {@code v-r} request parameter with a type matching a
+     * {@link HandlerHelper.RequestType}.
+     *
+     * @param request the servlet request
+     * @return {@code true} iff the request is Vaadin internal.
+     */
+    public static boolean isFrameworkInternalRequest(HttpServletRequest request) {
+        final String parameterValue = request
+                .getParameter(ApplicationConstants.REQUEST_TYPE_PARAMETER);
+        return parameterValue != null && Stream
```
Artur-, line 64: You need to look at the URL also. A request to `/admin/verysecret.png?some=parameter` is not an internal framework request

flow-server/src/main/java/com/vaadin/flow/internal/SecurityHelper.java
Artur-: Moved the method to HandlerHelper but `isRequestType` cannot be used as it requires a `VaadinRequest`

flow-server/src/main/java/com/vaadin/flow/internal/SecurityHelper.java
Artur-: Fixed

flow-server/src/main/java/com/vaadin/flow/internal/SecurityHelper.java
joheriks: Shouldn't `/myview?v-r=init` count as an internal request? At least Fusion client sends such requests to initialize Flow when opening a non-root server-side route:
https://github.com/vaadin/flow/blob/a685db13dc42012400c448834b2286ea595d8167/flow-client/src/main/resources/META-INF/resources/frontend/Flow.ts#L338

Artur-: I see, the problem is that you cannot use only pathInfo for this. Spring Security matchers always combine `getServletPath()` with `getPathInfo` so you get the full path relative to the context path. 

When e.g. requesting `"/VAADIN/build/vaadin-devmodeGizmo-2aa6cd8571de3aa0f313.cache.js"` pathInfo is `null` and `servletPath` is set to `"/VAADIN/build/vaadin-devmodeGizmo-2aa6cd8571de3aa0f313.cache.js"`. Similarly for `http://localhost:9090/bar` where `bar` should the path info but with Spring Security you get `"/bar"` as `servletPath` and `null` as `pathInfo`

We cannot ignore the path completely because that would mean that a secure resource like `/secret/topsecret.pdf` would become an internal Vaadin request by adding `?v-r=init` and probably then set to public access at the same time.

Artur-: Is it correct that Flow.ts sends the `v-r=init` request to the current location or should it always send it to the context root?

Artur-: So you can trust `pathInfo` in Spring when calling a servlet it seems. When fetching a static resource or calling an endpoint, it is `null`. So I think this approach should work if we can merge #10431

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -107,6 +110,31 @@ public static boolean isRequestType(VaadinRequest request,
...
+     */
+    public static boolean isFrameworkInternalRequest(HttpServletRequest request) {
+        String pathInfo = request.getPathInfo();
+        if (pathInfo == null || pathInfo.isEmpty() || "/".equals(pathInfo)) {
+            // According to the spec, pathInfo should be null but not all servers implement
+            // it like that...
+            // Additionally the spring servlet is mapped as /vaadinServlet right now it
+            // seems but requests are sent to /vaadinServlet/, causing the "/" path info
+            final String parameterValue = request.getParameter(ApplicationConstants.REQUEST_TYPE_PARAMETER);
+            return Stream.of(RequestType.values()).anyMatch(r -> r.getIdentifier().equals(parameterValue));
```
knoobie, line 132: This code looks like a hot code paths. I would personally don't use `Stream::of` here and use the gold old `for(RequestType r : RequestType.values()` for better performance.

Artur-: I think we could actually simplify it to `return parameterValue != null`

knoobie: I would argue against this, because in the end `parameterValue` is a user value and should be validated / sanitized against the valid values? Or am I too overly cautious?

Artur-: In this case we probably want to block all requests to `/` with an unknown `v-r` value and make it clear that is reserved for Vaadin internal use

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -107,6 +110,58 @@ public static boolean isRequestType(VaadinRequest request,
...
+     * The requests listed in {@link RequestType} are considered internal.
+     * Especially bootstrap requests for any route or static resource requests
+     * are not internal, even resource requests for the JS bundle.
+     *
+     * @param request
+     *            the servlet request
+     * @return {@code true} if the request is Vaadin internal, {@code false}
+     *         otherwise
+     */
+    public static boolean isFrameworkInternalRequest(
```
Artur-, line 142: This really needs an additional check. If you map a `VaadinServlet` as `/myvaadin/*` then this would still return true for a request to `/?v-r?init`

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -194,8 +195,8 @@ private void includeInitialUidl(JsonObject initialJson,
 
     @Override
     protected boolean canHandleRequest(VaadinRequest request) {
-        return request.getService().getBootstrapUrlPredicate()
```
pleku, line 197: Should this API in the service be killed since it is not used from anywhere after this PR - both the setter and the getter are dead code ?

flow-server/src/test/java/com/vaadin/flow/server/BootstrapHandlerTest.java
```diff
@@ -1535,6 +1535,20 @@ public void testUIConfiguration_usingPageSettings() throws Exception {
                 testUI.getPushConfiguration().getPushMode());
     }
 
+    @Test
+    public void internal_request_no_bootstrap_page() {
```
pleku, line 1539: nit: please stick to the existing test method naming convention next time

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -109,6 +111,114 @@ public static boolean isRequestType(VaadinRequest request,
...
+            return true;
+        }
+
+        return false;
+    }
+
+    private static boolean isInternalRequestInsideServlet(
+            String requestedPathWithoutServletMapping,
+            String requestTypeParameter) {
+        if (requestedPathWithoutServletMapping == null
```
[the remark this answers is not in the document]
Artur-: They actually check different things. The check on 172 returns false if the path is not inside the servlet mapping. This method considers a `null` `pathInfo` the same as `""` and checks the request type

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
Artur-: It is still used

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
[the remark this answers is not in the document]
pleku: Doesn't that mean that for the same `VaadinServletRequest` request, whether it is considered internal or not, can be different for the two `isFrameworkInternalRequest` overload methods then ?
Since the one which takes `VaadinRequest` reports _an internal request_ when `requestedPathWithoutServletMapping == null` but the `requestTypeParameter` is not.
And one that takes `HttpServletRequest` reports _not internal_ `requestedPathWithoutServletMapping == null` regardless of the presence of the `requestTypeParameter`. So I take it that this is intentional, but I don't understand exactly why based on the code.

Maybe what is missing for being able to understand this even later on, is the intention on when each method should be called or how do they differ, which should be in the javadocs. Now they just take different parameters but have the same name and javadocs.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
pleku: But there is nothing ever setting the predicate to anything else than the default which always return true ? Or is it being set from Spring ?

Artur-: It has nothing to do with this change

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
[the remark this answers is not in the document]
Artur-: The only difference between the two methods is that the method taking `VaadinRequest` knows that the request is targeting a vaadin servlet. The other method first checks that the request is targeting a Vaadin servlet, using the given `servletMappingPath`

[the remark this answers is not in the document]
Artur-: The version taking a `VaadinRequest` is only needed to do an extra check in `BootstrapHandler` and `IndexHtmlRequestHandler` to ensure that they do not bootstrap something for an "internal request". It could be better to move it to `IndexHtmlRequestHandler` instead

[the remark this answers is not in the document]
Artur-: But that would mean that ```
isInternalRequestInsideServlet(
            String requestedPathWithoutServletMapping,
            String requestTypeParameter) {
```
would need to be public, which it shouldn't be...

File: flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java (hunk 5 above)
Comment (pleku) [reply flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java:197]: Sure, it can be axed in a following PR

File: flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java (hunk 7 above)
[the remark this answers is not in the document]
Comment (pleku) [reply flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java:185]: Please just add the javadoc that documents that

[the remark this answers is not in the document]
Comment (Artur-) [reply flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java:167]: Ok, I moved one of the methods to `BootstrapHandler` instead so it is out of sight. Changed some docs and made the `getPathIfInsideServlet` method return an `Optional` instead so the return value from that is not confused with a `null` `pathInfo` value

File: flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java [hunk 8]
```diff
@@ -109,6 +111,114 @@ public static boolean isRequestType(VaadinRequest request,
...
+        }
+        return false;
+    }
+
+    /**
+     * Checks whether the request is an internal request.
+     *
+     * The requests listed in {@link RequestType} are considered internal.
+     * Especially bootstrap requests for any route or static resource requests
+     * are not internal, even resource requests for the JS bundle.
```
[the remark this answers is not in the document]
Artur-: Rewrote it

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -109,6 +112,96 @@ public static boolean isRequestType(VaadinRequest request,
...
+                .of(requestedPath.substring(servletMappingPath.length()));
+    }
+
+    private static String getRequestPathInsideContext(
+            HttpServletRequest request) {
+        String servletPath = request.getServletPath();
+        String pathInfo = request.getPathInfo();
+        String url = "";
+        if (servletPath != null) {
+            url += servletPath;
```
joheriks, line 197: This branch lacks test coverage, would it make sense to add a test like:
```
    @Test
    public void test() {
        HttpServletRequest request = createRequest("/", RequestType.INIT);
        Mockito.when(request.getServletPath()).thenReturn("/abc");

        Assert.assertTrue(
                HandlerHelper.isFrameworkInternalRequest("/abc", request));
        Assert.assertFalse(
                HandlerHelper.isFrameworkInternalRequest("/", request));
    }
```

Artur-: Added

flow-server/src/test/java/com/vaadin/flow/server/HandlerHelperTest.java
```diff
@@ -119,49 +119,86 @@ public void isFrameworkInternalRequest_validType_withPath() {
...
         VaadinRequest request = createVaadinRequest("/hello", "/myservlet",
                 null);
 
-        Assert.assertFalse(HandlerHelper.isFrameworkInternalRequest(request));
+        Assert.assertFalse(
+                BootstrapHandler.isFrameworkInternalRequest(request));
+    }
+
+    @Test
+    public void test() {
```
pleku, line 194: `test()` ?

pleku (dismissed):
...

flow-server/src/test/java/com/vaadin/flow/server/HandlerHelperTest.java
Artur-: Removed

pleku (approved):

Issue #10364, opened 2021-03-23
WebComponentIT.downloadLinkHasCorrectBaseURL is flaky


mshabarov:
https://bender.vaadin.com/viewLog.html?buildId=220268&buildTypeId=Flow_Flow25_Nightly_ValidationWithWindows

mshabarov:
```
java.lang.RuntimeException: downloadLinkHasCorrectBaseURL[ANY_Chrome_](com.vaadin.flow.webcomponent.WebComponentIT): run failed 2 times
	at com.vaadin.testbench.RetryRule$1.evaluate(RetryRule.java:69)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.apache.maven.surefire.junitcore.pc.Scheduler$1.run(Scheduler.java:410)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: org.openqa.selenium.NoSuchElementException: No element with tag <a> found with the attributes [id='link'] using index 0
For documentation on this error, please visit: https://www.seleniumhq.org/exceptions/no_such_element.html
Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:17:03'
System info: host: '8A2A47E8BF40', ip: '172.22.248.86', os.name: 'Windows Server 2019', os.arch: 'amd64', os.version: '10.0', java.version: '1.8.0_252'
Driver info: driver.version: unknown
	at com.vaadin.testbench.ElementQuery.get(ElementQuery.java:303)
[...]

PR #10365, opened 2021-03-23
test: Fix flaky tests in WebComponentIT

mshabarov:
Waits for the elements in 'show-message' web component shadow-dom to be appear before testing them.

Fixes https://github.com/vaadin/flow/issues/10364

### DIFF
--- a/flow-tests/test-embedding/embedding-test-assets/src/test/java/com/vaadin/flow/webcomponent/WebComponentIT.java
+++ b/flow-tests/test-embedding/embedding-test-assets/src/test/java/com/vaadin/flow/webcomponent/WebComponentIT.java
@@ -40,11 +40,8 @@ public class WebComponentIT extends ChromeBrowserTest implements HasById {
         waitForElementVisible(By.id("show-message"));
 
         TestBenchElement showMessage = byId("show-message");
-        waitUntil(driver -> {
-            List<WebElement> selects = findInShadowRoot(showMessage, By.tagName(
-                    "select"));
-            return selects != null && !selects.isEmpty();
-        });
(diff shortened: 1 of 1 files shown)


pleku (approved):
I wonder if this type of things should be fixed instead by making sure that testbench works with changes related to the embedded web component. Now that I look at this test, it seems like that is probably broken/not supported ? If you agree, can you create an issue to /testbench ?

pleku:
Also, what branches should this end up to @mshabarov ?

mshabarov:
> I wonder if this type of things should be fixed instead by making sure that testbench works with changes related to the embedded web component. Now that I look at this test, it seems like that is probably broken/not supported ? If you agree, can you create an issue to /testbench ?

Yep, I would expect the testbench does wait until elements are available in shadow DOM. I will check how the selector method works in testbench. 

By the way, I didn't run this test several times. Maybe it's worth to do it in order to be sure it's fixed.

mshabarov:
> I wonder if this type of things should be fixed instead by making sure that testbench works with changes related to the embedded web component. Now that I look at this test, it seems like that is probably broken/not supported ?

After looking into the `TestBenchElement` code and taking into account that the other places, where the shadow DOM of the embedded component is being queried for elements (like this one `TestBenchElement content = themedComponent.$("div").first();`), work stable, i.e. without any random failures, I don't think the testbench code is kind of broken in respect to embedded components.
Maybe it's a matter of 'waiting' lack in the `ElementQuery` implementation, but for me it's maybe not that necessary, because you can always use `isPresentInShadowRoot` if the test is faster that the browser.

pleku (approved):

PR #10380, opened 2021-03-23
chore: Add missing import and header

mshabarov:


denis-anisimov:
Don't think that this is necessary : this is a test class.
And PR validation passes this.

But anyway.

denis-anisimov (approved):

mshabarov:
Snapshot build fails, because of missed import for class reference in javadoc.
I will try to copy that rule to PR validation configuration as well.

PR #10387, opened 2021-03-24
fix: Prevent NPE when writing field value to bean

knoobie:
Forward port vaadin/framework#12230


knoobie:
@pleku I tried to create a test for this change, but couldn't get it to work without adding a lot of code.. this doesn't felt reasonable for this, that's why I resigned from committing it. After this is merged, I can create a cherry pick pr with all binder related changes to pick to 2.6

pleku:
I kind of understand the situation, but it is still a bit unfortunate that the code is not easily testable 😬 

This change is a no-brainer though and could have been picked by static code analysis too.

pleku (approved):

pleku:
> After this is merged, I can create a cherry pick pr with all binder related changes to pick to 2.6

That would be great, thanks.

knoobie:
@pleku can you trigger a rebuild? Looks kinda unrelated 😅

pleku:
> @pleku can you trigger a rebuild? Looks kinda unrelated 😅

✅

PR #10388, opened 2021-03-24
fix: Make shortcuts not break writing on inputs

### LINKED ISSUE #10362: Shortcut event delegate should prevent default only if the shortcut registration doesn't
The change introduced in #10264 prevents writing on a input/textfield/... when used together with a `Shortcuts.addShortcutListener`. Click shortcuts added via `ClickNotifier` have `allowBrowserDefault()` by default in them and thus those work.

This can be fixed by only calling `oldEvent.preventDefault()` in the JS execution when the shortcut listener has been configured with `allowBrowserDefault()`

pleku:
Only passes the delegated event forward when it maps to the shortcut key.
For the shortcut key, preventDefault() is the default behavior and default
can be allowed with allowBrowserDefault() in the ShortcutRegistration.

Fixes #10362


pleku:
Will need to pick the test refactoring change before this, but seeing that it actually fixes the flaky test first.

caalador (approved):

Issue #8177, opened 2020-04-27
Redirect on logout not working correctly with longpolling


mkn1988:
When using Vaadin 14 with SpringSecurity and push transport long polling, the redirect on logout is not working correctly. After calling the spring logout url, another request from the browser to the last url the vaadin ui was using is fired. This request will be redirected by spring to the login page. So instead of displaying the logout success page, the browser displays the login page again.

See [my github](https://github.com/mkn1988/MWE_logout_longpolling) to recreate the bug. The Project is based on [Spring Secured Vaadin Starter](https://github.com/vaadin-learning-center/spring-secured-vaadin ). I changed the vaadin version to `14.1.25` and the spring version to `2.2.2`. I also added some simple static html pages for login and successful logout. Furthermore I added a simple logout button to the MainView.

After logging in and using the logout button, the browser should display `/login/logout.html`. But instead of the browser will display `/login/login.html` when using long polling as push transport mode. When using WEBSOCKET_XHR everything works as expected.

denis-anisimov:
Confirmed.

Thank you very much for providing the source code which can be used to reproduce the issue.

It looks like in Long polling case the session is closed invalidated : that's why the initial page is opened instead of the button action.

denis-anisimov:
The application works in expected way without any Push as well.
So this is really Long Polling only issue.
What happens under the hood: the configuration is done so that `/logout` URI is used to close the session and the redirects to a URL .
If `/logout` is changed to `/login/logout.html` (the final redirection URI) then everything is fine.

denis-anisimov:
The behavior is very specific to Chrome.
Safari doesn't have this issue.
The redirect which happens from `/logout`  to `/login/logout.html` is cancelled in Chrome.

alump:
We also have this issue on our application. Vaadin-Refresh stoped working when we moved from web socket to long poll.

denis-anisimov:
The reason of this issue is : session expired message comes after the redirect and Push is closed too late (which cancels the  redirect).
In FW8 there is a dedicated code specific to the `setLocation` method: if we set a new location then the client app should be terminated which closes the Push. So Push will be closed on `setLocation` before the redirect and it will happen as expected.
This is rather a workaround in fact because `setLocation` just calls the JS and if user  calls some custom JS which sets a new location then this issue still be there. So the fix will be just for `Page::setLocation` . But I don't see what we can do more for this.
[...]

PR #10389, opened 2021-03-24
chore: enable SessionCloseLogoutIT

denis-anisimov:


pleku (approved):

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/SessionCloseLogoutView.java
```diff
@@ -30,9 +31,7 @@
     public SessionCloseLogoutView() {
         NativeButton btn = new NativeButton("Logout!");
         btn.addClickListener(evt -> getUI().ifPresent(ui -> {
-
-            ui.getPage().executeJs(String.format("window.location.href='%s'",
-                    BaseHrefView.class.getName()));
+            UI.getCurrent().getPage().setLocation(BaseHrefView.class.getName());
```
mshabarov, line 34: So one could fix it this way? :)))
Does this mean that we now prohibit users to use `window.location.href` for redirect, but recommend to use `getPage().setLocation`?

mshabarov (approved):

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/SessionCloseLogoutView.java
denis-anisimov: https://github.com/vaadin/flow/issues/9647#issuecomment-805803796
https://github.com/vaadin/flow/issues/8177

PR #10398, opened 2021-03-24
test: Remove the test because it relies on Webpack startup time

### LINKED ISSUE #10101: Flakey test DevModeHandlerTest.start_twoInstances_secondInstanceUsesAnotherPort
Spot the flaky test in [this build](https://bender.vaadin.com/viewLog.html?buildId=214755&tab=buildResultsDiv&buildTypeId=Flow_Flow70_Flow60Snapshot)

mshabarov:
Removes DevModeHandlerTest::start_twoInstances_secondInstanceUsesAnotherPort because it relies on web pack startup time: if it starts fast enough, then the second handler instance may see that it occupies the port and just reuse that port, test will fail then. If webpack is starting slowly, the second handler sees that the port is not used by webpack and starts another instance of webpack. Then the tests passes.

Fixes #10101
Related-to #9061

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
@@ -582,45 +582,6 @@ public class DevModeHandlerTest {
     }
 
     @Test
-    public void startTwoInstances_secondInstanceStartsWhenWebpackStarts_secondInstanceUsesSamePort()
-            throws Exception {
-
-        // start the first instance
-        DevModeHandler handler = DevModeHandler.start(0, createDevModeLookup(),
-                npmFolder, CompletableFuture.completedFuture(null));
-
-        // remove the "singleton" instance to be able to start another one
-        removeDevModeHandlerInstance();
-
-        Assert.assertNotNull(
-                "Not expected null instance of started dev mode handler",
-                handler);
-
-        // Waits until first handler starts web pack server, which means that
-        // the port should have already been saved into the local file. Thus,
-        // the second handler should be able to read this file, reuse the
-        // port and not to start another web pack server.
-        while (!handler.checkWebpackConnection()) {
-            Thread.sleep(100);
-        }
-
-        // since the timeout is quite big the server port still should be
-        // available and the second instance should try to reuse it
-        DevModeHandler anotherHandler = DevModeHandler.start(0,
-                createDevModeLookup(), npmFolder,
-                CompletableFuture.completedFuture(null));
-
-        int firstPort = handler.getPort();
-
-        anotherHandler.join();
-
-        int secondPort = anotherHandler.getPort();
-
-        // the second instance was able to start with the same port value
-        Assert.assertEquals(firstPort, secondPort);
-    }
-
-    @Test
     public void start_serverPortDoesNotWork_throws() throws Exception {
         exception.expect(CompletionException.class);
         exception.expectCause(Matchers.instanceOf(IllegalStateException.class));
--- a/flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
@@ -582,8 +582,7 @@ public class DevModeHandlerTest {
     }
 
     @Test
-    @Ignore("Flaky test, see #10101")
-    public void start_twoInstances_secondInstanceUsesAnotherPort()
+    public void startTwoInstances_secondInstanceStartsWhenWebpackStarts_secondInstanceUsesSamePort()
             throws Exception {
 
         // start the first instance
@@ -593,15 +592,23 @@ public class DevModeHandlerTest {
         // remove the "singleton" instance to be able to start another one
         removeDevModeHandlerInstance();
 
+        Assert.assertNotNull(
+                "Not expected null instance of started dev mode handler",
+                handler);
+
+        // Waits until first handler starts web pack server, which means that
+        // the port should have already been saved into the local file. Thus,
+        // the second handler should be able to read this file, reuse the
+        // port and not to start another web pack server.
+        while (!handler.checkWebpackConnection()) {
+            Thread.sleep(100);
+        }
+
         // since the timeout is quite big the server port still should be
         // available and the second instance should try to reuse it
         DevModeHandler anotherHandler = DevModeHandler.start(0,
                 createDevModeLookup(), npmFolder,
                 CompletableFuture.completedFuture(null));
-
-        while (handler.getPort() == 0) {
-            Thread.sleep(100);
-        }
 
         int firstPort = handler.getPort();
 
@@ -609,10 +616,8 @@ public class DevModeHandlerTest {
 
         int secondPort = anotherHandler.getPort();
 
-        // the second instance was able to start with another port value
-        // even though the first port number is not bound to webpack server
-        // instance
-        Assert.assertNotEquals(firstPort, secondPort);
+        // the second instance was able to start with the same port value
+        Assert.assertEquals(firstPort, secondPort);
     }
 
     @Test
--- a/flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/DevModeHandler.java
@@ -384,7 +384,14 @@ public final class DevModeHandler implements RequestHandler {
         return true;
     }
 
-    private boolean checkWebpackConnection() {
+    /**
+     * Checks the status of connection with Webpack dev server.
+     * 
+     * @return <code>true</code> if connection established, <code>false</code>
+     *         otherwise.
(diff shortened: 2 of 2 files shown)


flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
```diff
@@ -582,8 +582,7 @@ public void start_twoTimes_onlyOneHandlerInstanceIsCreated() {
     }
 
     @Test
-    @Ignore("Flaky test, see #10101")
-    public void start_twoInstances_secondInstanceUsesAnotherPort()
+    public void startTwoInstances_secondInstanceStartsWhenWebpackStarts_secondInstanceUsesSamePort()
```
denis-anisimov, line 585: This change effectively removes the `start_twoInstances_secondInstanceUsesAnotherPort` test.

`startTwoInstances_secondInstanceStartsWhenWebpackStarts_secondInstanceUsesSamePort` has completely different semantic which looks quite similar to the existing test `should_reuseWebpackPort_AfterRestart`.
I'm not sure whether it's worth to have such test or not but the PR should fix the `start_twoInstances_secondInstanceUsesAnotherPort` instead introducing totally different test.

`startTwoInstances_secondInstanceStartsWhenWebpackStarts_secondInstanceUsesSamePort` checks that the second server reuses the same port.

`start_twoInstances_secondInstanceUsesAnotherPort` checks that it is possible to start second web pack server if there is already one started.

These are two absolutely different things.

The problem with `start_twoInstances_secondInstanceUsesAnotherPort` : `DevModeHandler` is  a singleton inside JVM and it's technically not possible to test what it tests without hacks.

So to be able to test this :
- one instance is created.
- the singleton file is cleared via Reflection
- now it's possible to create another DevModeHandler instance (even though the previous one is not stopped and has webpack process running)
- the second instance is created
- Everything else is checking that this is really another instance : the web pack is not reused (which would be a bug since another JVM which starts `DevModeHandler` may not reuse the same web pack dev server).

This is the test and it's opposite to `startTwoInstances_secondInstanceStartsWhenWebpackStarts_secondInstanceUsesSamePort`.

I assume that the test fails because the port is not yet used properly by the first web pack dev server
So may be it will be fixed via moving wait code :
```
while (handler.getPort() == 0) {
            Thread.sleep(100);
        }
```

right after the first handler start

denis-anisimov (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
mshabarov: Yes, I agree that `startTwoInstances_secondInstanceStartsWhenWebpackStarts_secondInstanceUsesSamePort` is another test with different semantics.

Seems I didn't get the meaning of `start_twoInstances_secondInstanceUsesAnotherPort`: _checks that it is possible to start second web pack server if there is already one started._

The test fails on CI because of the second instance reuses the port ('Reuse port' messages in the logs). First instance writes to file faster, than the second one reads the file. Thus, I'm not sure that moving that block and waiting for port would improve the timing issue.

To be able to start another instance of webpack and not reusing the existing one, the first instance should 'wait' for the second one to read the file. I don't know how to do it without hacking the `DevModeHandler` code.

But, yes, my changes duplicates the test `should_reuseWebpackPort_AfterRestart`, so they are not needed.

I will think more how to keep the existing semantics and fix the race between instances.

denis-anisimov: >The test fails on CI because of the second instance reuses the port ('Reuse port' messages in the logs). First instance writes to file faster, than the second one reads the file. 

Sounds like self-contradictory statement.
If the second instance is not able to read the file which is written by the first instance then how it may reuse the port ?
How it may reuse anything which is not simply known yet ? 

I don't recall details but every instance should check whether the port is free regardless of the presence the file.
And I think the first instance hasn't been able to occupy the port YET (the port is not occupied before it's written to the file).
And then the second instance reads the file, find the port and checks whether the port is free. And since it's free it uses it.
This is my assumption.

But this test may be just removed if it's not possible to make it working properly.

mshabarov: > Sounds like self-contradictory statement.
If the second instance is not able to read the file which is written by the first instance then how it may reuse the port ?
How it may reuse anything which is not simply known yet ?

No contradictions.
What happens on CI (according to logs and code analysis): the first instance is fast enough and it get a new port and writes it to a file BEFORE the second instance tries to read the port from file 
`port = getRunningDevServerPort(npmFolder);`
So it reads the file and gets the port => reuses it and does start a new webpack => assertion fails, because ports are the same.

Locally, the second instance ALWAYS tries to read the file BEFORE the first one writes to it. So it decided to start a new webpack. So, both instances tries to read the file and they both get a new port BEFORE any of them actually write to file.

denis-anisimov: The statement is contradictory since the second sentence `First instance writes to file faster, than the second one reads the file`  doesn't mention local execution. And the first says that the port is reused on CI , the second says that the port is not even read (having CI as a context).

Doesn't matter.

Apparently I missed that preconditions of the test: it assumes that the first instance is written the port into the file
but it's not yet started (not ready to handle requests).
The second instance should read the written file , get the port number from there and then check the connection.
If the connection doesn't work ( means that the first web pack server is not yet ready) another port should be chosen and the second instance should be started with another port.

For me locally the file is written by the first server, the file is read by the second handler. 
And locally the first web pack is not yet started indeed.
So the second handler tries to use the port and fails. So it chooses another one.

Apparently on CI the first web pack server is fast enough to start . So when the second instance reads the port it's already functional and it's just reused.

As a result : I don't see how the web pack dev server may be slow down to guarantee that the second won't be able to establish the connection.
So : I suggest to remove the test.

mshabarov: >I don't recall details but every instance should check whether the port is free regardless of the presence the file.
And I think the first instance hasn't been able to occupy the port YET (the port is not occupied before it's written to the file).
And then the second instance reads the file, find the port and checks whether the port is free. And since it's free it uses it.

Hmmm, I couldn't find the place where it's checking whether the port is free. Is the port is present in the file, then it only checks the connection with webpack and reset the watchdog.

But if you say it is supposed to work as you described, and the current logic is different, then maybe there is something to fix in the `doStartDevModeServer`?

mshabarov: > The second instance should read the written file , get the port number from there and then check the connection.
If the connection doesn't work ( means that the first web pack server is not yet ready) another port should be chosen and the second instance should be started with another port.
For me locally the file is written by the first server, the file is read by the second handler.
And locally the first web pack is not yet started indeed.
So the second handler tries to use the port and fails. So it chooses another one.
Apparently on CI the first web pack server is fast enough to start . So when the second instance reads the port it's already functional and it's just reused.

Great explanations! Thanks! So your machine behaves as you expected when the webpack is not started while the second instance tries to use the port. My machine passes the test because of another code branch/block executed (second instance doesn't see the file at all). CI behaves in a third way :)))

So, yeah. This test rely on non-deterministic behaviour that we cannot predict.
I agree to remove the test, because see no way to make it pass 100%.

denis-anisimov: `checkWebpackConnection` effectively checks that port is occupied .
In `2.x`  this check is almost the same as check for free port: in fact it checks that HTTP connection can be established.
In the `master`  it's more complicated but it still first checks whether the port is free or not.
More complicated check may be the reason why the test became flaky in the `master`.

mshabarov: Yes, `master` checks connection + manifest, `2.x` check only connection.

`master` and `2.x` have basically the following differences:

2.x: prepareConnection("/", "GET").getResponseCode();
master: prepareConnection("/manifest.json", "GET");

Maybe this one causing successful (returns `true`) `checkWebpackConnection` for `master`.

By the way, I double checked master behaviour in CI and found this record:
`INFO com.vaadin.flow.server.DevModeHandler - Reusing webpack-dev-server running at http://localhost:35765`

which proves your thought about that the webpack is fast enough, and the second instance sees that it is already started.

Well. I can dig into why `prepareConnection("/manifest.json", "GET");` leads to successful  `checkWebpackConnection` check in master. But not sure I have to change it in favor of fixing this test.

denis-anisimov: Just remove the test.
It's not able to work as it's supposed to work and there is no way to make it reliable.

denis-anisimov (approved):

mshabarov:
Yes, I'll pick it manually.

PR #10399, opened 2021-03-24
CP: Binder Changes to 2.6

knoobie:


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -250,6 +250,32 @@
...
+         * As of version 6.0, when a converter is used on a binding and the user
+         * input value is modified by the converter, the value from the converter
+         * is applied back to the input. It is possible to control this behavior
+         * with this API.
+         *
+         * @see BindingBuilder#withConverter(Converter)
+         * @see BindingBuilder#withConverter(SerializableFunction, SerializableFunction)
+         * @see BindingBuilder#withConverter(SerializableFunction, SerializableFunction, String)
+         *
+         * @param convertBackToPresentation
```
pleku, line 267: It should be highlighted here in javadocs that the default behavior in 2.x / v14 series is still that things are not written back (default value is `false`). Now that I think about this, it could say instead "as of version 6.0" could be removed or maybe clarified that "as of Flow 6.0 / Vaadin 19 .."

knoobie: What do you think about this?

```
         * The default behavior (do not convert back to presentation) changes to the
         * opposite as of Vaadin 19 / Flow 6.0, when a converter is used on a binding
         * and the user input value is modified by the converter, the value from the
         * converter is applied back to the input. It is possible to control this
         * behavior with this API.
```

pleku: That is great.

knoobie: Done! Gonna have to catch a train and have bad connection there, if you need a quick change just force push it :) Should be back by the evening if there is more to change

pleku (approved):

pleku:
New build triggered.

PR #10417, opened 2021-03-25
fix: remember original state after LOADING (#10099)

joheriks:
Added method loadingFailed transitioning to CONNECTION_LOST if it
is the last active loading. Also removed the exposed
Vaadin.Flow.clients.TypeScript.loadingStarted / loadingFinished pair, as
blocking TB is anyway not working for multiple ongoing backend calls
(and it would be better handled by register ConnectionState as a Flow
client with isActive flag).


Haprog (approved):

Haprog:
For reference. This is a cherry-pick of https://github.com/vaadin/flow/pull/10099.

PR #10424, opened 2021-03-26
fix: make spring test not dependent on platform

caalador:
Update spring test dependencies so
it doesn't bring in a dependency to platform.


taefi (approved):

flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/MiscSingleModuleTest.kt
```diff
@@ -255,7 +255,9 @@ class MiscSingleModuleTest : AbstractGradleTest() {
             }
             
             dependencies {
-                implementation('com.vaadin:vaadin-spring-boot-starter:20.0-SNAPSHOT')
+                implementation('com.vaadin:flow:$flowVersion')
+                implementation('com.vaadin:vaadin-spring:17.0-SNAPSHOT')
```
knoobie, line 259: Wouldn't it be better to move this hard coded version to the top as well for easier discoverability?

caalador: Done. 
Hopefully we could get rid of this in some way at some point.

taefi (dismissed):

taefi (approved):

PR #10432, opened 2021-03-26
set deploy-file config url to flow.release.repo.url property

taefi:
The property is defined in flow/pom.xml and
points to the correct prerelease repo.

### DIFF
--- a/flow-plugins/flow-gradle-plugin/pom.xml
+++ b/flow-plugins/flow-gradle-plugin/pom.xml
@@ -134,7 +134,7 @@
                         <configuration>
                             <packaging>jar</packaging>
                             <generatePom>true</generatePom>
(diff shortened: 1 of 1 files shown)


flow-plugins/flow-gradle-plugin/pom.xml
```diff
@@ -134,7 +134,7 @@
                         <configuration>
                             <packaging>jar</packaging>
                             <generatePom>true</generatePom>
-                            <url>${project.distributionManagement.repository.url}</url>
+                            <url>${flow.release.repo.url}</url>
```
caalador, line 137: It doesn't match the snapshot.target in bender.

Perhaps we should see that we have the jar in the correct place and have it use the deploy:deploy target instead of the deploy:deploy-file target.

PR #10438, opened 2021-03-26
fix: Upload did not respond with correct content length for large files

### LINKED ISSUE #10392: Upload of big files: StartedEvent and ProgressUpdateEvent report -1 file size
### Description
When uploading large files, eg 2Gb, the StartedEvent and the ProgressUpdateEvent report -1 as contentLength.  When the filesize is small, the contentlength is correct.
Our application needs to accept those huge uploads but must implement a server-side free-space check before accepting the file, so thie -1 is very inconvenient.

### Live Demo

https://glitch.com/edit/#!/vaadin-upload-issue

### Steps to reproduce
```
        Upload upload = new Upload(new Receiver() {
            @Override
            public OutputStream receiveUpload(String s, String s1) {
                return new OutputStream() {
                    @Override
                    public void write(int b) throws IOException {
                        //noop, good enough to demonstrate the issue
                    }
                };
            }
        });
        upload.addProgressListener(new ComponentEventListener<ProgressUpdateEvent>() {
            @Override
            public void onComponentEvent(ProgressUpdateEvent event) {
                System.out.println("ProgressUpdateEvent "+event.getReadBytes() +" / "+event.getContentLength());
            }
        });

        upload.addStartedListener(new ComponentEventListener<StartedEvent>() {
            @Override
            public void onComponentEvent(StartedEvent event) {
                System.out.println("StartedEvent "+event.getFileName() +" of "+event.getContentLength());
            }
        });
        body.add(upload);
```

Note that is noticed that in the the javascript it does know the filesize correctly. Today i have this ugly workaround in place, which works:
```
    private void determineFileSizeThroughJsOfLastFile() {
        try {
            PendingJavaScriptResult jsResult = getElement().executeJs("if (this.files == null || this.files.length ==0) {return -2} else { return this.files[this.files.length-1].size}");
            jsResult.then(result -> {
                long length = Double.valueOf(((JreJsonNumber)result).getNumber()).longValue();
                System.out.println("determineFileSizeThroughJsOfLastFile file size is "+length );
            });
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
```

### Expected outcome
The actual file size should be there, not -1.

### Actual outcome
-1
### Browsers Affected
Seen on Chrome, not tested further.

### Version
Vaadin Flow 14.4.0

knoobie:
fixes #10392

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
@@ -575,12 +575,16 @@ public class StreamReceiverHandler implements Serializable {
 
     /**
      * The request.getContentLength() is limited to "int" by the Servlet
-     * specification (<= 3.0). To support larger file uploads we check if
+     * specification (&lt;= 3.0). To support larger file uploads we check if
      * the given {@link VaadinRequest} is of type {@link VaadinServletRequest}
      * and returns the appropriated value of
      * {@link VaadinServletRequest#getContentLengthLong()}, otherwise
      * manually evaluate the Content-Length header which can contain
      * long values.
+     *
+     * @param request
+     *            request to get content length from
+     * @return content length in bytes
      */
     protected long getContentLength(VaadinRequest request) {
         try {
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
@@ -49,6 +49,7 @@ import com.vaadin.flow.server.UploadException;
 import com.vaadin.flow.server.VaadinRequest;
 import com.vaadin.flow.server.VaadinResponse;
 import com.vaadin.flow.server.VaadinSession;
+import com.vaadin.flow.server.VaadinServletRequest;
 import com.vaadin.flow.server.communication.streaming.StreamingEndEventImpl;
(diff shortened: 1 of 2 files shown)


knoobie:
@pleku Can you please take a look at the build next week? I'm out of options here.. I kinda feel bullied by the agent 😅

denis-anisimov:
Sorry, didn't notice this PR: please assign yourself to the ticket if you are working on it to avoid collisions.

Thanks for your contribution but I would better use existing API instead of custom header parsing: https://github.com/vaadin/flow/pull/10506.

knoobie:
@denis-anisimov  Sorry, I can't assign myself :) I wanted to update the `VaadinRequest` as well, but I wasn't sure if it is worth it to change such a crucial part of the API just for this fix - when there was already a custom method `getContentLength` in `StreamReceiverHandler` which could be used instead of changing the API.

denis-anisimov:
>Sorry, I can't assign myself :) 

OK, got it :(

Don't know then how to avoid collisions.......

>when there was already a custom method getContentLength

yeah, I missed that there is already such method since the line which required modification didn't use it.

But anyway, the solution makes sense but since we have now `default` methods in interfaces it's possible to do non-breaking changes. It's better to have shared `getContentLengthLong` which may be used from anywhere instead of doing custom impl for it in every place where it can be needed.

knoobie:
I'm on your side :) I wasn't sure if using `default` methods would be enough to retain binary compatibility with other implementations of `VaadinRequest` that's why I used this approach.

PR #10444, opened 2021-03-29
Do not use the request charset when reading a file from the classpath

Artur-:


denis-anisimov (commented):
I'm bit worried that this changes existing methods: this is fine in the `master` but it may be an issue for backporting....
The methods not public in fact so it's not API in that sense: they should not be used outside of the Flow.

Artur-:
I think it is fine as it is a package private helper, especially as I cannot imagine when you would need the functionality to read a file using an encoding passed from the browser.

denis-anisimov (approved):

PR #10507, opened 2021-03-30
fix: use warn instead of trace for WC exporters discovering

### LINKED ISSUE #9855: Logging in WebComponentExporterUtils should be using WARN and not TRACE
Say that you create this
```
public static class HelloWorldExporter extends WebComponentExporter<HelloWorld> {

        protected HelloWorldExporter() {
            super("hello-world");
        }

        @Override
        protected void configureInstance(WebComponent<HelloWorld> webComponent, HelloWorld component) {

        }

    }
```
The intent is pretty clear: you want `HelloWorld` to be exported as `<hello-world>`
It does not work and no warnings are logged. It is unclear what to do next.

`WebComponentExportedUtils.isEligible` notices the problem and logs, with TRACE level
```
                    "Class {} has no public no-argument constructor, and won't be instantiated as a '{}' by default",
```
 
This and the other errors in `isEligible` should be at least WARN so that you see what is wrong.

denis-anisimov:
fixes #9855

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/webcomponent/WebComponentExporterUtils.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/webcomponent/WebComponentExporterUtils.java
@@ -91,7 +91,7 @@ public final class WebComponentExporterUtils {
                 .filter(ctor -> ctor.getParameterCount() == 0).findFirst();
         if (!constructor.isPresent()
                 || !Modifier.isPublic(constructor.get().getModifiers())) {
-            getLogger().trace(
+            getLogger().warn(
                     "Class {} has no public no-argument constructor, and won't be instantiated as a '{}' by default",
                     clazz.getName(),
                     WebComponentExporter.class.getSimpleName());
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/webcomponent/WebComponentExporterUtils.java
```diff
@@ -91,15 +91,15 @@ private static boolean isEligible(Class<?> clazz) {
                 .filter(ctor -> ctor.getParameterCount() == 0).findFirst();
         if (!constructor.isPresent()
                 || !Modifier.isPublic(constructor.get().getModifiers())) {
-            getLogger().trace(
+            getLogger().warn(
```
mshabarov, line 94: Should those two
`Class {} is abstract, and won't be instantiated as a '{}' by default`
`Class {} is not public, and won't be instantiated as a '{}' by default`
also come with WARN level?

mshabarov: Yes, I talk about those occasional mistakes, when somebody has only an abstract class and gets no result. But from the other side: if we enable it with WARN, it would bloat the logs with false warning.

mshabarov (approved):

PR #10521, opened 2021-03-31
Add support for parsing query string in QueryParameters

Artur-:
This is a prerequisite for #10431 split out to a separate PR for easier review

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -16,6 +16,8 @@
 package com.vaadin.flow.router;
 
 import java.io.Serializable;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -125,6 +127,12 @@ public class QueryParameters implements Serializable {
         }
         String param = paramAndValue.substring(0, index);
         String value = paramAndValue.substring(index + 1);
+        try {
+            value = URLDecoder.decode(value, "utf-8");
+        } catch (UnsupportedEncodingException e) {
+            throw new RuntimeException(
+                    "Unable to decode parameter value: " + value, e);
+        }
         return Arrays.asList(param, value);
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -16,14 +16,19 @@
 package com.vaadin.flow.router;
 
 import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+
+import org.apache.http.NameValuePair;
+import org.apache.http.client.utils.URLEncodedUtils;
 
 /**
  * Holds query parameters information.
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -97,6 +102,35 @@ public static QueryParameters simple(Map<String, String> parameters) {
...
+     * @return query parameters information
+     */
+    public static QueryParameters fromString(String queryString) {
+        return new QueryParameters(parseQueryString(queryString));
+    }
+
+    private static Map<String, List<String>> parseQueryString(
+            String queryString) {
+        List<NameValuePair> params = URLEncodedUtils.parse(queryString,
+                StandardCharsets.UTF_8);
```
pleku, line 119: The only place where apache.http.client is used is in devmode for downloading node - that has been copied from frontend-maven-plugin. And that code should be removed from `flow-server` to its own place so that would not be leaking to user applications. But this is not your fault.

But if you introduce a further dependency to it, then it will be more difficult for us to make this refactoring.
The code not the same exactly, but close to what `Location` does too to query parameters. Would it be possible to reuse-modify the code there instead ?

pleku (changes requested):
(just changing status to show that this is not waiting for review)

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -97,6 +102,35 @@ public static QueryParameters simple(Map<String, String> parameters) {
...
     }
 
+    /**
+     * Creates parameters from a query string.
+     *
+     * @param queryString
+     *            the query string
+     * @return query parameters information
+     */
+    public static QueryParameters fromString(String queryString) {
```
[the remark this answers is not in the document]
Artur-: Added mentioned that it is needed by #10431

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
Artur-: I will move the code from `Location.parseParams` to replace this. Would have helped a lot if it had been in the correct class earlier. 

Sadly the implementation is partly broken so it also needs to be fixed at the same time

Artur-: Overall I don't understand why we have this generic util methods implemented in the framework in a slightly broken way when there are perfectly fine and tested util libraries out there that you can use.

The current method needed two fixes to be compatible with what `HttpServletRequest.getParameters` returns:
1. It needs to decode values
2. A parameter like "?hello" should map to an empty string and not no value at all like the old method did

Hopefully this does not break something else then

pleku: > Overall I don't understand why we have this generic util methods implemented in the framework in a slightly broken way when there are perfectly fine and tested util libraries out there that you can use

The problem is the broken code in the code base. We should not try reinvent the wheel, but also not bring in libraries only to depend on a subset of the code there. It is a balancing act.

Artur-: If you want to avoid external dependencies, there is always the possibility of rebasing the given class and including it directly in the project. Then it is possible to update it etc for new releases, there are no conflicts and no dependencies either

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -97,6 +99,66 @@ public static QueryParameters simple(Map<String, String> parameters) {
                         entry -> Collections.singletonList(entry.getValue())));
     }
 
+    /**
+     * Creates parameters from a query string.
+     *
```
pleku, line 104: I think the code is fine. But as the string can come from the user, this parsing might be a subject for DoS exploit, but I think the computations done inside the stream-map-collect are so simple that the query string limit of 1024 characters should not make it exploitable. Do you agree ?

I would consider adding some clarification to the method javadocs that this is code here is not checking for the length of the given query string - it is the responsibility of the caller.

pleku (changes requested):
One question and suggestion.

The changes in this PR are fine, but when I look at the code flow (which this PR doesn't change, just the implementation), it does smell a bit. It seems like the code in the framework could be (later) improved to 
1) only parse the query parameters when needed
2) cache the query strings for the ongoing request, now parse again and again
3) see if the query string parsing code from the http request implementation could be reused, like from `AtmosphereRequest` since it handles both 1+2 from above
I can create an issue for this after this is merged.

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
pleku: Yes I think that is the least bad solution, to copy-paste the code and update it manually, in case the code cannot be / is not released as a standalone module.

Artur-: So should we do it here? In that case I would revert to the original version more or less to use the lib and then make a separate PR about including it directly in the project instead of as a dependency

Artur-: Because of the internal dependencies, it might not make sense to copy paste... It would lead to extracting the whole `httpcore` jar into the project

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
Artur-: Where would the DoS exploit possibility come from?

pleku: Like discussed separately, from using this code to parse a string that originates from client and it is very very long and because it is for some reason limited to 1024. This comment is enough for now.

pleku (approved):

PR #10526, opened 2021-03-31
feat!: endpoint access is denied by default

### LINKED ISSUE #10328: change the default Endpoint access to "deny all"
When adding a Fusion `@Endpoint` to a Vaadin app, as a developer I want the default access level to the new endpoint to be 'deny all' so that the applicaiton is secure by default and I do not accidentially expose access to the backend to unintended users. In order to enable access I want to explicitly use one of the [access control annotations](https://vaadin.com/docs-beta/latest/fusion/security/configuring/#other-security-options): `@RolesAllowed` to allow access to selected roles, `@PermitAll` to allow access to any logged-in (non-anonymous) users, or `@AnonymousAllowed` to allow unrestricted access to _everyone_. As in Vaadin 15-19 I want to set these annotations either on an entire endpoint class or on individual public methods of that class.

**NOTE: this is a breaking change. In Vaadin 15-19 the default endpoint access is "any logged-in (non-anonymous) user"**

joheriks:
Fixes #10328

### DIFF
--- a/fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
+++ b/fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
@@ -136,7 +136,7 @@ public class VaadinConnectAccessChecker {
     private boolean cannotAccessMethod(Method method,
             HttpServletRequest request) {
         return requestForbidden(request)
-                || entityForbidden(getSecurityTarget(method), request);
+                || !entityAllowed(getSecurityTarget(method), request);
     }
 
     private boolean requestForbidden(HttpServletRequest request) {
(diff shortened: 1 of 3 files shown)


haijian-vaadin:
The code change looks good to me. Just one remind from Leif regarding the error message for access denied:
For production mode, the current message is fine.
For dev mode, in case a dev forgets to add any annotations, we could show an error message like
"The access is denied by default, you can use the following annotations to the endpoint to allow access"

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
```diff
@@ -174,20 +174,19 @@ private boolean requestForbidden(HttpServletRequest request) {
...
+                && !entity
+                .isAnnotationPresent(DenyAll.class);
     }
 
     private boolean roleAllowed(RolesAllowed rolesAllowed,
             HttpServletRequest request) {
-        if (rolesAllowed == null) {
-            return true;
-        }
-
```
knoobie, line 190: I think this should stay and changed to `return false` otherwise you could run into NPE later on with `rolesAllowed.value()`.

joheriks: It's guarded by the caller a few lines above (`rolesAllowed != null && roleAllowed(rolesAllowed, request)`). I'd prefer non-nullable parameter by default (like `request`).

knoobie: Ahh thanks! I was looking at the original code, where it's called without the guard.

joheriks:
> The code change looks good to me. Just one remind from Leif regarding the error message for access denied:
> 
> * For production mode, the current message is fine.
> * For dev mode, in case a dev forgets to add any annotations, we could show an error message like
>   "The access is denied by default, you can use the following annotations to the endpoint to allow access"

The message shown in dev mode is now more detailed, nudging the dev towards using the annotations.

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessCheckerTest.java
```diff
@@ -510,4 +527,51 @@ public void test() {
...
+            assertTrue(accessDeniedMessage.contains(PermitAll.class.getSimpleName()));
+            assertTrue(accessDeniedMessage.contains(RolesAllowed.class.getSimpleName()));
+            assertTrue(accessDeniedMessage.contains(AnonymousAllowed.class.getSimpleName()));
+        } finally {
+            CurrentInstance.clearAll();
+        }
+    }
+
+    @Test
+    public void should_notShowHelpfulMessage_When_accessDeniedInDevMode()
```
haijian-vaadin, line 557: ProductionMode?

haijian-vaadin (changes requested):

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessCheckerTest.java
joheriks: A copy-paste typo, fixed.

haijian-vaadin (approved):

PR #10534, opened 2021-03-31
refactor: Split public resources according to security context

Artur-:


pleku (approved):

pleku:
As the removed class was not released in 6.0 this is not a breaking change.

Issue #8885, opened 2020-08-25
Maven plugin: add support for cleaning npm-related folders


mvysny:
Vaadin 14.3.4

Implement something akin to the [Vaadin Gradle Plugin's vaadinClean](https://github.com/vaadin/vaadin-gradle-plugin) task. Sometimes the node_modules may become corrupted or npm install may not run for mysterious reasons. While Vaadin team figures out how to tackle these issues properly, as a workaround it would be good to have a task that would delete the npm-related files and folders: node_modules, package*.json and webpack.*.js (and pnpm-related files; and TS-related files as well).

garaei:
It is really a need to  have a task that would perform the Vaadin Dance and delete the npm-related en pnpm-related files and folders.

alexweirig:
+1

I totally agree

marcushellberg:
Deleting `package.json` would also remove info about installed npm dependencies in Fusion projects, please don't delete that.

Artur-:
Presumably if you take a start.vaadin.com project, run it and/or build it for production and then run clean and this clean target, you should be back with the downloaded project?

mvysny:
I agree with Artur. The vaadin clean goal should be there to help the developer when Vaadin doesn't start for some reason. In this case, preserving custom-added packages is secondary; primary is to get the dev env up and running. You can always checkout package.json from git later on.

marcushellberg:
Should it also reset `pom.xml` in that case?

Touching `package.json` or `pom.xml` from the clean target would be very unexpected behavior in my opinion.

Artur-:
Yeah I did not at all say "preserving custom-added packages is secondary". You should not delete `package.json` as it is not an autogenerated file - it is included in the generated project. It would be the same as removing stuff from the `<dependencies>` section from pom.xml.

The lock files, `node_modules` and `*.generated.js` are generated and can be removed. It is unlikely that removing `webpack.generated.js` will help with many problems though. Removing `node_modules` and the package lock file should be enough to reinstall npm dependencies from scratch

mvysny:
> You should not delete package.json as it is not an autogenerated file - it is included in the generated project. It would be the same as removing stuff from the <dependencies> section from pom.xml.

[...]

PR #10542, opened 2021-04-01
feat: Add gradle plugin

caalador:
Add gradle plugin into the plugins-module.


flow-plugins/flow-gradle-plugin/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+| Vaadin Gradle Plugin version | Supports |
+|------------------------------|----------|
+| -                            | Vaadin 13 and lower are unsupported |
+| 0.6.0 and lower              | Vaadin 14.1.x LTS and lower |
+| 0.7.0                        | Vaadin 14.2.x LTS |
+| 0.14.3.7                     | All other Vaadin 14 LTS versions (recommended) |
+| -                            | Vaadin 15 and 16 are unsupported |
+| 0.17.0.1                     | Vaadin 17 and higher |
+| 0.20.0.0.alpha3              | Vaadin 20 and higher, experimental support for Vaadin 19 |
```
pleku, line 45:
```suggestion
| 0.20.0.0.alpha3              | experimental support for Vaadin 19 and 20 |
| 20.0.0                            | Vaadin 20 and higher |
```

flow-plugins/flow-gradle-plugin/README.md
```diff
@@ -0,0 +1,283 @@
...
+   then packages everything into the war/jar archive. Automatically calls the
+   `vaadinPrepareFrontend` and `vaadinBuildFrontend` tasks.
+
+*Note* (after you built the project in production mode): In order to prepare the project
+setup back to development mode, you must run `./gradlew vaadinPrepareFrontend`
+with the `productionMode` effectively set to false (e.g. by ommitting the `-Pvaadin.productionMode` flag).
+
+All tasks provided by the plugin:
+
+* `vaadinClean` will clean the project completely and removes JavaScript packaging-related
```
pleku, line 63: Just commenting that we don't have this for the maven plugin but should make it unified.

flow-plugins/flow-gradle-plugin/README.md
```diff
@@ -0,0 +1,283 @@
...
+  need to be invoked explicitly. Note: this task is only triggered automatically if `productionMode` is set to true.
+
+## Configuration
+
+To configure the plugin, you can use the following snippet in your `build.gradle` file:
+
+`build.gradle` in Groovy:
+```groovy
+vaadin {
+  pnpmEnable = true
```
pleku, line 84: I don't get this - how does `pnpmEnable` relate to enabling vaadin gradle plugin? Or is it surplus ? Should it be production mode instead ?

flow-plugins/flow-gradle-plugin/README.md
```diff
@@ -0,0 +1,283 @@
...
+}
+```
+
+All configuration options follow. Note that you **RARELY** need to change anything of the below.
```
pleku, line 88: I disagree, production mode is something that you should be aware of.

flow-plugins/flow-gradle-plugin/README.md
```diff
@@ -0,0 +1,283 @@
...
+then produce the `META-INF/VAADIN/config/flow-build-info.json` file with the `"productionMode": true` setting,
+which will then cause Vaadin to start in production mode at runtime. To revert this setting
+back to false, simply recompile your project without the `-Pvaadin.productionMode` switch.
+
+# Developing The Plugin
+
+Please read the Gradle Tutorial on [Developing Custom Gradle Plugins](https://docs.gradle.org/current/userguide/custom_plugins.html)
+to understand how Gradle plugins are developed.
+
+The main entry to the plugin is the `VaadinPlugin` class. When applied to the project, it will register
```
pleku, line 217: I wonder if this is still valid, will see soon

flow-plugins/flow-gradle-plugin/README.md
```diff
@@ -0,0 +1,283 @@
...
+all necessary tasks and extensions into the project.
+
+Use Intellij (Community is enough) to open the project.
+
+## Branches
```
pleku, line 222: This is not up to date, should just remove it

flow-plugins/flow-gradle-plugin/pom.xml
```diff
@@ -0,0 +1,179 @@
...
+    <parent>
+        <groupId>com.vaadin</groupId>
+        <artifactId>flow-plugins</artifactId>
+        <version>7.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>flow-gradle-plugin</artifactId>
+    <name>Flow Gradle Plugin</name>
+    <packaging>jar</packaging>
+
```
pleku, line 12: Description missing - please add with information on how this relates to `vaadin-gradle-plugin`

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinBuildFrontendTask.kt
```diff
@@ -0,0 +1,82 @@
...
+ * annotations defined in the classpath,
+ * * Copy resource files used by flow from `.jar` files to the `node_modules`
+ * folder
+ * * Install dependencies by running `npm install`
+ * * Update the [FrontendUtils.IMPORTS_NAME] file imports with the
+ * [com.vaadin.flow.component.dependency.JsModule] [com.vaadin.flow.theme.Theme] and [com.vaadin.flow.component.dependency.JavaScript] annotations defined in
+ * the classpath,
+ * * Update [FrontendUtils.WEBPACK_CONFIG] file.
+ *
+ * @since 2.0
```
pleku, line 39:
```suggestion
```

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
```diff
@@ -0,0 +1,241 @@
...
+import com.vaadin.flow.server.InitParameters
+import com.vaadin.flow.server.frontend.FrontendTools
+import com.vaadin.flow.server.frontend.FrontendUtils
+import com.vaadin.flow.server.frontend.installer.NodeInstaller
+import org.gradle.api.Project
+import org.gradle.api.plugins.JavaPluginConvention
+import org.gradle.api.tasks.SourceSetContainer
+import java.io.File
+
+public open class VaadinFlowPluginExtension(project: Project) {
```
pleku, line 28: Just commenting that there are a lot of things that are not configurable via maven here - are these configurations still valid for gradle ? Like deciding if bundle should be created or npm install run after updating the dependencies

pleku (changes requested):
Couple of comments + questions

flow-plugins/flow-gradle-plugin/README.md
caalador: I think it's from a time when npm was the default instead of pnpm so before v14.3?

flow-plugins/flow-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,221 @@
+/*
+ * Copyright 2018-2019 Devsoap Inc.
```
mstahv, line 2: I wonder what is left from the Devsoap version here in the build script that is worth copyright notice. Also, this is build script, maybe remove the license header altogether 🤔

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
mstahv: @mvysny You might be the best source to comment.

flow-plugins/flow-gradle-plugin/README.md
caalador: We should probably only have the development info in the README and the rest should live in the docs.

flow-plugins/flow-gradle-plugin/build.gradle
pleku: I missed this, but nor do I understand how build scripts would even have license header (similar to pom.xml) and we'll just drop the header

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
mvysny: Those settings were taken from Maven plugin 14.3.x's sources and they were most probably not cleaned up when adding support for Vaadin 20. We should find such settings and remove them, so that Gradle plugin has the same set of settings as Maven plugin does.

flow-plugins/flow-gradle-plugin/README.md
mvysny: True, we should reword this to "...apart from the production mode, there's rarely any need to change..." or similar. The point was to stop users from tinkering with the settings, shooting themselves to the foot.

flow-plugins/flow-gradle-plugin/README.md
mvysny: In such case I propose to add similar functionality to Maven plugin as requested here: https://github.com/vaadin/flow/issues/8885 . I have seen Vaadin Dance being helpful to users, and `vaadinClean` essentially performs the dance very quickly.

pleku: At least now the issue has a reference to here so it should be OK for now

flow-plugins/flow-gradle-plugin/README.md
taefi: As the structure of gradle plugin was not changed during the process of adopting its source code under flow project, I believe this is still valid.

pleku (changes requested):
Readme changes still missing, outdated content to be 🪓

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
taefi: I scanned all the setting properties defined in `VaadinFlowPluginExtension.kt` and noticed all of them are present either in `FlowModeAbstractMojo.java` or in `BuildFrontendMojo.java`, especially `runNpmInstall` and `generateBundle`. Although there may be no way to configure them even in the maven and always their default values are in use, I propose to keep them for now, and clean them up in a separate task for both maven and gradle plugins at the same time.

flow-plugins/flow-gradle-plugin/build.gradle
taefi: The license header is removed.

flow-plugins/flow-gradle-plugin/pom.xml
taefi: Added.

flow-plugins/flow-gradle-plugin/README.md
taefi: Done.

flow-plugins/flow-gradle-plugin/README.md
taefi: `## Configuration` heading is removed from README and will be added to the docs.

flow-plugins/flow-gradle-plugin/README.md
taefi: Removed from README.

flow-plugins/flow-gradle-plugin/README.md
```diff
@@ -0,0 +1,210 @@
...
+This applies to Visual Studio Code (VSCode) as well since it also uses Eclipse bits and BuildShip
+underneath - see [https://github.com/mvysny/vaadin14-embedded-jetty-gradle/issues/4](https://github.com/mvysny/vaadin14-embedded-jetty-gradle/issues/4)
+for more details.
+
+In order to run your project in production mode from your IDE, simply compile the
+project in production mode, e.g. by using `-Pvaadin.productionMode`. The plugin will
+then produce the `META-INF/VAADIN/config/flow-build-info.json` file with the `"productionMode": true` setting,
+which will then cause Vaadin to start in production mode at runtime. To revert this setting
+back to false, simply recompile your project without the `-Pvaadin.productionMode` switch.
+
```
pleku, line 147: Several topics from above here are too not related to developing the plugin ... but you can revisit those after working on the docs

pleku (approved):

PR #10553, opened 2021-04-02
fix: don't use ErrorHandler for IOExceptions in StreamReceiverHandler

### LINKED ISSUE #10351: TerminalErrorHandler triggered when Upload is cancelled
In a web application I am currently working on, I am using a terminalErrorHandler (`VaadinSession#setErrorHandler`).
I am also using the `Upload` component as well as a custom upload component based on it.
Now when I cancel the upload on the client side `StreamVaribale#streamingFailed` and the terminalErrorHandler get triggered.
In my opinion, the terminalErrorHandler should not be triggered in this case, as a failed upload is somehow expected and already handled by `StreamVariable#streamingFailed`.

While trying to find the cause, I have seen the the  `StreamReceiverHandler` handles an `UploadInterruptedException` ([here](https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java#L524-L536))  and this exception is not propagated to the terminalErrorHandler. However, this exception only occurs, if you cancel the upload from the server.
All other Exceptions, which includes the `EOFException` I get, when the client cancels the upload, are handled by the next `catch` block ([here](https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java#L537-L546)) and there the Exceptions is wrapped in an `UploadException` and thrown.

There is even a comment about that `throw new UploadException(e)` but I am not sure if thats the correct behavior, as an Exception is somehow expected in this case.
If in your opinion the exception should propagate to the terminalErrorHandler if it is not cause of an cancelled upload, then the solution might be to add a `if (!streamVariable.isInterrupted())` before the `throw`.
The `Upload` component triggers a `UploadAbort` event, when the client aborts the upload and `StreamVariable#isInterrupted` would then return true in this caseand as I said already handled by `StreamVariable#streamingFailed`.

1) Set a terminalErrorHandler like this: `vaadinSession.setErrorHandler(e -> System.out.println("Error"))`
2) Add a Vaadin Upload with a `FailedListener` to your application:
```
Upload u = new Upload();
u.addFailedListener(e -> {
	System.out.println("Upload Failed");
});
u.setReceiver(new MemoryBuffer());
```
3) Start an upload of a file and cancel the upload before it finishes.

The output in the console looks like this:

> Upload Failed
> Error

### Expected behavior
Only the `FailedListener` should be triggered

### Actual behavior
The terminalErrorHandler gets triggered too.

### Versions:
    - Vaadin / Flow version: 18.0.7

denis-anisimov:
fixes #10351

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
@@ -193,7 +193,7 @@ public class StreamReceiverHandler implements Serializable {
         } catch (Exception exception) {
             session.lock();
             try {
-                // Report not IO exceptions (which are mistakes) via
+                // Report other than IO exceptions, which are mistakes, via
                 // ErrorHandler
                 session.getErrorHandler().error(new ErrorEvent(exception));
             } finally {
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
@@ -189,7 +189,7 @@ public class StreamReceiverHandler implements Serializable {
             }
         } catch (IOException exception) {
             // do not report IO exceptions via ErrorHandler
-            getLogger().warn("IO Exception during file upload", exception);
+            getLogger().warn("IO Exception during file upload, fired as StreamingErrorEvent", exception);
         } catch (Exception exception) {
             session.lock();
             try {
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
@@ -187,9 +187,14 @@ public class StreamReceiverHandler implements Serializable {
                 success = handleMultipartFileUploadFromInputStream(session,
                         request, streamReceiver, owner);
             }
+        } catch (IOException exception) {
+            // do not report IO exceptions via ErrorHandler
+            getLogger().warn("IO Exception during file upload", exception);
         } catch (Exception exception) {
             session.lock();
             try {
+                // Report not IO exceptions (which are mistakes) via
+                // ErrorHandler
                 session.getErrorHandler().error(new ErrorEvent(exception));
             } finally {
                 session.unlock();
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
```diff
@@ -187,9 +187,14 @@ protected void doHandleMultipartFileUpload(VaadinSession session,
                 success = handleMultipartFileUploadFromInputStream(session,
                         request, streamReceiver, owner);
             }
+        } catch (IOException exception) {
+            // do not report IO exceptions via ErrorHandler
+            getLogger().warn("IO Exception during file upload", exception);
         } catch (Exception exception) {
             session.lock();
             try {
+                // Report not IO exceptions (which are mistakes) via
```
pleku, line 196:
```suggestion
                // Report other than IO exceptions, which are mistakes, via
```

flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
```diff
@@ -187,9 +187,14 @@ protected void doHandleMultipartFileUpload(VaadinSession session,
                 success = handleMultipartFileUploadFromInputStream(session,
                         request, streamReceiver, owner);
             }
+        } catch (IOException exception) {
+            // do not report IO exceptions via ErrorHandler
+            getLogger().warn("IO Exception during file upload", exception);
```
pleku, line 192:
```suggestion
            getLogger().warn("IO Exception during file upload, fired as StreamingErrorEvent", exception);
```

flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
```diff
@@ -522,37 +527,40 @@ private void cleanStreamVariable(VaadinSession session,
...
+            // be handled (to be passed to terminalErrorHandler): such
+            // exceptions mean mistakes in the implementation logic (not upload
+            // I/O operations).
+            throw new UploadException(e);
         }
         return new Pair<>(startedEvent.isDisposed(),
                 success ? UploadStatus.OK : UploadStatus.ERROR);
     }
 
+    private void fireStreamingFailedEvent(VaadinSession session,
```
pleku, line 550: Non blocking: this method is not just about firing the event so it should be named `onStreamingFailed` or such

pleku (changes requested):
Two suggestions for added clarity and one suggested rename. Otherwise OK

flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
denis-anisimov: Done.

pleku (approved):

PR #10556, opened 2021-04-06
fix: consider first added route as main route and always return it first

### LINKED ISSUE #10528: [BUG] RouterLink is using RouteAlias value instead of Route value in 14.5.1
**Describe the bug**

A view class with a route `something` and a route alias `""` (empty) was passed into `RouterLink` to create a router-link anchor.

Before 14.5.1, the `href` would be `something`, but after upgrading to 14.5.1, it has become empty. This is causing some of our BDD tests to fail.

This is only an issue when the RouteAlias value is empty. If the RouteAlias is a non-empty value, the routerlink uses the route value as expected.

**To Reproduce**
1. Create a class as follow
```
@Route(value = "something", layout = MainLayout.class)
public class SomethingView extends VerticalLayout {
  /* some content */
}
```
2. Create a router link and use it on a view
```
RouterLink routerLink = new RouterLink(name, navigationTarget, parameter);
```
3. Check the href of the anchor

**Expected behavior**
An anchor with `something` as href.

I was upgrading from 14.4.7 to 14.5.1

denis-anisimov:
fixes #10528

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
@@ -31,6 +32,8 @@ import java.util.function.Function;
 import java.util.regex.Pattern;
 
 import com.vaadin.flow.component.Component;
+import com.vaadin.flow.router.Route;
+import com.vaadin.flow.router.RouteAlias;
 import com.vaadin.flow.router.RouteParameters;
 import com.vaadin.flow.server.AmbiguousRouteConfigurationException;
 
@@ -101,11 +104,18 @@ final class RouteSegment implements Serializable {
      */
     private Map<String, RouteSegment> allSegments;
 
-    private RouteSegment() {
-    }
+    private final boolean isRoot;
 
-    private RouteSegment(String segmentTemplate) {
+    /**
+     * Main route segment should be returned first in the {@link #getRoutes()}
+     * method to support {@link Route}/{@link RouteAlias} annotation semantic :
+     * {@link Route} should take precedence over {@link RouteAlias}.
+     */
+    private boolean isMainRouteSegment;
+
+    private RouteSegment(String segmentTemplate, boolean isRoot) {
         this.template = segmentTemplate;
+        this.isRoot = isRoot;
 
         if (RouteFormat.isParameter(segmentTemplate)) {
             info = new RouteFormat.ParameterInfo(segmentTemplate);
@@ -124,6 +134,8 @@ final class RouteSegment implements Serializable {
         this.info = original.info;
         this.pattern = original.pattern;
         this.target = original.target;
+        this.isRoot = original.isRoot;
+        this.isMainRouteSegment = original.isMainRouteSegment;
 
         original.getStaticSegments().entrySet()
                 .forEach(e -> this.addSegment(new RouteSegment(e.getValue()),
@@ -146,7 +158,7 @@ final class RouteSegment implements Serializable {
      * Create a new root segment instance.
      */
     static RouteSegment createRoot() {
-        return new RouteSegment("");
+        return new RouteSegment("", true);
     }
 
     String getName() {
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
```diff
@@ -205,22 +217,36 @@ boolean isEligible(String value) {
      */
     Map<String, RouteTarget> getRoutes() {
```
caalador, line 218: Not explicitly connected to the change, but are we trusting that the given Map is ordered an just get the first iteration target from it? 
Not even the javaDoc mentions that this would be a ordered Map.

denis-anisimov: Kind of.......
I trust :)
It's an impl detail: the result is `LinkedHashMap` .
Technically there was intention to have support for Route/RouteAlias because of ordering but it's broken because of 
```
if (target != null) {
            result.put("", target);
        }
```
which breaks the order.

Javadocs are not necessary here since this is not a `public`  class and it's indeed is a bit out of the scope of the PR.
I may update the return value to `LinkedHashMap` even though it's not really clear which order it is:
as I said the order should have been the order of routes addition (considering categorization like static, parameter, etc) but the check `target != null`  ruins the order.
So the only order which work for sure is: if there is a first added route (not removed) then it goes first.

I don't know how to express this anyhow : please let me know if you see how it can be claridied.
Otherwise I would consider this just an impl detail.

`Route`  precedence over `RouteAlias`  with dynamic routes is quite questionable and fragile: `@Route`  may be removed at any point. So annotation based routing doesn't fit in fact into the dynamic routing model . And that's why I consider this as an impl detail.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
```diff
@@ -205,22 +217,36 @@ boolean isEligible(String value) {
...
+        Map<String, RouteSegment> leafSegments = getLeafSegments();
+
+        String mainRoutePath = null;
+        RouteTarget mainRouteTarget = null;
+
+        /*
+         * Find first the main route segment which should go first to be able to
+         * support Route/RouteAlias semantic. Main route doesn't have to exist
+         * though: it may be removed at any point. In this case the order
+         * doesn't matter since the RouteAlias ordering is not defined.
```
mshabarov, line 229: > RouteAlias ordering is not defined.

It sounds a bit confusing, because according the [route template priority](https://vaadin.com/docs/v14/flow/routing/tutorial-router-templates/#route-template-priority), the priority of `@RouteAlias` is the same as for `@Route`.  Or you mean something else under 'ordering' here?

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
```diff
@@ -205,22 +217,36 @@ boolean isEligible(String value) {
...
-        collectRoutes(result, getVarargsSegments());
+        leafSegments
+                .forEach((path, segment) -> result.put(path, segment.target));
 
-        if (getTemplate().isEmpty()) {
-            return Collections.unmodifiableMap(result);
-        } else {
-            return result;
-        }
+        return Collections.unmodifiableMap(result);
```
mshabarov, line 249: Perhaps, this is fine, but I don't get why is `getTemplate().isEmpty()` condition has been removed?

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
caalador: Dynamic routes don't need that much consideration here as they can be added even without any route annotations and should consider the initial registration as main.

But changing the return value to LinkedHashMap would make it clearer that the ordering can be trusted to some degree.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
caalador: We should always return an unmodifiable map. It shouldn't matter if the map is empty or not.

mshabarov: > We should always return an unmodifiable map

Yes, I agree. I just saw no relation between empty template and unmodifiable result, and wondered if there is any "hidden" behaviour behind it.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
```diff
@@ -101,11 +104,18 @@
...
-    private RouteSegment() {
-    }
+    private final boolean isRoot;
+
+    /**
+     * Main route segment should be returned first in the {@link #getRoutes()}
+     * method to support {@link Route}/{@link RouteAlias} annotation semantic :
+     * {@link Route} should take precedence over {@link RouteAlias}.
+     */
+    private boolean isMainRouteSegment;
```
mshabarov, line 114: Since the `isMainRouteSegment` and `isRoot` meanings look very close to each other (judging their names, 'root' is similar to 'main route'), would it be useful to add one more sentence that the `isMainRouteSegment` is true when the current route segment is a route segment and does not have children segments, is that correct?

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
denis-anisimov: The documentation is written according to the current impl which doesn't distinguish `Route`  and `RouteAlias` .
This design is not backward compatible and the ticket `#10528` is exactly about this.

One thing is correct though: templates in `Route` and `RouteAlias`  are introduced only in `2.5`  and they may behave in any way as we want. And there is no sense do not support what it says.

But the static routes should respect `Route` precedence over `RouteAlias`  so I will correct the code to use main route only with static routes.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
denis-anisimov: They have no any relation to each other.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
```diff
@@ -309,6 +351,16 @@ private void collectRoutes(Map<String, RouteTarget> result,
         }
     }
 
+    private RouteSegment getFirstLeafSegments() {
```
mshabarov, line 354: Should it be `getFirstLeafSegment` (with no 's' in the end)? The returned result is a single route segment, not the collection of route segments. However, a route segment might have a children, so It might be treated as a 'collection', thus, having a plural in the name might be technically correct.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
denis-anisimov: `root`  means `""` : empty route.
Route means route.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
denis-anisimov: Agreed.

denis-anisimov: Arghhh.
This is silly. 
The return type is `Map`  because `unmodifiableMap`  is used which is unable to return `LinkedHashMap`.
So if I change the return type then I cannot use `unmodifiableMap`.

Well, I guess it's not too bad to use modifiable map here since it's not a field and this is not a public API.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
denis-anisimov: Leftover.
Will correct.

mshabarov (approved):

PR #10561, opened 2021-04-06
chore: remove license header from build.gradle

taefi:
As build.gradle script is just a build-related
file, it does not need a license header.


caalador (approved):

pleku:
Why the double PR @taefi - why not just push the change to #10542 ?

taefi:
> Why the double PR @taefi - why not just push the change to #10542 ?
> 
> When we already have PR open to merge the feature branch, I don't see why we would make individual PRs targeting the feature branch

Isn't `feature/gradle` protected? Didn't consider using my admin privileges to push directly into this protected branch...

PR #10583, opened 2021-04-07
chore: disable gradle daemon thread

taefi:
As we use gradle for a single module build,
keeping a long-running gradle process, which
is not being stopped after the build completes,
would not have much impact to improve the 
performance. Disabling it may alleviate the 
insufficient memory errors and JVM crashes on
the CI.


caalador (approved):

taefi:
!DO NOT MERGE!

taefi:
Closed as another approach used.

Issue #10544, opened 2021-04-01
Custom theme live-reload doesn't work for newly added stylesheet files


mshabarov:
### Description of the bug / feature
Since V19.0.0.alpha5 the custom theme live-reload (webpack re-compilation) is not being triggered upon adding a new stylesheet files (or any other files) into `frontend/themes/my-custom-theme/` (including `components` folder).
It works fine with V19.0.0.alpha4.

### Minimal reproducible example
1. Go to start.vaadin.com
2. Choose `Vaadin 19` in the Settings tab
3. Download the project and run `mvn`
4. Add a new file `vaadin-button.css` (or any other name) to the `{project.root}/frontend/themes/myapp/components/`
5. Notice no webpack re-compilation in the console logs and no app restart
6. Add a new file `foo.css` (or any other name) to the `{project.root}/frontend/themes/myapp/`
7. Notice no webpack re-compilation in the console logs and no app restart
8. Change `vaadin.version` to 19.0.0.alpha4 in `pom.xml`, delete `node_modules` and any generated files in project root, then run `mvn clean spring-boot:run`
9. Do steps 4, 6 and check that the application restarts upon adding the files.

### Expected behavior
Webpack re-compiles and the application restarted, applying a newly added styles.
### Actual behavior
Nothing happens upon adding a new file.
### Versions:
    - Vaadin / Flow version: V19.0.2, Flow 6.0.3
    - Java version: 8
    - OS version: Mac OS
    - Browser version (if applicable): Chrome 89.0.4389.90
    - Application Server (if applicable): spring-boot
    - IDE (if applicable): IntelliJ IDEA 2020.3

mshabarov:
When the generated theme files were moved into separate folder, this code did not changed (webpack config file):
```javascript
const themeName = extractThemeName(flowFrontendThemesFolder);
```
It should have extracted the theme name from `frontendGeneratedFolder`, not from `flowFrontendThemesFolder`.
As a result, the watch and live reload plugins were not activated at all, because the theme name was empty.

This also shows how necessary it is to maintain the tests in a good condition.
Theme live reload test was disabled because of flakiness, while the change was done and didn't remind to change this folder either.

pleku:
> Add a new file foo.css (or any other name) to the {project.root}/frontend/themes/myapp/

@mshabarov I'd like to note that this is not even supposed to work - you have to reference the added file in `styles.css` with `@import url('foo.css');` and then live reload works with v19.0.3

[...]

Issue #9948, opened 2021-02-02
Server restart is required after removing a file from the theme 'components' folder


Artur-:
### Description of the bug / feature
If you delete e.g. `frontend/themes/components/vaadin-button.css`, webpack compilation will then fail with
```
ERROR in ./themes/myproject/components/vaadin-button.css
Module build failed (from ../node_modules/@vaadin/theme-loader/theme-loader.js):
Error: ENOENT: no such file or directory, open '/.../frontend/themes/myproject/components/vaadin-button.css'
```

### Minimal reproducible example
1. https://start.vaadin.com/?preset=prerelease&dl
2. Add vaadin-button.css to the folder
3. `mvn`
4. Remove vaadin-button.css from the folder

### Expected behavior
The theme is recompiled without the button css

### Versions:
    - Vaadin / Flow version: 19 alpha 5

caalador:
https://github.com/vaadin/flow/issues/9596

pleku:
The documented limitation says that _manual refresh in browser is needed_, and this is not enough as the error doesn't go away after refresh and a manual webpack restart is needed instead.

So by minimum, we should look into a) fixing the documented limitation b) figuring out if we can catch the failure somehow and force a restart or improve the error message somehow ?

pleku:
To be investigated more - we should be able to pick the removal of the file before webpack compilation and react to that, or then make sure that our plugin is not causing the failure due to the missing file. Timeboxed investigation for this is a good way to start.

mshabarov:
I reproduced this issue with starter app from start.vaadin.com using Vaadin `20.0.0.alpha7` and latest Flow snapshot `7.0-SNAPSHOT` and also with Jetty server.
There was no need for me to restart the server, but only the click on the Webpack error overlay (spring-boor case):
<img width="1783" alt="Screenshot 2021-04-23 at 10 13 28" src="https://user-images.githubusercontent.com/61410877/115835643-295f8780-a41f-11eb-881c-0a267c96efad.png">

However, for running in Jetty, there was no error overlay at all.
In both cases I was able to continue my work with styles further.

Regarding the issue itself, tried to recognise the component styles file event and then delete `theme-mytheme.generated.js`, or clear its content. I also tried `remove-files-webpack-plugin` plugin to delete this file before compilation.

[...]

PR #10584, opened 2021-04-07
test: Add a test for component styles live reload

mshabarov:
Related-to #10544


mshabarov:
Test for switching a theme name is in progress, but this fix can be validated manually and merged, as the issue is severe.

flow-server/src/main/resources/plugins/theme-live-reload-plugin/package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "repository": "vaadin/flow",
   "name": "@vaadin/theme-live-reload-plugin",
-  "version": "1.0.0",
+  "version": "2.0.0",
```
pleku, line 10: Not sure if it makes any difference, but should this just be a maintenance/minor bump ?

pleku (changes requested):

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
```diff
@@ -52,7 +47,15 @@ class ThemeLiveReloadPlugin {
...
-              if (file.indexOf(this.themeName + '.generated.js') > -1) {
+              // Webpack watches to the changes in theme-[my-theme].generated.js
+              // because it is referenced from theme.js. Changes in this file
+              // should not trigger the theme handling callback (which
+              // re-generates theme-[my-theme].generated.js),
+              // otherwise it will get into infinite re-compilation loop.
+              // There might be several theme generated files in the
+              // generated folder, so the condition does not contain the exact
+              // theme name
+              if (file.endsWith('.generated.js')) {
```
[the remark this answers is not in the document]
mshabarov: done

flow-server/src/main/resources/plugins/theme-live-reload-plugin/package.json
mshabarov: I followed the rule that changing the API (constructor) is a breaking change, and, thus, is a `+1.0.0`.

pleku (commented):
Testing the changes:
- adding /components and new files there seems to work
- adding & removing css that is imported to `styles.css` seems to work
- switching the theme on the fly:
1) first try failed with
```ERROR in ./generated/theme.js
Module not found: Error: Can't resolve './theme-myapp.generated.js' in 'C:\Users\Pekka\workspaces\my-app\frontend\generated'
 @ ./generated/theme.js 1:0-69 2:26-37
 @ ./generated/vaadin.ts
Child HtmlWebpackCompiler:
     1 asset
    Entrypoint HtmlWebpackPlugin_0 = __child-HtmlWebpackPlugin_0
       1 module
Child InjectManifest:
     1 asset
    Entrypoint InjectManifest = sw.js
    [../target/sw.ts] 3.1 KiB {InjectManifest} [built]
        + 92 hidden modules
i ?wdm?: Failed to compile.

------------------ Frontend compilation failed. ------------------
```
2. second try later on, it worked
3. third try, back to original theme, worked

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
```diff
@@ -30,32 +30,41 @@
...
+        createDirectoryIfAbsent(fontsDir);
+
+        File componentsDir = new File(themeFolder, "components");
+        createDirectoryIfAbsent(componentsDir);
 
         stylesCSSFile = new File(themeFolder, "styles.css");
         testStylesCSSFile = new File(themeFolder, "test-styles.css");
         fontFile = new File(themeFolder, "fonts/ostrich-sans-regular.ttf");
+        componentCSSFile = new File(new File(themeFolder, "components"),
+                "vaadin-text-field.css");
```
pleku, line 67: Should it test instead creating the file after the webpack is running ? Or should there be another test that does that? To cover both: changing existing component css and adding a new component css file with theme ?

mshabarov: > Should it test instead creating the file after the webpack is running ?

Currently, a new component css file is being created. And it happens after webpack is running.

> To cover both: changing existing component css and adding a new component css file with theme ?

I will add one more test to check the existing component css file changing.

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
```diff
@@ -252,4 +272,36 @@ private void createDirectoryIfAbsent(File dir) {
...
+    /**
+     * Converts rgba(X, Y, Z, A) to 6-digit hex format in the given style
+     * string.
+     * 
+     * @param styleString
+     *            color or border style value to be converted to color hex
+     *            format.
+     * @return converted style string.
+     */
+    private String convertRgbToHex(String styleString) {
```
pleku, line 285: I think there is plenty of CSS that could be used without involving colors and complex converting code for test like this, it feels like a waste of time for me to check the color of something when you could just change something else ...

pleku (approved):

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
```diff
@@ -89,9 +91,9 @@ public void webpackLiveReload_newCssAndFontCreatedAndDeleted_stylesUpdatedOnFly(
                 "Red background is not expected before applying the styles",
```
pleku, line 91: Would it make sense to split the test into two:
- create css - reload works
- create font - reload works

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
mshabarov: Yes, I've reverted it because it doesn't work anyway. Trying to use an attached element (similar to java live reload tests) and check its counter. Checking the client callback in Flow seems to work unstable and sometimes the checking code is faster than the live reload.

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
```diff
@@ -103,22 +112,22 @@ public void webpackLiveReload_newCssAndFontCreatedAndDeleted_stylesUpdatedOnFly(
...
+        });
         waitUntilCustomFont();
         waitUntilCustomBackgroundColor();
 
         // Live reload upon adding a new component styles file
         doActionAndWaitUntilLiveReloadComplete(
-                () -> createOrUpdateComponentCSSFile(CUSTOM_COMPONENT_BORDER));
-        waitUntilComponentCustomColor(CUSTOM_COMPONENT_BORDER);
+                () -> createOrUpdateComponentCSSFile(BORDER_RADIUS));
+        waitUntilComponentCustomColor(BORDER_RADIUS);
```
pleku, line 125: ..CustomColor(BORDER_RADIUS); ?

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
```diff
@@ -103,22 +112,22 @@ public void webpackLiveReload_newCssAndFontCreatedAndDeleted_stylesUpdatedOnFly(
...
 
         // Live reload upon updating component styles file
         doActionAndWaitUntilLiveReloadComplete(
-                () -> createOrUpdateComponentCSSFile(
-                        OTHER_CUSTOM_COMPONENT_BORDER));
-        waitUntilComponentCustomColor(OTHER_CUSTOM_COMPONENT_BORDER);
+                () -> createOrUpdateComponentCSSFile(OTHER_BORDER_RADIUS));
+        waitUntilComponentCustomColor(OTHER_BORDER_RADIUS);
```
pleku, line 130: ...CustomColor(BORDER_RADIUS); ?

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
```diff
@@ -137,39 +146,28 @@ private void waitUntilCustomFont() {
         waitUntil(driver -> isCustomFont());
     }
 
-    private void waitUntilComponentCustomColor(String border) {
-        waitUntil(driver -> isComponentCustomStyle(border));
+    private void waitUntilComponentCustomColor(String borderRadius) {
+        waitUntil(driver -> isComponentCustomStyle(borderRadius));
```
pleku, line 150: Should this be just removed or ?

pleku (changes requested):

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
mshabarov: Divided into two tests.

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
mshabarov: Renamed

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
mshabarov: Renamed

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
mshabarov: Why? I believe not

pleku (approved):

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
pleku: Not done ? I meant that there would be several `@Test` for different changes. But that was a question/suggestion, it is just refactoring

mshabarov: Unfortunately, I cannot split those tests, because TB runs them concurrently (despite `@NotThreadSafe` on the class), or I don't know who is a culprit, but they erase styles of each other when running from different methods.
I'm going to create an issue in TB regarding that, it's not a first time I notice that behaviour, WDYT @pekka?

pleku: Ok. That sounds very bad indeed. It should be trivial (for users, like us) to make a TB test not executed in parallel.

Please make an issue and we'll have to look at that later on

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
```diff
@@ -122,6 +117,20 @@ public void webpackLiveReload_newCssAndFontCreatedAndDeleted_stylesUpdatedOnFly(
...
         waitUntilCustomFont();
 
+        // Live reload upon file deletion
+        doActionAndWaitUntilLiveReloadComplete(
+                this::deleteBackgroundAndFontStyles);
+        waitUntilInitialStyles();
+    }
+
+    @Test
+    public void webpackLiveReload_newComponentCssCreatedAndDeleted_stylesUpdatedOnFly() {
```
pleku, line 127: Now I'm confused - you commented above that this cannot be done as the tests might be run parallel - which would also imply that in this case there might be a random-sequence of reloads in the browser since the execution of the each `@Test` could cause a reload when run in parallel in the CI ?

pleku (commented):
I'll have to retest things as the watched configuration was changed

flow-tests/test-application-theme/test-theme-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ThemeLiveReloadIT.java
mshabarov: Hmm, you're right. Another test can trigger live reload right after another test changes the styles. That could lead to 'false positive' test result, i.e. live reload doesn't work for one of those two cases, but another test triggers it => test passed erroneously.
Well, then they should be in the same method.

pleku (approved):
Retested live reload and seems to work. Only get an unrelated error when swapping the theme from `readManifestPaths` but when I remove the `@PWA` then the theme swap works again.

Just one thing to note is that any changes to `theme.json` do not trigger a recompilation until the resources are referenced from the css or Java code is recompiled, but this is fine.

pleku:
Would it be time to separate the fix into another PR and take the fix in ?

mshabarov:
> Would it be time to separate the fix into another PR and take the fix in without the not-working integration tests ?
Yes, will do

mshabarov:
Apparently, the problem was in having https://github.com/vaadin/flow/issues/9948, and the test failed when it was waiting until rendering the attached identifier element after deleting component style - the page probably was refreshed too early before actual webpack re-compilation, thus, the identifier was obsolete and wait condition failed.

flow-tests/test-application-theme/test-theme-component-live-reload/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-test-lumo</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-html-components-testbench</artifactId>
+            <version>${project.version}</version>
```
pleku, line 39: nit: this probably should be in test scope but it is very little difference

flow-tests/test-application-theme/test-theme-component-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ComponentThemeLiveReloadIT.java
```diff
@@ -0,0 +1,217 @@
...
+        createDirectoryIfAbsent(componentsDir);
+
+        componentCSSFile = new File(new File(themeFolder, "components"),
+                "vaadin-text-field.css");
+    }
+
+    @After
+    public void cleanUp() {
+        if (componentsDir.exists()) {
+            doActionAndWaitUntilLiveReloadComplete(() -> {
```
pleku, line 67: It seems weird that the cleanup code does wait for live reload to happen - is there a reason for that ? Is it part of the test ? Probably _the why_ should commented

pleku (changes requested):

flow-tests/test-application-theme/test-theme-component-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ComponentThemeLiveReloadIT.java
mshabarov: I wrapped these delete actions in a wait loop because:
1. To not leave the server in a reloading state after the test. There might be a second test rerun and the reload can affect it, and it then would be confusing for investigator what actually happened.
2. In case when in the future there comes any other `@Test` methods potentially (after solving TB parallel execution), it's good to start it without any live reload ongoing.

Will add an explanation comment there.

flow-tests/test-application-theme/test-theme-component-live-reload/pom.xml
mshabarov: done

flow-tests/test-application-theme/test-theme-component-live-reload/src/test/java/com/vaadin/flow/uitest/ui/ComponentThemeLiveReloadIT.java
mshabarov: added

pleku (approved):

PR #10585, opened 2021-04-07
Disable gradle plugin until build issues are sorted out

Artur-:


caalador:
As this seems to be more of an issue in teamcity than bender added a skip for the gradle plugin in the reactor build order for TC pull requests for now.

caalador:
Except the skip module has a problem when the project doesn't contain the gradle-plugin module...

PR #10569, opened 2021-04-06
fix: Do not use hardcoded target directory

caalador:
Any code writing to the build directory should
not depend on a hardcoded `target` directory
as for instance gradle uses `build` and it may
be configured also in Maven.

Fixes #10335

PR #10596, opened 2021-04-08
fix: Use (p)npm to install flow plugins

### LINKED ISSUE #10306: Vaadin 19 projects break if you add a dependency when using npm
```
vaadin init hello-vaadin --latest
cd hello-vaadin
# Add <pnpmEnable>false</pnpmEnable> to vaadin-maven-plugin in the production profile
mvn install -Pproduction
npm install open
node_modules/.bin/webpack
```

### Expected behavior
Webpack compiles the bundle

### Actual behavior
```
$ node_modules/.bin/webpack
/Users/artur/tmp/hello-vaadin/node_modules/webpack-cli/bin/cli.js:93
				throw err;
				^

Error: Cannot find module '@vaadin/stats-plugin'
```

### Reason

It seems that Flow messes with `node_modules/@vaadin` manually instead of using `npm`. 

`npm install open` cleans up the mess and logs 
```
removed 4 packages, and audited 1338 packages in 2s
```

### Versions:
    - Vaadin / Flow version: 19.0.0
    - npm: 7.6.0

caalador:
This also simplifies plugin development
by having the plugins in the build folder.

Fixes #10306


caalador:
Depends on #10569 to not use hardcoded folders.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
```diff
@@ -46,16 +46,18 @@
  */
 public class TaskInstallWebpackPlugins implements FallibleCommand {
 
-    private File nodeModulesFolder;
+    private File targetFolder;
+
+    protected static final String PLUGIN_TARGET = "plugins";
 
     /**
      * Copy Flow webpack plugins into the given nodeModulesFolder.
```
pleku, line 54: Javadocs not updated

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
```diff
@@ -46,16 +46,18 @@
  */
 public class TaskInstallWebpackPlugins implements FallibleCommand {
 
-    private File nodeModulesFolder;
+    private File targetFolder;
+
+    protected static final String PLUGIN_TARGET = "plugins";
```
pleku, line 51: Nit: missing javadocs for protected

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
```diff
@@ -77,7 +79,7 @@ public void execute() {
      *
      * @return names of plugins to install
      */
-    protected List<String> getPlugins() {
+    protected static List<String> getPlugins() {
```
pleku, line 82: Even though this API might be possible to be converted to be static, I'm not sure if it should be just done so and still kept in this class - doesn't seem clean as this is supposed to be a _task to install webpack plugins_ but now it has some static API that is used for something else. Should these be in an util or some other class instead ?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
```diff
@@ -325,13 +325,17 @@ File getGeneratedFallbackFile() {
...
+     *            the used build directory
      */
     TaskUpdateImports(ClassFinder finder,
             FrontendDependenciesScanner frontendDepScanner,
             SerializableFunction<ClassFinder, FrontendDependenciesScanner> fallBackScannerProvider,
             File npmFolder, File generatedPath, File frontendDirectory,
-            File tokenFile, JsonObject tokenFileData, boolean disablePnpm) {
-        super(finder, frontendDepScanner, npmFolder, generatedPath, null);
+            File tokenFile, JsonObject tokenFileData, boolean disablePnpm,
+            String buildDir) {
```
pleku, line 336: Refactoring: should there be some context/configuration object or some pojo that encapsulates the needed things instead of introducing more and more parameters ? The thing goes for other classes like `NodeUpdater`

pleku (changes requested):
This change is a bit weird in the sense that the issue is about being able to
1) install dependencies with npm
2) run webpack manually

which is a "new feature". The suggested change here fixes things so that it possible to now achieve 1+2 (?) but it is not tested in any way that it is actually possible now or it won't get broken later on. So this seems like this is not a new feature (as there are no tests that validate this) but just refactoring that makes things work for now.

caalador:
>  not tested in any way
There is the test that the plugins are added to the package.json and for the functionality we have all tests in flow-tests that test and depend on this functionality as webpack.generated.js depends on the existence of the plugins.

pleku:
> > not tested in any way

> There is the test that the plugins are added to the package.json and for the functionality we have all tests in flow-tests that test and depend on this functionality as webpack.generated.js depends on the existence of the plugins.

Is that the same as what I described as 1) and 2) in my comment, the "feature" ?  I don't understand how those tests validate that the feature of using _npm install and webpack manually_ is working and done. I understand that it just tests our webpack plugins are installed and where those are installed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
caalador: The whole `Task*` should be updated and refactored. I would perhaps add a ticket about it (and then do it perhaps even before this feature) as it will not be a minor extra on top.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
caalador: Moved and Docs added.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
caalador: Moved and Docs added.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskInstallWebpackPlugins.java
caalador: Created new utility class.

caalador:
Before this change the plugins were installed after npm install task was executed so that npm doesn't clear the plugins (especially if the node_modules should be empty) and now the plugins need to be copied before the npm install is executed.

Our npm install and a manual npm install doesn't differ in actual execution. 

Also our running of webpack doesn't differ from the manual running of webpack except for speed and log output.

I don't see how this could be tested "outside" the framework except by running a shellscript on an existing project if the fact that the task execution order is not seen as a functionality change.

flow-server/src/main/java/com/vaadin/flow/server/frontend/WebpackPluginsUtil.java
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Load and parse the requested Json file.
+     * 
+     * @param jsonFilePath
+     *            path to json file
+     * @return parsed Json for file if found
+     * @throws IOException
+     *             thrown for problems reading file
+     */
+    protected static JsonObject getJsonFile(String jsonFilePath)
```
pleku, line 88: First thought this could be moved to `JsonUtils`, but then it would be weird if that one has a dependency to `FrontendUtils` which should in theory be outside of `flow-server` anyway ... so maybe this is fine

pleku (approved):

pleku:
> I don't see how this could be tested "outside" the framework except by running a shellscript on an existing project if the fact that the task execution order is not seen as a functionality change.

@caalador I'd just mark this as either a fix or refactor - I don't see how from users perspective we're adding a new feature that should be highlighted in the release notes. We can change the linked issue to be a bug fix instead

pleku:
Just to clarify - we're not going to backport this to v19. Right @caalador ?

caalador:
No. I don't think this should be backported to 19.

Issue #10451, opened 2021-03-29
Switching between theme folders causes endless reload loop


OlliTietavainenVaadin:
### Description of the bug / feature

Adding two (or more) Theme folders and switching the `@Theme` annotation from one `themeFolder` value to another after starting the app once causes the browser to end up in an endless reload loop.

### Minimal reproducible example
https://github.com/OlliTietavainenVaadin/v14-6-pit
* Started from a Vaadin 14.5 project from Vaadin Start, updated to `14.6.0.alpha1` to test the new theme system. 
* Created two theme folder, `custom` and `custom2`
* Added a `styles.css` with different values to each theme
* Added a `@Theme(themeFolder="custom")` annotation to the `MainView`, started the app, and opened it in the browser
* Switched the theme annotation to `@Theme(themeFolder="custom2")`, restarted the app, and opened it in the browser

### Expected behavior
App should switch to the theme `custom2`
### Actual behavior
App ends up in a reload loop
### Versions:
    - Vaadin / Flow version: 14.6.0.alpha1
    - Java version: 11
    - OS version: Windows 10
    - Browser version (if applicable): Chrome
    - Application Server (if applicable): spring-boot
    - IDE (if applicable): IntellJ Idea

pleku:
Switching the theme from one to another requires restarting the server, if it is running already. This is at least how I've understood it. And that is also the workaround for now so I'm not marking this as a blocker as it is trivial to apply the workaround.

OlliTietavainenVaadin:
Yes, I can confirm, you need to stop the server completely and restart it (the autoreload's restart isn't enough). So it's annoying but not a blocker.

mshabarov:
I've not yet tried the app example to reproduce it, but I got this step:
`Switched the theme annotation to @Theme(themeFolder="custom2"), restarted the app, and opened it in the browser`
as a manual restart of the server, so put a blocker in there.

If the manual restart does the trick, then ofc it's not a blocker.

BTW, the custom theme documentation states that you cannot switch the theme on the fly https://vaadin.com/docs/v14/flow/styling/custom-theme/#limitations

pleku:
First we should check that what is the reload loop about (probably by webpack?) and try to avoid it and fail more graciously and make it clear that a restart is needed. We can then spend time on investigating how we could just make the theme (annotation) scanning work during development so that we could enable swapping the theme on the fly. There are some options for this. We should spend a few days timebox to investigate our options for improving this.

[...]

PR #10599, opened 2021-04-08
fix: Skip any generated files in theme live reload plugin

mshabarov:
Switching the theme name in `@Theme` annotation triggered/created a new [other-theme].generated.js file, which is being watched by webpack and trigger another live-reload in turn. 
Thus, the endless loop took place. 
This fix skips the live reload for all generated theme files, no matter which name it has.

Related-to #10451


caalador (approved):

mshabarov:
Let's merge this change for now without theme switch test, to address this issue quickly.
Jetty plugin does not trigger the live reload upon changing the java class file, probably something is broken with configuration.
I need to look into it more.

PR #10605, opened 2021-04-09
chore: clarify VaadinContext::getAttribute semantic

### LINKED ISSUE #9517: VaadinServletContext::getAttribute(Class, Supplier) set the default value
The method `VaadinContext::getAttribute(Class<T> type, Supplier<T> defaultValueSupplier) `  sets the value rom the supplier instead just using the value to return.
It has some sense though I don't have a strong opinion about this.

The issue is: javadoc of `VaadinContext::getAttribute(Class, Supplier)` doesn't mention this.
So either the method should not set the attribute but just return the value from the supplier (then I'm not sure why supplier is needed at all?) or this behavior should be documented.

In case the current behavior of `getAttribute` is changed (so that it doesn't set the value anymore) the API of `VaadinContext`  should be extended so that it provides a method to atomically execute a command.
Currently `getAttribute` allows to check the value presence and set the value atomically (under the context lock).
If behavior is changed then there will be no way to execute such operation properly (it will be thread unsafe).

denis-anisimov:
Fixes #9517

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinContext.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinContext.java
@@ -31,8 +31,12 @@ import java.util.function.Supplier;
 public interface VaadinContext extends Serializable {
 
     /**
-     * Returns value of the specified attribute, creating a default value if not
-     * present.
+     * Returns value of the specified attribute, creating and set a default
+     * value if not present.
+     * <p>
+     * If attribute is not yet available the {@code defaultValueSupplier} is
+     * used to get the default value which is set as the attribute value and the
+     * value is returned. The operation is executed atomically.
      *
      * @param type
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinContext.java
```diff
@@ -31,8 +31,12 @@
 public interface VaadinContext extends Serializable {
 
     /**
-     * Returns value of the specified attribute, creating a default value if not
-     * present.
+     * Returns value of the specified attribute, creating and set a default
+     * value if not present.
```
caalador, line 35:
```suggestion
     * Returns value of the specified attribute, creating and storing a default
     * value if attribute not present.
```

caalador (approved):

PR #10622, opened 2021-04-12
fix: Extract theme name from proper folder and skip generated files

mshabarov:
Extracts the custom theme name from frontend generated folder and skips theme handling when the theme generated file is being created/updated.

Related-to #10451, #10544

IT test to cover component styles live reload is being worked in a separate PR https://github.com/vaadin/flow/pull/10584, it is flaky still.
IT test to cover themes switch is under work still.


pleku (approved):
Maybe add the comment in the PR description that tests are being worked in another PR so it doesn't look so bad later on :)

mshabarov:
> Maybe add the comment in the PR description that tests are being worked in another PR so it doesn't look so bad later on :)

Done :)

PR #10627, opened 2021-04-13
Consider generated icons public resources

Artur-:


flow-server/src/test/java/com/vaadin/flow/server/HandlerHelperTest.java
```diff
@@ -201,4 +204,29 @@ private VaadinRequest createVaadinRequest(String requestPath,
...
+        expected.add("/icons/icon-512x512.png");
+        expected.add("/icons/icon-16x16.png");
+        expected.add("/icons/icon-32x32.png");
+        expected.add("/icons/icon-96x96.png");
+        expected.add("/icons/icon-180x180.png");
+        expected.add("/icons/icon-1125x2436.png");
+        expected.add("/icons/icon-750x1334.png");
+        expected.add("/icons/icon-1242x2208.png");
+        expected.add("/icons/icon-640x1136.png");
+        Assert.assertArrayEquals(expected.toArray(),
```
joheriks, line 229: Compare as sets to reduce sensitivity of test to order changes?

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -312,14 +316,22 @@ public static boolean isPathUnsafe(String path) {
...
-        return new String[] { //
-                "/favicon.ico", //
-                "/" + PwaConfiguration.DEFAULT_PATH, //
-                "/" + FrontendUtils.SERVICE_WORKER_SRC_JS, //
-                PwaHandler.SW_RUNTIME_PRECACHE_PATH, //
-                "/" + PwaConfiguration.DEFAULT_OFFLINE_PATH, //
-                "/" + PwaHandler.DEFAULT_OFFLINE_STUB_PATH //
-        };
+        List<String> resources = new ArrayList<>();
+        resources.add("/favicon.ico");
```
knoobie, line 320: What do you think about changing this to a `private static` field that is just returned here instead of re-creating this whole construct on every call to `getPublicResources`?

Artur-: Changed

flow-server/src/test/java/com/vaadin/flow/server/HandlerHelperTest.java
Artur-: Changed

joheriks (approved):

PR #10630, opened 2021-04-13
fix: Only override older package versions

### LINKED ISSUE #9793: PNPM overrides version with platform version
When using pnpm if you want to depend on a non platform version of a vaadin component the npm version will be overridden with the platform defined version.

e.g. if the platform npm version of vaadin-charts is 6.3.1 and you add a dependency for vaadin-charts 9.0.0 that has the npm version 7.0.0 then as the version is still framework managed as NpmPackage will update the version versions.json  for forcing versions will get the platform version 6.3.1 instead of the expected 7.0.0

If the version is set manually to 7.0.0 in package.json it will override the platform version but in this case also the java package need to be updated.

For a sample project see https://github.com/vaadin/vaadin-charts/issues/557

caalador:
If package dependency version in
package json is newer than in versions.json
then we leave the newer version unchanged.

Fixes #9793

### DIFF
--- a/flow-server/src/main/resources/pnpmfile.js
+++ b/flow-server/src/main/resources/pnpmfile.js
@@ -25,9 +25,13 @@ function readPackage(pkg) {
   const {dependencies} = pkg;
   
   if (dependencies) {
-    for (let k in versions) {
-      if (dependencies[k] && dependencies[k] !== versions[k]) {
-        pkg.dependencies[k] = versions[k];
+    for (let key in versions) {
+      if (dependencies[key]) {
+        if (dependencies[key].startsWith('./') || compareVersions(dependencies[key], versions[key]) === -1) {
+          // only rewrite to versions version if dependency older
+          console.debug("Updating " + key + " to " + versions[key] + " from " + pkg.dependencies[key]);
+          pkg.dependencies[key] = versions[key];
+        }
       }
     }
   }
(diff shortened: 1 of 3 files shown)


flow-server/src/main/resources/pnpmfile.js
```diff
@@ -25,9 +25,13 @@ function readPackage(pkg) {
   const {dependencies} = pkg;
   
   if (dependencies) {
-    for (let k in versions) {
-      if (dependencies[k] && dependencies[k] !== versions[k]) {
-        pkg.dependencies[k] = versions[k];
+    for (let key in versions) {
+      if (dependencies[key]) {
+        if (dependencies[key].startsWith('./') || compareVersions(dependencies[key], versions[key]) === -1) {
```
pleku, line 30: Would it be impossible to just use something like https://docs.npmjs.com/cli/v7/using-npm/semver in here so that we don't have to have our own parsing ?

flow-server/src/main/resources/pnpmfile.js
```diff
@@ -40,3 +44,77 @@ function readPackage(pkg) {
...
+  for (let i = 0; i < Math.max(v1Array.length - 1, v2Array.length - 1); i++) {
+    const number1 = parseInt(v1Array[i] || 0, 10);
+    const number2 = parseInt(v2Array[i] || 0, 10);
+
+    if (number1 > number2) return 1;
+    if (number2 > number1) return -1;
+  }
+
+  const sp1 = v1Array[v1Array.length - 1];
+  const sp2 = v2Array[v2Array.length - 1];
```
pleku, line 102: sp?

pleku (changes requested):

flow-server/src/main/resources/pnpmfile.js
```diff
@@ -40,3 +44,77 @@ function readPackage(pkg) {
...
+function tryParse(v) {
+  return isNaN(Number(v)) ? v : Number(v);
+}
+
+function validate(version) {
+  if (typeof version !== 'string') {
+    throw new TypeError('Invalid argument expected string got ' + version);
+  }
+  if (!semver.test(version)) {
+    throw new Error('Invalid argument not valid semver (\'' + version + '\' received)');
```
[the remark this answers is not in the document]
caalador: It would show up in the logs as `ERROR? Invalid argument not valid semver ('./target/flow-frontend' received)at validate                 pnpmfile.js:81` failing pnpm install

caalador:
I'll rework the version comparison if this doesn't cause new problems.

caalador:
Closing this as it brings more problems and enables getting mixed versions.

Issue #10357, opened 2021-03-22
Spring login helper should take request cache into account


haijian-vaadin:
As a developer, to redirect the user to the unauthenticated requests, I want to use a [request cache](https://github.com/vaadin/bakery-app-starter-flow-spring/blob/master/src/main/java/com/vaadin/starter/bakery/app/security/SecurityConfiguration.java#L87), for example in the Bakery app.

Currently, the request cache how no effect on the[ login helper](https://github.com/vaadin/flow/blob/master/flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts#L26), one would have to do the authenticated request redirection manually, e.g., with a [returnUrl](https://github.com/vaadin-learning-center/crm-tutorial-typescript/blob/master/frontend/components/login-view/login-view.ts#L23) as done in the CRM TS tutorial.

DoD:
Given that a request cache configured, calling the `login` helper will also redirect the users to the authenticated access attempt, as a developer, I don't need to handle it manually with a [returnUrl](https://github.com/vaadin-learning-center/crm-tutorial-typescript/blob/master/frontend/components/login-view/login-view.ts#L23).

This now turns out to be also a bug, since now we have a default request cache. A user cannot log in anymore when first trying to access an unauthenticated view.

PR #10634, opened 2021-04-14
fix: cannot login to authentication protected view

haijian-vaadin:
The [Vaadin security helper](https://github.com/vaadin/spring/blob/master/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurerAdapter.java) now has a request cache by default, which means that if a user tries to access an authentication protected view, the request will be saved and has it as a `redirectUrl` in the response. 
The logic in the `login()` helper didn't take this request cache into account.

Part of #10357, which fixes the bug.

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
@@ -7,6 +7,7 @@ export interface LoginResult {
   token?: string;
   errorTitle?: string;
   errorMessage?: string;
+  redirectUrl?: string;
 }
 
 export interface LoginOptions {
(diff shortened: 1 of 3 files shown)


flow-client/src/test/frontend/AuthenticationTests.ts
```diff
@@ -71,13 +71,11 @@ describe('Authentication', () => {
     it('should return a CSRF token on valid credentials', async () => {
       fetchMock.post('/login', {
         body: happyCaseResponseText,
-        redirectUrl: '/'
+        redirectUrl: 'localhost:8080/'
```
platosha, line 74: Incorrect URL, `localhost:` could be recognised as a protocol (http?). Although likely not an issue for this test, let’s change to a correct form.

```suggestion
        redirectUrl: '//localhost:8080/'
```

flow-client/src/test/frontend/AuthenticationTests.ts
```diff
@@ -105,6 +103,39 @@ describe('Authentication', () => {
...
+
+    it('should redirect based on request cache after login', async () => {
+      // An unthenticated request attempt would be captured by the default
+      // request cache, so after login, it should redirect the user to that 
+      // request
+      fetchMock.post('/login', {
+        body: happyCaseResponseText,
+        // mock the unthenticated attempt, which would be 
+        // saved by the default request cache
+        redirectUrl: 'localhost:8080/protected-view'
```
platosha, line 128:
```suggestion
        redirectUrl: '//localhost:8080/protected-view'
```

platosha (changes requested):

flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
```diff
@@ -37,23 +37,22 @@ export async function login(username: string, password: string, options?: LoginO
     const response = await fetch(loginProcessingUrl, { method: 'POST', body: data, headers });
 
     const failureUrl = options && options.failureUrl ? options.failureUrl : '/login?error';
-    const defaultSuccessUrl = options && options.defaultSuccessUrl ? options.defaultSuccessUrl : '/';
```
joheriks, line 40: Does it mean the `defaultSuccessUrl` field of `LoginOptions` should be removed, as it is no longer a relevant option?

haijian-vaadin: yes, the `defaultSuccessUrl ` is now useless. Not removing it because removing it would be a breaking change, and the fix needs to be cherry-picked to v19. 
So plan to remove it in a separate PR.

haijian-vaadin: Marked the `defaultSuccessUrl` field as deprecated.

platosha (approved):

PR #10635, opened 2021-04-14
fix: Accept newer framework managed versions

### LINKED ISSUE #9793: PNPM overrides version with platform version

caalador:
Any framework managed version that is newer
than the version in platform versions should be
left as is as the version probably is from an updated
annotation.

Fixes #9793

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
@@ -86,14 +86,7 @@ class VersionsJsonFilter {
      */
     private JsonObject collectUserManagedDependencies(JsonObject packageJson) {
         JsonObject json = Json.createObject();
-        JsonObject vaadinDep;
-        if (packageJson.hasKey(VAADIN_DEP_KEY) && packageJson
-                .getObject(VAADIN_DEP_KEY).hasKey(dependenciesKey)) {
-            vaadinDep = packageJson.getObject(VAADIN_DEP_KEY)
-                    .getObject(dependenciesKey);
-        } else {
-            vaadinDep = Json.createObject();
-        }
+        JsonObject vaadinDep = collectFrameworkVersions(packageJson);
 
         if (packageJson.hasKey(dependenciesKey)) {
             JsonObject dependencies = packageJson.getObject(dependenciesKey);
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
@@ -15,6 +15,8 @@
  */
 package com.vaadin.flow.server.frontend;
 
+import org.slf4j.LoggerFactory;
+
 import elemental.json.Json;
 import elemental.json.JsonObject;
 
@@ -33,6 +35,8 @@ class VersionsJsonFilter {
 
     private final String dependenciesKey;
 
+    private final String OLDER_VERSION_WARNING = "Using user pinned version '{}' of '{}' which is older than the current platform version '{}'";
+
     VersionsJsonFilter(JsonObject packageJson, String dependenciesKey) {
         this.dependenciesKey = dependenciesKey;
         userManagedDependencies = collectUserManagedDependencies(packageJson);
@@ -49,6 +53,13 @@ class VersionsJsonFilter {
         JsonObject json = Json.createObject();
         for (String key : versions.keys()) {
             if (userManagedDependencies.hasKey(key)) {
+                if (isNewer(versions.getString(key),
+                        userManagedDependencies.getString(key))) {
+                    LoggerFactory.getLogger("Versions").warn(
+                            OLDER_VERSION_WARNING,
+                            userManagedDependencies.getString(key), key,
+                            versions.getString(key));
+                }
                 json.put(key, userManagedDependencies.getString(key));
             } else if (vaadinVersions.hasKey(key) && isNewer(
                     vaadinVersions.getString(key), versions.getString(key))) {
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
@@ -29,12 +29,14 @@ import static com.vaadin.flow.server.frontend.NodeUpdater.VAADIN_DEP_KEY;
 class VersionsJsonFilter {
 
     private final JsonObject userManagedDependencies;
+    private final JsonObject vaadinVersions;
 
     private final String dependenciesKey;
 
     VersionsJsonFilter(JsonObject packageJson, String dependenciesKey) {
         this.dependenciesKey = dependenciesKey;
         userManagedDependencies = collectUserManagedDependencies(packageJson);
+        vaadinVersions = collectFrameworkVersions(packageJson);
     }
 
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
```diff
@@ -99,4 +120,20 @@ private boolean isUserChanged(String key, JsonObject vaadinDep,
...
+     * @return Vaadin dependencies or empty object
+     */
+    private JsonObject collectFrameworkVersions(JsonObject packageJson) {
+        if (packageJson.hasKey(VAADIN_DEP_KEY) && packageJson
+                .getObject(VAADIN_DEP_KEY).hasKey(dependenciesKey)) {
+            return packageJson.getObject(VAADIN_DEP_KEY)
+                    .getObject(dependenciesKey);
+        }
+        return Json.createObject();
+    }
```
pleku, line 138: This is just code duplication from the beginning of `collectUserManagedDependencies`, maybe just make that method reuse the object fetched here ?

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
```diff
@@ -46,15 +48,34 @@
     JsonObject getFilteredVersions(JsonObject versions) {
         JsonObject json = Json.createObject();
         for (String key : versions.keys()) {
-            if (!userManagedDependencies.hasKey(key)) {
-                json.put(key, versions.getString(key));
-            } else {
+            if (userManagedDependencies.hasKey(key)) {
                 json.put(key, userManagedDependencies.getString(key));
```
pleku, line 52: Should we opt for newer platform managed dependency when possible ?
One scenario is when people downgrade due to an issue - then I think this will work. But then when they update the platform version later on, the old version stays there forever ?
Another scenario is when people upgrade a version before it is released in platform - then this works but the override stays there forever and the downside of the first scenario happens and the version is not updated anymore ? 

For pnpm the node_modules etc. are never deleted (at least by `TaskUpdatePackages`) since it only checks for the version upgrade in `!enablePnpm`

pleku: Should we just make the upgrade possible but not downgrade ? Then basically they have to downgrade the whole platform version to be able to downgrade a component.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
```diff
@@ -558,6 +557,95 @@ public void runPnpmInstall_customNpmRcFileFound_npmRcFileIsNotGenerated()
...
+                        Json.parse("{ \"@vaadin/vaadin-overlay\":\""
+                                + PINNED_VERSION + "\"}"))
+                        .toJson());
+        task.execute();
+
+        File overlayPackageJson = new File(getNodeUpdater().nodeModulesFolder,
+                "@vaadin/vaadin-overlay/package.json");
+
+        // The resulting version should be the one specified via platform
+        // versions file
```
pleku, line 594: This comment is the opposite of what is expected below ?

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
```diff
@@ -558,6 +557,95 @@ public void runPnpmInstall_customNpmRcFileFound_npmRcFileIsNotGenerated()
...
+        JsonObject overlayPackage = Json.parse(FileUtils
+                .readFileToString(overlayPackageJson, StandardCharsets.UTF_8));
+        Assert.assertEquals(customOverlayVersion,
+                overlayPackage.getString("version"));
+    }
+
+    @Test
+    public void runPnpmInstall_vaadinVersionNewerThanPinned_installedOverlayVersionIsNotSpecifiedByPlatform()
```
pleku, line 602: (same comment as in code earlier) Is this the expectancy that we would want to have ? That users stay on the old version until they manually remove it. Doesn't seem like a good way

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
caalador: :+1:

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
caalador: ? This tests that when we find a npm annotation with a version newer than the one in platform versions that we don't take the version from platform, but use the found annotated version.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
caalador: fixed

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
caalador: I can't say. 
At the time of making there was quite a long debate on if it should be possible for the user to set only newer or also give the choice to have an older version.
The end result then was that the user can put in what he want's and we won't touch it if it never matches the framework found version.

pleku: Yeah, then it is actually the same as with Maven - when you add your own overrides, you have to remember to remove those too or swear when you don't. So lets keep it.

But it might help if we could log a warning that "you've pinned an older version of vaadin-party-parrot than what comes with vaadin x.y.z" so that people could mayhaps realize that they can remove the pin. Can you add something like that ?

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
pleku: Right, I misunderstood the test context as `vaadin` is not the same as `platform` ..

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
caalador: Added warning into logs for the older version.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
caalador: Changed the test name to be more clear that it's the version that flow has found and written to the package.json

pleku (approved):

PR #10662, opened 2021-04-15
refactor: clarify naming and simplify logic

Artur-:


fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
```diff
@@ -117,41 +117,60 @@ public AnnotatedElement getSecurityTarget(Method method) {
...
-                        + "@RolesAllowed";
-            } else {
-                accessDeniedMessage = "Unauthorized access to Vaadin endpoint";
-            }
-            return accessDeniedMessage;
+        if (canAccessMethod(method, request)) {
+            return null;
+        }
+
+        if (isDevMode()) {
```
knoobie, line 134: Is it worth to save the result as field in the class to reduce the calls to `VaadinService.getCurrent().getDeploymentConfiguration().isProductionMode()` on every endpoint call? The `devMode` should not change in between calls, or?

Artur-: I doubt it would make any difference. The problem is that production mode is defined per servlet/service instead of per application so this is the academically correct way to check it. In theory there could be multiple services although in practice there is not. I think it should be fixed in Flow so that development mode is a flag for the application

knoobie: You are right, that makes more sense!

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
```diff
@@ -117,41 +117,60 @@ public AnnotatedElement getSecurityTarget(Method method) {
...
+     * <p>
+     * Note! If CSRF protection is disabled, this method will always return
+     * {@code true}.
+     * 
+     * @param request
+     *            the request to validate
+     * @return {@code true} if the CSRF token is ok or checking is disabled or
+     *         there is no HTTP session, {@code false} otherwise
+     */
+    private boolean isValidCsrfToken(HttpServletRequest request) {
```
haijian-vaadin, line 173: maybe better to change `isValidCsrfToken` to `hasValidCsrfToken`

haijian-vaadin (changes requested):
Looks good, just a small renaming suggestion.

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
Artur-: Would it be strange if `hasValidCsrfToken` returns `true` when there is no session and thus no token?

haijian-vaadin: That's true, would `validateCsrfTokenInRequest()` be better? with `isValidCsrfToken()` I would expect the parameter to be a token.

Artur-: Not optimal but I can live with that. Done

haijian-vaadin (approved):

PR #10663, opened 2021-04-15
fix: include redirectUrl in the return value of login()

haijian-vaadin:
It includes the `redirectUrl` in the return value of the `login()` helper. So that if the user has defined a request cache or a defaultSuccessUrl in the Spring Security configuration, user can use the `redirectUrl` in `LoginResult` to redirect instead of maintaining a `returnUrl` manually as in the current doc, which also doesn't work when a request cache or a defaultSuccessUrl is configured. 

Part of https://github.com/vaadin/flow/issues/10357

### DIFF
--- a/flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
+++ b/flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
@@ -7,6 +7,7 @@ export interface LoginResult {
   token?: string;
   errorTitle?: string;
   errorMessage?: string;
+  redirectUrl?: string;
 }
 
 export interface LoginOptions {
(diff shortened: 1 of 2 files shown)


flow-client/src/main/resources/META-INF/resources/frontend/Authentication.ts
```diff
@@ -53,7 +54,8 @@ export async function login(username: string, password: string, options?: LoginO
         if (vaadinCsrfToken) {
           result = {
             error: false,
-            token: vaadinCsrfToken
+            token: vaadinCsrfToken,
+            redirectUrl: response.url,
```
Artur-, line 58:
```suggestion
            redirectUrl: response.url
```

platosha (approved):

PR #10669, opened 2021-04-15
chore: explain JavaScript annotation behavior

### LINKED ISSUE #8285: The @JavaScript annotation should explain that it's included in a module scope
In general, the following forum topic poses an interesting question:

https://vaadin.com/forum/thread/18116726/javascript-function-defined-in-jsfile-used-by-javascript-annotation-is-n

The question is: why does not `@JavaScript()` annotation run given javascript file in a global scope, but in a module scope? I think it's really important to document this straight in the `@JavaScript` annotation javadoc, since Java developers are usually not JavaScript developers.

    - Vaadin / Flow version: 14.1.27

denis-anisimov:
fixes #8285

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/dependency/JavaScript.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/dependency/JavaScript.java
@@ -52,9 +52,35 @@ import com.vaadin.flow.shared.ui.LoadMode;
  * {@code RouteA} annotated with {@code @Route("route-a")} and
  * {@code @JavaScript("./src/javascript.js")}, the {@code javascript.js} will be
  * run on the root route as well.
+ * <p>
+ * External JavaScript dependencies (e.g. "http://example.com/some.js") are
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/component/dependency/JavaScript.java
```diff
@@ -52,9 +52,35 @@
...
+ * <pre>
+ *  
+ * <code>
+ * UI.getCurrent().getPage().executeJs("some_function_definied_in_module()");
+ * </code>
+ * </pre>
+ * 
+ * because the function is private there (unless it's explicitly exposed
+ * somehow). The JavaScript where the function is defined should be either
+ * external or it should be added using {@link Page#addJavaScript(String)}: in
```
caalador, line 76:
```suggestion
 * because the function is private there (unless it's explicitly exposed).
 * The JavaScript where the function is defined should be either
 * external or it should be added using {@link Page#addJavaScript(String)}: in
```

caalador (approved):

Issue #10598, opened 2021-04-08
Validation support for warnings (or in general more ErrorLevels)


stefanuebe:
Using validators together with binders currently only allow validation / handling of errors. There are user scenarios, where entering a value should not lead to an invalid form, but at least should show a warning or allow the developer to react programmatically to such warnings. In a more general way, it could be extended to allow using all existing ErrorLevels as a ValidationResult (e. g. when an app shall differentiate between normal errors and critical ones).

Creating a ValidationResult with for instance ErrorLevel.WARNING is already possible, but the Binder / Validator API does neither support showing this warning nor does it allow to check, whether there have been warnings or not.

Therefore the requested enhancement is to extend the Validator/Binder API to allow a more generic way of using ErrorLevels, for instance by 
* adding new convenience methods isWarning() etc. for ValidationResult 
* adapting the methods hasErrors()/getValidationErrors() from BinderValidationStatus to new methods hasResults(ErrorLevel) / getValidationResults(ErrorLevel); plus convenience methods hasWarnings/getValidationWarnings(), etc.
* Binder#validate should not only "react" to errors, but all types of ErrorLevels, fire a respective StatusChangeEvent and return a BinderValidationStatus containing all occurred ErrorLevels (e.g. warning and error). 
* StatusChangeEvent also should provide the validation ErrorLevels , not only a boolean "hasValidationError".
* Validation result messages of other ErrorLevels should be shown as they are for errors, but with a different styling (with a Lumo default style, e.g. something yellow-ish for warnings, blue for info, etc.; which can be customized)
* adding Icons indicating the ErrorLevel (exclamation mark for warning, x in a circle for error, **i** in a circle for info)
* Not sure if needed, but maybe necessary for backward compatibility: Binder can be configured on which ErrorLevel (plus above) a status change event shall be fired (default is all, but then it can easily be set to ERROR)
* The handling of ValidationResults, which are "ok" (ErrorLevel null), should not change

knoobie:
We had this requirement as well, instead of failing the whole process the customer wanted to show that those fields "aren't filled optimal" like e.g.

- `INFO`, `WARN`, `ERROR` with different color codings to show that the given value is technically valid but can still be problematic later on
- Example:  Age > 100 - could be a typo or correct. Only the user can know this. Show `WARN` "Given age is greater than 100. Are you sure this person is still alive?"

javier-godoy:
[...]

PR #10683, opened 2021-04-15
feat: add styling support for validation warnings

javier-godoy:
Part of #10598

Implements the following part of #10598:

> Validation result messages of other ErrorLevels should be shown as they are for errors, but with a different styling (with a Lumo default style, e.g. something yellow-ish for warnings, blue for info, etc.; which can be customized). ✔

The error level is added to the `theme` list, thereby allowing the components to be styled. 

> The handling of ValidationResults, which are "ok" (ErrorLevel null), does not change ✔
Warnings and info results are not considered errors for the validation logic:  
https://github.com/vaadin/flow/blob/59fd55ec5f0e51d10ec806c655e8dae628f00cef/flow-data/src/main/java/com/vaadin/flow/data/binder/ValidationResult.java#L107-L110

This approach requires some support from the theme (not included in this PR):

- Define global CSS variables: something yellow-ish for warnings, blue for info (critical and system are not included, thus they fall back to the default style for `invalid`)

```
html {
	--lumo-warning-text-color: #9a6e1f; 	 
	--lumo-warning-color-50pct: #ffb53380;
	--lumo-warning-color-10pct: #ffb53333;
	
	--lumo-info-text-color: var(--lumo-primary-color);
	--lumo-info-color-50pct: var(--lumo-primary-color-50pct);
	--lumo-info-color-10pct: var(--lumo-primary-color-10pct);
}
```

- Style vaadin-text-field (which covers many other textfield-based components: vaadin-combobox, vaadin-date-picker, etc.) with:

```
:host([theme~='warning']) {
	--lumo-error-text-color:  var(--lumo-warning-text-color);
	--lumo-error-color-10pct: var(--lumo-warning-color-10pct);
	--lumo-error-color-50pct: var(--lumo-warning-color-50pct);	
}

:host([theme~='info']) {
	--lumo-error-text-color:  var(--lumo-info-text-color);
	--lumo-error-color-10pct: var(--lumo-info-color-10pct);
	--lumo-error-color-50pct: var(--lumo-info-color-50pct);	
}
```

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -42,8 +42,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.HasElement;
 import com.vaadin.flow.component.HasText;
 import com.vaadin.flow.component.HasValidation;
+import com.vaadin.flow.component.HasTheme;
 import com.vaadin.flow.component.HasValue;
 import com.vaadin.flow.component.HasValue.ValueChangeEvent;
 import com.vaadin.flow.component.HasValue.ValueChangeListener;
@@ -51,6 +53,7 @@ import com.vaadin.flow.component.UI;
 import com.vaadin.flow.data.converter.Converter;
 import com.vaadin.flow.data.converter.StringToIntegerConverter;
 import com.vaadin.flow.data.validator.BeanValidator;
+import com.vaadin.flow.dom.ThemeList;
 import com.vaadin.flow.function.SerializableConsumer;
 import com.vaadin.flow.function.SerializableFunction;
 import com.vaadin.flow.function.SerializablePredicate;
@@ -2545,6 +2548,7 @@ public class Binder<BEAN> implements Serializable {
             HasValidation fieldWithValidation = (HasValidation) field;
             fieldWithValidation.setInvalid(true);
             fieldWithValidation.setErrorMessage(result.getErrorMessage());
+            setErrorTheme(fieldWithValidation, result);
         }
     }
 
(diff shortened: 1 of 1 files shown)


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -3156,4 +3161,33 @@ public void setValidatorsDisabled(boolean validatorsDisabled) {
     public boolean isValidatorsDisabled() {
         return validatorsDisabled;
     }
+
+    private static String getThemeName(ErrorLevel errorLevel) {
+        return errorLevel.name().toLowerCase();
+    }
+
+    private static Optional<ThemeList> getThemes(HasValidation field) {
+        if (field instanceof HasTheme) {
```
denis-anisimov, line 3170: This line shows that this PR overall makes `Binder` effectively depend on `UI` (even though there is already `HasValidation` dependency but this is also bad thing).

`Binder` should not know and depend on any specifics of the `UI`. It's initially designed to work with `HasValue`  which is not a `UI`  class at all.

`HasValidation` has been introduced as a necessary evil here assuming that this is the only one UI knowledge which should be used here.

Now it's not true anymore: you are trying to add `HasTheme`  UI abstraction.
And now the question arises : do we want to modify `Binder`  every time when new `UI`  abstraction is invented ?
My answer is **no**.

So what I suggest: 
- introduce a new abstraction for a Binder which will be executed inside  `Binder::handleError` and `Binder::clearError`
- it may be `BiConsumer`  or some custom interface (may be it will require two methods).
- Make accessors for this interface in `Binder`  : setter and getter
- Use this interface in the  `Binder::handleError` and `Binder::clearError` if it's available and execute its methods.
- Provide a default implementation for it.

That will allow to the user to change the implementation of the interface to anything he wants. And it allows to extract UI classes dependencies from the Binder via abstraction layer.

You can do it here in this PR or make another one.
Also you may use a ticket to discuss the details/design of this suggestion.

Also any PR requires tests.

denis-anisimov (changes requested):

denis-anisimov:
Closed in favor of https://github.com/vaadin/flow/pull/11384

PR #10700, opened 2021-04-16
test: Add support for passing additional options to headless chrome

Artur-:


flow-test-util/src/main/java/com/vaadin/flow/testutil/AbstractParallelTestBenchTest.java
```diff
@@ -71,12 +71,17 @@ public void setup() throws Exception {
...
         } else if (Browser.CHROME == getRunLocallyBrowser()
                 && !isJavaInDebugMode()) {
-            setDriver(createHeadlessChromeDriver());
+            setDriver(createHeadlessChromeDriver(getAdditionalHeadlessChromeOptions()));
             return;
         }
         super.setup();
     }
 
+    protected String[] getAdditionalHeadlessChromeOptions() {
```
mshabarov, line 80: Would be useful to add a javadoc to this method and mention that the given options are only applied to a locally running browser.

mshabarov (changes requested):

flow-test-util/src/main/java/com/vaadin/flow/testutil/AbstractParallelTestBenchTest.java
joheriks: In addition, in the interest of generality and reusability, would it be an idea to return a `ChromeOptions` instead of an array of strings and use `ChromeOptions:merge`?

Artur-: Fixed to use `ChromeOptions`

mshabarov (approved):

flow-test-util/src/main/java/com/vaadin/flow/testutil/AbstractParallelTestBenchTest.java
Artur-: But this is actually in a way much worse as it does not solve the original problem. Merging two options will overwrite `args` from the first one so you cannot add a new parameter, you need to redefine them all

Artur-: So how about this version instead?

joheriks: 😞 didn't check, sorry, then the original was flag merge better.

Artur-: The latest version should support both cases

mshabarov (approved):

PR #10701, opened 2021-04-16
fix: properly remove block comments from templates

### LINKED ISSUE #10673: Comments in LitTemplate can produce an invalid warning
In a LitTemplate, adding a comment (probably a multiline one that includes '*' character) produces a misleading warning: `No lit template data found in foo.ts sources`

Create a LitElement that includes a multiline comment. For example,
```
import { html, LitElement } from 'lit-element';

export class HelloLit extends LitElement {
  /* ******************************************************
   * comment
   * ******************************************************/

  render() {
    return html` <div>Some content</div>`;
  }
}

customElements.define('hello-lit', HelloLit);
```

Then create a corresponding Java class. For example,
```
@Tag("hello-lit")
@JsModule("./src/hello-lit.ts")
public class HelloLit extends LitTemplate {
}
```

and add this LitTemplate inside a view and navigate to this view.

### Expected behavior
The view is rendered without any error.

### Actual behavior
The view is rendered correctly. However, the following error is produced in the logs:
```
2021-04-15 14:16:17.060  WARN 3832 --- [nio-8080-exec-7] c.v.f.c.littemplate.BundleLitParser      : No lit template data found in ./src/hello-lit.ts sources.
2021-04-15 14:16:17.060  INFO 3832 --- [nio-8080-exec-7] c.v.f.c.l.i.LitTemplateParserImpl        : Couldn't find the definition of the element with tag 'hello-lit' in any lit template file declared using '@JsModule' annotations. Check the availability of the template files in your WAR file or provide alternative implementation of the method LitTemplateParser.getTemplateContent() which should return an element representing the content of the template file
2021-04-15 14:16:17.060  INFO 3832 --- [nio-8080-exec-7] c.v.f.c.l.LitTemplateDataAnalyzer        : Couldn't parse template for class org.vaadin.tarek.components.custom.HelloLit class. Only specific Lit template format is supported. Please check that your template definition directly contains 'render' method which returns html`_template_content_`.
```
### Versions:
    - Vaadin / Flow version: 14.5.2
    - Java version: 11
    - OS version: Windows

denis-anisimov:
fixes #10673


flow-server/src/test/java/com/vaadin/flow/internal/StringUtilTest.java
```diff
@@ -63,4 +65,29 @@ public void commentRemoval_emojiInString_removalDoesnotThrowResultIsTheSame() {
...
+               + "}\n"
+               + "\n"
+               + "customElements.define('hello-lit', HelloLit);");
+        // @formatter:on
+        MatcherAssert.assertThat(result, CoreMatchers.allOf(
+                CoreMatchers.containsString("<div>Some content</div>"),
+                CoreMatchers.containsString(
+                        "customElements.define('hello-lit', HelloLit);"),
+                CoreMatchers.not(CoreMatchers.containsString("comment"))));
+    }
```
caalador, line 91: Even though this exhibits the error it's not clear why this would fail when the other block tests do not.
The actual problem comes when there is a n%2 amount of asterisks before the / as the check eats up one character after each asterisk.

Also the test didn't show in the failing assertion that the actual problem is that after the block comment there is no string left.

A clearer test would be:
```
    @Test
    public void endCommentCheck_shouldNotEatCharacters(){
        String blockComment = StringUtil
            .removeComments("return html'/* comment **/';");
        Assert.assertEquals("return html'';", blockComment);
    }
```
Which on failure will give a clearer exception where one can see that the whole end of the string is missing.
```
org.junit.ComparisonFailure: 
Expected :return html'Code';
Actual   :return html'
```

flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
```diff
@@ -80,11 +80,21 @@ public final static String removeComments(String code) {
...
-            case IN_BLOCK_COMMENT:
-                if (character.equals("*") && scanner.hasNext()
-                        && scanner.next().equals("/")) {
+            case IN_BLOCK_COMMENT_EVEN_ASTERISKS:
+                if (character.equals("/")) {
                     state = State.NORMAL;
-                    break;
+                }
+                break;
+            case IN_BLOCK_COMMENT:
```
caalador, line 88:
```
            case IN_BLOCK_COMMENT:
                if (character.equals("*") && scanner.hasNext("/")) {
                    scanner.next();
                    state = State.NORMAL;
                    break;
                }
            }
```
Fixes this issue as the problem is not the asterisks inside the block, but the amount of asterisks just before the end as the use of next was not thought to be a problem even if it eats up a charater.

caalador (changes requested):

flow-server/src/test/java/com/vaadin/flow/internal/StringUtilTest.java
denis-anisimov: Agreed with this.

flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
denis-anisimov: I've been looking for something like `putBack`  but `hasNext`  does similar thing.
Apparently `Scanner`  is more advanced than I thought.

OK, done. Thanks.

caalador (approved):

PR #10707, opened 2021-04-18
refactor: split AccessAnnotationChecker out from VaadinConnectAccessChecker

Artur-:


fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/EndpointUtil.java
```diff
@@ -104,7 +104,8 @@ public boolean isAnonymousEndpoint(HttpServletRequest request) {
             return false;
         }
 
-        return accessChecker.getSecurityTarget(method.get())
+        return accessChecker.getAccessAnnotationChecker()
+                .getSecurityTarget(method.get())
                 .isAnnotationPresent(AnonymousAllowed.class);
```
haijian-vaadin, line 109: Could this be wrapped into the AccessAnnotationChecker? feeling a bit uncomfortable that the chain is getting too long here. 
Would it better if it is `accessChecker.getAccessAnnotationChecker().isAnonymousEndpoint(method.get());`?

Artur-: If we change the checker interface to take "principal" and "roleLookup" instead of a HttpRequest, then we can skip the separate logic and just check if the method is accessible without a Principal.

Artur-: Refactored

haijian-vaadin (approved):

PR #10716, opened 2021-04-19
fix: report the parameter name in the exception message

### LINKED ISSUE #9994: RouteParameters.getLong("param") should provide the parameter name in the exception message
When `event.getRouteParameters().getLong("param")` fails, it should provide the parameter name in the exception message

```
@Route("foo/:myparam/bar")
public class ExampleRoute extends VerticalLayout implements BeforeEnterObserver {
    @Override
        System.out.println(event.getRouteParameters().getLong("myparam").get());
    }
}
```

### Expected behavior
The `NumberFormatException` should be wrapped in an `IllegalArgumentException` which should clearly state the parameter name.
### Actual behavior
Current message doesn't contain the parameter name `myparam`
```
java.lang.NumberFormatException: For input string: "kvak"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Long.parseLong(Long.java:692)
	at java.base/java.lang.Long.valueOf(Long.java:1144)
	at com.vaadin.flow.router.RouteParameters.getLong(RouteParameters.java:155)
	at org.vaadin.example.ExampleRoute.beforeEnter(Cicina.java:13)
```

### Versions:
    - Vaadin / Flow version: 14.5.0.alpha3

denis-anisimov:
fixes #9994

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/RouteParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/RouteParameters.java
@@ -162,7 +162,7 @@ public final class RouteParameters implements Serializable {
             return Optional.of(Long.valueOf(value));
         } catch (NumberFormatException e) {
             throw new IllegalArgumentException(String.format(
-                    "Couldn't parse '%s' paramter value '%s' as long",
+                    "Couldn't parse '%s' parameter value '%s' as long",
                     parameterName, value), e);
         }
     }
--- a/flow-server/src/main/java/com/vaadin/flow/router/RouteParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/RouteParameters.java
@@ -137,7 +137,7 @@ public final class RouteParameters implements Serializable {
             return Optional.of(Integer.valueOf(value));
         } catch (NumberFormatException e) {
             throw new IllegalArgumentException(String.format(
-                    "Couldn't parse '%s' paramter value '%s' as integer",
+                    "Couldn't parse '%s' parameter value '%s' as integer",
                     parameterName, value), e);
         }
     }
--- a/flow-server/src/main/java/com/vaadin/flow/router/RouteParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/RouteParameters.java
@@ -133,7 +133,13 @@ public final class RouteParameters implements Serializable {
             return Optional.empty();
         }
 
-        return Optional.of(Integer.valueOf(value));
+        try {
+            return Optional.of(Integer.valueOf(value));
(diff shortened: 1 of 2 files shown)


flow-server/src/test/java/com/vaadin/flow/router/RouteParametersTest.java
```diff
@@ -78,48 +84,63 @@ public void getters_provide_empty_values() {
...
-    public void integer_getter_throws_exception() {
+    public void integer_getter_stringParameter_throws() {
         RouteParameters parameters = getParameters();
 
-        try {
-            parameters.getInteger("string");
+        exception.expect(IllegalArgumentException.class);
+        exception.expectMessage(
+                CoreMatchers.allOf(CoreMatchers.containsString("'string'"),
+                        CoreMatchers.containsString("'foo'")));
```
caalador, line 93: I would rather expect the full message in all tests as it's not that long.

flow-server/src/main/java/com/vaadin/flow/router/RouteParameters.java
```diff
@@ -152,7 +158,13 @@ public static RouteParameters empty() {
             return Optional.empty();
         }
 
-        return Optional.of(Long.valueOf(value));
+        try {
+            return Optional.of(Long.valueOf(value));
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException(String.format(
+                    "Couldn't parse '%s' paramter value '%s' as long",
```
caalador, line 165:
```suggestion
                    "Couldn't parse '%s' parameter value '%s' as long",
```

flow-server/src/main/java/com/vaadin/flow/router/RouteParameters.java
```diff
@@ -133,7 +133,13 @@ public static RouteParameters empty() {
             return Optional.empty();
         }
 
-        return Optional.of(Integer.valueOf(value));
+        try {
+            return Optional.of(Integer.valueOf(value));
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException(String.format(
+                    "Couldn't parse '%s' paramter value '%s' as integer",
```
caalador, line 140:
```suggestion
                    "Couldn't parse '%s' parameter value '%s' as integer",
```

caalador (changes requested):

flow-server/src/test/java/com/vaadin/flow/router/RouteParametersTest.java
denis-anisimov: done

caalador (approved):

PR #10718, opened 2021-04-19
chore: Move theme live reload tests to it's own profile

mshabarov:
Adds a new category for theme live reload tests and skips them by default.
Creates a new profile for running only those tests separately.


pleku (commented):
So how are the theme reload tests executed ?

mshabarov:
> So how/when are the theme reload tests executed ?

I plan to create one more step in Flow snapshot and PR validation builds, which will do `mvn clean verify -Ptheme-live-reload-tests
` on target `flow-tests/test-application-theme/pom.xml`.
This step will go right after main `Run validation` step.

pleku (approved):
OK I suppose you're ready to do it immediately

mshabarov:
> OK I suppose you're ready to do it immediately

Sure thing

PR #10734, opened 2021-04-20
chore: Clarify idempotence of Registrations

### LINKED ISSUE #9978: Clarify idempotence of Registrations in JavaDoc
Since #8506 has been fixed, a `Registration` is supposed to be idempotent (or [idempotent-ish](https://github.com/vaadin/flow/issues/8466#issuecomment-636516695) at least), i.e. calling `remove()` multiple times has the same effect as calling it once.

However, this is not mentioned explicitely in the documentation:

https://github.com/vaadin/flow/blob/c009afc5c24d8a3d6abc14f6f5cd3faea80c1932/flow-server/src/main/java/com/vaadin/flow/shared/Registration.java#L26-L37

My interpretation of the current JavaDoc would be exactly the idempotent-ish behavior described by @Legioth, but I think it would be better to be more explicit about it to provide clear guidelines how on to use or implement `Registration`s.

For example, one may wonder if a `Registration` field should be nulled out after calling `remove`, or whether a custom callback must be wrapped with `Registration.once` or not.

denis-anisimov:
fixes #9978

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/shared/Registration.java
+++ b/flow-server/src/main/java/com/vaadin/flow/shared/Registration.java
@@ -33,6 +33,10 @@ import com.vaadin.flow.server.Command;
 public interface Registration extends Serializable {
     /**
      * Removes the associated listener from the event source.
+     * <p>
+     * The {@code remove} method called after removal does nothing.
+     * 
+     * @see #once(Command)
      */
     void remove();
 


flow-server/src/main/java/com/vaadin/flow/shared/Registration.java
```diff
@@ -33,6 +33,10 @@
 public interface Registration extends Serializable {
     /**
      * Removes the associated listener from the event source.
+     * <p>
+     * The {@code remove} method called after removal does nothing.
```
caalador, line 37: Is this really right? Shouldn't it say that the method should not do anything after first call. As this is an interface method it's up to the implementation to decide if calling remove multiple times does different things.

denis-anisimov: Kind of yes and no.

I agree that it's up to implementation to decide.

But on the other hand: it's indication for implementation how to implement the method and also it's API method which describe what behavior should be expected.

It'a aligned with `InputStream::close`  method: even though it's not an interface but an abstract class.
The method is abstract : not implemented. And the javadocs says that the method is idempotent.

So even though I kind of agree that it's up to impl, but on the other hand I see this as a part of contract: the impl has to respect idempotence of the method.

caalador (approved):

PR #10745, opened 2021-04-20
chore: Define default profile on root level

mshabarov:
Sets the `default` profile on a maven parent level to be able to skip some test categories (theme live reload in this case) by default when building flow snapshot in CI.

### DIFF
--- a/flow-server/pom.xml
+++ b/flow-server/pom.xml
@@ -261,9 +261,6 @@
     <profiles>
         <profile>
             <id>default</id>
-            <activation>
-                <activeByDefault>true</activeByDefault>
-            </activation>
             <properties>
                 <test.excludegroup>com.vaadin.flow.testcategory.ContainerTests,
                     com.vaadin.flow.testcategory.SlowTests
--- a/flow-server/pom.xml
+++ b/flow-server/pom.xml
@@ -261,6 +261,9 @@
     <profiles>
         <profile>
             <id>default</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
             <properties>
                 <test.excludegroup>com.vaadin.flow.testcategory.ContainerTests,
                     com.vaadin.flow.testcategory.SlowTests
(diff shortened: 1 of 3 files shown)


caalador:
It shouldn't work that way as profiles in parent and child POM do not interact so having activeByDefault in parent doesn't make it active in the child.
See https://stackoverflow.com/a/63174395/13523483 for more details.

mshabarov:
> It shouldn't work that way as profiles in parent and child POM do not interact so having activeByDefault in parent doesn't make it active in the child.
See https://stackoverflow.com/a/63174395/13523483 for more details.

Yes, you're right and thanks for the link!
These changes don't make sense.
For testing, I put some echoing into maven poms
```
            <plugin>
                <artifactId>maven-antrun-plugin</artifactId>
                <executions>
                    <execution>
                        <phase>generate-resources</phase>
                        <goals>
                            <goal>run</goal>
                        </goals>
                        <configuration>
                            <tasks>
                                <echo>flow-tests property: ${test.excludegroup}</echo>
                            </tasks>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```

and it shows that even without my changes, all the needed properties are visible 

```
[INFO] Executing tasks

main:
     [echo] flow-tests property: com.vaadin.flow.testcategory.ScreenshotTests,
     [echo]                     com.vaadin.flow.testcategory.PushTests,
     [echo]                     com.vaadin.flow.testcategory.ThemeLiveReloadTests

```

But for some reason the IT's marked by `PushTests` and `ThemeLiveReloadTests` are still executed in flow snap validation.

This is not the case for `-Pvalidation` profile in PR validation though.

Don't know what is the reason.

mshabarov:
Tested with Flow snapshot build: adding `-Pdefault` to the `Run validation` build activates the profile (obviously) and skips the desired tests.

So, to have skipped theme live reload tests by default on a parent pom level I simply added a duplicate of this profile with only theme test category.

caalador (approved):

PR #10750, opened 2021-04-21
test: unignore tests in PopStateHandlerIT

### LINKED ISSUE #10485: Fusion : PopStateHandlerIT fails
`PopStateHandlerIT` had a number of `hasClientIssue("7572")` which effectively disabled some piece of test in every test method.

I'm going to make a PR and change the test to disable the parts which are failing .

But the all the tests should work in the end. The fix for this ticket should remove all disabled parts (and may be just return the test state to the same as it has e.g. in `2.5` branch).

Similar ticket: https://github.com/vaadin/flow/issues/7572.

For some reason #7572 doesn't mention `PopStateHandlerIT` anyhow.

platosha:
Fixes #10485

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/JavaScriptBootstrapUI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/JavaScriptBootstrapUI.java
@@ -49,7 +49,7 @@ import elemental.json.JsonValue;
 /**
  * Custom UI for {@link JavaScriptBootstrapHandler}. This class is intended for
  * internal use in client side bootstrapping.
- * 
+ *
  * <p>
  * For internal use only. May be renamed or removed in a future release.
  */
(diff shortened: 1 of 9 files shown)


flow-server/src/main/java/com/vaadin/flow/component/page/History.java
```diff
@@ -182,11 +182,14 @@ public void pushState(JsonValue state, String location) {
      *            to only change the JSON state
      */
     public void pushState(JsonValue state, Location location) {
+        final String pathWithQueryParameters = Optional.ofNullable(location)
```
haijian-vaadin, line 185: could extract this into a separate method? since it's also used in the `replaceState()`?

haijian-vaadin: maybe would be a bit better to use simple if-else with some comments to explain why it's now needed to get the lastHandledLocation from UI (not before), which would help to understand the code better than the Optional.orElse()

flow-server/src/main/java/com/vaadin/flow/component/internal/JavaScriptBootstrapUI.java
```diff
@@ -117,23 +120,34 @@ public void connectClient(String clientElementTag, String clientElementId,
...
+                    event -> renderViewForRoute(event.getLocation(),
+                            NavigationTrigger.CLIENT_SIDE));
+
+            // Render the flow view that the user wants to navigate to.
+            renderViewForRoute(new Location(flowRoute),
+                    NavigationTrigger.CLIENT_SIDE);
+        } else {
+            History.HistoryStateChangeHandler handler = getPage().getHistory()
+                    .getHistoryStateChangeHandler();
+            // FIXME(platosha): pass the history state from the client side
```
haijian-vaadin, line 141: Does this work without client-side change already? If not, let's remove the code and make a separate ticket about it.

haijian-vaadin (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/page/History.java
platosha: I’m trying to keep the code style consistent, there are other lines like that in the same file:
```java
        pushState(state,
                Optional.ofNullable(location).map(Location::new).orElse(null));
```

Actually it is not necessary to get last location from UI, removed that.

What I’m after here is to support `null` location argument in this method. as the jsdoc explicitly allowed.

flow-server/src/main/java/com/vaadin/flow/component/internal/JavaScriptBootstrapUI.java
platosha: It does work, but does not pass the state. That was just a self reminder to get back to that — now fixed that.

flow-server/src/main/java/com/vaadin/flow/component/internal/JavaScriptBootstrapUI.java
```diff
@@ -117,23 +124,33 @@ public void connectClient(String clientElementTag, String clientElementId,
...
+            if (!trimmedRoute.equals(flowRoute)) {
+                // See InternalRedirectHandler invoked via Router.
+                getPage().getHistory().replaceState(null, trimmedRoute);
+                flowRoute = trimmedRoute;
+            }
 
-        // Render the flow view that the user wants to navigate to.
-        renderViewForRoute(new Location(trimmedRoute),
-                NavigationTrigger.CLIENT_SIDE);
+            getPage().getHistory().setHistoryStateChangeHandler(
```
haijian-vaadin, line 135: Not sure about this. What if the user has defined a `HistoryStateChangeHandler` in the app, this would override the one defined by the user?

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/PopStateHandlerIT.java
```diff
@@ -78,7 +77,7 @@ public void testSamePathHashChanges_noServerSideEvent() {
     }
 
     @Test
-    @Ignore("Ignored because of fusion issue: https://github.com/vaadin/flow/issues/10485")
+    @Ignore("Ignored because if the issue https://github.com/vaadin/flow/issues/10825")
```
haijian-vaadin, line 80:
```suggestion
    @Ignore("Ignored because of the issue https://github.com/vaadin/flow/issues/10825")
```

haijian-vaadin (changes requested):

haijian-vaadin (changes requested):
Looks like the failing test is about a trailing slash, if I change the `assertLastMessage("/routerlink",`  to `assertLastMessage("/routerlink/",`
in line 48 of `NavigationTriggerIT`, should we do that and then merge the PR?

haijian-vaadin (approved):

Issue #10525, opened 2021-03-31
DX testing feedback for custom theme in 14.6


pleku:
Docs fixes to be done:
- [x] `@Theme` annotation docs should point out in (v14) that is should be placed in the root layout. It is mentioned in the annotation javadocs, but only as last line
  - Check the same for master docs 
- [x] `@Theme` annotation javadocs should mention **earlier** that is should be in the root layoyut, now it is at the end of a very long javadoc and nobody reads that far
  - Check the same for Flow 6.0/master
- [ ] It should be highlighted at least for 14.6 that the `@Theme` annotation works differently in 19+ because the default value is not the theme class but the theme folder name

Further feedback:
- Supporting `TextField.css` instead of `vaadin-text-field.css` inside `themes/my-theme/components/` ? IMO sounds like a bad idea since we cannot know the needed tag name from the Java component name (easily) and there might be several components with the same name on the classpath, so to achieve uniqueness it would have to use package names too, and that would be painful compared to the tag name that is unique already and can be seen from the component docs

caalador:
`It should be highlighted at least for 14.6 that the @Theme annotation works differently in 19+` 
Where should this be highlighted?

pleku:
Lets leave this type of things to a migration documentation between 14 and nextLTS

PR #10751, opened 2021-04-21
chore: update theme javadoc

caalador:
Update the javadoc to have the annotation
placement earlier and more visible.


taefi:
This fixes the first part of #10525

taefi (approved):

PR #10754, opened 2021-04-21
feat!: generate nullable TS types for Java types by default

### LINKED ISSUE #8849: Generate nullable TS types for nullable Java types
When accessing properties of a TypeScript type generated from a Java type, I want properties that are nullable in Java to remain nullable in TypeScript, so that any existing Java POJO type can be used in an endpoint without any modifications (even if the some properties are `null` in Java at the run time).

This effectively reverts the decision made in https://github.com/vaadin/flow/issues/6717 (to make the TS types non-nullable by default). The main reasons to revert that decision are:

1. When all TS types are non-nullable by default, there needs to be a way to change the default to generate a nullable type sometimes. The current implementation relies on the `Optional<T>` wrapping, which is not ideal for the reasons explained in https://github.com/vaadin/flow/issues/7371 and https://vaadin.slack.com/archives/CDNTYMS9H/p1592226597202000. That issue suggest to rely on `@Nullable` instead of `Optional<T>`.

    Relying on `@Nullable` leads to a situation where one would need to write very confusing code like `@Nullable @NotNull MyEntity nullableNotNullEntity` in order to create a form with a combo-box that's initially empty, but requires a value (see https://github.com/vaadin/flow/pull/8835#discussion_r468081943).

    Dropping `@Nullable` and making all properties nullable by default resolves that issue.

1. When all TS types are non-nullable by default, the Java types that are outside of developers' control (3rd party, external modules, etc) may become unusable with endpoints. If they include a single `null` property, an attempt to return such a POJO from an endpoint would result in a runtime exception (see https://github.com/vaadin/flow/pull/6891).

    Dropping `@Nullable` and making all properties nullable by default resolves that issue.

The downsides of making all properties of the generated TS types nullable by default are explained in https://github.com/vaadin/vaadin-connect/issues/404:

1. When all TS types are nullable by default, the generated types become more complicated. For example, `List<Map<String, Map<String, Version>>> prop` in Java becomes `prop?: Array<{ [key: string]: { [key: string]: Version | undefined; } | undefined; } | undefined>` in TypeScript.

1. When all TS types are nullable by default, in TypeScript code developers need to check for `undefined` every time when they address any (sub) property of a POJO returned from an endpoint. For example, `const pojo = await Endpoint.getMeMyPojo(); if (pojo.name) { doStuff(pojo.name) }`.

    This downside is somewhat alleviated by the [optional chaining (`?.`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and [nullish coallescing (`??`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_Coalescing_Operator) operators, that were added to the language recently.

Not a downside, but also relevant:

1. When all TS types are nullable by default, there needs to be a way to change the default and generate a non-nullable property sometimes. A reasonable approach would be to rely on some form of `@Nonnull` or `@NotNull` annotation. For example `@Notnull List<@Notnull Map<String, @Notnull Map<String, @Notnull Version>>> prop` to generate `prop: Array<{ [key: string]: { [key: string]: Version; }; }>`.

1. ~When all TS types are nullable by default, there is no need to run run-time checks on the values returned by the endpoints for `null` property values. The [null checker](https://github.com/vaadin/flow/pull/6891) will need to be removed.~

    UPD: The null checker is still needed to check that `@Nonnull`-annotated properties are indeed non null at the run time.

Lodin:
Given a Java class as
```java
public class Contact {
    private String name = "";
    private Company company;
}
```
the generated TypeScipt properties would be nullable by default, like
```typescript
export default interface Contact {
  company?: Company;
  name?: string;
}
```
instead of non-nullable by default, as in previous versions
```typescript
export default interface Contact {
  company: Company;
  name: string;
}
```
One can still mark a field as non-nullable by using the `@Nonnull` annotation, for example, 
```java
public class Contact {
    @Nonnull
    private String name = "";
    private Company company;
}
```

Resolves #8849.


fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeCheckerHelper.java
```diff
@@ -240,18 +237,12 @@ private boolean isPropertySubjectForChecking(
             Field field = readMethod.getDeclaringClass().getDeclaredField(name);
             return !Modifier.isStatic(field.getModifiers())
                     && !Modifier.isTransient(field.getModifiers())
-                    && !field.isAnnotationPresent(JsonIgnore.class)
-                    && !isNullable(field);
+                    && !field.isAnnotationPresent(JsonIgnore.class);
```
haijian-vaadin, line 240: Missing a `isRequired()` Check on the field? so that property would subject for checking if it's marked as NotNull/Nonnull explicitly.

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ConnectUtils.java
```diff
@@ -0,0 +1,13 @@
+package com.vaadin.flow.server.connect;
+
+import java.lang.reflect.AnnotatedElement;
+
+import javax.annotation.Nonnull;
+import javax.validation.constraints.NotNull;
+
+public final class ConnectUtils {
```
haijian-vaadin, line 8: Connect is the legacy name for Fusion, let's avoid using that. Sth like EndpointGeneratorutils.
Or maybe we can move this into the existing GeneratorUtils and mark that as public.

haijian-vaadin (changes requested):

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeCheckerHelper.java
Lodin: Should I check it this way then (pseudo names used)?
```java
return (!Modifier.isStatic(field.getModifiers())
                    && !isNullable(field);
                    && !field.isAnnotationPresent(JsonIgnore.class)) 
                    || isRequired(field);
```

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/ExplicitNullableTypeCheckerTest.java
```diff
@@ -456,7 +458,16 @@ public Long methodWithIdAnnotation() {
     /**
      * Method for testing
      */
-    public String stringNotNullable() {
+    @Nonnull
+    public String stringNotNull() {
```
platosha, line 462: Identical to the one below. Maybe remove?

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/Annotation.java
```diff
@@ -0,0 +1,17 @@
+package com.vaadin.flow.server.connect;
+
+import java.lang.reflect.AnnotatedElement;
+
+import javax.annotation.Nonnull;
+
+import com.github.javaparser.ast.nodeTypes.NodeWithAnnotations;
+
+public final class Annotation {
```
haijian-vaadin, line 9: Annotation is too general here, maybe `NonNullChecker` or sth similar would be better?

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/generator/endpoints/AbstractEndpointGenerationTest.java
```diff
@@ -349,16 +350,14 @@ private void assertRequestSchema(Schema requestSchema,
             assertSchema(stringSchemaEntry.getValue(), parameterTypes[index],
                     parameterTypeArguments.get(index));
             List requiredList = requestSchema.getRequired();
-            if (parameters[index].isAnnotationPresent(Nullable.class)
-                    || Optional.class
-                            .isAssignableFrom(parameters[index].getType())) {
+            if (parameters[index].isAnnotationPresent(NotNull.class)
```
haijian-vaadin, line 353: shouldn't we only check `@Nonnull` as agreed?

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/generator/endpoints/AbstractEndpointGenerationTest.java
```diff
@@ -509,15 +508,14 @@ private void assertSchemaProperties(Class<?> expectedSchemaClass,
             Type type = expectedSchemaField.getGenericType();
             assertSchema(propertySchema, expectedSchemaField.getType(),
                     extractTypeArguments(type, typeArguments));
-            if (Optional.class.isAssignableFrom(expectedSchemaField.getType())
-                    || expectedSchemaField
-                            .isAnnotationPresent(Nullable.class)) {
+            if (expectedSchemaField.isAnnotationPresent(NotNull.class)
```
haijian-vaadin, line 511: shouldn't we only check `@Nonnull` as agreed?

haijian-vaadin (changes requested):

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeCheckerHelper.java
```diff
@@ -238,20 +234,15 @@ private boolean isPropertySubjectForChecking(
             }
 
             Field field = readMethod.getDeclaringClass().getDeclaredField(name);
-            return !Modifier.isStatic(field.getModifiers())
+            return (!Modifier.isStatic(field.getModifiers())
                     && !Modifier.isTransient(field.getModifiers())
-                    && !field.isAnnotationPresent(JsonIgnore.class)
-                    && !isNullable(field);
+                    && !field.isAnnotationPresent(JsonIgnore.class))
+                    || isElementRequired(field);
```
haijian-vaadin, line 240: should be && instead of ||?

haijian-vaadin (changes requested):

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/generator/IndentationUtils.java
```diff
@@ -149,9 +176,23 @@ public void addSymbol(char symbol) {
         public String getContent() {
             return content.toString().trim();
         }
+
+        public boolean isEmpty() {
```
haijian-vaadin, line 180: also here, no need to have it as public.

fusion-endpoint/src/test/resources/com/vaadin/flow/server/connect/generator/endpoints/collectionendpoint/expected-CollectionEndpoint.ts
```diff
@@ -9,7 +9,7 @@ import client from './connect-client.default';
 import Collection from './com/vaadin/flow/server/connect/generator/endpoints/collectionendpoint/CollectionEndpoint/Collection';
 
 /**
- * Get a collection by author name. The generator should not mix this type with the Java's Collection type.
+ * Get a collection by author name. The generator should not mix this type  with the Java's Collection type.
```
haijian-vaadin, line 12: looks like we shouldn't add the extra space, the test is failing on CI because of this, the test goes green after removing the extra space.

fusion-endpoint/src/test/resources/com/vaadin/flow/server/connect/generator/endpoints/modelpackage/expected-ModelPackageEndpoint.ts
```diff
@@ -19,7 +19,7 @@ function _getListOfUserName(): Promise<Array<string> | undefined> {
 export {_getListOfUserName as getListOfUserName};
 
 /**
- * Get a collection by author name. The generator should not mix this type with the Java's Collection type.
+ * Get a collection by author name. The generator should not mix this type  with the Java's Collection type.
```
haijian-vaadin, line 22: looks like we shouldn't add the extra space, the test is failing on CI because of this, the test goes green after removing the extra space.

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/NonnullAnnotationChecker.java
```diff
@@ -6,7 +6,7 @@
 
 import com.github.javaparser.ast.nodeTypes.NodeWithAnnotations;
 
-public final class Annotation {
+public final class NonnullAnnotationChecker {
```
haijian-vaadin, line 9: Sorry, after another thought, I was thinking, maybe we can just move these two methods into `ExplicitNullableTypeChecker`? since the required checks are quite relevant to that class.

Also as I mentioned in the previous review, both methods could be named as `isRequired()`, no need to put the parameter type into the method name.

haijian-vaadin (changes requested):

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
```diff
@@ -16,16 +16,35 @@
...
+     * 
+     * @param element
+     *            an element to be required
+     * @return a result of check
+     */
+    public static boolean isRequired(AnnotatedElement element) {
+        return element.isAnnotationPresent(Nonnull.class);
+    }
+
+    public static boolean isRequired(NodeWithAnnotations<?> node) {
```
haijian-vaadin, line 44: Missing JavaDoc for this method. Each public method should have JavaDoc, otherwise, CI would fail.

haijian-vaadin (changes requested):

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/generator/IndentationUtils.java
```diff
@@ -0,0 +1,325 @@
+package com.vaadin.flow.server.connect.generator;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Stack;
+
+final class IndentationUtils {
```
haijian-vaadin, line 8: hmm, looks like this class is only used for test? Let's move it from the src/main/java to src/main/test

haijian-vaadin (changes requested):

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/generator/IndentationUtils.java
Lodin: Yes, that class is only for tests, and it is already in `src/test/java`.

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
Lodin: Hmm, how should I do that? Just copy-paste the JavaDoc from the sibling method?

haijian-vaadin: Yes

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/generator/IndentationUtils.java
haijian-vaadin: ah, sorry, my bad.

haijian-vaadin (dismissed):

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
```diff
@@ -16,20 +16,43 @@
...
+    /**
+     * Checks if the element should be required (not nullable) in the generated
+     * Typescript code.
+     * 
+     * @param element
+     *            an element to be required
+     * @return a result of check
+     */
+    public static boolean isRequired(AnnotatedElement element) {
+        return element.isAnnotationPresent(Nonnull.class);
```
knoobie, line 41: What do you think about changing this from `javax.annotation.Nonnull` to a string check against "Nonnull" /  "NonNull" and "Notnull" / "NotNull" ? This would remove the direct Dependency on javax (having Jakarta namespace in mind) and additionally allowing the usage of other annotations like e.g.:

- custom annotation 
- Jetbrains annotation 
- Spring annotation 
- Findbugs annotation 
- Lombok annotation 
- Micrometer annotation 

and so on.

haijian-vaadin: Yes, that's a very good idea, we should do that. ping @Lodin

haijian-vaadin: Just a small clarification, I think we would go with ignore case string comparison with `Nonnull`, but not using `NotNull`. `NotNull` is for validation, which means that there would be validation errors if the value is null, but the value itself could be null, otherwise, the validation would never fail.

haijian-vaadin (changes requested):
Let's update the `isRequired()` check to use string comparison.

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
knoobie: I was not sure if `isAnnotationPresent` can handle ignore case comparison, that's why I mentioned both spellings I have seen in the wild

haijian-vaadin: I was thinking sth simple as
```
Stream.of(element.getAnnotations()).anyMatch(anno -> "Nonnull"
                .equalsIgnoreCase(anno.annotationType().getSimpleName()))
```

knoobie: I wanted to argue against using `Stream`, but I just saw what `NodeWithAnnotations::isAnnotationPresent` did under the hood - so nvm :) Sounds like a good way!

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
```diff
@@ -16,19 +16,55 @@
...
+     * @param node
+     *            a node to be required
+     * @return a result of check
+     */
+    public static boolean isRequired(NodeWithAnnotations<?> node) {
+        return node.getAnnotations().stream().anyMatch(annotation -> {
+            return nonNullPattern.matcher(annotation.getName().getIdentifier())
+                    .find()
+                    || annotation.resolve().getQualifiedName()
+                            .equals("org.jetbrains.annotations.NotNull");
```
joheriks: Does this (erroneously) match some annotation that for contains "nonnull" as a substring? Would perhaps be simpler to have a single regexp that covers everything.

Lodin: Used `equalsIgnoreCase` instead of regex. It is simpler and less bug-prone.

joheriks: The magic constant `"nonnull"` is repated, would still put it in a `final static String`.

haijian-vaadin (approved):

Issue #10366, opened 2021-03-23
Support servlet 5 aka jakarta.servlet


tarekoraby:
The newer versions of many app-servers (e.g. Tomcat 10 and Jetty 11) have shifted to use the `jakarta.servlet` namespace. Consider the possibility of supporting those in Vaadin apps.

Legioth:
There is no known good way of supporting both the `jakarta.servlet` and `javax.servlet` namespaces at the same time.

Our options are therefore probably limited to one of these:
* Publish separate releases for each standard. We could in theory split out the differing parts to a separate library, but right now it would require separate versions of at least the `flow-server` dependency.
* Drop support for `javax.servlet`.
* Provide some kind of supported integration to transform between the two standards. This could be based on  https://github.com/eclipse/transformer or https://github.com/apache/tomcat-jakartaee-migration.

pleku:
I think we need to do something for this for the next-LTS - not maybe immediately but at least have a plan for what/when.

knoobie:
Just a service notification; Spring 6 / Spring Boot 3 to be released in Q4 2022 requires Java 17+ and Jakarta 9+

https://spring.io/blog/2021/09/02/a-java-17-and-jakarta-ee-9-baseline-for-spring-framework-6

Artur-:
Some fundamental issues:
* Atmosphere has no version for servlet 5: https://github.com/Atmosphere/atmosphere/issues/2467
* Jetty websocket support has changed in Jetty 10 so it needs some refactoring, e.g. the `websocket-server` does not exist in Jetty 11
* Apache commons file upload is for Servlet 3: https://issues.apache.org/jira/browse/FILEUPLOAD-345

mperktold:
>     * Jetty websocket support has changed in Jetty 10 so it needs some refactoring, e.g. the `websocket-server` does not exist in Jetty 11

We are already running on Jetty 10 without any problems in that regard, using the new `websocket-javax-server` module.

I have a [repo](https://github.com/mperktold/vaadin14-embedded-jetty) that replicates our setup. It is forked from a Vaadin 14 project, but it has been upgraded to Vaadin 19, and the same setup works for us with Vaadin 22.

mshabarov:
Implemented in Vaadin 24.0.

PR #10758, opened 2021-04-21
refactor!: Move access checker and @AnonymousAllowed to flow-server

Artur-:
`@AnonymousAllowed` is moved from `com.vaadin.flow.server.connect.auth` package to `com.vaadin.flow.server.auth`.
This allows them to be used for view access control also.


flow-server/pom.xml
```diff
@@ -69,6 +68,12 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
+        <!-- Needed for security annotations -->
+        <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+            <version>${javax.annotation.api.version}</version>
```
knoobie, line 75: Should this be changed to 

```xml
<dependency>
    <groupId>jakarta.annotation</groupId>
    <artifactId>jakarta.annotation-api</artifactId>
    <version>1.3.5</version>
</dependency>
```

to align with Spring Boot and others that have adopted the current release of the latest non-jakarta-namespace jars?

Artur-: Sounds like that should be part of #10366

knoobie: That's something different - the `1.3.5` of `jakarta` still has the old `javax` namespace. The ticket is about the new API with `jakarta`.

haijian-vaadin (changes requested):
For @AnonymousAllowed annotation, instead of just moving, would it worth having a duplicate and mark the one in connect package as deprecated for the sake of easier migration?

Artur-:
Any idea on how that would work since the checker is in `flow-server` and don't want to depend on `fusion-endpoint` where the old annotation is?

Artur-:
My assumption was that `@AnonymousAllowed` is not the mostly used annotation and therefore updating imports would be fine

flow-server/pom.xml
pleku, line 75: Could/should this be `optional` scope instead

flow-server/src/test/java/com/vaadin/flow/server/auth/AccessControlTestClasses.java
```diff
@@ -0,0 +1,235 @@
+package com.vaadin.flow.server.auth;
+
+import javax.annotation.security.DenyAll;
+import javax.annotation.security.PermitAll;
+import javax.annotation.security.RolesAllowed;
+
+public class AccessControlTestClasses {
+
+    public static class NoAnnotationEndpoint {
```
haijian-vaadin, line 9: Rename these XXXXEndpoint to XXXXClass? Since they are not endpoints, just generic classes?

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
```diff
@@ -24,6 +24,8 @@
 import javax.servlet.http.HttpServletRequest;
 
 import com.vaadin.flow.server.VaadinService;
+import com.vaadin.flow.server.auth.AccessAnnotationChecker;
+import com.vaadin.flow.server.auth.AnonymousAllowed;
```
pleku, line 28: So now`fusion-endpoint` will be even more tied to `flow-server`. Is this ideal at this point that these things are in `flow-server` at this point ? @haijian-vaadin @joheriks

haijian-vaadin (changes requested):

pleku (commented):
I'm not sure exactly if this change is even necessary at this point as it might be that this code will not end up to `flow-server` anyway.

haijian-vaadin:
> Any idea on how that would work since the checker is in `flow-server` and don't want to depend on `fusion-endpoint` where the old annotation is?

Right, then I update the title and description of the PR, so the breaking change would be mentioned in the release note.

pleku:
>>Any idea on how that would work since the checker is in flow-server and don't want to depend on fusion-endpoint where the old annotation is?

>Right, then I update the title and description of the PR, so the breaking change would be mentioned in the release note.

Or should it be in some `vaadin-base` or `frameworks-base` or whatever instead ? I mean if the desire is to NOT have `fusion-endpoint` depend on `flow-server` in the end

Artur-:
If classes are one day moved from `flow-server` elsewhere, then that is a problem for a lot of classes. How is it related to this change?

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
Artur-: Not really "more" but it sure is tied after this, just like it is today.

haijian-vaadin: Should be OK to leave them in flow-server for now.
If we split in the way that Flow specific parts are moved away from the current flow-server, then the remaining would become the common/base part, which would include this.

flow-server/pom.xml
Artur-: Why?

flow-server/src/test/java/com/vaadin/flow/server/auth/AccessControlTestClasses.java
Artur-: Done

haijian-vaadin (approved):

pleku:
> If classes are one day moved from flow-server elsewhere, then that is a problem for a lot of classes. How is it related to this change?

It is related since this topic is relevant at the moment. It is not mandatory, but I'm just raising the question

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/auth/VaadinConnectAccessChecker.java
joheriks: This just means that `AccessAnnotationChecker` and `AnonymousAllowed` belong in the future common `flow` and `fusion` module (and hence, at the moment, in `flow-server`).

flow-server/pom.xml
pleku: To avoid potential conflicts and issues down the road due to introducing a dependency that existing projects might be using already, but actually optional doesn't even help there.

Issue #10281, opened 2021-03-13
Development Mode Performance Regression 14.x vs 19


knoobie:
Update:

Further investigation https://github.com/vaadin/flow/issues/10281#issuecomment-802839323

|Topic                                        | Pull Request                          |
| --------------------------------------- |:--------------------------------------:|
| Service Worker Generation             | https://github.com/vaadin/flow/pull/10361 (20.x / 7.x) & https://github.com/vaadin/flow/pull/10384 (19.x / 6.x)  |
| Shipping precompiled JS             | https://github.com/vaadin/flow/pull/10765 (20.x / 7.x) & TBA (19.x / 6.x)  |
---

Original Ticket:

### Description of the bug / feature

Starting an application (v19) with `mvn` downloaded from start.vaadin.com takes a lot more time than the same application with v14. On my machine 3x slower..

#### V14 Example 

- Download v14 app from https://start.vaadin.com
- Run it with `mvn`
- Application launches **fast**

Example attached: [my-app(1).zip](https://github.com/vaadin/flow/files/6134418/my-app.1.zip) 

#### V19 Example 

- Download v19 app from https://start.vaadin.com
- Run it with `mvn`
- Application launches **slow**

Example attached: [my-app(2).zip](https://github.com/vaadin/flow/files/6134417/my-app.2.zip)

### Expected behavior

Both application have the same starting time.

#### v14 - 7 sec webpack execution
```log
2021-03-13 09:32:19.629  INFO 22488 --- [nio-8080-exec-1] c.v.f.s.DefaultDeploymentConfiguration   : 
Vaadin is running in DEBUG MODE.
When deploying application for production, remember to disable debug features. See more from https://vaadin.com/docs/
2021-03-13 09:32:21.440  INFO 22488 --- [nio-8080-exec-1] c.vaadin.flow.spring.SpringInstantiator  : The number of beans implementing 'I18NProvider' is 0. Cannot use Spring beans for I18N, falling back to the default behavior
2021-03-13 09:32:26.009  INFO 22488 --- [         task-2] dev-webpack                              : Started webpack-dev-server. Time: 6972ms

```

#### v19 - 21 sec webpack execution (and a lot of unrelated logging)

```log
2021-03-13 09:33:18.310  INFO 22804 --- [onPool-worker-1] com.vaadin.flow.server.DevModeHandler    : Starting webpack-dev-server

------------------ Starting Frontend compilation. ------------------
[...]

PR #10765, opened 2021-04-22
fix: don't publish flow-client TS sources in JAR

Haprog:
Improve performance of webpack build by only shipping precompiled JS of the flow-frontend TS sources, so webpack doesn't try to use the TS sources which is slightly slower than using JS (and `.d.ts`) directly. The TS sources shouldn't be needed in the JAR anyway and this has the benefit of making the JAR a bit smaller in size also. Related to #10281.


Legioth:
I guess one drawback of this is that it would give worse source maps for anyone trying to locally debug anything related to the code for which the original TS files are missing.

Haprog:
> I guess one drawback of this is that it would give worse source maps for anyone trying to locally debug anything related to the code for which the original TS files are missing.

I don't think that's an issue since we have:
```json
    "sourceMap": true,
    "inlineSources": true,
```
in `tsconfig.json` (for the flow-client build)

The `.js.map` files seem to contain the full TS sources too.

flow-client/src/main/frontend/Flow.ts
```diff
@@ -251,7 +251,7 @@ export class Flow {
       const { appId } = appConfig;
 
       // Load bootstrap script with server side parameters
-      const bootstrapMod = await import('./FlowBootstrap');
+      const bootstrapMod = await import('../resources/META-INF/resources/frontend/FlowBootstrap');
```
haijian-vaadin, line 254: Tried the snapshot build, got frontend compile errors 
ERROR in ../target/flow-frontend/Flow.js
Module not found: Error: Can't resolve '../resources/META-INF/resources/frontend/FlowBootstrap'

haijian-vaadin (changes requested):

Haprog:
After moving also `FlowClient` and `FlowBootstrap` (`.js` and `.d.ts`) it seems to work now. I tested locally in a Flow app. Let's see if the CI also passes now.

Haprog:
I forgot to mvn clean before installing flow-client locally when testing so I didn't catch some problems. New try now. Tested locally working again.

flow-client/.eslintrc.js
```diff
@@ -18,6 +18,13 @@ module.exports = {
       },
     },
 
+    {
+      "files": ["**/.eslintrc.js"],
+      "globals": {
+        "module": "readonly",
+      },
+    },
```
Haprog, line 26: Add an entry for `.eslintrc.js` files so it doesn't complain about `module.exports = {` line in the eslint configs.

flow-client/scripts/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  "env": {
+    "browser": false,
+    "node": true,
```
Haprog, line 4: New eslint config for the `scripts` dir with `node` env so it doesn't complain about usage of `require()` in node scripts.

flow-client/scripts/client.js
```diff
@@ -35,10 +36,13 @@ ${clientSource}
...
-fs.writeFileSync(toSourceDir + toFile, clientSource, 'utf8');
+fs.writeFileSync(sourceDir + toFile, clientSource, 'utf8');
 
-// Write to target
-fs.writeFileSync(toTargetDir + toFile, clientSource, 'utf8');
+// Write to target (copy '.d.ts' and '.js' files from sourceDir)
+mkdirp(targetDir);
+fs.readdirSync(sourceDir)
+  .filter(s => s.endsWith('.d.ts') ||s.endsWith('.js'))
+  .forEach(file => fs.copyFileSync(sourceDir + file, targetDir + file));
```
Haprog, line 45: Use `mkdirp` to create this target dir as it's not automatically populated by Maven anymore due to moving the source files.

Now we also need to copy the FlowClient and FlowBootstrap files (`.js` and `.d.ts`) from source to target dir due to moving the source files. `tsc` (TS compiler) outputs the compiled results of TS files into target but doesn't copy the JS sources for these files that aren't written in TS. Here we just copy all `.d.ts` and `.js` from source to target to be more future proof.

Previously this script was only writing the generated `FlowClient.js` into both source and target. Now it's written into source first and then it gets copied to target also as part of this step.

flow-client/src/main/frontend/Flow.ts
Haprog: Problem fixed.

haijian-vaadin:
I just tried, now it indeed works. Good job!

haijian-vaadin (approved):

PR #10767, opened 2021-04-22
feat: Add helpers for using current request and simplify naming

Artur-:
Requires #10758

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
@@ -27,6 +27,8 @@ import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
 import javax.servlet.http.HttpServletRequest;
 
+import com.vaadin.flow.server.VaadinServletRequest;
+
 /**
  * Checks if a given user has access to a given method.
  * <p>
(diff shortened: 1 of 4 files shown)


denis-anisimov:
The first two commits are entirely fusion related changes,
They should be reviewed by the fusion team.

I don't get why some classes have been moved to `flow-server` (what's the usecase for app without fusion in the `@AnonymousAllowed` annotation) but this is a part of another PR anyway.

haijian-vaadin:
> The first two commits are entirely fusion related changes,
> They should be reviewed by the fusion team.
> 
> I don't get why some classes have been moved to `flow-server` (what's the usecase for app without fusion in the `@AnonymousAllowed` annotation) but this is a part of another PR anyway.

It depends on https://github.com/vaadin/flow/pull/10758, might be better to review this PR after 10758 is merged and this PR is rebased.

Artur-:
Rebased

> I don't get why some classes have been moved to flow-server (what's the usecase for app without fusion in the @AnonymousAllowed annotation) but this is a part of another PR anyway.

The same annotations can be used for view access control

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
```diff
@@ -52,6 +54,46 @@
...
+     * {@link HttpServletRequest#isUserInRole(String)} has access to the given
+     * method.
+     * 
+     * @param method
+     *            the method to check access to
+     * @return {@code true} if the user has access to the given method,
+     *         {@code false} otherwise
+     */
+    public boolean hasAccess(Method method) {
+        VaadinServletRequest request = VaadinServletRequest.getCurrent();
```
denis-anisimov, line 69: I would remove any API which assumes there is something implicit.
`VaadinServletRequest.getCurrent()` is a magic assumption: I recall it was your attempt to clean up the code
to avoid using `Current`  whenever it's possible.

There can be a helper method outside of this class which may be used by a developer (not by Flow internal code).
But the original code should avoid `Current`  implicits.

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
```diff
@@ -52,6 +54,46 @@
...
+     * {@link HttpServletRequest#isUserInRole(String)} has access to the given
+     * class.
+     * 
+     * @param cls
+     *            the class to check access to
+     * @return {@code true} if the user has access to the given method,
+     *         {@code false} otherwise
+     */
+    public boolean hasAccess(Class<?> cls) {
+        VaadinServletRequest request = VaadinServletRequest.getCurrent();
```
denis-anisimov, line 89: Current implicit

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
Artur-: This is a helper to be used by the developer

Artur-: https://github.com/vaadin/spring/pull/780/files#diff-8ee7e28928834d56e852315edccd85c18bab05e3ca39fb7de7c037e38b31b2ceR114 is an example of how you would use it

denis-anisimov (dismissed):

haijian-vaadin (dismissed):

joheriks (approved):

haijian-vaadin (approved):

PR #10774, opened 2021-04-22
Fixed outdated documentation about Push annotation

mstahv:


enver-haase (commented):
Better say 'the' AppShellConfigurator class if only one is possible. Or 'an' if there can be multiple.

mstahv:
See @enver-haase you could have done better job by yourself 🤪 Made a minor change that is probably better now.

mshabarov (approved):

PR #10697, opened 2021-04-16
Assume a non-redirect response for login requests

Artur-:
Partly fixes #10681
Partly fixes #10682

Connected to https://github.com/vaadin/spring/pull/770

PR #10787, opened 2021-04-23
fix: update CSRF tokens with the ones returned in the login call

Artur-:
This fixes a problem introduced in #10697

### DIFF
--- a/flow-client/src/main/frontend/Authentication.ts
+++ b/flow-client/src/main/frontend/Authentication.ts
@@ -47,6 +47,17 @@ export async function login(username: string, password: string, options?: LoginO
 
     if (loginSuccessful) {
       const vaadinCsrfToken = response.headers.get('Vaadin-CSRF') || undefined;
+      updateVaadinCsrfToken(vaadinCsrfToken);
+
+      const springCsrfHeader = response.headers.get('Spring-CSRF-header') || undefined;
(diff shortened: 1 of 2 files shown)


joheriks (approved):

flow-client/src/main/frontend/Authentication.ts
```diff
@@ -47,6 +47,17 @@ export async function login(username: string, password: string, options?: LoginO
 
     if (loginSuccessful) {
       const vaadinCsrfToken = response.headers.get('Vaadin-CSRF') || undefined;
+      updateVaadinCsrfToken(vaadinCsrfToken);
+
+      const springCsrfHeader = response.headers.get('Spring-CSRF-header') || undefined;
+      const springCsrfToken = response.headers.get('Spring-CSRF-token') || undefined;
```
haijian-vaadin, line 53: Would be better to have a comment to mention where the headers  (Spring-CSRF-header, Spring-CSRF-token) are set, otherwise looks quite magic here.

haijian-vaadin (changes requested):

flow-client/src/main/frontend/Authentication.ts
Artur-: Added comment

haijian-vaadin (approved):

PR #10800, opened 2021-04-26
View access checker

### LINKED ISSUE #10609: Support annotating Flow views with the same annotations used for endpoints

Endpoints use 
* `@AnonymousAllowed` - public access
* `@PermitAll` - requires login but not any specific role
* `@DenyAll` - no access
* `@RolesAllowed` - access if you have one of the given roles

The default is `@DenyAll`

It should be possible to use the same annotations for Flow views with exactly the same rules. Checking the annotations should be done by Flow before creating a view instance.

Note that this feature cannot be enabled by default as it would mean that a new application with a single view would not work. Instead it can be implemented in Flow as a feature that is off by default. In a Spring Boot app, `VaadinWebSecurityConfigurerAdapter` can enable the feature.

```java
@RolesAllowed("admin")
@Route("admin")
public class AdminView extends VerticalLayout {
   public AdminView() {
     add(new Div("You can only see this if you are an admin"));
  }
}

### Versions:
    - Vaadin / Flow version:
    - Java version:
    - OS version:
    - IDE (if applicable):

### PR DESCRIPTION
**Pull request by Artur-**
Fixes #10609

Based on #10767

### CODE REVIEW

File: flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java [hunk 1]
```diff
@@ -0,0 +1,176 @@
...
+                        .setAttribute(SESSION_STORED_REDIRECT, beforeEnterEvent.getLocation().getPathWithQueryParameters());
+                if (loginView != null) {
+                    beforeEnterEvent.forwardTo(loginView);
+                } else {
+                    beforeEnterEvent.getUI().getPage().setLocation(loginUrl);
+                }
+            } else {
+                beforeEnterEvent.rerouteToError(NotFoundException.class);
+            }
+        }
```
taefi, line 170: I suggest to reduce the indentation of these lines by ommitting `} else {` block after `if (ok) { return }`:

```suggestion
        getLogger().debug("Denied access to view {}", targetView.getName());
        if (httpServletRequest.getUserPrincipal() == null) {
            httpServletRequest.getSession()
                    .setAttribute(SESSION_STORED_REDIRECT, beforeEnterEvent.getLocation().getPathWithQueryParameters());
            if (loginView != null) {
                beforeEnterEvent.forwardTo(loginView);
            } else {
                beforeEnterEvent.getUI().getPage().setLocation(loginUrl);
            }
        } else {
            beforeEnterEvent.rerouteToError(NotFoundException.class);
        }
```

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
```diff
@@ -0,0 +1,176 @@
...
+            if (httpServletRequest.getUserPrincipal() == null) {
+                httpServletRequest.getSession()
+                        .setAttribute(SESSION_STORED_REDIRECT, beforeEnterEvent.getLocation().getPathWithQueryParameters());
+                if (loginView != null) {
+                    beforeEnterEvent.forwardTo(loginView);
+                } else {
+                    beforeEnterEvent.getUI().getPage().setLocation(loginUrl);
+                }
+            } else {
+                beforeEnterEvent.rerouteToError(NotFoundException.class);
```
taefi, line 168: Are we assuming if the `AccessAnnotationChecker::hasAccess` returns `false` and the `userPrincipal != null`, then route does not exist essentially? Shouldn't this reroute to an `UnAuthorized` error page instead of  `NotFoundException`? Or am I missing something?

taefi (changes requested):
Some questions:

- How the ViewAccessChecker would be enabled? Should the user add it as `beforeEnterListener` through `VaadinServiceInitListener` or ...?

- Shouldn't some unit tests or IT's be added for these changes, especially the ViewAccessChecker class?

And the `ViewAccessChecker` does not comply with line wrapping length of 80 defined for flow repo.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
```diff
@@ -0,0 +1,176 @@
...
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.router.BeforeEnterEvent;
+import com.vaadin.flow.router.BeforeEnterListener;
+import com.vaadin.flow.router.NotFoundException;
+import com.vaadin.flow.server.VaadinServletRequest;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
```
taefi, line 28:
```suggestion
```

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.server.auth;
+
+import javax.servlet.http.HttpServletRequest;
```
taefi, line 18:
```suggestion
import javax.servlet.http.HttpServletRequest;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
```

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
```diff
@@ -0,0 +1,176 @@
...
+                }
+            } else {
+                beforeEnterEvent.rerouteToError(NotFoundException.class);
+            }
+        }
+    }
+
+    private Logger getLogger() {
+        return LoggerFactory.getLogger(getClass());
+    }
```
taefi, line 175: Asking out of curiosity, is there a benefit to call `LoggerFactory.getLogger(getClass())` each time we need to log something, or is it better to hold the reference in a `private final` field and just use that throughout the class? I know that `LoggerFactory::getLogger` is returning the logger reference from a cached map, just wanted to learn why this way is preferred here.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: Is there a reason to disclose that the view exists when you do not have access to it?

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: There are some issues with class loaders and redeploy if you keep logger references, at least if you keep them in static fields. Calling `getLogger(cls)` is very cheap.

Artur-:
> How the ViewAccessChecker would be enabled? (I mean not using spring add-on) Should the user add it as beforeEnterListener through VaadinServiceInitListener or ...?

Yes, as it is used by the Spring integration in https://github.com/vaadin/spring/blob/456d653ca5743b81cdea4ce672170942296dee9a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/ViewAccessCheckerInitializer.java

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
taefi: If you look at it from the end-user point of it is correct and I like this attitude. But from a user (developer) perspective, isn't this confusing during the development? How the user should distinguish between a bad route or security setup and not having enough permissions? 

I think the ideal form would enable the users to configure the message/page as a parameter, so they can decide to which level they want to hide their pages against unauthorized users or attackers. I think normally an application would filter (hide) menus based on user's privileges, and you are right it does not make sense to reveal that "there is a hidden thing you don't have access to see", but the developers would be confused IMO.

Artur-:
Added unit tests, fixed formatting

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: I think we should document how to enable the debug logging for this so you see from the logs what happens. If there are requests, we can consider adding more features

flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    @Test
+    public void anonAccessToAnonViewAllowed() {
+        Result result = checkAccess(AnonymousAllowedView.class, null);
+        Assert.assertTrue(result.wasAccessGranted());
+    }
+
+    @Test
+    public void anonAccessToNoAnnotationViewDenied() {
```
taefi, line 68:
```suggestion
    public void anonymousAccessToNoAnnotationViewDenied() {
```

flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    @Test
+    public void anonAccessToRolesAllowedAdminViewDenied() {
+        Result result = checkAccess(RolesAllowedAdminView.class, null);
+        Assert.assertFalse(result.wasAccessGranted());
+    }
+
+    @Test
+    public void loggedInNoRolesAccessToAnonViewAllowed() {
```
taefi, line 98:
```suggestion
    public void loggedInNoRolesAccessToAnonymousViewAllowed() {
```

flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
```diff
@@ -0,0 +1,412 @@
...
+
+    @Test
+    public void loggedInNoRolesAccessToRolesAllowedAdminViewDenied() {
+        Result result = checkAccess(RolesAllowedAdminView.class,
+                User.USER_NO_ROLES);
+        Assert.assertFalse(result.wasAccessGranted());
+    }
+
+    @Test
+    public void loggedInUserRoleAccessToAnonViewAllowed() {
```
taefi, line 137:
```suggestion
    public void loggedInUserRoleAccessToAnonymousViewAllowed() {
```

flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
```diff
@@ -0,0 +1,412 @@
...
+
+    @Test
+    public void loggedInUserRoleAccessToRolesAllowedAdminViewDenied() {
+        Result result = checkAccess(RolesAllowedAdminView.class,
+                User.NORMAL_USER);
+        Assert.assertFalse(result.wasAccessGranted());
+    }
+
+    @Test
+    public void loggedInAdminRoleAccessToAnonViewAllowed() {
```
taefi, line 176:
```suggestion
    public void loggedInAdminRoleAccessToAnonymousViewAllowed() {
```

flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
```diff
@@ -0,0 +1,412 @@
...
+        Assert.assertTrue(checkAccess(TestLoginView.class, User.NORMAL_USER)
+                .wasAccessGranted());
+        Assert.assertTrue(checkAccess(TestLoginView.class, User.USER_NO_ROLES)
+                .wasAccessGranted());
+        Assert.assertTrue(checkAccess(TestLoginView.class, User.ADMIN)
+                .wasAccessGranted());
+    }
+
+    @Test
+    public void redirectUrlStoredForAnonUsers() {
```
taefi, line 224:
```suggestion
    public void redirectUrlStoredForAnonymousUsers() {
```

flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
```diff
@@ -0,0 +1,412 @@
...
+    public void disabledAccessCheckerAlwaysPasses() {
+        this.viewAccessChecker.setEnabled(false);
+        Assert.assertTrue(checkAccess(RolesAllowedAdminView.class, null)
+                .wasAccessGranted());
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void setLoginViewCannotBeCalledTwice() {
+        this.viewAccessChecker.setLoginView("/foo");
+    }
```
taefi, line 251: Minor (can be ignored): 
This test only tests the following case: 
1. Having called `setLoginView(Class<? extends Component> loginView)` in setup,
2. Calling `setLoginView(String loginUrl)` would result in an exception.
and not the other permutations.

taefi (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
taefi: Fair enough for now, but I suggest making the debug scoped logged message shout out louder in some way (e.g. adding some `!` or `*` to the message string) to enable the user to find it easier at least. ;)

flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
```diff
@@ -31,13 +31,15 @@
...
 import com.vaadin.flow.component.Tag;
+import com.vaadin.flow.server.auth.AnonymousAllowed;
 
 /**
  * This is a basic default error view shown on routing exceptions.
  *
  * @since 1.0
  */
 @Tag(Tag.DIV)
+@AnonymousAllowed
```
denis-anisimov, line 42: This should not be necessary: otherwise I will have to modify each and every my existing views in the existing application.
The absence of the view access annotation should cause a warning via the logger (to be aware if this and apply some action) but by default it should allow anonymous access.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
```diff
@@ -0,0 +1,176 @@
...
+        this.accessAnnotationChecker = accessAnnotationChecker;
+    }
+
+    /**
+     * Enables or disables the access checker.
+     * 
+     * @param enabled
+     *            {@code true} to enable, {@code false} to disable
+     */
+    public void setEnabled(boolean enabled) {
```
denis-anisimov, line 71: So I may disable/enable the checker at any time ?
Should anyone be notified when this happens ?

Is checker thread-safe ?
Does it assume session lock ?
If yes: then it should have an assert.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * Sets the Flow login view to use.
+     * <p>
+     * Note that the access checker needs to be separately enabled using
+     * {@link #setEnabled(boolean)}
+     * 
+     * @param loginView
+     *            the Flow view to use as login view
+     */
+    public void setLoginView(Class<? extends Component> loginView) {
```
denis-anisimov, line 93: Is this method supposed to be called at any time by anyone or this is an internal API?

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
```diff
@@ -0,0 +1,176 @@
...
+
+        HttpServletRequest httpServletRequest = vaadinServletRequest
+                .getHttpServletRequest();
+        getLogger().debug("Checking access for view {}", targetView.getName());
+        if (loginView != null && targetView == loginView) {
+            getLogger().debug("Allowing access for login view {}",
+                    targetView.getName());
+            return;
+        }
+        boolean ok = accessAnnotationChecker.hasAccess(targetView,
```
denis-anisimov, line 149: `ok` means it's value `true` .
The name should be `hasAccess` or something like that.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: Who would want to be notified about this?

It is only meant for explicitly enabling the checker, not really for disabling it. It can be changed into something like `enable()` if anybody finds that more useful.

It is a `BeforeEnterListener`, the framework handles the thread safety one would hope

flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
Artur-: Yes, you will have to modify public views in your existing application if you take this into use. I think this is ok because:
1. There is no must to take this feature into use
2. The default should be secure, not insecure
3. The default should be the same as for endpoints, which is deny access
4. You typically have a super class where you set the annotation so you do not modify every view

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: The user calls it to define the login view, once. Like e.g. https://github.com/vaadin/spring/blob/bf6cec487745dfb6018c44b512f5dba687451eed/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/SecurityConfig.java#L39-L48

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: Changed to `hasAccess`

flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
Artur-: Added

flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
taefi: This View Access Checker mechanism would only be enabled in a Spring Boot-based application if the SecurityConfigurerAdapter class extends the new class `VaadinWebSecurityConfigurerAdapter` (instead of `WebSecurityConfigurerAdapter`), otherwise, it would not be enabled by default. So, already existing applications should not change anything.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
taefi: As its very first use-case, it is called inside the [VaadinWebSecurityConfigurerAdapter](https://github.com/vaadin/spring/pull/780/files#diff-edf3f38b026aff886d241960104383d96c558a28be315c8a4651047e49491d51R243) to enable users to define their login view in the configurer adapter's configure method. 

But for non Spring Boot applications, I guess it is supposed to be called by the user code, as they have to do the configurations manually.

flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
denis-anisimov: OK, if there is away to avoid using checker mechanism then it's fine.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
denis-anisimov: The method is `public` and it's not a CTOR.

It means that _anyone_ (who develops an app) may call this method at _any time_.

If it's expected then it's fine: be prepared that there will be someone who calls it at unexpected time.

If it may be called only _once_ then this should be passed via CTOR: only CTOR allows to force the semantic of calling only once.

Another option : factory which again delegates to the CTOR. It may complicate the code a bit but it allows to avoid unexpected sututations.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
denis-anisimov: This is a general issue with exposing a state via `public` accessors. 
It doesn't matter whether this is a checker or not.
This is something which you should be aware of :

YourAPIClas.state

someones code somwhere:

// set some developer state based on this state e.g.
myComponent.setVisible(YourAPIClas.state );

Now if I can't be notified about state change then I cannot update properly my derived state, so in this case the UI
won't ever be updated.

>It is a BeforeEnterListener, the framework handles the thread safety one would hope

Hmmm.  You are mixing the contract and the usage. This class is `public` and may be used any way : the methods may be called by anyone at any time.
_When_ it's used as a  `BeforeEnterListener` then everything will be fine.
But you can't protect this class from using it not as a `BeforeEnterListener`. 
If this class had had been private then there would have been no issue.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: Constructor parameters are not really compatible with DI, especially when somebody else injects the instance and you get access to it for configuration, like in this Spring Security case

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: I am not sure if you are suggesting a change here or not?

Any API can be called at any time, sure. In this case it means you can disable the checker from anywhere at any time assuming you have a reference to it. Then the next time it is used as a `BeforeEnterListener`, it will allow access to any view. 

How it works if you e.g. call `beforeEnter` manually doesn't sound very interesting, this is something done in unit tests.

So is there something here that should be changed? As mentioned, I think we can change  `setEnable(boolean)` to `enable()` as the only known use case is that you want to enable the checker which is disabled by default

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
denis-anisimov: Factory is perfectly compatible.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: Changed this to `enable()` so it can be opened up later if needed

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: To me, it sounds fine that it throws if you call it multiple times. With Spring Security, you would call it once in your `configure` method like here https://github.com/vaadin/spring/blob/dfc14e5c40602198dabfb5f211c8c5ecaf97a1a3/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/SecurityConfig.java#L61 through a super class helper.

The instance is created and enabled way before that, to ensure view access control is used even if no login view is set

Artur-: This could at least use a unit test though because I suspect something might be broken if you never set the login view, neither as a Flow view nor as a URL

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
denis-anisimov: I suggest: 
- add assertions that session is locked when it's expected to be locked or prevent somehow `ViewAccessChecker` methods be called at any time . By the way `setEnabled` is not a part of `BeforeEnterListener` contract. So it may be called outside of any lock. Other way : initially protect the state via using `AtomicBoolean` or `AtomicReference`.

`setEnabled` /`enabled` : I don't know the whole picture of usage: I don't see why it should be disabled at all if I set it.
If I set the checker then it should be enabled.

My concern is : every time when state is exposed via dumb public accessors looks suspicious. It's almost the same as using `public boolean enabled` field: the internal state is exposed without any protection.

Artur-: There is nothing that modifies the UI state so why would the session lock be needed?

denis-anisimov: If method can be called from another thread then there can be a problem.
E.g. the easiest problem is: the value is cached in the thread.
The thread which mutated the value will use the updated value and another thread which reads the value may use the old one.

denis-anisimov: It doesn't matter which state is modified via several threads: if the state is not final mutable object you have to care about thread safety.

caalador (commented):
Would it make sense to have an IT module with full usage including the pluging into a UI instance?

Else seems fine after the 2 minor additions/changes

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
```diff
@@ -0,0 +1,181 @@
...
+     * Sets the Flow login view to use.
+     * <p>
+     * Note that the access checker needs to be separately enabled using
+     * {@link #enable()}
+     * 
+     * @param loginView
+     *            the Flow view to use as login view
+     */
+    public void setLoginView(Class<? extends Component> loginView) {
+        throwIfLoginViewSet();
```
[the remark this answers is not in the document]
Artur-: It sounds like a real edge case that you have different login URLs. I would consider that when there is a use case described in an issue. Updated the javadoc.

Artur-:
> Would it make sense to have an IT module with full usage including the pluging into a UI instance?
It would. Right now it is in https://github.com/vaadin/spring/pull/780

Artur-:
Added "access denied" in dev mode.

caalador (dismissed):

taefi (dismissed):
Other than the comment on [nullability of configuration](https://github.com/vaadin/flow/pull/10800/files#r635938763), looks good to me.

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
caalador: Going with the usecase it will be hard work to use this from multiple threads. As I haven't seen a usecase needing the isEnabled() method it could just be dropped as then there would be no issue of reading the value.

Artur-:
Can configuration ever be null really?

Artur-:
e.g. https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/server/communication/JavaScriptBootstrapHandler.java#L129 and https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java#L169 do not consider it

caalador:
Only if the UI is not connected to a VaadinSession. I don't think its probable that this would ever be the case here.

Artur-:
Yeah, not in a real "before enter" event

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: Removed `isEnabled` also

caalador (approved):

taefi (approved):

PR #10812, opened 2021-04-26
refactor!: revert "feat: use superclass of VaadinServlet/Request (#9699)"

pleku:
This reverts commit c581668c894e3527e5f727d64bdc97f690d03901, except
intentionally leaves out the protected API introduced to StaticFileServer to get a static
resource, so it may be overridden to use something else than VaadinServletService.


flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -70,12 +70,12 @@
...
      *
-     * @param vaadinService
-     *            vaadin service for the deployment, not <code>null</code>
+     * @param servletService
+     *            servlet service for the deployment, not <code>null</code>
      */
-    public StaticFileServer(VaadinService vaadinService) {
-        this.vaadinService = vaadinService;
-        deploymentConfiguration = vaadinService.getDeploymentConfiguration();
+    public StaticFileServer(VaadinServletService servletService) {
```
denis-anisimov, line 76: Is this necessary to revert ?
It will break OSGi 7.0 (even though it will be easy to fix this).

pleku: Good catch, you're right that it should not be reverted. Fixed

flow-server/src/main/java/com/vaadin/flow/server/communication/PushHandler.java
```diff
@@ -147,15 +148,15 @@
...
-    private VaadinService service;
+    private VaadinServletService service;
 
     /**
      * Creates an instance connected to the given service.
      *
      * @param service
      *            the service this handler belongs to
      */
-    public PushHandler(VaadinService service) {
```
denis-anisimov, line 158: Not sure whether this is a forgotten change as well: since the PR comment mentions that revert for this  is left out.
Technically, I don't see any issues here regardless of the revert or not , so I'm just checking whether this is forgotten or not.

pleku: You're right - this was a mistake but it is also problematic because `PushHandler` creates a `VaadinServletRequest` later (line 188) and that requires a servlet service - though later on there is absolutely no need to use a servlet request because any other API is dependent on `VaadinRequest`. I'm trying to see if `VaadinRequest` could be used somehow

denis-anisimov: OK, it's not blocking.

denis-anisimov (approved):

pleku:
I'll fix the commit message, the `PushRequestHandler` + `PushHandler` parts were also reverted because this change also requires that `VaadinServletRequest` would accept `VaadinService` since for some reason (at least for detecting the session), `PushHandler` creates a `VaadinServletRequest` from the received `AtmosphereRequest`. The code (existing before contribution) is a bit off here - only a `VaadinRequest` would actually be needed since the `PushHandler` uses only `VaadinService` which only needs `VaadinRequest`, not `VaadinServletRequest`. But since there is no way to just make an instance of `VaadinRequest`, the `PushHandler` code is tied to `VaadinServletRequest` instead.

This part is reverted for now. Making `PushRequestHandler` and `PushHandler` not rely on `VaadinServletService` is doable separately later on, but not a priority. It should be enough to make some kind of wrapper for `VaadinRequest` (edited)

PR #10830, opened 2021-04-28
fix: Adapt warning message to 2.6

mshabarov:
Related-to: #9948

### DIFF
--- a/flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
+++ b/flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
@@ -80,10 +80,11 @@ class ThemeLiveReloadPlugin {
           // styles and theme generated file in one run to not have webpack
           // compile error
           if (deletedComponentStyleFile && !themeGeneratedFileDeleted) {
+            const themeGeneratedPath = "./frontend/themes/" + themeName + "/" + themeName + ".generated.js";
+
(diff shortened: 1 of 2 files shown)


flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
```diff
@@ -80,10 +80,11 @@ class ThemeLiveReloadPlugin {
           // styles and theme generated file in one run to not have webpack
           // compile error
           if (deletedComponentStyleFile && !themeGeneratedFileDeleted) {
+            const themeGeneratedPath = "./frontend/themes/" + themeName + "/" + themeName + ".generated.js";
```
taefi, line 83: Isn't the theme .generated.js file recently moved to `/frontend/generated/` and with the following name pattern:
`theme-[theme name].generated.js`? Or probably that change isn't back ported to 2.6?

taefi (changes requested):

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
mshabarov: It isn't back ported to 2.6. For 2.6 the pattern is still `[theme name].generated.js` and it is in the theme folder, not in generated.

mshabarov: I overlooked this difference while the changes were automatically picked from master to 2.6.

taefi (approved):

PR #10831, opened 2021-04-28
fix: don't set text based on template parsing data

### LINKED ISSUE #10106: All content of a @Id mapped element are removed from the DOM

denis-anisimov:
If @Id-mapped component has no children then it gets the text from the template which is
an empty string in many cases. This may break the content of the comonent when the
content is set on the client side (e.g. Lit template without shadow root).

fixes #10106


flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/template/ChildOrderIT.java
```diff
@@ -163,28 +163,29 @@ public void prependTextsFromServer_textsAreAddedBeforeExistingOnes() {
         assertNodeOrder(container, "Client text");
 
         clickAndWaitForContainerToChange(container, "prependChildToContainer2");
-        assertNodeOrder(container, "Server text 1Client text");
+        assertNodeOrder(container, "Client text", "Server text 1");
```
caalador, line 166: Why has the order changed so that server text is appended when it should be prepended as per test name and click target?

caalador (changes requested):

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/template/ChildOrderIT.java
denis-anisimov: I don't know about the test name: this is a revert.
The behavior is exactly the same as it was before introducing parsing template and set Text value:
https://github.com/vaadin/flow/blob/b735ebb640416bd84a6dc519ca2deb5811592779/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/template/ChildOrderIT.java#L166

caalador: Thanks. Yes the naming is confusing, but the PR explains it clearer that it prepends server texts, but client is always first:
```
* Change the order of how nodes are attached to the client Now the index 0 considers the first server-side mapped node after all
the client-side-only nodes. * Add a test with DOM manipulation from the CTOR.
```

caalador (approved):

PR #10832, opened 2021-04-28
chore: run formatter:format

manolo:


mshabarov:
Hmm, I've recently merged formatting PR targeted to `6.0` and it were ~600 files touched. Here we have only 15, which is weird. But anyway, the changes look good.

mshabarov (approved):

PR #10834, opened 2021-04-28
fix: Watch component folders in parent themes

mshabarov:
Adds parent themes component folders to be watched by webpack.

Related-to: #9948


mshabarov:
@caalador I added an extra method to `theme-handle` for parent themes lookup. In a perfect world, it should be merged with existing theme handling traversing to avoid codes duplicate, but there are some reasons I didn't go that way: 
1. avoid regressions in theme handling and overcomplicate it with parents collecting and conditional blocks whether to handle the theme or it's just a parent collecting.
2. theme handling also looks into the JAR, that doesn't need a watch. 

So what do you think, is it fine to have it as a separate method?

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -166,4 +166,50 @@ function extractThemeName(frontendGeneratedFolder) {
...
+  let themeFound = false;
+  for (let i = 0; i < options.themeProjectFolders.length; i++) {
+    const themeProjectFolder = options.themeProjectFolders[i];
+    if (fs.existsSync(themeProjectFolder)) {
+      const themeFolder = path.resolve(themeProjectFolder, themeName);
+      if (fs.existsSync(themeFolder)) {
+        if (themeFound) {
+          throw new Error("Found theme files in '" + themeProjectFolder + "' and '"
+            + themeFound + "'. Theme should only be available in one folder");
+        }
```
caalador, line 193: This check is already in the theme build so I wouldn't have it here as it makes it confusing that where does it come from and the actual handling gives out more data than this phase.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -166,4 +166,50 @@ function extractThemeName(frontendGeneratedFolder) {
...
+ * given custom theme
+ */
+function findParentThemes(themeName, options) {
+  return findParentThemesInternal(themeName, options, false);
+}
+
+function findParentThemesInternal(themeName, options, isParent) {
+  let foundParentThemes = [];
+  let themeFound = false;
+  for (let i = 0; i < options.themeProjectFolders.length; i++) {
```
caalador, line 185: Missing theme resource folder from the check as the parent could be from a jar.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -166,4 +166,50 @@ function extractThemeName(frontendGeneratedFolder) {
...
+        // is supposed to be a parent theme
+        if (isParent) {
+          foundParentThemes.push(themeFolder);
+        }
+        themeFound = themeProjectFolder;
+      }
+    }
+  }
+  return foundParentThemes;
+}
```
caalador, line 213: Would update and simplify to something like:
```
function findParentThemes(themeName, options) {
    return collectParentThemes(themeName, getThemeProjectFolders(options), false);
}

function getThemeProjectFolders(options) {
    const folders = options.themeProjectFolders.filter((folder) => fs.existsSync(folder));

    if (fs.existsSync(options.themeResourceFolder)) {
        folders.put(options.themeResourceFolder);
    }
    return folders;
}

function collectParentThemes(themeName, themeProjectFolders) {
    let foundParentThemes = [];
    themeProjectFolders.forEach(folder => {
        const themeFolder = path.resolve(folder, themeName);
        if (fs.existsSync(themeFolder)) {
            const themeProperties = getThemeProperties(themeFolder);

            if (themeProperties.parent) {
                foundParentThemes.push(collectParentThemes(themeProperties.parent, themeProjectFolders));
                if (!foundParentThemes.length) {
                    throw new Error("Could not locate files for defined parent theme '" + themeProperties.parent + "'.\n" +
                        "Please verify that dependency is added or theme folder exists.")
                }
            }
            // Add a theme path to result collection only if a given themeName
            // is supposed to be a parent theme
            if (isParent) {
              foundParentThemes.push(themeFolder);
            }
        }
    });

    return foundParentThemes;
}
```

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -189,6 +183,23 @@ const themeOptions = {
...
+  themeName = extractThemeName(frontendGeneratedFolder);
+  const parentThemePaths = findParentThemes(themeName, themeOptions);
+  const currentThemeFolders = projectStaticAssetsFolders
+    .map((folder) => path.resolve(folder, "themes", themeName));
+  // Watch the components folders for component styles update in both
+  // current theme and parent themes. Other folders or CSS files except
+  // 'styles.css' should be referenced from `styles.css` anyway, so no need
+  // to watch them.
+  themeWatchFolders = [...currentThemeFolders, ...parentThemePaths]
+    .map((themeFolder) => path.resolve(themeFolder, "components"));
```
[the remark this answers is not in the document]
mshabarov: I couldn't find any such a checks in `extra-watch-webpack-plugin` implementation.
Concerning parent themes, they all should exist, because it is our code who actually crawled through the theme directories.
Current themes yes, might be absent. But if the project has two duplicated theme folders (or developer creates one more duplicated folder) it should be caught by theme handling code.
So I think it's fine to have theme, even though they might be absent.

flow-tests/test-application-theme/test-theme-component-live-reload/frontend/themes/app-theme/theme.json
```diff
@@ -0,0 +1,3 @@
+{
+  "parent": "parent-theme"
+}
```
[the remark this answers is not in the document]
mshabarov: done

flow-tests/test-application-theme/test-theme-component-live-reload/frontend/themes/parent-theme/components/empty
```diff
@@ -0,0 +1 @@
+Empty file here to ensure the components folder is created
```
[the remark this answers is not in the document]
mshabarov: done

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
mshabarov: Ok, will be excluded.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
mshabarov: Hmm, is it a real scenario that someone would want to change the files coming from JAR?
Well, sometimes it can happen. Will add theme resource folder also.

caalador: That's true.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -166,4 +166,51 @@ function extractThemeName(frontendGeneratedFolder) {
...
+  return collectParentThemes(themeName, getThemeProjectFolders(options), false);
+}
+
+function getThemeProjectFolders(options) {
+  const folders = options.themeProjectFolders.filter(
+    (folder) => fs.existsSync(folder));
+
+  if (fs.existsSync(options.themeResourceFolder)) {
+    folders.put(options.themeResourceFolder);
+  }
```
caalador, line 188: We should just drop this as we wouldn't want to watch the jar parent theme.

mshabarov: Excluded

caalador (commented):
Could you rebase the PR on top of latest as the formatter changes should be there as this PR doesn't touch anything except the plugin javascript

mshabarov:
Sure, I reverted all the formatings already done in the `master` and keep only formatting for my changes in the test class.

caalador (approved):

PR #10836, opened 2021-04-28
feat: Backport 'Inject Lumo CSS also into shadow roots' to 2.6

mshabarov:
Adds support for
"lumoImports": ["color", "typography", "badge", ... ]
to theme.json to override the default ["color", "typography"]

If you do not have an application theme specified, Lumo imports are injected by generated-flow-imports.js like before.
If you have an application theme, Lumo imports are injected as part of applying the theme.

In both cases, Lumo imports are added to the beginning of <head> so they are considered before any other styles, such as @CssImport:ed stylesheets. This allows overriding Lumo CSS rules no matter how you add the CSS.

Related-to #9767 #9983

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
@@ -61,7 +61,7 @@ import static com.vaadin.flow.server.frontend.FrontendUtils.IMPORTS_NAME;
 public class TaskUpdateImports extends NodeUpdater {
 
     private static final String THEME_PREPARE = "const div = document.createElement('div');";
-    private static final String THEME_LINE_TPL = "div.innerHTML = '%s';%n"
+    private static final String THEME_LINE_TPL = "%sdiv.innerHTML = '%s';%n"
             + "document.head.insertBefore(div.firstElementChild, document.head.firstChild);";
     private static final String THEME_VARIANT_TPL = "document.documentElement.setAttribute('%s', '%s');";
     // Trim and remove new lines.
@@ -138,19 +138,26 @@ public class TaskUpdateImports extends NodeUpdater {
             Collection<String> lines = new ArrayList<>();
             AbstractTheme theme = getTheme();
             ThemeDefinition themeDef = getThemeDefinition();
+            final boolean hasApplicationTheme =
+                    themeDef != null && !"".equals(themeDef.getName());
 
-            if (themeDef != null && !"".equals(themeDef.getName())) {
+            if (hasApplicationTheme) {
                 // If we define a theme name we need to import theme/theme-generated.js
                 lines.add("import {applyTheme} from 'themes/theme-generated.js';");
                 lines.add("applyTheme(document);");
             }
 
             if (theme != null) {
+                // There is no application theme in use, write theme includes here.
+                // Otherwise they are written by the theme
                 if (!theme.getHeaderInlineContents().isEmpty()) {
                     lines.add(THEME_PREPARE);
+                    if (hasApplicationTheme) {
+                        lines.add("// Handled in the application theme");
+                    }
                     theme.getHeaderInlineContents()
                             .forEach(html -> addLines(lines,
-                                    String.format(THEME_LINE_TPL, NEW_LINE_TRIM
+                                    String.format(THEME_LINE_TPL, hasApplicationTheme ? "// ":"", NEW_LINE_TRIM
                                             .matcher(html).replaceAll(""))));
                 }
                 if (themeDef != null) {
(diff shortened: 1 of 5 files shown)


mshabarov:
Alright, I did a manual testing with `V14.6 pre` generated project having these changes https://github.com/vaadin/start/pull/1160 and local flow `2.6-SNAPSHOT` having these lumo imports backporting, and the badges styles are rendered properly and `myapp.generated.js` now includes all necessary lumo imports injections.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
```diff
@@ -138,20 +138,28 @@ protected void writeImportLines(List<String> lines) {
...
+                        lines.add("// Handled in the application theme");
+                    }
+                    String commentToken = hasApplicationTheme ? "// " : "";
+                    lines.add(String.format(THEME_PREPARE, commentToken));
                     theme.getHeaderInlineContents()
                             .forEach(html -> addLines(lines,
-                                    String.format(THEME_LINE_TPL, NEW_LINE_TRIM
-                                            .matcher(html).replaceAll(""))));
+                                    String.format(THEME_LINE_TPL, commentToken, NEW_LINE_TRIM
+                                            .matcher(html).replaceAll(""), commentToken)));
```
taefi, line 162: Why to write those lines as commented in the first place, while an Application Theme is going to take care of them in another place? Isn't this making the code a little hard to understand? Are those lines supposed to be uncommented by the user later?

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -32,14 +32,49 @@ const documentCssFile = 'document.css';
...
-  target.adoptedStyleSheets = [...target.adoptedStyleSheets, sheet];
+  if (first) {
+    target.adoptedStyleSheets = [sheet, ...target.adoptedStyleSheets];
+  } else {
+    target.adoptedStyleSheets = [...target.adoptedStyleSheets, sheet];
+  }
+};
+`;
+
+// This is copied from flow-generated-import
```
taefi, line 63: Is this comment helping with understanding something? I suggest to either remove it or we add some clear descriptions about what it's doing.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -87,8 +132,31 @@ function generateThemeFile(themeFolder, themeName, themeProperties, productionMo
...
+
+    lumoCssCode.push(`} else if (!document['${lumoCssFlag}']) {\n`);
+    lumoImports.forEach((lumoImport) => {
+      lumoCssCode.push(`addStyleInclude("lumo-${lumoImport}", target);\n`);
+    });
+    lumoCssCode.push(`document['${lumoCssFlag}'] = true;\n`);
+    lumoCssCode.push(`}\n`);
+  }
+
+  globalCssCode.push(`injectGlobalCss(${variable}.toString(), target);\n    `);
```
taefi, line 159: Is `\n` working properly for the generated files on Windows machines? 
Is it an overkill if we consider the the `os` package to get the line break character? 
```JS
var os = require('os')
// ...
var eolChar = JSON.stringify(os.EOL);
```

taefi (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
caalador: I would also skip adding anything instead of adding thing as commented.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
caalador: `\n` works fine for windows also and it will become a new line when the string is written to the file.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
mshabarov: I skipped writing these lines in case of application theme.
As I understand it, theme generated file is not needed to be explored or changed by developers.
For us, there is a comment in the source code
```
// There is no application theme in use, write theme includes here.
// Otherwise they are written by the application theme
```
explaining the behaviour.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
mshabarov: Agree, I see no worth for developers to see this comment.
Removed.

taefi (approved):

PR #10851, opened 2021-04-29
merge `master` into `feature/snowpack3`

vlukashov:
The actual changes that need a review are:
 - a88fbd4: Merge branch 'master' into feature/snowpack
 - f5b0a53: fix: include source maps for flow-frontend .js files
 - eed4dc1: chore: fix PropertyDeploymentConfigurationTest failures


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -693,14 +694,20 @@ private static InputStream getFileFromSnowpack(String filePath)
         if ("index.html".equalsIgnoreCase(filePath)) {
             // special case treatment for the index.html file
```
manolo, line 695: shouldn't be better not to use single-line comment and instead  use multi-line block`/* */`, probably formatter does the work better.

vlukashov:
see https://github.com/vaadin/flow/pull/10854 instead

PR #10853, opened 2021-04-29
feat!: use regular string enum instead of a const enum

### LINKED ISSUE #8963: It should be possible to loop through all defined enum values in TS
### Description of the bug
A common use case for `enum` is to define hard coded options. In many cases, you want to show these options also in the UI. Based on the way the TS enums are generated right now, it seems to be impossible to loop through them as e.g.
```
${Object.keys(MyEnum).map(enum => html`something`)}
```

```
public enum Access {
   USER, ADMIN
}
```

you would want to be able to show in the UI as a select for defining the role of a user without hard coding the roles in TS also.

### Versions:
    - Vaadin version: 17

Lodin:
Given a Java enum as
```java
pubic enum Environment {
  DEV, TEST, PROD
}
```
the generated Typescript source will be
```typescript
enum Environment {
  DEV = "DEV",
  TEST = "TEST",
  PROD = "PROD"
}
```
which translates to JavaScript in a following way:
```javascript
var Environment;

(function (Environment) {
  Environment["DEV"] = "DEV";
  Environment["TEST"] = "TEST";
  Environment["PROD"] = "PROD";
})(Environment || (Environment = {}));
```
So now it is possible to do any object action with the final `Environment` enum:
```javascript
Object.keys(Environment).forEach(enumValue => console.log(enumValue));
```

Resolves #8963

### DIFF
--- a/fusion-endpoint/src/main/resources/com/vaadin/flow/server/connect/generator/EntityTemplate.mustache
+++ b/fusion-endpoint/src/main/resources/com/vaadin/flow/server/connect/generator/EntityTemplate.mustache
@@ -17,7 +17,7 @@ import {{~#importAs}} {{importAs}}{{/importAs}}{{~^importAs}} {{className}}{{/im
  }}{{/vendorExtensions.x-vaadin-file-path}}
  */
 {{#if vendorExtensions.x-is-enum}}
-const enum {{{getClassNameFromImports classname ../../imports}}} {
+enum {{{getClassNameFromImports classname ../../imports}}} {
 {{#allowableValues.enumVars}}
   {{name}} = {{{value}}},
 {{/allowableValues.enumVars}}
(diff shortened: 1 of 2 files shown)


haijian-vaadin:
Looks good to me, but is this really a breaking change? Would the existing code stop working after this change?

Lodin:
@haijian-vaadin, right, that's my mistake. Basically, nothing should change.

haijian-vaadin:
OK, thanks, I just tried myself, looks everything still works fine. I would also argue maybe this is a fix rather than a feat? I think we can safely backport this to V19?

Lodin:
Yes, I think, we can backport it easily because only the code emitted by `tsc` will change and new opportunities will appear.

haijian-vaadin (approved):

Haprog (approved):

platosha (approved):

PR #10855, opened 2021-04-30
chore: fix probable merge failures

caalador:
Fix some lines in webpack.generated
that have duplicated or are used strangely
possibly due to merge resolution.


pleku (approved):
Looks kind of scary change but if the build passes 🤷🏻‍♂️

PR #10863, opened 2021-04-30
feat: Make helper method public to allow Spring to use it

Artur-:


taefi (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -190,7 +190,19 @@ static boolean isInternalRequestInsideServlet(
...
+     * 
+     * @param servletMappingPath
+     *            the servlet mapping from the servlet configuration
+     * @param requestedPath
+     *            the request path relative to the context root
+     * @return an optional containing the path relative to the servlet if the
+     *         request is inside the servlet mapping, an empty optional
+     *         otherwise
+     */
+    public static Optional<String> getPathIfInsideServlet(
```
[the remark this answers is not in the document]
Artur-: Fixed

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -190,7 +190,19 @@ static boolean isInternalRequestInsideServlet(
...
             String servletMappingPath, String requestedPath) {
         if (servletMappingPath.endsWith("/*")) {
```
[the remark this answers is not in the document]
Artur-: Fixed

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -191,29 +192,75 @@ static boolean isInternalRequestInsideServlet(
...
+                servletMappingWithoutSlash = servletMappingPath.substring(1);
+            } else {
+                servletMappingWithoutSlash = servletMappingPath;
+            }
+            if (requestedPath.equals(servletMappingWithoutSlash)) {
+                return Optional.of(requestedPath
+                        .substring(servletMappingWithoutSlash.length()));
+            } else {
+                return Optional.empty();
+            }
```
taefi, line 240: Minor: I would avoid unnecessary `else` and the resulting indented codes inside the `else` block when those cases in the above have explicit `return` statements. Just makes it more readable IMO.

Artur-: Removed

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -191,29 +192,77 @@ static boolean isInternalRequestInsideServlet(
...
+                servletMappingWithoutSlash = servletMappingPath;
+            }
+
+            if (requestedPath.equals(servletMappingWithoutSlash)) {
+                return Optional.of(requestedPath
+                        .substring(servletMappingWithoutSlash.length()));
+            }
+
             return Optional.empty();
         }
```
taefi, line 243: This would simplify it even further:
```suggestion
        if ("/*".equals(servletMappingPath) || "/".equals(servletMappingPath)) {
            // All paths are inside a /* servlet and a / servlet
            return Optional.of(requestedPath);
        } 
        
        if (servletMappingPath.endsWith("/*")) {
            String servletPrefix = servletMappingPath.substring(0,
                    servletMappingPath.length() - 1); // Only remove "*"
            if (requestedPath.startsWith(servletPrefix)) {
                return Optional
                        .of(requestedPath.substring(servletPrefix.length()));
            }
            return Optional.empty();
        } 
        
        // Servlet is mapped only to a static path such as "" or /foo/bar
        String servletMappingWithoutSlash;

        if (servletMappingPath.startsWith("/")) {
            // Requested path should not contain the initial slash
            servletMappingWithoutSlash = servletMappingPath.substring(1);
        } else {
            servletMappingWithoutSlash = servletMappingPath;
        }

        if (requestedPath.equals(servletMappingWithoutSlash)) {
            return Optional.of(requestedPath
                    .substring(servletMappingWithoutSlash.length()));
        }

        return Optional.empty();
```

taefi (approved):

PR #10866, opened 2021-05-01
draft: proposal for caching feature in data communicator

TatuLund:
If you set page size = 100 in Grid, and set cache time to be suitable (say anything between 5 seconds or more), and are using @PreserveOnRefresh feature, refreshing the page will not reload data from data provider until cache expires. This may be useful if your data source is static by nature and you want to prevent user bombarding the data base with refreshes.


pleku:
@TatuLund please create an issue first describing the problem this tries to solve ?

TatuLund:
@pleku I will do it later, this is related to an internal discussion. At the moment this is just a wild idea. One can solve the problem also by using generic backend cache like ehcache.

pleku:
OK. Design wise, I think things like this start as opt-in as in there might be a "simple solution" to it that one can use and test with, but for most apps they would probably want to plug in their own caching solution - some might be provided even as add-ons via directory (at least first).

pleku:
Lets continue with a PR later on. Next I would suggest opening an enhancement issue which describes the use case and we can discuss the design there first.

PR #10869, opened 2021-05-03
chore: add missing cleanup to tests

caalador:
Add missing cleanup of current instances
to tests.

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/communication/PushHandlerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/communication/PushHandlerTest.java
@@ -206,6 +206,7 @@ public class PushHandlerTest {
         mockConnectionLost(new MockVaadinSession(), false);
 
         Assert.assertNull(VaadinSession.getCurrent());
+        VaadinSession.setCurrent(null);
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


flow-server/src/test/java/com/vaadin/flow/server/communication/PushHandlerTest.java
```diff
@@ -216,6 +217,11 @@ public void connectionLost_sessionIsSetViaCurrent_currentInstancesAreCleared()
         Assert.assertNotNull(VaadinSession.getCurrent());
     }
 
+    @After
+    public void cleanup() {
+        VaadinSession.setCurrent(null);
```
denis-anisimov, line 222: There is no `@Before`  set up method. So if there is something which sets `VaadinSession`  then it's done for some tests only (not generally).
Should these tests clear the current by themselves instead ?

caalador: Yeah it was only in 2 tests. Updated.

denis-anisimov: I'm not sure in fact which way is better:
- every time when it's reset it has to be done in `finally` : to avoid unexpected exceptions.
- Then it requires `try{}finallly`  is it become too verbose.
- on the other hand it's not clear why `tearDown` is used at all and it's easy to remove it because of that.

`tearDown` is less verbose but more questionable.

denis-anisimov (changes requested):

denis-anisimov (approved):

PR #10878, opened 2021-05-03
fix: upgrade to LitElement 2.5.0 and lit-html 1.4.0 (#10874) (CP: 7.0)

### LINKED ISSUE #10826: Upgrade to LitElement 2.5 once available
Lit 2 contains one significant breaking change for users: `@internalProperty` is renamed to `@state`.
This has been backported to LitElement 2.5 so that this version contains both `@internalProperty` and `@state`.

We need to update from LitElement 2.4 to 2.5 as soon as it is available and replace all usages of `@internalProperty` with `@state`.

This update is needed for both V20 and V14.

This will enable e.g. start.vaadin.com to generate the same template for both Lit 1 and Lit 2 projects.

Artur-:
Fixes #10826


Artur-:
Apparently this is one is in #10877

PR #10894, opened 2021-05-04
fix: log limitation of adding new styles while switching themes

### LINKED ISSUE #10680: Custom theme component styles are not watched by webpack after theme swapping
When swapping a custom theme name in `@Theme("my-theme")` and after live reload, webpack still watches to the previous theme component folder (`frontend/themes/my-theme/components/`) and doesn't watch for changes (add/delete component CSS files) in a new theme component folder.
1. Download project from start.vaadin.com (choose v19, v20 pre or v14 pre)
2. Create another theme folder, say `other-theme`
3. Run project `mvn`
4. Change theme name from `myapp` to `other-theme`
5. Wait until live reload finishes
6. Add any css file into `other-theme/components`, notice no live-reload happens.
7. Add any css file into `myapp/components`, notice live-reload still works.
### Expected behavior
Live reload works only for active theme  folder.
### Actual behavior
Live reload works for old, but not for actual theme folder.
### Versions:
    - Vaadin / Flow version: V14.6.0.beta1, V19.0.4, V20.0.0.alpha7
    - Java version: 8
    - OS version: MacOS
    - Browser version (if applicable): Chrome 89
    - IDE (if applicable): IntellijJ Idea

taefi:
## Description

If there are more than one theme available in
themes folder, user can switch between them in
development mode by changing the name of the
theme in `@Theme`. As having a watch for adding
new styles in component folder is 'static' and could
not be changed to another theme folder dynamically, 
at least a log about this limitation can help the users.

Fixes #10680 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

This PR does not change any existing behavior, just adds some logs that can help users to find out when a restart is needed to see the effect of their changes in the available themes.

### DIFF
--- a/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
+++ b/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
@@ -26,6 +26,9 @@ const {copyStaticAssets, copyThemeResources} = require('./theme-copy');
 // matches theme name in './theme-my-theme.generated.js'
 const nameRegex = /theme-(.*)\.generated\.js/;
 
+let prevThemeName = undefined;
+let firstThemeName = undefined;
+
 /**
  * Looks up for a theme resources in a current project and in jar dependencies,
  * copies the found resources and generates/updates meta data for webpack
@@ -39,8 +42,26 @@ const nameRegex = /theme-(.*)\.generated\.js/;
 function processThemeResources(options, logger) {
   const themeName = extractThemeName(options.frontendGeneratedFolder);
   if (themeName) {
+    if (!prevThemeName && !firstThemeName) {
+      prevThemeName = themeName;
+      firstThemeName = themeName;
+    } else if ((prevThemeName && prevThemeName !== themeName && firstThemeName !== themeName)
+        || (!prevThemeName && firstThemeName !== themeName)) {
+      const warning = `Attention: Active theme is switched to '${themeName}'.`
+      const description = `
+      Note that adding new style sheet files to '/themes/${themeName}/components', 
+      may not be taken into effect until the next application is restart.
+      Changes to existing style sheet files are being reloaded as before.`;
+      logger.warn("*******************************************************************");
+      logger.warn(warning);
+      logger.warn(description);
+      logger.warn("*******************************************************************");
+    }
+    prevThemeName = themeName;
+
     findThemeFolderAndHandleTheme(themeName, options, logger);
   } else {
+    prevThemeName = undefined;
     logger.debug("Skipping Vaadin application theme handling.");
     logger.trace("Most likely no @Theme annotation for application or only themeClass used.");
   }
--- a/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
+++ b/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
@@ -43,7 +43,6 @@ function processThemeResources(options, logger) {
   const themeName = extractThemeName(options.frontendGeneratedFolder);
   if (themeName) {
     if (!prevThemeName && !firstThemeName) {
-      prevThemeName = themeName;
       firstThemeName = themeName;
     } else if ((prevThemeName && prevThemeName !== themeName && firstThemeName !== themeName)
         || (!prevThemeName && firstThemeName !== themeName)) {
--- a/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
+++ b/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
@@ -49,7 +49,7 @@ function processThemeResources(options, logger) {
       const warning = `Attention: Active theme is switched to '${themeName}'.`
       const description = `
       Note that adding new style sheet files to '/themes/${themeName}/components', 
-      may not be taken into effect until the next application is restart.
+      may not be taken into effect until the next application restart.
       Changes to existing style sheet files are being reloaded as before.`;
       logger.warn("*******************************************************************");
       logger.warn(warning);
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -39,8 +42,25 @@ const nameRegex = /theme-(.*)\.generated\.js/;
 function processThemeResources(options, logger) {
   const themeName = extractThemeName(options.frontendGeneratedFolder);
   if (themeName) {
+    if (!prevThemeName && !firstThemeName) {
+      firstThemeName = themeName;
+    } else if ((prevThemeName && prevThemeName !== themeName && firstThemeName !== themeName)
```
mshabarov, line 47: Just in case: have you tested that this message does not appear after CSS change live reload _before_ theme swap and after CSS change live reload _after_ theme swap?

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -39,8 +42,25 @@ const nameRegex = /theme-(.*)\.generated\.js/;
...
+      logger.warn("*******************************************************************");
+      logger.warn(warning);
+      logger.warn(description);
+      logger.warn("*******************************************************************");
+    }
+    prevThemeName = themeName;
+
     findThemeFolderAndHandleTheme(themeName, options, logger);
   } else {
+    prevThemeName = undefined;
```
mshabarov, line 63: Why is it necessary? If no @Theme annotation or theme name string is specified, then this variable should be undefined.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -39,8 +42,25 @@ const nameRegex = /theme-(.*)\.generated\.js/;
...
   const themeName = extractThemeName(options.frontendGeneratedFolder);
   if (themeName) {
+    if (!prevThemeName && !firstThemeName) {
+      firstThemeName = themeName;
+    } else if ((prevThemeName && prevThemeName !== themeName && firstThemeName !== themeName)
+        || (!prevThemeName && firstThemeName !== themeName)) {
+      const warning = `Attention: Active theme is switched to '${themeName}'.`
+      const description = `
+      Note that adding new style sheet files to '/themes/${themeName}/components', 
+      may not be taken into effect until the next application is restart.
```
mshabarov, line 52:
```suggestion
      may not be taken into effect until the next application restart.
```

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -39,8 +42,25 @@ const nameRegex = /theme-(.*)\.generated\.js/;
...
+      Changes to existing style sheet files are being reloaded as before.`;
```
mshabarov, line 53: Good to describe what does 'existing' mean. Does it refer to files inside /themes/${prevThemeName}/components ?

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
taefi: Yes, it is not appear on those situations. I tested the styles in styles.css and also the existing [component].css in both themes.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
taefi: `prevThemeName = undefined;` is needed in the situation that the user decides to comment or remove the `@Theme(...)` completely to see how the application looks without any theme. Then when the user brings back one of the themes, the previous theme should be `undefined` to enable us to detect the change.

taefi: `!prevThemeName && firstThemeName !== themeName` would be `true` when the user brings back the `@Theme(...)` with the themeName that is not the first theme that application started with.

taefi: The reason for having `prevThemeName` is that I was warning the user with a message like this: `The current theme is changed from 'prevThemeName' to 'themeName' ...`. 
Now that I changed the message to show only the current theme, maybe keeping the `prevThemeName` is irrelevant, right?

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
taefi: I'm referring to the `already existing` css files under /components in both themes, no matter it is the first theme or not. It just cannot detect adding new files being added. Existing css files are watched because of the dependency tree.

taefi:
```suggestion
      Changes to already existing style sheet files are being reloaded as before.`;
```

mshabarov: Ah, thanks! You're right, reload works for all existing files, no matter of old or new theme.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
mshabarov: Thanks for clarifications! It does make sense.
Could you please add a comment explaining that case and what for `(!prevThemeName && firstThemeName !== themeName)` and `prevThemeName = undefined;` are used. It would help improve the readability.

mshabarov:
```java
      // Warning message is shown to the developer when:
      // 1. He is switching to any theme, which is differ from one being set up 
      // on application startup, by changing theme name in `@Theme()`
      // 2. He removes or comments out `@Theme()` to see how the app 
      // looks like without theming, and then again brings `@Theme()` back.
     const warning = `Attention: Active theme is switched to '${themeName}'.`

     // This is needed in the situation that the user decides to comment or 
     // remove the @Theme(...) completely to see how the application looks 
     // without any theme. Then when the user brings back one of the themes, 
     // the previous theme should be undefined to enable us to detect the change.
     prevThemeName = undefined;

```

taefi: Done.

mshabarov (approved):

PR #7608, opened 2020-02-17
Url parameter template support.

bogdanudrescu:
Fixes https://github.com/vaadin/flow/issues/2740
Fixes https://github.com/vaadin/flow/issues/4213

Fixes the 2 above issues by defining a low level API that works with `@Route`, `@RoutePrefix` and `@RouteAlias`, which supports parameters within the route string in following format:
- `path/to/:paramName`
- `path/to/:paramName/foo`

Parameter accepting only values matching a regular expression:
- `path/to/:paramName(regex)/foo`

Optional parameters may be defined as:
- `path/to/:paramName?/foo`
- `path/to/:paramName?(regex)/foo`

Wildcard parameter (it is accepted only as the last segment template):
- `path/to/:paramName*`
- `path/to/:paramName*(regex)`

Multiple parameters may be combined within the path:
- `path/to/:paramName1/:paramName2?`
- `path/to/:paramName1/:paramName2?/foo`
- `path/to/:paramName1/foo/:paramName2/bar`
- `path/to/:paramName1(regex)/foo/:paramName2?/bar/:paramName3*`

Parameters are provided to views using `BeforeEnterEvent.getRouteParameters()` which returns a `RouteParameters` instance.

NOTE: In case of `HasUrlParameter`, the last parameter will be automatically defined so the value of `@Route` should not contain any parameter template defined as the last segment. In other words nothing changes to how `HasUrlParameter` is being used.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/flow/7608)

PR #10910, opened 2021-05-05
test: enable ignored cases in NavigationTriggerIT

### LINKED ISSUE #10513: Fusion: NavigationTriggerIT fails
Related ticker : https://github.com/vaadin/flow/issues/7572

The `NavigationTriggerIT` has been disabled because of `7572`.
The latter ticket is closed but `NavigationTriggerIT` still fails.

I'm going to refactor the test to disable its part which is failing in fusion.
The functionality should be fixed and restored. The test should be enabled after that.

platosha:
Fixes #10513

### DIFF
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/NavigationTriggerIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/NavigationTriggerIT.java
@@ -22,6 +22,7 @@ import org.junit.Ignore;
 import org.junit.Test;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
 
 import com.vaadin.flow.router.NavigationTrigger;
 import com.vaadin.flow.testutil.ChromeBrowserTest;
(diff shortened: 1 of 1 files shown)


flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/NavigationTriggerIT.java
```diff
@@ -87,7 +77,8 @@ public void testNavigationTriggers_back_forward() {
 
         findElement(By.id("rerouteButton")).click();
         assertMessageCount(7);
-        assertLastMessage("/", NavigationTrigger.PROGRAMMATIC, "rerouted");
+        assertLastMessage("/rerouted", NavigationTrigger.PROGRAMMATIC,
```
platosha, line 80: The new behaviour seems to be related with the feature https://github.com/vaadin/flow/pull/7608

Debugging revealed, that the `BeforeEvent` now contains pre-resolved `Location` with the URL parameters filled. So the location has `rerouted` in this case.

See https://github.com/vaadin/flow/blob/872bbf45429f1d617a56396c885d3fed8fbd426b/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java#L703-L758

haijian-vaadin (approved):

PR #10912, opened 2021-05-06
fix: upgrade to LitElement 2.5.1 and lit-html 1.4.1

Artur-:


Haprog (approved):

knoobie:
@Artur- looks like this is missing in the latest 2.5.4 release: https://github.com/vaadin/flow/releases/tag/2.5.4

Haprog:
@knoobie I think this won't make it to 2.5 anymore. Vaadin 14.6.0 LTS (with Flow 2.6 containing this fix) should be out in a few days so you can upgrade to that.

PR #10921, opened 2021-05-06
fix: allow empty segments at start

### LINKED ISSUE #10729: After 14.5, Router HasUrlParameter @WildcardParameter doesn't support empty parameters anymore
After 14.5, Router HasUrlParameter @WildcardParameter doesn't support empty parameters anymore. Works with 14.4.10. 
```
@Route(value = "about", layout = MainView.class)
@PageTitle("About")
@CssImport("./views/about/about-view.css")
public class AboutView extends Div implements HasUrlParameter<String> {
    public AboutView() {
        addClassName("about-view");
        add(new Text("Content placeholder"));
    }
    @Override
    public void setParameter(BeforeEvent beforeEvent, @WildcardParameter String s) {
        System.out.println(">>>"+s);
    }
}
```
### Expected behavior
With code above the standard output for url `http://localhost:8080/about//2/3`, should be just:
`>>>/2/3`
### Actual behavior
The standard output is
```
>>>2/3
>>>VAADIN/build/webcomponentsjs/webcomponents-loader.js
>>>VAADIN/build/vaadin-gizmo-1f1e5ed60cb793a57317.cache.js
>>>VAADIN/static/client/client-4C23589935A74206212F474A719FC41C.cache.js
>>>VAADIN/build/vaadin-bundle-098d1f74dc147d6e9258.cache.js
>>>icons/icon-32x32.png
>>>manifest.webmanifest
>>>sw.js
>>>icons/icon-16x16.png
>>>icons/icon-96x96.png
```
### Versions:
    - Vaadin version: 14.5.3
    - Java version: Amazon Corretto 11.0.10
    - OS version: Windows 10 Pro

caalador:
Allow having empty segment as first segment
for parameters.

fixes #10729

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
@@ -19,6 +19,7 @@ import java.io.Serializable;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Utility class which contains various methods for parsing a route url into
@@ -39,7 +40,7 @@ public class PathUtil implements Serializable {
      * @return a List containing the segments of the path.
      */
     public static List<String> getSegmentsList(String path) {
-        path = trimPathEnding(path);
+        path = path == null ? "" : trimSegmentsString(path);
 
         final String[] segments = path.split("/");
         if (segments.length == 1 && segments[0].isEmpty()) {
@@ -59,7 +60,7 @@ public class PathUtil implements Serializable {
      * @return path form from input segments.
      */
     public static String getPath(List<String> segments) {
-        return trimPathEnding(
+        return trimSegmentsString(
                 segments == null ? "" : String.join("/", segments));
     }
 
@@ -106,18 +107,16 @@ public class PathUtil implements Serializable {
     }
 
     /**
-     * Trim the path by removing any leading whitespaces and trailing slashes
-     * and whitespaces.
+     * Trim the path by removing any leading and trailing whitespaces and
+     * trailing slashes.
      *
      * @param path
-     *            url path to trim.
+     *            url path to trim, not null
      * @return a String representing the input path without any leading and
-     *         trailing whitespaces and trailing slash.
+     *         trailing whitespaces or trailing slash.
      */
-    public static String trimPathEnding(String path) {
-        if (path == null) {
-            return "";
-        }
+    public static String trimSegmentsString(String path) {
+        Objects.requireNonNull(path);
 
         path = path.trim();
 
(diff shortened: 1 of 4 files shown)


flow-server/src/test/java/com/vaadin/flow/router/RouterConfigurationUrlResolvingTest.java
```diff
@@ -146,6 +146,28 @@ public void url_resolves_correctly_for_optional_and_wild_parameters()
                 .getUrl(WildParameter.class, "there/are/many/of/us"));
     }
 
+    @Test
+    public void wildcardPathWithEmptyParameter_emptyParameterIsAvailable() {
```
pleku, line 150: Can it be so that the wildcard route is mapped to `""` ? Then you would have "//" or "//foo" ?
If yes then should it be tested too ?

flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
```diff
@@ -105,4 +105,26 @@ public static String trimPath(String path) {
...
+    /**
+     * Trim the path by removing any leading whitespaces and trailing slashes
+     * and whitespaces.
+     *
+     * @param path
+     *            url path to trim.
+     * @return a String representing the input path without any leading and
+     *         trailing whitespaces and trailing slash.
+     */
+    public static String trimPathEnding(String path) {
```
pleku, line 117: The name of this method is slightly odd compared to the javadocs: "trim path ending" VS "trim whitespace and remove trailing slashes"

flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
```diff
@@ -105,4 +105,26 @@ public static String trimPath(String path) {
         return path;
     }
 
+    /**
+     * Trim the path by removing any leading whitespaces and trailing slashes
+     * and whitespaces.
+     *
+     * @param path
+     *            url path to trim.
```
pleku, line 113: apparently can be null which is considered as ""

pleku (changes requested):
Couple of javadocs nits and one question

flow-server/src/test/java/com/vaadin/flow/router/RouterConfigurationUrlResolvingTest.java
caalador: There is a problem that for the location there is the restiction that `Relative path cannot start with /` so a `Route("")` can not accept a `//Parameters/path` as it will leave a / as the initial parameter.

flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
caalador: renamed

flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
caalador: Changed so we don't accept null.

pleku (approved):

PR #10933, opened 2021-05-06
Always update webpack.generated.js

### LINKED ISSUE #10932: webpack.generated.js is not overwritten when upgrading from 14.0.10 to 14.5.4

If you have a 14.0.10 project and have committed `webpack.generated.js` to the repository, it seems that upgrading to 14.5.4 keeps using the old `webpack.generated.js`, which causes the project to fail completely because webpack output path has been changed between 14.0 and 14.5

Artur-:
## Description

Fixes #10932

## Type of change

- [X] Bugfix
- [ ] Feature


pleku (changes requested):
LGTM but a unit test would be great for catching regressions, any chance for adding that to https://github.com/vaadin/flow/blob/master/flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java ?

Artur-:
Added in #10934

PR #10938, opened 2021-05-07
fix: allow empty segments at start (#10921)

### LINKED ISSUE #10729: After 14.5, Router HasUrlParameter @WildcardParameter doesn't support empty parameters anymore

caalador:
Fixes a regression with HasUrlParameter by allowing to have an empty segment as first segment for parameters.

Fixes #10729

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
@@ -19,6 +19,7 @@ import java.io.Serializable;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Utility class which contains various methods for parsing a route url into
@@ -39,7 +40,7 @@ public class PathUtil implements Serializable {
      * @return a List containing the segments of the path.
      */
     public static List<String> getSegmentsList(String path) {
-        path = trimPath(path);
+        path = path == null ? "" : trimSegmentsString(path);
 
         final String[] segments = path.split("/");
         if (segments.length == 1 && segments[0].isEmpty()) {
@@ -59,8 +60,8 @@ public class PathUtil implements Serializable {
      * @return path form from input segments.
      */
     public static String getPath(List<String> segments) {
-        return trimPath((segments == null || segments.isEmpty()) ? ""
-                : String.join("/", segments));
+        return trimSegmentsString(
+                segments == null ? "" : String.join("/", segments));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
```diff
@@ -59,8 +60,8 @@
      * @return path form from input segments.
      */
     public static String getPath(List<String> segments) {
-        return trimPath((segments == null || segments.isEmpty()) ? ""
-                : String.join("/", segments));
+        return trimSegmentsString(
+                segments == null ? "" : String.join("/", segments));
```
pleku, line 64: nit: I would have gone with same as above
```java
return segments == null ? "" : trimSegmentsString(String.join("/", segments));
```

pleku (approved):

flow-server/src/main/java/com/vaadin/flow/router/internal/PathUtil.java
caalador: Well the change went to master already  and it is cognitively clearer. Else would change it to 
```
if(segments == null || segments.isEmpty()) {
  return "";
}
return String.join("/", segments);
```
But now checking the String.join contents I would have expected it to be smarter about empty iterables.

PR #10945, opened 2021-05-07
Fix incorrect import

Artur-:
Lit does not export `lit/directive`, it only exports `lit/directive.js`


Artur-:
Update `decorators` -> `decorators.js` and `directive` -> `directive.js`

Haprog:
I pushed a commit updating some more imports. I used VS Code search with regex `['"]lit/[^'"]+(?<!\.js)['"]` to essentially find all occurrences of `lit/<something>` imports that don't end in `.js` and updated those.

Haprog (approved):

Haprog:
@Artur- Should this be cherry-picked to older versions?

PR #10953, opened 2021-05-10
fix: Enable use of external css using `@import`

### LINKED ISSUE #10228: [Vaadin 19] Embedded Applications with Custom Theme and external font: error in dev tool console when there are two different servers
### Description of the bug
I have an Vaadin 19 application with my custom theme (called zaiks) and some components exported via `WebComponentExporter`. This web components (I call them 'widgets') are then embedded in other external application.

But from the beginning what I have:

**A. I configured all according to the:**
1) custom theme documentation (https://vaadin.com/docs-beta/latest/ds/customization/custom-theme/)
2) custom theme demo source (https://github.com/vaadin/custom-theme-demo)
3) test embedding source (https://github.com/vaadin/flow/tree/master/flow-tests/test-embedding)

**B. Like in demo and test source I'm using an external font**
I'm using Google Poppins font. I put all needed ttf files into my theme folder:
image

**C. I'm including this font** 
In `document.css`:
```
@font-face {
  font-family: "Poppins";
  font-style: normal;
  font-weight: 400;
  src: url("./webbase/font/poppins/Poppins-Regular.ttf") format("TrueType");
}
@font-face {
  font-family: "Poppins";
  font-style: normal;
  font-weight: 500;
  src: url("./webbase/font/poppins/Poppins-Medium.ttf") format("TrueType");
}
....
```
**D. All widgets classes (which extends `WebComponentExporter`) use my theme** 
via `@Theme("zaiks")` annotation.

**E. The same components (widgets) are added to the normal Views in my vaadin app**
so I can route to this view and see my widgets as normal internal components.

**F. My Vaadin app is deployed / served on serv1** (Tomcat)

**G. The external app is deployed / served on serv2** (IIS)
This exteernal app embeds my widgets served from serv1.
 
**The result:**
When I'm running my Vaadin app from serv1 and route to a View everything is ok (external font are used) .

But when I open a page from the external app on serv2 which embeds my widget is shown, but without my fonts, and in dev console I can see an error:
`GET https://<serv2>/<external app>/VAADIN/static/themes/zaiks/webbase/font/poppins/Poppins-Regular.ttf net::ERR_ABORTED 404 (Not Found)
`
Its because font is search in `<serv2>/<external app>` but it is in `<serv1>/<vaadin app>`:
`http://<serv1>/<vaadin app>/VAADIN/static/themes/zaiks/webbase/font/poppins/Poppins-Regular.ttf`

Just run test embedding (https://github.com/vaadin/flow/tree/master/flow-tests/test-embedding) but in scenario in which embedding app is on a different server than embedded web components are deployed.

caalador:
Add fix for using a external css file by having
the `@import url(...)` in a css file.

fixes #10228


flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -47,11 +47,37 @@ const getStyleModule = (id) => {
   return cssText;
 };
 `;
+const createLinkReferences = `
+const createLinkReferences = (css, target) => {
+  // Unresolved urls are written as '@import url(text);' to the css
+  const importMatcher = /\\@import\\surl\\((.+?)\\);/g;
```
pleku, line 53: What about "conditional imports", do they work like intended, like
```css
@import url('https://fonts.googleapis.com/css?family=Poppins') print;
```
in case someone wants to use a different font based on certain conditions
https://developer.mozilla.org/en-US/docs/Web/CSS/@import#importing_css_rules_conditionally

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -47,11 +47,37 @@ const getStyleModule = (id) => {
...
+  
+  var match;
+  var styleCss = css;
+  
+  // For each external url import add a link reference
+  while((match = importMatcher.exec(css)) !== null) {
+    styleCss = styleCss.replace(match[0], "");
+    const link = document.createElement('link');
```
pleku, line 61: So basically any file that webpack fails to discover locally, ends up here and is made to a link. Should it be documented ? I considered giving a warning/error if a relative url is discovered here but I guess it can be intentional.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
caalador: No we don't support that as we don't have any good way to match conditionals.

caalador: Seem webpack wraps it as `@media print {@import url(https://fonts.googleapis.com/css?family=Poppins);}`

caalador: So it would be more complex, but in theory if it would be needed then we could probably parse the `@media` and add it as an attribute `media=""` to the generated link.

pleku: Hmm. And then it will be broken ? I think it is (not that common) case anyway so it should be enough if the behavior documented

caalador: Having the media query there will not make the import disappear, but it will now be added without the query string.

Pushed a fix for media queries as in the end it wasn't that much of a problem.

PR #10968, opened 2021-05-10
fix: make ts email validator accept undefined value

haijian-vaadin:
This change makes the TypeScript email validator accept undefined values. It fixes the runtime error when validating an empty email field on the UI, since the field is now undefined by default.
Added IT tests in https://github.com/vaadin/spring/pull/799

### DIFF
--- a/flow-client/src/main/frontend/form/Validators.ts
+++ b/flow-client/src/main/frontend/form/Validators.ts
@@ -66,8 +66,8 @@ abstract class NumberValidator<T> extends AbstractValidator<T> {
   constructor(attrs?: ValidatorAttributes) {
     super(attrs);
   }
-  validate(value: T) {
-    return isNumeric(String(value));
+  validate(value: T | undefined) {
+    return value === undefined || isNumeric(String(value));
   }
 }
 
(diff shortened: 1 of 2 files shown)


Haprog:
The change looks ok, but I'd like to see code for an actual use case with an input/field where this fixes an issue to verify that this is needed/useful in reality. The new test just calls validator directly to check that it works with a mock value which is a bit too low level.

I'm just wondering how this gets executed in an actual use case since native `<input>` and input based Vaadin components like `<vaadin-email-field>` can't have `undefined` as the value. If you set the `nativeInput.value = undefined` the value will automatically be converted to string `"undefined"` and in case of `<vaadin-email-field>` it could be automatically converted to empty string `""` by the component. The value of `<vaadin-text-field>` based components are always of type string.

haijian-vaadin:
Good comment, I will add some IT tests for this. For now, to verify the fix, you can do it manually with the crm tutorial project
1. `git clone git@github.com:vaadin-learning-center/crm-tutorial-typescript.git`
2. `git checkout test-default-nullable`
3. Go to List view, click the **Add Contact** button, in the contact form, fill the first name with some value, then click Save
4. See that there is an js error in the console as shown in the attached screenshot and the form is not validated correctly.
Screenshot 2021-05-12 at 10 43 29
5. Change the flow version in the pom.xml to be the PR snapshot `8.0.pr-10968-SNAPSHOT`
6. Rerun the application, see that the form shows validation errors. (The error in the console is now the expected validation error)

Note, due to a bug in the offline login logic, you might need to clear the cache and refresh the page before logging in.

platosha (approved):

Issue #10892, opened 2021-05-04
Extract development mode server into a separate Maven module 


joheriks:
`DevModeHandler`, `DevModeInitializer` and related Java classes managing the Webpack development server should be moved out of `flow-server` and into a new module called `development-mode`.

Considerations:
- Move current `flow-server` dependencies related only to development mode (e.g. swagger) to this module, so these are excluded from production builds.
- Package root needs rename (currently `DevModeInitializer` is in `com.vaadin.flow.server.startup`, `DevModeHandler` in `com.vaadin.flow.server`)

fluorumlabs:
Suggestion to use `foundation-` prefix for common modules

PR #10969, opened 2021-05-10
feat: extract DevModeHandler and BrowserLiveReload into a separate module

joheriks:
The main modifications:
- `DevModeHandler` is now an interface in `flow-server`, and the implementation class in `vaadin-dev-server` has been renamed to `DevModeHandlerImpl`.
- `BrowserLiveReloadImpl` has been moved to the new module.
- There are two factory interfaces in `flow-server`: `DevModeHandlerAccess` and `BrowserLiveReloadAccess`. These are supplied with concrete instances via `LookupServletContainerInitializer `.
- `DevModeHandlerImpl` is registered via a `VaadinServiceInitListener` (impl in `DevModeRequestHandlerInitializer`).
 
TODO:
- [ ] Fix Spring add-on (in vaadin/spring#817).
- [x] Add module to dev-mode ITs.
- [x] Live reload in JRebel breaks (expects `com.vaadin.flow.internal.BrowserLiveReloadAccess` to be interface), need to make backwards-compatible solution.
- [x] Verify compatibility with HotswapAgent.

Part of #10892


flow-server/src/main/java/com/vaadin/flow/internal/BrowserLiveReloadAccess.java
```diff
@@ -30,39 +34,30 @@
...
-                    "BrowserLiveReloadAccess::getLiveReload is called in production mode.");
-            return null;
-        }
-        if (!service.getDeploymentConfiguration()
-                .isDevModeLiveReloadEnabled()) {
-            LoggerFactory.getLogger(BrowserLiveReloadAccess.class).debug(
-                    "BrowserLiveReloadAccess::getLiveReload is called when live reload is disabled.");
-            return null;
-        }
+    static BrowserLiveReload getLiveReloadIfAvailable(VaadinService service) {
```
denis-anisimov, line 56: I would better use `Optional<> getLiveReload` instead.

flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerAccess.java
```diff
@@ -0,0 +1,44 @@
...
+    DevModeHandler getDevModeHandler(VaadinService service);
+
+    /**
+     * Create a {@link DevModeHandler} is factory available.
+     *
+     * @param service
+     *            a service
+     * @return a <code>DevModeHandler</code> instance or null if disabled
+     */
+    static DevModeHandler getDevModeHandlerIfAvailable(VaadinService service) {
```
denis-anisimov, line 37: Same thing : return `Optional` instead and avoid `IfAvailable` .

flow-server/src/main/java/com/vaadin/flow/internal/TestInterface.java
```diff
@@ -0,0 +1,6 @@
+package com.vaadin.flow.internal;
+
+public interface TestInterface {
```
denis-anisimov, line 3: Leftover ?

flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
```diff
@@ -95,12 +95,6 @@ protected VaadinServletService() {
                         e);
             }
         }
-        if (getDeploymentConfiguration().enableDevServer()) {
```
denis-anisimov, line 98: Is it intentional ?
What is the replacement ?
There will be issues without using dev mode handler as a `RequestHandler`.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdateTestUtil.java
```diff
@@ -71,87 +68,6 @@ static ClassFinder getClassFinder(Class<?>... classes)
         return classPaths.toArray(new URL[0]);
     }
 
-    // Creates stub versions of `node` and `npm` in the ./node folder as
-    // frontend-maven-plugin does
-    // Also creates a stub version of webpack-devmode-server
-    public static void createStubNode(boolean stubNode, boolean stubNpm,
```
denis-anisimov, line 77: Question: Is it moved somewhere ?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java
```diff
@@ -386,10 +391,12 @@ public static void initDevModeHandler(Set<Class<?>> classes,
         CompletableFuture<Void> nodeTasksFuture = CompletableFuture
                 .runAsync(runnable);
 
-        DevModeHandler.start(
+        BrowserLiveReload liveReload = context
```
denis-anisimov, line 394: Wrong access to live reload.
It uses impl instead of SPI/API.
Impl may be changed . SPI/API is a contract.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java
```diff
@@ -386,10 +391,12 @@ public static void initDevModeHandler(Set<Class<?>> classes,
...
+                .getAttribute(BrowserLiveReload.class);
+        DevModeHandlerImpl.start(
```
denis-anisimov, line 396: lookup passed as an argument can be extended with `liveReload` instead of using extra arg.
Or may be use (provided) `lookup` to get the liveReload from there directly.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeRequestHandlerInitializer.java
```diff
@@ -0,0 +1,37 @@
...
+import com.vaadin.flow.function.DeploymentConfiguration;
+import com.vaadin.flow.server.ServiceInitEvent;
+import com.vaadin.flow.server.VaadinServiceInitListener;
+
+public class DevModeRequestHandlerInitializer
+        implements VaadinServiceInitListener {
+
+    @Override
+    public void serviceInit(ServiceInitEvent event) {
+        DevModeHandlerImpl devModeHandler = DevModeHandlerImpl
```
denis-anisimov, line 28: Impl is used to get the dev mode handler.
SPI/API must be used.

The reason is quite simple: the whole idea of SPI is providing a way to implement things differently.
In OSGi e.g. there can be totally different implementation for things.
That's why impl may not be used directly.

denis-anisimov (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdateTestUtil.java
denis-anisimov: Never-mind, found it.

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -64,7 +64,7 @@
     private final List<String> manifestPaths;
 
     // Matcher to match string starting with '/themes/[theme-name]/'
-    protected static final Pattern APP_THEME_PATTERN = Pattern
+    public static final Pattern APP_THEME_PATTERN = Pattern
```
denis-anisimov, line 67: Any specific reason to expose this as `public` ?

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -51,7 +51,7 @@
      * The pattern of error message shown when the URL path contains unsafe
      * double encoding.
      */
-    static final String UNSAFE_PATH_ERROR_MESSAGE_PATTERN = "Blocked attempt to access file: {}";
+    public static final String UNSAFE_PATH_ERROR_MESSAGE_PATTERN = "Blocked attempt to access file: {}";
```
denis-anisimov, line 54: Any specific reason to expose it as `public` ?

flow-server/src/main/java/com/vaadin/flow/internal/BrowserLiveReloadAccess.java
joheriks: Done.

flow-server/src/main/java/com/vaadin/flow/internal/TestInterface.java
joheriks: Yep, removed.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
joheriks: There is a registration of `DevModeHandlerImpl` as a request handler instead (via `VaadinServiceInitListener`) (see `DevModeRequestHandlerInitializer` and last comment).

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java
joheriks: Done.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeRequestHandlerInitializer.java
joheriks: I disagree here. That a `DevModeRequestHandlerInitializer` is used to register it as a handler (and even that `DevModeHandlerImpl` implements `RequestHandler`) is specific to `DevModeHandlerImpl`. It is not part of the contract given by the interface `DevModeHandler` (a replacement `DevModeHandlerImpl` would not necessarily register itself as a `RequestHandler`).

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
joheriks: It is because of package move and split into modules. Dev server needs this pattern to know if it should proxy to webpack instead of serve the file statically.

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
joheriks: Same reason as above... but here the case is weaker since it's only used for log message. Duplicated the string instead and revert to package-private.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeRequestHandlerInitializer.java
denis-anisimov: A  number of issues here:

- `event.addRequestHandler(devModeHandler)` does it guarantee the order in which `RequestHandler`s are called?  It's important that dev mode handler is called before the bootstrap handler e.g.
- how `DevModeRequestHandlerInitializer` instance is registered ? I can't find this here via Github.

Again: keep in mind OSGi.
The things which you expect to work out of the box in plain WAR case doesn't work sometimes in OSGi.

E.g. OSGi doesn't invoke `ServletContextInitListener` at all.
So you may not relay on this.

We don't support dev server/dev mode in OSGi at the moment but the design should be done so that it requires to add only one piece of things instead if rewriting everything .
Previously the code which registered the `RequestHandler` was inside the service itself and there was no need to write it once again for OSGi.

The question is : will this code work in OSGi as well ?
If it doesn't then will I have to write it once again specifically for OSGi ?
Or is it possible to make an abstraction so that I don't need to write the same code twice ?

Another question : can `DevModeHandlerImpl` be reused in OSGi as is ?
If yes then it's fine.
If no and `DevModeRequestHandlerInitializer` is executed in OSGi as well then you may not use it here.

There should be SPI and API. SPI should be used to be able to implement things differently in different containers.
API should be common.
The generic/common functionality should be extracted out of the SPI so there is no need to duplicate the same code for another impl.

I'm not sure that the current design makes SPI which can be implemented differently.
Most likely it allows to use only one impl which doesn't have to be present at runtime and it means that total redesign will be required once again for OSGi case.

denis-anisimov:
Another thing which I don't understand: how do you register services in Lookup ?
I can't find any code which stores the impls inside Lookup.

You should either modify `LookupServletContainerInitializer` or use Java SPI services mechanism to load services (and the latter one always creates a new service instance being asked for a service).

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java
joheriks: Done.

joheriks:
> Another thing which I don't understand: how do you register services in `Lookup` ?
> I can't find any code which stores the impls inside Lookup.
> 
> You should either modify `LookupServletContainerInitializer` or use Java SPI services mechanism to load services (and the latter one always creates a new service instance being asked for a service).
> 
> There is no any magic: the service won't appear automagically. It needs to be registered explicitly.

`Lookup` delegates to Java's `ServiceLoader` (SPI), and the service mappings are in the following files:
https://github.com/vaadin/flow/pull/10969/files#diff-9c8ce3cb6bb6a74b88e4a412b3c00aac1d7600f94529332989dfe1814c03e217
https://github.com/vaadin/flow/pull/10969/files#diff-6699fbbd3893fdc01cedf94dde531857f4ca9cefa79fef110ba683e7cac1bdf9

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeRequestHandlerInitializer.java
joheriks: > event.addRequestHandler(devModeHandler) does it guarantee the order in which RequestHandlers are called? It's important that dev mode handler is called before the bootstrap handler e.g

The API of `ServiceInitEvent::addRequestHandler` does give this guarantee:

https://github.com/vaadin/flow/blob/83ebf7b7a920ce098b9048d10a7d32813d2719e4/flow-server/src/main/java/com/vaadin/flow/server/ServiceInitEvent.java#L57-L70

Not sure about OSGI---but I can check it. What would be the standard way of adding a request handler in another module / bundle? I assume I will also need to add a `bnd.bnd` file in `vaadin-dev-server`.

denis-anisimov:
> > Another thing which I don't understand: how do you register services in `Lookup` ?
> > I can't find any code which stores the impls inside Lookup.
> > You should either modify `LookupServletContainerInitializer` or use Java SPI services mechanism to load services (and the latter one always creates a new service instance being asked for a service).
> > There is no any magic: the service won't appear automagically. It needs to be registered explicitly.
> 
> `Lookup` delegates to Java's `ServiceLoader` (SPI), and the service mappings are in the following files:
> https://github.com/vaadin/flow/blob/d51d8d29cdb2188c8235496254ff7c2670100b0d/vaadin-dev-server/src/main/resources/META-INF/services/com.vaadin.flow.internal.DevModeHandlerAccess
> https://github.com/vaadin/flow/blob/d51d8d29cdb2188c8235496254ff7c2670100b0d/vaadin-dev-server/src/main/resources/META-INF/services/com.vaadin.flow.internal.BrowserLiveReloadAccess

Ah, now I see, thanks.

`ServiceLoader` (SPI) doesn't work in OSGi and this should be avoided as much as possible in the internal Flow code (unless it's supposed to work _ONLY_ in WAR case).

Couple of things to be aware of:
- Java SPI (`ServiceLoader`) mechanism used in `Lookup`  impl doesn't cache the services . Every time when service is asked via `lookup` method the new instance is created. Most likely it's OK in this specific case. 
- Java SPI mechanism has been designed more for the app developer usage who is not able to modify `LookupServletContainerInitializer` . For our internal usage I would recommend to use `LookupServletContainerInitializer`  where the classes are scanned via `ServletContainerInitializer` Servlet API.
- Neither (Java SPI/`ServletContainerInitializer` )  works in OSGi.

Now let's got through the services:
- `com.vaadin.flow.internal.BrowserLiveReloadAccess` service: doesn't work in OSGi _NOW_. Previously it worked even though I'm not sure whether it does anything currently at all in OSGI.
- `com.vaadin.flow.internal.DevModeHandlerAccess` service doesn't work in OSGi but dev mode handler never worked in OSGi. This is fine.
- `com.vaadin.flow.server.VaadinServiceInitListener ` service doesn't work in OSGi. This functionality before it has been extracted worked in OSGi .  Now if dev mode is implemented for OSGi one has to care about this quite not-obvious extra thing. This complicates dev mode impl for OSGi and I would like to have it better designed: dev mode handler would have one Dev mode interface which being implemented once doesn't require any extra action. It means that I don't want to care about `com.vaadin.base.devserver.startup.DevModeRequestHandlerInitializer` OSGi specific implementation to register an init listener which register a handler. I would like implement one interface and it will be added as a `RequestHandler` out of the box.
- `javax.servlet.ServletContainerInitializer` doesn't work in OSGi so `DevModeInitializer` won't work . But this is fine : it's never worked.

So at the moment : `BrowserLiveReloadAccess` is questionable since the current approach breaks this functionality in OSGi.
This can be fixed via implementing this access for OSGI as well. But then this requires a ticket: this issue should be known.

Everything related to dev mode handler doesn't work in OSGi: that's fine since it didn't work.

`DevModeRequestHandlerInitializer`  complicates dev mode implementation for OSGi since extra action is required : add a handler and this can be done with a proper design without need to have `DevModeRequestHandlerInitializer`  at all.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeRequestHandlerInitializer.java
denis-anisimov: The ordering : it's fine.

`DevModeRequestHandlerInitializer`  in OSGI : it won't work (at least without additional dance which should be avoided).
Now I see that it's registered via Java standard SPI which doesn't work in OSGi out of the box.

And I already commented this generic approach of using Java SPI: it doesn't work in OSGi.
So `DevModeRequestHandlerInitializer` won't do its job in OSGi and the implementor of dev mode has to care about this explicitely as an extra action.

Even though generally I agree that avoiding any usage of dev mode classes in flow-server is preferable approach I would like to implement only one SPI and register it properly.
Here I have to implement `DevModeHandler` and then also care about registering `RequestHandler`'s part of its functionality in addition.

Question: does this PR completely removed all dev mode related classes from `flow-server` module ?

As I understand there are still extracted `BrowserReload` interface references in `flow-server`.

So :
- if there is no any single dev mode class/interface referenced from `flow-server` then let's go with the current approach. Because otherwise Dev mode `RequestHandler` interface should be returned back and `flow-server` will know about this specific case.
- if `flow-server` already has some references to dev mode via some interfaces then I don't see any point to have extra `DevModeRequestHandlerInitializer`: just dereference the dev mode handler `RequestHandler` implementation via some interface (in fact `Lookup` may be asked for any `RequestHandler` in the code which creates them) and use it via this interface inside `VaadinService`. The only problem here is : request handlers may not be referenced directly via the List of objects since on OSGi any service object may disappear at any time when bundle is undeplpyed.

flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
joheriks: Restored the handler addition (`DevModeHandler` interface now extends `RequestHandler`, no longer use `VaadinServiceInitListener`).

joheriks:
I've made the following changes to the PR:
- Added `BrowserLiveReloadAccess` and `DevModeHandlerAccess` to `LookupServletContainerInitializer` instead of relying on SPI service discovery.
- Don't use `VaadinServiceInitListener` to register the devmode handler as a request handler

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/BrowserLiveReloadFactoryImpl.java
```diff
@@ -0,0 +1,61 @@
...
+ * Default implementation for {@link BrowserLiveReloadFactory} that stores the
+ * instance in the Vaadin context.
+ */
+public class BrowserLiveReloadFactoryImpl implements BrowserLiveReloadFactory {
+
+    @Override
+    public BrowserLiveReload getLiveReload(VaadinContext context) {
+        BrowserLiveReload liveReload = context
+                .getAttribute(BrowserLiveReload.class);
+        if (liveReload == null) {
```
denis-anisimov, line 35: This is not thread-safe.

denis-anisimov: So is it a factory or an access ?
factory semantic is `create` and this one caches.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerFactoryImpl.java
```diff
@@ -0,0 +1,53 @@
...
+ * <p>
+ * For internal use only. May be renamed or removed in a future release.
+ *
+ * @author Vaadin Ltd
+ * @since
+ */
+public class DevModeHandlerFactoryImpl implements DevModeHandlerFactory {
+
+    @Override
+    public DevModeHandler getDevModeHandler(VaadinService service) {
```
denis-anisimov, line 36: Is it a `factory` or an `access` ?
Not a factory semantic.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/internal/BrowserLiveReloadFactory.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Returns a {@link BrowserLiveReload} instance for the given
+     * {@code context}.
+     *
+     * @param context
+     *            a Vaadin context
+     * @return a <code>BrowserLiveReload</code> instance or null if disabled
+     */
+    BrowserLiveReload getLiveReload(VaadinContext context);
```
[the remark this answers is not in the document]
joheriks: The `BrowserLiveReloadAccess` class is there for backwards compatibility with JReben and HotswapAgent---otherwise these tools won't work since they assume there is class (not interface) with this name on the classpath.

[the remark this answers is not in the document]
joheriks: The -> `*Factory` rename was just to avoid this clash; better name suggestion welcome.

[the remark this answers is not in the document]
denis-anisimov: Aaaah.
OK.

I suggest don't use factory because of the semantic.
Let's keep just access classes which do both get (if exists) and create (if not yet).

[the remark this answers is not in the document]
denis-anisimov: Even though I would prefer to use something better than `Access`

flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerFactory.java
```diff
@@ -0,0 +1,62 @@
...
+public interface DevModeHandlerFactory {
+
+    /**
+     * Returns a {@link DevModeHandler} instance for the given {@code service}.
+     *
+     * @param service
+     *            a Vaadin service
+     * @return a {@link DevModeHandler} instance
+     */
+    DevModeHandler getDevModeHandler(VaadinService service);
```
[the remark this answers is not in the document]
joheriks: Renamed the two interfaces now to `*Accessor` (`BrowserLiveReloadAccess` not workable as already taken by the backwards-compatibility class)

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/BrowserLiveReloadFactoryImpl.java
joheriks: Added sync on context.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerFactoryImpl.java
joheriks: Renamed interfaces.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -460,7 +461,10 @@ public static String getStatsContent(VaadinService service)
 
         try {
             if (!config.isProductionMode() && config.enableDevServer()) {
-                content = getStatsFromWebpack();
+                Optional<DevModeHandler> devModeHandler = DevModeHandlerAccessor
+                        .getDevModeHandlerFromService(service);
```
denis-anisimov, line 465: I would avoid long method name :`FromService`  suffix is extra here.
But this is not a blocker.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/BrowserLiveReloadAccessorImpl.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Default implementation for {@link BrowserLiveReloadAccessor} that stores the
+ * instance in the Vaadin context.
+ */
+public class BrowserLiveReloadAccessorImpl
+        implements BrowserLiveReloadAccessor {
+
+    @Override
+    public BrowserLiveReload getLiveReload(VaadinContext context) {
+        synchronized (context) {
```
denis-anisimov, line 34: There is already synchronization in the impl 
`public <T> T getAttribute(Class<T> type, Supplier<T> defaultValueSupplier)` 
And the semantic of the interface method assumes that the method is called atomically.
So I suggest to use the existing way of lazy init instead of duplicating the code .

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
joheriks: Renamed.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/BrowserLiveReloadAccessorImpl.java
joheriks: Nice suggestion, I missed that one. Fixed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
joheriks: I've also introduced some additional changes to `DevModeHandlerAccessor` that I needed for Spring Boot add-on ([draft PR](https://github.com/vaadin/spring/pull/817) gives the motivation):
- rename`DevModeHandlerAccessor` -> `DevModeHandlerManager`, as is no longer a simple accessor.
- added `getHandlesTypes` and `initDevModeHandler` to `DevModeHandlerManager` to allow custom discovery implementation.
- removed `VaadinService` parameter from `DevModeHandler::getDevModeHandler` as it was not used expect for logging.
- added `stop()` method to the `DevModeHandler `interface.

denis-anisimov (approved):

joheriks:
@denis-anisimov Can I bother you for a new approval, rebased on top of latest master and added a test for the JRebel / HotswapAgent compatibility class.

fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java
```diff
@@ -64,7 +64,7 @@
     @SuppressWarnings({ "unchecked", "rawtypes" })
     public void setup() throws Exception {
         assertNull("No DevModeHandler should be available at test start",
-                DevModeHandler.getDevModeHandler());
+                DevModeHandlerImpl.getDevModeHandler());
```
denis-anisimov, line 67: I'm not sure whether it's important or can be done later but I think this test should be rewritten instead 
just replacing `DevModeHandler` to `DevModeHandlerImpl`.

The test is inside `fusion` . It doesn't (or it should not ) know anything about dev mode handler real imp.
And it should not depend on the impl module at all.
`DevModeHandler` has an interface: it should be mocked and communication with mock should be tested instead of testing the real `DevModeHandlerImpl`: this is the additional point of the extracting things. Once dependency is removed then there is no anymore to care about real implementation: only interface communication which can be implemented anyhow.

denis-anisimov (changes requested):

fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java
joheriks: The tests here are verifying that endpoint TS is generated when `DevModeInitializer::onStartup` is called, so as such they test the specific implementation `DevModeHandler`. I don't see a way to mock  `DevModeHandler` and still have a relevant tests. I would rather suggest that the test should be moved to `vaadin-dev-server`.

denis-anisimov: >I would rather suggest that the test should be moved to vaadin-dev-server.

I'm fine with that.

I didn't go into details but:
>The tests here are verifying that endpoint TS is generated when 

Do you _really_ need that the endpoint TS _is generated_ or you need to check that some endpoint method is called ?

As I said: I'm not familiar with details but once abstraction layer is extracted you may test two things separately:
- some interface method is invoked (being mocked) in the place where the communication is done
- this interface is implemented properly (so the implementation does something which you expect e.g. generates a file) : this should be done in completely another place where the implementation is provided.

May be in this case the test indeed should be just moved into another place.

joheriks: Moved the test to `vaadin-dev-server` but didn't change the test so `fusion-endpoint` functionality is still tested. So the issue of module isolation in tests is not solved, but it should be easier to mock only `EndpointGeneratorTaskFactory` interface later on if needed.

joheriks: I'll look into mocking `EndpointGeneratorTaskFactory` still, but I don't wan to spend too much time on it

denis-anisimov (approved):

fusion-endpoint/src/test/java/com/vaadin/flow/server/startup/fusion/DevModeInitializerEndpointTest.java
denis-anisimov: Ok

joheriks: Gave up on mocking the interface, it is easily doable but weakens the test case too much (in fact, this should be considered an IT test).

PR #10973, opened 2021-05-11
fix: Remove old router layouts content before navigation

mshabarov:
## Description

Recursively removes content of the all nested router layouts of the given old content when the navigation occurs. 
This is needed to let Dependency Injection frameworks (Spring and CDI add-ons) to re-create managed components with no duplicates/leftovers.
Consider the following router target chain before navigation:
```
MainRouterLayout
|__SubRouterLayout
     |__MyRouteView
```
And after navigation the whole chain can be replaced:
```
AnotherRouterLayout
|__AnotherRouteView
```
Or some part of it:
```
MainRouterLayout
|__AnotherRouteView
```
Route target `MyRouteView` will be removed before navigation in order to let DI framework to insert it by itself with no leftovers, depending on the scope of `MyRouteView`. It might be a new instance of `MyRouteView` or a preserved one.

Fixes https://github.com/vaadin/cdi/issues/345

Co-authored-by: @roeltje25

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -691,7 +692,7 @@ public class UIInternals implements Serializable {
 
         // Assemble previous parent-child relationships to enable detecting
         // changes
-        Map<RouterLayout, HasElement> oldChildren = new HashMap<>();
+        Map<RouterLayout, HasElement> oldChildren = new IdentityHashMap<>();
         for (int i = 0; i < routerTargetChain.size() - 1; i++) {
             HasElement child = routerTargetChain.get(i);
             RouterLayout parent = (RouterLayout) routerTargetChain.get(i + 1);
@@ -1107,18 +1108,19 @@ public class UIInternals implements Serializable {
     }
 
     private void removeChildrenContentFromRouterLayout(
-            RouterLayout routerLayout,
-            Map<RouterLayout, HasElement> oldChildren) {
-        HasElement oldContent = oldChildren.get(routerLayout);
+            final RouterLayout targetRouterLayout,
+            final Map<RouterLayout, HasElement> oldChildren) {
+        HasElement oldContent = oldChildren.get(targetRouterLayout);
+        RouterLayout removeFrom = targetRouterLayout;
         // Recursively remove content of the all nested router
         // layouts of the given old content to be detached. This
         // is needed to let Dependency Injection frameworks to
         // re-create managed components with no
         // duplicates/leftovers.
         while (oldContent != null) {
-            routerLayout.removeRouterLayoutContent(oldContent);
+            removeFrom.removeRouterLayoutContent(oldContent);
             if (oldContent instanceof RouterLayout) {
-                routerLayout = (RouterLayout) oldContent;
+                removeFrom = (RouterLayout) oldContent;
             }
             oldContent = oldChildren.get(oldContent);
         }
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -691,7 +692,7 @@ public class UIInternals implements Serializable {
 
         // Assemble previous parent-child relationships to enable detecting
         // changes
-        Map<RouterLayout, HasElement> oldChildren = new HashMap<>();
+        Map<RouterLayout, HasElement> oldChildren = new LinkedHashMap<>();
         for (int i = 0; i < routerTargetChain.size() - 1; i++) {
             HasElement child = routerTargetChain.get(i);
             RouterLayout parent = (RouterLayout) routerTargetChain.get(i + 1);
@@ -704,6 +705,17 @@ public class UIInternals implements Serializable {
 
         if (layouts != null) {
             routerTargetChain.addAll(layouts);
+        }
+
+        // If the old and the new router target chains are not intersect,
+        // meaning that the new chain doesn't contain the root router
+        // layout node of the old chain, this aims to recursively remove
+        // content of the all nested router layouts of the given old content
+        // to be detached. This is needed to let Dependency Injection
+        // frameworks to re-create managed components with no
+        // duplicates/leftovers.
+        if (oldRoot == null || !routerTargetChain.contains(oldRoot)) {
+            oldChildren.forEach(RouterLayout::removeRouterLayoutContent);
         }
 
         // Ensure the entire chain is connected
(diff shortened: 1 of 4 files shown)


denis-anisimov (commented):
It might be that the change makes sense but it requires a lot of unit tests which checks that
all the layouts (the topmost and interim) are cleaned up properly.

It also requires ITs even though they will be quite hacky since there are no scopes in Flow and the components are recreated every time (removed are not reused).

One option would be to implement some custom scope without any DI (for pure Flow) and use a custom `Instantiator` which use the scope. That's doable and not that hard: just make sure that other tests are not affected.

Or hacky way: use `UI.getCurrent` in detach listeners of the removed components to set info in the "attached" UI even from the component which is detached.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -691,7 +692,7 @@ public void showRouteTarget(Location viewLocation, Component target,
 
         // Assemble previous parent-child relationships to enable detecting
         // changes
-        Map<RouterLayout, HasElement> oldChildren = new HashMap<>();
+        Map<RouterLayout, HasElement> oldChildren = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
mshabarov: The order is needed here `oldChildren.forEach(RouterLayout::removeRouterLayoutContent);` when I want to remove the children content from their parents, starting from most bottom layouts and ending with top most layouts.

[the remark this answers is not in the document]
mshabarov: However, I've double checked and it indeed doesn't matter in which order to detach the contents from its parents.
Will revert it back to `HashMap`

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -725,9 +737,19 @@ public void showRouteTarget(Location viewLocation, Component target,
...
+                    // layouts of the given old content to be detached. This
+                    // is needed to let Dependency Injection frameworks to
+                    // re-create managed components with no
+                    // duplicates/leftovers.
+                    while (oldContent != null) {
                         layout.removeRouterLayoutContent(oldContent);
+                        if (oldContent instanceof RouterLayout) {
+                            layout = (RouterLayout) oldContent;
+                        }
+                        oldContent = oldChildren.get(oldContent);
```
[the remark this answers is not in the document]
mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -692,7 +692,7 @@ public void showRouteTarget(Location viewLocation, Component target,
 
         // Assemble previous parent-child relationships to enable detecting
         // changes
-        Map<RouterLayout, HasElement> oldChildren = new LinkedHashMap<>();
+        Map<RouterLayout, HasElement> oldChildren = new HashMap<>();
```
denis-anisimov: I recommend to use `IdentityHashMap` here since you don't need `equals` and `hashCode` but the identity of the objects.
Don't forget that `RouterLayout` is implemented by the app developer and he may do mistakes in `equals` and `hashCode` which may break the implementation.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -1118,4 +1106,22 @@ private void configurePush(HasElement root) {
...
+        HasElement oldContent = oldChildren.get(routerLayout);
+        // Recursively remove content of the all nested router
+        // layouts of the given old content to be detached. This
+        // is needed to let Dependency Injection frameworks to
+        // re-create managed components with no
+        // duplicates/leftovers.
+        while (oldContent != null) {
+            routerLayout.removeRouterLayoutContent(oldContent);
+            if (oldContent instanceof RouterLayout) {
+                routerLayout = (RouterLayout) oldContent;
```
denis-anisimov, line 1121: Don't mutate the parameter provided in the method .
Make a new one for mutation. The good code practice: make method parameters effectively final.

denis-anisimov (changes requested):
No ITs ?

mshabarov:
> No ITs ?

Here is an IT:
https://github.com/vaadin/flow/pull/10973/files#diff-fb6cd17d229cc75410811ed67e2a0c86e3d7c3b95e31dcd83a1f128e9ad77cef

flow-tests/test-common/src/main/java/com/vaadin/flow/uitest/servlet/RouterLayoutCustomScopeServlet.java
```diff
@@ -0,0 +1,236 @@
...
+            // Cleanup the components context upon detaching UI
+            ui.addDetachListener(event -> {
+                Map<Class<? extends Component>, Component> removed = components
+                        .remove(event.getUI().getUIId());
+                removed.clear();
+            });
+        }
+
+        Component getComponent(Class<? extends Component> componentType,
+                SerializableFunction<Class<? extends Component>, Component> instanceSupplier) {
```
denis-anisimov, line 104: This is a factory. So the parameter name should be `factory` .

flow-tests/test-common/src/main/java/com/vaadin/flow/uitest/servlet/RouterLayoutCustomScopeServlet.java
```diff
@@ -0,0 +1,236 @@
...
+            throws ServiceException {
+        RouterLayoutCustomScopeService routerLayoutCustomScopeService = new RouterLayoutCustomScopeService(
+                this, deploymentConfiguration);
+        routerLayoutCustomScopeService.init();
+        return routerLayoutCustomScopeService;
+    }
+
+    @Target(ElementType.TYPE)
+    @Retention(RetentionPolicy.RUNTIME)
+    public @interface CustomUIScope {
```
denis-anisimov, line 68: Since this is a test I would go without annotation which allows to configure the behavior.
It's enough to make the default behavior (UI) scoped e.g. : always store the instances .
That will work only for this servlet so there should be no any issues with that.

But that's not blocking , it's just complicates things a bit.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/RemoveRoutersLayoutContentIT.java
```diff
@@ -0,0 +1,79 @@
...
+import static com.vaadin.flow.uitest.servlet.RouterLayoutCustomScopeServlet.NAVIGATE_TO_ANOTHER_ROUTE_OUTSIDE_MAIN_LAYOUT_BUTTON_ID;
+import static com.vaadin.flow.uitest.servlet.RouterLayoutCustomScopeServlet.NAVIGATE_TO_ANOTHER_ROUTE_INSIDE_MAIN_LAYOUT_BUTTON_ID;
+import static com.vaadin.flow.uitest.servlet.RouterLayoutCustomScopeServlet.NAVIGATE_BACK_FROM_ANOTHER_ROUTE_OUTSIDE_MAIN_LAYOUT_BUTTON_ID;
+import static com.vaadin.flow.uitest.servlet.RouterLayoutCustomScopeServlet.NAVIGATE_BACK_FROM_ANOTHER_ROUTE_INSIDE_MAIN_LAYOUT_BUTTON_ID;
+import static com.vaadin.flow.uitest.servlet.RouterLayoutCustomScopeServlet.SUB_LAYOUT_ID;
+
+public class RemoveRoutersLayoutContentIT extends ChromeBrowserTest {
+
+    @Override
+    protected void open() {
```
denis-anisimov, line 38: Isn't it only enough to override `getTestPath` method ?

denis-anisimov (changes requested):

denis-anisimov:
> > No ITs ?
> 
> Here is an IT:
> https://github.com/vaadin/flow/pull/10973/files#diff-fb6cd17d229cc75410811ed67e2a0c86e3d7c3b95e31dcd83a1f128e9ad77cef
> 
> I'm also going to add a regression IT to the CDI add-on repo.

Ah, github showed my only a piece of the changes by the link from e-mail.
OK, thanks.

flow-tests/test-common/src/main/java/com/vaadin/flow/uitest/servlet/RouterLayoutCustomScopeServlet.java
mshabarov: Renamed

flow-tests/test-common/src/main/java/com/vaadin/flow/uitest/servlet/RouterLayoutCustomScopeServlet.java
mshabarov: Annotation excluded and appropriate comment added

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/RemoveRoutersLayoutContentIT.java
mshabarov: Yes, it's enough, thanks! Fixed.

denis-anisimov (approved):

PR #11033, opened 2021-05-19
fix: Using a custom RouteRegistry

### LINKED ISSUE #11022: Not possible to extend AbstractRouteRegistry from Vaadin 14.5 onwards
If you want to implement your own routing, one needs to have CustomRouteRegistry extending AbstractRouteRegistry. 

In Vaadin 14.5 and newer versions this will require implementing

public NavigationRouteTarget getNavigationRouteTarget ( String url )
public RouteTarget getRouteTarget ( Class<? extends Component> target, RouteParameters parameters ) 

This will not be straightforward because NavigationRouteTarget does not have a public constructor.

caalador:
## Description

Make it possible to create a custom
RouteRegistry and add test module for it.

Fixes #11022

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/ErrorRouteRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/ErrorRouteRegistry.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.vaadin.flow.server;
+
+import java.io.Serializable;
+import java.util.Optional;
+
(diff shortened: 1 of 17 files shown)


flow-server/src/main/java/com/vaadin/flow/server/ErrorRouteRegistry.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.vaadin.flow.router.internal.ErrorTargetEntry;
+
+/**
+ * Interface class for RouteRegistries that can be used to request for error
+ * navigation views for Exceptions.
+ *
+ * @since
+ */
+public interface ErrorRouteRegistry extends Serializable {
```
pleku, line 30: 🦆 So previously error views feature was only used if you had extended `ApplicationRouteRegistry` but not if you had extended `AbstractRouteRegistry`. So now you can use either since `ApplicationRouteRegistry` extends this interface and now any custom registry can decide itself whether or not enables those. 👍🏻 as this is not a behavior or API change. But if anyone had managed to extend `ApplicationRouteRegistry`, they just need to compile it again.

pleku (approved):
LGTM, but I'm just wondering if we could avoid the _yet another IT module_ and instead just trust that a "unit test" would be enough - as in if you can write the code, we can expect it to work. And basically just add code that checks the logic for the added parts (error route registry)

caalador:
I feel that for this level of a feature the IT test module is justified as it needs to work on so many levels and be used in navigation etc. I wouldn't feel comfortable with only an it test as on that level we do test the usage with test registries that extend abstract registry.

pleku:
Just noting that this is safe to be picked to 2.6 as even if it adds new API (interface) it actually fixes two bugs with this.

pleku:
Maybe we don't have to pick this to `6.0` as `7.0` is not far away.

PR #11034, opened 2021-05-19
test: fix mockito errors accessing protected API in JDK16

manolo:
related to https://github.com/vaadin/flow-components/issues/997 
related to https://github.com/vaadin/flow-components/pull/998

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/component/webcomponent/WebComponentTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/webcomponent/WebComponentTest.java
@@ -16,11 +16,15 @@
 
 package com.vaadin.flow.component.webcomponent;
 
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
-import org.mockito.Matchers;
+import org.mockito.ArgumentMatchers;
 import org.mockito.Mockito;
 
 import com.vaadin.flow.component.Component;
@@ -30,9 +34,6 @@ import com.vaadin.flow.server.webcomponent.WebComponentBinding;
 
 import elemental.json.Json;
 import elemental.json.JsonValue;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.verify;
 
 public class WebComponentTest {
 
@@ -145,21 +146,24 @@ public class WebComponentTest {
                 element);
 
         webComponent.setProperty(intConfiguration, 1);
-        verify(element, Mockito.times(1)).executeJs(Matchers.anyString(),
-                Matchers.any(), Matchers.any());
+        verify(element, Mockito.times(1)).executeJs(
+                ArgumentMatchers.anyString(), ArgumentMatchers.any(),
+                ArgumentMatchers.any());
         webComponent.setProperty(doubleConfiguration, 1.0);
-        verify(element, Mockito.times(2)).executeJs(Matchers.anyString(),
-                Matchers.any(), Matchers.any());
+        verify(element, Mockito.times(2)).executeJs(
+                ArgumentMatchers.anyString(), ArgumentMatchers.any(),
+                ArgumentMatchers.any());
         webComponent.setProperty(stringConfiguration, "asd");
-        verify(element, Mockito.times(3)).executeJs(Matchers.anyString(),
-                Matchers.any(), Matchers.any());
+        verify(element, Mockito.times(3)).executeJs(
+                ArgumentMatchers.anyString(), ArgumentMatchers.any(),
+                ArgumentMatchers.any());
         webComponent.setProperty(booleanConfiguration, true);
-        verify(element, Mockito.times(4)).executeJs(Matchers.anyString(),
-                Matchers.any(), Matchers.any());
-        // JsonValue has a different number of arguments, so the invocation
-        // will be registered to a different overload
+        verify(element, Mockito.times(4)).executeJs(
+                ArgumentMatchers.anyString(), ArgumentMatchers.any(),
+                ArgumentMatchers.any());
+        // JsonValue has a different number of arguments
         webComponent.setProperty(jsonConfiguration, Json.create(true));
-        verify(element, Mockito.times(1)).executeJs(Matchers.anyString(),
-                Matchers.any());
+        verify(element, Mockito.times(5)).executeJs(
+                ArgumentMatchers.anyString(), ArgumentMatchers.any());
     }
 }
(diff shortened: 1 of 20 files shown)


manolo:
Notice that removing easymock is out of the scope of this PR, it should be done separately https://github.com/vaadin/flow/issues/11064

flow-server/src/test/java/com/vaadin/flow/component/webcomponent/WebComponentTest.java
```diff
@@ -145,21 +146,24 @@ public void setProperty_attemptsToWriteSupportedTypes() {
...
+                ArgumentMatchers.any());
         webComponent.setProperty(booleanConfiguration, true);
-        verify(element, Mockito.times(4)).executeJs(Matchers.anyString(),
-                Matchers.any(), Matchers.any());
-        // JsonValue has a different number of arguments, so the invocation
-        // will be registered to a different overload
+        verify(element, Mockito.times(4)).executeJs(
+                ArgumentMatchers.anyString(), ArgumentMatchers.any(),
+                ArgumentMatchers.any());
+        // JsonValue has a different number of arguments
```
mshabarov, line 164: This is a bit unclear: previously, a different `executeJs` overload was invoked (with 2 params), and that's why the expected `Mockito.times()` was 1. And this was explained in the comment.
Now the same overload is invoked (2 params) but for some reason it is expected to be invoked 5 times. This is not explained, but instead `// JsonValue has a different number of arguments` is present giving basically no explicit clarifications on why is the times = 5.
Do we really expect times = 5 here? And if yes, could you please add more comments explaining that.

mshabarov (changes requested):

flow-server/src/test/java/com/vaadin/flow/component/webcomponent/WebComponentTest.java
manolo: Actually, I don't know the exact cause, I guess mockito implenentation now does not differentiates by the signature of the method but the name, or something. The fact is that the test needs this change, and it still checks whether the correct code is called, so I think the number of internal count calls is not important at this point.

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerClassLoaderTest.java
```diff
@@ -1,5 +1,8 @@
 package com.vaadin.flow.server.startup;
 
+import static com.vaadin.flow.server.Constants.TARGET;
+import static com.vaadin.flow.server.frontend.FrontendUtils.DEFAULT_FLOW_RESOURCES_FOLDER;
```
mshabarov, line 4: `import static` methods should go after all 'normal' imports according to our code styles https://sites.google.com/a/vaadin.com/fw-flow-team-s-wiki/contributing-code/code-style/intellij-idea

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerClassLoaderTest.java
```diff
@@ -9,16 +12,15 @@
...
 
 import com.vaadin.flow.server.frontend.TestUtils;
 
-import static com.vaadin.flow.server.Constants.TARGET;
-import static com.vaadin.flow.server.frontend.FrontendUtils.DEFAULT_FLOW_RESOURCES_FOLDER;
-
 public class DevModeInitializerClassLoaderTest {
 
     @Test
+    @Ignore("mockito 3.0 does not use the instance classloader, probably because it uses bytebuddy, see https://github.com/mockito/mockito/issues/2304")
```
mshabarov, line 23: I believe this needs a ticket on flow side to modify and enable this test. Should I create it (if not exists yet) ?

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/VaadinConnectControllerTest.java
```diff
@@ -740,10 +742,17 @@ public void should_ReturnCorrectResponse_When_EndpointClassIsProxied() {
 
         ApplicationContext contextMock = mock(ApplicationContext.class);
         TestClass endpoint = new TestClass();
-        TestClass proxy = mock(TestClass.class, CALLS_REAL_METHODS);
-        when(contextMock.getBeansWithAnnotation(Endpoint.class)).thenReturn(
-                Collections.singletonMap(endpoint.getClass().getSimpleName(),
-                        proxy));
+
+        // CGLib proxies are supported as entry-point classes
+        Enhancer enhancer = new Enhancer();
```
mshabarov, line 747: Why is this needed?

flow-server/src/test/java/com/vaadin/flow/component/webcomponent/WebComponentTest.java
mshabarov: > mockito implenentation now does not differentiate the signature of the method but the name

Yeah, this makes sense. Thanks.

fusion-endpoint/src/test/java/com/vaadin/flow/server/connect/VaadinConnectControllerTest.java
manolo: Context: https://vaadin.slack.com/archives/CDNTYMS9H/p1621592967091900

Basically, fusion only supports cglib proxies, since mockito was using it in previous version the test simply worked, now mockito uses bytebuddy proxies, hence we need to create a cglib proxy by hand, and this is the code

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerClassLoaderTest.java
manolo: Please create it, so as we do not forget about this.

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerClassLoaderTest.java
manolo: Ups, my eclipse did it. Will revert the movement.

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerClassLoaderTest.java
mshabarov: Sure, will do.

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerClassLoaderTest.java
```diff
@@ -19,6 +20,7 @@
...
```
mshabarov, line 23:
```suggestion
    @Ignore("mockito 3.0 does not use the instance classloader, probably because it uses bytebuddy, see https://github.com/vaadin/flow/issues/11071, https://github.com/mockito/mockito/issues/2304")
```

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerClassLoaderTest.java
mshabarov: Added a link to corresponding issue.

mshabarov (approved):

mshabarov (approved):

ZheSun88:
FYI: as this has been reverted from 7.0 branch, i update the label manually.

PR #11039, opened 2021-05-20
fix: fix external url reference

### LINKED ISSUE #11015: 'css-loader' can't resolve the relative path in production mode since 14.6.0
'css-loader' can't resolve the relative path in production mode.

- My frontend folder looks like this:

```
-resources
  -frontend
    -css
        leaflet-map-styles.css
  -img
     map_layers_icon.png
```

- leaflet-map-styles.css

```css
[class="leaflet-control-layers-toggle"] {
	background-image: url('img/map_layers_icon.png') !important;
	background-size: 26px 26px;
}
```

- GoogleMap.java

```java
@CssImport(value = "./css/leaflet-map-styles.css", themeFor = "leaflet-map")
public class GoogleMap extends Div {...}
```

- addon: [Leaflet4Vaadin](https://vaadin.com/directory/component/leaflet4vaadin)

### Expected behavior
No problem when running/building image

### Actual behavior
```
ERROR in ../node_modules/@vaadin/flow-frontend/css/leaflet-map-styles.css
Module build failed (from ../node_modules/css-loader/dist/cjs.js):
Error: Can't resolve 'img/map_layers_icon.png' in 'D:\test\node_modules\@vaadin\flow-frontend\css'
    at D:\test\node_modules\enhanced-resolve\lib\Resolver.js:209:21
    at D:\test\node_modules\enhanced-resolve\lib\Resolver.js:285:5
    
    ...
    
@ ../target/frontend/generated-flow-imports-fallback.js 33:0-71 34:89-96
@ ../target/frontend/generated-flow-imports.js?babel-target=es6

...

[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  18:38 min
[INFO] Finished at: 2021-05-18T13:23:50+08:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.vaadin:vaadin-maven-plugin:14.6.1:build-frontend (default) on project test: Webpack process exited with non-zero exit code.
[ERROR] Stderr: ''
[ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.vaadin:vaadin-maven-plugin:14.6.1:build-frontend (default) on project test: Webpack process exited with non-zero exit code.
Stderr: ''
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    
    ...    
```

### Versions:
    - Vaadin / Flow version: 14.6.1 (brokenness starts with 14.6.0)
    - Java version: 11
    - OS version: Windows 10

caalador:
## Description

Fix the handing of external url reference in files that go into node_modules.
`flow-frontend` is copied directly to node_modules making css-loader
try to resolve urls for files that shouldn't be resolved.

Fixes #11015

## Type of change

- [x] Bugfix


flow-server/src/main/resources/webpack.generated.js
```diff
@@ -194,7 +194,8 @@ module.exports = {
             options: {
               url: (url, resourcePath) => {
                 // Only translate files from node_modules
-                const resolve = resourcePath.match(/(\\|\/)node_modules\1/);
+                const resolve = resourcePath.match(/(\\|\/)node_modules\1/)
+                  && fs.existsSync(path.resolve(path.dirname(resourcePath), url));
```
mshabarov, line 198: I don't get the meaning of this fix. The added line of code return `true` if the file exists, meaning that its path is an absolute path resolved from relative url and path to resource. Not digging deeply into what is resourcePath, I would want to understand why this condition is not attached to both existing conditions (either in node_modules, or in themes folder), but only to the first one, i.e. why isn't it like this: 
`return (resolve || themeResource) && fs.existsSync(path.resolve(path.dirname(resourcePath), url));` ?

mshabarov: Or maybe it goes like this because we are sure about anything within `themes` and we don't need to re-check its existence ?

caalador: The `theme-loader` already searches for file and checks their existence when it translates ./ and ../ urls to start with themes/... which is then seen here as themeResource that should be resolved. Also as the themeResource has been rewritten it would not be found with the relative path from the url here.

For node_modules we see if we can find the file against the css file the url is in (which is the resourcePath). In V14 local files are copied directly into node_modules and this creates the problem of trying to resolve a url that is not available with the path and if doing so will make css-loader throw a file not found.
By adding this check we do not try to resolve a url that is in a css file inside node_modules, but can not find the targeted file. another choice would be to ignore anything that is in `node_modules/@vaadin/flow-fronted`, but that might in theory skip something that should be then translated and copied.

mshabarov: Thanks for clarifications, Mikael! So the theme resources case is covered by `theme-loader` plugin. And for V14, due to files copy to node_modules, `css-loader` throws file not found and we skip that.

mshabarov (approved):

Issue #9378, opened 2020-11-11
Publish the @vaadin/form module to npm


vlukashov:
As a developer I want to install the `@vaadin/form` package from npm with `npm i @vaadin/form` so that I can use it in my projects without having the full Vaadin Fusion project setup.

Related discussions:
 - https://twitter.com/vberchet/status/1323651226228043778
   ![image](https://user-images.githubusercontent.com/22416150/98787512-46237800-2408-11eb-8a63-b61ad0733ac1.png)
 - https://polymer.slack.com/archives/C03PF4L4L/p1602844343473100
   <img width="382" alt="image" src="https://user-images.githubusercontent.com/22416150/98787543-50de0d00-2408-11eb-9875-f4a65575d7f9.png">

**Implementation notes:**
Publishing the Vaadin Fusion JS packages like `@vaadin/form` and `@vaadin/connect` to npm is complicated because we still want to keep the ability to create SNAPSHOT builds of Vaadin that use the locally-modified version of e.g. the `@vaadin/form` package (from this same repo, not from npm).
 - npm publishing could as well happen as a side step of the Vaadin release pipeline, without changing how the dependency to `@vaadin/form` is included into a Fusion project (which is _not_ via the npm registry). That would keep the SNAPSHOT builds working as before, but would lead to a situation where there would be different ways to install the `@vaadin/form` package, and that's a possible source of confusion and hard-to-debug issues.

vicb:
Thanks @vlukashov !

> because we still want to keep the ability to create SNAPSHOT

I do not exactly understand why the form package needs to be handled differently from other packages (i.e. the form web components) but may git subtrees could help here ?

One thing that would also be nice is to make sure that the package works server side so that node backend could validate the forms. The only limitation I have found so far is that some properties [are patched on the `window`](https://github.com/vaadin/flow/blob/775db42f470f41f1e84fdcfd6f7cbdd978ab6866/flow-client/src/main/resources/META-INF/resources/frontend/form/index.ts#L7). This is easily fixed by creating separate entry points for the browser or node in package.json - or by checking if windows is defined.

vlukashov:
> I do not exactly understand why the form package needs to be handled differently from other packages (i.e. the form web components) but may git subtrees could help here ?

[...]

PR #11042, opened 2021-05-20
Move TypeScript client code to separate npm packages 

### LINKED ISSUE #10898: Fusion: move Fusion client-side code to an npm package
Currently, the Fusion client-side code (`Connect.ts`, `Authentication.ts` and `form/`)is under the flow-client module, would be good to move them into an npm package e.g. `fusion-client`, like the [connect-client](https://www.npmjs.com/package/connect-client), so that a Flow user could exclude it from their application.

Note, both Fusion and Flow depend on the connection indicator, we need to separate that into a different npm package as well.

### LINKED ISSUE #10901: Fusion: move Fusion Flow shared client-side code into an npm package
The connection indicator and devmode gizmo are  used by both Fusion and Flow, there is a plan to [move fusion client-side code into an npm package](https://github.com/vaadin/flow/issues/10898), for that to work, we need to move the connection indicator (`ConnectionState.ts` and `ConnectionIndicator.ts`) which will be used by both Fusion and Flow.

Lodin:
Creates two npm packages that will be used as dependencies in both Flow and Fusion:

- `@vaadin/client-commons` - this package contains common client code shared by Flow and Fusion.
- `@vaadin/client-fusion` - this package contains Fusion-specific client code

Resolves #10901 
Resolves #10898


joheriks:
Two divergent versions of `ConnectionState.ts` and `ConnectionIndicator.ts`. E.g. the former:
https://github.com/vaadin/flow/blob/feat/fusion/npm-client/flow-client/src/main/frontend/ConnectionState.ts
https://github.com/vaadin/flow/blob/feat/fusion/npm-client/client/packages/client-common/src/ConnectionState.ts
Could this be somehow avoided?

Lodin:
@joheriks, yes, the first one should be removed eventually. But I guess I cannot do it right now before the packages are published to npm because it will break the flow-client. I think it should be the next step after publishing packages to npm where we remove the old versions and add the new ones from npm.

joheriks:
> @joheriks, yes, the first one should be removed eventually. But I guess I cannot do it right now before the packages are published to npm because it will break the flow-client. I think it should be the next step after publishing packages to npm where we remove the old versions and add the new ones from npm.

Right. Would be good to add a deprecation notice to the top of the `flow-client` files to avoid accidental divergence of the files.

vicb:
Sorry I have been lazy and didn't check all the changes in this PR.

My question is: would the form component be releases as a standalone package ?

I have discussed that with @vlukashov before (see https://github.com/vaadin/flow/issues/9378) and it would be awesome.

As of today I am using the forms in my client side app with my own custom node server on the backend. There are very few things that I had to change for this to work. The form package is great and I think would be very valuable for the JS/TS community in general.

Thanks

haijian-vaadin:
Hi @vicb, thanks a lot for the feedback, it's really great to hear that you like the form package.

For now, we don't plan to publish the form as a separate package yet. We could do it one day once we have a clear picture of what to do with it as a product.

ping our new PO @gilberto-torrezan.

haijian-vaadin (changes requested):
According to https://github.com/vaadin/fusion/issues/28, it needs sth more to manage the versions.

haijian-vaadin (dismissed):

haijian-vaadin:
> According to [vaadin/fusion#28](https://github.com/vaadin/fusion/issues/28), it needs sth more to manage the versions.

As agreed in the sprint planning session, we can do  [vaadin/fusion#28](https://github.com/vaadin/fusion/issues/28) in a separate PR. So removed my request change. Up to @platosha and @Haprog to check if the code in this PR is fine.

platosha (changes requested):

platosha (approved):

platosha (approved):

haijian-vaadin:
ping @vicb, @vaadin/form is finally published as a separate package at `https://www.npmjs.com/package/@vaadin/form`.

vicb:
Thanks a lot @haijian-vaadin !

I love this form framework, simple but feature packed and powerful.

Out of curiosity, is there a plan to decouple it from the vaadin backend ? (i.e. I had to remove a dependency on the DOM to be able to run the validators on a nodejs server)

PR #11065, opened 2021-05-25
fix: x64 version for mac if node old

### LINKED ISSUE #11060: Vaadin not working on apple silicon
Vaadin not working on apple silicon

gradle:vvadinPrepareFrontend - install node

https://nodejs.org/dist/v14.15.4/node-v14.15.4-darwin-arm64.tar.gz - file not foud

 - OS: Mac OS 11.3.1
 - arm64
- vaadinVersion=14.6.1

May be to temporarily install the x86 version?

no problem with rosette :)

caalador:
Use the x64 package for mac if the installed
node version is < 16.0.0 as ARM package
is not available before this.

Fixes #11060


pleku (approved):

pleku (approved):
Changes are only in "internal classes" used for devmode handler so this is good to go for maintenance releases.

Issue #11047, opened 2021-05-21
`StaticFileServer` dumps file listing when directory is accessed and application is running in (at least) Tomcat


fluorumlabs:
Steps to reproduce:

Development mode:
1. Create SVC project
2. Run it
3. Open `http://localhost:8080/frontend/.;` in browser
4. Observe list of files

Production mode:
1. Open `https://start.vaadin.com/frontend/.;` in browser
2. Observe list of files

miguelatvaadin:
Hi @fluorumlabs ,

coudl you please provide versions?

Thanks

miguelatvaadin:
ok. Thanks, @fluorumlabs

caalador:
I couldn't replicate this running `skeleton-starter-flow`, but it does happen with the spring-boot based starter `skeleton-starter-flow-spring`

fluorumlabs:
As far as I remember, `skeleton-starter-flow` use jetty

caalador:
Deploying the starter war to tomcat 9.0.39 I only get the result `Could not navigate to 'frontend/.;'`

fluorumlabs:
Same behavior can be observed with `/VAADIN/static/.;`

Not sure how `labs.vaadin.com` is configured, but probing two apps there works just "fine":
`https://labs.vaadin.com/edit-multiple/.;`, `https://labs.vaadin.com/theme-example/.;`

caalador:
That is the same it can only be done on a spring-boot project as far as I can find.
You can even start from `start.vaadin.com/.;` and continue from there with adding folders that you see.

fluorumlabs:
Seems it is Spring boot specific. However, my observations/debugging shown that the actual file list is served by StaticFileServer. In fact, this is the default behavior of FileInputStream when File points to a directory

caalador:
Would seem so. Getting a Spring project to break on break points seems like a game of luck at times.
In the spring-skeleton `request.getPathInfo()` returns `/.` for `/.;` and we only check for `endsWith("/")` to determine a directory

fluorumlabs:
`getPathInfo()` simply strips path parameters in Tomcat (the ones starting with `;`), I add `;` simply to prevent web browser from normalizing URL. The same effect can be achieved with `curl https://labs.vaadin.com/edit-multiple/. --path-as-is`

fluorumlabs:
Regarding checking if path points to a directory or not -- all these URLs point to the same directory:
[...]

PR #11072, opened 2021-05-25
fix: don't serve directories as static files

caalador:
Add check to the resource to see
if it is a directory and do not serve
if this is the case.

Fixe #11047


flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -107,6 +107,11 @@ public boolean isStaticResourceRequest(HttpServletRequest request) {
         }
         resource = getStaticResource(requestFilename);
 
+        if (resource != null && resource.getPath().endsWith("/")) {
```
denis-anisimov, line 110: Does `getPath()` always ends with "/" for a directory ?
In the test the URL may be `"new URL("file:///C:/frontend"` and not the `"new URL("file:///C:/frontend/"`.
The path is different but the semantic is the same : it's still a directory.

So the assumption looks quite fragile.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
caalador: the `getResource(path)` url returns always (as far as I've found out) directories ending with '/' it's the same expectation as we have for the request file name up top which also fails if the request is only `/frontend` also /frontend might be a file named only frontend and not a directory.

I couldn't find a better way to make certain it's actually a directory as most of the folders are inside jars  and you can't make a File for those to check isDirectory. One will get `java.nio.file.FileSystemNotFoundException` for example when using `Files.isDirectory(Paths.get(resource.toUri()))` if the file is inside a jar as we can't generate a Path for it.

caalador: Another way would be to have something like:
```
if (resource != null && resource.getProtocol().equals("jar")) {
            FileSystem fileSystem = null;
            try {
                URI resourceUri = resource.toURI();
                fileSystem = FileSystems
                    .newFileSystem(resourceUri, Collections.emptyMap());
                final Path path = fileSystem.getPath(resource.getPath()
                    .substring(resource.getPath().lastIndexOf("!") + 1));
                if (Files.isDirectory(path)) {
                    // Directory resources should not be served.
                    return false;
                }
            } catch (URISyntaxException | IOException e) {
                e.printStackTrace();
            } finally {
                if (fileSystem != null) {
                    try {
                        fileSystem.close();
                    } catch (IOException e) {
                        e.printStackTrace();
                    }
                }
            }
        }
```
But for the use it feels a bit heavy to have as it will be called quite often.

denis-anisimov: >getResource(path) 

I cannot find this method.

There is call `resource = getStaticResource(requestFilename);`.

The `getStaticResource` method :
- may be overriden anyhow
- its default impl delegates to `vaadinService.getStaticResource(path);` which also some internal API and may return anything


So this code protects the directory listing _only_ in some cases with some assumptions.

I'm not sure how the code should be written but the fix in the PR doesn't fix the issue generally for me.

denis-anisimov: The related question is: 
why the very first check 

```
        String requestFilename = getRequestFilename(request);
        if (requestFilename.endsWith("/")) {
            // Directories are not static resources although
            // servletContext.getResource will return a URL for them, at
            // least with Jetty
            return false;
        }
```
doesn't help here ?

There is some logic which make a redirect for some requested paths and the redirect goes with `/` in the end.
So  `getRequestFilename(request)` may be checked against `/` in the end (and this is also not a generic case).

Apparently `requestFilename` doesn't end with `/` in the considered case and in fact this is exactly the same situation with `getStaticResource(requestFilename)` : one more unreliable check.

caalador: The request goes to the servlerContext.getResource(path) returning the URL for the resource if found.
The first check actually already fails for the simple request `localhost:8080/frontend` which will not be seen as a directory, but will when gotten from the servlet context be found as a directory and be served as a directory listing.

caalador: But the first check is that if the request ends with `/` then it most certainly is not a static resource that we should serve.

denis-anisimov: >The request goes to the servlerContext.getResource(path)

No it doesn't.

It goes via 
- `StaticFileServer::getStaticResource` 
- `VaadinService::getStaticResource`
- and only as an impl detail of `VaadinServletService` it goes to `getServlet().getServletContext().getResource`.

so this fix assumes two things:
- `StaticFileServer::getStaticResource`  is not reimplemented or its implementation behaves in the certain way.
- `VaadinService::getStaticResource`  is in fact `VaadinServletService::getStaticResource`  which is not reimplemented or if it's reimplemented then it behaves in a certain way.


I think it's just incorrect to solve this issue in the `StaticFileSever` at all. At least in the `isStaticResourceRequest`.

It should be considered as a static resource request but it should not render the directory.
The logic which needs to be changed is rendering the directory listing: not sure which place it is.
`ResponseWriter` may be.

caalador: I don't see how the ResponseWriter could differentiate between a directory listing and a normal file content.
Also one point with not handling a directory request as a StaticFile request is that you can't go around guessing what folders may exist in a project.

denis-anisimov: There should be some logic which shows the file content.

Technically the directory is also a file in UNIX but I think the directory is handled via a specific way.
The logic is most likely checks whether this is a regular file or not and if it's a directory then it _renders_ a listing (like directory is listed by pure Web server).
So the result of the rendering is a specific HTML which contains elements like `<li>` or whatever for the files inside the directory.
The real directory on the FS doesn't contain any HTML tags of course. So there should be something which decides how to show the URL : as a listing of files or the content of the file.

I don't know how exactly the list of files is shown since the original ticket doesn't say that. But I assume this is a listing done using HTML markup. Then there is some code which does this.
This code should not do this for the directory and that's it.
This is theoretically. It might be not that easy in practice.

But the changes in the current PR doesn't protected from the dir listing generally. This can be used only as a lightweight preliminary check : if this true then deny. But if it's not true (doesn't end with slash) still try to check whether this is a directory or not.

caalador: I updated it to check for actual directory, with test for dir in file system and dir inside a zip.

denis-anisimov: OK, I've checked how the response is written and apparently it's delegated `URLConnection` subclasses like that
`DirectoryURLConnection`.
So there is no code which controls the response :(

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -107,6 +115,33 @@ public boolean isStaticResourceRequest(HttpServletRequest request) {
         }
         resource = getStaticResource(requestFilename);
 
+        if (resource != null) {
+            try {
+                URI resourceURI = resource.toURI();
```
denis-anisimov, line 120: This can be extracted to a several methods to avoid having everything here ilined.
The current code contains too many nested `if-try-else-if`.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
caalador: Cleaned up code and fixed the test.

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -116,6 +128,35 @@ public boolean isStaticResourceRequest(HttpServletRequest request) {
         return resource != null;
     }
 
+    private boolean resourceIsDirectory(URL resource) {
+        URI resourceURI = null;
```
denis-anisimov, line 132: I think the original check (against whether the path ends with `/`) can be the done in the very beginning: it's lightweight and in most cases it will be enough to identify the directoty.
But if it doesn't end with `/` then this logic should be applied.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
caalador: Added simple check and updated test to go all paths.

denis-anisimov (approved):

PR #11075, opened 2021-05-26
fix: avoid NPE in Location parsing

### LINKED ISSUE #11019: NPE in QueryParameters.parseQueryString if location contains colon
Steps to reproduce:
1. Generate project with Flow views using SVC
2. Start project
3. Navigate to `http://localhost:8080/a:`
4. Observe NPE in server logs:

```java
java.lang.NullPointerException: Cannot invoke "String.split(String)" because "query" is null
	at com.vaadin.flow.router.QueryParameters.parseQueryString(QueryParameters.java:119) ~[flow-server-7.0.0.beta3.jar:7.0.0.beta3]
	at com.vaadin.flow.router.QueryParameters.fromString(QueryParameters.java:114) ~[flow-server-7.0.0.beta3.jar:7.0.0.beta3]
	at com.vaadin.flow.router.Location.parseParams(Location.java:247) ~[flow-server-7.0.0.beta3.jar:7.0.0.beta3]
	at com.vaadin.flow.router.Location.<init>(Location.java:60) ~[flow-server-7.0.0.beta3.jar:7.0.0.beta3]
	at com.vaadin.flow.server.communication.JavaScriptBootstrapHandler$JavaScriptBootstrapContext.initRoute(JavaScriptBootstrapHandler.java:90) ~[flow-server-7.0.0.beta3.jar:7.0.0.beta3]
	at com.vaadin.flow.server.BootstrapHandler$BootstrapContext.<init>(BootstrapHandler.java:224) ~[flow-server-7.0.0.beta3.jar:7.0.0.beta3]
	at com.vaadin.flow.server.communication.JavaScriptBootstrapHandler$JavaScriptBootstrapContext.<init>(JavaScriptBootstrapHandler.java:83) ~[flow-server-7.0.0.beta3.jar:7.0.0.beta3]
	at com.vaadin.flow.server.communication.JavaScriptBootstrapHandler.createBootstrapContext(JavaScriptBootstrapHandler.java:160) ~[flow-server-7.0.0.beta3.jar:7.0.0.beta3]
	at com.vaadin.flow.server.BootstrapHandler.createAndInitUI(BootstrapHandler.java:1309) ~[flow-server-7.0.0.beta3.jar:7.0.0.beta3]
```

denis-anisimov:
fixes #11019

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/Location.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/Location.java
@@ -246,7 +246,7 @@ public class Location implements Serializable {
         }
 
         if (query == null) {
-            // decoding of parameters is done in QueryParameters::parseParams
+            // decoding of parameters is done in QueryParameters
             query = path.substring(beginIndex + 1);
         }
 
--- a/flow-server/src/main/java/com/vaadin/flow/router/Location.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/Location.java
@@ -15,8 +15,6 @@
  */
 package com.vaadin.flow.router;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
-
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
@@ -28,6 +26,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Represents a relative URL made up of path segments and query parameters, but
@@ -244,6 +244,9 @@ public class Location implements Serializable {
             query = path.substring(beginIndex + 1);
         }
 
+        if (query == null) {
+            query = path.substring(beginIndex + 1);
+        }
         return QueryParameters.fromString(query);
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/router/Location.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/Location.java
@@ -238,15 +238,17 @@ public class Location implements Serializable {
             return QueryParameters.empty();
         }
         String query;
+
         try {
             query = new java.net.URI(path).getQuery();
         } catch (URISyntaxException ignore) { // NOSONAR
-            query = path.substring(beginIndex + 1);
+            query = null;
         }
 
         if (query == null) {
             query = path.substring(beginIndex + 1);
         }
+
         return QueryParameters.fromString(query);
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/router/Location.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/Location.java
@@ -246,6 +246,7 @@ public class Location implements Serializable {
         }
 
         if (query == null) {
+            // decoding of parameters is done in QueryParameters::parseParams
             query = path.substring(beginIndex + 1);
         }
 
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/Location.java
```diff
@@ -244,6 +244,9 @@ private static QueryParameters parseParams(String path) {
             query = path.substring(beginIndex + 1);
         }
 
+        if (query == null) {
+            query = path.substring(beginIndex + 1);
```
pleku, line 248: Wouldn't it be just the same to merge this to the case from above when the error has been thrown ? Now the same line of code is there twice, if the `URI` constructor threw and exception or `null` was thrown by `URI::getQuery`. 

I think it would make it the code more cleaner.

flow-server/src/main/java/com/vaadin/flow/router/Location.java
```diff
@@ -244,6 +244,9 @@ private static QueryParameters parseParams(String path) {
             query = path.substring(beginIndex + 1);
```
pleku, line 244: I have an issue with this code (that you also added below) - it doesn't escape any special characters like the `getQuery()` decodes - so you can pass in malicious query parameter ?

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/Location.java
denis-anisimov: I cannot answer to this question.
I don't understand the code either.
But this is a fix which just avoid NPE #11019 and nothing else.
I copied the fallback here as is since there was already such code.

That's the reason why fixes should be done by someone who has written this initially or the one who is responsible for this code.

pleku: Actually answering this myself - the code in  `QueryParameters` does the decoding of the parameters later on - maybe it would be enough to just add a comment there.

flow-server/src/main/java/com/vaadin/flow/router/Location.java
denis-anisimov: The question is _how_ ?
I wouldn't say it become better with the latest change.

flow-server/src/main/java/com/vaadin/flow/router/Location.java
pleku: > That's the reason why fixes should be done by someone who has written this initially or the one who is responsible for this code.

Yes it would be better so - but that is just an excuse. The job is not just about writing new code, but also about maintaining it. In this case it is now fine, but we can just make it a bit more maintainable so that others don't run into the same question without knowing the answer before going through a lot of code elsewhere.

flow-server/src/main/java/com/vaadin/flow/router/Location.java
```diff
@@ -238,9 +238,14 @@ private static QueryParameters parseParams(String path) {
...
         String query;
+
         try {
             query = new java.net.URI(path).getQuery();
         } catch (URISyntaxException ignore) { // NOSONAR
+            query = null;
+        }
+
+        if (query == null) {
             query = path.substring(beginIndex + 1);
```
pleku, line 249:
```suggestion
            // decoding of parameters is done in QueryParameters::parseParams
            query = path.substring(beginIndex + 1);
```

pleku (approved):

PR #11078, opened 2021-05-26
Reapply "test: fix mockito errors accessing protected API in JDK16 (#11034) (CP: 7.0)"

manolo:


mshabarov:
@manolo are these changes same as before, but now we want to reapply it because the necessary changes on the components side are done?

mshabarov (approved):

PR #11086, opened 2021-05-26
test: fix mockito errors accessing protected API in JDK16 (#11034) (CP: 6.0)

manolo:


flow-client/src/main/frontend/form/index.ts
```diff
@@ -9,5 +9,5 @@ $wnd.Vaadin = $wnd.Vaadin || {};
 $wnd.Vaadin.registrations = $wnd.Vaadin.registrations || [];
 $wnd.Vaadin.registrations.push({
   is: '@vaadin/form',
-  version: /* updated-by-script */ '7.0-SNAPSHOT'
+  version: /* updated-by-script */ '6.0-SNAPSHOT'
```
pleku, line 12: I don't get how is this change related ?

pleku (changes requested):

pleku:
ping @manolo

flow-client/src/main/frontend/form/index.ts
manolo: I reverted this, but still shouldn't that version say 6.0 since it's flow 6.0 branch ?

pleku (approved):

PR #11089, opened 2021-05-27
fix: Annotate default error handlers. (#11063)

caalador:
Annotate default error handlers with
the new annotaton DefaultErrorHandler
so that we can also in add-ons have default
handlers that can be overridden by user
handlers.

Part of vaadin/spring#661


pleku (approved):

pleku:
I don't know why APIDiff has not been run at all

PR #11091, opened 2021-05-27
chore: add bnd.bnd for OSGi

joheriks:
Related to vaadin/flow#10892


denis-anisimov:
Dev server doesn't work in OSGi anyhow.
So I don't think it should declare itself as an OSGi bundle: it lies that it's semantically OSGi compatible.

joheriks:
> Dev server doesn't work in OSGi anyhow.
> So I don't think it should declare itself as an OSGi bundle: it lies that it's semantically OSGi compatible.

Alright.

PR #11096, opened 2021-05-27
Use simple space removing for CI

Lodin:
Because of recent CI failures that are not reproducible on my Windows machine, let's run a previous version of code comparison where all spaces are simply removed. It will work only for CI, and the developer will get the human-readable output instead.


haijian-vaadin (approved):

haijian-vaadin:
Fixes https://github.com/vaadin/fusion/issues/25

PR #11097, opened 2021-05-27
feat: allow defining access control annotations on view super classes

### LINKED ISSUE #10705: Allow defining endpoint access control annotations on super class

#10609 suggests using the same access control annotations used today for Fusion endpoints also for Flow views however to be useful it should check super classes also (e.g. `AbstractPublicView`)

Suggestion:
Also check super classes for endpoints if the method and endpoint class does not have any access control annotations.

By unifying these checks we have an easier to understand security model for Vaadin, both for views and endpoints.

NOTE: The suggestion is not to support access control annotations on the super class version of a method. E.g. the following should log a warning and interpret `SubClass.hello` as `@DenyAll`:

```java
public class SuperClass {

    @AnonymousAllowed
    public String hello() {
        return "hello";
    }
}
```

```java
public class SubClass extends SuperClass {
  public String hello() {
     return "goodbye";
  }
}
```

taefi:
## Description

This allows leaving child views unannotated 
(without any of the access annotations) and the 
access annotations to be defined on any of the 
view's superclasses. 

Fixes #10705

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
@@ -210,10 +210,12 @@ public class AccessAnnotationChecker implements Serializable {
      *         returned.
      *         <p>
      *         Access annotations that being checked are:
-     *         - <code>@AnonymousAllowed</code>
-     *         - <code>@PermitAll</code>
-     *         - <code>@RolesAllowed</code>
-     *         - <code>DenyAll</code>
+     *         <ul>
+     *         <li><code>@AnonymousAllowed</code>
+     *         <li><code>@PermitAll</code>
+     *         <li><code>@RolesAllowed</code>
+     *         <li><code>DenyAll</code>
+     *         </ul>
      *
      * @throws NullPointerException
      *             if the input <code>cls</code> is null
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
@@ -201,8 +201,8 @@ public class AccessAnnotationChecker implements Serializable {
      * @param cls
      *            the class to check
      * @return the first annotated class in {@code cls}'s hierarchy that
-     *         annotated with one of the access annotations, starting from
-     *         the input {@code cls} class itself, going up in the hierarchy.
+     *         annotated with one of the access annotations, starting from the
+     *         input {@code cls} class itself, going up in the hierarchy.
      *         WARNING: Note that interfaces are ignored.
      *         <p>
      *         If no class in the hierarchy was annotated with any of the access
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2000-2020 Vaadin Ltd.
+ * Copyright 2000-2021 Vaadin Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -20,6 +20,7 @@ import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.security.Principal;
+import java.util.Objects;
 import java.util.function.Function;
 
 import javax.annotation.security.DenyAll;
@@ -199,12 +200,34 @@ public class AccessAnnotationChecker implements Serializable {
      *
      * @param cls
      *            the class to check
-     * @return the entity that is responsible for security settings for the
-     *         method passed
-     * @throws IllegalArgumentException
-     *             if the method is not public
+     * @return the first annotated class in <code>cls</code>'s hierarchy with
+     *          that annotated with one of the access annotations,
+     *          starting from the input <code>cls</code> class itself, going up
+     *          in the hierarchy. Note that interfaces are ignored.
+     *          <p>
+     *          If no class in the hierarchy was annotated with any of the access
+     *          annotations, the <code>cls</code> input parameter itself would be
+     *          returned.
+     *          <p>
+     *          Access annotations that being checked against are:
+     *          <code>@AnonymousAllowed</code>
+     *          <code>@PermitAll</code>,
+     *          <code>@RolesAllowed</code>
+     *          <code>DenyAll</code>
+     *
+     * @throws NullPointerException
+     *             if the input <code>cls</code> is null
      */
     public AnnotatedElement getSecurityTarget(Class<?> cls) {
+        Objects.requireNonNull(cls, "The input Class must not be null.");
+
+        Class<?> clazz = cls;
+        while (clazz != null && clazz != Object.class) {
+            if (hasSecurityAnnotation(clazz)) {
+                return clazz;
+            }
+            clazz = clazz.getSuperclass();
+        }
         return cls;
     }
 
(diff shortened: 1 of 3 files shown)


Artur-:
I think this changes how endpoint security checks work in an undesired way. It should be scoped to only change view aka class based access control

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
```diff
@@ -199,12 +200,36 @@ public AnnotatedElement getSecurityTarget(Method method) {
...
+     *         Access annotations that being checked are:
+     *         <ul>
+     *         <li><code>@AnonymousAllowed</code>
+     *         <li><code>@PermitAll</code>
+     *         <li><code>@RolesAllowed</code>
+     *         <li><code>DenyAll</code>
+     *         </ul>
+     *
+     * @throws NullPointerException
+     *             if the input <code>cls</code> is null
```
taefi, line 221:
```suggestion
     * @return the first annotated class in {@code cls}'s hierarchy that
     *         annotated with one of the access annotations, starting from
     *         the input {@code cls} class itself, going up in the hierarchy.
     *         WARNING: Note that interfaces are ignored.
     *         <p>
     *         If no class in the hierarchy was annotated with any of the access
     *         annotations, the {@code cls} input parameter itself would be
     *         returned.
     *         <p>
     *         Access annotations that being checked are:
     *         <ul>
     *         <li>{@code @AnonymousAllowed}
     *         <li>{@code @PermitAll}
     *         <li>{@code @RolesAllowed}
     *         <li>{@code @DenyAll}
     *         </ul>
     *
     * @throws NullPointerException
     *             if the input {@code cls} is null
```

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
```diff
@@ -199,12 +200,36 @@ public AnnotatedElement getSecurityTarget(Method method) {
...
      */
     public AnnotatedElement getSecurityTarget(Class<?> cls) {
+        Objects.requireNonNull(cls, "The input Class must not be null.");
+
+        Class<?> clazz = cls;
+        while (clazz != null && clazz != Object.class) {
+            if (hasSecurityAnnotation(clazz)) {
+                return clazz;
+            }
+            clazz = clazz.getSuperclass();
```
haijian-vaadin, line 231: For the case that a view doesn't have any access control annotations, and also non of the ancestors have any annotations, should we return the view class itself instead of the Object? The end result is probably the same but feels more right that way.

haijian-vaadin (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
mshabarov: It returns the view class itself currently,  doesn't it?
Consider class hierarchy `Object <- A <- B`, where `B` is an input and no-one is annotated. Then `clazz` finally would be an `Object` and method returns `cls` which is a view class `B`.

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
```diff
@@ -199,12 +200,36 @@ public AnnotatedElement getSecurityTarget(Method method) {
...
      * @param cls
      *            the class to check
-     * @return the entity that is responsible for security settings for the
-     *         method passed
-     * @throws IllegalArgumentException
-     *             if the method is not public
+     * @return the first annotated class in {@code cls}'s hierarchy that
+     *         annotated with one of the access annotations, starting from
+     *         the input {@code cls} class itself, going up in the hierarchy.
+     *         WARNING: Note that interfaces are ignored.
```
mshabarov, line 206:
```suggestion
     *         <em>Note:</em> interfaces in the {@code cls}'s hierarchy are ignored.
```

flow-server/src/test/java/com/vaadin/flow/server/auth/AccessControlTestClasses.java
```diff
@@ -270,4 +270,68 @@ public void rolesAllowedUserAdmin() {
...
+    public static class DenyAllGrandParentView {
+    }
+
+    public static class DenyAllParentView extends DenyAllGrandParentView {
+    }
+
+    @Route("no-annotation-denyall-by-grandparent")
+    public static class NoAnnotationDenyAllByGrandParentView
+            extends DenyAllParentView {
+    }
```
mshabarov, line 336: I would also add a test case when a tested class is an interface or it's parent is an interface:
```java
@AnonymousAllowed
public static interface CustomComponent extends HasElement {
}

public static class CustomComponentImpl implements CustomComponent {
}
```
This is not maybe a real case, but since the API is public and javadocs warns about interfaces skipping, good to tests it either.

mshabarov:
> I think this changes how endpoint security checks work in an undesired way. It should be scoped to only change view aka class based access control

@Artur- could you please elaborate more what do you mean, or point me to the docs/classes/ticket where the behaviour is being changed by this implementation?

flow-server/src/test/java/com/vaadin/flow/server/auth/AccessControlTestClasses.java
taefi: Done.

mshabarov (approved):

taefi:
> I think this changes how endpoint security checks work in an undesired way. It should be scoped to only change view aka class-based access control

I should argue a little to deepen my understanding - or misunderstanding ;)

If I understood @Artur-'s discussion about this change correctly, he is concerned about the behavior change of the following method of `AccessAnnotationChecker`:
```Java
public AnnotatedElement getSecurityTarget(Class<?> cls){ ... }
```
Before changes in this PR (as it is in master), the security target for endpoints was working as follows:
- If the endpoint's methods weren't annotated with any of the access annotations, the security target has been falling back to Endpoint class, and the above method was resolving it to the class itself, so if the Endpoint class also wasn't annotated with any access annotation, The `DenyAll` logic was applying. 

Now (after this change), when the endpoint security target for unannotated methods falls back to the Endpoint class, it would go up in the hierarchy and ... (rest of the story would be the same as flow views). 

I'm not sure why is this behavior change should be "undesired", and even I'm in favor of this change as an enhancement for the fusion endpoint security, but **it is a breaking change** anyway.

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
haijian-vaadin: right, my bad, misread that it returns clazz at the end.

haijian-vaadin:
@taefi @mshabarov The discussion is here in the fusion channel https://vaadin.slack.com/archives/CDNTYMS9H/p1622458536274100

I think this change looks good, but we need to how this security checks work for both Fusion and Flow and make them aligned, also to be aligned how it works with Spring Security.

Artur-:
I have not tested how it affects endpoints but as the same access checker is used for views and endpoints, it is likely that it changes how endpoint access checks work.

Endpoint access checks need to be in sync with how the annotation checking works in e.g. Spring Boot, which should follow JSR 250. Otherwise you might end up in a situation where you are using the same annotation in different parts of your application, say `@Endpoint` and `@Service` classes, but the meaning of the annotations are different depending on the case. We do not want that.

If I interpreted JSR 250 correctly it seems to say:
1. You can always figure out the annotation based on looking at only one file
2. The file to use is the one where the method is declared (or overridden)
3. Method annotation always override class annotation
4. You never look at any super (overridden) method or super class
5. You never look at interfaces

If this PR changes it so that 4. is no longer true because the super class of an endpoint is checked then you have a mismatch in how Vaadin and Spring decides if access is allowed or denied. This would make it hard to understand if the application is properly secured or not.

Artur-:
JSR 250 only considers method level security with the class being the place where you can do a fallback configuration. For view based access control we need class based security with another, more generic place as the fallback. For this, the only logical place is a super class.

haijian-vaadin (approved):

PR #11099, opened 2021-05-27
fix: prevent passing bad character to dev-server

pleku:
The webpack dev-server does not escape " character, as it is not valid
URL. This limitation was not checked when passing request to it via
DevModeHandlerImpl.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.server;
 
+import javax.servlet.http.HttpServletRequest;
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
@@ -27,8 +28,6 @@ import java.util.Optional;
 import java.util.function.BiConsumer;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-
-import javax.servlet.http.HttpServletRequest;
 
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.server.communication.PwaHandler;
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -409,7 +412,8 @@ public static boolean isPathUnsafe(String path) {
             throw new RuntimeException("An error occurred during decoding URL.",
                     e);
         }
-        return PARENT_DIRECTORY_REGEX.matcher(path).find();
+        return PARENT_DIRECTORY_REGEX.matcher(path).find()
+                || ILLEGAL_CHARACTER_REGEX.matcher(path).find();
```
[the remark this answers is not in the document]
pleku: Moved to devmodehandler as the check is just related to webpack devserver and what can be passed to it

caalador (commented):
just needs running `formatter:format`

pleku:
Formatter run and hopefully ccdm tests pass this time too

caalador (approved):

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerImpl.java
```diff
@@ -393,6 +398,7 @@ private boolean checkWebpackConnection() {
     @Override
     public HttpURLConnection prepareConnection(String path, String method)
             throws IOException {
+        // path should have been checked at this point for any outside requests
         URL uri = new URL(WEBPACK_HOST + ":" + getPort() + path);
```
fluorumlabs, line 402: According to JavaDoc, the recommended way of dealing with encoding is to use `new URI(...).toURL()`:

> Note, the `java.net.URI` class does perform escaping of its
component fields in certain circumstances. The recommended way
to manage the encoding and decoding of URLs is to use `java.net.URI`,
and to convert between these two classes using `toURI()` and
`toURL()`.
```

fluorumlabs (commented):
Denylisting is generally a bad approach for dealing with incorrect inputs

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerImpl.java
pleku: It would make sense, yes. But the code needs more cleanup to be able to do this properly.

PR #11102, opened 2021-05-28
chore: warn about missing DevModeHandlerManager

joheriks:
Gives a clue in case the vaadin-dev-server module is accidentally excluded in dev mode.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
@@ -32,6 +32,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.function.DeploymentConfiguration;
+import com.vaadin.flow.internal.DevModeHandler;
 import com.vaadin.flow.internal.DevModeHandlerManager;
 import com.vaadin.flow.internal.UsageStatistics;
 import com.vaadin.flow.server.communication.FaviconHandler;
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
```diff
@@ -98,8 +99,17 @@ protected VaadinServletService() {
...
-                    .ifPresent(handlers::add);
+            Optional<DevModeHandler> handlerManager = DevModeHandlerManager
+                    .getDevModeHandler(this);
+            if (handlerManager.isPresent()) {
+                handlers.add(handlerManager.get());
+            } else {
+                getLogger()
+                        .warn("no DevModeHandlerManager implementation found "
+                                + "but dev server enabled. Include the "
+                                + "com.vaadin.vaadin-dev-server maven "
```
knoobie, line 110:
```suggestion
                                + "com.vaadin:vaadin-dev-server:${vaadin.version} "
```

I think maven is redundant in this case, as gradle has this dependency as well. The syntax for the property was adopted from maven and matches the current property defined by start.vaadin.com.

joheriks: Reworded.

joheriks: (Skipped the version info because it is different from platform and bom-managed)

taefi (approved):

PR #11107, opened 2021-05-28
fix: set routes first into the template

denis-anisimov:
fixes psi#26

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
@@ -60,11 +60,11 @@ public class RouteNotFoundError extends Component
                 .isProductionMode();
 
         String template = getErrorHtml(productionMode);
-        template = template.replace("{{path}}", path);
-        template = template.replace("{{additionalInfo}}", additionalInfo);
         if (template.contains("{{routes}}")) {
             template = template.replace("{{routes}}", getRoutes(event));
         }
+        template = template.replace("{{additionalInfo}}", additionalInfo);
+        template = template.replace("{{path}}", path);
 
         getElement().appendChild(new Html(template).getElement());
         return HttpServletResponse.SC_NOT_FOUND;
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
```diff
@@ -60,11 +60,11 @@ public int setErrorParameter(BeforeEnterEvent event,
                 .isProductionMode();
 
         String template = getErrorHtml(productionMode);
-        template = template.replace("{{path}}", path);
-        template = template.replace("{{additionalInfo}}", additionalInfo);
         if (template.contains("{{routes}}")) {
             template = template.replace("{{routes}}", getRoutes(event));
         }
+        template = template.replace("{{additionalInfo}}", additionalInfo);
```
fluorumlabs, line 66: Please add comment emphasizing that the order of replacements is important (for the future you), similar to https://github.com/vaadin/flow/blob/30886fd96c0cfdb8cc7d8546972666528126fddb/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java#L1148-L1150

fluorumlabs (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
denis-anisimov: I've added the comment but this is extra.
There is a test which tests that it works as expected.
Comment doesn't force anything: too easy to ignore it.

fluorumlabs: The test is not ideal as it only tests injection via `{{path}}` (not `{{additionalInfo}}` which is also present in production template). And obviously it won't catch the cases when, at some point, a new placeholder will be added.

So it's better to have the comment. While it won't prevent issues, it will at least lower the chance.

mshabarov (approved):

mshabarov:
@fluorumlabs could you please review this PR? Merge is blocked since you requested changes. Thanks!

fluorumlabs (approved):

PR #11114, opened 2021-05-31
fix: prevent passing bad character to wp (#11099)

pleku:
The webpack dev-server does not escape " character, as it is not valid
URL. This limitation was not checked when passing request to it via
DevModeHandlerImpl.

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
@@ -26,7 +26,6 @@ import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.net.ConnectException;
-import java.net.HttpURLConnection;
 import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


caalador (dismissed):

flow-server/src/test/java/com/vaadin/flow/server/DevModeHandlerTest.java
```diff
@@ -21,12 +21,12 @@
 import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.net.ConnectException;
+import java.net.HttpURLConnection;
```
caalador, line 29: Should this be added here as we have static imports for the expected codes HTTP_OK and HTTP_FORBIDDEN?

pleku: Should not, but honestly - this is a test class, should make no difference and there are more important things to do

caalador (approved):

PR #11118, opened 2021-05-31
fix: make primitive types non-nullable by default in generated TS file

Lodin:
Fixes vaadin/fusion#23.

### DIFF
--- a/fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
+++ b/fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
@@ -41,35 +41,13 @@ public class ExplicitNullableTypeChecker {
      * @return a result of check
      */
     public static boolean isRequired(AnnotatedElement element) {
-        return Stream.of(element.getAnnotations())
-                .anyMatch(annotation -> "nonnull".equalsIgnoreCase(
-                        annotation.annotationType().getSimpleName()));
-    }
-
-    /**
-     * Checks if the reflected element should be required (not nullable) in the
-     * generated Typescript code.
-     *
-     * @param element
-     *            an element to be required
-     * @return a result of check
-     */
-    public static boolean isRequired(Field element) {
-        return element.getType().isPrimitive()
-                || isRequired((AnnotatedElement) element);
-    }
-
-    /**
-     * Checks if the reflected element should be required (not nullable) in the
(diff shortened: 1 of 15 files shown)


fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
```diff
@@ -43,16 +47,48 @@ public static boolean isRequired(AnnotatedElement element) {
...
+     * Checks if the reflected element should be required (not nullable) in the
+     * generated Typescript code.
+     *
+     * @param element
+     *            an element to be required
+     * @return a result of check
+     */
+    public static boolean isRequired(Field element) {
+        return element.getType().isPrimitive()
+                || isRequired((AnnotatedElement) element);
```
haijian-vaadin, line 59: No need for this cast, since `Field` implements the `AnnotatedElement` interface

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
```diff
@@ -43,16 +47,48 @@ public static boolean isRequired(AnnotatedElement element) {
...
+     * generated Typescript code.
+     *
+     * @param element
+     *            an element to be required
+     * @return a result of check
+     */
+    public static boolean isRequired(Parameter element) {
+        return element.getType().isPrimitive()
+                || isRequired((AnnotatedElement) element);
+    }
```
haijian-vaadin, line 73: No need for this cast, since `Parameter` implements the `AnnotatedElement` interface

haijian-vaadin (changes requested):

fusion-endpoint/src/main/java/com/vaadin/flow/server/connect/ExplicitNullableTypeChecker.java
haijian-vaadin: Also, it would run into an endless loop here?

Lodin: Won't the `isRequired` method we are currently within will be called instead leading to a endless loop?

haijian-vaadin (approved):

haijian-vaadin (approved):

PR #11124, opened 2021-06-01
fix: make generated TS array be able to have undefined values

Lodin:
Fixes vaadin/fusion#24.


Haprog (approved):

Haprog (changes requested):
Should make sure that arrays of primitive types (from Java) won't get `undefined` inside the array in the TS type, as we discussed in the daily.

I guess these:

```java
int[] array1;
List<int> array2;
```

should result in something like this?:

```typescript
let array1: Array<number>;
let array2: Array<number> | undefined;
```

Lodin:
I suppose we can ignore @vaadin-bot here because all the issues are not caused by my changes

platosha (approved):
Great job here! 👍

platosha (approved):

Haprog (approved):

PR #11133, opened 2021-06-01
chore: add dummy class to get javadocs generated

pleku:


pleku:
@caalador I don't know if we should merge this or test the dokka plugin if it works on top of master branch ?

manolo:
JVM crashed, re-running

manolo (approved):

caalador:
If you want to test there is the branch `gradle/javaDoc` with dokka for master.
Else as the JavaDoc is not used I would be fine with this also.

pleku:
This caused other issues, and it seems that the dokka plugin started to work today so we're retrying in 7.0 with that https://bender.vaadin.com/viewLog.html?buildId=234529&buildTypeId=Flow_Flow70_Flow70Snapshot

PR #11147, opened 2021-06-03
fix: Handle concurrent file requests

### LINKED ISSUE #11144: java.nio.file.FileSystemAlreadyExistsException: null with latest snapshot
This is from the build log of a Spring snapshot
```
[16:23:06] :	 [Step 3/5] java.nio.file.FileSystemAlreadyExistsException: null
[16:23:06] :	 [Step 3/5] 	at com.sun.nio.zipfs.ZipFileSystemProvider.newFileSystem(ZipFileSystemProvider.java:113) ~[zipfs.jar:1.8.0_201]
[16:23:06] :	 [Step 3/5] 	at java.nio.file.FileSystems.newFileSystem(FileSystems.java:326) ~[na:1.8.0_201]
[16:23:06] :	 [Step 3/5] 	at java.nio.file.FileSystems.newFileSystem(FileSystems.java:276) ~[na:1.8.0_201]
[16:23:06] :	 [Step 3/5] 	at com.vaadin.flow.server.StaticFileServer.resourceIsDirectory(StaticFileServer.java:146) ~[flow-server-8.0-SNAPSHOT.jar:8.0-SNAPSHOT]
[16:23:06] :	 [Step 3/5] 	at com.vaadin.flow.server.StaticFileServer.isStaticResourceRequest(StaticFileServer.java:118) ~[flow-server-8.0-SNAPSHOT.jar:8.0-SNAPSHOT]
[16:23:06] :	 [Step 3/5] 	at com.vaadin.flow.server.VaadinServlet.serveStaticOrWebJarRequest(VaadinServlet.java:343) ~[flow-server-8.0-SNAPSHOT.jar:8.0-SNAPSHOT]
[16:23:06] :	 [Step 3/5] 	at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:289) ~[flow-server-8.0-SNAPSHOT.jar:8.0-SNAPSHOT]
[16:23:06] :	 [Step 3/5] 	at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:109) ~[vaadin-spring-18.0-SNAPSHOT.jar:na]
[16:23:06] :	 [Step 3/5] 	at javax.servlet.http.HttpServlet.service(HttpServlet.java:733) [tomcat-embed-core-9.0.45.jar:4.0.FR]
```

Seems like Flow is trying to create multiple file system instances for the same zip file which based on this code can never work: https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/jdk.zipfs/share/classes/jdk/nio/zipfs/ZipFileSystemProvider.java#L90-L121

Potentially this happens when you make two requests at the same time for the same static resource

caalador:
Fixes issue with multiple concurrent
requests to jar resource.

Fixes #11144

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
@@ -78,8 +78,8 @@ public class StaticFileServer implements StaticFileHandler {
             .compile("^\\/themes\\/[\\s\\S]+?\\/");
 
     // Mapped uri is for the jar file
-    Map<URI, Integer> openFileSystems = new HashMap<>();
-    private Object fileSystemLock = new Object();
+    static final Map<URI, Integer> openFileSystems = new HashMap<>();
+    private static final Object fileSystemLock = new Object();
 
     /**
      * Constructs a file server.
@@ -188,8 +188,9 @@ public class StaticFileServer implements StaticFileHandler {
     FileSystem getFileSystem(URI resourceURI) throws IOException {
         synchronized (fileSystemLock) {
             URI fileURI = getFileURI(resourceURI);
-            if (openFileSystems.containsKey(fileURI)) {
-                openFileSystems.put(fileURI, openFileSystems.get(fileURI) + 1);
+
+            if (openFileSystems.computeIfPresent(fileURI,
+                    (key, value) -> value + 1) != null) {
                 // Get filesystem is for the file to get the correct provider
                 return FileSystems.getFileSystem(resourceURI);
             }
@@ -206,19 +207,13 @@ public class StaticFileServer implements StaticFileHandler {
         synchronized (fileSystemLock) {
             try {
                 URI fileURI = getFileURI(resourceURI);
-                if (!openFileSystems.containsKey(fileURI)) {
-                    getLogger().debug("Tried to close non existent key {}",
-                            fileURI);
-                    return;
-                }
-                final Integer integer = openFileSystems.get(fileURI);
-                if (integer == 1) {
+                final Integer locks = openFileSystems.computeIfPresent(fileURI,
+                        (key, value) -> value - 1);
+                if (locks != null && locks == 0) {
                     openFileSystems.remove(fileURI);
                     // Get filesystem is for the file to get the correct
                     // provider
                     FileSystems.getFileSystem(resourceURI).close();
-                } else {
-                    openFileSystems.put(fileURI, integer - 1);
                 }
             } catch (IOException ioe) {
                 getLogger().error("Failed to close FileSystem for '{}'",
--- a/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
@@ -148,7 +148,7 @@ public class StaticFileServer implements StaticFileHandler {
             return false;
         }
 
-        if (resource.getProtocol().equals("jar")) {
+        if ("jar".equals(resource.getProtocol())) {
             // Get the file path in jar
             final String pathInJar = resource.getPath()
                     .substring(resource.getPath().lastIndexOf("!") + 1);
@@ -165,12 +165,32 @@ public class StaticFileServer implements StaticFileHandler {
             }
         }
 
-        // If not a jar check if a file path direcotry.
-        return resource.getProtocol().equals("file")
+        // If not a jar check if a file path directory.
+        return "file".equals(resource.getProtocol())
                 && Files.isDirectory(Paths.get(resourceURI));
     }
 
+    /**
+     * Get the file URI for the resource jar file. Returns give URI if
+     * URI.scheme is not of type jar.
+     *
+     * The URI for a file inside a jar is composed as
+     * 'jar:file://...pathToJar.../jarFile.jar!/pathToFile'
+     *
+     * the first step strips away the initial scheme 'jar:' leaving us with
+     * 'file://...pathToJar.../jarFile.jar!/pathToFile' from which we remove the
+     * inside jar path giving the end result
+     * 'file://...pathToJar.../jarFile.jar'
+     *
+     * @param resourceURI
+     *            resource URI to get file URI for
+     * @return file URI for resource jar or given resource if not a jar schemed
+     *         URI
+     */
     private URI getFileURI(URI resourceURI) {
+        if (!"jar".equals(resourceURI.getScheme())) {
+            return resourceURI;
+        }
         try {
             String scheme = resourceURI.getRawSchemeSpecificPart();
             int jarPartIndex = scheme.indexOf("!/");
--- a/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
@@ -35,7 +35,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.locks.ReentrantLock;
 import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
@@ -80,7 +79,7 @@ public class StaticFileServer implements StaticFileHandler {
 
     // Mapped uri is for the jar file
     Map<URI, Integer> openFileSystems = new HashMap<>();
-    private ReentrantLock lock = new ReentrantLock();
+    private Object fileSystemLock = new Object();
 
     /**
      * Constructs a file server.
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -75,6 +78,10 @@
     public static final Pattern APP_THEME_PATTERN = Pattern
             .compile("^\\/themes\\/[\\s\\S]+?\\/");
 
+    // Mapped uri is for the jar file
+    Map<URI, Integer> openFileSystems = new HashMap<>();
+    private ReentrantLock lock = new ReentrantLock();
```
fluorumlabs, line 83: Could this be replaced with a synchronization marker and `synchronized(...){}` blocks? The `ReentrantLock` seems to be excessive here as the code doesn't really re-enter.

caalador: changed

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -160,6 +170,64 @@ private boolean resourceIsDirectory(URL resource) {
...
+            throw new IllegalArgumentException(syntaxException.getMessage(),
+                    syntaxException);
+        }
+    }
+
+    // Package protected for feature verification purpose
+    FileSystem getFileSystem(URI resourceURI) throws IOException {
+        synchronized (fileSystemLock) {
+            URI fileURI = getFileURI(resourceURI);
+            if (openFileSystems.containsKey(fileURI)) {
```
fluorumlabs, line 191: `Map.computeIfPresent()` seems to suit better here

fluorumlabs, line 191: At the first glance `Map.computeIfPresent()` suits better this use

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -160,6 +170,64 @@ private boolean resourceIsDirectory(URL resource) {
...
+            return fileSystem;
+        }
+    }
+
+    // Package protected for feature verification purpose
+    void closeFileSystem(URI resourceURI) {
+        synchronized (fileSystemLock) {
+            try {
+                URI fileURI = getFileURI(resourceURI);
+                if (!openFileSystems.containsKey(fileURI)) {
```
fluorumlabs, line 209: At the first glance `Map.computeIfPresent` suits better this use as well

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -75,6 +77,10 @@
     public static final Pattern APP_THEME_PATTERN = Pattern
             .compile("^\\/themes\\/[\\s\\S]+?\\/");
 
+    // Mapped uri is for the jar file
+    Map<URI, Integer> openFileSystems = new HashMap<>();
+    private Object fileSystemLock = new Object();
```
fluorumlabs, line 82: Makes me thinking... How many `StaticFileServer` instances could be up simultaneously, serving content from same jars? Let's say, if I have two servlets running.

Probably better to make this `static final`

fluorumlabs: Same for the Map

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -160,6 +170,59 @@ private boolean resourceIsDirectory(URL resource) {
                 && Files.isDirectory(Paths.get(resourceURI));
     }
 
+    private URI getFileURI(URI resourceURI) {
+        try {
+            String scheme = resourceURI.getRawSchemeSpecificPart();
```
mshabarov, line 175: Does the method `getRawSchemeSpecificPart` return the string like `jar:{url}!/`? It's maybe a question to `getRawSchemeSpecificPart` docs, but I expect `scheme` string to contain `jar`, `http`, `file` or other 'scheme' in terms of URI definition. Could you please comment what options can be for 'scheme' variable here (in JAR case in the other cases)? It would really improve the readability.

caalador: The `getRawSchemeSpecificPart` returns it without the `jar:` so only  `file://....` as we at this point know that it is a jar schemed uri that we handle.

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -160,6 +170,59 @@ private boolean resourceIsDirectory(URL resource) {
...
+            int jarPartIndex = scheme.indexOf("!/");
+            if (jarPartIndex != -1) {
+                scheme = scheme.substring(0, jarPartIndex);
```
mshabarov, line 178: In case of JAR, what would be the content of 'scheme'? From the code above, it's going to be path to JAR itself, not to the file inside JAR. I propose to reflect it in the method name or in the javadocs to this method, despite it is private.

mshabarov: As at this point we certainly know that `resourceURI` pointing to a JAR, maybe it better to name this method `getJarFileURI`?

mshabarov: Or, well, it can be used in other places in the code...

caalador: `Components of all URIs: [<scheme>:]<scheme-specific-part>[#<fragment>]` so it has 2 schemes `jar:` and `file:`  so for instance for a resource URI `jar:file:///C:/Users/mikae/AppData/Local/Temp/junit3995612489514253689/fake.jar!/frontend`
The rawSchemeSpecificPart is `file:///C:/Users/mikae/AppData/Local/Temp/junit3995612489514253689/fake.jar!/frontend`  and the end scheme which we generate a new URI for is `file:///C:/Users/mikae/AppData/Local/Temp/junit3995612489514253689/fake.jar`

mshabarov: Thanks for clarifications! For me it is something that needs a comment in the code containing similar clarifications like you provided above. But up to you to add it. I've never used `rawSchemeSpecificPart` to be honest and these evolutions of the URI in this method are really hard to understand without some knowledge of `rawSchemeSpecificPart`.

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
caalador: documented method

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
```diff
@@ -323,6 +339,193 @@ public void directoryIsNotResourceRequest() throws Exception {
...
+                        } catch (Exception e) {
+                            return new Result(e);
+                        }
+                        return new Result(null);
+                    };
+                    return callable;
+                }).collect(Collectors.toList());
+
+        ExecutorService executor = Executors.newFixedThreadPool(THREADS);
+        List<Future<Result>> futures = executor.invokeAll(folderNotResource);
```
mshabarov, line 444: not blocking: Not sure this method start the threads simultaneously.
To make the threads call file server concurrently with a higher probability, I would propose to use a synchronisers, for example `CountDownLatch `.

mshabarov (approved):

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
caalador: It executes 5 threads at once and without the fix the failure rate for the test was 100% of the executions.

mshabarov: Thanks! Make sense, so let's ignore my comment then.

Issue #11181, opened 2021-06-08
lit/quoted-expressions rule in eslint-plugin-lit v1.5.0 breaks flow-client


manolo:
This error happens in CI after a  eslint-plugin-lit  1.5.0 was released
![Screenshot 2021-06-08 at 07 33 52](https://user-images.githubusercontent.com/161853/121176206-3cf27080-c85c-11eb-9549-9d38e5e0ece6.png)

knoobie:
@manolo could that be a duplicate of https://github.com/vaadin/flow/issues/11176?

Haprog:
Duplicate of #11176

PR #11182, opened 2021-06-08
chore: disable lit/quoted-expressions rule in eslint

manolo:
This should be reverted after #11181 is fixed


mshabarov (approved):

manolo:
Does not pass validation, thus not the solution for the issue, closing this.

Issue #11149, opened 2021-06-03
Failing client package build


caalador:
Quite often client-package builds fail on `Error: ENOENT: no such file or directory, copyfile 'src/main/frontend/FlowBootstrap.d.ts' -> 'target/classes/META-INF/resources/frontend/FlowBootstrap.d.ts'`

```
13:54:14]	[com.vaadin:flow-client] npm ERR! code ELIFECYCLE
[13:54:14]	[com.vaadin:flow-client] npm ERR! errno 1
[13:54:14]	[com.vaadin:flow-client] npm ERR! @vaadin/flow-deps@0.0.1 client: `node scripts/client.js`
[13:54:14]	[com.vaadin:flow-client] npm ERR! Exit status 1
[13:54:14]	[com.vaadin:flow-client] npm ERR! 
[13:54:14]	[com.vaadin:flow-client] npm ERR! Failed at the @vaadin/flow-deps@0.0.1 client script.
[13:54:14]	[com.vaadin:flow-client] npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

The client script should probably force the creation of the target folders so we can copy the existing file there in every case and not be trusting that the folder will be created by someone else.

manolo:
> client script should probably force the creation 

It actually was doing that, but async.

PR #11191, opened 2021-06-09
chore: fix random failure on copying files when building flow-client

manolo:
fixes: #11149

Note, I couldn't reproduce this in my local, but happens quite often in the build platform does for JDKs tests.
I added there an extra step for patching the file with this approach and couldn't see the failure any more.
There is no way for actually test this with an IT, or at least I do not see  any.

### DIFF
--- a/flow-client/scripts/client.js
+++ b/flow-client/scripts/client.js
@@ -15,7 +15,6 @@
  */
 
 const fs = require('fs');
-const mkdirp = require('mkdirp');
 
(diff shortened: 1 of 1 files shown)


haijian-vaadin (approved):
Thanks for the fix.

manolo:
So, I figured out the problem `mkdirp` is async, hence the target folder was not created yet when copying the files. Replacing it with `mkdirSync` is the solution

haijian-vaadin (approved):

Issue #11201, opened 2021-06-10
PwaTestIT::testPwaResources fails because of incorrect manifest URL


mshabarov:
### Description of the bug / feature

`PwaTestIT::testPwaResources` suddenly starting to fail (locally and in Flow 6,7,8 snap and PR validation) because of no protocol in the PWA manifest URL:

`<link rel="manifest" href="manifest.webmanifest">`

![Screenshot 2021-06-10 at 10 03 48](https://user-images.githubusercontent.com/61410877/121480014-4ce09080-c9d3-11eb-997d-0d50880b83e7.png)

```
java.lang.RuntimeException: testPwaResources[ANY_Chrome_](com.vaadin.flow.pwatest.ui.PwaTestIT): run failed 2 times
	at com.vaadin.testbench.RetryRule$1.evaluate(RetryRule.java:78)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.apache.maven.surefire.junitcore.pc.Scheduler$1.run(Scheduler.java:410)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.net.MalformedURLException: no protocol: manifest.webmanifest
	at java.net.URL.(URL.java:593)
	at java.net.URL.(URL.java:490)
	at java.net.URL.(URL.java:439)
	at com.vaadin.flow.pwatest.ui.PwaTestIT.readStringFromUrl(PwaTestIT.java:254)
[...]

PR #11211, opened 2021-06-10
fix: Apply theme correctly even when no Vaadin components are used

Artur-:


Haprog (dismissed):

Haprog:
The CI failure seems to be this #11201

Haprog (approved):

PR #11232, opened 2021-06-14
fix: make TS validators accept null values

haijian-vaadin:
## Description
Fixes the bug that TypeScript form validators throw exceptions when validating null values. This could happen when loading an empty value from the backend. Related to https://github.com/vaadin/fusion/issues/44

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-client/src/main/frontend/form/Validators.ts
+++ b/flow-client/src/main/frontend/form/Validators.ts
@@ -79,7 +79,7 @@ export class IsNumber extends NumberValidator<number | null | undefined> {
   }
 
   validate(value: number | null | undefined) {
-    return (this.optional && (value === null || value === undefined)) || super.validate(value);
+    return (this.optional && value == null) || super.validate(value);
   }
 }
 
--- a/flow-client/src/main/frontend/form/Validators.ts
+++ b/flow-client/src/main/frontend/form/Validators.ts
@@ -71,15 +71,15 @@ abstract class NumberValidator<T> extends AbstractValidator<T> {
   }
 }
 
-export class IsNumber extends NumberValidator<number | undefined> {
+export class IsNumber extends NumberValidator<number | null | undefined> {
   optional: boolean;
   constructor(optional: boolean, attrs?: ValidatorAttributes) {
     super({ message: 'must be a number', ...attrs });
     this.optional = optional;
   }
 
-  validate(value: number | undefined) {
-    return (this.optional && value === undefined) || super.validate(value);
+  validate(value: number | null | undefined) {
+    return (this.optional && (value === null || value === undefined)) || super.validate(value);
   }
 }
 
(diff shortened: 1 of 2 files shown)


flow-client/src/main/frontend/form/Validators.ts
```diff
@@ -97,8 +97,8 @@ export class Email extends AbstractValidator<string> {
   constructor(attrs?: ValidatorAttributes) {
     super({ message: 'must be a well-formed email address', ...attrs });
   }
-  validate(value: string | undefined) {
-    return value === undefined || isEmail(value);
+  validate(value: string | null | undefined) {
+    return !value || isEmail(value);
```
Lodin, line 101: It also rejects the empty string. Is it ok? If not, we can replace it with the loose null check:
```suggestion
    return value != null || isEmail(value);
```

flow-client/src/main/frontend/form/Validators.ts
```diff
@@ -71,15 +71,15 @@ abstract class NumberValidator<T> extends AbstractValidator<T> {
...
   optional: boolean;
   constructor(optional: boolean, attrs?: ValidatorAttributes) {
     super({ message: 'must be a number', ...attrs });
     this.optional = optional;
   }
 
-  validate(value: number | undefined) {
-    return (this.optional && value === undefined) || super.validate(value);
+  validate(value: number | null | undefined) {
+    return (this.optional && (value === null || value === undefined)) || super.validate(value);
```
Lodin, line 82: We can use the loose null check here:
```suggestion
    return (this.optional && value == null) || super.validate(value);
```

flow-client/src/main/frontend/form/Validators.ts
haijian-vaadin: I also thought about this. I think the Email validator should accept the empty string, which could happen when the user first type sth, then remove the value. 
This kind of required check should be done in the Required/NotNull validator, not this Email validator.

flow-client/src/main/frontend/form/Validators.ts
haijian-vaadin: cool, didn't know about this trick, thanks

flow-client/src/main/frontend/form/Validators.ts
Lodin: Do I understand correctly that results of `!value` and `value != null` are the same then?

haijian-vaadin: No. The difference is `''`. `''` passes the 'Email' validation with `!value`, fails with `value !=null`. As indicated in the changes of another file.
```
const validator = new Email();
assert.isTrue(validator.validate(''));
```

platosha (approved):

PR #11235, opened 2021-06-15
fix: do not open FileSystem for unknown schemes

### LINKED ISSUE #11230: Fail accessing static resources from other jar:  java.nio.file.FileSystemNotFoundException: Provider "war" not installed
### Description of the bug
The new method resourceIsDirectory in StaticFileServer class generates an exception for resources that have protocol "jar:war:file", when vaadin spring boot application is started standalone (java -jar xxxx.jar)

the exception is raised when a static resource enclosed in another jar (different than the running one) is tried to be accessed
e.g: 
Access to resource logo.png: http://localhost:8080/other-folder/logo.png

Maven project other-component has the image:
src/main/resources/META-INF/resources/other-folder/logo.png

Maven spring boot application other-vaadin-app has compilation dependency other-component 

This worked in flow-server version 2.5.3.

### Expected behavior
No exception raised
### Actual behavior
Exception raised: java.nio.file.FileSystemNotFoundException: Provider "war" not installed
### Versions:
    - Vaadin / Flow version: 14.6.3, Flow: 2.6.3
    - Spring boot: 2.4.5
    - Java version: 11
    - OS version: Ubuntu 20.04
    - Browser version (if applicable): not applicable
    - Application Server (if applicable): vaadin spring boot maven structure
    - IDE (if applicable): not applicable

caalador:
Don't open a filesystem for unknown schemes
that are probably custom for container and
have no FileSystem Provider implementation added.

Fixes #11230

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
@@ -142,7 +142,7 @@ public class StaticFileServer implements StaticFileHandler {
         if ("jar".equals(resource.getProtocol())) {
             // Get the file path in jar
             final String pathInJar = resource.getPath()
-                    .substring(resource.getPath().lastIndexOf("!") + 1);
+                    .substring(resource.getPath().indexOf('!') + 1);
             try {
                 FileSystem fileSystem = getFileSystem(resourceURI);
                 // Get the file path inside the jar.
@@ -150,7 +150,7 @@ public class StaticFileServer implements StaticFileHandler {
 
                 return Files.isDirectory(path);
             } catch (IOException e) {
-                getLogger().debug("failed to read zip file", e);
+                getLogger().debug("failed to read jar file contents", e);
             } finally {
                 closeFileSystem(resourceURI);
             }
@@ -200,8 +200,9 @@ public class StaticFileServer implements StaticFileHandler {
         synchronized (fileSystemLock) {
             URI fileURI = getFileURI(resourceURI);
             if (!fileURI.getScheme().equals("file")) {
-                throw new IOException(
-                        "Can not read scheme '" + fileURI.getScheme() + "'");
+                throw new IOException("Can not read scheme '"
+                        + fileURI.getScheme() + "' for resource " + resourceURI
+                        + " and will determine this as not a folder");
             }
 
             if (openFileSystems.computeIfPresent(fileURI,
--- a/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
@@ -30,10 +30,8 @@ import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
@@ -199,6 +199,10 @@ public class StaticFileServer implements StaticFileHandler {
     FileSystem getFileSystem(URI resourceURI) throws IOException {
         synchronized (fileSystemLock) {
             URI fileURI = getFileURI(resourceURI);
+            if (!fileURI.getScheme().equals("file")) {
+                throw new IOException(
+                        "Can not read scheme '" + fileURI.getScheme() + "'");
+            }
 
             if (openFileSystems.computeIfPresent(fileURI,
                     (key, value) -> value + 1) != null) {
--- a/flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
@@ -343,19 +343,8 @@ public class StaticFileServerTest implements Serializable {
                 "Fake should not check the file system nor be a static resource.",
                 fileServer.isStaticResourceRequest(request));
 
-        File archiveFile = new File(folder.getRoot(), "fake.jar");
-        archiveFile.createNewFile();
-        Path tempArchive = archiveFile.toPath();
+        Path tempArchive = generateZipArchive(folder);
 
-        try (ZipOutputStream zipOutputStream = new ZipOutputStream(
-                Files.newOutputStream(tempArchive))) {
-            // Create a file to the zip
-            zipOutputStream.putNextEntry(new ZipEntry("/file"));
-            zipOutputStream.closeEntry();
-            // Create a directory to the zip
-            zipOutputStream.putNextEntry(new ZipEntry("frontend/"));
-            zipOutputStream.closeEntry();
-        }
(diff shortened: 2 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -199,6 +199,10 @@ private URI getFileURI(URI resourceURI) {
     FileSystem getFileSystem(URI resourceURI) throws IOException {
         synchronized (fileSystemLock) {
             URI fileURI = getFileURI(resourceURI);
+            if (!fileURI.getScheme().equals("file")) {
+                throw new IOException(
+                        "Can not read scheme '" + fileURI.getScheme() + "'");
```
pleku, line 204: Should it output the full uri for debugging purposes to be able to figure out the causing resource ? The catching block only logs this error its message.

Or since the code is not actually doing anything itself, but just checking if it is a directory, maybe the message could reflect that "this is fine" and the error can be "ignored".

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
```diff
@@ -374,6 +376,56 @@ public void directoryIsNotResourceRequest() throws Exception {
...
+                "jar:war:" + warFile.toURI().toURL().toString() + "!/"
+                        + archiveFile.getName() + "!/frontend");
+
+        setupRequestURI("", "", "/frontend/.");
+        Mockito.when(servletService.getStaticResource("/frontend/."))
+                .thenReturn(folderResourceURL);
+
+        Assert.assertTrue(
+                "Request should return as static request as we can not determine non file resources in jar files.",
+                fileServer.isStaticResourceRequest(request));
```
pleku, line 426: How difficult would it be to add a unit test that actually verifies that the serving of the jar from the jar works, meaning that also `serveStaticResource` method call would be checked for this case ? As we don't have a regression test for that either but the added code here could help in adding the test case

pleku (commented):
Looking good, couple of questions

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
knoobie: Just a question or hint - this whole "resource" handling is quite tricky.. I've seen multiple issues on spring / spring boot over the last years with the same problems.. the amount of "protocol" and combinations is just absurd.. (e.g. jar, war, wsjar, zip, bundle..)

I'm not sure if you have looked around, but there is a handy spring framework class that handles this in spring, which could be used to get some ideas how they do it?

https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/io/support/PathMatchingResourcePatternResolver.java

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
caalador: It would seem like we would need to unzip the initial jar to be able to check the secondary jar content.

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
caalador: Yes I'm certain that we can not support everything, but should only take care of the basic one and use the old resolution for anything else.

caalador: Added more info to the exception.

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
caalador: Added test to show expectation of feature.

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
```diff
@@ -374,6 +376,110 @@ public void directoryIsNotResourceRequest() throws Exception {
...
+        warFile.createNewFile();
+        Path warArchive = warFile.toPath();
+
+        try (ZipOutputStream warOutputStream = new ZipOutputStream(
+                Files.newOutputStream(warArchive))) {
+            // Create a file to the zip
+            warOutputStream.putNextEntry(new ZipEntry(archiveFile.getName()));
+            warOutputStream.write(Files.readAllBytes(tempArchive));
+
+            warOutputStream.closeEntry();
```
pleku, line 462: This creation code could be extracted and reused for both added tests instead of 1-1 duplication

pleku (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
```diff
@@ -374,6 +376,110 @@ public void directoryIsNotResourceRequest() throws Exception {
...
+
+        Assert.assertTrue(
+                "Request should return as static request as we can not determine non file resources in jar files.",
+                fileServer.isStaticResourceRequest(request));
+
+        folder.delete();
+    }
+
+    @Test
+    public void jarInaJar_fileIsLoadedCorrectly()
```
pleku, line 431: To me test name is invalid, it should be `isStaticResource_jarInAJar_detectsAsStaticResources`

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
```diff
@@ -374,6 +376,110 @@ public void directoryIsNotResourceRequest() throws Exception {
         folder.delete();
     }
 
+    @Test
+    public void jarWarFileScheme_() throws IOException {
```
pleku, line 380: To me test name is invalid, it should be `isStaticResource_jarWarFileScheme_detectsAsStaticResources`

pleku (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
caalador: done

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
caalador: done

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
pleku: Just noting that we don't still have tests that the actual serving of the files will work (which I intended), only that those are detected properly.

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
caalador: done

flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
caalador: Serving of files is tested for instance in `serveStaticResource` and jar resources are mainly served by the ServletContainer due to the spec. Any classResource that we serve should have their own test already.

pleku (approved):

knoobie:
Has the bot a problem because this isn't targeting master?

caalador:
Could be. I'll make a PR for master and have that trickle down instead. Even though I have used the bot for automating from  2.7 to 2.6.

estronque:
Thanks, the fix works!

Issue #9862, opened 2021-01-21
Add jandex index to all necessary artifacts


mvysny:
Vaadin 14.4.6: in order to support Quarkus, so-called Jandex index needs to be created. The reason is that Quarkus doesn't rely on classpath scanning to discover classes; instead it uses Jandex index. Alternatively, the jar files needs to have an empty `beans.xml` file present. Please see https://quarkus.io/guides/cdi-reference for more details.

These files need to be present in all `flow-*` jar files.

mvysny:
As a workaround I've created a project which builds the Vaadin Jandex index; I'm also publishing the artifact on jcenter and maven central. See https://github.com/mvysny/vaadin-jandex/ for more details.

pleku:
### Acceptance Criteria
- [ ] /platform project will build Jandex index for all necessary Vaadin artifacts
    - flow + flow-components artifacts, theme integrations, everything that has an annotation that the Vaadin application is going to read
    - not needed: cdi, spring, mpr ...

pleku:
Not yet done as #11241 is pending

PR #11241, opened 2021-06-15
feat: generate jandex index for Flow

caalador:
Generate a jandex index for Flow
to use with quarkus.

part of vaadin/quarkus#1
Flow support for #9862 when no platform in use

### DIFF
--- a/flow-jandex/pom.xml
+++ b/flow-jandex/pom.xml
@@ -63,22 +63,12 @@
         </dependency>
         <dependency>
             <groupId>com.vaadin</groupId>
-            <artifactId>flow-client</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>com.vaadin</groupId>
             <artifactId>flow-html-components</artifactId>
             <version>${project.version}</version>
         </dependency>
         <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>flow-server-production-mode</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>com.vaadin</groupId>
-            <artifactId>flow-html-components-testbench</artifactId>
             <version>${project.version}</version>
         </dependency>
         <dependency>
--- a/flow-jandex/pom.xml
+++ b/flow-jandex/pom.xml
@@ -1,0 +1,147 @@
+<!--
+  ~ Copyright 2000-2021 Vaadin Ltd.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License"); you may not
+  ~ use this file except in compliance with the License. You may obtain a copy of
+  ~ the License at
+  ~
+  ~ http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


flow-jandex/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-server-production-mode</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-html-components-testbench</artifactId>
```
denis-anisimov, line 81: Is this needed at runtime (not for tests) ?

flow-jandex/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-data</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-client</artifactId>
```
denis-anisimov, line 66: Is there any Java file which should be included into the index and used at runtime ?
Shouldn't be.

flow-jandex/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+                        <exclude>com</exclude>
+                        <exclude>com/**/*</exclude>
+                        <exclude>OSGI-INF</exclude>
+                        <exclude>OSGI-INF/*</exclude>
+                    </excludes>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+</project>
```
denis-anisimov, line 147: new line

flow-jandex/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+                    <executions>
+                        <execution>
+                            <id>unpack-dependencies</id>
+                            <phase>generate-resources</phase>
+                            <goals>
+                                <goal>unpack-dependencies</goal>
+                            </goals>
+                            <configuration>
+                                <includes>**/com/vaadin/**/*.class</includes>
+                                <outputDirectory>${project.build.outputDirectory}</outputDirectory>
```
denis-anisimov, line 112: is it a correct place for unpacking ?
This is a jar project, so won't the unpacked classes be packaged into one big jar file in the result ?

denis-anisimov (changes requested):

flow-jandex/pom.xml
caalador: Shouldn't be needed. Removed.

flow-jandex/pom.xml
caalador: Shouldn't be needed. Removed.

flow-jandex/pom.xml
caalador: We unpack any com.vaadin so that they get read as the jandex plugin doesn't go through jar file contents as far as I could configure it and we exclude all classes from the generated jar.

flow-jandex/pom.xml
caalador: added

denis-anisimov (approved):

PR #11278, opened 2021-06-17
fix: Make upload requests pass Spring Security checks

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
```diff
@@ -175,11 +175,19 @@ private static boolean isFrameworkInternalRequest(String servletMappingPath,
...
+            return true;
         }
 
         return false;
     }
 
+    private static boolean isUploadRequest(
+            String requestedPathWithoutServletMapping) {
+        return requestedPathWithoutServletMapping
+                .matches("VAADIN/dynamic/resource/.*/.*/upload");
```
pleku, line 188: With this check, one can make a call to `VAADIN/dynamic/resource/../../../upload` and this will return `true`.
I don't see any encoding done here (see `isPathUnsafe` method). In theory, one could have mapped something to `/upload` or `/some-prefix-path/upload` that is secured. That makes at least the contract for the `HandlerHelper` API method a bit broken as I don't think that URL should return `true`. There is also no mention of encoding/prechecking the request in the javadocs.

I would fix it just to be safe, even if it is not exploitable through `VaadinWebSecurityConfigurerAdapter.java` in Spring add-on (I don't know if that is the case) by bypassing security measures there for a fake upload request.

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java
Artur-: Changed to check the whole string for valid characters

pleku (approved):

darjag:
I have this issue in Vaadin 20.0.3 with VaadinWebSecurityConfigurerAdapter. In my case I use with a property vaadin.urlMapping and don't test without.

Artur-:
The wast majority of Flow related features and fixed for Spring Security is only in Vaadin 21. For Vaadin 20 you can manually add a rule for allowing requests to `VAADIN/dynamic/resource/[0-9]*/[a-z0-9-]**/upload` to bypass the CSRF check

PR #11304, opened 2021-06-22
feat: Do not serialize VaadinSession by default in development

### LINKED ISSUE #11151: Allow disabling serialization of VaadinSession and disable it by default
When you are using authentication in your application, the authentication information is by default stored in the HTTP session. If you use Spring dev tools during development then one every Java change, the server will be restarted. When this happens, the session is serialized and then deserialized again. If the whole application is serializable, the session will be retained and you will stay logged in. However, in most cases the Flow application is not serializable as you can very easily create code that causes serialization of the component tree to fail. When serialization of the Flow application fails, the authentication info will not be serialized either but instead the whole session will be thrown away.

### Actual behavior
The development workflow is
1. Start app
2. Login and navigate to the view you want to modify
3. Make code changes
4. Auto reload kicks in
5. Login and navigate to the view
6. Check if you did the correct changes

### Expected behavior
The workflow should be

1. Start app
2. Login and navigate to the view you want to modify
3. Make code changes
4. Auto reload kicks in
5. Check if you did the correct changes

### Suggested solution

Add a parameter called `serializeVaadinSession` with a default value of `false`.

When `serializeVaadinSession` is `false`, the `VaadinSession` and related locks that are stored in the HTTP session are not serialized when the session is serialized. Everything else in the HTTP session is serialized.

When `serializeVaadinSession` is explicitly set to `true` in the project, use the current behavior.

Artur-:
Fixes #11151


Artur-:
Seems like this approach has a problem at least with Jetty. Jetty (in the included test) serializes the session after each request so after the first serialization, when `sessionWillPassivate` has been called, the `VaadinSession` is no longer found and the application does not work.

Artur-:
Took another approach which skips serialization instead of removing it from the session

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -2180,8 +2180,13 @@ protected VaadinSession loadSession(WrappedSession wrappedSession) {
      * @return the VaadinSession or null if no session was found
      */
     protected VaadinSession readFromHttpSession(WrappedSession wrappedSession) {
-        return (VaadinSession) wrappedSession
+        VaadinSession session = (VaadinSession) wrappedSession
                 .getAttribute(getSessionAttributeName());
+        if (session != null && !session.valid) {
```
pleku, line 2185: non-blocking: In this context the `session.valid` gives me a different impression, as in the session was invalidated. Some other name would be more understandable but the name eludes me

pleku (changes requested):
I'm going to test with some other projects+servers too

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
```diff
@@ -26,15 +36,6 @@
...
 
-import static com.vaadin.flow.server.InitParameters.BUILD_FOLDER;
-import static com.vaadin.flow.server.InitParameters.SERVLET_PARAMETER_CLOSE_IDLE_SESSIONS;
-import static com.vaadin.flow.server.InitParameters.SERVLET_PARAMETER_DEVMODE_ENABLE_LIVE_RELOAD;
-import static com.vaadin.flow.server.InitParameters.SERVLET_PARAMETER_DISABLE_XSRF_PROTECTION;
-import static com.vaadin.flow.server.InitParameters.SERVLET_PARAMETER_PRODUCTION_MODE;
-import static com.vaadin.flow.server.InitParameters.SERVLET_PARAMETER_REQUEST_TIMING;
-import static com.vaadin.flow.server.InitParameters.SERVLET_PARAMETER_SEND_URLS_AS_PARAMETERS;
-import static com.vaadin.flow.server.InitParameters.SERVLET_PARAMETER_SYNC_ID_CHECK;
-
```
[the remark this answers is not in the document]
Artur-: Fixed

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -2215,7 +2220,7 @@ protected void removeFromHttpSession(WrappedSession wrappedSession) {
      *
      * @return the attribute name used for storing the VaadinSession
      */
-    protected String getSessionAttributeName() {
+    String getSessionAttributeName() {
```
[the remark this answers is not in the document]
Artur-: Reverted

flow-tests/test-jetty-reload/src/test/java/com/vaadin/flow/uitest/ui/SessionValueIT.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.Test;
+import org.openqa.selenium.By;
+
+import net.jcip.annotations.NotThreadSafe;
+
+@NotThreadSafe
+public class SessionValueIT extends ChromeBrowserTest {
+
+    @Test
+    public void sessionValuePreservedOnRefresh() throws InterruptedException {
```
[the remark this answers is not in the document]
Artur-: Changed to reload

flow-tests/test-jetty-reload/src/test/java/com/vaadin/flow/uitest/ui/SessionValueIT.java
```diff
@@ -0,0 +1,47 @@
...
+                .replace("The custom value in the session is: ", "");
+
+        // trigger jetty reload
+        findElement(By.id(WebpackDevServerPortView.TRIGGER_RELOAD_ID)).click();
+
+        waitForElementPresent(By.id("customAttribute"));
+        div = $("div").id("customAttribute");
+        String customAttributeAfterReload = div.getText()
+                .replace("The custom value in the session is: ", "");
+        Assert.assertEquals(customAttribute, customAttributeAfterReload);
```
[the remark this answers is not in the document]
Artur-: Done

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
Artur-: Inverted and changed to `deserializedAsEmpty`

pleku (dismissed):
Tested with TomEE and it worked there too. Tried with Jboss/Wildfly but run into other issues and could not get to test hot deploy capabilities easily ... 

@denis-anisimov are you going to take a look or not ?

pleku:
Also, maybe there should be a mention of this behavior in https://github.com/vaadin/docs/blob/latest/articles/guide/live-reload/index.asciidoc, can you make a PR there @Artur- ? Essentially I would just add a note that this behavior is there to make it easier to develop apps while keeping authentication features enabled, and that the behavior will not be the same for production apps.

denis-anisimov:
Immediate question: is it only targets the `master` version or it will be picked into other versions ?

Artur-:
Only master

flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java
```diff
@@ -314,11 +314,21 @@ default boolean isEagerServerLoad() {
...
 
+    /**
+     * Checks if development mode session serialization is enabled or not.
+     * <p>
+     * By default session serialization is disabled in development mode.
+     *
+     * @return {@code true} if dev mode session serialization is enabled,
+     *         {@code false} otherwise
+     */
+    boolean isDevModeSerializeSession();
```
denis-anisimov, line 332: The name is cryptic so much that I can't understand what's that without reading javadocs.

I don't think there is a sense to care about prefix `is` for boolean property.
We already have e.g `disableAutomaticServletRegistration` method.
So getting rid of `if` allows to make the proper name for the method.

denis-anisimov:
>@denis-anisimov are you going to take a look or not ?

yes.

Is there any rush ?

flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java
denis-anisimov: This introduces a new API.
And the question related to my initial question: what versions this PR targets?

If it should be picked to other versions then there should be 2 PRs most likely: one which doesn't change API and just introduce a property which can be used even without `DeploymentConfiguration` and then change in `DeploymentConfiguration`.
Otherwise it won't be trivial to pick this change.
Or it you are going to pick the change in other branches by yourself then it's fine.

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
```diff
@@ -264,6 +265,13 @@ public boolean isDevModeLiveReloadEnabled() {
                 && enableDevServer(); // gizmo excluded from prod bundle
     }
 
+    @Override
+    public boolean isDevModeSerializeSession() {
```
denis-anisimov, line 269: The question here but may be it should be asked in the `DeploymentConfiguration`:
- is this the proper class to have this method ?

`DeploymentConfiguration` is 1:1 with a VaadinServlet in the app.
`ApplicationConfiguration` is 1:1 with the _app_.

So the question is :  do you really want to disable/enable session serialiation per VaadinServlet or per applicaiton?
So is the code ready for the case when application has several servlets and one servlet allows serialization and another doesn't ?

And does this have any sense at all ?

If it should be a setting per application then this is a wrong class along with `DeploymentConfiguration` interface.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -2180,8 +2180,13 @@ protected VaadinSession loadSession(WrappedSession wrappedSession) {
      * @return the VaadinSession or null if no session was found
      */
     protected VaadinSession readFromHttpSession(WrappedSession wrappedSession) {
-        return (VaadinSession) wrappedSession
+        VaadinSession session = (VaadinSession) wrappedSession
                 .getAttribute(getSessionAttributeName());
+        if (session != null && session.deserializedAsEmpty) {
```
denis-anisimov, line 2185: What is `session.deserializedAsEmpty`?
a public field in the session ?

There can't be any `public` fields.

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -138,6 +138,10 @@
 
     private final StreamResourceRegistry resourceRegistry;
 
+    private transient boolean serialize = true;
+
+    transient boolean deserializedAsEmpty = false;
```
denis-anisimov, line 143: Every field may be only `private` .
Make an getter or setter with a proper modifier.

denis-anisimov: `false` is a default value for a boolean field.
It's extra

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -334,6 +338,10 @@ public void setConfiguration(DeploymentConfiguration configuration) {
         }
         assert this.configuration == null : "Configuration can only be set once";
         this.configuration = configuration;
+        if (!configuration.isProductionMode()
+                && !configuration.isDevModeSerializeSession()) {
+            this.serialize = false;
```
denis-anisimov, line 343: `serialize` can be calculated at any time.
Is there any reason to introduce this as a field and then care about its `transient` nature ?

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
Artur-: You don't, just like you don't want to enable production mode / dev mode per servlet and a lot of (all?) other parameters in `DeploymentConfiguration`. But I think it falls out of scope for this PR.

flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java
Artur-: What are you suggesting as the name? `devModeSerializeSession`?

Artur-: There is `isDevModeLiveReloadEnabled` right above this. It is named in the same way except with an extra `Enabled` that gives no additional information

Artur-:
Docs in https://github.com/vaadin/docs/pull/584

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -1018,6 +1026,11 @@ public String getPushId() {
      */
     private void readObject(ObjectInputStream stream)
             throws IOException, ClassNotFoundException {
+        boolean wasSerialized = stream.readBoolean();
+        this.deserializedAsEmpty = !wasSerialized;
+        if (!wasSerialized) {
```
denis-anisimov, line 1031: I don't understand this code.
It just returns if serialization has not happened.

It means that the state of the object is not restored proeprly.
It means e.g. that transient `pendingAccessQueue` has `null` value which means that the code becomes broken.

So: shouldn't this code just thrown an exception if an object is deserialized being not serialized ?

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -1027,6 +1040,14 @@ private void readObject(ObjectInputStream stream)
         }
     }
 
+    private void writeObject(java.io.ObjectOutputStream stream)
+            throws IOException {
+        stream.writeBoolean(this.serialize);
```
denis-anisimov, line 1045: `this` is extra here and below.

flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java
Artur-: Documented in https://github.com/vaadin/docs/pull/584 as 

`devmode.serializeSession`::
Determines whether the `VaadinSesion(s)` of the users are serialized on server shutdown when running in development mode. 

Naming here and in the docs should match

flow-tests/test-jetty-reload/src/main/java/com/vaadin/flow/uitest/ui/SessionValueView.java
```diff
@@ -0,0 +1,41 @@
...
+            customAttribute = UUID.randomUUID().toString();
+            session.setAttribute("custom-attribute", customAttribute);
+        }
+        Div div = new Div();
+        div.setId("customAttribute");
+        div.setText("The custom value in the session is: " + customAttribute);
+        add(div);
+
+    }
+}
```
denis-anisimov, line 41: new line

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
denis-anisimov: No.
It's a part of the PR.
There is already `ApplicationConfiguration` and `AbstractConfiguration`.
And there is `isProductionMode` in the `AbstractConfiguration`.
So you should just use the proper class and decide which scope for the config should be used: application or sevlet.
There is nothing else to do.
Since this is a new property it should be decided now, here in this PR.
Only existing properties can be considered separately.

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
Artur-: The whole change is about conditionally not including `VaadinSession` in the serialized HTTP session.

If you throw an exception during serialization or deserialization, then you abort the whole process.

When the session is deserialized, it is marked as "empty" and then ignored and removed by `VaadinService` when encountered.

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
Artur-: I think there was a reason in the first version of the PR but probably not anymore. Shouldn't be anything preventing reading `configuration` in `writeObject`

flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java
denis-anisimov: `isDevModeSessionSerializationEnabled` is already better.
But I would go with `enableDevModeSessionSerialization` (`disable` ) or something like this.

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
Artur-: Fixed

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
Artur-: Fixed

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
Artur-: Fixed

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
Artur-: `ApplicationConfiguration` sounds better then. I will move it.

flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java
Artur-: Changed to `isDevModeSessionSerializationEnabled` which is similar to `isDevModeLiveReloadEnabled`. The property then is also similar: `devmode.sessionSerialization.enabled`

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -2180,8 +2180,13 @@ protected VaadinSession loadSession(WrappedSession wrappedSession) {
      * @return the VaadinSession or null if no session was found
      */
     protected VaadinSession readFromHttpSession(WrappedSession wrappedSession) {
-        return (VaadinSession) wrappedSession
+        VaadinSession session = (VaadinSession) wrappedSession
                 .getAttribute(getSessionAttributeName());
+        if (session != null && session.isDeserializedAsEmpty()) {
```
denis-anisimov, line 2185: Does this mean that the deserialized `VaadinSession` instance will be disregarded and a new one will be created instead ?
Would be good to have this in the comments then.

denis-anisimov: Is this tested by the way anyhow ?

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
denis-anisimov: Would be good to have this in the comments.

Another thing which I would like to see here: protection from using the instance which is not properly deserialized.
It should make itself as invalid (via some flag or somehow else) and then every access to the instance should throw an exception which says that the instance is invalid: a new one should have been created and used instead.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
Artur-: Added

Artur-: The whole thing is tested by the IT which makes it impossible to serialize the test UI but checks that the session is still serialized and deserialized. If this did not work, then you would not get a new UI after server restart + browser refresh

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
Artur-: Added an assert to `checkHasLock` which is called from most methods

Artur-: Added javadoc

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -544,6 +547,7 @@ public boolean hasLock() {
      *            and the session is not locked
      */
     public void checkHasLock(String message) {
+        assert deserializedAsEmpty == false : "The session was discarded during serialization and should not be used";
```
denis-anisimov, line 550: I think it should not be an `assert`.
`assert` is for Flow developers: it most likely will be disabled for the app developer (since you have to enable it explicitely via `-ea` ).
it should throw an exception.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
Artur-: Yes and this can only fail because of a bug in Flow. The error message has no meaning for anybody else

denis-anisimov: May be you are right.

denis-anisimov (dismissed):

pleku (approved):

PR #11309, opened 2021-06-22
fix: provide info about navigation target in NotFoundException

### LINKED ISSUE #7071: NotFoundException better error handling
Maybe default error handler could provide more information when handling `NotFoundException`

Lets assume we have this class
```
public class ViewWithoutRouteAnnotation extends Div {
}
```
and we are trying to get its URL
```
@Route
public class MainView extends Div {
	public MainView() {
		RouteConfiguration.forApplicationScope().getUrl(ViewWithoutRouteAnnotation.class);
	}
}
```
Because class does NOT have `@Route` annotation `NotFoundException` is thrown. `RouteNotFoundError` class handles exception and in development mode we get: 
image
There is no exception logged in console.
In this example everything is clear. But if we have more code in constructor we don't know which line of code is causing problem.

 - Suggestions:

First in `RouteConfiguration`  in method `getUrlForTarget` add target class name. Existing code:
```
private String getUrlForTarget(Class<? extends Component> navigationTarget,
            RouteRegistry registry) {
        Optional<String> targetUrl = registry.getTargetUrl(navigationTarget);
        if (!targetUrl.isPresent()) {
            throw new NotFoundException(
                    "No route found for given navigation target!");
        }
        return targetUrl.get();
}
```
Suggested change:
```
throw new NotFoundException("No route found for navigation target " + navigationTarget.getName() + "!");
        
```

Second in `RouteNotFoundError` add logger. For example:
```
@Override
public int setErrorParameter(BeforeEnterEvent event, ErrorParameter<NotFoundException> parameter) {
		// existing code...
	log.error(parameter.getCaughtException().getMessage(), parameter.getCaughtException());
	return HttpServletResponse.SC_NOT_FOUND;
}
```

denis-anisimov:
fixes #7071

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
@@ -48,6 +48,10 @@ public class RouteNotFoundError extends Component
     @Override
     public int setErrorParameter(BeforeEnterEvent event,
             ErrorParameter<NotFoundException> parameter) {
+        LoggerFactory.getLogger(RouteNotFoundError.class)
+                .error(parameter.hasCustomMessage()
+                        ? parameter.getCustomMessage()
+                        : "Route is not found", parameter.getCaughtException());
         String path = event.getLocation().getPath();
         String additionalInfo = "";
         if (parameter.hasCustomMessage()) {
--- a/flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
@@ -49,7 +49,7 @@ public class RouteNotFoundError extends Component
     public int setErrorParameter(BeforeEnterEvent event,
             ErrorParameter<NotFoundException> parameter) {
         LoggerFactory.getLogger(RouteNotFoundError.class)
-                .error(parameter.hasCustomMessage()
+                .info(parameter.hasCustomMessage()
                         ? parameter.getCustomMessage()
                         : "Route is not found", parameter.getCaughtException());
         String path = event.getLocation().getPath();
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/RouteNotFoundError.java
```diff
@@ -48,6 +48,10 @@
     @Override
     public int setErrorParameter(BeforeEnterEvent event,
             ErrorParameter<NotFoundException> parameter) {
+        LoggerFactory.getLogger(RouteNotFoundError.class)
+                .error(parameter.hasCustomMessage()
```
knoobie, line 52: I would propose to lower this to info, to reduce log spam with such severity.

denis-anisimov: Well, `error` is indeed too much here. But I'm not sure whether `info` is good enough : the main point of the ticket is to see the problem via logger.
Reducing spam means that it will not be shown.......
On the other hand in production info level is normally not written into the log.
But during  the development `info` should be enabled by default.
So may be you are right.

knoobie: Thanks for considering!

caalador (approved):

joelpop:
Scenario: 21.0.0.alpha9, Spring Security, custom `HasErrorParameter<NotFoundException>`’ed hander class.

1. When I attempt to navigate to an existing, unauthorized view, my hander class’s `setErrorParameter` method is called and I get a PROGRAMMATIC trigger with its `getCustomMessage` returning “Access denied”. So far so good.
2. When I attempt to navigate to a non-existent view, my hander class’s `setErrorParameter` method is called and I get a PROGRAMMATIC trigger with its `getCustomMessage` returning “Access denied” PLUS a my class’s `setErrorParameter` method (same instance, of course) is called a second time as a CLIENT_SIDE trigger with its `getCustomMessage` returning the empty string and the `get[Caught]Exception` having “Couldn’t find route for ‘xxxxxx’“.

PR #11318, opened 2021-06-23
fix: order only dependencies inside package.json

### LINKED ISSUE #11294: Do not sort main keys in package.json
Please do not touch the main keys of `package.json`. The order that `npm init` creates the keys is just fine, you should not sort those alphabetically. `npm install` also only sorts the dependencies, I have never seen it change the other parts of `package.json`

This has apparently been introduced as a side effect of #7431 

```
vaadin init --pre packagejson
cd packagejson
mvn
```

### Expected behavior
The `name` field should still be first in `package.json`

### Actual behavior
The `name` field is moved below `dependencies` and `devDependencies`

### Versions:
    - Vaadin / Flow version: 21 alpha 7

denis-anisimov:
fixes #11294

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -33,14 +33,15 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
-import com.vaadin.flow.server.Constants;
-import com.vaadin.flow.server.frontend.scanner.ClassFinder;
-import com.vaadin.flow.server.frontend.scanner.FrontendDependencies;
-import com.vaadin.flow.server.frontend.scanner.FrontendDependenciesScanner;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.vaadin.flow.server.Constants;
+import com.vaadin.flow.server.frontend.scanner.ClassFinder;
+import com.vaadin.flow.server.frontend.scanner.FrontendDependencies;
+import com.vaadin.flow.server.frontend.scanner.FrontendDependenciesScanner;
 
 import elemental.json.Json;
 import elemental.json.JsonException;
@@ -253,6 +254,13 @@ public abstract class NodeUpdater implements FallibleCommand {
             packageJson.put(DEP_LICENSE_KEY, DEP_LICENSE_DEFAULT);
             packageJson.put(DEPENDENCIES, Json.createObject());
             packageJson.put(DEV_DEPENDENCIES, Json.createObject());
+        } else {
+            if (!packageJson.hasKey(DEPENDENCIES)) {
+                packageJson.put(DEPENDENCIES, Json.createObject());
+            }
+            if (!packageJson.hasKey(DEV_DEPENDENCIES)) {
+                packageJson.put(DEV_DEPENDENCIES, Json.createObject());
+            }
         }
 
         addVaadinDefaultsToJson(packageJson);
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -252,21 +252,18 @@ public abstract class NodeUpdater implements FallibleCommand {
             packageJson = Json.createObject();
             packageJson.put(DEP_NAME_KEY, DEP_NAME_DEFAULT);
             packageJson.put(DEP_LICENSE_KEY, DEP_LICENSE_DEFAULT);
-            packageJson.put(DEPENDENCIES, Json.createObject());
-            packageJson.put(DEV_DEPENDENCIES, Json.createObject());
-        } else {
-            if (!packageJson.hasKey(DEPENDENCIES)) {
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -253,6 +254,13 @@ JsonObject getPackageJson() throws IOException {
             packageJson.put(DEP_LICENSE_KEY, DEP_LICENSE_DEFAULT);
             packageJson.put(DEPENDENCIES, Json.createObject());
             packageJson.put(DEV_DEPENDENCIES, Json.createObject());
+        } else {
+            if (!packageJson.hasKey(DEPENDENCIES)) {
+                packageJson.put(DEPENDENCIES, Json.createObject());
+            }
+            if (!packageJson.hasKey(DEV_DEPENDENCIES)) {
+                packageJson.put(DEV_DEPENDENCIES, Json.createObject());
+            }
```
caalador, line 263: Why do we specifically need these?
Would probably be better to remove both from the null case and have a `addDefaultObjects(packageJson)`
```
    static void addDefaultObjects(JsonObject json) {
        computeIfAbsent(json, DEPENDENCIES, Json::createObject);
        computeIfAbsent(json, DEV_DEPENDENCIES, Json::createObject);
   }
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -125,7 +125,31 @@ public void execute() {
...
+            sortObject(deps);
+        }
+        if (json.hasKey(DEV_DEPENDENCIES)) {
+            JsonObject deps = json.get(DEV_DEPENDENCIES);
+            sortObject(deps);
+        }
+        if (json.hasKey(VAADIN_DEP_KEY)) {
+            JsonObject vaadin = json.get(VAADIN_DEP_KEY);
+            sortObject(vaadin);
+        }
```
caalador, line 139: sortObject should handle the check so we would here only have
```
sortObject(json, DEPENDENCIES);
sortObject(json, DEV_DEPENDENCIES);
sortObject(json, VAADIN_DEP_KEY);
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -125,7 +125,31 @@ public void execute() {
...
+    private void sortObject(JsonObject object) {
+        JsonObject ordered = orderKeys(object);
+        for (String key : object.keys()) {
+            object.remove(key);
+        }
+        // add ordered keys back
+        for (String key : ordered.keys()) {
+            JsonValue value = ordered.get(key);
+            object.put(key, value);
+        }
```
caalador, line 152: I would rather than removing and adding just create a new object and set that
```
private void sortObject(JsonObject json, String key) {
  if(json.hasKey(key)) {
    json.put(key, orderKeys(json.get(key)));
  }
}
```

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
denis-anisimov: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
denis-anisimov: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
denis-anisimov: If you are talking about adding and creating an object for `dependencies` and other keys then you are breaking the order via adding and removing.
-Removing will remove the object.
-Adding will add it into the end.

So adding and removing is _exactly_ what I'm trying to avoid here: I don't want to break the existing order which is exactly one of the complains in the ticket.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -125,7 +125,24 @@ public void execute() {
...
+            return;
+        }
+        JsonObject object = json.get(key);
+        JsonObject ordered = orderKeys(object);
+        Stream.of(object.keys()).forEach(object::remove);
+        // add ordered keys back
+        Stream.of(ordered.keys()).forEach(prop -> {
+            JsonValue value = ordered.get(prop);
+            object.put(prop, value);
+        });
```
caalador, line 145: What is the added value of removing and adding the values when we already have the correct ordered object?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
caalador: Still shouldn't they always stay in the order          DEPENDENCIES - DEV_DEPENDENCIES - VAADIN_DEP_KEY
as they are always set in that order. 
But I guess they might expect moving them around wouldn't make us touch the top object ordering.

caalador (approved):

PR #11326, opened 2021-06-24
fix: Always use  unix convention for the path

### LINKED ISSUE #11220: Vaadin 20: Couldn't find the definition of the element with tag 'collapsible-vaadin-split-layout'
I just upgraded a Vaadin Flow Project from 19.0.8 to 20.0.1 and now I get the following error, when accessing a view, which contains the `CollapsibleSplitLayout`:

> java.lang.IllegalStateException: Couldn't find the definition of the element with tag 'collapsible-vaadin-split-layout' in any template file declared using '@JsModule' annotations. Check the availability of the template files in your WAR file or provide alternative implementation of the method getTemplateContent() which should return an element representing the content of the template file

I checked and the files are there inside `node_modules/@vaadin/flow-frontend/src`.
I already tried to solve the issue using the help from the chat as I was expecting a problem with our project.
What we found so far is that the issue is not related to the used add-on (in my case `CollapsibleSplitLayout`) and can be reproduced using a vaadin starter project.
It also seems to be an issue that only occurs on Windows the same project works fine on a Mac

1. Create a new starter on https://start.vaadin.com/app/
2. Add the following dependency to the pom.xml:
```
        <dependency>
            <groupId>org.vaadin.tarek</groupId>
            <artifactId>collapsible-splitlayout</artifactId>
            <version>1.0.2</version>
        </dependency>
```
3. Create a view with following code:
```
@Route(value = "Test")
@RouteAlias(value = "")
@PageTitle("Test")
public class TestView extends Div {

    public TestView() {
        CollapsibleSplitLayout cs = new CollapsibleSplitLayout();
        cs.setSizeFull();
        cs.addToPrimary(new Span("Primary"));
        cs.addToSecondary(new Span("Secondary"));
        add(cs);
    }
}
```
4. Start the application and open the view.

### Expected behavior
The view should show a `CollapsibleSplitLayout`.

### Actual behavior
Following error is shown instead:
> There was an exception while trying to navigate to '' with the exception message 'Error creating bean with name 'com.example.application.views.test.TestView': Instantiation of bean failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.example.application.views.test.TestView]: Constructor threw exception; nested exception is java.lang.IllegalStateException: Couldn't find the definition of the element with tag 'collapsible-vaadin-split-layout' in any template file declared using '@JsModule' annotations. Check the availability of the template files in your WAR file or provide alternative implementation of the method getTemplateContent() which should return an element representing the content of the template file'

### Versions:
    - Vaadin / Flow version: 20.0.1
    - Java version: 11.0.9
    - OS version: Windows 10 version 2004
    - IDE (if applicable): IntelliJ IDEA 2021.1.2

### LINKED ISSUE #11325: JsModules from addons are not loaded when running the Application class for Spring Boot
Based on new comments from #7752 from @stefanuebe:
"
This issue seems to still exists somehow, we have that now in a V20 project. I'll try to create a sample project with the reproducable issue.

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'sample.app.SampleCardView': Instantiation of bean failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [sample.app.SampleCardView]: Constructor threw exception; nested exception is java.lang.IllegalStateException: Couldn't find the definition of the element with tag 'content-card' in any template file declared using '@JsModule' annotations. Check the availability of the template files in your WAR file or provide alternative implementation of the method getTemplateContent() which should return an element representing the content of the template file
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateBean(AbstractAutowireCapableBeanFactory.java:1316) ~[spring-beans-5.3.8.jar:5.3.8]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1214) ~[spring-beans-5.3.8.jar:5.3.8]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:564) ~[spring-beans-5.3.8.jar:5.3.8]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:524) ~[spring-beans-5.3.8.jar:5.3.8]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:314) ~[spring-beans-5.3.8.jar:5.3.8]
	at com.vaadin.flow.spring.SpringInstantiator.getOrCreate(SpringInstantiator.java:117) ~[vaadin-spring-17.0.1.jar:na]
	at com.vaadin.flow.di.Instantiator.createRouteTarget(Instantiator.java:193) ~[flow-server-7.0.2.jar:7.0.2]
```

Reproduced with
1) clone https://github.com/vaadin/skeleton-starter-flow-spring (v20 branch is default)
2) add for instance
```
<dependency>
    <groupId>com.github.appreciated</groupId>
    <artifactId>card</artifactId>
    <version>2.0.0</version>
</dependency>
```
4) then clean install and run as a Spring Boot Java App.
"

caalador:
Add on templates didn't resolve correclty on windows
due to an issue in the path to flow-frontend getting
`\` instead of `/`.

Fixes #11220
Fixes #11325

### DIFF
--- a/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
+++ b/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
@@ -199,7 +199,8 @@ public final class BundleLitParser {
             // "node_modules/@vaadin/flow-frontend/" instead of "./"
             // "target/flow-frontend/" instead of "./"
             if (name.contains(FLOW_NPM_PACKAGE_NAME) || name.contains(service
-                    .getDeploymentConfiguration().getFlowResourcesFolder())) {
+                    .getDeploymentConfiguration().getFlowResourcesFolder()
+                    .replaceAll("\\\\", "/"))) {
                 alternativeFileName = alternativeFileName.replaceFirst("\\./",
                         "");
             }
--- a/flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
@@ -120,10 +120,29 @@ public interface AbstractConfiguration extends Serializable {
                 false);
     }
 
+    /**
+     * Return the defined build folder for the used build system.
+     * <p>
+     * Default value is <code>target</code> used by maven and the gradle plugin
+     * will set it to <code>build</code>.
+     *
+     * @return build folder name, default {@code target}
+     */
     default String getBuildFolder() {
         return getStringProperty(InitParameters.BUILD_FOLDER, Constants.TARGET);
     }
 
+    /**
+     * Get the location for flow resources inside the build folder.
+     * <p>
+     * Default will be <code>target/flow-frontend</code> where target is the
+     * defined buildFolder see {@link #getBuildFolder()}.
+     * <p>
+     * Note! The path is made using Paths.get which will use File.separator that
+     * is OS dependent.
+     *
+     * @return flow resources folder, default {@code target/flow-frontend}
+     */
     default String getFlowResourcesFolder() {
         return Paths.get(getBuildFolder(),
                 FrontendUtils.DEFAULT_FLOW_RESOURCES_FOLDER).toString();
--- a/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/internal/LitTemplateParserImplTest.java
+++ b/flow-lit-template/src/test/java/com/vaadin/flow/component/littemplate/internal/LitTemplateParserImplTest.java
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.component.littemplate.internal;
 
+import java.nio.file.Paths;
 import java.util.Properties;
 import java.util.stream.Stream;
 
@@ -58,7 +59,8 @@ public class LitTemplateParserImplTest {
         Properties properties = new Properties();
         Mockito.when(configuration.getInitParameters()).thenReturn(properties);
         Mockito.when(configuration.getFlowResourcesFolder()).thenReturn(
-                "target/" + FrontendUtils.DEFAULT_FLOW_RESOURCES_FOLDER);
+                Paths.get("target", FrontendUtils.DEFAULT_FLOW_RESOURCES_FOLDER)
+                        .toString());
 
         Instantiator instantiator = Mockito.mock(Instantiator.class);
         Mockito.when(instantiator.getServiceInitListeners())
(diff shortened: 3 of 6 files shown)


flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
```diff
@@ -199,7 +199,8 @@ private static String getSourceFromObject(JsonObject module,
             // "node_modules/@vaadin/flow-frontend/" instead of "./"
             // "target/flow-frontend/" instead of "./"
             if (name.contains(FLOW_NPM_PACKAGE_NAME) || name.contains(service
-                    .getDeploymentConfiguration().getFlowResourcesFolder())) {
+                    .getDeploymentConfiguration().getFlowResourcesFolder()
```
pleku, line 202: Should it be somehow highlighted in javadocs, which are by the way missing completely, that the `getFlowResourcesFolder` will return different result for different OS ? Same goes for the `getBuildFolder` method.
Maybe check if any other places need similar fixes ?

pleku (changes requested):
This will need to be manually picked to 7.0 since the `lit` import will be different there

flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
caalador: Added javaDocs with highlight on the system dependent path. Interesting that no one noted that they were missing in the original pr.

flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
```diff
@@ -120,10 +120,29 @@ default boolean isXsrfProtectionEnabled() {
...
     }
 
+    /**
+     * Get the location for flow resources inside the build folder.
+     * <p>
+     * Default will be <code>target/flow-frontend</code> where target is the
+     * defined buildFolder see {@link #getBuildFolder()}.
+     * <p>
+     * Note! The path is made using Paths.get which will use File.separator that
+     * is OS dependent.
```
pleku, line 142: I wonder if it could be just made OS independent here from the start ...

pleku (approved):

flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
pleku: Sonar should have blocked it

flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
caalador: I guess it could be changed as I think the idea was more to not get file paths like `c:\\users\\home\\code\\target/flow-frontend` even if Windows should be able to handle it.

Issue #10965, opened 2021-05-10
`setErrorParameter` of error view with `HasErrorParameter<NotFoundException>` called for non-HTML requests


OlliTietavainenVaadin:
### Description of the bug / feature
If a Vaadin 14 application contains a class that `implements HasErrorParameter<NotFoundException>`, the `setErrorParameter` method gets invoked for any request that returns a 404 error, even for e.g. image requests (when the requested image is missing). 

### Minimal reproducible example
`PageNotFoundView.java`:
```
@ParentLayout(MainLayout.class)
public class PageNotFoundView extends VerticalLayout implements HasErrorParameter<NotFoundException> {

    @Override
    public int setErrorParameter(BeforeEnterEvent event, ErrorParameter<NotFoundException> parameter) {
        add(new Span("Oh no, not found"));
        return 404;
    }
}
```

`OtherView.java`:
```
    Image image = new Image("not_found.jpeg", "alt");
    add(image);
```

### Expected behavior

`setErrorParameter` doesn't get invoked when navigating into `OtherView`

### Actual behavior

`setErrorParameter` gets invoked, even though the `PageNotFoundView` is never displayed in the UI. The request for the image also gets a `text/html` response, which is against the request's accept string.

### Versions:
    - Vaadin / Flow version: 14.5.4
    - Java version: 11
    - OS version: Windows
    - Browser version (if applicable): n/a
    - Application Server (if applicable): Spring Boot / Tomcat (n/a) 
    - IDE (if applicable): n/a

denis-anisimov:
It's weird : I cannot reproduce this with the `master` but can with Flow `2.6`.

denis-anisimov:
I don't see this as a bug in fact.
It's rather an enhancement and I think it can't be fixed properly.

`not_found.jpeg` is relative URI which will be something `http://hostname/context/not_found.jpeg`.
This URL for image makes a separate request to the server.
And there are two cases:
- the URL is considered as a resource URL (the current usecase) 
- the URL has no semantical context.

If I ask `http://hostname/context/not_found.jpeg` in the browser address bar then there is no way to say whether this is a view name (some Route navigation target) or just an attempt to ask a resource.
The URL looks like a resource but I don't see any reason why I can't use `not_found.jpeg` as a route target.

So when you asks `http://hostname/context/not_found.jpeg` directly then static file server doesn't handle the request: there is no such resource and the request is handled by the last Request Handler in the list of handlers (`BootstrapHandler` ).
[...]

PR #11334, opened 2021-06-25
Don't use NotFoundException view for resources

denis-anisimov:
## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -26,8 +26,10 @@ import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
@@ -35,11 +37,14 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.ServiceLoader;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.apache.http.HttpHeaders;
+import org.apache.http.entity.ContentType;
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.DataNode;
 import org.jsoup.nodes.Document;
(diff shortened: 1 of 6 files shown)


pleku (changes requested):
Like discussed in the ticket, not sure if this should be taken in or not, so blocking for now.

Personally I can't say which is better, to allow certain accept headers or to block certain accept headers. To be continued in the ticket.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -483,6 +493,34 @@ public boolean synchronizedHandleRequest(VaadinSession session,
...
+        if (!headers.contains(HttpHeaders.REFERER)) {
+            return true;
+        }
+        Set<String> typesWithWeight = getAcceptTypes(request);
+        // text/html or application/xhtml+xml mime types should have weight 1.0
+        return typesWithWeight.contains(ContentType.TEXT_HTML.getMimeType())
+                || typesWithWeight.contains(TEXT_HTML_TYPE_WEIGHT)
+                || typesWithWeight.contains(
+                        ContentType.APPLICATION_XHTML_XML.getMimeType())
+                || typesWithWeight.contains(XHTML_TYPE_WEIGHT);
```
pleku, line 511: I think the [suggestion](https://github.com/vaadin/flow/issues/10965#issuecomment-870460570) from @Legioth makes sense,  what do you think @denis-anisimov ?

> Anything with text/html in the group with highest priority should certainly cause a UI to be created.
If we want to slightly reduce the risk of regressions, then we could consider to also create a UI for requests with text/html at a lower priority, or with only */* as the only value.
We should definitely not create an UI if the browser doesn't at all ask for text/html or if it has */* only to supplement some more specific request.

denis-anisimov: I don't think I understand what you mean.

>Anything with text/html in the group with highest priority should certainly cause a UI to be created.

Implemented here in this PR.

>If we want to slightly reduce the risk of regressions, then we could consider to also create a UI for requests with text/html at a lower priority, or with only / as the only value.
We should definitely not create an UI if the browser doesn't at all ask for text/html or if it has / only to supplement some more specific request.

It is possible to configure `Accept` header manually in the browser so technically it may contain anything (and  unexpected things) but any meaningful request contains `Accept` header which accepts _everything_ with lower weight.
E.g. image request `Accept` header is `image/avif,image/webp,image/apng,image/svg+xml,image/*,*/*;q=0.8`.
`*/*` matches `text/html`.

So it doesn't add anything to clarify the PR/direction to improve it.

If you see some exact steps which may be done here please just describe them in the details.

Legioth: If the browser specifically asks for `text/html` at any priority, then that's a clear indication that the browser might intend to use the response body as HTML.

Browsers usually include `*/*` at a low priority as a fallback (e.g. `text/css,*/*;q=0.1`), so it would defeat the purpose if we would interpret that to match `text/html` even though it does match from a strictly technical sense.

As an exception to the exception, asking for only `*/*` could still be defined to return a Flow bootstrap page since that's a very generic request for which we cannot make any assumptions for what the browser might do.

> It is possible to configure `Accept` header manually in the browser so technically it may contain anything

If the user has configured the browser to send an arbitrary `Accept` header value, then it's quite acceptable that the error handling might not work in the best possible way. The only result is that the user will see a generic 404 response instead of seeing a Flow view that says that nothing was found for the given URL. If the value doesn't match any of the patterns that we look for, then we can treat it the same way as if no `Accept` header was present.

pleku:
Closing this PR for now as it is not deemed important enough to make it better.

Please comment on the issue #10965 if this is important for someone.

PR #11341, opened 2021-06-28
feat: add Vaadin Nonnull annotation

Lodin:
Adds a `@Nonnull` annotation that is applicable to not only fields/methods/parameters like `javax.annotation.Nonnull` but also to type parameters. It gives more control over the generated Typescript code.

It is still possible to use any other annotation with the case-insensitive `nonnull` name. 

Resolves vaadin/fusion#50


fusion-endpoint/src/main/java/com/vaadin/fusion/ExplicitNullableTypeChecker.java
```diff
@@ -40,22 +41,29 @@
...
+            annotations = Stream.concat(annotations, Stream
+                    .of(((Field) element).getAnnotatedType().getAnnotations()));
+        } else if (element instanceof Parameter) {
+            isPrimitive = ((Parameter) element).getType().isPrimitive();
+            annotations = Stream.concat(annotations, Stream.of(
+                    ((Parameter) element).getAnnotatedType().getAnnotations()));
+        }
+
+        return isPrimitive || annotations.anyMatch(annotation -> "nonnull"
+                .equalsIgnoreCase(annotation.annotationType().getSimpleName()));
```
gilberto-torrezan, line 58: The entire logic can be optimized in case the field/parameter is primitive. The entire collection of annotations and concat operations could be skipped.

fusion-endpoint/src/main/java/com/vaadin/fusion/Nonnull.java
```diff
@@ -0,0 +1,13 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
+public @interface Nonnull {
+}
```
gilberto-torrezan, line 13: There's this thing that every Java file should end with a line break.

fusion-endpoint/src/test/java/com/vaadin/fusion/NonnullEntity.java
```diff
@@ -0,0 +1,16 @@
...
+public class NonnullEntity {
+    @Nonnull
+    private final List<@Nonnull String> nonNullableField = new ArrayList<>();
+
+    @Nonnull
+    public String nonNullableMethod(
+            @Nonnull Map<String, @Nonnull String> nonNullableParameter) {
+        return nonNullableParameter.getOrDefault("test", "");
+    }
+}
```
gilberto-torrezan, line 16: New line missing.

fusion-endpoint/src/main/java/com/vaadin/fusion/Nonnull.java
```diff
@@ -0,0 +1,13 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
+public @interface Nonnull {
```
gilberto-torrezan, line 12: Javadocs are missing, since this is a public class in the framework.

It should explain that the annotation is about, when to use it, and should point out that any other `@Nonnull` is also fine.

fusion-endpoint/src/test/java/com/vaadin/fusion/generator/VaadinConnectClientGeneratorTest.java
```diff
@@ -28,6 +28,8 @@
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
+import com.vaadin.fusion.utils.TestUtils;
```
gilberto-torrezan, line 31: Unnecessary import?

fusion-endpoint/src/test/java/com/vaadin/fusion/NonnullTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+        @Test
+        public void should_haveMethodParameterWithNonNullableCollectionItemType() {
+            assertTrue(parameter.getType().asClassOrInterfaceType()
+                    .getTypeArguments().get().get(1).getAnnotations().stream()
+                    .anyMatch(annotation -> "Nonnull"
+                            .equals(annotation.getName().asString())));
+        }
+    }
+}
```
gilberto-torrezan, line 124: New line missing.

gilberto-torrezan (changes requested):

fusion-endpoint/src/test/java/com/vaadin/fusion/generator/VaadinConnectClientGeneratorTest.java
Lodin: No, it is used through the tests.

fusion-endpoint/src/main/java/com/vaadin/fusion/Nonnull.java
Lodin: Could you please check if the documentation I added is correct?

gilberto-torrezan: There's no collection item type - the need for this annotation is to allow using it for type parameters, which the traditional Nonnull doesn't support.

By the way, according to the [javadocs](https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/ElementType.html), the `TYPE_USE` also includes `TYPE_PARAMETER` - can you check if `TYPE_PARAMETER` can be removed?

gilberto-torrezan (changes requested):

gilberto-torrezan (approved):

PR #11342, opened 2021-06-28
fix: use ServletContext classloader as a default classloader for serv…

### LINKED ISSUE #4367: VaadinService classloader is incorrectly set in the CTOR
There is a method `setDefaultClassLoader` which is called in the `VaadinService`CTOR.

(By the way this is horrible that we have protected method which is _intended to be overwritten_ in subclasses which is called in the CTOR but that's another issue).

This method sets the classloader for the `VaadinService` to the `Thread.currentThread().getContextClassLoader()`.

Apparently this is intentional code which is done to be able to acquire appropriate class loader to load application's classes (e.g. UI) (from javadocs).

Not sure whether this is done for OSGi case or not but exactly for OSGi it doesn't work.

It might be that it depends on how OSGi declares and uses Vaadin HTTP servlet.

I have an Activator class which register the servlet via HTTP service (see code below).
I  have to do it as it's in the code (with one more call for `setClassLoader`) otherwise it fails with inability to load `UI` class.

We should take care about this way of activating servlet.

```
public class Activator implements BundleActivator {

    @Override
    public void start(BundleContext ctx) throws Exception {
        Hashtable<String, String> props = new Hashtable<>();
        props.put("osgi.http.whiteboard.servlet.pattern", "/");
        VaadinServlet servlet = new VaadinServlet() {
            @Override
            public void init(ServletConfig servletConfig)
                    throws ServletException {
                super.init(servletConfig);

                getService().setClassLoader(getClass().getClassLoader());
            }
        };

        ctx.registerService(Servlet.class.getName(), servlet, props);

    }

    @Override
    public void stop(BundleContext context) throws Exception {
    }

}
```

denis-anisimov:
…ice (#11308)

Fixes #4367

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
@@ -16,7 +16,9 @@
 
 package com.vaadin.flow.server;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
+import javax.servlet.Servlet;
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletResponse;
 
 import java.io.BufferedWriter;
 import java.io.IOException;
@@ -49,10 +51,6 @@ import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
-
-import javax.servlet.Servlet;
-import javax.servlet.ServletContext;
-import javax.servlet.http.HttpServletResponse;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -2395,4 +2376,39 @@ private void runWithServiceContext(Runnable runnable) {
             setCurrent(null);
         }
     }
+
+    /**
+     * Get the name of the CSRF Token attribute in HTTP session.
+     *
+     * @return the attribute name string
+     */
+    public static String getCsrfTokenAttributeName() {
```
pleku, line 2385: This should not be picked ?

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
denis-anisimov: Git is absolutely unable to pick properly.

Thanks.

pleku (approved):

PR #11353, opened 2021-06-29
fix: Update buildIdentifier comparison

### LINKED ISSUE #11352: Faulty comparison for buildIdentifier in FrontendVersion
FrontendVersion compares wrong the version when comparing the buildIdentifier string.

With input
`new FrontendVersion("2.0.0-alpha6").isOlderThan(new FrontendVersion("2.0.0-alpha13"))`
Expected 
  `true`
Actual
  `false`

caalador:
Fix buildIdentifier comparison so
values compare correctly.

fixes #11352

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
@@ -326,7 +326,8 @@ public class FrontendVersion
         final Matcher otherMatcher = buildIdentifierParser
                 .matcher(other.buildIdentifier);
         if (thisMatcher.find() && otherMatcher.find()) {
-            if (thisMatcher.group(1).compareTo(otherMatcher.group(1)) != 0) {
+            if (thisMatcher.group(1)
+                    .compareToIgnoreCase(otherMatcher.group(1)) != 0) {
                 // If we do not have a text identifier assume newer
                 // If other doesn't have text identifier assume older
                 if (thisMatcher.group(1).isEmpty()) {
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
@@ -17,6 +17,8 @@ package com.vaadin.flow.server.frontend;
 
 import java.io.Serializable;
 import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Version object for frontend versions comparison and handling.
@@ -27,6 +29,13 @@ import java.util.Objects;
  */
 public class FrontendVersion
         implements Serializable, Comparable<FrontendVersion> {
+
+    /**
+     * Parses the buildIdentifier to String + Integer for instance beta1 =>
+     * 'beta' and '1'
+     */
+    private final Pattern buildIdentifierParser = Pattern
+            .compile("(\\D*)(\\d*)");
 
     /**
      * The version number of this release. For example "6.2.0". Always in the
@@ -306,9 +315,27 @@ public class FrontendVersion
                     && other.buildIdentifier.isEmpty()) {
                 return -1;
             }
-            return buildIdentifier.compareToIgnoreCase(other.buildIdentifier);
+            return compareBuildIdentifier(other);
         }
         return 0;
+    }
+
+    private int compareBuildIdentifier(FrontendVersion other) {
+        final Matcher thisMatcher = buildIdentifierParser
+                .matcher(buildIdentifier);
+        final Matcher otherMatcher = buildIdentifierParser
+                .matcher(other.buildIdentifier);
+        if (thisMatcher.groupCount() == 2 && otherMatcher.groupCount() == 2) {
+            thisMatcher.find();
+            otherMatcher.find();
+            if (thisMatcher.group(1).compareTo(otherMatcher.group(1)) != 0) {
+                return thisMatcher.group(1)
+                        .compareToIgnoreCase(otherMatcher.group(1));
+            }
+            return Integer.parseInt(thisMatcher.group(2))
+                    - Integer.parseInt(otherMatcher.group(2));
+        }
+        return buildIdentifier.compareToIgnoreCase(other.buildIdentifier);
     }
 
     private String getInvalidVersionMessage(String name, String version) {
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
@@ -31,7 +31,7 @@ public class FrontendVersion
         implements Serializable, Comparable<FrontendVersion> {
 
     /**
-     * Parses the buildIdentifier to String + Integer for instance beta1 =>
+     * Parses the buildIdentifier to String + Integer for instance beta1 returns
      * 'beta' and '1'
      */
     private final Pattern buildIdentifierParser = Pattern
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
```diff
@@ -306,11 +315,28 @@ public int compareTo(FrontendVersion other) {
                     && other.buildIdentifier.isEmpty()) {
                 return -1;
             }
-            return buildIdentifier.compareToIgnoreCase(other.buildIdentifier);
+            return compareBuildIdentifier(other);
         }
         return 0;
     }
 
+    private int compareBuildIdentifier(FrontendVersion other) {
```
pleku, line 323: Just out of curiousity - what if someone uses e.g. `2.0.0.1` for some reason - does it end up to the build identifier ? Can you add a test that it wins over prerelease versions like `2.0.0-alpha5`.

caalador: But ain't that eXo Versioning of Major.Minor.Maintenance.Patch-Qualifier and we support semantic versioning of Major.Minor.Revision-BuildIdentifier (we ofc actually accept . . . . instead of . . . - because it's simpler)
Also no it returns wrong as did the previous version.

caalador: That should cover it also.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
```diff
@@ -306,11 +315,35 @@ public int compareTo(FrontendVersion other) {
...
+                .matcher(buildIdentifier);
+        final Matcher otherMatcher = buildIdentifierParser
+                .matcher(other.buildIdentifier);
+        if (thisMatcher.groupCount() == 2 && otherMatcher.groupCount() == 2) {
+            thisMatcher.find();
+            otherMatcher.find();
+            if (thisMatcher.group(1).compareTo(otherMatcher.group(1)) != 0) {
+                // Empty alphabetical matcher is worth more than one with value
+                // so we need to reverse the check orderin that case.
+                if (otherMatcher.group(1).isEmpty()) {
```
pleku, line 334: I don't get this code and I think it is not just me - if the string for `otherMatcher.group(1)` is empty, why is it then compared ? Isn't it just the case that there is no _string_ group for that one and thus the numeric value will mean that it is newer ?

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendVersionTest.java
```diff
@@ -199,6 +199,28 @@ public void newerVersionByBuildIdentifier() {
...
+
+        assertTrue("same versions should equal",
+                test.isEqualTo(new FrontendVersion("2.0.0.alpha20")));
+    }
+
+    @Test
+    public void testAgainstVersionWithValueInBuildInfo() {
+        FrontendVersion test = new FrontendVersion("2.0.0-alpha3");
+        assertTrue("2.0.0-alpha3 should be older than 2.0.0.5",
+                test.isOlderThan(new FrontendVersion("2.0.0.5")));
```
pleku, line 221: Please test also that it works the other way around

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendVersionTest.java
```diff
@@ -199,6 +199,28 @@ public void newerVersionByBuildIdentifier() {
...
+    }
```
pleku, line 222: Please add a test that `2.0.0.6` is older than `2.0.0.11` for both ways

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendVersionTest.java
```diff
@@ -199,6 +199,28 @@ public void newerVersionByBuildIdentifier() {
                 test.isNewerThan(new FrontendVersion("2.0.0-RC1")));
     }
 
+    @Test
+    public void buildIdentifierNumbers_returnsAsExpected() {
+
+        FrontendVersion test = new FrontendVersion("2.0.0-alpha6");
+        assertTrue("2.0.0-alpha6 should be older than 2.0.0-alpha13",
+                test.isOlderThan(new FrontendVersion("2.0.0-alpha13")));
```
pleku, line 207: Looking at the code logic, I wonder if all the comparisons here should be double checked the other way too, e.g. 
```java
assertFalse("2.0.0-alpha13 should not be older than 2.0.0-alpha20",
                new FrontendVersion("2.0.0-alpha13").isOlderThan(test));
```

pleku (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
caalador: Fixed. Also other issues in the check.

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendVersionTest.java
caalador: added older newer checks for multiple choices.

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendVersionTest.java
caalador: added

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendVersionTest.java
caalador: added older newer checks for multiple choices.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
```diff
@@ -306,11 +315,34 @@ public int compareTo(FrontendVersion other) {
...
         return 0;
     }
 
+    private int compareBuildIdentifier(FrontendVersion other) {
+        final Matcher thisMatcher = buildIdentifierParser
+                .matcher(buildIdentifier);
+        final Matcher otherMatcher = buildIdentifierParser
+                .matcher(other.buildIdentifier);
+        if (thisMatcher.find() && otherMatcher.find()) {
+            if (thisMatcher.group(1).compareTo(otherMatcher.group(1)) != 0) {
```
pleku, line 329: Shouldn't this ignore the case like below? In the unlikely event that someone would have caps on?

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendVersionTest.java
```diff
@@ -199,6 +199,56 @@ public void newerVersionByBuildIdentifier() {
...
+
+        test = new FrontendVersion("2.0.0.alpha20");
```
pleku, line 209: There is a dot here instead of dash but that is fine

pleku (changes requested):
Almost, one minor comment

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendVersion.java
caalador: done

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendVersionTest.java
caalador: done

manolo (approved):

manolo:
Tested manually in my local and works ok.

pleku (approved):

PR #11384, opened 2021-07-06
feat: extract validation error handler from Binder

denis-anisimov:
part of #10598, replacement for #10683

## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -2427,7 +2427,7 @@ public class Binder<BEAN> implements Serializable {
      * Gets the validation error of this form.
      * <p>
      * If none has been set with
-     * {@link #setValidationErrorHandler(BinderValidationErrorHandler), the
+     * {@link #setValidationErrorHandler(BinderValidationErrorHandler)}, the
      * default implementation is returned.
      *
      * @return the error handler used, never <code>null</code>
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -43,7 +43,6 @@ import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.HasText;
-import com.vaadin.flow.component.HasValidation;
 import com.vaadin.flow.component.HasValue;
 import com.vaadin.flow.component.HasValue.ValueChangeEvent;
 import com.vaadin.flow.component.HasValue.ValueChangeListener;
@@ -1519,6 +1518,8 @@ public class Binder<BEAN> implements Serializable {
 
     private BinderValidationStatusHandler<BEAN> statusHandler;
 
+    private BinderValidationErrorHandler errorHandler = new DefaultBinderValidationErrorHandler();
+
     private Set<Binding<BEAN, ?>> changedBindings = new LinkedHashSet<>();
 
     private boolean validatorsDisabled = false;
@@ -2423,6 +2424,46 @@ public class Binder<BEAN> implements Serializable {
     }
 
     /**
+     * Gets the validation error of this form.
+     * <p>
+     * If none has been set with
+     * {@link #setValidationErrorHandler(BinderValidationErrorHandler), the
+     * default implementation is returned.
+     *
+     * @return the error handler used, never <code>null</code>
+     * @see #handleError(HasValue, ValidationResult)
+     * @see #clearError(HasValue)
+     * @see DefaultBinderValidationErrorHandler
+     */
+    public BinderValidationErrorHandler getValidationErrorHandler() {
+        return errorHandler;
+    }
+
+    /**
+     * Sets the validation error handler to update error status for fields.
+     * <p>
+     * The error handler is invoked by methods
+     * {@link #handleError(HasValue, ValidationResult)} and
+     * {@link #clearError(HasValue)}.
+     * <p>
+     * {@link DefaultBinderValidationErrorHandler} instance is used if the
+     * handler is not explicitly set.
+     *
+     * @param handler
+     *            the status handler to set, not <code>null</code>
+     * @throws NullPointerException
+     *             for <code>null</code> status handler
+     * @see #handleError(HasValue, ValidationResult)
+     * @see #clearError(HasValue)
+     */
+    public void setValidationErrorHandler(
+            BinderValidationErrorHandler handler) {
+        Objects.requireNonNull(handler, "Cannot set a null "
+                + BinderValidationErrorHandler.class.getSimpleName());
+        errorHandler = handler;
+    }
+
+    /**
      * Adds status change listener to the binder.
      * <p>
      * The {@link Binder} status is changed whenever any of the following
@@ -2540,6 +2581,8 @@ public class Binder<BEAN> implements Serializable {
     /**
      * Handles a validation error emitted when trying to write the value of the
      * given field.
+     * <p>
+     * TODO:
      *
      * @param field
      *            the field with the invalid value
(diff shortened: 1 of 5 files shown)


flow-data/src/main/java/com/vaadin/flow/data/binder/DefaultBinderValidationErrorHandler.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Default implementation of {@link BinderValidationErrorHandler}.
+ * <p>
+ * This handler applies visual effects for the field if its type allows this:
+ * <ul>
+ * <li>If the field's class implements {@link HasValidation} interface then its
+ * validity and error message is set based on a {@link ValidationResult}
+ * <li>If the field's class may have a theme (e.g. implements {@link HasTheme}
+ * or {@link HasElement}) then theme name derived from the {@link ErrorLevel} of
+ * the {@link ValidationResult} instance is applied.
```
pleku, line 36: Please give an example of what the theme name would be, e.g. 
> For example, for `ErrorLevel.WARNING` the element will get the `"theme"="warning"` attribute and value in HTML.

flow-data/src/main/java/com/vaadin/flow/data/binder/BinderValidationErrorHandler.java
```diff
@@ -0,0 +1,56 @@
...
+ * @see Binder#setValidationErrorHandler(BinderValidationErrorHandler)
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+public interface BinderValidationErrorHandler extends Serializable {
+
+    /**
+     * Handles a validation error emitted when trying to write the value of the
+     * given field.
```
pleku, line 40: Could add a mention that "See #clearError for clearing the error."

flow-data/src/main/java/com/vaadin/flow/data/binder/BinderValidationErrorHandler.java
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @param field
+     *            the field with the invalid value
+     * @param result
+     *            the validation error result
+     */
+    void handleError(HasValue<?, ?> field, ValidationResult result);
+
+    /**
+     * Clears the error condition of the given field, if any.
```
pleku, line 50:
```suggestion
     * Clears the error condition of the given field, if one has been
     * previously set with {@link #handleError}.
```

flow-data/src/main/java/com/vaadin/flow/data/binder/BinderValidationErrorHandler.java
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @param field
+     *            the field with an invalid value
```
pleku, line 53:
```suggestion
     *            the field to clear the previous error from
```

flow-data/src/main/java/com/vaadin/flow/data/binder/BinderValidationErrorHandler.java
```diff
@@ -0,0 +1,56 @@
...
+ * handling.
+ *
+ * 
+ * @see DefaultBinderValidationErrorHandler
+ * @see Binder#setValidationErrorHandler(BinderValidationErrorHandler)
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+public interface BinderValidationErrorHandler extends Serializable {
```
pleku, line 36: ### API review comment/speculation:

An alternative would have been to make this a functional interface so a lambda could have been used.
For example:
```java
binder.setValidationErrorHandler(field, result -> {
    if (result.isError()) {
        ...
    } else { // OK
        ... 
    }
});
```
But to be honest, when I think about this, one would probably make a reusable class for this and use it for all `Binder`s in their application. Or extend and create their own `MyBinder` that changes is explicitly. And now the default implementation can be customized quite easily if required as it protected API.

So I'm OK with the taken approach here by Denis. Do you have any strong opinion on this @knoobie, would the suggested approach work for you ?

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2422,6 +2423,46 @@ public void setValidationStatusHandler(
...
+     * @see #handleError(HasValue, ValidationResult)
+     * @see #clearError(HasValue)
+     * @see DefaultBinderValidationErrorHandler
+     */
+    public BinderValidationErrorHandler getValidationErrorHandler() {
+        return errorHandler;
+    }
+
+    /**
+     * Sets the validation error handler to update error status for fields.
```
pleku, line 2443:
```suggestion
     * Sets the validation error handler to update error status for fields
     * when the user input is validated.
```

pleku (changes requested):
I feel like we should also check/test whether this causes any UX changes to the components.
Also adding documentation to the binder chapter would be good too, but can be done later.

flow-data/src/main/java/com/vaadin/flow/data/binder/BinderValidationErrorHandler.java
knoobie: We already have a custom `_ProductName_Binder` class that extends `Binder` because of some internal helper methods we have added for i18n. 

I would personally prefer the current implementation over functional interface style. People familiar with the binder concept already know both names `handleError` and `clearError` as both are already available as `protected` methods. So the whole `ValidationErrorHandler` is just a nice abstraction on top - "No new concept to learn". :)

pleku: Thanks for the input, lets go with this.

flow-data/src/main/java/com/vaadin/flow/data/binder/BinderValidationErrorHandler.java
denis-anisimov: done

flow-data/src/main/java/com/vaadin/flow/data/binder/DefaultBinderValidationErrorHandler.java
denis-anisimov: done

pleku (approved):
Tested that the UX has not changed for the components. In my opinion, this can be added to 8.0 and 2.7 directly as it basically retains existing behavior and makes it just more customizable. The added part is applying the HTML attribute e.g. `"theme"="warning"`.
And I think it is safe for the LTS version too as the added API is
1) In `Binder` very straight forward
2) In CSS it is based on `theme` attribute and the `ErrorLevel` from Java which has been as-is since Vaadin 8.0 and not likely to change.

(FYI @Artur- )

pleku:
Can be tested with eg. 
```java
@Route("binder-test")
public class BinderErrorLevelTest extends VerticalLayout {

    private class Bean {
        private String email;

        public String getEmail() {
            return email;
        }

        public void setEmail(String email) {
            this.email = email;
        }
    }

    private final TextField field;

    public BinderErrorLevelTest() {
        field = new TextField("empty:system, < 2 critical, < 4 error, < 6 warning, < 8 info");
        field.setValueChangeMode(ValueChangeMode.EAGER);
        field.setWidthFull();
        final Binder<Bean> binder = new Binder<>(Bean.class);
        binder.forField(field).withValidator((value, context) -> {
            if (value == null || value.length() == 0) {
                return ValidationResult.create("system", ErrorLevel.SYSTEM);
            }
            if (value.length() < 2) {
                return ValidationResult.create("critical", ErrorLevel.CRITICAL);
            } else if (value.length() < 4) {
                return ValidationResult.create("error", ErrorLevel.ERROR);
            } else if (value.length() < 6) {
                return ValidationResult.create("warning", ErrorLevel.WARNING);
            } else if (value.length() < 8) {
                return ValidationResult.create("info", ErrorLevel.INFO);
            }
            return ValidationResult.ok();
        }).bind("email");

        add(field);
    }
}
```

PR #11410, opened 2021-07-13
fix: avoid merge conflicts during properties collection

### LINKED ISSUE #11404: BeanPropertySet.get(...) fails on types that inherit the same interface multiple times
### Description of the bug
BeanPropertySet.get(...) fails on types that inherit the same interface multiple times.

```
  interface Iface3
    extends
      Iface2,
      Iface
  {
  }

  interface Iface2
    extends
      Iface
  {
  }

  interface Iface
  {
    Object getPropertyValue();
  }

  public static void main(String[] args)
    throws IntrospectionException
  {
    // lists the property twice
    BeanUtil.getBeanPropertyDescriptors(Iface3.class).stream().forEach(property -> {
      System.out.println(property.getName());
    });

    // fails because of the duplicate property
    BeanPropertySet.get(Iface3.class, false, PropertyFilterDefinition.getDefaultFilter());
  }

```

### Expected behavior
Bean properties declared multiple times in the hierarchy shouldn't cause problems.
### Actual behavior
Exception is thrown:
` java.lang.IllegalStateException: Duplicate key com.vaadin.flow.data.binder.BeanPropertySet$BeanPropertyDefinition@64cd705f`

### Versions:
    - Vaadin / Flow version: Vaadin 20
    - Java version: tested with 8 and 11
    - OS version: Windows

denis-anisimov:
fixes #11404

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/BeanPropertySet.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/BeanPropertySet.java
@@ -323,8 +323,8 @@ public class BeanPropertySet<T> implements PropertySet<T> {
                     .map(descriptor -> new BeanPropertyDefinition<>(this,
                             instanceKey.type, descriptor))
                     .collect(Collectors.toMap(PropertyDefinition::getName,
-                            Function.identity(), this::mergePropertyDefinitions,
-                            HashMap::new));
+                            Function.identity(),
+                            this::mergePropertyDefinitions));
         } catch (IntrospectionException e) {
             throw new IllegalArgumentException(
                     "Cannot find property descriptors for "
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/BeanPropertySet.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/BeanPropertySet.java
@@ -323,13 +323,31 @@ public class BeanPropertySet<T> implements PropertySet<T> {
                     .map(descriptor -> new BeanPropertyDefinition<>(this,
                             instanceKey.type, descriptor))
                     .collect(Collectors.toMap(PropertyDefinition::getName,
-                            Function.identity()));
+                            Function.identity(), this::mergePropertyDefinitions,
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/binder/BeanPropertySet.java
```diff
@@ -323,7 +323,8 @@ private BeanPropertySet(InstanceKey<T> instanceKey) {
                     .map(descriptor -> new BeanPropertyDefinition<>(this,
                             instanceKey.type, descriptor))
                     .collect(Collectors.toMap(PropertyDefinition::getName,
-                            Function.identity()));
+                            Function.identity(), this::mergePropertyDefinitions,
+                            HashMap::new));
```
mshabarov, line 327: Why do you specify `HashMap::new` here and not use overloaded method where it is used by default?

mshabarov (dismissed):

flow-data/src/main/java/com/vaadin/flow/data/binder/BeanPropertySet.java
denis-anisimov: Apparently I missed the method which doesn't require the last argument.

denis-anisimov: done

mshabarov (approved):

mshabarov:
@denis-anisimov I'm not sure is it an inconsistency, but if `BeanPropertySet.get` returns only one property definition for the considered example, then I would expect `BeanUtil.getBeanPropertyDescriptors` also returns only one descriptor for inherited beans.

denis-anisimov:
`BeanUtil` is a totally different util class in a different module.
If it has an issue then this is a totally different issue and it doesn't cause any exception.
The PR should address only one issue, not both.

PR #11416, opened 2021-07-14
fix: reset session field only when HTTP session is invalidated

### LINKED ISSUE #6959: Cannot invalidate HTTP session from SessionDestroyListener
In #6763 it's suggested that HttpSession can be invalidated from Vaadin SessionDestroyListener, however, when the listener is executed the wrapped session is null.

- Expected behavior:
  I should be able to invalidate the HttpSession after some period of inactivity.
- Actual behavior
  The HttpSession is not available after the Vaadin session has been destroyed.

Vaadin 14.0.11

[Example application](https://github.com/vaadin/flow/files/3851644/v14.zip)

The SessionDestroyListener is:
```
	public void serviceInit(ServiceInitEvent event) {
		event.getSource().addSessionDestroyListener(ev->{
			System.out.println("Vaadin session destroyed");
			VaadinSession vsession = ev.getSession();
			WrappedSession wsession = vsession.getSession();
			if (wsession==null) {
				System.out.println("wrapped session is null");
			} else {
				wsession.invalidate();
			}
		});
	}
```

denis-anisimov:
fixes #6959

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
@@ -205,11 +205,11 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
             // it as soon as we acquire the lock.
             service.fireSessionDestroy(this);
         }
-        // Vaadin session attribut is removed from HTTP session in two cases:
+        // Vaadin session attribute is removed from HTTP session in two cases:
         // either Vaadin session is closed explicitly by VaadinService or HTTP
         // session is invalidated (and all attributes are removed from it). In
         // the latter case the session field should be set to {@code null}
-        // Immediately to avoid using HTTP session object which is invalid (its
+        // immediately to avoid using HTTP session object which is invalid (all
         // methods throw exceptions).
         try {
             lock();
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
@@ -78,6 +78,9 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
     @Deprecated
     public static final String UI_PARAMETER = InitParameters.UI_PARAMETER;
 
+    static final String CLOSE_SESSION_EXPLICITLY = "vaadin.close-session."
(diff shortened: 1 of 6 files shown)


flow-server/src/test/java/com/vaadin/flow/server/VaadinServiceTest.java
```diff
@@ -416,6 +417,46 @@ public void createRequestHandlers_pwaHandlerIsInList_webComponentHandlersAreInLi
...
+            }
+        };
+
+        service.fireSessionDestroy(vaadinSession);
+
+        Assert.assertEquals(VaadinSessionState.CLOSED, stateRef.get());
+    }
+
+    @Test
+    public void removeFromHttpSession_setExplciitSessionCloseAttribute()
```
caalador, line 439:
```suggestion
    public void removeFromHttpSession_setExplicitSessionCloseAttribute()
```

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -202,7 +205,21 @@ public void valueUnbound(HttpSessionBindingEvent event) {
             // it as soon as we acquire the lock.
             service.fireSessionDestroy(this);
         }
-        session = null;
+        // Vaadin session attribut is removed from HTTP session in two cases:
+        // either Vaadin session is closed explicitly by VaadinService or HTTP
+        // session is invalidated (and all attributes are removed from it). In
+        // the latter case the session field should be set to {@code null}
+        // Immediately to avoid using HTTP session object which is invalid (its
+        // methods throw exceptions).
```
caalador, line 213:
```suggestion
        // Vaadin session attribute is removed from HTTP session in two cases:
        // either Vaadin session is closed explicitly by VaadinService or HTTP
        // session is invalidated (and all attributes are removed from it). In
        // the latter case the session field should be set to {@code null}
        // immediately to avoid using HTTP session object which is invalid (all
        // methods throw exceptions).
```

caalador (changes requested):
Some minor typos.

caalador (approved):

PR #11422, opened 2021-07-15
feat: make registerEndpoint public

### LINKED ISSUE #11421: Make EndpointRegistry.registerEndpoint() public
It would be great to be able to call the method [`registerEndpoint()`](https://github.com/vaadin/flow/blob/master/fusion-endpoint/src/main/java/com/vaadin/fusion/EndpointRegistry.java#L90) manually. If endpoints are bundled in dependencies, and the .ts are generated externally from the current project, then the endpoints can be registered in Spring, but they will not be taken into account by the `EndpointRegistry`. By calling this method from somewhere (ie: a `VaadinServiceInitListener `or something like that) they can be used.

mlopezFC:
## Description

This will allow manual endpoint registration for covering the case where endpoints are provided by dependencies outside the project.

Fixes #11421

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/fusion-endpoint/src/main/java/com/vaadin/fusion/EndpointRegistry.java
+++ b/fusion-endpoint/src/main/java/com/vaadin/fusion/EndpointRegistry.java
@@ -87,7 +87,13 @@ public class EndpointRegistry {
                 .orElse(beanType.getSimpleName());
     }
 
-    void registerEndpoint(Object endpointBean) {
+    /**
+     * Registers the specified endpoint.
+     * 
+     * @param endpointBean
+     *            the endpoint that is going to be added in the registry
+     */
+    public void registerEndpoint(Object endpointBean) {
         // Check the bean type instead of the implementation type in
         // case of e.g. proxies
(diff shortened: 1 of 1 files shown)


haijian-vaadin:
quote from @mlopezFC 

> This is not for a Vaadin customer, but for an addon that I'm building that will include some endpoints. The addon is already working because I'm using reflection, so no problems if this is not merged. Regarding the typescript generation, that is already covered because if I call the goal "build-frontend" of vaadin-maven-plugin, then the typescript  is generated and it's just a matter of some maven magic to move them to META-INF/resources/frontend and that's it. I think that producing ts sources in dependencies instead of producing them in the war file has the advantage that will reduce source code generation time and it's not necessary to carry the java sources around,

link to the addon: https://github.com/FlowingCode/AppLayoutAddon

Based on the information, we will close this PR, since we plan to work on the multi-module support, this would be part of it.

Thanks Martin for bringing this up.

PR #11447, opened 2021-07-20
fix: Don't revisit visited classes.

### LINKED ISSUE #9861: FrontendDependencies loses dependencies in some cases

Due to the way the class is written, `visitClass` can be called multiple times for the same class. The results returned by `visitClass` are different though because there is a class field `visitedClasses` tracking what has been visited. 

The first call to `visitClass` will visit the given class and all classes it refers to and gather data correctly.
The second call to `visitClass` for the same class will only visit that one class and not any classes it refers to. It will gather data only for that particular class.

So if you have class A that refers to B, and B contains e.g. `@JsModule` it will be found during the first call and not during the second.

This is not considered at all by the rest of the code, e.g.
`collectEndpoints` always overwrites any previous result with the new one (the one that is missing data): https://github.com/vaadin/flow/blob/790122df06e186e7ce2aa5a682dacd2ec2d0cdff/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java#L266-L270

`computeExporterEndpoints` always overwrites data with what it found (on a potential second scan): https://github.com/vaadin/flow/blob/790122df06e186e7ce2aa5a682dacd2ec2d0cdff/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java#L524

The broken behavior can be seen in `flow-tests/test-embedding/test-embedding-generic` by running
```
mvn flow:prepare-frontend flow:build-frontend
```
and looking at `target/classes/META-INF/VAADIN/config/flow-build-info.json`

It should not say
```
  "chunks": {
    "fallback": {
      "jsModules": [
...
        "@vaadin/flow-frontend/ConnectionIndicator.js"
      ],

```

As `@JsModule("@vaadin/flow-frontend/ConnectionIndicator.js")` is set on UI.java

### Versions:
    - Vaadin / Flow version: master

caalador:
Do not revisit visited classes as
visited childer will not be rechecked
and possibly overwriten.

Always visit UI once as else we get
UI JsModule in fallback.

Fixes #9861

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
@@ -535,25 +535,37 @@ public class FrontendDependencies extends AbstractDependenciesScanner {
 
         for (Class<?> exporter : exporterClasses) {
             String exporterClassName = exporter.getName();
-            EndPointData exporterData = new EndPointData(exporter);
-            exportedPoints.put(exporterClassName,
-                    visitClass(exporterClassName, exporterData, false));
+            if (!endPoints.containsKey(exporterClassName)) {
+                EndPointData exporterData = new EndPointData(exporter);
+                exportedPoints.put(exporterClassName,
+                        visitClass(exporterClassName, exporterData, false));
+            }
 
             if (!Modifier.isAbstract(exporter.getModifiers())) {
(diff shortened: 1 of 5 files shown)


denis-anisimov (commented):
Artur also mentions similar issue in the `computeExporterEndpoints` method.
This PR doesn't contain any changes related to this method and tests.

denis-anisimov (approved):

PR #11450, opened 2021-07-21
fix: Show only single permanent notification

### LINKED ISSUE #11074: VaadinDevmodeGizmo preserved on the page when @PreserveOnRefresh is used
Old VaadinDevmodeGizmo pop ups are preserved on the page when navigate to the route component annotated with `@PreserveOnRefresh`:

Screenshot 2021-05-25 at 17 18 06

[skeleton-starter-flow-spring-preserve.zip](https://github.com/vaadin/flow/files/6540033/skeleton-starter-flow-spring-preserve.zip)
1. Open and run the project.
2. Go to http://localhost:8080
3. Navigate to the child view by clicking on the navigate button
4. Navigate back to the main view
5. Notice that there are so many dev mode gizmo pop ups as many times you navigate to main view.

### Expected behavior
Single VaadinDevmodeGizmo popup
### Actual behavior
Multiple VaadinDevmodeGizmo popups
### Versions:
- Vaadin / Flow version: 21.0.0.alpha2 / 8.0.0.alpha3
- Java version: 15
- OS version: Mac OS
- Browser version (if applicable): Chrome 90
- IDE (if applicable): IntelliJ IDEA Ultimate

caalador:
If a persistent message is open
do not open a new permanent message
for the same persistentId.

Closes #11074

### DIFF
--- a/flow-client/src/main/frontend/VaadinDevmodeGizmo.ts
+++ b/flow-client/src/main/frontend/VaadinDevmodeGizmo.ts
@@ -808,6 +808,13 @@ export class VaadinDevmodeGizmo extends LitElement {
 
   showNotification(type: MessageType, message: string, details?: string, link?: string, persistentId?: string) {
     if (persistentId === undefined || !VaadinDevmodeGizmo.notificationDismissed(persistentId!)) {
+      // Do not open persistent message if another is already visible with the same persistentId
+      const matchingVisibleNotifications = this.notifications
+        .filter((notification) => notification.persistentId === persistentId)
+        .filter((notification) => !notification.deleted);
+      if (matchingVisibleNotifications.length > 0) {
+        return;
+      }
       const id = this.nextMessageId;
       this.nextMessageId += 1;
       this.notifications.push({


flow-client/src/main/frontend/VaadinDevmodeGizmo.ts
```diff
@@ -808,6 +808,13 @@ export class VaadinDevmodeGizmo extends LitElement {
 
   showNotification(type: MessageType, message: string, details?: string, link?: string, persistentId?: string) {
     if (persistentId === undefined || !VaadinDevmodeGizmo.notificationDismissed(persistentId!)) {
+      // Do not open persistent message if another is already visible with the same persistentId
+      const matchingVisibleNotifications = this.notifications
+        .filter((notification) => notification.persistentId === persistentId)
```
mshabarov, line 813: Does this make it so that it is possible to have several notification visible, but only one at maximum for each notification type/id, e.g. one for `preserveOnRefreshWarning` and one for `liveReloadUnavailable`, but not two of `liveReloadUnavailable` type?

caalador: Yes it only stops multiple notification with the same persistentId of showing so having different persistentId notifications visible at the same time still works.

mshabarov:
Tested this fix with the steps described in the ticket, using `21.0.0.alpha10` and flow `9.0-SNAPSHOT` built locally on top of `issues/11074_show_single_permanent_notification`.

PreserveOnRefresh warning message is still present in the gizmo, but this time it's being shown in the gizmo message list, not as a separate pop-up:

Screenshot 2021-07-22 at 11 37 19

This is not the behaviour I expected (I expected only one POR message regardless of navigation count), but IMO it can be kept as is. For POR it's not that informative, because POR is always present since the application starts. But there can be message types in the future for whose it's worth to show a new message when some event is happening. What do you think, @caalador ?

caalador:
Well this fix was for notifications that stacked. Also if the notification is not closed no message should be added as notifications are added to messages only when the are dismissed-

Messages would probably need something else as they are not ID'd in any way and just add to the messages list and should those need blocking it would be another issue.

caalador:
Now I see that is you mark it as `don't show again` it will keep on adding to the message log. I'll add a fix for that so they don't stack there.

mshabarov (approved):

PR #11462, opened 2021-07-23
chore: update test log check

caalador:
Update the way the test checks for
log output.


denis-anisimov (changes requested):
I've been working on the same thing.

I don't see any reliable way to check logs safely.

The test should not check logs at all. It should just check that `validateToolVersion` doesn't throw and nothing else.

PR #11464, opened 2021-07-23
fix: update mock logging

caalador:
Update the log check to not use
a file as it will most likely bring in
problems in concurrent execution


denis-anisimov (approved):
Very good.

👍

PR #11471, opened 2021-07-26
fix: register data provider listener only if it has not been registered

TatuLund:
fixes: https://github.com/vaadin/flow-components/issues/1914


denis-anisimov (commented):
This looks good but I spent some time to understand why it's needed.
From the code I see paired `handleDetach` method which do the clean up.
So assuming the expected workflow `detach` should always happen after `attach` and this check should not be needed at all.

The problem is in the method names which are uses in two cases:
- when the node is attached/detached (this is a normal expected workflow). And method names are fine for this situation.
- when data provider is set: `handleDetach` and then `handleAttach` are called .

In the second case the method names are not suitable. Since the same methods are used in two totally different cases they should be renamed.

This is one thing which I would like to see changed here.

Another : it's still not obvious that `setDataProvider` resets the listener in the same way as attach/detach does.
So even though the `null` check is quite safe change I would like to see a unit test which will allow to prevent introducing such regression back.

Legioth:
>This looks good but I spent some time to understand why it's needed.

That's because the PR doesn't contain any tests

Legioth:
The conventional way of dealing with cases like this is to `remove()` the old registration if one exists before writing a new value to the field, instead of assuming that the old registration can be reused. That approach is slightly more robust since it would also cover the case of changing from one data provider to another without "detach" in between - this isn't present in the code today but there's no guarantee that it won't occur as a result of future changes.

TatuLund:
@Legioth , good point, I improved my solution.

mshabarov:
@TatuLund I'm going to add a test for this fix to move this important change forward, so if you currently are working on it also, please let me know :)

TatuLund:
@mshabarov Please do, I have not figured out good test yet.

denis-anisimov (approved):

PR #11481, opened 2021-07-27
fix: remove extra assertion

### LINKED ISSUE #11480: InvalidateHttpSessionIT::invalidateHttpSession_vaadinSessionIsClosed sometimes fails
InvalidateHttpSessionIT.invalidateHttpSession_vaadinSessionIsClosed[ANY_Chrome_]
 
The failure is caused by the assertion error:

```
Caused by: org.junit.ComparisonFailure: expected:<node01[kpjv2lj8jwnu17545l4eubki9235]-default> but was:<node01[65jhgodhllnfuwi3bzrh59zs234]-default>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at com.vaadin.flow.uitest.ui.InvalidateHttpSessionIT.invalidateHttpSession_vaadinSessionIsClosed(InvalidateHttpSessionIT.java:39)
```

See screenshot:

invalidateHttpSession_vaadinSessionIsClosed ANY_Chrome_ (com vaadin flow uitest ui InvalidateHttpSessionIT)

This specific failure happened in OSGi build validation but it should not be related to OSGi and it's common failure.
It may happen in regular build as well.

denis-anisimov:
fixes #11480

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/InvalidateHttpSessionIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/InvalidateHttpSessionIT.java
@@ -27,8 +27,6 @@ public class InvalidateHttpSessionIT extends ChromeBrowserTest {
     public void invalidateHttpSession_vaadinSessionIsClosed() {
         open();
 
-        String sessionId = findElement(By.id("current-session-id")).getText();
-
         findElement(By.id("invalidate-session")).click();
 
         waitForElementPresent(By.id("invalidated-session-id"));
@@ -36,7 +34,7 @@ public class InvalidateHttpSessionIT extends ChromeBrowserTest {
         String invalidatedSessionId = findElement(
                 By.id("invalidated-session-id")).getText();
 
-        sessionId = findElement(By.id("current-session-id")).getText();
+        String sessionId = findElement(By.id("current-session-id")).getText();
         Assert.assertNotEquals(sessionId, invalidatedSessionId);
     }
 
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/InvalidateHttpSessionIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/InvalidateHttpSessionIT.java
@@ -36,8 +36,6 @@ public class InvalidateHttpSessionIT extends ChromeBrowserTest {
         String invalidatedSessionId = findElement(
                 By.id("invalidated-session-id")).getText();
 
-        Assert.assertEquals(sessionId, invalidatedSessionId);
-
         sessionId = findElement(By.id("current-session-id")).getText();
         Assert.assertNotEquals(sessionId, invalidatedSessionId);
     }


flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/InvalidateHttpSessionIT.java
```diff
@@ -36,8 +36,6 @@ public void invalidateHttpSession_vaadinSessionIsClosed() {
         String invalidatedSessionId = findElement(
                 By.id("invalidated-session-id")).getText();
 
-        Assert.assertEquals(sessionId, invalidatedSessionId);
-
         sessionId = findElement(By.id("current-session-id")).getText();
         Assert.assertNotEquals(sessionId, invalidatedSessionId);
```
mshabarov, line 40: Would it better to compare current session (the new one, after invalidation) and the initial one?
I concern that if `Assert.assertEquals(sessionId, invalidatedSessionId);` fails sometimes, then this one also will fail sometimes:
`Assert.assertNotEquals(sessionId, invalidatedSessionId);`
Thus, comparing with `invalidatedSessionId` will potentially cause random failures.

denis-anisimov: I don't really understand what you suggest.
The last assertion is the subject of the test.
It may not fail if there is no bug: the new session after invalidation may not be the same as it was before invalidation.
So if the assertion fails then something needs to be fixed in the Flow code. 
The current change is just a test correction and nothing else.

I don't know the exact reason of the failure but it seems the page is just refreshed after the session invalidated so the previous session is the same as after the refresh/reload since it doesn't invalidate the session.

But it doesn't matter what happens there : refresh/reload or whatever. The current/invalidated session may not be the same as a previous one. This is what assertion tests.
Everything else is far-fetched.

mshabarov: Ok, I didn't look into the corresponding View implementation before review.
Invalidated session id is being recorded via session destroy listener, which execution is most likely one of the goal of this test to check.
Assuming that, this check looks good:
`Assert.assertNotEquals(sessionId, invalidatedSessionId);`

If it fails, then the listener or session invalidation do not work apparently.

As you deleted this assertion, the line 30 in not needed anymore and can be removed:
`String sessionId = findElement(By.id("current-session-id")).getText();`

mshabarov (approved):

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/InvalidateHttpSessionIT.java
denis-anisimov: done

mshabarov (approved):

PR #11482, opened 2021-07-27
fix: don't override web.xml, annotation config with classpath flow-build-info.json settings

### LINKED ISSUE #11479: Build info on classpath overrides parameters set in web.xml
When using DevModeInitializer to get classes from annotations, a flow-build-info.json found on the classpath will override parameters set in web.xml.

Here's the code in question (DeploymentConfigurationFactory.java): 

```
  protected static Properties createInitParameters(
              Class<?> systemPropertyBaseClass, VaadinConfig vaadinConfig)
              throws VaadinConfigurationException {
          Properties initParameters = new Properties();
          readUiFromEnclosingClass(systemPropertyBaseClass, initParameters);
          readConfigurationAnnotation(systemPropertyBaseClass, initParameters);
  
          // Read default parameters from server.xml
          final VaadinContext context = vaadinConfig.getVaadinContext();
          for (final Enumeration<String> e = context.getContextParameterNames(); e
                  .hasMoreElements();) {
              final String name = e.nextElement();
              initParameters.setProperty(name, context.getContextParameter(name));
          }
  
          // Override with application config from web.xml
          for (final Enumeration<String> e = vaadinConfig
                  .getConfigParameterNames(); e.hasMoreElements();) {
              final String name = e.nextElement();
              initParameters.setProperty(name,
                      vaadinConfig.getConfigParameter(name));
          }
  
          readBuildInfo(systemPropertyBaseClass, initParameters,
                  vaadinConfig.getVaadinContext());  // this overrides config from web.xml, for example the frontend folder setting
          return initParameters;
      }

```

Create two dependent projects in eclipse, build one then try to build the other. DevModeInitializer will override parameters in the second project if the flow-build-info.json for the first one is on the classpath. 

### Expected behavior

web.xml parameters should override any existing flow-build-info.json on the classpath

### Actual behavior
parameters from an existing flow-build-info.json on the classpath override those set in web.xml

If this is intended behavior, feel free to close. Otherwise, I can submit a PR changing the order, because I think that might make more intuitive sense for users. Right now, we have to delete the flow-build-info.json for our dependent, standalone Vaadin project so that we can compile a second project that uses the first as a dependency. 

### Versions:
    - Vaadin / Flow version: 14.6.5
    - Java version: OpenJDK 8
    - OS version: Manjaro
    - Application Server (if applicable): Jetty
    - IDE (if applicable): Eclipse

rudiejd:
## Description

flow-server was overriding parameters set in web.xml/annotations with those found in *any flow-build-info.json* on the classpath. Now it only overrides parameters which weren't already set. Fixes #11479.

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
@@ -124,9 +124,9 @@ public class DeploymentConfigurationFactory extends AbstractConfigurationFactory
             Map<String, String> properties = getConfigParametersUsingTokenData(
                     buildInfo);
             // only insert properties that haven't been defined
-            for (String p : properties.keySet()) {
-                if (!initParameters.containsKey(p)) {
-                    initParameters.put(p, properties.get(p));
+            for (Map.Entry<String, String> e : properties.entrySet()) {
+                if (!initParameters.containsKey(e.getKey())) {
+                    initParameters.put(e.getKey(), e.getValue());
                 }
             }
             fallbackChunk = FrontendUtils.readFallbackChunk(buildInfo);
--- a/flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
@@ -123,8 +123,12 @@ public class DeploymentConfigurationFactory extends AbstractConfigurationFactory
             JsonObject buildInfo = JsonUtil.parse(json);
             Map<String, String> properties = getConfigParametersUsingTokenData(
                     buildInfo);
-            initParameters.putAll(properties);
-
+            // only insert properties that haven't been defined
+            for (String p : properties.keySet()) {
+                if (!initParameters.containsKey(p)) {
+                    initParameters.put(p, properties.get(p));
+                }
+            }
             fallbackChunk = FrontendUtils.readFallbackChunk(buildInfo);
         }
         if (fallbackChunk == null) {
--- a/flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
@@ -5,6 +5,7 @@ import javax.servlet.ServletContext;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.file.Files;
@@ -13,9 +14,12 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.ArrayList;
 
 import org.apache.commons.io.FileUtils;
 import org.easymock.Capture;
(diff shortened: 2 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/DeploymentConfigurationFactory.java
```diff
@@ -123,8 +123,12 @@ private void readBuildInfo(Properties initParameters,
             JsonObject buildInfo = JsonUtil.parse(json);
             Map<String, String> properties = getConfigParametersUsingTokenData(
                     buildInfo);
-            initParameters.putAll(properties);
-
+            // only insert properties that haven't been defined
+            for (Map.Entry<String, String> e : properties.entrySet()) {
```
denis-anisimov, line 127: Don't use one letter variable.
`entry` instead of `e` would be better here.

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
```diff
@@ -330,6 +330,29 @@ public void createInitParameters_valuesFromContextAreIgnored_valuesAreTakenFroms
         Assert.assertFalse(parameters.contains("bar"));
     }
 
+    @Test
+    public void createInitParameters_valuesFromTokenAreIgnored_valuesAreTakenFromservletConfig()
```
denis-anisimov, line 334: `valuesFromTokenAreIgnored`:
the second part of the name is usually precondition.
So it should say that properties are provided via init parameters and token file at the same time.
And the last part is a result : `valuesFromTokenFileOverridenByInitParams` e.g.

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
```diff
@@ -330,6 +330,29 @@ public void createInitParameters_valuesFromContextAreIgnored_valuesAreTakenFroms
...
+        DeploymentConfigurationFactory factory = new DeploymentConfigurationFactory();
+
+        VaadinConfig config = mockTokenFileViaContextParam(
+                "{ '" + SERVLET_PARAMETER_PRODUCTION_MODE + "': true}");
+        Mockito.when(config.getConfigParameterNames())
+                .thenReturn(Collections.enumeration(
+                        Arrays.asList(SERVLET_PARAMETER_PRODUCTION_MODE,
+                                FrontendUtils.PARAM_TOKEN_FILE)));
+        Mockito.when(
+                config.getConfigParameter(SERVLET_PARAMETER_PRODUCTION_MODE))
```
denis-anisimov, line 345: That checks only one parameter.
 Don't know how hard it is but would be better to check every init parameter.
I think the values has only  two types: boolean and String.

So it should be possible to make to arrays of properties and go through these arrays and do the check

denis-anisimov (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
rudiejd: I added another boolean parameter, and added a string parameter so we can capture both types. I also added a parameter that isn't overridden by the servlet config.

rudiejd: should I do *all* init parameters? wasn't sure whether that is necessary since the behavior is the same

denis-anisimov: I would recommend to do it for all init parameters, yes.
It might be that the code may be modified so it works properly for most of parameters but is broken for one.

What I suggest is : split parameter names into two complementary set . One set contains parameters whose type is String and another with boolean type.
`InitParameters` class can be used to get all parameter names via reflection.
Then just use a couple of unit test methods: go through all the parameters and make a check.

rudiejd: okay, I think I'll just be hardcoding a set of string parameters (not sure how else to tell if value is a string, can't use getAssignableType or anything) and getting the boolean ones through reflection, then making assertions for each

flow-server/src/test/java/com/vaadin/flow/server/DeploymentConfigurationFactoryTest.java
```diff
@@ -335,38 +339,77 @@ public void createInitParameters_valuesAreTakenFromservletConfigAndTokenFile_val
...
+            if (i < initParamFields.length - 1) {
+                mockTokenJsonString += ",";
+            }
+        }
+        mockTokenJsonString += " }";
+        VaadinConfig config = mockTokenFileViaContextParam(mockTokenJsonString);
+        List<String> allParamsList = new ArrayList<String>();
+        for (int i = 0; i < initParamFields.length; i++) {
+            String paramName = (String) initParamFields[i]
+                    .get(new InitParameters());
```
denis-anisimov, line 376: The field is static so there is no need to instantiate an object. `null` can be passed as an arg here.
But it's minor.

denis-anisimov (changes requested):
Since there is an important request: add message to every assertion with a parameter name I would also like to ask to refactor the code to extract cycles which initializes mocks into separate methods to improve readability: the test itself may be very short if it just calls initializer methods.

rudiejd:
> Since there is an important request: add message to every assertion with a parameter name I would also like to ask to refactor the code to extract cycles which initializes mocks into separate methods to improve readability: the test itself may be very short if it just calls initializer methods.

I have refactored the generation of the token and the mocking into two different private methods. I tried to make them a bit generic to encourage reuse for similar test cases

rudiejd:
https://teamcity.vaadin.com/viewLog.html?buildId=85356&buildTypeId=Flow_FlowTests
can someone tell me why run validation is failing and/or re run? everything's passing locally

denis-anisimov:
`flow-client` failed with unknown reason.
I've restarted the build.

rudiejd:
@denis-anisimov same thing again. the client unit/integration tests run locally though, so i can't figure out if i broke something... it doesn't show any test as failing on teamcity either

denis-anisimov (approved):

denis-anisimov:
There were some issues with builds.
Now they should be fixed.
At least `flow-client` validation has passed.

Issue #11502, opened 2021-07-30
Add public method to check whether navigation target has mandatory parameters


TatuLund:
Consider following use case.

I have complex application with many routes and they configuration increases over time as developers add new views. I want to have automatically generated menu in my application. It is rather easy to do auto generated menu with Vaadin.

```
    private void populateMenu(MenuType type) {
        RouteRegistry reg = SessionRouteRegistry
                .getSessionRegistry(VaadinSession.getCurrent());
        List<RouteData> routes = reg.getRegisteredRoutes();
        routes.forEach(route -> {
            PageTitle title = route.getNavigationTarget()
                    .getAnnotation(PageTitle.class);
            String titleString = "";
            if (title != null) {
                titleString = title.value();
            } else {
                titleString = route.getNavigationTarget().getTypeName();
            }
            RouterLink link = new RouterLink(titleString , navigationTarget);
            addToDrawer(link); // For example if I am using AppLayout
        });
    }
```

Now this approach has a shortcoming. If I have navigationTarget that requires mandatory parameters, I should be able to check that as RouterLink creation will throw an exception if mandatory parameter is needed, one should use alternative constructor where you can give e.g. default parameter.

This is however not possible easily as such methods `HasUrlParameterFormat ` are package protected.

As a workaround it is possible to introduce custom annotation or other mechanism for bookeeping such urls.

PR #11503, opened 2021-07-30
documentation: Add missing throws in JavaDocs [skip ci]

TatuLund:
See RouteConfiguration#getUrl for reference

see: https://github.com/vaadin/flow/issues/11502

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/RouterLink.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/RouterLink.java
@@ -131,7 +131,7 @@ public class RouterLink extends Component implements HasText, HasComponents,
      *             if navigation target requires parameters
      */
     public RouterLink(Router router, String text,
-            Class<? extends Component> navigationTarget) {
+            Class<? extends Component> navigationTarget) throws IllegalArgumentException {
         this(router, text, navigationTarget, RouteParameters.empty());
     }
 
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/router/RouterLink.java
```diff
@@ -127,6 +127,8 @@ public RouterLink(String text, Class<? extends Component> navigationTarget,
      *            link text
      * @param navigationTarget
      *            navigation target
+     * @throws IllegalArgumentException
```
denis-anisimov, line 130: `throws IllegalArgumentException` could be added to the constructors to show explicitly that it may be thrown.

Also please add `throws IllegalArgumentException` to  `RouterLink::setRoute`

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/RouterLink.java
TatuLund: > Also please add throws IllegalArgumentException to RouterLink::setRoute

I did that already

denis-anisimov: No.

```
public RouterLink(Router router, String text,
            Class<? extends Component> navigationTarget) throws IllegalArgumentException`

TatuLund: Done

denis-anisimov (approved):

denis-anisimov:
Formatter validation fails.
Please apply formatter.

denis-anisimov (approved):

Issue #10084, opened 2021-02-18
`npx pnpm` fails on Windows if home directory contains space


joheriks:
As  `Windows User` on Windows with home directory `C:\Users\Windows User`, when Flow is fetching a compatible pnpm version via `npx` (using a command prefixed by `npx pnpm ...`) it fails with error `EPERM: operation not permitted, mkdir 'C:\Users\Windows'`. This seems to be due to an issue with npm 6 not accepting spaces in the home directory:
https://stackoverflow.com/questions/55450386/npm-eperm-mkdir-operation-not-permitted-on-windows

**Note: fixed in npm 7 (node.js 15+) (EDIT:) which is installed by default by Flow 8.0 (V21)**. Installing Node.js 15+ globally fixes the issue.

Versions:
- Vaadin 19.0.0.beta3
- Windows 10
- npm 6.14.8 
- node v14.15.5

johannest:
Force upgrading to the latest npm 7 through PowerShell with following commands (https://stackoverflow.com/questions/18412129/how-can-i-update-npm-on-windows) helped me:
```
Set-ExecutionPolicy Unrestricted -Scope CurrentUser -Force
npm install -g npm-windows-upgrade
npm-windows-upgrade
```

johannest:
In case the issue manifests itself with `Found too old 'pnpm'. If installed into the project 'node_modules', upgrade 'pnpm' to at least 5.0.0`

Then the above workaround is not enough, but you might need to 
1. Run `npm config set cache`
2. And run `npm config set cache`

pleku:
Because Vaadin 20 is still coming with Node.js 14 + npm 6 and we are not going to update that on the fly for Vaadin 20, we should probably do something about this for Vaadin 20. There is another issue automatically upgrading the node.js version when updating Vaadin versions #9864.

Thus it might make sense in the meanwhile to 
a) detect windows + space in the home folder
b) check that npm version 6 or older
And fail with a clear message that tells them to update their node.js version and why it should be updated.

Haprog:
> And fail with a clear message that tells them to update their node.js version and why it should be updated.

I don't think you need to update Node.js. Updating `npm`  should be enough.

johannest:
>And fail with a clear message that tells them to update their node.js version and why it should be updated.

It is unfortunately not enough to upgrade the node version. I have v14.7.0, and this is failing to 
[...]

PR #11517, opened 2021-08-04
fix: Fail-fast if the path to npm cache contains whitespaces

mshabarov:
## Description

Checks that the npm version accepts whitespaces in user home path and explains how to fix that otherwise.

Fixes https://github.com/vaadin/flow/issues/10084

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -86,10 +86,27 @@ public class FrontendTools {
             + FrontendUtils.DISABLE_CHECK //
             + MSG_SUFFIX;
 
+    private static final String USER_HOME_CONTAINS_WHITESPACES = MSG_PREFIX
+            + "%nYour Windows user home directory path contains whitespaces, and the currently installed npm"
+            + "%nversion (%s) doesn't accept this."
+            + "%nPlease exclude whitespaces from your user home path or upgrade npm version to 7 (or newer) by:"
+            + "%n 1) Running 'npm-windows-upgrade' tool with Windows PowerShell:"
+            + "%n        Set-ExecutionPolicy Unrestricted -Scope CurrentUser -Force"
+            + "%n        npm install -g npm-windows-upgrade"
+            + "%n        npm-windows-upgrade"
+            + "%n 2) Manually installing a newer version of npx: npm install -g npx"
+            + "%n 3) Manually installing a newer version of pnpm: npm install -g pnpm"
+            + "%n 4) Deleting the following files from your Vaadin project's folder (if present):"
+            + "%n        node_modules, package-lock.json, webpack.generated.js, pnpm-lock.yaml, pnpmfile.js"
+            + MSG_SUFFIX;
+
     private static final List<FrontendVersion> NPM_BLACKLISTED_VERSIONS = Arrays
             .asList(new FrontendVersion("6.11.0"),
                     new FrontendVersion("6.11.1"),
                     new FrontendVersion("6.11.2"));
+
+    private static final FrontendVersion WHITESPACE_ACCEPTING_NPM_VERSION = new FrontendVersion(
+            7, 0);
 
     private static final int SUPPORTED_NODE_MAJOR_VERSION = 10;
     private static final int SUPPORTED_NODE_MINOR_VERSION = 0;
@@ -390,14 +407,10 @@ public class FrontendTools {
         }
 
         try {
-            List<String> npmVersionCommand = new ArrayList<>(
-                    getNpmExecutable(false));
-            npmVersionCommand.add("--version"); // NOSONAR
-            FrontendVersion npmVersion = FrontendUtils.getVersion("npm",
-                    npmVersionCommand);
-            FrontendUtils.validateToolVersion("npm", npmVersion,
+            FrontendVersion foundNpmVersion = getNpmVersion();
+            FrontendUtils.validateToolVersion("npm", foundNpmVersion,
                     SUPPORTED_NPM_VERSION, SHOULD_WORK_NPM_VERSION);
-            checkForFaultyNpmVersion(npmVersion);
+            checkForFaultyNpmVersion(foundNpmVersion);
         } catch (UnknownVersionException e) {
             getLogger().warn("Error checking if npm is new enough", e);
         }
@@ -463,6 +476,41 @@ public class FrontendTools {
                     "by updating your global npm installation with `npm install -g npm@latest`");
             throw new IllegalStateException(badNpmVersion);
         }
+    }
+
+    /**
+     * Checks whether it's fine to have a whitespaces in user home path, and, if
+     * not, fails-fast with a message to developer how to fix it.
+     *
+     * @throws IllegalStateException
+     *             if <code>pnpm install</code> is blocked by whitespaces in
+     *             user home
+     */
+    void checkNpmAcceptsWhitespacesInUserHome() {
+        if (FrontendUtils.isWindows() && userHomeContainsWhitespaces()) {
+            try {
+                FrontendVersion foundNpmVersion = getNpmVersion();
+                // npm < 7.0.0 doesn't accept whitespaces in user home
+                boolean accepted = FrontendUtils.isVersionAtLeast(
+                        foundNpmVersion, WHITESPACE_ACCEPTING_NPM_VERSION);
+                if (!accepted) {
+                    throw new IllegalStateException(
+                            String.format(USER_HOME_CONTAINS_WHITESPACES,
+                                    foundNpmVersion.getFullVersion()));
+                }
+            } catch (UnknownVersionException e) {
+                getLogger().warn(
+                        "Error checking if npm supports whitespaces in Windows user home path",
+                        e);
+            }
+        }
+    }
+
+    private FrontendVersion getNpmVersion() throws UnknownVersionException {
+        List<String> npmVersionCommand = new ArrayList<>(
+                getNpmExecutable(false));
+        npmVersionCommand.add("--version"); // NOSONAR
+        return FrontendUtils.getVersion("npm", npmVersionCommand);
     }
 
     private File getExecutable(String dir, String location) {
@@ -723,5 +771,11 @@ public class FrontendTools {
         } else {
             return getNodeExecutable();
         }
+    }
+
+    private boolean userHomeContainsWhitespaces() {
+        String userDirectoryPath = FileUtils.getUserDirectoryPath();
+        return userDirectoryPath != null
+                && userDirectoryPath.matches(".*[\\s+].*");
     }
 }
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -521,9 +521,7 @@ public class FrontendTools {
             npmCacheCommand.add("cache");
             npmCacheCommand.add("--global");
             String output = FrontendUtils.executeCommand(npmCacheCommand);
-            getLogger().info("Command execution output: " + output);
             output = parseCommandOutput(output);
-            getLogger().info("Command execution output after parsing: " + output);
             if (output.isEmpty()) {
                 throw new IllegalStateException(
                         String.format("Command '%s' returned an empty path",
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -518,19 +518,14 @@ public class FrontendTools {
      * installed npm.
      *
      * @return the file object representing path to npm cache directory.
-     * @throws IOException
-     *             if an I/O occurs while getting npm cache directory.
-     * @throws InterruptedException
-     *             if the current thread is being interrupted while getting the
-     *             npm cache directory.
      * @throws CommandExecutionException
      *             if getting the npm cache directory completes exceptionally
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -465,6 +478,41 @@ void checkForFaultyNpmVersion(FrontendVersion npmVersion) {
...
 
+    /**
+     * Checks whether it's fine to have a whitespaces in user home path, and, if
+     * not, fails-fast with a message to developer how to fix it.
+     *
+     * @throws IllegalStateException
+     *             if <code>pnpm install</code> is blocked by whitespaces in
+     *             user home
+     */
+    void checkNpmAcceptsWhitespacesInUserHome() {
```
denis-anisimov, line 489: `throws IllegalStateException`

denis-anisimov: Why is it user home only ?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -314,8 +314,12 @@ private void runNpmInstall() throws ExecutionFailedException {
             if (requireHomeNodeExec) {
                 tools.forceAlternativeNodeExecutable();
             }
-            executable = enablePnpm ? tools.getPnpmExecutable()
-                    : tools.getNpmExecutable();
+            if (enablePnpm) {
+                tools.checkNpmAcceptsWhitespacesInUserHome();
```
denis-anisimov, line 318: Why user home is so specific in this case ?

And quite related question: why tools should be aware of this specific case (if any) ?
User home is just a directory which has no any specific meaning for `FrontendTools`. So the existence of such method is very weird.
 `FrontendTools` may have method  `checkNpmAcceptsWhitespacesInDir` or just `checkNpmAcceptsWhitespaces` and accept a File as a parameter.

denis-anisimov: Another question: should the method throw or just return a `boolean` ?
I can't answer for sure to this question but it might be that the caller of the method should decide how to handle the case of spaces in the path instead of the callee : the callee  defines only one hardcoded handler but the every caller may decide how to handle this instead only one way.

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
```diff
@@ -497,6 +504,127 @@ public void getSuitablePnpm_tooOldVersionInstalledAndSkipVersionCheck_accepted()
...
+        Assume.assumeTrue(
+                "This test is only for Windows, as the whitespaces in user "
+                        + "home issue reproduced only on Windows",
+                FrontendUtils.isWindows());
+
+        String originalUserHome = System.getProperty(USER_HOME);
+        try {
+            // given
+            // Whitespace in user name
+            System.setProperty(USER_HOME, "C:\\Users\\Windows User");
```
denis-anisimov, line 519: quite dangerous test since it modifies globally user home and any other test which relies or modifies the same property may fail itself or fail this test.
Though I don't know how make it better.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
```diff
@@ -603,6 +608,52 @@ public void runPnpmInstall_frameworkCollectedVersionNewerThanPinned_installedOve
                 overlayPackage.getString("version"));
     }
 
+    @Test
+    public void runNpmInstall_getPnpmExecutable_whitespacesInUserHomeChecked()
```
denis-anisimov, line 612: `npm` ?
It this test needed at all ?
`TaskRunPnpmInstallTest` extends `TaskRunNpmInstallTest` so it runs all its methods.
If this test should tests _PNPM_ then it should be renamed.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
mshabarov: > `throws IllegalStateException`

Not sure I got this comment. Exception description is added to javadoc and methods signature doesn't require it's declaration.

mshabarov: > Why is it user home only ?

Because the original issue description and relevant SO link to npm issue tell about problems exactly with whitespaces in user name (=> user home) reported by developers, see https://github.com/zkat/npx/issues/146.
Npm cache directory is located in app data folder inside user home directory by default, so if user name contains spaces - it fails.
However, you right that the method should take the path as a parameter and check it.
And it's better to check the path to npm cache folder instead of user nome, since the cache folder can be customised and contains whitespaces as well.
Then, the `TaskRunNpmInstall` should call this method and give a path to npm cache directory, provided by `FrontendTools`.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
mshabarov: Yes, I agree. I'm going to improve it as follows:
- rename the method to be not connected to user home, but receive the `File` to be checked.
- throw the exception on the `TaskRunNpmInstall`, because it is a caller code and it should decide how to handle this.

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
mshabarov: I'm going to check npm cache folder instead of user home, it's technically more precise. So, this one won't be needed soon and the npm cache getter can be stubbed then.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
mshabarov: The test method is about PNPM. I'll rename it.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
denis-anisimov: > Exception description is added to javadoc and methods signature doesn't require it's declaration.

Javadocs also are not required for the exception which is `RuntimeException`.
It's optional.
And the last thing which I do when I use a method: read javadocs and almost never check the exceptions in the javadocs.
Ideally there should not be any unchecked exceptions: everything should be checked. Then every such exception has to be declared in the signature.
The reality in our code is different.
But if you mention unchecked exception in the javadocs then it's a good thing to add throws in the method declaration.
Because I will not even be aware of the exception when I use a method: I never read javadocs. But I can check the method signature and see that there is a `throws` and then this may be an indicator that I there are cases when it's thrown and this is the only reason why I can find javadocs description for `throws`.

Shortly : javadocs `throws` is useless without `throws` in the method declaration. Even `throws` in the method declaration for unchecked exception is quite useless since it doesn't force me to do anything about it but at least it may give a hint . Javadocs `throws` doesn't give anything if you are not aware of it.

mshabarov:
`TaskRunPnpmInstallTest` and `FrontendToolsTest` tests successfully run on Windows agent and shows the message as expected.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -465,6 +465,81 @@ void checkForFaultyNpmVersion(FrontendVersion npmVersion) {
...
 
+    /**
+     * Checks whether it's fine to have a whitespaces in user home path, and, if
+     * not, fails-fast with a message to developer how to fix it.
+     *
+     * @throws IllegalStateException
+     *             if <code>pnpm install</code> is blocked by whitespaces in
+     *             user home
+     */
+    boolean npmAcceptsWhitespaces(File dirToBeChecked)
```
denis-anisimov, line 476: The method name is confusing.
I don't see how the parameter is related to the method: it looks like the method doesn't require any parameter.
The method says something about npm and it's not clear how the parameter is used in the context.

denis-anisimov: By the way : whitespace is just one of possible checks.
There is no reason to have this word in the name.
The method could have name `folderIsAllowedByNpm` or something like this.
The impl may include other checks.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -724,4 +799,11 @@ private String doGetNodeExecutable() {
             return getNodeExecutable();
         }
     }
+
+    private String parseCommandOutput(String output) {
```
denis-anisimov, line 803: parse ?
What kind of parsing?
This is rather a conversation or just a "normalization"

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -966,21 +966,66 @@ public UnknownVersionException(String tool, String extraInfo,
...
-                        "Using command " + String.join(" ", versionCommand));
+                throw new CommandExecutionException(exitCode);
             }
-            String output = streamToString(process.getInputStream());
-            return new FrontendVersion(parseVersionString(output));
-        } catch (InterruptedException | IOException e) {
-            throw new UnknownVersionException(tool,
-                    "Using command " + String.join(" ", versionCommand), e);
+            return streamToString(process.getInputStream());
+        } catch (Exception e) {
```
denis-anisimov, line 1027: Please never catch `Exception` unless there are serious well understood reasons for this.
There are no reasons to catch `Exception` here: only exceptions related to the process execution may be caught.
Everything else indicate a problem in the code which may no be thrown as a `CommandExecutionException`.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -314,8 +331,16 @@ private void runNpmInstall() throws ExecutionFailedException {
             if (requireHomeNodeExec) {
                 tools.forceAlternativeNodeExecutable();
             }
-            executable = enablePnpm ? tools.getPnpmExecutable()
-                    : tools.getNpmExecutable();
+            if (enablePnpm) {
```
denis-anisimov, line 334: So it does matter only when Pnpm is used and the check can be skipped in case of NPM ?

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
```diff
@@ -31,16 +32,20 @@
...
 import elemental.json.JsonObject;
 
 import static com.vaadin.flow.server.Constants.PACKAGE_JSON;
+import static com.vaadin.flow.testutil.FrontendStubs.createStubNode;
 
 public class TaskRunPnpmInstallTest extends TaskRunNpmInstallTest {
 
     private static final String PINNED_VERSION = "3.2.17";
 
+    private static final String USER_HOME = "user.home";
```
denis-anisimov, line 47: Is this still needed ?

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -465,6 +465,81 @@ void checkForFaultyNpmVersion(FrontendVersion npmVersion) {
         }
     }
 
+    /**
+     * Checks whether it's fine to have a whitespaces in user home path, and, if
```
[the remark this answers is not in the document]
mshabarov: Fixed

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
mshabarov: Renamed to `folderIsAcceptableByNpm`

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
mshabarov: Yes, it's not a parsing, renamed to `removeLineBreaks`.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
mshabarov: Good comment! I reworked it in a way to handle all the checked exceptions in the caller's code (`TaskRunNpmInstall`), it ignores the whole check if something goes wrong and just logs it.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
mshabarov: Yes. The error occurs when `pnpm` is installed via `npx`.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
mshabarov: no. Removed

denis-anisimov (approved):

PR #11520, opened 2021-08-05
fix: Prevent deadlock in findOrCreateVaadinSession

TatuLund:
It is possible findOrCreateVaadinSession leaves session locked if HTTP session is invalidated in an another thread between lockSession and unlockSession calls. This PR changes the implementation so that this can no longer happen.

This is a cherry pick from: https://github.com/vaadin/framework/pull/12355

See details: https://github.com/vaadin/framework/issues/12352

TODO: Update framework version in the comments when target release is known.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
@@ -763,13 +763,22 @@ public abstract class VaadinService implements Serializable {
     /**
      * Locks the given session for this service instance. Typically you want to
      * call {@link VaadinSession#lock()} instead of this method.
+     * <p>
+     * Note: The method and its signature has been changed to return lock 
+     * instance in Vaadin 8.14.0. If you have overriden this method, you need
+     * to update your implementation.
+     * <p>
+     * Note: Overriding this method is not recommended, for custom lock storage
+     * strategy override {@link #getSessionLock(WrappedSession)} and
+     * {@link #setSessionLock(WrappedSession,Lock)} instead.
      *
      * @param wrappedSession
      *            The session to lock
+     * @return Lock instance
      * @throws IllegalStateException
      *             if the session is invalidated before it can be locked
      */
-    protected void lockSession(WrappedSession wrappedSession) {
+    protected Lock lockSession(WrappedSession wrappedSession) {
         Lock lock = getSessionLock(wrappedSession);
         if (lock == null) {
             /*
@@ -799,21 +808,31 @@ public abstract class VaadinService implements Serializable {
             lock.unlock();
             throw e;
         }
+        return lock;
     }
 
     /**
      * Releases the lock for the given session for this service instance.
      * Typically you want to call {@link VaadinSession#unlock()} instead of this
      * method.
+     * <p>
+     * Note: The method and its signature has been changed to get lock instance
+     * as parameter in Vaadin X.X.0. If you have overriden this method, you need
+     * to update your implementation.
+     * <p>
+     * Note: Overriding this method is not recommended, for custom lock storage 
+     * strategy override {@link #getSessionLock(WrappedSession)} and
+     * {@link #setSessionLock(WrappedSession,Lock)} instead.
      *
      * @param wrappedSession
      *            The session to unlock
+     * @param lock
+     *            Lock instance to unlock
      */
     protected void unlockSession(WrappedSession wrappedSession) {
-        assert getSessionLock(wrappedSession) != null;
-        assert ((ReentrantLock) getSessionLock(wrappedSession))
+        assert ((ReentrantLock) lock)
                 .isHeldByCurrentThread() : "Trying to unlock the session but it has not been locked by this thread";
-        getSessionLock(wrappedSession).unlock();
+        lock.unlock();
     }
 
     private VaadinSession findOrCreateVaadinSession(VaadinRequest request)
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -799,21 +808,31 @@ protected void lockSession(WrappedSession wrappedSession) {
...
      *
      * @param wrappedSession
      *            The session to unlock
+     * @param lock
+     *            Lock instance to unlock
      */
     protected void unlockSession(WrappedSession wrappedSession) {
-        assert getSessionLock(wrappedSession) != null;
-        assert ((ReentrantLock) getSessionLock(wrappedSession))
+        assert ((ReentrantLock) lock)
```
denis-anisimov, line 833: Is this something with github or you missed the `lock` parameter in the method signature ?

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -799,21 +808,31 @@ protected void lockSession(WrappedSession wrappedSession) {
...
+     * Note: Overriding this method is not recommended, for custom lock storage
+     * strategy override {@link #getSessionLock(WrappedSession)} and
+     * {@link #setSessionLock(WrappedSession,Lock)} instead.
      *
      * @param wrappedSession
      *            The session to unlock
+     * @param lock
+     *            Lock instance to unlock
      */
     protected void unlockSession(WrappedSession wrappedSession) {
```
denis-anisimov, line 832: Seems `lock` parameter is missed.
The question is : which version do you target this fix?
Only `master` ?
This won't go into 2.x series ?

Because this is API change technically.

So : is it only for `master`?
If no: then how you are going to deal with API change and why there is no labels for cherry-picks ?

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -763,13 +763,22 @@ protected Lock getSessionLock(WrappedSession wrappedSession) {
...
+     * {@link #setSessionLock(WrappedSession,Lock)} instead.
      *
      * @param wrappedSession
      *            The session to lock
+     * @return Lock instance
      * @throws IllegalStateException
      *             if the session is invalidated before it can be locked
      */
-    protected void lockSession(WrappedSession wrappedSession) {
+    protected Lock lockSession(WrappedSession wrappedSession) {
```
denis-anisimov, line 781: This is also API change (see comments in `unlock`).

denis-anisimov (changes requested):

denis-anisimov:
>fix: Prevent deadlock in findOrCreateVaadinSession


Strictly speaking you are not preventing any deadlock. Deadlock happens when two mutexes are acquired in a different order and two threads may not proceed because they are waiting each other.

Here the mutex is not released in the thread which acquired it.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -799,21 +808,31 @@ protected void lockSession(WrappedSession wrappedSession) {
...
      *
      * @param wrappedSession
      *            The session to unlock
+     * @param lock
+     *            Lock instance to unlock
      */
-    protected void unlockSession(WrappedSession wrappedSession) {
-        assert getSessionLock(wrappedSession) != null;
-        assert ((ReentrantLock) getSessionLock(wrappedSession))
+    protected void unlockSession(WrappedSession wrappedSession, Lock lock) {
```
denis-anisimov, line 832: Another question: is `wrappedSession` argument needed here ?
The implementation doesn't need it. 
Do you have a cases when this parameter can be needed ?
Shouldn't it be just removed ?

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
denis-anisimov: Please don't "Resolve" the comment by yourself.
You should just answer somehow that this is done.

It's the question/comment maker who should "Resolve" the comment,

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
TatuLund: With Leif we came to conclusion that if for some unlikely reason someone has overriden the method, it is good to offer this partial backwards compatibility.

denis-anisimov: The method has to be implemented anyhow since it's a new one.
So it's not backward compatible.
But if there is a vanishing chance that in the overriden method `wrappedSession` is used for something else than just getting the attribute the new signature will help to preserve the impl.
So OK.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
TatuLund: Fixed

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
TatuLund: Yes, as mentioned in comments.

denis-anisimov: API change makes this impossible to cherry-pick this fix into 2.x series.
That was the point of the comment.

It's fine for this PR in the `master` branch.

denis-anisimov (approved):

PR #11529, opened 2021-08-06
fix: Use byte[] instead of String to store stats.json

TatuLund:
Cherry pick for Flow 2.7: https://github.com/vaadin/flow/commit/841244967cc662619370e4ab6e689https://github.com/vaadin/flow/commit/841244967cc662619370e4ab6e6890e704bf7f610e704bf7f61


mshabarov (approved):

mshabarov (changes requested):
Compilation error

mshabarov:
Most likely because of missing this change from original fix?

```
        if (statistics != null) {
            return new ByteArrayInputStream(statistics.statsJson);
        }
```

TatuLund:
> Most likely because of missing this change from original fix?

True, fixed

mshabarov (approved):

PR #11552, opened 2021-08-10
fix: do not serve UI for invalid location

### LINKED ISSUE #9443: Malformed URL "/..**" causes Error 500 instead of 404
Steps to reproduce:
1. Navigate to [https://cookbook.vaadin.com/..**](https://cookbook.vaadin.com/..**)

Expected behavior:
404 error

Actual behavior:
500 error

Example stacktrace:
```
java.lang.IllegalArgumentException: Relative path cannot contain .. segments
	at com.vaadin.flow.router.Location.verifyRelativePath(Location.java:322) ~[flow-server-5.0.0.beta1.jar:5.0.0.beta1]
	at com.vaadin.flow.router.Location.parsePath(Location.java:281) ~[flow-server-5.0.0.beta1.jar:5.0.0.beta1]
	at com.vaadin.flow.router.Location.<init>(Location.java:56) ~[flow-server-5.0.0.beta1.jar:5.0.0.beta1]
	at com.vaadin.flow.router.Router.getLocationForRequest(Router.java:148) ~[flow-server-5.0.0.beta1.jar:5.0.0.beta1]
	at com.vaadin.flow.router.Router.resolveNavigationTarget(Router.java:163) ~[flow-server-5.0.0.beta1.jar:5.0.0.beta1]
	at com.vaadin.flow.server.BootstrapUtils.resolvePageConfigurationHolder(BootstrapUtils.java:285) ~[flow-server-5.0.0.beta1.jar:5.0.0.beta1]
	at com.vaadin.flow.server.BootstrapHandler$BootstrapContext.<init>(BootstrapHandler.java:195) ~[flow-server-5.0.0.beta1.jar:5.0.0.beta1]
```

pleku:
Fixes #9443 for legacy bootstrapping mode.

IT and fix for Flow 3+ bootstrapping will come next. This commit can be picked to Flow 2.x with the IT from the other PR.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
@@ -108,28 +108,49 @@ public class LocationUtil {
     }
 
     /**
-     * Parses query parameters from the given path.
+     * Handles given location when it is either {@code null} or starts with "/".
      * 
-     * @param path
-     *            the path to parse the query parameters from
+     * @param location
+     *            the location to handle
+     * @return the cleaned up location, not {@code null}
+     */
+    public static String ensureRelativeNonNull(String location) {
+        if (location == null) {
+            return "";
+        }
+        if (location.startsWith("/")) {
+            location = location.substring(1);
+        }
+        return location.trim();
+    }
+
+    /**
+     * Parses query parameters from the given location.
+     * 
+     * @param location
+     *            the location to parse the query parameters from
      * @return the query parameters
      */
-    public static QueryParameters parseQueryParameters(String path) {
-        int beginIndex = path.indexOf(Location.QUERY_SEPARATOR);
+    public static QueryParameters parseQueryParameters(String location) {
+        if (location == null) {
+            return QueryParameters.empty();
+        }
+
+        int beginIndex = location.indexOf(Location.QUERY_SEPARATOR);
         if (beginIndex < 0) {
             return QueryParameters.empty();
         }
         String query;
 
         try {
-            query = new java.net.URI(path).getQuery();
+            query = new java.net.URI(location).getQuery();
         } catch (URISyntaxException ignore) { // NOSONAR
             query = null;
         }
 
         if (query == null) {
             // decoding of parameters is done in QueryParameters
-            query = path.substring(beginIndex + 1);
+            query = location.substring(beginIndex + 1);
         }
 
         return QueryParameters.fromString(query);
--- a/flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.vaadin.flow.router;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * Utility class exposing reusable utility methods for location.
+ *
+ * @since 2.7
+ */
+public class LocationUtil {
+
+    // Prevent instantiation of util class
+    private LocationUtil() {
+    }
+
+    /**
+     * Throws {@link InvalidLocationException} if the provided path is not
+     * parseable as a relative path. A relative path should be parseable as a
+     * URI without a scheme or host, it should not contain any <code>..</code>
+     * segments and it shouldn't start with <code>/</code>.
+     *
+     * @param path
+     *            the (decoded) path to check, not null
+     */
+    public static void verifyRelativePath(String path) {
+        assert path != null;
+
+        try {
+            // Ignore forbidden chars supported in route definitions
+            String strippedPath = path.replaceAll("[{}*]", "");
+
+            URI uri = new URI(URLEncoder.encode(strippedPath, UTF_8.name()));
+            if (uri.isAbsolute()) {
+                // "A URI is absolute if, and only if, it has a scheme
+                // component"
+                throw new InvalidLocationException(
+                        "Relative path cannot contain an URI scheme");
+            } else if (uri.getPath().startsWith("/")) {
+                throw new InvalidLocationException(
+                        "Relative path cannot start with /");
+            } else if (hasIncorrectParentSegments(uri.getRawPath())) {
+                throw new InvalidLocationException(
+                        "Relative path cannot contain .. segments");
+            }
+        } catch (URISyntaxException | UnsupportedEncodingException e) {
+            throw new InvalidLocationException("Cannot parse path: " + path, e);
+        }
+
+        // All is OK if we get here
+    }
+
+    /**
+     * Parses the given path to parts split by the path separator, ignoring the
+     * query string if present. The path is verified with
+     * {@link #verifyRelativePath(String)}.
+     * 
+     * @param path
+     *            the path to parse
+     * @return tha path split into parts
+     */
+    public static List<String> parsePathToSegments(String path) {
+        final String basePath;
+        int endIndex = path.indexOf(Location.QUERY_SEPARATOR);
+        if (endIndex >= 0) {
+            basePath = path.substring(0, endIndex);
+        } else {
+            basePath = path;
+        }
+
+        verifyRelativePath(basePath);
+
+        List<String> splitList = Arrays
+                .asList(basePath.split(Location.PATH_SEPARATOR));
+        if (basePath.endsWith(Location.PATH_SEPARATOR)) {
+            // Explicitly add "" to the end even though it's ignored by
+            // String.split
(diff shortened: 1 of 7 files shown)


denis-anisimov (commented):
That fix handles only one case: open a page with location which we don't accept.
But as I understand it won't do anything if navigation happens from the correct location to the location ".." using a router link.

So the code which catch the invalid location should be somewhere inside Router handling code.
And there is already some code which handles invalid location. But it does it in a specific way.
And then there is a question how it should handle it: 503 vs 404 is a semantical difference.
".." is invalid because we are trying to prevent traversing resources via waling using parent directory. So every such attempt is disallowed and 503 is used to indicate that: it's disallowed.
On the other hand ".." is an attempt to go to the parent which doesn't exist in this specific case described in the ticket. And it should be 404 because it doesn't exist.

There is a number of issue with that:
- ".." currently is always "invalid" URL and 503 is returned,
- ".." should return 404 for "/.." according to the ticket
- if "foo" an existing route and "/" is a route then ".." should work fine if you are at the moment in the "foo" route since it becomes "foo/.." and it should be translated to "/" which is working route.

So the current 503 error handles everything semantically as: wrong URL and you should not try to use ".." ever inside the URL.
The ticket wants to distinguish ".." segment inside URL and it's significant efforts to implement.
I personally don't see serious benefit to have 404 code instead of 503.
The current approach allows to handle all this cases in a similar way.

I cannot say that the expected behavior in the ticket is really expected even though it makes sense.
The current behavior : 503 for every URL which contains ".."  makes sense for me as well. So we disallow every URL which contains "..".

pleku:
> That fix handles only one case: open a page with location which we don't accept.
But as I understand it won't do anything if navigation happens from the correct location to the location ".." using a router link.

That would be another thing, which I'm not sure how relevant/important it is, but anyway it can be another ticket if necessary.

> So the current 503 error handles everything semantically as: wrong URL and you should not try to use ".." ever inside the URL.

@denis-anisimov I don't understand 503 error like that, can you please elaborate how you see it like that ? To me 503 error is about _service not available_ and in e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503 it says
> The HyperText Transfer Protocol (HTTP) 503 Service Unavailable server error response code indicates that the server is not ready to handle the request.
Common causes are a server that is down for maintenance or that is overloaded. This response should be used for temporary conditions and the Retry-After HTTP header should, if possible, contain the estimated time for the recovery of the service.

To me there is a clear distinction between server getting an _invalid URL that won't work_ and _service being down for maintenance_. Thus 503 is not the right thing, and I think 400 (not 404) which e.g. Tomcat uses is something we should align with.
> The HyperText Transfer Protocol (HTTP) 400 Bad Request response status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).

So to me there is a clear difference between the two that is being fixed for initial bootstrapping.

denis-anisimov:
Well, I don't know which error code is more suitable.
I don't insist on 503 and don't really care what is it.

What is important is: 404 which is Page not found (without any doubt) and some error which says that there is an error on the server side. This is the only important thing.

I don't see this is as a separate issue since it's exactly the question: whether we treat ".." as invalid generally (and then it's a some error code which is not 404) or depending on request it's either valid or not.

The invalid URL is handled inside the Router logic and the fix should go there instead of boostrapping. The current fix location is just in a wrong place. So instead of covering all the cases including the router navigation it's located on the top  of the stacktrace with catching only one specific case and it uses a hack: instantiate a `Location` which is never needed.
This is an artificial code and looks like a hack.

The fix moved into the Router on the other hand will handle every case properly and generally.

And again: the main complain is semantic. Why 404 ?
404 means page not found.
So is I have two routes:
- "/foo"
- "/"
And "/foo/.." URL means "/". So it's not 404 . And the fix is invalid from this point of view.
But if we treat every URL with ".." as invalid (and not 404) then it's fine.

I don't know this functionality enough. @caalador knows thus much better and the code is written mostly by him.
So I suggest him as a reviewer.

denis-anisimov:
I was referring to 404 since the original ticket description talks about 404.

But now I see 400 in the fix. I've missed that .
I though that by default 503 is used instead of 500. But it's 500 in the ticket description.

So the error code 400 is fine I think. But anyway: the logic should be somewhere inside router code which creates a `Location` instead having a fake `Location` instantiation inside boostrap.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -540,6 +544,32 @@ public boolean synchronizedHandleRequest(VaadinSession session,
...
+     *         {@code false} if not (location was valid)
+     * @throws IOException in case writing to response fails
+     */
+    protected boolean writeErrorCodeIfRequestLocationIsInvalid(
+            VaadinRequest request, VaadinResponse response) throws IOException {
+        try {
+            // #9443 Use error code 400 for bad location and don't create UI
+            new Location(request.getPathInfo(),
+                    QueryParameters.full(request.getParameterMap()));
+        } catch (InvalidLocationException invalidLocationException) {
```
caalador, line 565: I would feel that instead of creating a Location object here we should open up (or create a LocationUtil class) `Location::verifyRelativePath` and use that in the catch block with `Location::ensureRelativeNonNull(request.getPathInfo())`

pleku: Added the new utility class and moved the static methods there from `Location`. There are some more candidates, but at least the campsite is now cleaner than when I arrived.

We could also validate the query string to be parseable before creating an UI, but I'd consider that to be extra at this point.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -540,6 +545,34 @@ public boolean synchronizedHandleRequest(VaadinSession session,
...
+     *         {@code false} if not (location was valid)
+     * @throws IOException
+     *             in case writing to response fails
+     */
+    protected boolean writeErrorCodeIfRequestLocationIsInvalid(
+            VaadinRequest request, VaadinResponse response) throws IOException {
+        try {
+            // #9443 Use error code 400 for bad location and don't create UI
+            new Location(request.getPathInfo(),
+                    QueryParameters.full(request.getParameterMap()));
```
caalador, line 566:
```suggestion
```
can be dropped as the util call does what we need/expect here.

flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Parses the given path to parts split by the path separator, ignoring the
+     * query string if present. The path is verified with
+     * {@link #verifyRelativePath(String)}.
+     * 
+     * @param path
+     *            the path to parse
+     * @return tha path split into parts
+     */
+    public static List<String> parsePath(String path) {
```
caalador, line 85: This would need the do not remove query part as per `LocationTest.locationWithParamsInUlAndParameters` so this should be overloaded with `parsePath(String path, boolean stripQueryString)` and `parsePath(String path) { parsePath(path, true); }`

This is only in the case that the test is actually correct.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
pleku: I don't understand how the test is even working correctly, it using the constructor that uses both path + separated query parameters. And the constructor javadocs says:
```java
   * The location string can not contain query parameters. To pass query
     * parameters, either specify them in {@link QueryParameters} in this
     * constructor, or use {@link Location#Location(String)}
```
And then the test verifies that it is using the query parameters from the path instead of the passed `QueryParams` object... WTF

pleku: 1) why (oh why) is there a test that tests that there can be different query parameters in the path and as preparsed
2) why (oh why) would the query parameters from the path be reported as part of a segment in `getSegments()` ?

I think the test is just 💩 and will fix it

pleku: Overloading the method makes no sense IMO. Renamed the method

caalador: Yes. After lookin really hard at the code and the exception it started to look like the test is not testing what it should.
The overload was only for if the test actually tests what it should test.

flow-server/src/test/java/com/vaadin/flow/router/LocationTest.java
```diff
@@ -224,8 +224,9 @@ public void locationWithParamsInUrl() {
     public void locationWithParamsInUrlAndParameters() {
         Location location = new Location("foo/bar/?one&two=222",
                 getQueryParameters());
-        assertEquals(Arrays.asList("foo", "bar", "?one&two=222"),
-                location.getSegments());
+        assertEquals(Arrays.asList("foo", "bar", ""), location.getSegments());
+        Assert.assertEquals("one=1&one=11&two=2&two=22&three=3",
```
caalador, line 228: Could have message saying why the expected is this and not `one&two=222`

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
pleku: oops 😅 done

flow-server/src/test/java/com/vaadin/flow/router/LocationTest.java
pleku: done

caalador (dismissed):

caalador:
The test probably has something to do with the strange `stripQueryString` functionality that has left the query from the string as part of the path.
basically it should test `%3faaa` instead of `?aaa`, but the junit test most likely uses ? instead of %3f as it simulates the path being decoded.

pleku:
> The test probably has something to do with the strange `stripQueryString` functionality that has left the query from the string as part of the path.
> basically it should test `%3faaa` instead of `?aaa`, but the junit test most likely uses ? instead of %3f as it simulates the path being decoded.

The test seems broken indeed as it is using `?` as part of rquest's `pathInfo` which is incorrect as `?` is a reserved character for requests

caalador (approved):

PR #11570, opened 2021-08-13
fix: set the lock field to a fake lock for unbound session

### LINKED ISSUE #11569: NPE after live reload
In Vaadin 21.0.0.beta1, a session that was [deserializedAsEmpty](https://github.com/vaadin/flow/blob/8.0/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java#L2175-L2183) (#11304) will be unbound. The `lock` however is transient, leading to the following NPE:
```
java.lang.NullPointerException: null
	at com.vaadin.flow.server.VaadinSession.unlock(VaadinSession.java:720) ~[flow-server-8.0.0.beta1.jar:8.0.0.beta1]
	at com.vaadin.flow.server.VaadinSession.valueUnbound(VaadinSession.java:221) ~[flow-server-8.0.0.beta1.jar:8.0.0.beta1]
	at org.apache.catalina.session.StandardSession.removeAttributeInternal(StandardSession.java:1831) ~[tomcat-embed-core-9.0.45.jar:9.0.45]
	at org.apache.catalina.session.StandardSession.removeAttribute(StandardSession.java:1317) ~[tomcat-embed-core-9.0.45.jar:9.0.45]
	at org.apache.catalina.session.StandardSession.removeAttribute(StandardSession.java:1289) ~[tomcat-embed-core-9.0.45.jar:9.0.45]
	at org.apache.catalina.session.StandardSessionFacade.removeAttribute(StandardSessionFacade.java:153) ~[tomcat-embed-core-9.0.45.jar:9.0.45]
	at com.vaadin.flow.server.WrappedHttpSession.removeAttribute(WrappedHttpSession.java:126) ~[flow-server-8.0.0.beta1.jar:8.0.0.beta1]
	at com.vaadin.flow.server.VaadinService.readFromHttpSession(VaadinService.java:2169) ~[flow-server-8.0.0.beta1.jar:8.0.0.beta1]
```

Because of #11416
https://github.com/vaadin/flow/blob/32d648beb96fddb2cb811d45f655037b0cfec603/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java#L220-L222

denis-anisimov:
fixes #11569

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
@@ -203,8 +203,8 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
         // the latter case the session field should be set to {@code null}
         // immediately to avoid using HTTP session object which is invalid (all
         // methods throw exceptions).
+        lock();
         try {
-            lock();
             if (getAttribute(CLOSE_SESSION_EXPLICITLY) == null) {
                 session = null;
             }
@@ -556,7 +556,6 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
      *            and the session is not locked
      */
     public void checkHasLock(String message) {
-        assert deserializedAsEmpty == false : "The session was discarded during serialization and should not be used";
         if (configuration == null || configuration.isProductionMode()) {
             assert hasLock() : message;
         } else if (!hasLock()) {
@@ -1041,6 +1040,10 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
         boolean wasSerialized = stream.readBoolean();
         this.deserializedAsEmpty = !wasSerialized;
         if (!wasSerialized) {
+            // even an empty unbound session still should support its API which
+            // requires a lock, so set a fake lock to be able to avoid
+            // unexpected issues on calling VaadinSession methods
+            lock = new ReentrantLock();
             return;
         }
         Map<Class<?>, CurrentInstance> old = CurrentInstance.setCurrent(this);
--- a/flow-server/src/test/java/com/vaadin/tests/server/SerializationTest.java
+++ b/flow-server/src/test/java/com/vaadin/tests/server/SerializationTest.java
@@ -7,18 +7,40 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
 
+import junit.framework.TestCase;
+import org.mockito.Mockito;
+
 import com.vaadin.flow.server.VaadinContext;
 import com.vaadin.flow.server.VaadinService;
 import com.vaadin.flow.server.VaadinSession;
 import com.vaadin.flow.server.startup.ApplicationConfiguration;
 
-import org.mockito.Mockito;
-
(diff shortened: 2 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -203,8 +203,8 @@ public void valueUnbound(HttpSessionBindingEvent event) {
         // the latter case the session field should be set to {@code null}
         // immediately to avoid using HTTP session object which is invalid (all
         // methods throw exceptions).
+        lock();
         try {
-            lock();
```
caalador, line 207: This change shouldn't be needed as it doesn't change anything apart from the exception message when the lock feature is broken so that both lock() and unlock() throw and unlock exception is shown instead of lock.

flow-server/src/test/java/com/vaadin/tests/server/SerializationTest.java
```diff
@@ -7,18 +7,40 @@
...
+                session.getPendingAccessQueue());
+    }
+
+    public void testSerializeVaadinSession_notProductionMode_disableDevModeSerialization_deserializedSessionHasLockInstance()
+            throws Exception {
+        VaadinSession session = mockSession(false);
+
+        session = serializeAndDeserialize(session);
+
+        assertNotNull(session.getLockInstance());
```
caalador, line 37: Could have failure message `"Lock should be available after empty deserialization"` or something to make the possible failure clearer

caalador (changes requested):
Else seems fine.

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
denis-anisimov: This is indeed not needed :
- in the perfect world 
- this PR

So when everything works as expected this change doesn't do anything.
But the reason is the exactly what you said: when/if `lock|()` throws an exception then code goes into `finally` and then 
it's a mixture of additional unexpected things like this issue (when `checkLock` doesn't throw anything then there is another `finally` ) and you get an exception which you can't understand how it happen...
This is exactly what happened here: NPE is thrown but it looks extremely weird since there were many places above where different NPE or ISE should have been thrown.
And I spent significant (far away from a few minutes) to understand how it's possible : why the NPE is thrown _only_ inside `unlock`.

This change doesn't do anything if everything works as expected.
But it will throw a proper exception earlier (before `finally`) when something is broken. And the resulting exception will be the proper exception.
So if in the next time someone accidentally break the code via having a `null` for a lock instance then the exception will be thrown immediately.

As a result : `lock` may not throw an exception. If it throws then it should be outside of `try` since this is not recoverable exception anyhow and it's better to see the exact place/line when the exception is thrown instead decoding the exception which is not a cause but a consequence.

flow-server/src/test/java/com/vaadin/tests/server/SerializationTest.java
denis-anisimov: done

caalador (approved):

PR #11577, opened 2021-08-16
Update to JSoup 1.14.2

Legioth:
Latest version improves handling of weird HTML and XML along with some smaller fixes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -114,7 +114,7 @@
         <flow.dev.dependencies.file>devDependencies.json</flow.dev.dependencies.file>
 
         <!-- Used in OSGi manifests -->
-        <jsoup.version>1.14.1</jsoup.version>
+        <jsoup.version>1.14.2</jsoup.version>
         <!-- Note that this should be kept in sync with the class Constants -->
         <atmosphere.runtime.version>2.4.30.slf4jvaadin1
         </atmosphere.runtime.version>


denis-anisimov (approved):

AB-xdev:
Will this get backported to Vaadin 14 (LTS)?

Asking because there was a CVE fixed:
* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-37714
* https://github.com/jhy/jsoup/security/advisories/GHSA-m72m-mhq2-9p6c

denis-anisimov:
I don't think so.
Jsoup 14 contain breaking API changes. We cannot break existing V14 applications.

But there was a fix in Flow to use the API which is available in Jsoup 14: https://github.com/vaadin/flow/commit/d12e31dc3927729ba04d6c11092fae177d3fb055
It means that there should not be issues to override the Jsoup version in your project via  explicit dependency declaration (not transitive).

Legioth:
Might still be warranted to make the upgrade in an upcoming minor release since it fixes a security issue that can be quite critical. It would just be great to understand whether there are any add-ons in Directory that would be affected. If we choose to make the upgrade, then it should be clearly documented as a breaking change.

AB-xdev:
Shouldn't existing Vaadin applications or add-ons that use JSoup have it also **explicitly** declared in there build files (e.g. ``pom.xml``)?

Legioth:
In theory yes. In practice quite seldom.

Though I'm more concerned about add-ons than applications. If you have direct use of the removed APIs in your application, then it doesn't take many minutes to get them fixed. If you're using an add-on that still uses old APIs, then you will have a much more difficult situation until a new version of that add-on is avilable.

Issue #11260, opened 2021-06-16
Disabling Anchor clears href


mvysny:
### Minimal reproducible example
Paste this into the Skeleton Starter for Vaadin 21.0.0.alpha5
```
package org.vaadin.example;

import com.vaadin.flow.component.html.Anchor;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.router.Route;

@Route("")
public class MainView extends VerticalLayout {

    public MainView() {
        final Anchor a = new Anchor("http://foo.com", "testing");
        a.setEnabled(false);
        System.out.println("HREF=" + a.getHref());
    }
}
```
### Expected behavior
Should print "HREF=http://foo.com"
### Actual behavior
Prints "HREF=" (the `a.getHref()` returns a blank string)
### Versions:
    - Vaadin / Flow version: 21.0.0.alpha5
    - Java version: 11
    - OS version: Ubuntu Linux 21.04
    - Browser version (if applicable): Firefox
    - Application Server (if applicable): Jetty
    - IDE (if applicable): Intellij

knoobie:
That's correct - see https://github.com/vaadin/flow/issues/10924

denis-anisimov:
Works as coded and designed in the https://github.com/vaadin/flow/issues/10924.

This ticket is a contradiction for https://github.com/vaadin/flow/issues/10924.

I personally don't have any preference how it should work. But it has been agreed in https://github.com/vaadin/flow/issues/10924 that it should work as designed.

If this is undesired  behavior then it's possible to override `onEnabledStateChanged` as it's implemented in the Component:
```
public void onEnabledStateChanged(boolean enabled) {
        if (getElement().getNode().hasFeature(ElementData.class)) {
            getElement().setAttribute("disabled", !enabled);
        }
    }
```

Issue #10924, opened 2021-05-06
Anchor component improper implement of HasEnabled


guygoldwasser:
B"H
follow this: https://stackoverflow.com/questions/13955667/disabled-href-tag/13955695
the setEnabled implementation not fit the html way of making <a> being disabled/enabled

Need some workaround implementation for this case e.g. set the href to none when it disabled

miguelatvaadin:
HI @guygoldwasser ,

thanks for submitting the issue.

As a workaround, just add this to your `styles.css`:

```css
a[disabled] {
    pointer-events: none;
    cursor: default;
   /* add whatever other css changes you want to highlight it is disabled (e.g. change color) */
}
```

I will redirect this issue to the proper team.

rolfsmeds:
Setting `pointer-events: none` is not the solution: the link will still be triggerable by keyboard, so this is not just a styling issue.

miguelatvaadin:
Yes. Good point. Thanks, @rolfsmeds .

At the first moment I thought css would be enough, but I understand that changing the `href` attribute must also be done.

Anyway, the behavior of a disabled link should also be modified so the user does not think that it is still clickable, when it has been disabled.

Should I still open an issue for this at the lumo repo, besides keeping this one for flow?

knoobie:
Some additional information from the HTML spec: 

> If the a element has no href attribute, then the element represents a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element's contents.

Source: https://html.spec.whatwg.org/#the-a-element 

Additionally, there is no mention of `disabled` as valid attribute - meaning `<a disabled>` is invalid html. Sounds like the `setEnabled(false)` method should be overwritten to not set the attribut and remove the `href` instead. (saving it somewhere, so that it can be restored, when calling `setEnabled(true)`(?))

Additional information regarding accessibility: https://www.scottohara.me/note/2019/07/17/placeholder-link.html

miguelatvaadin:
Thank you very much for your comments @knoobie ,

yes, the idea is to remove the `href` attribute and restore it when the anchor is enabled again, and the lumo theme would handle the style accordingly.

Legioth:
Should be noted that one should not override `setEnabled`, but rather `onEnabledStateChanged` to make this kind of change in how the component updates itself when it's disabled.

joheriks:
## Acceptance criteria:
[...]

PR #11578, opened 2021-08-16
fix: avoid setting null href and handle removeHref for disabled Anchor

### LINKED ISSUE #11571: Calling setEnabled(true) on anchor causing an IllegalArgumentException
Calling setEnabled(true) on an anchor without a href set causing an IllegalArgumentException

```
Anchor anc = new Anchor();
anc.setEnabled(true);
<VertircalLayout>.add(anc);
```

### Expected behavior
Anchor component enabled without any exception

### Actual behavior
```
java.lang.IllegalArgumentException: Cannot set href to null
at com.vaadin.flow.component.PropertyDescriptors$PropertyDescriptorImpl.set(PropertyDescriptors.java:67)
at com.vaadin.flow.component.PropertyDescriptor.set(PropertyDescriptor.java:53)
at com.vaadin.flow.component.Component.set(Component.java:481)
at com.vaadin.flow.component.html.Anchor.setHref(Anchor.java:119)
at com.vaadin.flow.component.html.Anchor.restoreHref(Anchor.java:210)
```

### Versions:
    - Vaadin / Flow version: 20.0.6
    - Java version: 11
    - OS version: Windows

denis-anisimov:
fixes #11571

## Description

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -47,6 +47,7 @@ public class Anchor extends HtmlContainer implements Focusable<Anchor> {
 
     private String disabledHref;
     private AbstractStreamResource resource;
+    private boolean isDisabledHrefRemoved;
 
     /**
      * Creates a new empty anchor component.
@@ -148,7 +149,10 @@ public class Anchor extends HtmlContainer implements Focusable<Anchor> {
      *
      */
     public void removeHref() {
-        getElement().removeAttribute("href");
+        if (!isEnabled()) {
+            isDisabledHrefRemoved = true;
+        }
+        doRemoveHref();
     }
 
     /**
@@ -264,15 +268,20 @@ public class Anchor extends HtmlContainer implements Focusable<Anchor> {
             disabledHref = null;
         } else {
             disabledHref = getHref();
-            removeHref();
+            doRemoveHref();
         }
     }
 
     private void restoreHref() {
-        if (!getElement().hasAttribute("href")) {
+        assert isEnabled();
+        if (!getElement().hasAttribute("href") && !isDisabledHrefRemoved) {
             // only set attribute value back if attribute has not been set
             // somehow when component was disabled
-            setHref(disabledHref);
+            if (disabledHref == null) {
+                doRemoveHref();
+            } else {
+                setHref(disabledHref);
+            }
         }
     }
 
@@ -282,6 +291,10 @@ public class Anchor extends HtmlContainer implements Focusable<Anchor> {
             // somehow when component was disabled
             setHref(resource);
         }
+    }
+
+    private void doRemoveHref() {
+        getElement().removeAttribute("href");
     }
 
 }
--- a/flow-html-components/src/test/java/com/vaadin/flow/component/html/AnchorTest.java
+++ b/flow-html-components/src/test/java/com/vaadin/flow/component/html/AnchorTest.java
@@ -246,6 +246,30 @@ public class AnchorTest extends ComponentTest {
         Assert.assertEquals("baz", anchor.getHref());
     }
 
+    @Test
+    public void setEnabled_anchorWithoutHref_doesNotThrow() {
+        Anchor anchor = new Anchor();
+        anchor.setEnabled(false);
+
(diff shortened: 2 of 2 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -148,7 +149,10 @@ public void setHref(String href) {
...
      */
     public void removeHref() {
-        getElement().removeAttribute("href");
+        if (!isEnabled()) {
+            isDisabledHrefRemoved = true;
+        }
+        doRemoveHref();
     }
 
     /**
```
pleku, line 158: I think the `setHref` javadocs should be updated to the fact that when the anchor is inside a disabled component container, the `href` is hidden and the anchor cannot be interacted with.

Also to me the following is odd design:
```java
Anchor anchor = new Anchor("google.com", "google");
anchor.setEnabled(false); // <-- hides href

anchor.setHref("vaadin.com"); // <-- href comes visible again ...
```
I would instead think that all `setHref`, `removeHref` and `getHref` (#11260) can be called whenever the anchor is disabled, but the changes are not applied to the client/DOM until it is enabled again. And this should be clarified in javadocs for both `setEnabled` and `setHref`.

pleku (changes requested):
I'm not happy with the original design done in #10924 nor the missing javadocs updates of due to the behavior. I would propose a change in the behavior completely

denis-anisimov:
I don't see how any of comments are related to the PR which is going to fix #11571.
This PR is not aiming to fix any design issues done in #10924 (it's not clear why the ticket was ready to do if the design is not confirmed).

This PR fixes #11571 and makes remove href behavoir consistent to update href functionality. It doesn't do anything else.

So only two actions can be done _here_:
- ?correct? and merge this PR as a fix for #11571 . Then may be create other tickets to correct unrelated things.
- close this PR since the whole functionality is originally poorly designed and it has no sense to fix #11571. Instead fix for  #10924 should be just reverted .

I'm totally fine with any approach. I also don't like design in #10924 but since the design looked accepted in the ticket I implemented it as described.

pleku:
> I'm totally fine with any approach. I also don't like design in #10924 but since the design looked accepted in the ticket I implemented it as described.

I don't see any/much design done in the issue itself, other than just what is the behavior of the element in the browser. The Java API has been left for "implementation and review to decide" but in my opinion it was not considered. I'm OK for closing this and making a revert for the original PR or just fixing it directly. Noting @miguelatvaadin as he was the original PR reviewer.

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
mstahv: I agree with Pekka that this is how it should work. Updating href while anchor is disabled should work fine, but the new value should end up in the dom only after it is enabled.

denis-anisimov: I kind of agree with this too except I don't see a proper way to implement this.

I insist on the fact of synchronization of Component API and Element API.
So the resulting design of the original ticket should clarify the question of synchronization.

1. Question: Anchor.setEnabled(false); Anchor:setHref("foo"). What happens with Anhcor::getHref and with Anchor::getElement::getAttribute("href");
2. Question Anchor.setEnabled(false), Anchor::getElement::setAttribute("href", "bar"); What happens with Anchor::getHref ?
3.  Question Anchor.setEnabled(false), Anchor::getElement::setAttribute("href", "bar");  What happens with Anhcor::getHref  and  Anchor::getElement::getAttribute("href");  _AFTER_  Anchor.setEnabled(true) ?
4.  Question: Anchor.setEnabled(false) . What do return Anhcor::getHref and Anchor::getElement::getAttribute("href"); ?

Since there is no way to synchronize Component and Element API I don't see how the original ticket may be done.
To be able to disable "href" attribute on the client side ONLY this required dramatic changes in the generic StateNode low level API and client side which is too much for such a small enhancement.

That's why I don't want to take responsibility of any decision made here and don't want to participate in design : there is no way to make it based on current StateNode API properly.

Issue #11584, opened 2021-08-17
Build fails in 14.x after NPM 7.20.6 update


caalador:
NPM released an updated latest version to 7.20.6 on 12.8.2021

Now a clean Vaadin 14 project fails with a webpack exception:

```
[INFO] --- flow-maven-plugin:2.7-SNAPSHOT:build-frontend (default) @ flow-test-themes ---
[INFO] Scanning classes to find frontend configurations and dependencies...
[INFO] Visited 797 classes. Took 11183 ms.
[INFO] Visited 6 classes. Took 36 ms.
[INFO] Running `npm install` to resolve and optionally download frontend dependencies. This may take a moment, please stand by...
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
npm WARN deprecated chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
npm WARN deprecated querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
npm WARN deprecated html-webpack-plugin@3.2.0: 3.x is no longer supported
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated core-js@2.6.12: core-js@<3.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Please, upgrade your dependencies to the actual version of core-js.
[INFO] Frontend dependencies resolved successfully.
[INFO] Copying frontend resources from jar files ...
[INFO] Visited 60 resources. Took 221 ms.
[INFO] Copying project local frontend resources.
[INFO] Copying frontend directory completed.
[INFO] Running webpack ...

Insufficient number of arguments or no entry found.
Alternatively, run 'webpack(-cli) --help' for usage info.

Hash: 6694ce404673f98a5369
Version: webpack 4.42.0
Time: 44ms
Built at: 17/08/2021 8.22.30

[...]

PR #11585, opened 2021-08-17
fix: blacklist npm versions

caalador:
Add new versions to the NPM blacklist.
Update information on that to do about
having an blacklisted version.

Touches #11584

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -91,7 +91,11 @@ public class FrontendTools {
     private static final List<FrontendVersion> NPM_BLACKLISTED_VERSIONS = Arrays
             .asList(new FrontendVersion("6.11.0"),
                     new FrontendVersion("6.11.1"),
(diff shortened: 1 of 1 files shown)


pleku (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -91,7 +91,11 @@
     private static final List<FrontendVersion> NPM_BLACKLISTED_VERSIONS = Arrays
             .asList(new FrontendVersion("6.11.0"),
                     new FrontendVersion("6.11.1"),
-                    new FrontendVersion("6.11.2"));
+                    new FrontendVersion("6.11.2"),
+                    new FrontendVersion("7.20.3"),
```
knoobie, line 95: FYI: Node includes one of the blacklisted versions by default in 16.6.2: https://nodejs.org/en/download/current/ (16.6.2 (includes npm 7.20.3))

pleku: This is just temporary to inform users instead of just letting things fail in an obscure way.

PR #8810, opened 2020-08-04
#7911 Implement Component.isAttached().

mvysny:
Fixes #7911

PR #11601, opened 2021-08-19
feat: Implement Component.isAttached(). (#8810)

### LINKED ISSUE #7911: Add Component.isAttached()
Currently in Vaadin 14.1.19 it's only possible to check whether a component is attached, by running the following method:

```kotlin
fun Component.isAttached(): Boolean {
    val ui: UI = ui.orElse(null) ?: return false
    return !ui.isClosing
}
```

It would be good to have `Component.isAttached()` directly provided by Vaadin.

caalador:
Closes #7911

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -432,6 +432,19 @@ public abstract class Component
     }
 
     /**
+     * Checks whether this component is currently attached to a UI.
+     * <p>
+     * When {@link UI#close()} is called, the UI and the components are not detached immediately;
+     * the UI cleanup is performed at the end of the current request which also detaches the
+     * UI and its components.
+     *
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/component/Component.java
```diff
@@ -431,6 +431,19 @@ protected void onDetach(DetachEvent detachEvent) {
...
     }
 
+    /**
+     * Checks whether this component is currently attached to a UI.
+     * <p>
+     * When {@link UI#close()} is called, the UI and the components are not detached immediately;
+     * the UI cleanup is performed at the end of the current request which also detaches the
+     * UI and its components.
+     *
+     * @return true if the component is attached to an active UI.
```
pleku, line 441:
```suggestion
     * @return true if the component is attached to an active UI.
     * @ since 2.7
```

pleku (approved):

pleku (approved):

PR #11630, opened 2021-08-24
fix: retain fragment when using push/replaceState

### LINKED ISSUE #11628: History.pushState drops fragment from location if query parameters are included
This is a regression after the fix for the `QueryParameters` behavior (#10521), the as now fragment part is dropped off from `Location`, as it used to work with fragment `#` in the location by chance. So starting from Vaadin 20, the `History.pushState` has been broken in Flow when the location has contained both query parameters and fragment. It is valid to push a location with fragment to the browser and many apps that tweak the scrolling location on the page may rely on this.

This would have been caught by tests RouterLinkIT, but as the test was already ignored by fusion-team (#7575), the regression has slipped through (I do admit a unit test should have been there too but it has not been). So now it probably needs to be fixed for both master and 2.7 (as part of #11590).

pleku:
Fixes #11628


flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
```diff
@@ -0,0 +1,156 @@
...
+    private class TestUI extends UI {
+        @Override
+        public Page getPage() {
+            return page;
+        }
+    }
+
+    private class TestPage extends Page {
+
+        private int count = 0;
```
caalador, line 41: What is the count needed for? I feel it should be removed as it's just there now,

caalador (changes requested):

flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
pleku:
```suggestion
```

flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
```diff
@@ -0,0 +1,156 @@
...
+        public TestPage(UI ui) {
+            super(ui);
+        }
+
+        @Override
+        public PendingJavaScriptResult executeJs(String expression,
+                Serializable... parameters) {
+            this.expression = expression;
+            this.parameters = parameters;
+            count++;
```
pleku, line 56:
```suggestion
```

flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
pleku: Not needed in this test, c/p leftover

caalador:
Formatter should have posted:
`>>> There are formatting issues in the following files:\nflow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java` but didn't show up for some reason

caalador (approved):

pleku:
Will manually pick to 2.7

pleku:
Build failed for some gradle issue on agent ...

PR #11635, opened 2021-08-25
fix: Remove removed items from array

### LINKED ISSUE #11608: Memory leak in theme generator
The logic [here](https://github.com/vaadin/flow/blob/fbcec8dc67543eed9ff2c0228c45432a6d0ace65/flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js#L267) seems to push strong references of themed shadow roots to an array in the global scope but they never get cleaned up.

Scroll the page at https://vaadin.com/docs/latest/ds/components/text-field up and down so that new example instances get created and old ones get discarded. Check `window.Vaadin._vaadintheme_docs_globalCss`, which seems to accumulate more and more references to the example shadow roots as you scroll.

### Expected behavior
The shadow root references get cleaned up

### Actual behavior
The array keeps accumulating shadow root references

caalador:
Remove any elements added to the
global array when the DOMNode is removed
from the dom.

Closes #11608

### DIFF
--- a/flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
+++ b/flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
@@ -265,7 +265,13 @@ window.Vaadin['${globalCssFlag}'] = window.Vaadin['${globalCssFlag}'] || [];
   if (injectGlobal) {
     ${globalCssCode.join('')}
     window.Vaadin['${globalCssFlag}'].push(target);
-  }
+    
+    const eventTarget = (target instanceof ShadowRoot) ? target.host : target;
+    eventTarget.addEventListener('DOMNodeRemoved', event => {
+      const index = window.Vaadin['${globalCssFlag}'].indexOf(target);
+      window.Vaadin['${globalCssFlag}'].splice(index, 1);
+    });
+   }
   if (!document['${componentCssFlag}']) {
     ${componentCssCode.join('')}
     document['${componentCssFlag}'] = true;
--- a/flow-tests/test-embedding/test-embedding-application-theme/src/test/java/com/vaadin/flow/webcomponent/ApplicationThemeComponentIT.java
+++ b/flow-tests/test-embedding/test-embedding-application-theme/src/test/java/com/vaadin/flow/webcomponent/ApplicationThemeComponentIT.java
@@ -221,4 +221,22 @@ public class ApplicationThemeComponentIT extends ChromeBrowserTest {
         Assert.assertTrue("Missing link for external url", linkUrls
                 .contains("https://fonts.googleapis.com/css?family=Itim"));
     }
+
+    @Test
+    public void removingEmbeddedComponent_arrayIsCleaned() {
+        open();
+        checkLogsForErrors();
+        Assert.assertEquals(
+                "Both embedded components should have registered globalCss", 2l,
+                getCommandExecutor().executeScript(
+                        "return window.Vaadin['_vaadintheme_embedded-theme_globalCss'].length"));
+        getCommandExecutor()
+                .executeScript("document.getElementById('second').remove()");
+
+        Assert.assertEquals(
+                "After removal of second should have removed its globalCss", 1l,
+                getCommandExecutor().executeScript(
+                        "return window.Vaadin['_vaadintheme_embedded-theme_globalCss'].length"));
+
+    }
 }


flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -265,7 +265,13 @@ window.Vaadin['${globalCssFlag}'] = window.Vaadin['${globalCssFlag}'] || [];
   if (injectGlobal) {
     ${globalCssCode.join('')}
     window.Vaadin['${globalCssFlag}'].push(target);
-  }
+    
+    const eventTarget = (target instanceof ShadowRoot) ? target.host : target;
+    eventTarget.addEventListener('DOMNodeRemoved', event => {
```
tomivirkki, line 270: Using "DOMNodeRemoved" event for this may have some drawbacks:
- The event is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/MutationEvent) and shows a warning in the console (Chrome): `Added synchronous DOM mutation listener to a 'DOMNodeRemoved' event. Consider using MutationObserver to make the page more responsive.`
- The event doesn't seem to invoke for child elements if their parent element is removed from the DOM (nested layouts may still suffer from the issue).

I wonder if using weak references (`WeakMap` maybe) would work better in this case?

pleku, line 270: Should this instead use a `MutationObserver` as `MutationEvent`s are deprecated ?

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -265,7 +265,13 @@ window.Vaadin['${globalCssFlag}'] = window.Vaadin['${globalCssFlag}'] || [];
...
+      const index = window.Vaadin['${globalCssFlag}'].indexOf(target);
+      window.Vaadin['${globalCssFlag}'].splice(index, 1);
+    });
```
pleku, line 273: If/when using an observer instead, it can be disconnected after the clean-up

pleku (changes requested):

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
pleku: Ah, I'm slow and github push updates apparently don't work

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -255,6 +255,21 @@ function generateThemeFile(themeFolder, themeName, themeProperties, productionMo
...
+        const index = window.Vaadin['${globalCssFlag}'].indexOf(node.shadowRoot);
+        if(index > -1) {
+          window.Vaadin['${globalCssFlag}'].splice(index, 1);
+        }
+      })
+    })
+  }
+);
+cleanupObserver.observe(document.body, {
+  childList: true
```
tomivirkki, line 271: This config only observes the immediate children of the body element.

You could add `subtree: true` to the config also mut it has some issues as well:
- Same as [here](https://github.com/vaadin/flow/pull/11635#discussion_r695442774), it wouldn't clear the child-views of a removed parent views.
- It wouldn't invoke on removals inside shadow roots (a child view is removed inside the shadow root of a parent view)

caalador: If I just leave this to you so we don't run in circles spending time?

tomivirkki: Sorry, I don't have a direct answer to how to properly fix it either, so would need investigation in every case. I Suggested trying WeakMap earlier.

caalador: Well the meaning of this is to not reset something we already have set so if there is a unique id for the target storing that would be enough as it doesn't need  to be the element that is in the array...

But now is the question that does it even work as it should in gatekeeping the document additions? As now for instance with the same document.css we get it in 2 times as the test has the same embedded component 2 times and the current block only blocks the issue if you applyTheme for the same target.

caalador: Is 
```
  const injectGlobal = (window.Vaadin.Flow['${globalCssFlag}'].length === 0) || (!window.Vaadin.Flow['${globalCssFlag}'].includes(target) && target !== document);
  if (injectGlobal) {
    ${globalCssCode.join('')}
    window.Vaadin.Flow['${globalCssFlag}'].push(target);
  }
```
really in any way better than
```
  if (!target['${globalCssFlag}']) {
    ${globalCssCode.join('')}
    target['${globalCssFlag}'] = true;
```

tomivirkki: If you want to just avoid duplicate records in an array, using a Set is one option:
```js
array = [...new Set(array.concat(target))];
```

PR #11659, opened 2021-08-27
Organize dependency versions and update vulnerable jackson-databind version for fusion-endpoint

fluorumlabs:
## Description

This fix rearranges version properties in `pom.xml` files: common versions are moved from `flow-server` and `fusion-endpoint` up to the root `pom.xml`; unused properties are cleared

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/pom.xml
+++ b/flow-server/pom.xml
@@ -11,17 +11,6 @@
     <name>Flow Server</name>
     <packaging>jar</packaging>
 
-    <properties>
-        <validation.api.version>2.0.1.Final</validation.api.version>
-        <jackson.version>2.11.2</jackson.version>
-        <spring.version>5.2.0.RELEASE</spring.version>
-        <spring.autoconfigure.version>2.2.0.RELEASE</spring.autoconfigure.version>
-        <javax.annotation.api.version>1.3.2</javax.annotation.api.version>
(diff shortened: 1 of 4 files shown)


mshabarov (approved):
The proposed change looks good to me.
However some fusion tests are failing because of breaking changes in jackson library.
The timezone format changes, example:

```
Expected :"2019-01-02T00:00:00.000+0000"
Actual   :"2019-01-02T00:00:00.000+00:00"
```

So if it satisfies Fusion, then we can just change the expectations in tests, @haijian-vaadin .

fluorumlabs:
This is the root cause for time serialization change: https://github.com/FasterXML/jackson-databind/issues/2643
And this is the (temporary) workaround: https://github.com/FasterXML/jackson-databind/issues/2643#issuecomment-808319443

Practically it should be verified if JavaScript deserializer is ok with extended ISO-8601 format - we can just update failing tests

Jackson 2.12 also brings case-insensitive Boolean deserialization: https://github.com/FasterXML/jackson-databind/issues/1852, which is harmless but breaks tests (note that this PR is for 2.11)

fluorumlabs:
Quick test shown that JS is parsing colon-separated TZ correctly.

haijian-vaadin (approved):

PR #11673, opened 2021-08-30
fix: keep track of css injected to document (#11635)

### LINKED ISSUE #11608: Memory leak in theme generator

caalador:
Stops css/fonts from being loaded multiple times to the document with embedded applications.

Closes #11608

Contains fix #11669

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -243,7 +243,7 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
 
         /**
          * Should custom theme be initialized.
-         * @return true if theme should initialize
+         * @return true if theme should be initialized
          */
         public boolean isInitTheme() {
             return initTheme;
@@ -251,7 +251,7 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
 
         /**
          * Set if custom theme should be initialized.
-         * @param initTheme initializeTheme
+         * @param initTheme enable or disable theme initialisation
          */
         public void setInitTheme(boolean initTheme) {
             this.initTheme = initTheme;
@@ -624,7 +624,7 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
                 head.prependElement("script").attr("type", "text/javascript")
                         .appendChild(new DataNode(
                                 "window.Vaadin = window.Vaadin || {}; window.Vaadin.theme = window.Vaadin.theme || {};"
-                                        + "window.Vaadin.theme.flowBootstrap = {}"));
+                                        + "window.Vaadin.theme.flowBootstrap = true;"));
             }
 
             if (!config.isProductionMode()) {
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -186,6 +186,8 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
         private JsonObject applicationParameters;
         private BootstrapUriResolver uriResolver;
 
+        private boolean initTheme = true;
+
         /**
          * Creates a new context instance using the given parameters.
          *
@@ -237,6 +239,22 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
          */
         public VaadinSession getSession() {
             return session;
+        }
+
+        /**
+         * Should custom theme be initialized.
+         * @return true if theme should initialize
+         */
+        public boolean isInitTheme() {
+            return initTheme;
+        }
+
+        /**
+         * Set if custom theme should be initialized.
+         * @param initTheme initializeTheme
+         */
+        public void setInitTheme(boolean initTheme) {
+            this.initTheme = initTheme;
         }
 
         /**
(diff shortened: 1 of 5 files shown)


flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -82,7 +82,13 @@ const createLinkReferences = (css, target) => {
 const injectGlobalCssMethod = `
 // target: Document | ShadowRoot
 export const injectGlobalCss = (css, target, first) => {
-  
+  if(target === document) {
+    const hash = getHash(css);
```
fluorumlabs, line 86: I wonder if it would be any faster to use `Set` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) of `css`. I suppose native hashing would be faster then custom logic + it will handle collisions in a safe manner by design.

caalador: That would put the css into memory as a duplicate also.
Any changes should be in a ticket as this is just a cherrypick to 14 which will need changes due to theme handling without Fusion.

fluorumlabs: We have to choose between:
- Memory overhead of having css in Set
- Performance overhead of computing hash every time css is about to be injected into document

fluorumlabs: > Any changes should be in a ticket as this is just a cherrypick to 14 which will need changes due to theme handling without Fusion.

The chosen approach is not mentioned anywhere in the ticket so I consider it to be an implementation detail, hence commenting in a PR. But fill free to escalate it to the ticket/other involved parties

caalador: What I was saying is that this is already in in 7,8,9.
Also seeing as this is to fix memory issue the end impact is minimal as it's only for css targeting document and those are usually 1/theme in a project (so 2 for when there is a parent theme etc.)

fluorumlabs: Right. Well, nevermind then :)

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -239,6 +241,22 @@ public VaadinSession getSession() {
             return session;
         }
 
+        /**
+         * Should custom theme be initialized.
+         * @return true if theme should initialize
```
mshabarov, line 246:
```suggestion
         * @return true if theme should be initialized
```

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -239,6 +241,22 @@ public VaadinSession getSession() {
...
+         */
+        public boolean isInitTheme() {
+            return initTheme;
+        }
+
+        /**
+         * Set if custom theme should be initialized.
+         * @param initTheme initializeTheme
```
mshabarov, line 254:
```suggestion
         * @param initTheme enable or disable theme initialisation
```

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -600,6 +618,15 @@ public Document getBootstrapPage(BootstrapContext context) {
                 handleThemeContents(context, document);
             }
 
+            // To not init theme for webcomponents we use a flag on the dom
+            if (context.isInitTheme() && context.getTheme().isPresent()
+                    && !"".equals(context.getTheme().get().getName())) {
+                head.prependElement("script").attr("type", "text/javascript")
+                        .appendChild(new DataNode(
+                                "window.Vaadin = window.Vaadin || {}; window.Vaadin.theme = window.Vaadin.theme || {};"
+                                        + "window.Vaadin.theme.flowBootstrap = {}"));
```
mshabarov, line 627: Using `flowBootstrap = {}` makes me think that some data will be included into this field later on. However it's just a flag that the bootstrap request is for application, not for web component. Maybe its better to set it as `window.Vaadin.theme.flowBootstrap = true`?

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -600,6 +618,15 @@ public Document getBootstrapPage(BootstrapContext context) {
                 handleThemeContents(context, document);
             }
 
+            // To not init theme for webcomponents we use a flag on the dom
+            if (context.isInitTheme() && context.getTheme().isPresent()
```
mshabarov, line 622: I don't fully understand how memory leaks in injecting global CSS is connected with not init theme for web components, and moreover, why now we decide not to init the theme for web components? Is that a separate bug which also leads to memory leaks, i.e. injecting flow theme (global css) for web components is not needed and is an excessive memory consumption?

caalador: WebComponents initialize the theme through the WebComponentBootstrap but this is also run meaning that when we should add the theme only to the shadowroot of the webcomponent we also add it to the document which is not what should happen as only the webComponent should init the theme for itself.

mshabarov: Ah, right, so the `WebComponentBootstrapHandler` skips execution of `applyTheme(document)` via this flag in DOM, because `applyTheme(document)` is for "global" CSS. Thanks Mikael!

mshabarov (approved):
Tested with 14.7 Pre starter project with all types of views. Works as expected. Also checked that `window.Vaadin.theme.injectedGlobalCss` contains one hashed value between navigations and scrolling.

Issue #11599, opened 2021-08-18
NullPointerException in StateNode when setting locale and using UI.getCurrent() inside localeChange callback.


mrgreywater:
### Description of the bug / feature
NullPointerException when changing locale
### Minimal reproducible example

Sometimes when changing the locale when Push is enabled, Vaadin throws a NullPointerException. It's difficult to reproduce.

### Expected behavior
No Exception should be thrown
### Actual behavior
A NullPointer Exception is thrown because in StateNode::dumpBeforeClientResponseEntries() there is a check missing for `entries == null`

![image](https://user-images.githubusercontent.com/2902403/129920187-86a21320-4fee-46d4-a007-f9a8543cfc38.png)

```
java.lang.RuntimeException: Push failed
	at com.vaadin.flow.server.communication.AtmospherePushConnection.push(AtmospherePushConnection.java:200) ~[flow-server-7.0.6.jar:7.0.6]
	at com.vaadin.flow.server.communication.AtmospherePushConnection.push(AtmospherePushConnection.java:175) ~[flow-server-7.0.6.jar:7.0.6]
	at com.vaadin.flow.component.UI.push(UI.java:686) ~[flow-server-7.0.6.jar:7.0.6]
	at com.vaadin.flow.server.VaadinSession.unlock(VaadinSession.java:708) ~[flow-server-7.0.6.jar:7.0.6]
	at com.vaadin.flow.server.VaadinService.requestEnd(VaadinService.java:1474) ~[flow-server-7.0.6.jar:7.0.6]
	at com.vaadin.flow.server.VaadinService.handleRequest(VaadinService.java:1549) ~[flow-server-7.0.6.jar:7.0.6]
	at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:299) ~[flow-server-7.0.6.jar:7.0.6]
	at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:109) ~[vaadin-spring-17.0.2.jar:na]
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:733) ~[tomcat-embed-core-9.0.45.jar:4.0.FR]
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:227) ~[tomcat-embed-core-9.0.45.jar:9.0.45]
[...]

PR #11726, opened 2021-09-02
fix: check the component belongs to UI before task execution

mshabarov:
## Description

Adds an assumption that the given component belongs to UI instance on which the task is going to be executed. This is needed to fail-fast and detect errors earlier before the actual task is executed and throws because of inconsistency in `StateTree`.

Related-to #11599

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -995,6 +995,8 @@ public Router getRouter() {
      *
      * @return a registration that can be used to cancel the execution of the
      *         task
+     * @throws IllegalArgumentException
+     *             if the given component doesn't belong to this UI
      */
     public ExecutionRegistration beforeClientResponse(Component component,
```
denis-anisimov, line 1001: the method should have `throws`: javadoc is not necessary but if you extend javadocs then the method should `throws` .

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/UI.java
mshabarov: Done

denis-anisimov (approved):

PR #11729, opened 2021-09-02
Improve generated code for Pageable

Artur-:
Improve generated code for Pageable
- Importing Pageable from 'mappedtypes' instead of 'endpointransfermapper' looks more sensible
- Importing Pageable instead of PageableDTO is more like the expectations

### DIFF
--- a/fusion-endpoint/src/main/java/com/vaadin/fusion/endpointransfermapper/PageableMapper.java
+++ b/fusion-endpoint/src/main/java/com/vaadin/fusion/endpointransfermapper/PageableMapper.java
@@ -41,19 +41,20 @@ public class PageableMapper
     @Override
     public Pageable toTransferType(
             org.springframework.data.domain.Pageable pageable) {
-        Pageable dto = new Pageable();
-        dto.setPageNumber(pageable.getPageNumber());
-        dto.setPageSize(pageable.getPageSize());
-        dto.setSort(sortMapper.toTransferType(pageable.getSort()));
+        Pageable transferPageable = new Pageable();
+        transferPageable.setPageNumber(pageable.getPageNumber());
+        transferPageable.setPageSize(pageable.getPageSize());
+        transferPageable.setSort(sortMapper.toTransferType(pageable.getSort()));
 
-        return dto;
+        return transferPageable;
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


fusion-endpoint/src/main/java/com/vaadin/fusion/endpointransfermapper/PageableMapper.java
```diff
@@ -16,31 +16,32 @@
...
+    public Class<? extends Pageable> getTransferType() {
+        return Pageable.class;
     }
 
     @Override
-    public PageableDTO toTransferType(Pageable pageable) {
-        PageableDTO dto = new PageableDTO();
+    public Pageable toTransferType(
+            org.springframework.data.domain.Pageable pageable) {
+        Pageable dto = new Pageable();
```
haijian-vaadin, line 44: rename to `mappedPageble` or sth like that?

fusion-endpoint/src/main/java/com/vaadin/fusion/endpointransfermapper/PageableMapper.java
```diff
@@ -49,8 +50,10 @@ public PageableDTO toTransferType(Pageable pageable) {
     }
 
     @Override
-    public Pageable toEndpointType(PageableDTO dto) {
-        Sort sort = sortMapper.toEndpointType(dto.getSort());
+    public org.springframework.data.domain.Pageable toEndpointType(
+            Pageable dto) {
```
haijian-vaadin, line 54: rename to `mappedPageble` or sth like that?

haijian-vaadin (changes requested):

fusion-endpoint/src/main/java/com/vaadin/fusion/endpointransfermapper/PageableMapper.java
Artur-: Done

fusion-endpoint/src/main/java/com/vaadin/fusion/endpointransfermapper/PageableMapper.java
Artur-: Done

haijian-vaadin (approved):

Issue #11713, opened 2021-09-01
Vaadin Starter App not run propaly in IE11 (development and production mode)


rhonibaer:
### Description of the bug / feature
Downloaded a fresh starter-app, with mostly all demo pages.
only started with "mvnw" and container runs up.
open now localhost:8080 in IE11

### Minimal reproducible example
simple open the app  localhost:8080 in IE11

### Expected behavior
Mostly all components are not rendered correctly. see attachments

### Actual behavior
![screen-1-ie11](https://user-images.githubusercontent.com/26479653/131649637-b9b3b0a3-fc82-477d-9322-18163ca27d8f.PNG)
![screen-2-ie11](https://user-images.githubusercontent.com/26479653/131649652-4e9a050c-8bc8-43a2-abc7-1c9e029d1d28.PNG)

### Versions:
    - Vaadin / Flow version: 14.6.8
    - Java version: Corretto 8
    - OS version: windows 10
    - Browser version (if applicable): IE11
    - Application Server (if applicable): 
    - IDE (if applicable): IntelliJ ultimate

mshabarov:
Judging by the errors in the console, the grid connector (`gridConnector.js`) wasn't initialised/loaded at all.
Is this observed only in IE11?

mshabarov:
I'm able to reproduce it as well. For Card List view, I don't see any type errors, but the components are not rendered at all.
Tried with `mvn -Dproduction` and `mvn -Dvaadin.devmode.transpile=true`.

caalador:
A quick test would say that the new theme feature is not working with IE11.
So a suggestion would be to use `@CssImport` and `@StyleSheet` when targeting IE11

caalador:
I noted that as soon as I added the map view using LeafletMap then no components rendered anymore and the console gets a `SCRIPT1002: Syntax error`.

With the new theme feature there is the console exceptions:
`message"Unable to get property 'appendChild' of undefined or null reference"`, `Could not find style data in module named lumo-spacing` and `SCRIPT5007: Unable to get property 'querySelectorAll' of undefined or null reference`

One issue that comes up with the map is that it is a .ts file which actually is not transpiled, but I can't get a configuration for transpilation of the `@customElement` in the ts file. Else without the theme the components work as they should.

For the ticket view selection I can see that ImageList and CheckoutForm utilise .ts files

So to fix the usage with no theme we should find a way to get the ts files transpiled

caalador:
[...]

PR #11792, opened 2021-09-08
fix: Fixes TypeScript transpilation to ES5

caalador:
Fix the transpilation of TypeScript files
to ES5 for use with IE11

Part of #11713


flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
```diff
@@ -182,6 +182,11 @@ private static String getSourceFromObject(JsonObject module,
                 && validKey(module, SOURCE, STRING)) {
             String name = module.getString(NAME);
 
+            // If the found module is
```
mshabarov, line 185: is ES5 module?

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -308,10 +308,12 @@ static void addVaadinDefaultsToJson(JsonObject json) {
         defaults.put("webpack", "4.42.0");
         defaults.put("webpack-cli", "3.3.11");
         defaults.put("webpack-dev-server", "3.11.0");
-        defaults.put("webpack-babel-multi-target-plugin", "2.3.3");
+        defaults.put("webpack-babel-multi-target-plugin", "2.5.0");
```
mshabarov, line 311: So the transpilation doesn't work in 2.3.3, but works in 2.5.0?

flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
caalador: It's what we have in BundleParser and it did leave me a bit confused. Elaborated on comment.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
caalador: The new configuration would seem to work with 2.3.3 also

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
mshabarov: ok, so this version bump is not needed for this fix?
I have nothing against versions update, let's upgrade it anyway.

PR #11849, opened 2021-09-14
fix: allow listener removal via DataChangeEvent

### LINKED ISSUE #6947: DataChangeEvent can't unregister its listener
`ComponentEvent`s have a useful `unregisterListener` method to be able to unregister themselves once invoked, but `DataChangeEvent` does not; it would be useful to set a "once" listener (not sure if there's another way to do it).

denis-anisimov:
fixes #6947

## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataChangeEvent.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataChangeEvent.java
@@ -18,6 +18,8 @@ package com.vaadin.flow.data.provider;
 import java.util.EventObject;
 import java.util.Objects;
 
+import com.vaadin.flow.server.Command;
+
 /**
  * An event fired when the data of a {@code DataProvider} changes.
  *
@@ -30,6 +32,8 @@ import java.util.Objects;
  *            the data type
  */
 public class DataChangeEvent<T> extends EventObject {
+
+    private Command unregisterListenerCommand = null;
 
     /**
      * An event fired when a single item of a {@code DataProvider} has been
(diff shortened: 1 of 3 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/DataChangeEvent.java
```diff
@@ -112,4 +116,31 @@ public DataChangeEvent(DataProvider<T, ?> source) {
...
+     *            the unregister command
+     */
+    void setUnregisterListenerCommand(Command unregisterListenerCommand) {
+        this.unregisterListenerCommand = unregisterListenerCommand;
+    }
+
+    /**
+     * Unregisters the event listener currently being invoked.
+     * <p>
+     * This method can only be called from within an event listener. Calling it
```
mshabarov, line 135: "can only be called" doesn't explain clearly what happens if devs do call it outside the listener, so I'd propose to reflect in javadoc that this method throws when called outside of the listener.

flow-data/src/test/java/com/vaadin/flow/data/provider/AbstractDataProviderTest.java
```diff
@@ -75,4 +73,30 @@ public void removeListener_listenerIsNotNotified() {
         dataProvider.refreshAll();
         Assert.assertNull(event.get());
     }
+
+    @Test
+    public void eventUnregisterListener_insideListener() {
```
mshabarov, line 78: Add test `eventUnregisterListener_outsideListener()`

mshabarov: I'd add one more test which adds a listener which throws an exception, and then check that unregister listener command in `null` despite of the exception thrown.

mshabarov (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/DataChangeEvent.java
denis-anisimov: Copied from here : https://github.com/vaadin/flow/blob/b6cf0b5ea13f9a7557853f480b7e49806d40f387/flow-server/src/main/java/com/vaadin/flow/component/ComponentEvent.java#L84

Would be better to have them consistent....
But OK.

flow-data/src/test/java/com/vaadin/flow/data/provider/AbstractDataProviderTest.java
denis-anisimov: done and done

mshabarov (approved):

PR #11860, opened 2021-09-15
fix: add possibility to set header values for StreamResource

### LINKED ISSUE #5471: StreamResource should have method for setting "Content-Disposition" response header
StreamResource currently have methods for setting response headers for contentType and cacheTime.

It would be useful if we could set content disposition also. This could, for example, be used to force download of dynamic content (instead of opening it in browser) with header: Content-Disposition: attachment; filename="filename.jpg".

Content disposition could be set with dedicated method ```setContentDisposition(String value)``` or more generic ```setHeader(String name, String value)```?

denis-anisimov:
fixes #5471

## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/StreamResource.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/StreamResource.java
@@ -245,7 +245,8 @@ public class StreamResource extends AbstractStreamResource {
      * 
      * @param name
      *            name of header to get value for
-     * @return an optional with header value, or an empty optional if it has not been set
+     * @return an optional with header value, or an empty optional if it has not
+     *         been set
      */
     public Optional<String> getHeader(String name) {
         if (headers != null) {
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/StreamResource.java
```diff
@@ -217,6 +222,53 @@ public ContentTypeResolver getContentTypeResolver() {
...
+        return this;
+    }
+
+    /**
+     * Gets the header {@code name} for the resource.
+     * 
+     * @param name
+     *            a header name
+     * @return an optional header, or an empty optional if it has not been set
+     */
```
caalador, line 249:
```suggestion
    /**
     * Gets the value for header {@code name} set for the resource.
     * 
     * @param name
     *            name of header to get value for
     * @return an optional with header value, or an empty optional if it has not been set
     */
```

caalador (approved):
Minor javadoc change that would make it a bit clearer on a read

caalador (dismissed):

caalador (approved):

PR #11892, opened 2021-09-20
feat: allow to customize exceptions thrown inside Binder

### LINKED ISSUE #4531: Enhancement request: Include field name in binder/field stack traces
In flow, if an exception occurs because of null value, incorrect combobox values or similar, the stack trace doesn't provide any information about which field caused the error. This can result in lots of debugging to narrow down which field caused the issue. 

It would be great if Flow at a minimum can include the field name/label/id in the stack trace.

stacktrace 1

denis-anisimov:
fixes #4531

## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-data/src/main/java/com/vaadin/flow/data/binder/DefaultBindingExceptionHandler.java
```diff
@@ -0,0 +1,108 @@
...
+import com.vaadin.flow.component.HasElement;
+import com.vaadin.flow.component.HasValue;
+import com.vaadin.flow.component.UI;
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.internal.NodeOwner;
+import com.vaadin.flow.internal.StateTree;
+import com.vaadin.flow.server.startup.ApplicationConfiguration;
+
+/**
+ * Default implementation of {@link DefaultBindingExceptionHandler}.
```
mshabarov, line 29:
```suggestion
 * Default implementation of {@link BindingExceptionHandler}.
```

flow-data/src/main/java/com/vaadin/flow/data/binder/DefaultBindingExceptionHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private UI getUI(Element element) {
+        NodeOwner owner = element.getNode().getOwner();
+        UI ui = null;
+        if (element.getNode().isAttached()) {
+            ui = ((StateTree) owner).getUI();
+        }
+        if (ui == null) {
+            ui = UI.getCurrent();
```
mshabarov, line 96: I don't get the meaning of traversing elements tree, if it finally falls into `ui = UI.getCurrent();` anyway.
Could you please explain why you don't just use `UI.getCurrent();` instead of `getUI`?

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -3201,4 +3239,28 @@ public void setValidatorsDisabled(boolean validatorsDisabled) {
     public boolean isValidatorsDisabled() {
         return validatorsDisabled;
     }
+
+    /**
+     * Sets a {@code handler} to customize the {@link RuntimeException} thrown
+     * by delegates (like {@link Setter}, {@link ValueProvider},
+     * {@link HasValue}) used inside {@link Binder} to be able to identify the
+     * bound field.
```
mshabarov, line 3247: I propose to add few words on when the developers would need a custom handler, in which cases. Or at least give a link to the default implementation and say something like "Binder's API allows one to set a custom binding exception handler in case of exception details provided be the default handler are not enough, or, in other way around, it should be reduced".

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
```diff
@@ -1738,6 +1739,173 @@ public void clearError(HasValue<?, ?> field) {
         Assert.assertSame(handler, binder.getValidationErrorHandler());
     }
 
+    @Test(expected = BindingException.class)
+    public void readBean_converterThrows_readBean_exceptionHandlerSet_bindingExceptionIsThrown() {
+        TestTextField testField = new TestTextField();
+        setExceptionHandler();
```
mshabarov, line 1745: All these newly added tests set a custom exception handler.
It seems one more test is needed which would check that the Binder's default handler is used by default.

mshabarov (commented):
Would be useful to mention this feature in the Bender's documentation as well https://vaadin.com/docs/latest/flow/binding-data/components-binder-beans.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
denis-anisimov: done

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
denis-anisimov: the most dumb test ever is added

denis-anisimov:
> Would be useful to mention this feature in the Bender's documentation as well https://vaadin.com/docs/latest/flow/binding-data/components-binder-beans.

One ticket/PR per repository.
If something needs to be done then please just create a ticket in the proper repository.

flow-data/src/main/java/com/vaadin/flow/data/binder/DefaultBindingExceptionHandler.java
mshabarov: Thanks for explanations.
Due to the method names are the same - `getUI`, I erroneously concluded that the method calls itself recursively but in fact it calls `getUI` of the owner.
 
> I would recommend to avoid questions unrelated to the review in the review. I see too many of them recently and they are blocking the PR unnecessary.

For me it's a new code and I don't always know the patterns used in the framework.
I believe it's better to ask rather that silently skip it, not having the idea how it works.

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
mshabarov: If you need more dumb tests, let me know :)

mshabarov (approved):

flow-data/src/main/java/com/vaadin/flow/data/binder/DefaultBindingExceptionHandler.java
denis-anisimov: I don't agree with this since the purpose of the review is not to learn the code by the reviewer.
It's just a wrong usage of the tool.
It's OK when the code is complicated and when it happens rarely.
But it happens regularly. And it's not that complicated since it's possible to spend some time and find out the answer by yourself via reading and checking the code: it's available.

PR #11898, opened 2021-09-21
fix: upgrade to lit 2.0.0

Artur-:


Haprog (dismissed):

haijian-vaadin (changes requested):
why it is fix?
Besides, we need to update the fusion and common frontend package as well, I will get those done and update this PR.

Haprog (approved):

Haprog:
The status of "Flow Pull Requests Run Validation (Flow Public)" build seems to be stuck but the build has in fact passed as green. Can we force that to recheck the status somehow? Also I think the change request by @haijian-vaadin is still blocking so I think he needs to approve even though latest changes were from him.

haijian-vaadin:
> The status of "Flow Pull Requests Run Validation (Flow Public)" build seems to be stuck but the build has in fact passed as green. Can we force that to recheck the status somehow? Also I think the change request by @haijian-vaadin is still blocking so I think he needs to approve even though latest changes were from him.

I blocked the PR on purpose, want to try it out on a starter project manually to see if the `peerdependency` change on the fusion-frontend works on both npm 7 and 6. Will approve it once it's verified.

haijian-vaadin (approved):

Issue #11730, opened 2021-09-02
FrontendUtils fails to find pnpm if only the newer version is installed


martaolszewskatusiek:
### Minimal reproducible example
1. Download Fusion app from start.vaadin.com for v20.
2. Have only pnpm version 6 available in system's path (in my case, 6.0.1).
3. Run `mvn spring-boot:run`.

### Expected behavior

mvn goal should run without issues, as pnpm 6 is more than the required minimum version of pnpm 5.

### Actual behavior

Error: `java.lang.IllegalStateException: Found too old 'pnpm'. If installed into the project 'node_modules', upgrade 'pnpm' to at least 5.0.0`

### Versions:
    - Vaadin / Flow version: v20
    - Java version: irrelevant
    - OS version: irrelevant
    - IDE (if applicable): Eclipse version: 2019-03 (4.11.0)

### Possible fix
The command used to check for pnpm version, `c:\_node_modules\npx.cmd --yes --quiet pnpm@5 --version` should probably be `c:\_node_modules\npx.cmd --yes --quiet pnpm@5+ --version`. In the current form it produces an empty string when only a newer version is available.

caalador:
For a non accepted version there should be an info log `pnpm '{}' is version {} which is not supported (expected >={})` is that available and what does it say

mshabarov:
I can't reproduce this issue on Mac OS.
Flow prints the following info message
`INFO 3792 --- [onPool-worker-9] c.v.flow.server.frontend.FrontendTools   : using '/Users/mikhail/.nvm/versions/node/v14.15.5/bin/npx --yes --quiet pnpm@5' for frontend package installation`
And
`/Users/mikhail/.nvm/versions/node/v14.15.5/bin/npx --yes --quiet pnpm@5 --version` gives `5.18.10`.

tarekoraby:
> For a non accepted version there should be an info log `pnpm '{}' is version {} which is not supported (expected >={})` is that available and what does it say

For me, the message exsists, and it says `pnpm 'C:\Program Files\nodejs\npx.cmd --yes --quiet pnpm@5' is version 4.5.0 which is not supported (expected >=5.0.0)`.

tarekoraby:
Which is confusing since `npm view pnpm version` indicates that my pnpm version is 6.14.5.

caalador:
@tarekoraby what does `"C:\Program Files\nodejs\npx.cmd" --yes --quiet pnpm@5 -v` say?

[...]

PR #11905, opened 2021-09-22
fix: Use proper npm/pnpm commands for Windows

mshabarov:
## Description

Related-to https://github.com/vaadin/flow/issues/11730

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -749,7 +749,9 @@ private String getNonNull(String... valueArray) {
         if (useGlobalPnpm) {
             // try to locate already installed global pnpm, throw an exception
             // if pnpm not found or its version is too old (< 5).
-            pnpmCommand = frontendToolsLocator.tryLocateTool("pnpm")
+            pnpmCommand = frontendToolsLocator
+                    .tryLocateTool(
+                            FrontendUtils.isWindows() ? "pnpm.cmd" : "pnpm")
```
denis-anisimov, line 754: `NpmCliTool` already contains something similar.
It makes sense to update this class and use it here.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
mshabarov: True. But I didn't do that because the name of this enum is `NpmCliTool` and I wasn't sure about pnpm cli.
But okay, this enum can be renamed and extended with pnpm. Will fix.

mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -152,6 +152,10 @@ String getCommand() {
         }
 
         String getScript() {
+            if (script == null) {
```
denis-anisimov, line 155: Private class....
It's too much for the private class.

I suggest to add `Objects.requireNotNull` in the CTOR and remove this exception from here.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
mshabarov: Adding `Objects.requireNotNull` would means quite different thing.
Pnpm cli javascript file is not applicable, it is not provided (at least it is not provided with the default installation and I didn't find any mentions of this JS file in the docs), so I added `null` into the script parameter. It can be `null` because it is not applicable for pnpm.
Then, If someone tries to get the script for pnpm, it would throw an exception saying that this is not supported.

denis-anisimov (approved):

denis-anisimov (approved):

PR #11906, opened 2021-09-22
fix: call processStyles when adding custom-style blocks

### LINKED ISSUE #11877: Lumo Utility Classes are not applied in production mode for LitElement "views"
### Description of the bug

Lumo utility classes are not applied in `production` mode in a `flow` project that has a `lit element` as view.

This does NOT happen with `flow` components as view NOR in development mode. In development `lit element` views look good as well.

 - Download given [example.zip](https://github.com/vaadin/flow/files/7178334/vaadin-bug.zip); created with start.vaadin.com

 - Build in production mode
 - Open localhost:8080 -> see black heading

### Expected behavior

Lumo utility classes work in `production` mode. (like in development mode)

image

### Actual behavior

Lumo utility classes are not loaded / do not work.

image

### Versions:
    - Vaadin / Flow version: 21.0.1
    - Java version: ---
    - OS version: ---
    - Browser version (if applicable): Only Tested with Firefox
    - Application Server (if applicable): ---
    - IDE (if applicable): ---

caalador:
Always when adding custom-style blocks with
module styles we should call the processStyles
to actually add style data.

Fixes #11877


Artur-:
Test?

caalador:
Can't get it to break in the theme test module. So it would need to have a application with no polymer components?

Artur-:
Yes, you cannot load any polymer based components nor the dev mode detector..

knoobie:
Just wondering: Shouldn't the theme have a dependency on the `ShadyCss`-thing?

denis-anisimov (dismissed):

caalador:
`@vaadin/vaadin-lumo-styles` depend on `@polymer/polymer` which depends on `@webcomponents/shadycss` so it's always available when lumo is used.

denis-anisimov (approved):

denis-anisimov (approved):

PR #11907, opened 2021-09-22
fix: auto register Vaadin servlet even if there is no routes

### LINKED ISSUE #4850: Validate automatic creation of main application servlet
After DynamicRoutes we should validate if a main application servlet should possibly be created even if we have no registered routes in the GlobalRouteRegistry.

The current implementation will not try to create a main application servlet if no Routes have been registered. With dynamic routes there is a possibility of adding the routes later or only to the session scope.

denis-anisimov:
fixes #4850

## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


caalador (dismissed):

caalador:
Conflicting files
flow-server/src/test/java/com/vaadin/flow/server/startup/ServletDeployerTest.java

caalador:
```
/opt/agent/work/a69251da2b2bb9e9/flow-server/src/test/java/com/vaadin/flow/server/startup/ServletDeployerTest.java:[118,37] method getContextEvent in class com.vaadin.flow.server.startup.ServletDeployerTest cannot be applied to given types;
  required: javax.servlet.ServletRegistration[]
  found: boolean,boolean,javax.servlet.ServletRegistration
  reason: varargs mismatch; boolean cannot be converted to javax.servlet.ServletRegistration
```

denis-anisimov:
done

flow-server/src/test/java/com/vaadin/flow/server/startup/ServletDeployerTest.java
```diff
@@ -381,25 +320,13 @@ private ServletRegistration getServletRegistration(String servletName,
...
+    private void assertservletsAreNotRegistered_registeredServlets(
             boolean compatibilityMode) throws Exception {
-        deployer.contextInitialized(getContextEvent(false, false,
+        deployer.contextInitialized(getContextEvent(
                 getServletRegistration("testServlet", TestServlet.class,
                         singletonList("/test/*"), Collections.emptyMap())));
 
-        assertMappingsCount(0, 0);
+        assertMappingsCount(1, 1);
     }
```
caalador, line 330: Doesn't this now assert that a servlet is registered and not no servlets registered. 
Also the compatibilityMode boolean doesn't do anything in the method so it seems a bit non functional, and makes
`registeredServlets_compatibilityMode_servletsAreNotRegistered` and `registeredServlets_servletsAreNotRegistered` the same test, so I'm not certain what these actually test.

caalador (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/startup/ServletDeployerTest.java
denis-anisimov: Yeah, a number of confusions.

compatibility mode is not checked anymore anyhow. This is a leftover from previous versions where `"frontend"` URL mapping servlet was registered as well.

The test has a precondition with non vaddin servlets and it checks that Vaadin servlet is registered. And previously the precondition was also no routes and no web components , that's why it checked that Vaadin servlet is not registered.
Now it should check that it's registered.
I've changed the names.

caalador (approved):

PR #11919, opened 2021-09-23
feat: add WhiteSpace setter and getter methods

### LINKED ISSUE #4577: Add a Java API for configuring white-space styles
When showing text inside any component or element, it's sometimes desirable to make it honor line breaks and other whitespace from the original String, so that e.g. `button.setText("Hello\nworld");` would be rendered on two lines.

There are currently two bad ways of achieving this:
1. Show the contents as HTML, e.g` button.add(new HTML("Hello<br>world"));`. The big drawback of this practice is that it opens up the application for XSS attacks for any text that may originate from users.
2. Use the Element API to set a suitable `white-space` style for the component, e.g. `button.getElement().getStyle().set("white-space", "pre-line");`. This approach requires that you are aware of this CSS feature and that you know which setting to actually use (the spec defines 5 different potential values).

I propose that `HasText` is enhanced with a typesafe API for configuring the `white-space` style based on an enum. With this, the example would be `button.setWhiteSpace(WhiteSpace.PRE_LINE);`. We could also consider an overload of the `setText`  method that also sets a white space mode: `button.setText("Hello\nWorld", WhiteSpace.PRE_LINE);`, and then maybe also supplement various String constructors that delegate to `setText`.

denis-anisimov:
fixes #4577

## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/HasText.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/HasText.java
@@ -32,8 +32,62 @@ import java.util.stream.Stream;
  */
 public interface HasText extends HasElement {
 
+    /**
+     * Represents <code>"white-space"</code> style values.
+     * 
+     * @author Vaadin Ltd
+     * @since
+     *
+     */
     enum WhiteSpace {
-        NORMAL, NOWRAP, PRE, PRE_WRAP, PRE_LINE, BREAK_SPACES, INHERIT, INITIAL;
+        /**
+         * Sequences of white space are collapsed. Newline characters in the
+         * source are handled the same as other white space. Lines are broken as
+         * necessary to fill line boxes.
+         */
+        NORMAL,
+        /**
+         * Collapses white space as for normal, but suppresses line breaks (text
+         * wrapping) within the source.
+         */
+        NOWRAP,
+        /**
+         * Sequences of white space are preserved. Lines are only broken at
+         * newline characters in the source and at &lt;br&gt; elements.
+         */
+        PRE,
+        /**
+         * Sequences of white space are preserved. Lines are broken at newline
+         * characters, at &lt;br&gt;, and as necessary to fill line boxes.
+         */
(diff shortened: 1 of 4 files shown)


knoobie:
Personally, I don't really see the need of this addition; it bloats the `HasText` interface and the `getWhiteSpace` does not really give the precise value (somebody would expect; calculated from the browser) instead it returns the value that was given by `set` or a default value that could even be wrong if styled with CSS.

mshabarov:
> the getWhiteSpace does not really give the precise value (somebody would expect; calculated from the browser) instead it returns the value that was given by set or a default value that could even be wrong if styled with CSS.

If it behaves like this, then maybe better to not add `getWhiteSpace` at all, or describe it's limitation in javadoc.

mshabarov (approved):
I'm not familiar with the `white-space` CSS properties that much, but the new API look good to me.

mshabarov:
Added Leif as he was the author of this enhancement.

Legioth:
The implementation seems to be in line with what I have expected. My only remark is that there's some new API in the enum that is implicitly public that doesn't have javadocs, but that's up to the team do decide what policy you have for such cases.

flow-server/src/main/java/com/vaadin/flow/component/HasText.java
```diff
@@ -28,6 +32,21 @@
  */
 public interface HasText extends HasElement {
 
+    enum WhiteSpace {
```
mshabarov, line 35: Add javadoc

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/HasText.java
denis-anisimov: done

mshabarov (approved):

PR #10582, opened 2021-04-07
fix: use upgraded Atmosphere version for Push

denis-anisimov:
fixes #10103

PR #11934, opened 2021-09-27
chore: Download Atmosphere JS instead of having it in the project

Artur-:


denis-anisimov:
So it passes the all tests ?

How does it differ from https://github.com/vaadin/flow/pull/10582 which has a disabled IT to pass the validation?

Artur-:
This does not upgrade Atmosphere, it just generates the JS in the build instead of having a copy in the project

denis-anisimov:
Ah, I confused it with another PR. OK.

denis-anisimov (approved):

PR #11938, opened 2021-09-27
feat: use array of values for @Route instead of only one

### LINKED ISSUE #11915: Rewrite Route related functionality to get rid of Route.NAMING_CONVENTION
Ugly `Route.NAMING_CONVENTION` is an indicator which is used to use a hardcoded strategy to generate a navigation target path for a `@Route` component.

This functionality relies in the fragile assumption that nobody is going to use `Route.NAMING_CONVENTION` value as a real path for the route. Even though it's quite improbable situation such reliance smells.

I suggest to make an improvement and rewrite this :
**let's use `String[]` as a type for `value`**

That gives three wins immediately:
- Java allows to use _one single_ String instead if array with one element in the annotation values. It means that all the existing code will be backward compatible.
- The very first path will be used as usual : main route. All other will be used as aliases. That allows to reduce `@RouteAlias` annotation usage: in most cases you don't need different configs for aliases but instead just list paths as aliases and it's better to use one annotation with array value instead of bunch of `@RouteAlias`.
- An empty array of paths is a clear indication that no path are provided and this indicator can be used to auto generate the path instead of  `Route.NAMING_CONVENTION` indicator.

Similar issue is https://github.com/vaadin/flow/issues/4117.

denis-anisimov:
fixes #11915

## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/router/RouteConfiguration.java
```diff
@@ -229,17 +231,26 @@ public void setAnnotatedRoute(Class<? extends Component> navigationTarget) {
...
         handledRegistry.setRoute(route, navigationTarget,
                 RouteUtil.getParentLayouts(navigationTarget, route));
 
-        for (RouteAlias alias : navigationTarget
-                .getAnnotationsByType(RouteAlias.class)) {
+        String[] value = routeAnnotation.value();
+        Stream<RouteAliasObject> aliases = Stream.of(value).skip(1)
+                .map(path -> new RouteAliasObject(path,
+                        routeAnnotation.layout(), routeAnnotation.absolute()));
+        aliases = Stream.concat(aliases,
```
mshabarov, line 245: Route aliases coming from RouteAlias annotation and from Route value can be cross checked for duplicates and warn the developer about that. Don't know whether having duplicates breaks anything in routing feature though.

Also, the concatenation code is present more than once, and it can be placed into util class.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
```diff
@@ -164,11 +188,11 @@ public static String getRouteAliasPath(Class<?> component,
         return list;
     }
 
-    static Optional<RouteAlias> getMatchingRouteAlias(Class<?> component,
-            String path, List<RouteAlias> routeAliases) {
-        return routeAliases.stream().filter(
+    static Optional<RouteAliasObject> getMatchingRouteAlias(Class<?> component,
```
mshabarov, line 191: I propose to retain the old method and deprecate it. Even though this is an internal util class, it still can be used by starter or others.

flow-server/src/main/java/com/vaadin/flow/router/RouteConfiguration.java
denis-anisimov: Yes, they can be checked and they are checked.
They are checked below, inside the `handledRegistry.setRoute` 

```
at com.vaadin.flow.router.internal.RouteSegment.ambigousException(RouteSegment.java:805)
	at com.vaadin.flow.router.internal.RouteSegment.ambigousTarget(RouteSegment.java:801)
	at com.vaadin.flow.router.internal.RouteSegment.setRouteTarget(RouteSegment.java:487)
	at com.vaadin.flow.router.internal.RouteSegment.addSubRoute(RouteSegment.java:458)
	at com.vaadin.flow.router.internal.RouteSegment.addSubRoute(RouteSegment.java:267)
	at com.vaadin.flow.router.internal.RouteModel.addRoute(RouteModel.java:131)
	at com.vaadin.flow.router.internal.ConfigureRoutes.setRoute(ConfigureRoutes.java:175)
	at com.vaadin.flow.router.internal.AbstractRouteRegistry.lambda$9(AbstractRouteRegistry.java:315)
	at com.vaadin.flow.router.internal.AbstractRouteRegistry.lambda$14(AbstractRouteRegistry.java:353)
	at com.vaadin.flow.router.internal.AbstractRouteRegistry.configure(AbstractRouteRegistry.java:98)
	at com.vaadin.flow.router.internal.AbstractRouteRegistry.configureWithFullTemplate(AbstractRouteRegistry.java:353)
	at com.vaadin.flow.router.internal.AbstractRouteRegistry.setRoute(AbstractRouteRegistry.java:313)
	at com.vaadin.flow.server.startup.ApplicationRouteRegistry.setRoute(ApplicationRouteRegistry.java:326)
```

This functionality already exists for `@RouteAlias` and it's just reused as is.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
denis-anisimov: This method is internal , inside `internal` package and should not be used since `@RouteAlias` annotation is not the only source of the aliases anymore.

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/EndPointData.java
```diff
@@ -54,7 +53,7 @@
     public String toString() {
         return String.format(
                 "%n view: %s%n route: %s%n%s%n layout: %s%n modules: %s%n scripts: %s%n css: %s%n",
-                name, route, theme, layout, col2Str(modules), col2Str(scripts),
+                name, theme, layout, col2Str(modules), col2Str(scripts),
```
mshabarov, line 56: Hmm, why is the route field was removed?
If it makes sense, then let's remove the '%s' leftover in the format string.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/RouterIT.java
```diff
@@ -93,7 +93,7 @@ public void faultyRouteShowsExpectedErrorScreen() {
     @Test
     public void routeWithRouteAliasHasNoParents() {
         openRouteUrl(RouterTestServlet.AliasLayout.class
-                .getAnnotation(Route.class).value());
+                .getAnnotation(Route.class).value()[0]);
```
mshabarov, line 96: I wonder whether this change is still a breaking change (formally), so anyone who calls `Route::value` in his code would not be able to do it anymore. However this should be rare case (or not the case for customer's projects at all), but I still think this change should go with `feat!:` and should be reflected in the release notes. What do you think?

denis-anisimov:
It is possible to write comments and publish them in the very end when review is completed.
It helps to other counterparts do not be interrupted every time when new round of review randomly comes.
The current way of review when nothing is changed but comments are coming all the time requires too much attention and very unproductive since I have to switch the context from the extremely complicated tasks to fix the review comments several times instead doing this once.

PR #11942, opened 2021-09-28
feat: add Object Component to html components module

### LINKED ISSUE #4303: Add Java API for <object>
Add Java API for `<object>` into `com.vaadin.flow.component.html`.
AFAIK currently developers would need to introduce an object component themselves, e.g.

```java
@Tag("object")
public class Object extends HtmlComponent
```
and write API like `setSource` themselves.

Once the API has been added, update the corresponding row in the migration table:
https://github.com/vaadin/flow-and-components-documentation/blob/master/documentation/migration/5-components.asciidoc

denis-anisimov:
fixes #4303

## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
@@ -1,0 +1,255 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.html;
+
+import java.util.Optional;
+
+import com.vaadin.flow.component.ClickNotifier;
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.Focusable;
+import com.vaadin.flow.component.HasOrderedComponents;
+import com.vaadin.flow.component.HtmlContainer;
+import com.vaadin.flow.component.PropertyDescriptor;
+import com.vaadin.flow.component.PropertyDescriptors;
+import com.vaadin.flow.component.Tag;
+import com.vaadin.flow.server.AbstractStreamResource;
+import com.vaadin.flow.server.StreamResource;
+
+/**
+ * Component representing a <code>&lt;object&gt;</code> element.
+ * 
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+@Tag(Tag.OBJECT)
+public class HtmlObject extends HtmlContainer implements ClickNotifier<Header>,
+        HasOrderedComponents, Focusable<HtmlObject> {
+
+    private static final PropertyDescriptor<String, String> dataDescriptor = PropertyDescriptors
+            .attributeWithDefault("data", "");
+
+    private static final PropertyDescriptor<String, Optional<String>> typeDescriptor = PropertyDescriptors
+            .optionalAttributeWithDefault("type", "");
+
+    /**
+     * Creates a new <code>&lt;object&gt;</code> component.
+     */
+    public HtmlObject() {
+    }
+
+    /**
+     * Creates a new <code>&lt;object&gt;</code> component with given data and
+     * type attribute values.
+     * 
+     * @see #setData(String)
+     * @see #setType(String)
+     *
+     * 
+     * @param data
+     *            a data attribute value
+     * @param type
+     *            a type attribute value
+     */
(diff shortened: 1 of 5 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+     * Removes "data" attribute.
+     * 
+     * @see #setData(String)
+     * @see #setData(AbstractStreamResource)
+     */
+    public void removeData() {
+        getElement().removeAttribute("data");
+    }
```
caalador, line 241: Is remove really needed as the component is nothing without the data.

flow-html-components/src/main/java/com/vaadin/flow/component/html/Param.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Removes "name" attribute.
+     * 
+     * @see #setName(String)
+     */
+    public void removeName() {
+        getElement().removeAttribute("name");
+    }
```
caalador, line 116: What use is there with removing name? shouldn't the action be to remove the whole param from object.

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
```diff
@@ -0,0 +1,243 @@
...
+
+/**
+ * Component representing a <code>&lt;object&gt;</code> element.
+ * 
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+@Tag(Tag.OBJECT)
+public class HtmlObject extends HtmlContainer implements ClickNotifier<Header>,
```
caalador, line 39: `ClickNotifier<Header>`? Shouldn't it be `ClickNotifier<HtmlObject>` also a quick look into object click events it seems to not work that well so I would perhaps not have the Click here.

caalador (changes requested):
IT tests for using StreamResource and Click (if not dropped)

denis-anisimov:
>IT tests for using StreamResource and Click (if not dropped)

IT ?

What would the IT test ? The IT makes sense to check that the resource is working in the browser in expected way but I don't see how I can check e.g. that PDF is shown in the browser.
The only test which can be done is checking the URL value. And I don't see a reason to check it via IT.
It's enough to test that the `Element` gets the  resource value.
I will add this test.

I don't see also a reason to check click since I don't see e.g. any tests for `H1` which is also `ClickNotifier`.

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
denis-anisimov: I'm not sure.....
I'm following the `Anchor` pattern. It also has this method.

caalador: Well an anchor without a href is still a placeholder, but I think that an Object without data is nothing.

flow-html-components/src/main/java/com/vaadin/flow/component/html/Param.java
denis-anisimov: Again: following the `Anchor` pattern for the non optional attribute.

The main idea here I think is: the component/element should be considered as is (localized, without any other context).
So it should be possible to do anything you want with even if it makes the element kind of "invalid"/"useless" (for a while).

BUT.......
Now I've got the difference between `href` in `Anchor` and attributes here.
`href` has `false` which prevents it to be removed if `null` is passed.

So you are right: there is no need in `remove` method since `null`  can be used to remove the attriute.

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
denis-anisimov: That's a mistake. Thanks.
Why not ?
That at least shows that it works http://www.htmlnest.com/html/object/event/onclick (at least somehow).

caalador: As long as it has text inside. If you drop the text from the sample it doesn't take the click. Don't know if there is image data.

But lets keep it.

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
denis-anisimov: done

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
denis-anisimov: `Header` is replaced with `HtmlObject`.

denis-anisimov:
> > IT tests for using StreamResource and Click (if not dropped)
> 
> IT ?
> 
> What would the IT test ? The IT makes sense to check that the resource is working in the browser in expected way but I don't see how I can check e.g. that PDF is shown in the browser. The only test which can be done is checking the URL value. And I don't see a reason to check it via IT. It's enough to test that the `Element` gets the resource value. I will add this test.
> 
> I don't see also a reason to check click since I don't see e.g. any tests for `H1` which is also `ClickNotifier`.

unit test is added for the resource.

caalador (approved):

caalador (approved):

PR #11951, opened 2021-09-30
refactor: Use settings object for TaskRunNpmInstall

caalador:
Use a settings object with TaskRunNpmInstall
so it is easier to see and controll what settings
are used.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
@@ -34,7 +34,6 @@ import org.apache.commons.io.FileUtils;
 import com.vaadin.flow.server.Constants;
 import com.vaadin.flow.server.ExecutionFailedException;
 import com.vaadin.flow.server.frontend.installer.NodeInstaller;
-import com.vaadin.flow.server.frontend.scanner.ClassFinder;
 import com.vaadin.flow.shared.util.SharedUtil;
 
 import elemental.json.Json;
@@ -90,7 +89,6 @@ public class TaskRunNpmInstall implements FallibleCommand {
     private final boolean enablePnpm;
     private final boolean requireHomeNodeExec;
     private final boolean autoUpdate;
-    private final ClassFinder classFinder;
 
     private final String nodeVersion;
     private final URI nodeDownloadRoot;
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -553,4 +527,165 @@ private void validateInstalledNpm(FrontendTools tools)
                     String.format(NPM_VALIDATION_FAIL_MESSAGE));
         }
     }
+
+    /**
+     * Settings to use when executing npm install task.
+     */
+    public static class NpmSettings {
```
denis-anisimov, line 534: The class contains public setters which allows to modify the object and confuses the contract.
In fact the setters are not needed: they are used only when the object is constructed even though it's 
better to avoid complicated CTOR with a bunch of args and use methods to set the values.
Once the object is constructed if becomes immutable.

`Builder` pattern is used normally for avoiding complicated CTORs. 
So since this should improve the contract of settings I suggest to avoid confusing contract with mutability.
Two options can be used:
- `Builder` pattern which produces an immutable object in the result
- `Builder` requires quite much code so setters still can be used but then they should be `private`. If settings class is moved as nested class inside NodeTasks then it's possible to access it's private methods and fields.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -99,41 +97,17 @@
...
-            URI nodeDownloadRoot, boolean useGlobalPnpm, boolean autoUpdate) {
-        this.classFinder = classFinder;
-        this.packageUpdater = packageUpdater;
-        this.enablePnpm = enablePnpm;
-        this.requireHomeNodeExec = requireHomeNodeExec;
-        this.nodeVersion = Objects.requireNonNull(nodeVersion);
-        this.nodeDownloadRoot = Objects.requireNonNull(nodeDownloadRoot);
-        this.useGlobalPnpm = useGlobalPnpm;
-        this.autoUpdate = autoUpdate;
+    TaskRunNpmInstall(NpmSettings settings) {
```
denis-anisimov, line 103: I don't see any point to use separate fields since they are already available in the settings class.
Just assign the settings class as a whole to a field and that's it. Any value can be received from there directly.
Just make the class immutable.

denis-anisimov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
caalador: The contract is that you can change the setter and it the setttings are locked in when creating the Task.

denis-anisimov: The builder which produces an immutable instance allows to do this more properly.
The implementation will be more verbose for the builder but then the fields inside the TaskRunNpmInstall
 are not required : the settings can be used as is since it's immutable.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
mshabarov: Let's deprecate the old CTOR and not remove it.
Never know who uses this API, although it's supposed to be internal API.
FrontendTools is an another example which is used in the start project.

mshabarov (changes requested):

PR #11953, opened 2021-09-30
Revert atmosphere changes

haijian-vaadin:
Revert atmosphere changes(https://github.com/vaadin/flow/commit/e6459b39d5deb1036923a54415c8fea77d1217ed and https://github.com/vaadin/flow/commit/bb04952947dc7b53bef9c21eb468dfb3bad68ab5) since it fails MPR and OSGi


denis-anisimov:
I think the reasons of failures should be investigated first. May be it's possible to fix MPR and OSGi instead of reverting this.

haijian-vaadin:
> I think the reasons of failures should be investigated first. May be it's possible to fix MPR and OSGi instead of reverting this.

Yes, it's only that we plan to make a Flow 9 and Spring release today for the stateless feature, so that we can get ready for the DX testing. After the release, the investigation and fixes should continue.

mshabarov (approved):
The investigation on why MPR and OSGi are failing is ongoing, but if you need the release now, go ahead and revert.

PR #11954, opened 2021-09-30
chore: Rename DevModeHandlerImpl to WebpackHandler to better describe what it does

Artur-:


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerManagerImpl.java
```diff
@@ -50,7 +50,7 @@ public void initDevModeHandler(Set<Class<?>> classes, VaadinContext context)
 
     @Override
     public DevModeHandler getDevModeHandler() {
```
denis-anisimov, line 52: `DevModeHandler` should be renamed as well and this is even more important since `DevModeHandlerImpl` is just an impl class and it doesn't even matter how it's called.
But then `DevModeHandler` may not be `WebpackHandler` since this will be reserved for a class name.

And I don't really see a need of this renaming now since the only important thing is the name of SPI/API  and the implementation class may have any name as I said.

denis-anisimov (changes requested):

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerManagerImpl.java
Artur-: The need is to be able to add a `ViteHandler` in parallel

Artur-: Why should `DevModeHandler` be renamed? It is an abstraction

denis-anisimov: Aaah.
That was exactly the question since I wasn't aware of the need.
At the moment there is only one dev mode handler implementation class and I don't recall exactly but it might be that the code which uses `DevModeHandler` assumes it's the only one available (`Lookup::lookup` method is called which ensures that there is only one service) (I'm not sure).

And it wasn't clear for me that you are going to add one more handler. If so then please make sure that `lookupAll` is used.

It's possible to extend `DevModeHandlerManagerImpl` with additional functionality (not only webpack) and that was the reason to the question.

So: please make sure that `lookupAll` is used for `DevModeHandler` SPI in the place where it's used.
And then this rename makes sense.

Artur-: DevModeHandler is never looked up through `Lookup`. All locations look up `DevModeHandlerManager`.

denis-anisimov: It may be considered as a typo.
Indeed the SPI is `DevModeHandlerManager`, the `DevModeHandler` is returned by the `DevModeHandlerManager`.
The problem is : `DevModeHandlerManager` is looked via `Lookup:lookup` and NOT `Lookup:lookupAll` .
Which means you _may not_  have several `DevModeHandlerManager` impls.
And that was something which I vague recalled and wondered why do you need to rename `DevModeHandlerManagerImpl` if it may be the only one impl for the `DevModeHandlerManager`. 
And since it may be the only one it doesn't really matter how it's called.

Artur-: And this is actually great because I can select which DevModeHandler to use in DMHMImpl. Exactly what I want to

denis-anisimov: So you mean the purpose is to use either `WebpackHandler` XOR  some another handler ?
Which means that you may not have dev mode module in your classpath together with abother module which has a different impl of `DevModeHandlerManager`?

If so then it makes sense. But that was far away from obvious from the very begining.

Artur-: Like this https://github.com/vaadin/flow/commit/924129550d12eae5660f0c38f35f1b66f17620de#diff-2fc39945b525bd9bdaebdb60e3e997e84e743d4c61b16f0685af285334a536afR54

denis-anisimov: Ah, so the `DevModeHandlerManager` is only one impl of SPI still but it may return different impls of `DevModeHandler`.
OK.
I would may be implement a different `DevModeHandlerManager` instead but if it shares a lot of functionality with the existing one and should be located in the same dev mode module then this is OK.

PR #11968, opened 2021-10-01
feat: Add clean-frontend goal to maven plugin

### LINKED ISSUE #8885: Maven plugin: add support for cleaning npm-related folders

caalador:
Add a goal to clean frontend files handled
by the framework. Most files are removed,
but package.json is only cleaned of framework
managed changes.

Fixes #8885


caalador:
Still missing tests for goal.

PR #11990, opened 2021-10-06
Util method for launching a browser with a given URL

Artur-:


Artur-:
Yes, that's how it works

PR #11992, opened 2021-10-07
feat: introduce event.target element to DOM event

### LINKED ISSUE #3356: "Converting circular structure to JSON" when using event.target
When trying to implement something like layout click listener in FW8 I ended up with

```
@DomEvent("click")
public class ClickEventWithTargetElement<T extends Component>
        extends ComponentEvent<T> {
    private Component target;

    public ClickEventWithTargetElement(T source, boolean fromClient,
            @EventData("event.target") Component target) {
        super(source, fromClient);
        this.target = target;
    }

    public Component getTarget() {
        return target;
    }
}
```

Adding this to a component and clicking on a child component results in (on client side)
```
(TypeError) : Converting circular structure to JSON
```
and the app stops working. Looks like the whole client engine stops processing events.

Clicking on the component root element results in (on server side)
```
Caused by: java.lang.IllegalArgumentException: Unknown type com.vaadin.flow.component.Component
	at com.vaadin.flow.internal.JsonCodec.decodeAs(JsonCodec.java:228)
```

pleku:
Makes it possible to obtain the Element instance on the server side
that corresponds to `event.target` for a DOM event. The mapping needs
to be explicitly enabled with DomListenerRegistration.mapEventTargetElement()
for it to work.

Next steps would be to introduce an API to certain component events that
leverage the API, or generalizing the API so that one could give out an
JS string that would be mapped instead of `event.target`.

Closes #3356

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
+++ b/flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
@@ -43,6 +43,7 @@ import com.vaadin.client.flow.collection.JsWeakMap;
 import com.vaadin.client.flow.dom.DomApi;
 import com.vaadin.client.flow.dom.DomElement;
 import com.vaadin.client.flow.dom.DomElement.DomTokenList;
+import com.vaadin.client.flow.dom.DomNode;
 import com.vaadin.client.flow.model.UpdatableModelProperties;
 import com.vaadin.client.flow.nodefeature.ListSpliceEvent;
 import com.vaadin.client.flow.nodefeature.MapProperty;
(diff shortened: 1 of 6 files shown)


caalador:
one test is failing:
```
DomApiAbstractionUsageTest.testDomApiCodeNotUsedcom.vaadin.client

java.lang.AssertionError: com.vaadin.client.flow.binding.SimpleElementBindingStrategy.getClosestStateNodeToTarget calls elemental.dom.Node.isEqualNode
  at com.vaadin.client.DomApiAbstractionUsageTest.verifyMethod(DomApiAbstractionUsageTest.java:148)
  at com.vaadin.client.DomApiAbstractionUsageTest.access$300(DomApiAbstractionUsageTest.java:36)
  at com.vaadin.client.DomApiAbstractionUsageTest$1$1.visitMethodInsn(DomApiAbstractionUsageTest.java:98)
```

pleku:
The bot is mistaken as the added interface method has a default implementation and is not thus a breaking change.

PR #12009, opened 2021-10-11
fix: use correct setup for tsconfig-paths-webpack-plugin

Lodin:
Solves issues that caused previous [PR revesion](https://github.com/vaadin/flow/pull/12007).


platosha:
Closing, as agreed in the daily: this is not a priority and getting tedious to fix.

PR #12024, opened 2021-10-13
test: Simplify tests

Artur-:
There is no need to test multiple components, only to test that a component styles file is applied


Artur-:
Isn't this something that the web components project tests?

Issue #11957, opened 2021-09-30
Generate index.html into the frontend folder


Artur-:
### Description of the bug / feature
Currently an `index.html` file is generated into `target` if non is present in `frontend`. There are very few benefits with this, it mostly makes it harder to understand how to customize the bootstrap page when you e.g. want to add some meta tags.

### Expected behavior
It should be generated into `frontend` instead.

knoobie:
What do you think about `frontend/generated` instead like already used for `theme.js` and other files. This could simplify the logic for the user as well. You want your own `index.html`? Create one at `frontend/index.html` otherwise the `frontend/generated/index.html` is used by the framework.

Artur-:
I am not sure hiding it in another place actually solves anything

Artur-:
If you do not want to see or customize it, you can always add it to .gitignore or similar

PR #12082, opened 2021-10-19
feat: Basic Vite support behind a feature flag

Artur-:
Depends on #11957  and #12072


caalador:
> > Just as a question, where do we reference to the themeFolder with theme?
> > Is it for template files?
> 
> The generated theme at least contains `import stylesCss from 'themes/my-theme/styles.css';`

But that should be resolved from ./frontend/ which should be a (the?) root entrypoint. Ofc if it needs defining with Vite then it should be defined.

Artur-:
Hopefully fixed now

PR #12089, opened 2021-10-19
refactor: Move dev server startup logic to AbstractDevServer

Artur-:


Artur-:
It doesn't complete when the first match is found, rather it keeps looking for future matches also and invokes the callback. The startup process only wants to know when the initial compilation is done so keeping the two concepts separate seems more sensible to me in this case.

PR #12103, opened 2021-10-20
feat: Do not use a loading screen for Vite

Artur-:


Artur-:
This would be a good solution for the case when you do not need to run `npm install`... Now that still blocks loading for a long long time

PR #12105, opened 2021-10-20
fix: Load dev mode gizmo for Vite when not in production

Artur-:


Artur-:
Added to `generated-flow-imports.js` also

PR #12206, opened 2021-11-01
test: Fix most 'scanned from multiple locations' warnings

Artur-:
containerIncludeJarPattern is "A pattern to match against the urls of the jars on the container path to select those that should be examined for things like META-INF fragments, resources, tlds and class inheritance hierarchy" but the container path is not the class path but also includes all Maven dependencies like

 org.apache.maven:maven-core:jar:3.8.2:compile
    com.google.inject:guice:jar:no_aop:4.2.2:compile
       com.google.guava:guava:jar:25.1-android:compile


caalador (approved):

Artur-:
Now it actually starts the server also...

caalador:
Has a merge conflict now.

Artur-:
Fixed

caalador (approved):

Issue #11, opened 2015-10-09
script tags cause TemplateParser to fail


OlliTietavainenVaadin:
If there's a non-empty < script> tag in a template, TemplateParser will fail. The createTemplate method throws a RuntimeException; node type for scripts is org.jsoup.nodes.DataNode.

PR #12207, opened 2021-11-01
Revert "feat: use array of values for @Route instead of only one (#11…

pleku:
…938)"

This reverts commit d202f290a99ac1517a7224f4b6b204bc4138079f (#11915)
Reverted because the change had way too little gain and was
breaking binary compatibility with unknown amount of projects and add-ons.


mshabarov:
This needs the revert in flow components https://github.com/vaadin/flow-components/pull/2297

mshabarov (dismissed):

mshabarov (approved):

PR #12228, opened 2021-11-01
fix: ensure Flow client container is pre-attached on navigation

### LINKED ISSUE #12080: Component's ShadowRoot is null on attach when navigating from a Typescript view.
Consider this view
```java
@Route(value = "shadow")
public class ShadowView extends VerticalLayout {

    public ShadowView() {
        TextField field = new TextField();
        field.addAttachListener(ui -> field.getElement().executeJs("alert(this.shadowRoot)"));
        add(field);
    }
}
```
If you navigate to this view from a typescript view, the `shadowRoot` will be null. Otherwise it will be available.

A more complex example:
1. Download the project from https://start.vaadin.com/?preset=all&preset=partial-prerelease
2. Go to one of typescript views
3. Go to the collaborative-master-detail-view
4. See that an error is shown.

ce-issue

The errors are caused by the field-highlighter trying to access the components shadow roots.

[my-app.zip](https://github.com/vaadin/flow/files/7367000/my-app.zip)
1. Run the application
2. Click in "Shadow". 
3. Verify that the alert shows `[object ShadowRoot]`
4. Click in "Hello TS"
5. Click in "Shadow". 
6. Verify that the alert shows `null`
### Expected behavior
The alert should show `[object ShadowRoot]` in all cases.

### Versions:
    - Vaadin / Flow version: 22.0.0.alpha8/9.0.0.alpha10
    - Java version: 8, 17
    - OS version: Linux, MacOs

platosha:
Fixes #12080

The client-side Vaadin Router does not attach components until the `onBeforeEnter` callback is run. However, when the Flow client initializes a server-side view, it assumes its container and the UI root are connected to the document. This concerns constructor timing as well as `AttachEvent` listeners.

This fixes the issue by pre-attaching the hidden container element created in the `Flow.ts` (the adapter for the client-side router) to the body.

### DIFF
--- a/flow-client/src/main/frontend/Flow.ts
+++ b/flow-client/src/main/frontend/Flow.ts
@@ -275,16 +275,18 @@ export class Flow {
         const tag = `flow-container-${appId.toLowerCase()}`;
         this.container = flowRoot.$[appId] = document.createElement(tag);
         this.container.id = appId;
-
-        // It might be that components created from server expect that their content has been rendered.
-        // Appending eagerly the container we avoid these kind of errors.
-        // Note that the client router will move this container to the outlet if the navigation succeed
-        this.container.style.display = 'none';
-        document.body.appendChild(this.container);
       }
 
       // hide flow progress indicator
       this.loadingFinished();
+    }
+
+    // It might be that components created from server expect that their content has been rendered.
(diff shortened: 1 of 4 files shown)


Haprog (approved):
LGTM. Just waiting to see if CI passes.

PR #12233, opened 2021-11-02
chore: extract URI encoding to a helper

Artur-:


flow-server/src/test/java/com/vaadin/flow/internal/UrlUtilTest.java
```diff
@@ -37,4 +39,21 @@ public void isExternal_URLContainsAnySchemaAsPrefix_returnsTrue() {
     public void isExternal_URLDoesnotContainSchema_returnsFalse() {
         Assert.assertFalse(UrlUtil.isExternal("foo"));
     }
+
+    @Test
+    public void encodeURI_shouldNotBeEscaped() {
```
caalador, line 44: This test has the problem that `+` is always changed to `%20` so I guess that should be handled so that any + is replaced before encoding and put back after changing the space + to %20

Artur-: Isn't that completely valid? "+" is not a special URI character as far as I know

Artur-: i.e. "+" in the string becomes "%2B"

Artur-: I will add an explicit test

Artur-: Added

caalador (approved):

PR #12271, opened 2021-11-04
fix: Changed copied server info to format "header: value" to be easier to read

Artur-:


jouni (approved):
Looks okay, even though `Live reload: Java active (Spring Boot Devtools): Front end active` might look a little weird.

PR #12272, opened 2021-11-04
chore: Move dev server connection checking to super class

Artur-:


vaadin-dev-server/src/test/java/com/vaadin/base/devserver/WebpackHandlerTest.java
```diff
@@ -276,6 +277,10 @@ public void should_HandleAnyAssetInManifestPaths() throws Exception {
         handler = startWebpack(port);
         waitForDevServer();
 
+        Method m = WebpackHandler.class.getDeclaredMethod("readManifestPaths");
+        m.setAccessible(true);
+        m.invoke(handler);
+
```
caalador, line 283: Here with the test comes a question. should the `isDevModeRequest` actually do the `readManifestPaths` that would call webpack as couldn't the files/contents change during runtime if things are added/removed else the read could just cache the things on first call to the method.

Artur-: This will be removed in a later PR along with everything manifest path related

caalador (approved):

PR #12278, opened 2021-11-04
refactor: Remove usage of manifest.json and method for checking if dev server can serve a request

Artur-:
The method is plain wrong in some cases, like when asking if it can serve the client engine files. It says yes but does not serve them. Instead, you should always ask the dev server to serve a file and check if it actually did serve it.

In development mode, we first request the dev server to serve a resource. If it doesn't serve it, then we continue as before.

In production, we first check META-INF/VAADIN/webapp and serve any match as a resource. If there isn't one, then we continue as before.

For this to work, index.html needs to be a special case as it exists in the dev server but should not be served from there.


joheriks:
So can the `WebpackManifestPlugin` also be removed from the Webpack configuration file now?

Artur-:
> So can the `WebpackManifestPlugin` also be removed from the Webpack configuration file now?

That should be possible, removed

Artur-:
Resolved conflicts

joheriks (approved):

PR #12279, opened 2021-11-05
fix: fix client requests with windows.

caalador:
Windows vite requests will
contain the path with the drive
which contains `:` so it cannot be
encoded to `%3A` as then the
devServer returns a 404

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
@@ -16,15 +16,11 @@
 
 package com.vaadin.flow.router;
 
-import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-
-import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Utility class exposing reusable utility methods for location.
@@ -53,7 +49,7 @@ public class LocationUtil {
             // Ignore forbidden chars supported in route definitions
             String strippedPath = path.replaceAll("[{}*]", "");
 
-            URI uri = new URI(URLEncoder.encode(strippedPath, UTF_8.name()));
+            URI uri = new URI(null, null, strippedPath, null);
             if (uri.isAbsolute()) {
                 // "A URI is absolute if, and only if, it has a scheme
                 // component"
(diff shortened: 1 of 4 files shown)


Artur- (approved):

Artur-:
We really really should not be using `URLEncoder.encode` here. It should be something equivalent to URIEncode in JavaScript: 
"
encodeURI() escapes all characters except:

Not Escaped:

    A-Z a-z 0-9 ; , / ? : @ & = + $ - _ . ! ~ * ' ( ) #
"

Artur- (dismissed):

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/AbstractDevServerRunner.java
```diff
@@ -695,7 +695,10 @@ public boolean serveDevModeRequest(HttpServletRequest request,
             requestFilename = "/VAADIN/static" + requestFilename;
         }
 
-        String devServerRequestPath = UrlUtil.encodeURI(requestFilename);
+        // Dev server request can have an absolute system path making
+        // it required to not encode ':'
```
Artur-, line 699: This is the wrong place to change. URLs are URLs, there is nothing special about the dev server?

Artur- (changes requested):

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/AbstractDevServerRunner.java
caalador: Well then the "fixed" url can not be used with Location as "abc:foo/bar" will read as has scheme so it's relative which is not accepted for a location. Wasn't this change made just to support spaces?

flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
```diff
@@ -51,7 +53,7 @@ public static void verifyRelativePath(String path) {
             // Ignore forbidden chars supported in route definitions
             String strippedPath = path.replaceAll("[{}*]", "");
 
-            URI uri = new URI(UrlUtil.encodeURI(strippedPath));
+            URI uri = new URI(URLEncoder.encode(strippedPath, UTF_8.name()));
```
Artur-, line 56: This is also wrong. What are you trying to fix here?

Artur-: Looks to me like this is using the wrong URI constructor. The parameter passed is not a full URL but a path and an appropriate constructor should be used

Artur-: Maybe it should use `    public URI(String scheme, String host, String path, String fragment)` with everything except the `path` parameter set to `null`

Artur-:
An alternative fix that tries to address all potential problems in #12291

PR #12291, opened 2021-11-05
fix: Make URLUtil.encodeURI behave like encodeURI in JavaScript and add encodeURIComponent

Artur-:


flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
```diff
@@ -51,7 +51,9 @@ public static void verifyRelativePath(String path) {
             // Ignore forbidden chars supported in route definitions
             String strippedPath = path.replaceAll("[{}*]", "");
 
-            URI uri = new URI(UrlUtil.encodeURI(strippedPath));
+            // : is completely valid in a path but URI will think that defines a
+            // protocol so skip it for the check
+            URI uri = new URI(UrlUtil.encodeURI(strippedPath).replace(":", ""));
             if (uri.isAbsolute()) {
```
caalador, line 57: Is the initial check un-necessary as we can never have a uri with a scheme at this point?

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/router/LocationUtil.java
Artur-: It would seem so if we first remove : and then check for :

Let's fix in a separate PR

PR #12292, opened 2021-11-05
chore: Maven and Gradle plugins should not be OSGi bundles

Artur-:


Artur-:
How do you run a Maven plugin through OSGi? 

The newer bnd plugin is not run at all for the plugins but only outputs
```
[13:05:56] :	 [Step 7/11] [INFO] --- bnd-maven-plugin:6.0.0:bnd-process (default) @ flow-maven-plugin ---
[13:05:56] :	 [Step 7/11] [INFO] skip project with packaging=maven-plugin
```

mstahv:
The actual Maven and Gradle plugins probably don't need to be bundles, but the shared library, flow-plugin-base, needs to. @stbischof did I got it right?

stbischof:
yes. It would help to have the shared lib ad an bundle.

Artur-:
Updated so `flow-plugin-base` is still a bundle

Artur-:
Looks good now?

stbischof:
LGTM

mstahv (approved):

Issue #4740, opened 2018-10-29
Determine how to enable Brotli compression by default


Legioth:
This is a follow-up to #4583 which introduces opt-in support for Brotli compression. Brotli compression should be enabled by default in cases where we can assume there won't be any interference issues similar to those that caused https://github.com/vaadin/flow/pull/4556 to be reverted.

In order to do that, these steps are needed:
* Introduce a third configuration parameter option so that `true` always uses precompressed files, `false` never uses, and a new `auto` configuration uses unless we can detect a situation where there will be interference. `auto` is used by default if no specific value is defined for the application.
* When `auto` mode is in use, look for Tomcat 8.x with built-in precompression e.g. based on the findings documented in https://github.com/vaadin/flow/issues/4583#issuecomment-418683599.
* Test with all other supported servlet containers. In particular, look for any built-in configuration options for enabling compression similar to the incompatible option in Tomcat 8.x and test with such options enabled.

Artur-:
How about we just do not support Brotli for Tomcat 8 and use Brotli for other servers. Unless the latest Tomcat 8 is fixed, then we just enable it. Then if you are using a broken server, you can disable it

Artur-:
Spring Boot nowadays uses Tomcat 9.x by default and there I do not see any issues when using Vite with Brotli compressed resources and
```
server.compression.enabled=true
server.compression.mime-types=application/json,application/xml,text/html,text/xml,text/plain,application/javascript,text/css
```

It is not possible to downgrade to using Spring Boot 2.5  + Tomcat 8.5.72 either so it cannot be tested
```
The following method did not exist:
    'void org.apache.tomcat.util.modeler.Registry.disableRegistry()'
```

Artur-:
Also tested Jetty and Undertow - no issues

Issue #4583, opened 2018-09-05
Determine how to re-enable Brotli compression


Legioth:
#4556 was reverted because it interfered with the built-in on-the-fly compression in Tomcat 8.x which still applies gzip compression to the response, thus making it double compressed even though the response header implies that it's only gzipped.

Compression is not enabled by default in Tomcat, but for instance Bakery has `server.compression.enabled=true` in `application.properties` which enables the feature. The same setting works fine with the embedded Jetty and Undertow servers since their built-in compression features properly handle the case.

We could make serving of precompressed Brotli files configurable, but then we'd also have to decide which way should be the default, or alternatively make the default setting vary depending on whether Tomcat 8.x is detected from the classpath.

knoobie:
Is that related to Tomcat 8.x or Tomcat 9.x as well?

Legioth:
Detecting Tomcat 8.x is easy:`servletContext.getServerInfo().startsWith("Apache Tomcat 8.");`.

Figuring out the compression settings is a little bit tricky. It seems like one working approach would be to query for MBeans with the name `Tomcat:type=ProtocolHandler,port=*` and then check the `compression` attribute value for each of them to look for the value `"on"` or `"force"`.

Legioth:
@knoobie Based on reading the code, Tomcat 9.x is fine, but I haven't yet tested that in practice.

Legioth:
I've crated a new PR that requires explicit opt-in to use precompressed Brotli files in https://github.com/vaadin/flow/pull/4739. I've created https://github.com/vaadin/flow/issues/4740 for making it used by default.

PR #12308, opened 2021-11-08
feat: Compress Vite output by default in production

### LINKED ISSUE #12055: [Vite] Bundle compression for production bundles
### How does it work today
The compression plugin is used in production builds to get a gzip bundle that can be served if supported by the requesting browser.

### How could it work
We need to look into what Rollup or Vite plugins there are for the same thing. Requires investigation.

Artur-:
As Brotli is disabled by default, you need to enable it to use this feature.
https://github.com/vaadin/flow/issues/4740 will enable Brotli by default once server issues are sorted out

Fixes #12055

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -374,6 +374,7 @@ public abstract class NodeUpdater implements FallibleCommand {
 
         if (featureFlags.isEnabled(FeatureFlags.VITE)) {
             defaults.put("vite", "v2.7.0-beta.3");
+            defaults.put("rollup-plugin-brotli", "3.1.0");
             defaults.put("mkdirp", "1.0.4"); // for application-theme-plugin
         } else {
             // Webpack plugins and helpers
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -11,6 +11,8 @@ import { processThemeResources } from '@vaadin/application-theme-plugin/theme-ha
 import settings from '#settingsImport#';
 import { UserConfigFn, defineConfig, HtmlTagDescriptor, mergeConfig } from 'vite';
 
+import brotli from 'rollup-plugin-brotli';
+
 const frontendFolder = path.resolve(__dirname, settings.frontendFolder);
 const themeFolder = path.resolve(frontendFolder, settings.themeFolder);
 const buildFolder = path.resolve(__dirname, settings.frontendBundleOutput);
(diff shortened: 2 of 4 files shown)


caalador:
Something is failing on the application start now: 
applicationStarts ANY_Chrome_ (com vaadin viteapp ProductionBasicsIT)

It's clearly loading, but I don't think that is the expected result for the page.

Artur-:
Updated to use `rollup-plugin-brotli` which is compatible with Vite. `rollup-plugin-gzip` compresses the output files in `generateBundle` and Vite still modifies the files after that plugin phase. `rollup-plugin-brotli` gathers the file names in `generateBundle` and compresses them in `writeBundle` which seemingly work correctly with Vite

flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java
```diff
@@ -227,7 +227,7 @@ default boolean getBooleanProperty(String propertyName,
      */
     default boolean isBrotli() {
         return getBooleanProperty(InitParameters.SERVLET_PARAMETER_BROTLI,
-                false);
+                true);
```
caalador, line 230: Does this change the webpack runner default now also?
Does this have an impact on something as it was defined as false before?
Should we bind the parameter with Vite so nothing changes for webpack?

Artur-: It enables Brotli support by default, which means that if StaticFileServer finds a ".br" version of a file, it serves that with Brotli headers. 

I don't think we should make this Vite specific

caalador: So the reason for not automatically enabling brotly was:

```
 Serving precompressed Brotli files is not enabled by default because some servlet containers are applying gzip encoding to the response even if there's already a header saying that the response is compressed using Brotli.
 ```

Artur-: Any idea which servers those were and how we can see if they were fixed or then disable compression only for them?

caalador: @Legioth do you remember what, where, who and how?

Artur-: Apparently some Tomcat version had this issue

Artur-: https://github.com/vaadin/flow/issues/4583

How about we disable it for Tomcat 8 like mentioned in that issue?

Artur-: Let's do that in a separate PR though

Artur-: Reverted this file

PR #12318, opened 2021-11-09
refactor: minimize stats.json for V14 bootstrap mode

joheriks:
Write only `assetsByChunkName` into the generated `stats.json`,
as template sources are no longer needed there.

Fixes #12318.


flow-server/src/main/resources/webpack.generated.js
```diff
@@ -334,13 +333,20 @@ module.exports = {
...
+      // V14 bootstrapping needs the bundle names
+      compiler.hooks.afterEmit.tapAsync("FlowStatsHelper", (compilation, done) => {
+        let miniStats = {
+          assetsByChunkName: compilation.getStats().toJson().assetsByChunkName
+        };
+        if (!devMode) {
+          fs.writeFile(statsFile, JSON.stringify(miniStats, null, 1), done);
+        } else {
+          stats = miniStats;
+          done();
```
Artur-, line 346: Shouldn't done(). be called also in the other branch?

joheriks: It's passed as the callback for when `writeFile` finished. But maybe it's better to add a lambda in between to prevent an unwanted argument from being passed.

Artur- (approved):

PR #12327, opened 2021-11-10
feat: Enable serving Brotli resources by default

### LINKED ISSUE #4740: Determine how to enable Brotli compression by default

Artur-:
Fixes #4740

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java
+++ b/flow-server/src/main/java/com/vaadin/flow/function/DeploymentConfiguration.java
@@ -227,7 +227,7 @@ public interface DeploymentConfiguration
      */
     default boolean isBrotli() {
         return getBooleanProperty(InitParameters.SERVLET_PARAMETER_BROTLI,
-                false);
+                true);
     }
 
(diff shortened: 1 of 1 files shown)


Artur-:
As far as I see in the code, this should not affect anything else than Vite because nothing else is generating `.br` files

joheriks (approved):

PR #12334, opened 2021-11-11
feat: add support for @EventData of type Component

pleku:
Makes it possible to use event data expressions that map to any Component / Element from server side state tree.
For example it will be possible to use:
- `@EventData("event.target") Component eventTarget` for `@DomEvent`
- `@EventData("event.target") Element eventTarget` for `@DomEvent`
- `DomEventRegistration::addEventDataElement(String expression)` to get any element mapped,
- and fetch it with `DomEvent::getEventDataElement(String expression)`

Follow up for #11992, #3356

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
+++ b/flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
@@ -1265,8 +1265,8 @@ public class SimpleElementBindingStrategy implements BindingStrategy<Element> {
             } else if (expressionString
                     .equals(JsonConstants.MAP_STATE_NODE_EVENT_DATA)) {
                 // map event.target to the closest state node
-                int targetNodeId = getClosestStateNodeIdToTarget(node,
-                        event.getTarget(), "event.target");
+                int targetNodeId = getClosestStateNodeIdToEventTarget(node,
+                        event.getTarget());
                 eventData.put(JsonConstants.MAP_STATE_NODE_EVENT_DATA,
                         targetNodeId);
             } else if (expressionString
@@ -1479,8 +1479,8 @@ public class SimpleElementBindingStrategy implements BindingStrategy<Element> {
     }
 
     // This method could be moved somewhere to be reusable
-    private int getClosestStateNodeIdToTarget(StateNode topNode,
-            EventTarget target, String eventDataExpression) {
+    private int getClosestStateNodeIdToEventTarget(StateNode topNode,
+                                                   EventTarget target) {
         if (target == null) {
             return -1;
         }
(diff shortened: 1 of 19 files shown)


flow-server/src/test/java/com/vaadin/flow/component/MappedDomEventWithRouterLinkData.java
```diff
@@ -0,0 +1,36 @@
...
+import com.vaadin.flow.router.RouterLink;
+
+public class MappedDomEventWithRouterLinkData
+        extends MappedDomEventWithComponentData {
+
+    private final RouterLink routerLink;
+
+    public MappedDomEventWithRouterLinkData(Component source,
+            boolean fromClient, @EventData("component") Component component,
+            @EventData("router.link") RouterLink routerLink) {
```
knoobie, line 28: This looks like "user generated" components should be able to be returned, like e.g. "MyCustomComponent extends Component" - is this intended? If yes, would be awesome to have a test for this; because this could be really awesome for type-safe component usage (probably problematic with generics..)

pleku: Do mean like https://github.com/vaadin/flow/pull/12334/files#diff-c6d2fb2b77355f6aa7c20d042332d2563c7e9313aff1a02809a5ee3d7857e342R94 where the `H3` component is mapped if it is there, or an error thrown when the type doesn't match. The implementation is in: https://github.com/vaadin/flow/pull/12334/files#diff-488cb30a1687bdf305e93b812187b637e6a3b01fce4f0c900ec73e49a7c79721R348 and it uses existing `ComponentMapping` server side feature, which says:
```java
"     * Gets the component this node has been mapped to, if any.
     *
     * @return an optional component, or an empty optional if no component has
     *         been mapped to this node
```
So in case you're creating things via the element API, but not as a component mapping, it will return `null` when you try to map it to a component. We do have code which will "map the component to an existing element", do you mean that I should also take this into account and make it possible ?

pleku: Maybe I misinterpreted your note/question 😅 

Yes, any component instance type is supported and there is an integration test at least with `H3` which should not be any different from any other component (in the eyes of the framework code).

But if you want to be able to map to things that are created with `new Element("foobar")` to `@EventData("event.target.getFoobarElement()") Foobar foobar` where `Foobar` is `@Tag("foobar") Foobar extends Component`, it will not work currently as there is _no existing mapping_. But the element instance has a component mapping as it has been created with `new Foobar()`, then it works.

knoobie: That make sense, as long as there is currently the target "in memory" / already instantiated it can be used. Sounds good! Totally forgot that e.g. RouterLink is already a "user generated" component for flow, but you are right - that should work :)

joheriks:
Two two Sonar critical issues about refactoring `getClosestStateNodeIdToTarget` to have less nesting and the unused import in `ComponentEventBus` could be addressed.

flow-tests/test-root-context/pom.xml
```diff
@@ -80,6 +80,7 @@
             <plugin>
                 <artifactId>maven-clean-plugin</artifactId>
                 <!-- This is because we run multiple modes on the same module -->
+                    <!--
```
caalador, line 83: I would just remove the extra execution completely if we do not need it.

pleku: "Oops", but I'll do it in a separate PR

flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
```diff
@@ -1495,13 +1512,47 @@ private int getClosestStateNodeToTarget(StateNode topNode,
...
+                    .wrap(WidgetUtil.crazyJsCast(domNodeReference));
+            StateNode stateNodeForDomNode = null;
+            do {
+                stateNodeForDomNode = stateTree
+                        .getStateNodeForDomNode(targetNode);
+                targetNode = DomApi.wrap(targetNode.getParentNode());
+            } while (stateNodeForDomNode == null && targetNode != null);
+            if (stateNodeForDomNode != null) {
+                return stateNodeForDomNode.getId();
+            }
```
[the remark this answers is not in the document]
pleku: Done

pleku:
> Two two Sonar critical issues about refactoring getClosestStateNodeIdToTarget to have less nesting and the unused import in ComponentEventBus could be addressed.

Done

joheriks (approved):

pleku:
Cherry-picked to 2.8

PR #12337, opened 2021-11-11
fix: Remove disused plugins from package.json

### LINKED ISSUE #12335: stats-plugin remains in package.json after upgrading from flow 9 beta2 to beta3
The `stats-plugin`  was removed but the code in `NodeUpdater.addWebpackPlugins` only adds the needed plugins and does not delete any old plugins

joheriks:
It does not delete them from `target` as this will go away with the next clean.

Fixes #12335

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
@@ -16,6 +16,7 @@
 package com.vaadin.flow.server.frontend;
 
 import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.net.URL;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
```diff
@@ -231,6 +232,16 @@ public void getJsonFileContent_incorrectPackageJsonContent_throwsExceptionWithFi
                 StringContains.containsString("broken-package.json"));
     }
 
+    @Test public void removedDisusedPlugins() throws IOException {
+        File packageJson = new File(npmFolder, "package.json");
+        FileWriter packageJsonWriter = new FileWriter(packageJson);
+        packageJsonWriter.write(
+                "{\"devDependencies\": {\"@vaadin/some-old-plugin\": \"./target/plugins/some-old-plugin\"}}");
```
Artur-, line 239: Would be good to test the real case also where you have plugin A and B in the file and B should remain

Artur- (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
joheriks: Thanks, the test was in fact a dud as it checked in the wrong JSON object. Now corrected.

Artur- (approved):

Artur- (approved):

PR #12341, opened 2021-11-11
refactor: Move CSRF checking from AccessChecker to FusionController

Artur-:
CSRF checking is related to the HTTP request and not at all related to the access checker. If you use the access checker in another context such as a websocket message, you cannot do the same CSRF check nor do you need any CSRF check.


platosha (approved):

Artur-:
Fixed the javadoc issue

platosha (approved):

PR #12352, opened 2021-11-12
fix: Use correct URL for generated/vaadin.ts

### LINKED ISSUE #12351: [vite] vaadin.ts loaded from wrong URL
When opening an app using 22 beta 2, you can see that the browser loads 

http://localhost:8080/generated/vaadin.ts

when it should be using

http://localhost:8080/VAADIN/generated/vaadin.ts

Artur-:
Fixes #12351


joheriks:
Should `enforce: pre` also be removed if it is ineffectual in its current place, and would actually make things worse if moved to the plugin object where it is effectual.

joheriks (approved):

Issue #12052, opened 2021-10-15
[Vite] PWA support


gilberto-torrezan:
PWA support is defined by 3 aspects:

- Web manifest
- Service worker
- Offline configuration

### How it works today
ServiceWorkerPlugin using InjectManifest for PWA (using workbox)

### How could it work
Service worker should be generated with the same configuration using the same workbox version

Note: offline support is something that might need to work in "webpack mode" for simplicity, if it doesn't work out of the box.

gilberto-torrezan:
The web manifest doesn't concern webpack at all, but the service worker and the offline configuration does.

We are already using workbox to process the output from webpack, so _in theory_ it would be just a matter of processing the output of vite instead for this to work. Needs investigation.

gilberto-torrezan:
The offline support will be tracked as a separate ticket: https://github.com/vaadin/flow/issues/12383

joheriks:
PWA in dev mode is tracked in #12393.

PR #12363, opened 2021-11-15
feat: enable pwa for vite

caalador:
Enable pwa installation for vite.

Part of #12052


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -98,6 +101,26 @@ export const vaadinConfig: UserConfigFn = (env) => {
...
       !devMode && brotli(),
+      settings.pwaEnabled &&
+      {
+        name: 'pwa',
+        enforce: 'post',
+        apply: 'build',
+        transformIndexHtml: {
+          enforce: 'post',
+          transform(html) {
+            return html.replace('</head>', '<link rel="manifest" href="manifest.webmanifest"></head>');
```
Artur-, line 112: Return an object instead, see `inject-entrypoint-script`

flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw-vite.ts
```diff
@@ -0,0 +1,6 @@
+importScripts("https://storage.googleapis.com/workbox-cdn/releases/3.6.3/workbox-sw.js");
```
Artur-, line 1: Does this make all applications load from googleapis.com? We should not do that

flow-server/src/main/resources/vite.generated.ts
caalador: Removed it altogether as noted that if pwa is enabled the link will be added from the server when index html is served.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw-vite.ts
caalador: Still trying to figure out how to not need the cdn, but haven't understood how to get the ts file tranpiled by vite so we could use : `import { precacheAndRoute } from 'workbox-precaching'`

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateSettingsFile.java
```diff
@@ -106,6 +123,42 @@ private String combinePath(String... parts) {
...
+
+    private String getOfflinePath() {
+        if (pwaConfiguration.isOfflinePathEnabled()) {
+            return getEscapedRelativeWebpackPath(
+                    Paths.get(pwaConfiguration.getOfflinePath()));
+        }
+        return ".";
+    }
+
+    private String getEscapedRelativeWebpackPath(Path path) {
```
Artur-, line 154: Webpack?

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -101,6 +108,53 @@ export const vaadinConfig: UserConfigFn = (env) => {
...
+            'vite:esbuild',
+            'replace',
+            'vite:define',
+            'rollup-plugin-dynamic-import-variables',
+            'vite:esbuild-transpile',
+            'vite:terser',
+          ]
+          const plugins = pwaConfig.plugins.filter(p => includedPluginNames.includes(p.name)) as Plugin[]
+          const bundle = await rollup.rollup({
+            input: path.resolve(settings.clientServiceWorkerSourceDir, settings.serviceWorkerSource),
```
Artur-, line 134: Can't we just pass the full path to sw.ts/sw.js in settings and not pass the dir at all?

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -101,6 +108,53 @@ export const vaadinConfig: UserConfigFn = (env) => {
...
+      {
+        name: 'pwa',
+        enforce: 'post',
+        apply: 'build',
+        async configResolved(config) {
+          pwaConfig = config;
+        },
+        async buildStart() {
+          // eslint-disable-next-line @typescript-eslint/no-var-requires
+          const rollup = require('rollup') as typeof Rollup
```
Artur-, line 121: Please add a comment here about what this is doing. I still have a feeling that this will cause problems at some point and that there is a much better way to do it but cannot say exactly what that way is right now. We can go with this for now but a comment will help in the future if we need to figure out another way

flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw-vite.ts
```diff
@@ -0,0 +1,91 @@
+/// <reference lib="webworker" />
```
Artur-, line 1: Do we need a separate copy? The only difference seems to be 
```
+declare var OFFLINE_PATH_ENABLED: boolean; // defined by webpack.generated.js
+declare var OFFLINE_PATH: string; // defined by webpack.generated.js
-const offlinePath = OFFLINE_PATH;
+const appShellPath = '.';
+const offlinePath = OFFLINE_PATH_ENABLED ? OFFLINE_PATH : appShellPath;
```

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateSettingsFile.java
caalador: Renamed

flow-server/src/main/resources/vite.generated.ts
caalador: done

flow-server/src/main/resources/vite.generated.ts
caalador: done

flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw-vite.ts
caalador: unified

Artur- (dismissed):

Artur-:
Some conflicts remain

Artur- (approved):

joheriks (approved):

PR #12364, opened 2021-11-15
fix: handle query params and whitespace in css-urls

### LINKED ISSUE #12287: theme @fontface urls don't support query params
When in a Vaadin theme i have a @fontface url as url('./roboto-v20-latin-ext_latin-regular.woff2') this leads to the browser making a request for http://localhost:8080/ui/VAADIN/static/themes/combinedtheme/fonts/roboto-v20-latin-ext_latin-regular.woff2 which is good and works as expected.
But with a query param as url('./roboto-v20-latin-ext_latin-regular.woff2?v=1.12') the browser makes a request to http://localhost:8080/ui/roboto-v20-latin-ext_latin-regular.woff2?v=1.12 which fails on a 404 as expected.
Why does it fail with a query parameter? Is Vaadin altering these urls during theme packaging during build? I'm on Vaadin 14.7.1
The issue in fact occurs also when there's a space in the url as url('./roboto-v20-latin-ext latin-regular.woff2') and when the url doesn't have a path as url('roboto-v20-latin-ext_latin-regular.woff2'). 2 cases that are supported in plain css.
This hampers us (development) consuming themes build by an other department (UX)...
A construct like this is a theme css : 
 @font-face {
    font-family: 'Roboto';
    font-style: normal;
    font-weight: 400;
    src: local(''),
    url('./roboto-v20-latin-ext_latin-regular.woff2?v=1.12') format('woff2')
}
body {
  font-family: Roboto
}
### Expected behavior
@fontface url values should support the same in theme css files as for plain css.
### Actual behavior
a url query parameter in the url makes it fail to load, as the url seems to be altered to load for the app's root folder, instead of from VAADIN/static
### Versions:
    - Vaadin / Flow version: 14.7.1
    - Java version: 11

pleku:
After this change, it is no longer to problem to have whitespace
and/or query parameters in CSS-URLs like:

url('./fonts/Roboto/roboto-v29-latin-ext_latin-regular.woff2?v=1.12') format('woff2')
url('./fonts/Roboto/roboto v29-latin-ext_latin-regular.woff2?v=1.12') format('woff2')
url('./fonts/Roboto/roboto v29-latin-ext_latin-regular.woff2') format('woff2')

After this change, something still strips the query parameter away but now it is not failing to load the font anymore.

Partly fixes #12287


flow-server/src/main/resources/plugins/theme-loader/theme-loader.js
```diff
@@ -3,7 +3,8 @@ const fs = require('fs');
 const path = require('path');
 
 // Collect groups [url(] [ |'|"]optional './|../', file part and end of url
```
caalador, line 5: The comment should be updated as end of url is preceded with query.
Also on line 35 it should be noted that the groups have gone from 5 to 6
```suggestion
// Collect groups [url(] [ |'|"]optional './|../', file part, query and end of url
```

flow-server/src/main/resources/plugins/theme-loader/theme-loader.js
```diff
@@ -3,7 +3,8 @@ const fs = require('fs');
...
-const urlMatcher = /(url\()(\'|\")?(\.\/|\.\.\/)(\S*)(\2\))/g;
+// takes into account that whitespace in file name and query parameters are valid
+const urlMatcher = /(url\()(['"])?(\.\/|\.\.\/)([^?'")]+)((\?\S*)?\2\))/g;
```
caalador, line 7:
```suggestion
const urlMatcher = /(url\()(['"])?(\.\/|\.\.\/)([^?\2)]+)(\?\S*)?(\2\))/g;
```
Would do a minor update for the regex so that we get end line and query parameter in their own groups (add the query to line 35) also change the internal `^?'"` to `^?\2`to enable the unlikely strings mixing the characters like: `url('./fonts/Roboto/roboto-v29-latin ext_"latin"-regular.woff2')`

caalador (changes requested):

flow-server/src/main/resources/plugins/theme-loader/theme-loader.js
pleku: >  also change the internal `^?'"` to `^?\2`

I had it like this, but changed it since I noticed that IDEA regex validation would not understand it and https://regexr.com/69jqg said that `\2` inside the `[^]` would "matches a STX character (char code 2)". 
It didn't occur to me how to use the backreference inside `[^]` so I just changed it to this but it is problematic indeed.
I'll need to investigate this more and think about what to do with `"` and `'` inside the url.

> Would do a minor update for the regex so that we get end line and query parameter in their own groups (add the query to line 35)

_end line_ ? Not sure I'm sure what you mean with that - the string can even continue on the same line with e.g. `format('woff2')` so please clarify what you mean. Also what is the point of separating the group, as the optional query part and the `")` ending need to be added back anyway.

caalador: Now the end result is a weird 
```
Group 5 |  ?v=1.12')
Group 6 |  ?v=1.12
```
so if you can get it to not generate 6 groups then it's fine to have the full end result, but the reading is hard with the current version.

I didn't note that it says matches the explicit character 2 instead of the group. I guess we can forget that part from this at least as I don't know if it is an actual use case.

pleku: Testing this, it seems to work but I'm not sure I understand _why_ it works 😅 as it claims that it is not matching
image

caalador: regex101 tester also shows that it works, but says that it matches a character 2 I guess the problem is if you have `2)` in the string? as ") doesn't match a url starting with ' and vice versa.

pleku: Well if I change it from `.woff2` to `.woff` then it still works ...

flow-server/src/main/resources/plugins/theme-loader/theme-loader.js
pleku: Changed to have only 5 groups

flow-server/src/main/resources/plugins/theme-loader/theme-loader.js
caalador: The current one just has the problem of `url('./fonts/Roboto/roboto-v29-latin-ext_la2)tin-regular.woff?v=1.12')` not matching

caalador: I guess the best would be to just get the whole url (query and all) and then for the file check drop any querypart, so have 2 regexes as this case seems too complex to make with a one liner (or it gets too complex to maintain)

pleku: At this point, it feels more painful to try to split it into two "reasonable" regexp instead of just going with one that works even when it is very complex: https://regexr.com/69jqg

Artur-: Why would we create our own URL parser when there are many perfectly fine ones out there?

pleku: Where is a new URL parser being created ? This PR tries to fix a broken regex that picks any `url( ... )` strings from css files and then replaces a part of the url address.

If there is a better way to do the url replacement, please suggest the way to do it.

caalador: In this case I would rather not support all the random cases as now the latest regex takes
image
in comparison to the simpler that fails in special cases where 2) is in the filename as the difference in complexity is quite noticable
image

pleku: Ok, good point. The JS code could then undergo further refactoring and for sure it would make sense to use something that is already available.

> I would rather not support all the random cases

Here I kind of disagree - whatever the solution is, it should support valid url() function parameters. In this case though, it can be limited to the fact that it should also be a valid file name as it only touches relative file urls.

pleku:
Closing in favor of other changes to the code that would be better instead of patching the current code.

PR #12376, opened 2021-11-16
fix: dev server serves only /VAADIN/* and sw.js

### LINKED ISSUE #12366: [Vite] Vite serves `index.html` directly when login view set in Spring Security configuration
Reproduction:
- Check out https://github.com/vaadin/fusion-crm-tutorial
- Update Vaadin version in `pom.xml` to `22.0-SNAPSHOT` (2021-11-16 or newer, as 22.0.0.beta3 has [an already fixed issue](https://github.com/vaadin/flow/issues/12351) that prevents this bug from manifesting)
- Run app and try to log in -> fails.
- Notice that the bootstrap page source returned for `/login` is the plain `index.html` as returned by Vite, hence missing the CSRF token added by `CsrfIndexHtmlRequestListener`, and other things.

Flow should never serve `index.html` from Vite, as the bootstrap page must go through the `IndexHtmlRequestHandler`.

joheriks:
Fixes #12366


Artur-:
This would prevent serving all other generated files like other enteypoints or other assets

Issue #12342, opened 2021-11-12
CallbackDataProvider blocking Session on long data Service anwser


lanmaster:
Vaadin 14. CallbackDataProvider. When service is slow and answer time is long then Grid connected to CallbackDataProvider is freezes with all UI.
Some example:
```
import com.example.application.views.MainLayout;
import com.vaadin.flow.component.grid.Grid;
import com.vaadin.flow.component.notification.Notification;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.data.provider.CallbackDataProvider;
import com.vaadin.flow.router.PageTitle;
import com.vaadin.flow.router.Route;
import com.vaadin.flow.router.RouteAlias;
import lombok.AllArgsConstructor;
import lombok.EqualsAndHashCode;
import lombok.Getter;
import lombok.Setter;
import org.apache.commons.lang3.RandomStringUtils;
import org.apache.commons.lang3.StringUtils;

import java.util.*;

@PageTitle("grid-with-sortable-columns")
@Route(value = "gridSortable", layout = MainLayout.class)
@RouteAlias(value = "gridSortable", layout = MainLayout.class)
public class GridWithSortableColumnsView extends VerticalLayout {

    public GridWithSortableColumnsView() {
        addClassName("dialogwithgrid-view");
        setSizeFull();

        // prepare Data
        List<GridRow> gridRows = new LinkedList<>();
        for (int i = 0; i < 10000; i++) {
            gridRows.add(new GridRow(i, RandomStringUtils.randomAlphanumeric(10)));
        }

        CallbackDataProvider<GridRow, Void> dataProvider = new CallbackDataProvider<>(
                query -> {

                    // Slow service answer is here
                    try {
                        Thread.sleep(2000);
                    } catch (InterruptedException e) {
                        e.printStackTrace();
                    }

                    Notification.show("Offset=" + query.getOffset() + " Limit=" + query.getLimit());
                    return gridRows.stream()
                            .sorted((o1, o2) -> StringUtils.compare(o1.getCol05(), o2.getCol05()))
                            .skip(query.getOffset())
                            .limit(query.getLimit());
                },
                query -> gridRows.size());

        Grid<GridRow> grid = new Grid<>();
        grid.setPageSize(10);
        grid.setSizeFull();
        grid.addColumn(GridRow::getId).setHeader("Id");
        grid.addColumn(GridRow::getCol05).setHeader("col05");
[...]

Issue #10709, opened 2021-04-19
Grid using DataProvider.fromCallbacks() freezing on scroll


lanmaster:
When i use `DataProvider.fromCallbacks()` scrolling down grid are freezing. As i think DataProvider start synchronous fetch operations. It is very unuseful. Can DataProvider use async mode?

miguelatvaadin:
Hi @lanmaster,

thank you very much for your question. DataProvider uses the flow communication mechanism which is synchronous. If you want to work in asynchronous mode then you should use the push feature (https://vaadin.com/docs/latest/flow/advanced/server-push).

Anyway I will add this to the enhancements backlog as I understand that this would be useful in some cases.

Best regards

lanmaster:
In my case CallbackDataProvider fetches data from heavy and slow service. There is blue line at page top occured almost always on fetching data. Can `fetchFromBackEnd` be more asyncronous? May be it can fetch data in parallel thread and use some callback when data is ready and it is time to update UI?

lanmaster:
Some example:
```
import com.example.application.views.MainLayout;
import com.vaadin.flow.component.grid.Grid;
import com.vaadin.flow.component.notification.Notification;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.data.provider.CallbackDataProvider;
import com.vaadin.flow.router.PageTitle;
import com.vaadin.flow.router.Route;
import com.vaadin.flow.router.RouteAlias;
import lombok.AllArgsConstructor;
import lombok.EqualsAndHashCode;
import lombok.Getter;
import lombok.Setter;
import org.apache.commons.lang3.RandomStringUtils;
import org.apache.commons.lang3.StringUtils;

import java.util.*;

@PageTitle("grid-with-sortable-columns")
@Route(value = "gridSortable", layout = MainLayout.class)
@RouteAlias(value = "gridSortable", layout = MainLayout.class)
public class GridWithSortableColumnsView extends VerticalLayout {

    public GridWithSortableColumnsView() {
        addClassName("dialogwithgrid-view");
        setSizeFull();

        // prepare Data
        List<GridRow> gridRows = new LinkedList<>();
        for (int i = 0; i < 10000; i++) {
            gridRows.add(new GridRow(i, RandomStringUtils.randomAlphanumeric(10)));
        }

        CallbackDataProvider<GridRow, Void> dataProvider = new CallbackDataProvider<>(
                query -> {

                    // Slow service answer is here
[...]

Issue #1066, opened 2016-06-27
When using web components in a template I want to be able to refer to the component class where imports are specified so that I do not have to specify them again for every template


Artur-:
This should be implemented as
`@Uses(Class<? extends Component.class>[] value)`

So that you can annotate any component class using `@Uses` and this will cause any HTML/JS/Stylesheet imports of the target class to be loaded also for this class.

PR #12378, opened 2021-11-17
feat: Add async mode to flush operation and API to turn it on

TatuLund:
In async mode data fetching is wrapped in future.

fixes: [#1066](https://github.com/vaadin/flow-components/issues/1066)
fixes: #10709
fixes: #12342

Draft for comments


flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -126,6 +130,11 @@
 
     private boolean fetchEnabled;
 
+    ExecutorService executor = Executors.newCachedThreadPool();
```
knoobie, line 133: What do you think about an method "setAsyncExecutorService"? This is probably needed if there is already a pool for such tasks created by the developer.

TatuLund: Yes, I just modified the solution

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -126,6 +129,10 @@
 
     private boolean fetchEnabled;
 
+    ExecutorService executor = null;
```
pleku, line 132: Nit: it should just be private

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -334,6 +341,21 @@ public void setRequestedRange(int start, int length) {
...
+     * defined and updates are done synchronously.
+     * <p>
+     * Note: This works only with Grid component. If set to true, Push needs to
+     * be enabled in order this to work.
+     * 
+     * @param executor
+     *            The ExecutorService used for async updates.
+     */
+    public void setExecutorForAsyncUpdates(ExecutorService executor) {
+        this.executor = executor;
```
pleku, line 356: Probably should do something to an old service or at least any pending request-future when for some reason a new executor is set

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -334,6 +341,21 @@ public void setRequestedRange(int start, int length) {
...
+     * component asynchronously or not. By default the executor service is not
+     * defined and updates are done synchronously.
+     * <p>
+     * Note: This works only with Grid component. If set to true, Push needs to
+     * be enabled in order this to work.
+     * 
+     * @param executor
+     *            The ExecutorService used for async updates.
+     */
+    public void setExecutorForAsyncUpdates(ExecutorService executor) {
```
pleku, line 355: This should validate via the component->getUI that the push mode is actually enabled, and probably also that it is _automatic_ mode since I'm not sure if this can ever work with manual push mode.

If push is not enabled, then it should IMO throw

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1026,6 +1048,7 @@ private String getInvalidContractMessage(String method) {
     }
 
     private void handleAttach() {
+        ui = UI.getCurrent();
```
pleku, line 1051: Using thread local should be avoided - instead use 
```suggestion
        ui = ((StateTree) stateNode.getOwner()).getUI();
```

pleku: Another option would be to create a method that gets the UI but maybe this is even safer in the long run as it always exists here and doesn't exist after `handleDetach` is called

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1048,6 +1071,7 @@ protected void handleDataRefreshEvent(DataRefreshEvent<T> event) {
     }
 
     private void handleDetach() {
+        ui = null;
```
pleku, line 1074: I think you need to cancel any pending future here too or things will go wrong if/when it returns
```suggestion
        ui = null;
        if (future != null) {
            future.cancel(true);
            future = null;
        }
```

pleku: Another question is that should a reference to the executor also be nullified - probably not desired as the grid could be just moved.

Anyway this needs to work so that a shared executor service never has tasks (to execute or to finish) that have references to data communicators that are detached already or its state has been invalidated. Thus you need to cancel the future also from `reset()` method and a few other setters.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1108,9 +1132,36 @@ private void flush() {
...
+            future = CompletableFuture
+                    .supplyAsync(() -> collectKeysToFlush(previousActive,
+                            effectiveRequested), executor);
+            future.thenAccept(activation -> {
+                if (ui == null) {
+                    return;
+                }
+                ui.access(() -> {
+                    performUpdate(oldActive, effectiveRequested, previousActive,
+                            activation);
```
pleku, line 1150: Based on the code, it should be so that a new request cannot change the state in the between as it would make any pending update cause unsync. issues. This is good.

However, this is not strictly enforced in any way in the code - what I'm trying to say is that I think there could be benefits for having
1) some encapsulation for the state of the data communicator update, like `oldActive, effectiveRequested, previousActive, activation` here but in a POJO
2) when the returning future is being processed, it would verify that the state that it expects is still valid - nothing has broken things in the meanwhile

This is maybe slightly overengineering, but I see that in the long run it could help making more maintainable code and not cause regressions if there are some changes done. Not sure if I'm able to explain my concern, but I tried.

flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorAsyncTest.java
```diff
@@ -0,0 +1,286 @@
...
+import com.vaadin.flow.component.UI;
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.internal.Range;
+import com.vaadin.flow.server.VaadinRequest;
+import com.vaadin.flow.server.VaadinService;
+import com.vaadin.flow.server.VaadinSession;
+
+import elemental.json.JsonValue;
+
+public class DataCommunicatorAsyncTest {
```
pleku, line 28: I would want to see a test, in Flow, where some component is waiting for the job to be executed, and the user refreshes the browser in the middle of that. It plays a bit differently depending on whether `@PreserveOnRefresh` is used or not as the component is / is not thrown away. Something like this would be bound to happen and it should just work, which I'm not sure based on the current implementation that it does (without some changes).

pleku (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
pleku: The API would be in the component to pass the items and the executor service or just the executor service itself.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
TatuLund: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
TatuLund: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
TatuLund: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
TatuLund: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
TatuLund: Done

flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorAsyncTest.java
TatuLund: Not possible to make this non-flake

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1108,9 +1147,36 @@ private void flush() {
...
+                future.cancel(true);
+            }
+            future = CompletableFuture
+                    .supplyAsync(() -> collectKeysToFlush(previousActive,
+                            effectiveRequested), executor);
+            future.thenAccept(activation -> {
+                if (ui == null) {
+                    return;
+                }
+                ui.access(() -> {
```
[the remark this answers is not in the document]
TatuLund: ignore

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1108,9 +1147,36 @@ private void flush() {
...
+        if (executor != null) {
+            // In async mode wrap fetching data in future, collectKeysToFlush
+            // will perfrom fetch from data provider with given range.
+            if (future != null) {
+                future.cancel(true);
+            }
+            future = CompletableFuture
+                    .supplyAsync(() -> collectKeysToFlush(previousActive,
+                            effectiveRequested), executor);
+            future.thenAccept(activation -> {
```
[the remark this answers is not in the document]
TatuLund: ignore

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1108,9 +1146,36 @@ private void flush() {
         resendEntireRange |= !(previousActive.intersects(effectiveRequested)
                 || (previousActive.isEmpty() && effectiveRequested.isEmpty()));
 
-        Activation activation = collectKeysToFlush(previousActive,
-                effectiveRequested);
+        if (executor != null) {
+            // In async mode wrap fetching data in future, collectKeysToFlush
+            // will perfrom fetch from data provider with given range.
```
joheriks, line 1151:
```suggestion
            // will perform fetch from data provider with given range.
```

joheriks (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -334,6 +342,30 @@ public void setRequestedRange(int start, int length) {
...
+     * defined and updates are done synchronously.
+     * <p>
+     * Note: This works only with Grid component. If set to true, Push needs to
+     * be enabled in order this to work.
+     * 
+     * @param executor
+     *            The ExecutorService used for async updates.
+     */
+    public void setExecutorForAsyncUpdates(ExecutorService executor) {
+        if (ui.getPushConfiguration().getPushMode() != PushMode.DISABLED) {
```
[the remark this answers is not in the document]
TatuLund: Actually I need to move the check from here to elsewhere in order it to work reliably. At this point of time there is some chance that ui is null, but even more so a timing conflict. If you have set Push with annotation, ui.getPushConfiguration().getPushMode() will still return disabled here. I will throw the exception in flush instead. I tested it and seemed to work reliably. Also that provoked the test to fail as there were insufficient mocks. I'll fix that too.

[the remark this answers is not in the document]
TatuLund: I will also add unit test for it.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -334,6 +342,30 @@ public void setRequestedRange(int start, int length) {
         requestFlush();
     }
 
+    /**
+     * Control whether DataCommunicator should push data updates to the
+     * component asynchronously or not. By default the executor service is not
+     * defined and updates are done synchronously.
+     * <p>
+     * Note: This works only with Grid component. If set to true, Push needs to
+     * be enabled in order this to work.
```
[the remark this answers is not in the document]
TatuLund: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -334,6 +342,30 @@ public void setRequestedRange(int start, int length) {
...
+     * 
+     * @param executor
+     *            The ExecutorService used for async updates.
+     */
+    public void setExecutorForAsyncUpdates(ExecutorService executor) {
+        if (ui.getPushConfiguration().getPushMode() != PushMode.DISABLED) {
+            throw new IllegalStateException(
+                    "Asynchronous DataCommunicator updatres require Push to be enabled and PushMode.AUTOMATIC");
```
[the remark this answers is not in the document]
TatuLund: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -334,6 +342,30 @@ public void setRequestedRange(int start, int length) {
...
+        }
+        if (this.executor != null) {
+            future.cancel(true);
```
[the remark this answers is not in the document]
TatuLund: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -334,6 +342,30 @@ public void setRequestedRange(int start, int length) {
...
+    /**
+     * Control whether DataCommunicator should push data updates to the
+     * component asynchronously or not. By default the executor service is not
+     * defined and updates are done synchronously.
+     * <p>
+     * Note: This works only with Grid component. If set to true, Push needs to
+     * be enabled in order this to work.
+     * 
+     * @param executor
+     *            The ExecutorService used for async updates.
```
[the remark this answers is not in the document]
TatuLund: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
TatuLund: Done

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -334,6 +342,29 @@ public void setRequestedRange(int start, int length) {
...
+     */
+    public void setExecutorForAsyncUpdates(ExecutorService executor) {
+        if (this.executor != null) {
+            if (future != null) {
+                future.cancel(true);
+                future = null;
+            }
+            this.executor.shutdown();
```
joheriks, line 363: Shutting down here would preclude having a single `ExecutorService` thread pool shared among several `DataCommunicator` and potentially other instances. Could this method instance accept an `Executor` instead since `DataCommunicator` only needs to create threads rather than manage the thread pool.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -334,6 +342,29 @@ public void setRequestedRange(int start, int length) {
...
+     * defined and updates are done synchronously. Setting to null will disable
+     * the feature.
+     * <p>
+     * Note: This works only with Grid component. If set to true, Push needs to
+     * be enabled and set to PushMode.AUTOMATIC in order this to work.
+     * 
+     * @param executor
+     *            The ExecutorService used for async updates.
+     */
+    public void setExecutorForAsyncUpdates(ExecutorService executor) {
```
joheriks, line 357: I'd suggest modifying the signature to `enablePushUpdates(Executor executor)` to more clearly inidicate that this method primarily enables the async updates. See next comment for argument type change motivation.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
TatuLund: Agree, semantically it is better. Done changes.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
TatuLund: Done

joheriks (approved):

PR #12385, opened 2021-11-17
test: Clear threadlocals before running tests

Artur-:


flow-test-generic/src/main/java/com/vaadin/flow/testutil/CurrentInstanceCleaner.java
```diff
@@ -0,0 +1,18 @@
+package com.vaadin.flow.testutil;
```
manolo, line 1: Failing in CI with error:
```
java.lang.ClassNotFoundException: com.vaadin.flow.testutil.CurrentInstanceCleaner
```
but `flow-server` depends on `flow-test-generic`, that is weird

manolo (changes requested):

flow-test-generic/src/main/java/com/vaadin/flow/testutil/CurrentInstanceCleaner.java
Artur-: It fails because of `build-tools` which runs tests but does not depend on `flow-test-generic`

Legioth:
Would it be better to instead explicitly check after each test that it hasn't leaked any of our thread locals? In that way, we could catch situations where framework code accidentally leaks.

manolo:
platform has already the test for versions https://github.com/vaadin/platform/pull/2549 you can remove `build-tools` module

Artur-:
> Would it be better to instead explicitly check after each test that it hasn't leaked any of our thread locals
It is not interesting in 99,9% of the tests if it leaks or not. What is interesting it that a test is run without thread locals that potentially breaks it

manolo (approved):

manolo (dismissed):

manolo:
flow-client does not depend on flow-test-generic, should we include the dependency in the module or remove the listener ?

manolo (approved):

manolo:
Now the dnd module

Artur-:
For `flow-tests` and `flow-plugin-base` it should probably be turned off

manolo (approved):

joheriks (approved):

Issue #12392, opened 2021-11-18
Vite projects do not work with npm 7+


Artur-:
### Description of the bug / feature
```
npm ERR! While resolving: no-name@0.9.4
npm ERR! Found: vite@2.7.0-beta.5
npm ERR! node_modules/vite
npm ERR!   dev vite@"v2.7.0-beta.5" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer vite@"^2.0.0" from vite-plugin-checker@0.3.4
npm ERR! node_modules/vite-plugin-checker
npm ERR!   dev vite-plugin-checker@"0.3.4" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

For now the workaround is to run `npm install -f` manually

This will be resolved at least when Vite 2.7.0 stable is out

joheriks:
Note issue is fixed temporarily by #12397, but keeping ticket open until stable release of Vite is used. The code that adds the `--legacy-peer-deps` parameter to npm should then be removed and this ticket closed.

Artur-:
The code was removed in #12522

PR #12397, opened 2021-11-18
fix: support Vite with npm 7

joheriks:
Pass --legacy-peer-deps when installing npm dependencies
to allow vite and vite-plugin-checker to coexist. This is a
temporary tweak to be reverted when the vite dep is a stable
version.

Temporary fix for #12392


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -351,7 +358,23 @@ private void runNpmInstall() throws ExecutionFailedException {
...
-        } catch (IllegalStateException exception) {
+
+            // Remove the following conditional block (together with the
+            // npmLegacyPeerDeps parameter) when vite has been updated a
+            // version accepted as a peer dependency by vite-plugin-checker.
+            // https://github.com/vaadin/flow/issues/12392
+            if (!enablePnpm && npmLegacyPeerDeps) {
+                ArrayList<String> versionCommand = new ArrayList<>(executable);
+                versionCommand.add("--version");
+                if (FrontendUtils.getVersion("node", versionCommand)
```
Artur-, line 369: Rather make `FrontendTools.getNpmVersion` public and use that

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
joheriks: Done

Artur- (approved):

PR #12400, opened 2021-11-19
CP: Element and Component as event data support for 2.8

pleku:
Do not squash!


pleku:
APIDiff gives a "false positive" as from the two added interface methods with default implementations. To my understanding, those only break API if an extending interface or implementing class have the exactly same method signature existing already, which is highly unlikely at this point.

<html>
<body>


Status | Modifier | Type | Method | Exceptions | Compatibility Changes: | Line Number
-- | -- | -- | -- | -- | -- | --
NEW (!) | public | com.vaadin.flow.dom.DomListenerRegistration | addEventDataElement(java.lang.String) | n.a. | Change : METHOD_NEW_DEFAULT,  METHOD_ADDED_TO_INTERFACE | 306
NEW (!) | public | com.vaadin.flow.dom.DomListenerRegistration | mapEventTargetElement() | n.a. | Change: METHOD_NEW_DEFAULT, METHOD_ADDED_TO_INTERFACE |  277

</body>
</html>

joheriks (approved):

PR #12402, opened 2021-11-19
feat: Make it easier to see the relevant part of InternalServerError

### LINKED ISSUE #9493: Make it easier to see the relevant part of a navigation error
When navigating to a view that throws an exception in its constructor, I'm presented with a basic HTML page that shows the exception message (and also the stacktrace if production mode is not enabled).

Screenshot 2020-11-25 at 13 24 05

An experienced developer looking at this message probably knows that the most relevant piece of information is the last part of the stacktrace and the end of the error message. A less experienced developer might not be aware of this gotcha and instead focus on the beginning of the message that conveys little useful information (e.g. `Error creating bean with name 'com.example.application.views.helloworld.HelloWorldView'`).

One suggestion for how to make the message easier to understand is to bold the last segment of the message to make it stand out.

Screenshot 2020-11-25 at 13 25 37

tarekoraby:
Fixes #9493

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/InternalServerError.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/InternalServerError.java
@@ -49,17 +49,17 @@ public class InternalServerError extends Component
     public int setErrorParameter(BeforeEnterEvent event,
             ErrorParameter<Exception> parameter) {
         String exceptionText;
-        String errorTextStem =
-                "There was an exception while trying to navigate to '%s'";
+        String errorTextStem = "There was an exception while trying to navigate to '%s'";
         String rootCause = getRootCause(parameter);
-        boolean isRootCauseAvailable = rootCause != null && !rootCause.isEmpty();
+        boolean isRootCauseAvailable = rootCause != null
+                && !rootCause.isEmpty();
         if (isRootCauseAvailable) {
-            exceptionText = String.format(errorTextStem
-                            + " with the root cause '%s'",
+            exceptionText = String.format(
+                    errorTextStem + " with the root cause '%s'",
                     event.getLocation().getPath(), rootCause);
         } else if (parameter!= null && parameter.hasCustomMessage()) {
-            exceptionText = String.format(errorTextStem
-                            + " with the exception message '%s'",
+            exceptionText = String.format(
+                    errorTextStem + " with the exception message '%s'",
                     event.getLocation().getPath(),
                     parameter.getCustomMessage());
         } else {
@@ -89,8 +89,7 @@ public class InternalServerError extends Component
     }
 
     private void reportException(Exception exception, String path,
-                                 String exceptionText,
-                                 boolean isRootCauseAvailable) {
+            String exceptionText, boolean isRootCauseAvailable) {
         if (isRootCauseAvailable) {
             getElement()
                     .appendChild(ElementFactory.createHeading3(exceptionText));
--- a/flow-server/src/main/java/com/vaadin/flow/router/InternalServerError.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/InternalServerError.java
@@ -49,7 +49,17 @@ public class InternalServerError extends Component
     public int setErrorParameter(BeforeEnterEvent event,
             ErrorParameter<Exception> parameter) {
         String exceptionText;
-        if (parameter.hasCustomMessage()) {
+        boolean isRootCauseAvailable = false;
+        String rootCause = getRootCause(parameter);
+
+        if (rootCause != null && !rootCause.isEmpty()) {
+            isRootCauseAvailable = true;
+            exceptionText = String.format(
+                    "There was an exception while trying to navigate to '%s'"
+                            + " with the root cause '%s'",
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/router/InternalServerError.java
```diff
@@ -49,7 +49,17 @@
     public int setErrorParameter(BeforeEnterEvent event,
             ErrorParameter<Exception> parameter) {
         String exceptionText;
-        if (parameter.hasCustomMessage()) {
+        boolean isRootCauseAvailable = false;
+        String rootCause = getRootCause(parameter);
+
+        if (rootCause != null && !rootCause.isEmpty()) {
+            isRootCauseAvailable = true;
```
joheriks, line 58:
```suggestion
        String rootCause = getRootCause(parameter);
        boolean isRootCauseAvailable = rootCause != null && !rootCause.isEmpty();
        if (isRootCauseAvailable) {
```

flow-server/src/main/java/com/vaadin/flow/router/InternalServerError.java
```diff
@@ -137,6 +152,29 @@ private void checkLogBinding() {
...
+        if (cause.getCause() == null) {
+            return cause.toString();
+        } else {
+            if (availableRecursiveLevels < 1) {
+                return null;
+            }
+            return doGetRootCause(cause.getCause(), --availableRecursiveLevels);
+        }
+    }
+
```
joheriks, line 176: I suggest writing this as one method and using iteration rather than (tail) recursion. Secondly, assuming the `availableRecursiveLevels` is there to catch recursive exception (`cause[.getCause()]*==cause`), I think this would be better caught with checking whether the cause in the loop is the original cause. An arbitrary constant like 10 may give an arbitrary root cause in a long recursive exception.

joheriks (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/InternalServerError.java
tarekoraby: > I suggest writing the two methods getRootCause and doGetRootCause as one method and using iteration rather than (tail) recursion. 

Sure, we can do that. But what's the advantage of using iteration over recursion? Is it a functional or aesthetic recommendation?

> An arbitrary constant like 10 may give an arbitrary root cause in a long exception chain (such as a stack overflow).

As it's currently implemented, if the arbitrary constraint (say 10) is met, then `getRootCause()` will return `null`, and the implementation will default to the current output from `parameter.hasCustomMessage()`.

knoobie: This looks rather complicated as a whole; there are a  lot of good "find root cause" "algorithmn" out there - you could get some inspiration from e.g. [Spring](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/NestedExceptionUtils.java) or [Apache](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java#L232)

tarekoraby: Thanks @knoobie! I pushed a refactoring based on the Spring implementation. 

The whole thing could be simplified even more if `parameter.hasCustomMessage()` is discarded altogether. I'm unsure though if there is any edge case in which the root cause would be unavailable for some reason. Any thoughts?

knoobie: Thanks for considering! Never say never ;) I would probably check for it, just in case somebody refactores the implementation of no route found to be a specific exception without a cause.

joheriks (changes requested):
In addition to the message addition `"with the root cause"` could we also have an explicit mention in the same string that the full stack trace can be seen in the server log. After this the code also need to be formatted correctly (run `mvn formatter:format` in the `flow` project).

tarekoraby:
Actually, the full exception will continue to be printed under this `h3` element. So then we will have to say something like "see full stack trace below or in the server logs". 
I slightly prefer not to add this extra note as I feel that the stack trace will not be missed (but I have no super-strong objections if you insist on adding it).

joheriks:
> Actually, the full exception will continue to be printed under this `h3` element. So then we will have to say something like "see full stack trace below or in the server logs". I slightly prefer not to add this extra note as I feel that the stack trace will not be missed (but I have no super-strong objections if you insist on adding it).

My main worry was that the beginning of the `<pre>` element could be misunderstood as being the stack trace of the root cause (tempting to copy-paste the `<h3>` line + only the first stack trace believing this to be the root cause into a ticket / chat) without some clear identification what is what. But on second thought the `<h3>` plus the gray background of the section serves to delineate them sufficiently, so I would be okay with this.

Screenshot 2021-11-24 at 15 57 57

(As a separate ticket/PR, a further DX improvement would be to make the stack trace easily copyable via the lower right-hand gizmo).

joheriks (approved):

PR #12409, opened 2021-11-19
feat: add HeartbeatListener to UI

### LINKED ISSUE #12295: Make HeartbeatListener public (moving outside of internal package)
### Feature Request
Currently the `HeartbeatListener` is inside the internal package "intended for internal use by the framework" which in turn implies, that it's not save to use in the long run, therefore I would request to make `HeartbeatListener` public.

Reasoning: In the past (v8) there were a lot of application that used to overwrite the `HeartbeatHandler` to calculate the time of idle in an application to show e.g. a counter to the user or some form of "idle warning". This used to be kinda invasive, because a lot of classes had to be extended. 

For example, the VaadinServletService with:
```java

  @Override
  protected List<RequestHandler> createRequestHandlers() throws ServiceException {
    List<RequestHandler> handlers = super.createRequestHandlers();
    for (int i = 0; i < handlers.size(); ++i) {
      if (handlers.get(i) instanceof HeartbeatHandler) {
        handlers.set(i, new CustomHearthbeatHandler());
      }
    }
    return handlers;
  }

```

Now that flow has a dedicated listener for such events in form of `HeartbeatListener` - I would like to use this, instead of overwriting yet again a lot of classes. This listener could be moved to the other listener like `UIInitListener` inside `VaadinService` or made accessible through the `UI`.

knoobie:
## Description

Moves `HeartbeatEvent` and `HeartbeatListener` from the internal package into the public packages.  
Creates a new method `UI::addHeartbeatListener` as public API to use instead of the internal API from `UIInternal`.

Fixes #12295 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [ x I have performed self-review and corrected misspellings.

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/communication/HeartbeatHandlerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/communication/HeartbeatHandlerTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.server.communication;
+
+import java.io.IOException;
+
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import com.vaadin.flow.component.UI;
+import com.vaadin.flow.component.internal.UIInternals;
+import com.vaadin.flow.server.VaadinRequest;
+import com.vaadin.flow.server.VaadinResponse;
+import com.vaadin.flow.server.VaadinService;
+import com.vaadin.flow.server.VaadinSession;
+
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.when;
+
+public class HeartbeatHandlerTest {
+
+    @Test
+    public void synchronizedHandleRequest_uiPresent_setLastHeartbeatTimestampIsCalledOnce()
+            throws IOException {
+        VaadinService service = mock(VaadinService.class);
+        VaadinSession session = mock(VaadinSession.class);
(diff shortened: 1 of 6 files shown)


flow-server/src/test/java/com/vaadin/flow/component/internal/UIInternalsTest.java
```diff
@@ -125,47 +121,6 @@ public void init() {
         Mockito.when(ui.getSession()).thenReturn(session);
     }
 
-    @Test
-    public void heartbeatTimestampSet_heartbeatListenersAreCalled() {
```
knoobie, line 129: Tests are just moved to the UI

flow-server/src/test/java/com/vaadin/flow/server/communication/HeartbeatHandlerTest.java
```diff
@@ -0,0 +1,57 @@
...
+import com.vaadin.flow.server.VaadinResponse;
+import com.vaadin.flow.server.VaadinService;
+import com.vaadin.flow.server.VaadinSession;
+
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.when;
+
+public class HeartbeatHandlerTest {
```
knoobie, line 35: Instead of creating a IT module to test the "new feature", I opted for this addition.  Because `HeartbeatHandler`  had no tests and it doesn't blow up the CI build-time for a feature that is here for ages. Now it's at least verified, that `setLastHeartbeatTimestamp` is called for a UI.

flow-server/src/test/java/com/vaadin/flow/component/internal/UIInternalsTest.java
caalador: Did these need moving at all as they test the internals feature so they should perhaps be here and only have the new test added to UI

knoobie: They didn't have to - my idea behind this was to allow easier deprecation and later removal of the method inside UIInterals. I can move them back and add more tests to the UI if you want

caalador: If you could it would be better to have the tests "close" to the implementation as it's clearer where the feature is tested.

knoobie: Done. (failed test looks unrelated)

caalador (approved):

caalador:
Seems you made the PR just before `fix: Stabilize vite dev-mode ITs` went in. Mind updating on top of master so that it makes the now test more stable and not so random.

knoobie:
Just merged master into it, let me know if that works for you :)

PR #12424, opened 2021-11-23
fix: Make linking to packages in the file system work

Artur-:
Fixes https://github.com/vaadin/fusion/issues/126


Artur-:
This would also help in creating tests for #12512

joheriks (approved):

joheriks:
This makes sense to have in the LTS but needs a manual cherry-pick since the test module doesn't exist there. Should be fine to merge untested into `2.8` since `NodeUpdater` changes are going to come from `master` so regressions would be caught earlier.

ZheSun88:
CP 2.8 : https://github.com/vaadin/flow/pull/14590

PR #12430, opened 2021-11-23
chore: Make formatter:format work again

Artur-:


pom.xml
```diff
@@ -296,20 +296,11 @@
                 <artifactId>formatter-maven-plugin</artifactId>
                 <version>2.16.0</version>
                 <configuration>
-                    <configFile>eclipse/VaadinJavaConventions.xml
-                    </configFile>
+                    <configFile>https://raw.githubusercontent.com/vaadin/flow/9.0/eclipse/VaadinJavaConventions.xml</configFile>
```
manolo, line 299: I would use `master` instead, to assure that latest conventions are loaded.
Contributor [doc guide](https://vaadin.com/docs/latest/contributing/overview) links to `master` as well.

manolo (changes requested):

pom.xml
Artur-: That's fine, assuming we will never ever move it out from `master`. Otherwise the build for all old Flow versions will fail the day it is removed from `master`

manolo: Makes sense too. My only concern is that devs would update the file in master and they might never cherry pick it to 9.0, probably no one remembers that.

Artur-: Changed to `master`. We can consider moving it later

manolo (approved):

PR #12431, opened 2021-11-23
fix: Make ClassFinder deterministic

### LINKED ISSUE #7556: Support reproducible flow-build-info.json and stats.json
Currently (v14) generated `flow-build-info.json` and `stats.json` entries are randomly reordered during build. In containerized environment it means that every build produces an unique image which creates problems both in development and CD workflows.

This should be also taken into account in relation to #6398

Artur-:
Fixes #7556

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/ClassFinder.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/ClassFinder.java
@@ -18,8 +18,10 @@ package com.vaadin.flow.server.frontend.scanner;
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -40,7 +42,7 @@ public interface ClassFinder extends Serializable {
      * list of classes.
      */
     class DefaultClassFinder implements ClassFinder {
-        private final Set<Class<?>> classes;
+        private final LinkedHashSet<Class<?>> classes;
 
         private final transient ClassLoader classLoader;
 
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/ClassFinder.java
```diff
@@ -70,7 +80,7 @@ public DefaultClassFinder(Set<Class<?>> classes) {
         public DefaultClassFinder(ClassLoader classLoader,
                 Class<?>... classes) {
             this.classLoader = classLoader;
-            this.classes = new HashSet<>();
+            this.classes = new LinkedHashSet<>();
             for (Class<?> clazz : classes) {
                 this.classes.add(clazz);
             }
```
caalador, line 86: Shouldn't this add also be sorted for the classes array.
So something like `Arrays.stream(classes).sorted(Comparator.comparing(Class::getName)).forEach(this.classes::add);`
As there is nothing guaranteeing that the classes is always ordered when given.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/ClassFinder.java
Artur-: It is not needed for ClassFinder to be deterministic. For the Set constructor it is needed as the iteration order for the same classes differ when run multiple times. Wouldn't it be expected that the classes are scanned in the order they are given?

caalador: Also it seems this constructor is used only in tests.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdateImportsTest.java
```diff
@@ -393,6 +395,45 @@ Logger log() {
...
+        JsonObject fallback = fallBackData.getObject("chunks")
+                .getObject("fallback");
+        JsonArray jsModules = fallback.getArray("jsModules");
+        JsonArray cssImports = fallback.getArray("cssImports");
+
+        String expectedJsModules = "[\"@polymer/e.js\",\"@polymer/D.js\",\"@polymer/c.js\",\"@polymer/b.js\",\"@polymer/a.js\",\"./extra-javascript.js\"]";
+        String expectedCssImports = "[{\"value\":\"./b-css.css\"},{\"include\":\"a-a\",\"value\":\"./a-css.css\"},{\"include\":\"extra-bar\",\"themeFor\":\"extra-foo\",\"value\":\"./extra-css.css\"}]";
+
+        Assert.assertEquals(expectedJsModules, jsModules.toJson());
+        Assert.assertEquals(expectedCssImports, cssImports.toJson());
```
caalador, line 434: Should also assert the tokenFile content for the written data here.

caalador (changes requested):
A more certain place to keep the build file in order would be to order the written imports in `TaskUpdateImports::populateFallbackData`

At least the token file should be asserted in the test

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdateImportsTest.java
Artur-: That can be done although the token file contains nothing else than the fallback chunk

Artur-: Added

caalador (approved):

PR #12432, opened 2021-11-24
fix: Do not decode + as space

### LINKED ISSUE #12354: Double decoding of query parameters when creating Location
When I open an URL with url encoded query parameters where a parameters contains a "+" symbol, the query parameter is incorrectly decoded twice and the "+" is removed from parameter value. This happens in `com.vaadin.flow.router.Location#Location(java.lang.String)` which calls the `com.vaadin.flow.router.LocationUtil#parseQueryParameters` where query string is first extracted `query = new java.net.URI(location).getQuery();` - this decodes the url encoded string for the first time, the string at this point is correct and "+" symbol gets decoded from "%2B". Next the `QueryParameters.fromString(query)` is called which calls `com.vaadin.flow.router.QueryParameters#makeQueryParamList` and this decodes the string again using `value = URLDecoder.decode(value, "utf-8");` - this removed the "+" symbol from the already decoded string and the parameter value is incorrect.

E.g. the encoded parameter value `someValue1%2BsomeValue2` which should be decoded as `someValue1+someValue2` gets decoded as `someValue1 someValue2` instead which is wrong.

### Versions:
    - Vaadin / Flow version: 14.7.5 / 2.7.4
    - Java version: 11

caalador:
Fix issue with URLDecoder handling
of '+' character.

Fixes #12354


flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -130,7 +130,7 @@ public static QueryParameters fromString(String queryString) {
             return Collections.singletonList(paramAndValue);
         }
         String param = paramAndValue.substring(0, index);
-        String value = paramAndValue.substring(index + 1);
+        String value = paramAndValue.substring(index + 1).replace("+", "%2B");
         try {
             value = URLDecoder.decode(value, "utf-8");
```
Artur-, line 135: The sustainable fix would be to add `UrlUtil.decode` and `UrlUtil.decodeURIComponent` that would work the same way as the `encode` and `encodeURIComponent` methods and then use the proper one here. It is not only about `+` but also a number of other characters

caalador: For decode the only cases are:
```
            case '+':
                sb.append(' ');
            case '%':
                /*
                 * Starting with this instance of %, process all
                 * consecutive substrings of the form %xy. Each
                 * substring %xy will yield a byte. Convert all
                 * consecutive  bytes obtained this way to whatever
                 * character(s) they represent in the provided
                 * encoding.
                 */
                 ...
```
Everything else is added as is.

caalador: Also in this specific flow when coming from Location we have the decoded query string from `java.net.URI(location).getQuery()`

Artur- (approved):

PR #12433, opened 2021-11-24
chore: Replace Assert.assertThat with non-deprecated MatcherAssert

Artur-:


flow-server/src/test/java/com/vaadin/flow/router/RouterTest.java
```diff
@@ -72,7 +74,6 @@
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.startsWith;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertThat;
```
caalador, line 75: Could have just updated the static import for this class to be `import static org.hamcrest.MatcherAssert.assertThat;`

caalador (approved):

PR #12434, opened 2021-11-24
feat: allow adding true modal dialogs to UI

### LINKED ISSUE #10172: An opened Dialog component does not disable the UI to which it belongs.
### Description
An opened Dialog component does not actually disable the UI to which it belongs. The Dialog's parent UI **remains active server-side** and can be interacted with, if the Dialog is removed from the DOM by using DevTools.

### Expected outcome
The parent UI of a Dialog should remain disabled **server-side** while a Dialog is opened, and not be interactive.

### Actual outcome
The parent UI can be enabled client-side by using Browser Developer Tools: delete the Dialog, disable "pointer-events: none" inline style on BODY element

### Live Demo

```
Dialog dialog = new Dialog();
dialog.add("this dialog");
dialog.setCloseOnOutsideClick(false);
dialog.setCloseOnEsc(false);
		
dialog.open();	
```

### Steps to reproduce
1. Create a Dialog
2. Disable closing the dialog on ESC and Outside Click
3. Open the Dialog
4. Enter Developer Tools
5. Delete the Dialog (remove it from the DOM)
6. Disable the "pointer-events: none" inline style from the BODY element

### Browsers Affected
- [x ] Chrome
- [ x] Firefox
- [ x] Safari
- [ x] Edge
- [ x] IE 11
- [ x] iOS Safari
- [ x] Android Chrome

pleku:
Introduces the possibility to add a modal dialog to a UI in a way
that the server side will ignore interaction with all other components
(from client side) until the modal dialog is removed. Modal dialogs can
be stacked on top of each other. This is similar to setting a component disabled,
but is not the same and does not show in the UI in any way - it would be
up to the modal dialog to show a modalityy curtain on top of the outside
UI to the dialog. The feature is thus a server side security feature,
and makes the framework enforce
https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal guidelines.

The implementation relies on a new server side only state tree feature InertData, which
makes it possible to make a component/element/statenode and its children not accept any
updates from client side, but it not the same as disabled. The inert
state is by default inherited to child nodes, except that any child can declare
itself to ignore parent inert state which means that updates are allowed for
itself and its children, unless those are explicitly made inert.

WIP:
- missing bunch of integration tests for trying to break inept UI,
which are TODOs in InertComponentIT
- this also disables all UI level listeners - like UI resize listener, which might not be desirable
- IT for checking that when `@PreserverOnRefresh` is used, the inert state of `UI` is still intact

Resolves #10172

### DIFF
--- a/flow-client/src/main/frontend/Flow.ts
+++ b/flow-client/src/main/frontend/Flow.ts
@@ -134,12 +134,9 @@ export class Flow {
   }
 
   loadingFinished() {
-    // NEVER CALLED WHEN INERT LINK CLICKED!
     // Make Testbench know that server request has finished
     this.isActive = false;
     $wnd.Vaadin.connectionState.loadingFinished();
-    // @ts-ignore
-    $wnd.console.log('loadingFinished');
   }
 
   private get action(): (params: NavigationParameters) => Promise<HTMLRouterContainer> {
@@ -188,8 +185,6 @@ export class Flow {
       // The callback to run from server side to cancel navigation
       this.container.serverConnected = (cancel) => {
         resolve(cmd && cancel ? cmd.prevent() : {});
-        // @ts-ignore
-        $wnd.console.log('flowLeave: this.container.serverConnected');
         this.loadingFinished();
       };
 
@@ -214,8 +209,6 @@ export class Flow {
             this.container.style.display = '';
             resolve(this.container);
           }
-          // @ts-ignore
-          $wnd.console.log('flowNavigate: this.container.serverConnected');
           this.loadingFinished();
         };
 
--- a/flow-client/src/main/frontend/Flow.ts
+++ b/flow-client/src/main/frontend/Flow.ts
@@ -134,9 +134,12 @@ export class Flow {
   }
 
   loadingFinished() {
+    // NEVER CALLED WHEN INERT LINK CLICKED!
     // Make Testbench know that server request has finished
     this.isActive = false;
     $wnd.Vaadin.connectionState.loadingFinished();
+    // @ts-ignore
+    $wnd.console.log('loadingFinished');
   }
 
   private get action(): (params: NavigationParameters) => Promise<HTMLRouterContainer> {
@@ -185,6 +188,8 @@ export class Flow {
       // The callback to run from server side to cancel navigation
       this.container.serverConnected = (cancel) => {
         resolve(cmd && cancel ? cmd.prevent() : {});
+        // @ts-ignore
+        $wnd.console.log('flowLeave: this.container.serverConnected');
         this.loadingFinished();
       };
 
@@ -209,6 +214,8 @@ export class Flow {
             this.container.style.display = '';
             resolve(this.container);
           }
+          // @ts-ignore
+          $wnd.console.log('flowNavigate: this.container.serverConnected');
           this.loadingFinished();
         };
 
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -16,6 +16,7 @@
 package com.vaadin.flow.component.internal;
 
 import java.io.Serializable;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -25,12 +26,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentUtil;
@@ -47,6 +46,7 @@ import com.vaadin.flow.component.internal.ComponentMetaData.DependencyInfo;
 import com.vaadin.flow.component.page.ExtendedClientDetails;
 import com.vaadin.flow.component.page.Page;
 import com.vaadin.flow.component.page.Push;
+import com.vaadin.flow.dom.ElementUtil;
 import com.vaadin.flow.dom.impl.BasicElementStateProvider;
 import com.vaadin.flow.function.DeploymentConfiguration;
 import com.vaadin.flow.internal.AnnotationReader;
@@ -78,6 +78,8 @@ import com.vaadin.flow.server.frontend.FallbackChunk;
 import com.vaadin.flow.server.frontend.FallbackChunk.CssImportData;
 import com.vaadin.flow.shared.Registration;
 import com.vaadin.flow.shared.communication.PushMode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Holds UI-specific methods and data which are intended for internal use by the
@@ -206,6 +208,8 @@ public class UIInternals implements Serializable {
     private ExtendedClientDetails extendedClientDetails = null;
 
     private boolean isFallbackChunkLoaded;
+
+    private ArrayDeque<Component> modalComponentQueue;
 
     /**
      * Creates a new instance for the given UI.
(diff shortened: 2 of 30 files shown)


knoobie:
Just a note: this could be really awesome for other use cases as well, like blocking the UI while a long running task is enqueued - e.g. pressing a button and ignoring every other client action until the server is done.

pleku:
> Just a note: this could be really awesome for other use cases as well, like blocking the UI while a long running task is enqueued - e.g. pressing a button and ignoring every other client action until the server is done.

Ok, thanks for the suggestion. I think then I'll retain the API in the `UI` to keep it visible, but I have to think about the naming.
I would think that in those cases - when the UI is blocked - there would be a notification/dialog shown to the user that tells them to wait, and typically that would be something that has a modality-curtain shown too (like `Dialog`).

knoobie:
Having this API inside the UI would be really good way to discover it! I wanted to mention the other use case because they probably work a little bit different like the dialog use case, as e.g. buttons aren't removed - they just finish they job. I had the following pattern in mind where no additional overlay is added https://github.com/vaadin/web-components/issues/842

Naming is probably the hardest part here - still awesome addition!

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -788,9 +788,9 @@ public Registration addPropertyChangeListener(String name,
      *
      * @since 1.3
      */
-    public DomListenerRegistration addPropertyChangeListener(
-            String propertyName, String domEventName,
-            PropertyChangeListener listener) {
+        public DomListenerRegistration addPropertyChangeListener(
+                String propertyName, String domEventName,
+                PropertyChangeListener listener) {
```
caalador, line 793: Faulty formatting change

flow-server/src/main/java/com/vaadin/flow/server/communication/ReturnChannelHandler.java
```diff
@@ -82,5 +82,4 @@ public String getRpcType() {
     private static Logger getLogger() {
         return LoggerFactory.getLogger(ReturnChannelHandler.class.getName());
     }
-
```
caalador, line 85: unneeded removal of line

pleku:
> Having this API inside the UI would be really good way to discover it! I wanted to mention the other use case because they probably work a little bit different like the dialog use case, as e.g. buttons aren't removed - they just finish they job. I had the following pattern in mind where no additional overlay is added vaadin/web-components#842

I think for this case it would be simply enough to make the UI inert and set nothing modal to it - but currently the API design in the `UI` doesn't directly allow this. This API can be added quite easily later on. It is also just possible to just add an invisible modal component to the UI 😄 and it will be the same effect. Or by getting the root layout (under UI) and setting it as modal via the setter in `UI`.

Like the current API's javadocs says - the developer/component needs to just take care of showing a clear indication to the user that something is happening and the UI won't work - so they don't try to do something while waiting.

flow-server/src/main/java/com/vaadin/flow/server/communication/ReturnChannelHandler.java
pleku: Done

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
pleku: Done

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/InertData.java
```diff
@@ -0,0 +1,205 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
```
joheriks, line 2:
```suggestion
 * Copyright 2000-2021 Vaadin Ltd.
```
I'm fine with 2022 as well 😝 But let's keep this consistent in so we don't miss it in the 2021 -> 2022 update

flow-server/src/test/java/com/vaadin/flow/internal/nodefeature/InertDataTest.java
```diff
@@ -0,0 +1,202 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
```
joheriks, line 2:
```suggestion
 * Copyright 2000-2021 Vaadin Ltd.
```
😛

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/ModalDialogView.java
```diff
@@ -0,0 +1,140 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
```
joheriks, line 2:
```suggestion
 * Copyright 2000-2021 Vaadin Ltd.
```

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/InertData.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import com.vaadin.flow.internal.StateNode;
+import com.vaadin.flow.internal.change.NodeChange;
+import com.vaadin.flow.shared.util.UniqueSerializable;
+
+/**
+ * Server side only feature about whether or not the node is inert or if it
+ * should ignore inhering inert state from parent. By default the node is not
```
joheriks, line 29:
```suggestion
 * should ignore inheriting inert state from parent. By default the node is not
```

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/InertData.java
```diff
@@ -0,0 +1,205 @@
...
+ * inert, and it will inherit the inert state from the parent. If the node lacks
+ * the inert data, then it will be just inheriting the state from parent.
+ * <p>
+ * The inert status is only updated when the changes are written to the client
+ * side because the inert state changes are applied for upcoming requests from
+ * the client side. Thus when an RPC call (like any DOM event) causes a node to
+ * become inert, the state is does not block any pending executions until
```
joheriks, line 36:
```suggestion
 * become inert, the inert state does not block any pending executions until
```

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/InertData.java
```diff
@@ -0,0 +1,205 @@
...
+     *            {@code false} for not
+     */
+    public void setInertSelf(boolean inertSelf) {
+        if (this.inertSelf != inertSelf) {
+            this.inertSelf = inertSelf;
+            markAsDirty();
+        }
+    }
+
+    public boolean isInertSelf() {
```
joheriks, line 105: Add JavaDoc

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/InertData.java
```diff
@@ -0,0 +1,205 @@
...
+        return inertSelf;
+    }
+
+    public boolean isIgnoreParentInert() {
```
joheriks, line 109: Add JavaDoc

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/InertData.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+    }
+
+    private void updateInertAndCascadeToChildren(Boolean resolvedParentInert) {
+        boolean newInert = resolveInert(resolvedParentInert);
+        if (cachedInert == null || cachedInert != newInert) {
+            // cascade update to all children unless those are ignoring parent
+            // value or have same value and thus don't need updating.
+            // (all explicitly updated nodes are visited separately)
+            LinkedList<StateNode> stack = new LinkedList<>();
```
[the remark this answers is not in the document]
joheriks: Would make sense to use `ArrayDeque`.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -1093,6 +1098,88 @@ public void setExtendedClientDetails(ExtendedClientDetails details) {
...
+     * inert after all modal components have been removed from it. Inert state
+     * is updated automatically when a component is removed from the UI, no need
+     * to call anything. Moving an inert component will remove inert status.
+     *
+     * @param child
+     *            the child component to toggle modal
+     */
+    public void toggleChildModal(Component child) {
+        if (modalComponentStack == null) {
+            modalComponentStack = new Stack<>();
```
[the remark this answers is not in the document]
joheriks: Would make sense to use `ArrayDeque`.

joheriks (changes requested):

flow-client/src/main/frontend/Flow.ts
```diff
@@ -214,6 +219,8 @@ export class Flow {
             this.container.style.display = '';
             resolve(this.container);
           }
+          // @ts-ignore
+          $wnd.console.log('flowNavigate: this.container.serverConnected');
           this.loadingFinished();
         };
```
joheriks, line 225: Remove debug statements from this method.

joheriks (dismissed):

joheriks (dismissed):

joheriks (approved):

PR #12440, opened 2021-11-24
feat: Bump minimum Java version to 11

Artur-:


manolo:
Agents in TC report : `Java version: 1.8.0_201, vendor: Oracle Corporation, runtime: /usr/lib/jvm/jdk1.8.0_201/jre`. 
What should we do here? 
- upgrade agents 
- modify builds so as they can detect flow version and set the appropriate javaHome

Artur-:
Stop using the TC build for PRs?

manolo:
Modified TC validations so as they run java11 in the case branch has these changes.

pom.xml
```diff
@@ -68,13 +68,13 @@
             https://maven.vaadin.com/vaadin-addons
         </flow.addons.repo.url>
 
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
```
manolo, line 72: What about producing bytecode for java 8 ? but still have java 11 as the source ?

Artur-: Why would we do that?

manolo: So as they can still deploy vaadin in their old systems.

Artur-: That would never work as you cannot use e.g. `String.strip` and any other Java API added after 8. It would be constantly broken.

manolo (approved):

Issue #12343, opened 2021-11-12
there are slow unit tests calling external tools


manolo:
There are unit tests in flow-server running external commands  like `npm install` hence they depend on those tools installed, and they take a long while to run: 
```
FrontendToolsTest
TaskRunPnpmInstallTest
TaskRunNpmInstallTest
```
They should either:
- be moved to a `flow-tests` submodule
- fixed by mocking system calls
- be retained to a test category and be run as ITs

caalador:
System calls can at least not be mocked as then we don't know if the features work with npm/pnpm which is quite integral in seeing that they install the expected packages

Issue #12412, opened 2021-11-22
categorise slow tests so as they are not run in regular validation.


manolo:
- IT Tests
There are some tests that penalise build times because of large usage of Thread.sleep().
`IdlePushChannelWebsocketIT.longWaitBetweenActions` and `IdlePushChannelLongPollingIT.longWaitBetweenActions` sleep for 7 minutes each, `PushLargeDataWebsocketIT` and `PushLargeDataLongPollingIT` last 1 minute each.
They are making that flow validation spend about 16 minutes of extra each time just sleeping the thread.

- Unit Tests
`HierarchyMapperWithNumerousDataTest` takes about 4 minutes on generating the 100,000 items of the data estructure needed for the test, apparently `TreeData.addItems` performs bad.

All of them should be marked in a `Slow` Category and run in separated build, probably snapshots.

See also https://github.com/vaadin/flow/issues/12343

caalador:
Closing this ticket as there already is a slow-tests execution so that they are excluded from the regular validation.

PR #12446, opened 2021-11-24
test: do not run slow tests in normal validations

manolo:
This PR is a refactor of pom files and category classes in order to share configuration between modules.
It also marks a bunch of slow tests and tests calling command line tools with the SlowTest annotation.

By default slow tests are skip in validation, it can still be run by using profiles: `-Pall-tests` or `-Pslow-tests` the second is thought for a separated build in CIs that will be run nightly.

Note that TC validation has been decreased by about 8-10 minutes https://teamcity.vaadin.com/buildConfiguration/Flow_FlowTests?mode=builds&state=successful#all-projects 

There is a TC validation for slow-tests that runs in parallel, it will be moved to the snapshot build

Fixes: #12343 #12412


flow-tests/pom.xml
```diff
@@ -403,38 +388,36 @@
             </properties>
         </profile>
         <profile>
-            <id>default</id>
+            <id>ie11</id>
```
ZheSun88, line 391: probably this profile can be removed totally, as IE11 is not part of support

ZheSun88: leave it for now, and remove it later

ZheSun88 (dismissed):

flow-test-generic/src/main/java/com/vaadin/flow/testcategory/PushTests.java
```diff
@@ -0,0 +1,27 @@
...
+ * the License.
+ */
+package com.vaadin.flow.testcategory;
+
+/**
+ * Tests which test push functionality should be annotated with
+ * {@code @Category(PushTests.class)}.
+ *
+ * @author Vaadin Ltd
+ * @since 1.0
```
ZheSun88, line 23: this since probably quite strange, but it is not something important i think.

ZheSun88 (approved):

ZheSun88 (commented):
do you want to do something with this test https://github.com/vaadin/flow/blob/330e87985c204dc55f0e42424079474ee9b7980d/flow-tests/test-root-ui-context/src/test/java/com/vaadin/flow/uitest/ui/push/IdlePushChannelIT.java#L11

PR #12474, opened 2021-11-29
fix(Fusion): support PathPattern parsed mapping for endpoint prefix

platosha:
Fixes compatibility with Spring Boot 2.6, which uses PathPattern matching.


platosha:
Should fix the build for https://github.com/vaadin/spring/pull/940

Artur- (approved):

PR #12487, opened 2021-11-30
Updated ASM8 to ASM9

mstahv:


flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
```diff
@@ -60,7 +60,7 @@
      *            The annotation class name to visit
      */
     FrontendAnnotatedClassVisitor(ClassFinder finder, String annotationName) {
-        super(Opcodes.ASM8);
```
knoobie, line 63: This (https://gitlab.ow2.org/asm/asm/-/issues/317854) could probably explain why the OpCodes has to be updated for newer java releases to work properly. Comment by Remi Forax: 

> I will close this issue as work as intended even if you may think it's surprising, currently ASM ClassReader support Java 11 (and even Java 12) so it recognizes Java 11 new attributes (here NestMembers). But here you are using a ClassVisitor which is configured to only recognize the ASM6 API, thus it fails at runtime with an UnsupportedOperation exception.

Artur-:
Can we merge this? Why is it a draft?

Artur- (approved):

mstahv:
I can't test that it actually works. I cannot reproduce the issue. Maybe going in blind is sometimes the option 😎

mstahv:
I can't test that it actually works. I cannot reproduce the issue. Maybe going in blind is sometimes the option 😎 @vaadin/flow Up to you to make the call.

caalador (approved):

mstahv:
@caalador Can you CP this for Vaadin 14 series as well?

mstahv:
@caalador For some reason the fix has not been picked to 9.0 (aka Vaadin 22) branch. https://github.com/vaadin/flow/blob/9.0/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java#L63

caalador:
Auto merge had not merged for some reason.

PR #12501, opened 2021-12-02
chore: replace exec-maven-plugin with gradle-maven-plugin

manolo:
This should fix problems when running the build in docker containers in windows.

### DIFF
--- a/flow-plugins/flow-gradle-plugin/pom.xml
+++ b/flow-plugins/flow-gradle-plugin/pom.xml
@@ -18,10 +18,6 @@
         central repo is done within platform final releases.
     </description>
 
-    <properties>
-        <gradle.executable>./gradlew</gradle.executable>
-    </properties>
-
     <dependencies>
         <dependency>
             <groupId>com.vaadin</groupId>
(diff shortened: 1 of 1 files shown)


flow-plugins/flow-gradle-plugin/pom.xml
```diff
@@ -55,34 +55,40 @@
...
+                        <arg>functionalTest</arg>
+                        <arg>-S</arg>
+                    </args>
+                </configuration>
                 <executions>
                     <execution>
-                        <id>gradle</id>
                         <phase>prepare-package</phase>
-                        <configuration>
-                            <executable>${gradle.executable}</executable>
```
ZheSun88, line 69: this property can be removed

ZheSun88:
this fix runs good on windows..(but it failed here with the flow validation/it-tests(6,3) )

image

.github/workflows/validation.yml
```diff
@@ -31,6 +31,8 @@ jobs:
             | sort > it.modules &
           P3=$!
           wait $P1 $P2 $P3
+      # - name: Sleep
+      #   run: echo Sleeping && sleep 300
```
ZheSun88, line 35: maybe remove this?

ZheSun88 (approved):

flow-plugins/flow-gradle-plugin/pom.xml
manolo: done

ZheSun88 (approved):

PR #12512, opened 2021-12-03
feat: Run postinstall only for selected npm packages

### LINKED ISSUE #12127: Run ”npm install” without running postinstall scripts
Supply chain attacks are becoming more common. To better protect everybody, we should not run postinstall scripts by default. If we have any dependencies that require postinstall (esbuild comes to mind) we should selectively, manually run postinstall for those. This way we at least limit a possible attack vector from thousands of packages to a handful

Artur-:
Fixes #12127


flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java
```diff
@@ -211,6 +211,13 @@
     @Parameter(property = "build.folder", defaultValue = "${project.build.directory}")
     private String projectBuildDir;
 
+    /**
+     * The npm packages to run postinstall for.
+     */
+    // Note: This must be kept in sync with the Gradle plugin default
+    @Parameter(property = "npm.postinstallPackages", defaultValue = "esbuild, @vaadin/vaadin-usage-statistics")
```
caalador, line 218: Shouldn't this be to be able to add post-install packages and the defaults be always executed and defined in BuildFrontendUtil. Then this would be in addition to as if you define this without the expected ones then it would fail the build.

Artur-: Then there would be no way to override the defaults, maybe it is not a problem

caalador: My thought here was that you add some packages and then everything fails as esbuild is not postinstalled because there was an expectation that the defaults are handled in any case.

Artur-: Done

caalador (approved):

caalador (approved):
Was there a documentation PR for this feature also so users know that they can request post-install for installed packages?

Artur-:
https://github.com/vaadin/docs/pull/1013

PR #12530, opened 2021-12-08
fix: output version string when version parsing fails

Artur-:
This should ease debugging of problems like "UnknownVersionException: Unable to detect version of pnpm"

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
@@ -975,20 +975,22 @@ public class FrontendUtils {
 
     protected static FrontendVersion getVersion(String tool,
             List<String> versionCommand) throws UnknownVersionException {
+        String output;
         try {
-            String output = executeCommand(versionCommand);
-            try {
-                return new FrontendVersion(parseVersionString(output));
-            } catch (IOException e) {
-                throw new UnknownVersionException(tool,
-                        "Expected a version number as output but got '" + output
-                                + "'" + " when using command "
-                                + String.join(" ", versionCommand),
-                        e);
-            }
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -977,8 +977,16 @@ protected static FrontendVersion getVersion(String tool,
...
+            try {
+                return new FrontendVersion(parseVersionString(output));
+            } catch (IOException e) {
+                throw new UnknownVersionException(tool,
+                        "Expected a version number as output but got '" + output
+                                + "'" + " when using command "
+                                + String.join(" ", versionCommand),
+                        e);
+            }
+        } catch (CommandExecutionException e) {
```
caalador, line 989: I would perhaps make it explicitly clearer by having them in different tries.
```
        String output;
        try {
            output = executeCommand(versionCommand);
        } catch (CommandExecutionException e) {
            throw new UnknownVersionException(tool, "Version command '"
                    + String.join(" ", versionCommand) + "' execution failed.",
                    e);
        }
        try {
            return new FrontendVersion(parseVersionString(output));
        } catch (IOException e) {
            throw new UnknownVersionException(tool,
                    "Expected a version number as output but got '" + output
                            + "'" + " when using command "
                            + String.join(" ", versionCommand),
                    e);
        }
```

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
Artur-: Done

caalador (approved):

PR #12533, opened 2021-12-08
chore: redistribute modules to avoid failures when running in the same folder

manolo:


Artur- (changes requested):
Are dependabot and issue template accidentally removed?

manolo:
@Artur-  you were very fast. That commit was pushed by accident, I had that disabled in my fork to avoid random PRs in my repo.

.github/workflows/validation.yml
```diff
@@ -5,15 +5,25 @@ jobs:
...
+          uname -a
+          id
+          pwd
+          type java && java -version
+          type mvn && mvn -version
+          type node && node --version
+          type npm && npm --version
       - name: Set flow version and Generate module lists
         run: |
+          # run 3 maven process at the same time and wait for them, saves half minute
```
Artur-, line 26: What does this comment refer to? Where does 3 come from?

.github/workflows/validation.yml
```diff
@@ -84,29 +97,26 @@ jobs:
...
         'test-root-context:0',
         'test-root-context:1',
         'test-root-context:2'
         ]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
Artur-, line 111: Why node 14 here and 16 above? Shouldn't it be 16 everywhere

Artur- (changes requested):

.github/workflows/validation.yml
manolo: There are three `mvn ... &` running in background in the block, each one has a line `PN=$!` at the end there is a `wait $P1 $P2 $P3`.
It means that when setting the version we can compute modules, and save some seconds

.github/workflows/validation.yml
manolo: yep, hard to see why frontend compilation failed, had to play with different versions, at the end, node version was not the cause

Artur- (approved):

PR #12537, opened 2021-12-08
chore: redistribute modules in CI actions (#12533 #12535) (CP: 9.0)

manolo:


.github/workflows/validation.yml
```diff
@@ -192,10 +203,18 @@ jobs:
...
       - name: Publish Unit Test Results
-        if: always()
         uses: EnricoMi/publish-unit-test-result-action/composite@v1
         with:
           files: "**/target/*-reports/TEST*.xml"
       - uses: geekyeggo/delete-artifact@v1
         with:
           name: saved-workspace
+      - name: The Build has Failed
+        if: ${{ failure() }}
```
ZheSun88, line 213: maybe simplify this place too? 
```
if: ${{ failure() || cancelled() }}
```

.github/workflows/validation.yml
```diff
@@ -84,29 +97,26 @@ jobs:
...
         'test-root-context:2'
         ]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Set up JDK 8
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+      - name: Set up JDK 11
```
ZheSun88, line 112: guess this name is wrong, as the java-version below is 8

.github/workflows/validation.yml
manolo: This is from the other PR, since 'm doing manually I can cherry pick it here

.github/workflows/validation.yml
manolo: yep

PR #12553, opened 2021-12-10
fix: outdated links to production mode docs

tarekoraby:


caalador (approved):

tarekoraby:
@caalador what does the above message by vaadin-bot mean? Does something have to be done?

caalador:
As we would want to have the link fixed also in 2.7 and 2.8 branches (so vaadin 14) the bot tried to do a cherry pick, but there was some merge issue so it needs to be picked manually.

I can do the pick for this.

PR #12561, opened 2021-12-13
Test: fix tesbench helper shadowroot

caalador:


flow-test-util/src/main/java/com/vaadin/flow/testutil/TestBenchHelpers.java
```diff
@@ -422,11 +426,25 @@ protected void waitForDevServer() {
...
+            // Based on https://github.com/SeleniumHQ/selenium/issues/10050#issuecomment-974231601
+            Map<String, Object> shadowRootMap = (Map<String, Object>) shadowObject;
+            String shadowRootKey = (String) shadowRootMap.keySet().toArray()[0];
+            return (WebElement)  shadowRootMap.get(shadowRootKey);
+        } else if(shadowObject instanceof  WebElement) {
+            return (WebElement) shadowObject;
+        } else if(shadowObject instanceof SearchContext) {
+            // ChromeDriver 96+
+            SearchContext shadowRoot = (SearchContext) shadowObject;
+            Assert.assertNotNull("Could not locate shadowRoot in the element",
```
Artur-, line 442: `shadowRoot` cannot be null here. `null instanceof SearchContext` is not true

caalador: Yes it was in the wrong position, but now that I look at it it will never fail as we will just have a TimeoutException from the waitUntil if it would be null.

Artur-:
Is this redundant now that #12573  is merged?

caalador:
Yes. Just forgot to close it yesterday.

PR #12562, opened 2021-12-13
Bump TypeScript to 4.5.3

Haprog:
A customer has been using `this.shadowRoot?.elementFromPoint` method which got broken in TS 4.4 (Vaadin 21) due to lib DOM type definition changes. In V22 with TS 4.4 it requires additional casting workaround to be able to call that method. This should be resolved by upgrading to TS 4.5 ([change in TS](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1123)). Related: https://github.com/WICG/webcomponents/issues/661#issuecomment-992262771

### DIFF
--- a/flow-client/package.json
+++ b/flow-client/package.json
@@ -44,7 +44,7 @@
     "prettier": "2.2.1",
     "sinon": "7.5.0",
     "sinon-chai": "3.6.0",
-    "typescript": "4.0.3",
+    "typescript": "4.5.3",
     "webpack": "4.46.0",
     "webpack-cli": "3.3.12",
     "workbox-core": "5.1.4",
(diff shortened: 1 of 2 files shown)


Artur-:
Are there incompatible changes in TS 4.5 that need to be taken into account somewhere?

Haprog:
@Artur- I'm not aware of any. Generally TS 4.5 seems to have less breaking changes compared to what 4.4 had. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#breaking-changes

If something related to those breaks, I assume it should be easy to fix.

Artur- (approved):

PR #12563, opened 2021-12-13
test: Split pnpm and npm tests into separate modules

Artur-:


manolo (approved):

manolo:
Failing in TC, it might be that order of the module matters in the build, wha happens if you move modules up to the place where the others were?

Artur-:
Now it should work..

manolo (dismissed):

manolo:
Rebased with master so as sonar action is run

manolo:
Rebased with master so as it runs sonar action

manolo (approved):

PR #12571, opened 2021-12-14
feat: Only send index.html when the browser requests a HTML page

### LINKED ISSUE #12569: Only serve index.html for unknown URLs if it is going to be used as an HTML document in the browser
### Description
The `Sec-Fetch-Dest` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Fetch-Dest ) tells what caused the browser to make a request to the server, i.e. what the result will be used for. If this header is set and it has a value of e.g. `script` we should never serve the index.html page but instead return 404. There are also a few other obvious ones listed in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Fetch-Dest that we should ignore, maybe:
* audio
* audioworklet
* font
* image
* manifest
* paintworklet
* script
* serviceworker
* sharedworker
* style
* track
* video
* worker
* xslt

Artur-:
Fixes #12569


caalador (commented):
Could we have a junit tests for "old browser", non html, html. and perhaps some for canHandleRequest in different situations.

Artur-:
Added tests

caalador (approved):

Issue #9986, opened 2021-02-08
use npm as the default client-side package manager


vlukashov:
When using Vaadin I expect that by default it makes a reasonable choice of the front-end package manager. Since Vaadin 14.2 and Vaadin 15 the default has been pnpm because in late 2019 it was determined to be significantly faster than npm for repeated builds (see #6150). In early 2021 the situation has changed: both npm and pnpm have improved build times and the speed benefit of using pnpm became negligible (see https://github.com/vaadin/flow/issues/9835).

Therefore, starting from Vaadin 20 (ETA June 2021) I expect Vaadin to use npm by default and to deprecate the pnpm support. Starting from Vaadin 21 (ETA September 2021) I expect pnpm support to be dropped completely.

DoD:
 - all new V20 projects generated by start.vaadin.com use npm
 - if the client-side package manager is not explicitly specified by the `vaadin.pnpm.enable` property, a new Vaadin 20 project uses npm to install client-side dependencies
 - if pnpm is explicitly selected as the client-side package manager (by the `vaadin.pnpm.enable` property) the project keeps using pnpm but prints a deprecation warning in the build log
 - a project upgraded from V19 to V20 'just works', i.e. Vaadin maven plugin automatically converts the project from pnpm to npm (i.e. removes `.npmrc` (if created [just for the `--shamefully-hoist` option](https://github.com/vaadin/flow/issues/9834)), pnpm files and the pnpm-specific `node_modules` structure)
 - there is a reliable way to lock versions of client-side Vaadin platform dependencies ([more context](https://docs.google.com/document/d/1UQ_JpxRSGBKeK2V48mJmM3B8ZViF6O_5dhwu6bRMguw))
 - the relevant docs are updated:
   - https://vaadin.com/docs-beta/latest/guide/installing/frontend/
   - https://vaadin.com/docs-beta/latest/flow/advanced/switch-npm-pnpm/
   - https://vaadin.com/docs-beta/latest/flow/advanced/all-vaadin-properties/

more context in the [feature design document](https://docs.google.com/document/d/1UQ_JpxRSGBKeK2V48mJmM3B8ZViF6O_5dhwu6bRMguw/edit#heading=h.2ckta28kw1pd)

haijian-vaadin:
**there is a reliable way to lock versions of client-side Vaadin platform dependencies** could be a separate investigation ticket. find tickets related to version pin tickets to make sure we don't introduce regressions (search by error messages like `dom-module`, `fs-event` etc).

[...]

PR #12572, opened 2021-12-14
feat!: use npm by default

haijian-vaadin:
## Description

Use npm as the default client-side package manager. To continue using `pnpm`, set the [`pnpm.enable` flag ](https://vaadin.com/docs/latest/flow/configuration/npm-pnpm/#switch-between-npm-and-pnpm)to `true`.

Fixes https://github.com/vaadin/flow/issues/9986


Artur-:
The constant `ENABLE_PNPM_DEFAULT` exists so it should be used in all those places you put `false`

caalador (commented):
Now a good test case for if it is plausible to use npm as the default is that can you get skeleton starter to work with https://vaadin.com/directory/component/pdf-browser as for some reason in this case we get polymer versions 3.2.0 and 3.4.1 which makes the project not start.

haijian-vaadin:
> Now a good test case for if it is plausible to use npm as the default is that can you get skeleton starter to work with https://vaadin.com/directory/component/pdf-browser as for some reason in this case we get polymer versions 3.2.0 and 3.4.1 which makes the project not start.

I was able to make it work with the following steps:
1. upgrade npm `npm install -g npm` (the overrides feature is in version 8.3.0)
2. in the `package.json` file, add an overrides section
```
"overrides": {
    "@polymer/polymer": "$@polymer/polymer"
  },
```
3. delete the `node_modules` folder and `package-lock.json` file

platosha (dismissed):

platosha (approved):

PR #12573, opened 2021-12-14
test: Remove unnecessary shadow root helpers

Artur-:

PR #12601, opened 2021-12-17
test: Remove unnecessary shadow root helpers (#12573)

ZheSun88:


caalador (approved):
2.7 fixes should always first go to 2.8 to be picked down, but I guess this should pick fine up also.

PR #12616, opened 2021-12-21
fix: Proper handling of href when the anchor is disabled/enabled

### LINKED ISSUE #10924: Anchor component improper implement of HasEnabled

mstahv:
## Description

Disabling and enabling component using the Java API now works as expected. href is always saved in the component implementation, but not stored in element if disabled.

Fixes #10924

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -28,6 +28,7 @@ import com.vaadin.flow.component.PropertyDescriptors;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.server.AbstractStreamResource;
 import com.vaadin.flow.server.StreamResource;
+import com.vaadin.flow.server.StreamResourceRegistry;
 
 /**
  * Component representing an <code>&lt;a&gt;</code> element.
@@ -184,9 +185,9 @@ public class Anchor extends HtmlContainer
         if (href instanceof String) {
             // let the method return the actual href string even if disabled
             return (String) href;
+        } else if (href instanceof AbstractStreamResource) {
+            return StreamResourceRegistry.getURI((AbstractStreamResource) href).toString();
         }
-        // this only happens if StreamResource is in use, in that case
-        // the disabled Anchor returns null, otherwise some String.
         return get(hrefDescriptor);
     }
 
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -129,8 +129,8 @@ public class Anchor extends HtmlContainer
     /**
      * Sets the URL that this anchor links to.
      * <p>
-     * A disabled Anchor removes the attribute from the HTML element, but it
-     * is stored (and reused when enabled again) in the server sided component.
+     * A disabled Anchor removes the attribute from the HTML element, but it is
+     * stored (and reused when enabled again) in the server sided component.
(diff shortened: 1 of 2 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -170,9 +185,36 @@ public void setHref(AbstractStreamResource href) {
...
+        super.onEnabledStateChanged(enabled);
+        if (enabled) {
+            assignHref();
+        } else {
+            getElement().removeAttribute("href");
+        }
+    }
+
+    private void assignHref() {
+        if (href != null) {
```
joheriks, line 208: Could this `if` expression also contain the guard `isEnabled()`, as it would simplify the callers.

joheriks (changes requested):

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
mstahv: Then I'd refactor the change as we'd be doing conditional checks twice in onEnabledStateChange.

joheriks (dismissed):

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -170,9 +181,36 @@ public void setHref(AbstractStreamResource href) {
      * @return the href value, or <code>""</code> if no href has been set
      */
     public String getHref() {
+        if (href instanceof String) {
+            // let the method return the actual href string even if disabled
+            return (String) href;
+        }
+        // this only happens if StreamResource is in use, in that case
+        // the disabled Anchor returns null, otherwise some String.
```
caalador, line 189: Just an idea. Would it break much if for a streamResource we would return the `AbstractStreamResource::getName`  as `StreamResource for 'AbstractStreamResource::getName'`

As then at least it would be clear that there is something set to the anchor.

mstahv: Even though it still wouldn't work perfectly, that might be a tiny enhancement. That could help if people are testing the existence of href when the component is disabled like this:

if(getHref().isEmpty()) {
   // do stuff with things
}

I guess we can't really get the actual uri for the resources as it (might be it) is not attached when ancsdhor is disabled 🤔

I guess the academically correct way would be to deprecate setHref(StreamResource) and have setStreamResource method separately. Then we could have a symmetric getter for it.

Up to you to decide. I'd get this in now as this and refine if we get more input from the users.

caalador: Well there is the static method `StreamResourceRegistry.getURI(AbstractStreamResource)` which if registered will give the actual href.

mstahv: Checked the implementation and to me it looks like it should always return something sane. Pushed a change like that.

mstahv: No test for it though 😬

mstahv: Now the tests cover this as well

caalador (approved):

PR #12617, opened 2021-12-21
test: fix memory leak in tests during build

Artur-:
Sets reuseDevServer to false for all modules to ensure that webpack is killed when jetty is stopped

### DIFF
--- a/flow-tests/pom.xml
+++ b/flow-tests/pom.xml
@@ -35,10 +35,12 @@
             com.vaadin.flow.testcategory.PushTests,
         </excludedGroups>
 
+        <jetty.scantrigger></jetty.scantrigger>
+
         <vaadin.devmode.liveReload.enabled>false</vaadin.devmode.liveReload.enabled>
         <vaadin.pnpm.enable>true</vaadin.pnpm.enable>
         <vaadin.allow.appshell.annotations>false</vaadin.allow.appshell.annotations>
-
+        <useDeprecatedV14Bootstrapping>false</useDeprecatedV14Bootstrapping>
         <vaadin.devmode.vite.options>--host</vaadin.devmode.vite.options>
 
         <!-- make sure we do not leave webpack-dev-server running after IT -->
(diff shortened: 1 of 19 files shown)


Artur-:
Memory usage in the build before this
image

after this
Image from iOS

caalador (dismissed):

Artur-:
Added back Jetty plugin to the csrf test module... maybe tests will now pass

caalador (dismissed):

Artur-:
This needs some more investigation because the fix doesn't actually fix the issue although the issue was fixed in the test branch. Probably something else still that is related to this

Artur-:
Now that system property handling is changed, it seems to work

flow-tests/pom.xml
```diff
@@ -256,11 +245,37 @@
...
+                                <value>${vaadin.reuseDevServer}</value>
+                            </systemProperty>
+                            <systemProperty>
+                                <name>vaadin.devmode.liveReload.enabled</name>
+                                <value>${vaadin.devmode.liveReload.enabled}</value>
+                            </systemProperty>
+                            <systemProperty>
+                                <name>vaadin.pnpm.enable</name>
+                                <value>${vaadin.pnpm.enable}</value>
+                            </systemProperty>
```
caalador, line 266: This should really be in the plugin configuration block and not as a system property.
``` 
                <configuration>
                    <pnpmEnable>${vaadin.pnpm.enable}</pnpmEnable>
                </configuration>
```

flow-tests/pom.xml
```diff
@@ -256,11 +245,37 @@
                         <stopKey>foo</stopKey>
                         <stopWait>5</stopWait>
                         <systemProperties>
+                            <!-- All system props are defined here as they need to be overridden 
+                                 in all modules so no non-standard value leaks to another module -->
+                            <force>true</force>
+                            <systemProperty>
+                                <name>vaadin.useDeprecatedV14Bootstrapping</name>
+                                <value>${vaadin.useDeprecatedV14Bootstrapping}</value>
+                            </systemProperty>
```
caalador, line 254: This should be in the maven plugin configuration block.
``` 
                <configuration>
                    <useDeprecatedV14Bootstrapping>${vaadin.useDeprecatedV14Bootstrapping}</useDeprecatedV14Bootstrapping>
                </configuration>
```

Artur-:
Can those improvmenets be in a separate PR as this blocks getting other PRs through?

Artur-:
Well, let's see if they pass without new issues...

flow-tests/pom.xml
Artur-: Done

flow-tests/pom.xml
Artur-: Done

caalador (approved):

PR #12635, opened 2021-12-23
feat: lock versions with npm overrides

haijian-vaadin:
## Description
How it works?
1. generate `versions.json` also for npm mode
2. if `npm` mode, create an `"overrides"` section in the `package.json` file if the file does not exist.
3. iterate through dependencies in the `version.json` file, add each dependency to the overrides section as @vaadin/xxx: $@vaadin/xxx if there is no @vaadin/xxx in the "overrides" section.

Requires npm 8.3.0 or newer which has the "overrids" feature. 
Fixes https://github.com/vaadin/fusion/issues/164

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -124,6 +125,39 @@ public void execute() {
         }
     }
 
+    private boolean lockVersionForNpm(JsonObject packageJson) throws IOException {
+        if(enablePnpm) {
```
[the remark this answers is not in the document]
Artur-: Ah right, this is used for `npm` only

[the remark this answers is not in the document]
Artur-: Should we write this info to `package.json` also for pnpm and read the overrides from the same place instead of having two different approaches?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -124,6 +125,39 @@ public void execute() {
...
+        
+        File generatedVersionsFile = new File(npmFolder, versionsPath);
+        final JsonObject versionsJson = Json.parse(FileUtils.readFileToString(
+                generatedVersionsFile, StandardCharsets.UTF_8));
+        
+        if (versionsJson != null) {
+            JsonObject overridesSection = getOverridesSection(packageJson);
+            for(String dependency: versionsJson.keys()) {
+                if(!overridesSection.hasKey(dependency)) {
+                    overridesSection.put(dependency, "$"+dependency);
```
Artur-, line 144: How can I manually override a version? This looks like it would overwrite all my manual changes?

haijian-vaadin: if you define a version in the dependencies section, e.g. `"@vaadin/button": "23.0.0-alpha3"`, then it will write into the overrides section `"@vaadin/button": "$@vaadin/button"`, so that the user-defined version is used. 

if you define sth in the overrides section, e.g. `"@vaadin/button": "23.0.0-alpha3"`, then NodeUpdater will leave it untouched.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -567,4 +568,59 @@ String writePackageFile(JsonObject json, File packageFile)
     Logger log() {
         return LoggerFactory.getLogger(this.getClass());
     }
+
+    /**
+     * Generate versions json file for pnpm.
```
Artur-, line 573: Isn't this also for npm now?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -101,7 +101,8 @@ public void execute() {
                     .getPackages();
             JsonObject packageJson = getPackageJson();
             modified = updatePackageJsonDependencies(packageJson,
-                    scannedApplicationDependencies);
+                    scannedApplicationDependencies)
+                    | lockVersionForNpm(packageJson);
```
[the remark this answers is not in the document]
Artur-: This really needs to be fixed or somebody will change it to `||` one day without realizing it breaks the code

Store the results into two separate booleans or something similar

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -310,7 +248,7 @@ private void runNpmInstall() throws ExecutionFailedException {
         Logger logger = packageUpdater.log();
         if (enablePnpm) {
             try {
-                createPnpmFile(generateVersionsJson());
+                createPnpmFile(packageUpdater.generateVersionsJson());
```
Artur-, line 251: We should find one place to generate the version json. It just makes everything harder if `pnpm` runs it as part of install and `npm` runs it in `TaskUpdatePackages`

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunNpmInstallTest.java
```diff
@@ -98,6 +98,17 @@ Logger log() {
                 return logger;
             }
 
+            @Override
```
Artur-, line 101: Where is the test for this whole feature? I see no test testing that an `overrides` section is even written

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
haijian-vaadin: good catch, will update the doc

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
haijian-vaadin: yes, will try to refactor

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunNpmInstallTest.java
haijian-vaadin: not yet, it's a draft PR, working in progress, doing it now

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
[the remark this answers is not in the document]
TomMD: Do I understand correctly that your preference would be to remove this warning message because it is inaccurate and could be harmful?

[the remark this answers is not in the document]
haijian-vaadin: Yes, the code has been refactored already, the warning is outdated.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunNpmInstallTest.java
haijian-vaadin: Added some tests, can you help to take another look?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
haijian-vaadin: does it sound good or do you have other concerns?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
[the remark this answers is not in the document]
haijian-vaadin: It works that both `npm` and `pnpm` read the pinned dependencies from the generated versions.json file. The difference is that `npm` locks the version with an `overrides` section in the `package.json` file, `pnpm` locks the version by creating a hook, the `pnpmfile.js`.
`npm` overrides uses the "$" syntax for the dependency version, where `pnpm` needs the explicit version, so it won't work.
Aslo the pnpm hook now works with the versions.json file directly `const versionsFile = require('path').resolve(__dirname, 'target/frontend/versions.json');`. To change it to read from `package.json` would actually make it more complicated.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
haijian-vaadin: I am now thinking maybe we do the refactor in a separate PR? So that we can get this PR merged and make a release before the PiT.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -117,13 +118,52 @@ public void execute() {
                     // https://github.com/pnpm/pnpm/issues/2587
                     // https://github.com/vaadin/flow/issues/9719
                     deletePnpmLockFile();
+                } else if (npmVersionLockingUpdated) {
+                    deleteNpmLockFile();
```
Artur-, line 122: When is this a problem? We really shouldn't be deleting the lock file, ever.

haijian-vaadin: It's needed for the existing project with npm mode. `npm` version locking only works after deleting the npm lock file.

haijian-vaadin: hmm, take another thought, seems it's only a problem with an existing project that has duplicated dependencies already. I will update the code.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
Artur-: Wouldn't it be better to make the final version and test that ?

haijian-vaadin: done, it took less effort than expected. Now it generates the `version.json` file in `TaskUpdatePackages`, the file path is passed as a field, same as `modified`.

Artur-:
I think we need to update the `hash` calculation to include the `overrides` section. Right now this does not work:
1. Run `mvn spring-boot:run`
2. Edit package.json and add an override like `"json-schema": ">=0.4.0"`
3. Run `mvn spring-boot:run`

If the original npm installation used 0.3.0 like in my app, it will still use 0.3.0. This is only because 
```
Skipping `npm install` because the frontend packages are already installed in the folder '.../node_modules' and the hash in the file '.../node_modules/.vaadin/vaadin.json' is the same as in 'package.json'
```

Artur-:
Then a small nit: I would place the `overrides` section before the `vaadin` section so that package.json first contains the "standard" things and then vaadin specifics at the end

haijian-vaadin:
> I think we need to update the `hash` calculation to include the `overrides` section. Right now this does not work:
> 
> 1. Run `mvn spring-boot:run`
> 2. Edit package.json and add an override like `"json-schema": ">=0.4.0"`
> 3. Run `mvn spring-boot:run`
> 
> If the original npm installation used 0.3.0 like in my app, it will still use 0.3.0. This is only because
> 
> ```
> Skipping `npm install` because the frontend packages are already installed in the folder '.../node_modules' and the hash in the file '.../node_modules/.vaadin/vaadin.json' is the same as in 'package.json'
> ```

I think this could done in a separate PR, made a [ticket](https://github.com/vaadin/flow/issues/12682). The main purpose of this PR is to lock the versions of Vaadin managed dependencies.

joheriks:
> Requires npm 8.3.0 or newer which has the "overrids" feature.
Should the version validation be updated then?
https://github.com/vaadin/flow/blob/5e19cb3bf2abd82e5ae5777403872c9f9509b545/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java#L113-L114

haijian-vaadin:
> > Requires npm 8.3.0 or newer which has the "overrids" feature.
> 
> Should the version validation be updated then?
> 
> https://github.com/vaadin/flow/blob/5e19cb3bf2abd82e5ae5777403872c9f9509b545/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java#L113-L114

@Artur- did a [change](https://github.com/vaadin/flow/pull/12660/files) to upgrade to Node 17 + npm 8.3. Is that enough?
I think npm 6.14 is still supported, as in V14, just that it doesn't have the overrides feature available.

Artur-:
It would be nice to let the users know that it will work better if you upgrade to npm 8.3+ if they are using an older globally installed node/npm. We cannot really require the latest version of npm though and we need to make sure that node 16 with its bundled npm works in a standard (i.e no addons) project

joheriks:
So if I have an older `npm` version that doesn't support `overrides` the section will be ignored, and there will be no version locking (as it is currently working when using `npm` instead of `pnpm`). Is there a risk of a silent regression if I have been relying on `pnpm` now that the default switched to `npm` but my pre-installed version is too old?

haijian-vaadin:
Is it enough to mention this in the upgrade guide? For people relying on the pnpm locking and cannot upgrade to npm 8.3, they can keep using pnpm.

Artur-:
The place you would run into this is most of the time when you get a browser warning that a custom element has already been defined. This would be the best place to provide additional info on how the solve the problem but it feels like a separate issue/pr

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -119,11 +122,47 @@ public void execute() {
...
         } catch (IOException e) {
             throw new UncheckedIOException(e);
         }
     }
 
+    boolean lockVersionForNpm(JsonObject packageJson, String versionsPath)
+            throws IOException {
+        if (enablePnpm) {
+            return false;
+        }
```
caalador, line 135: I feel we should not return for pnpm, but just wrap the end result in `pnpm` JsonObject and clear out the code for pnpmfile.js

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -119,11 +122,47 @@ public void execute() {
...
+                generatedVersionsFile, StandardCharsets.UTF_8));
+
+        if (versionsJson != null) {
+            JsonObject overridesSection = getOverridesSection(packageJson);
+            final JsonObject dependencies = packageJson.getObject(DEPENDENCIES);
+            for (String dependency : versionsJson.keys()) {
+                if (!overridesSection.hasKey(dependency)
+                        && dependencies.hasKey(dependency)
+                        && !NodeUpdater.DEP_NAME_FLOW_JARS.equals(dependency)) {
+                    overridesSection.put(dependency, "$" + dependency);
```
caalador, line 149: If `pnpm { override { } }` doesn't work with `$[dependencyKey]` then we should just write the version here.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunNpmInstallTest.java
```diff
@@ -447,6 +447,47 @@ protected NodeUpdater getNodeUpdater() {
...
+                        FileUtils.write(new File(npmFolder, "versions.json"),
+                                versionsContent, StandardCharsets.UTF_8);
+                    }
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+                return "./versions.json";
+            }
+        };
+    }
```
caalador, line 489: Shouldn't this be in `TaskRunPnpmInstallTest` as it is only used from there?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -119,11 +122,47 @@ public void execute() {
...
+        final JsonObject versionsJson = Json.parse(FileUtils.readFileToString(
+                generatedVersionsFile, StandardCharsets.UTF_8));
+
+        if (versionsJson != null) {
+            JsonObject overridesSection = getOverridesSection(packageJson);
+            final JsonObject dependencies = packageJson.getObject(DEPENDENCIES);
+            for (String dependency : versionsJson.keys()) {
+                if (!overridesSection.hasKey(dependency)
+                        && dependencies.hasKey(dependency)
+                        && !NodeUpdater.DEP_NAME_FLOW_JARS.equals(dependency)) {
```
caalador, line 148: Shouldn't the skip, skip any dependency that targets a directory e.g. anything that has the value as `startsWith("./" + buildFolder())` and not only `@vaadin/flow-frontend` also what about `DEP_NAME_FLOW_DEPS = "@vaadin/flow-deps"`?

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdatePackagesNpmVersionLockingTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+        String versionsPath = packageUpdater.generateVersionsJson();
+        packageUpdater.lockVersionForNpm(packageJson, versionsPath);
+
+        Assert.assertNull(
+                packageJson.getObject(OVERRIDES).get(TEST_DEPENDENCY));
+    }
+
+    @Test
+    public void shouldNotNotUpdatesOverrides_whenHasUserModification()
```
caalador, line 130:
```suggestion
    public void shouldNotUpdateOverrides_whenHasUserModification()
```
What is the expected place to have user modificatio? Override or dependencies? should override be populated for user dependency that is overriding framework version?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
Artur-: Wouldn't that mean duplicating every single version in one more place? Also that if you manually update a version of a dependency, then you would still get an old version unless you remember to update the overrides also?

caalador: It should ofc be automatically updated.

caalador:
Also missing is cleaning of the override in `CleanFrontendMojo`

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
joheriks: I guess it depends on whether we consider a valid use case to manually update Vaadin managed dependencies. If so, it would indeed be surprising to have to update the `overrides`section as well. If not, and it is all automated it shouldn't matter. At least `pnpm` (`5.18.10`)  doesn't seem to fail on the `overrides: { @dep: "$@dep" }` syntax. Since we have already version pinning in `pnpmfile.js`, would it anyway be better to keep this (then `pnpmfile.js` could just later be removed if [pnpm adds the compatible option](https://github.com/pnpm/pnpm/issues/4190)).

caalador: It should be 100% ok to update a vaadin dependency if there is an incompatibility with for instance an add-on that would else work or if you want to use something newer in the project (for instance charts version)

joheriks: I would be inclined to keep `@dep: "$@dep"`, adding duplicate versions seems like setting up for more trouble ahead.

caalador: Well my point was just that if it doesn't work with pnpm (is changed so that both generate the overrides) then it should lock it with the version. else it should reference.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
haijian-vaadin: I think it's better to leave pnpm lock alone in this PR, so that if npm locking for some reason doesn't work, they can switch back to pnpm.
We could do the pnpm locking change in a different PR.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunNpmInstallTest.java
haijian-vaadin: good point, move it there

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdatePackagesNpmVersionLockingTest.java
haijian-vaadin: the overrides section might have user modifications. A user could update the values in the overrides section manually, if there are such changes, we just leave them alone.

haijian-vaadin:
> Also missing is cleaning of the override in `CleanFrontendMojo`

Good catch, done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
haijian-vaadin: good point, updated

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdatePackagesNpmVersionLockingTest.java
joheriks: If the update has pinned a Vaadin managed dep (e.g. updated in the `overides` section `"@polymer/iron-a11y-announcer": "$@polymer/iron-a11y-announcer"` -> `"@polymer/iron-a11y-announcer": "3.1.0"`), this is also cleaned. But maybe this is fine.

haijian-vaadin: right, it's an edge case and should be fine.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
```diff
@@ -165,6 +168,7 @@ public void runPnpmInstall_versionsJsonIsFound_pnpmHookFileIsGenerated()
 
         TaskRunNpmInstall task = createTask();
         getNodeUpdater().modified = true;
+        getNodeUpdater().versionsPath = "./version.json";
```
platosha, line 171: Let’s align the filename:

```suggestion
        getNodeUpdater().versionsPath = "./versions.json";
```

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
```diff
@@ -180,6 +184,7 @@ public void runPnpmInstall_versionsJsonIsNotFound_pnpmHookFileIsGeneratedFromPac
             throws IOException, ExecutionFailedException {
         TaskRunNpmInstall task = createTask();
         getNodeUpdater().modified = true;
+        getNodeUpdater().versionsPath = "./version.json";
```
platosha, line 187:
```suggestion
        getNodeUpdater().versionsPath = "./versions.json";
```

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
platosha, line 171: Maybe align the filename?

```suggestion
        getNodeUpdater().versionsPath = "./versions.json";
```

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunPnpmInstallTest.java
```diff
@@ -719,17 +645,53 @@ private JsonObject getGeneratedVersionsContent(File versions)
...
 
     }
 
+    private NodeUpdater createAndRunNodeUpdater(String versionsContent) {
+        NodeUpdater nodeUpdater = createNodeUpdater(versionsContent);
+        try {
+            nodeUpdater.execute();
+        } catch (Exception e) {
+            throw new IllegalStateException(
+                    "NodeUpdater failed to genereate the version.json file");
```
platosha, line 662:
```suggestion
                    "NodeUpdater failed to genereate the versions.json file");
```

platosha (approved):

joheriks (dismissed):

Artur- (approved):

imsys:
> now that it will work better if you upgrade to npm 8.3+ if they are using an older globally installed node/npm. We cannot really require the latest version of npm though and we need to make sure that node 16 with its bundled npm works in a standard (i.e no addons) project

As someone who fell in this thread almost by accident, when I saw the issue at https://github.com/pnpm/pnpm/issues/4190, and I don't know much about this project, but I thought to give my two cents, that you could add some version specification in the package.json

```json
"engines": {
    "npm": ">=8.3.0"
    "pnpm": ">=5.10.1"
}
```
> "Unless the user has set the `engine-strict` config flag (see `.npmrc`), this field is advisory only and will only produce warnings when your package is installed as a dependency." - from npmjs documentation

haijian-vaadin:
Hi @imsys, thanks a lot for the proposal, that's a very good idea, I made an enhancement [ticket](https://github.com/vaadin/flow/issues/12824) out of your suggestion.

PR #12669, opened 2022-01-03
Webpack 5 support behind feature flag

Artur-:


Artur-:
Something related to theming is broken here

PR #12670, opened 2022-01-04
chore!: remove webpackTemplate and webpackGeneratedTemplate options

Artur-:
These options are not really useful for anything and just makes the code more complex

### DIFF
--- a/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
+++ b/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
@@ -164,16 +164,4 @@ public class BuildFrontendMojo extends FlowModeAbstractMojo
         return runNpmInstall;
     }
 
-    @Override
-    public String webpackGeneratedTemplate() {
-
-        return webpackGeneratedTemplate;
-    }
-
-    @Override
-    public String webpackTemplate() {
-
-        return webpackTemplate;
-    }
-
 }
--- a/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
+++ b/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
@@ -92,22 +92,6 @@ public class BuildFrontendMojo extends FlowModeAbstractMojo
     @Parameter(defaultValue = "true")
     private boolean optimizeBundle;
 
-    /**
-     * Copy the `webpack.config.js` from the specified URL if missing. Default
-     * is the template provided by this plugin. Set it to empty string to
-     * disable the feature.
-     */
-    @Parameter(defaultValue = FrontendUtils.WEBPACK_CONFIG)
-    private String webpackTemplate;
-
-    /**
-     * Copy the `webpack.generated.js` from the specified URL. Default is the
-     * template provided by this plugin. Set it to empty string to disable the
-     * feature.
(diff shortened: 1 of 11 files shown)


flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
```diff
@@ -164,16 +164,4 @@ public boolean runNpmInstall() {
         return runNpmInstall;
     }
 
-    @Override
-    public String webpackGeneratedTemplate() {
-
-        return webpackGeneratedTemplate;
```
caalador, line 170: Also remove the `webpackGeneratedTemplate` and `webpackTemplate` variables +javadoc on lines 95-109

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -707,8 +694,7 @@ private NodeTasks(Builder builder) {
                 builder.classFinder);
         FrontendDependenciesScanner frontendDependencies = null;
 
-        boolean enableWebpackConfigUpdate = builder.webpackTemplate != null
-                && !builder.webpackTemplate.isEmpty();
+        boolean enableWebpackConfigUpdate = builder.webappResourcesDirectory != null;
```
caalador, line 697: Do we need this at all. Instead the `else if (enableWebpackConfigUpdate) {` on line 799 (new code) should be changed to `} else if (frontendDependencies != null) {`

caalador (changes requested):

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
Artur-: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
Artur-: Is there a case where `builder.enablePackagesUpdate || builder.enableImportsUpdate` would be false but `enableWebpackConfigUpdate` would be true and thus `frontendDependencies` would remain `null`?

caalador: When running prepare-frontend they are both false and at that time I think also enableWebpackConfigUpdate is also false which would lead to frontendDependencies to be null. So this way the behaviour would stay the same as it is now.

Artur-: Changed

caalador (dismissed):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
caalador: Seems that we have tests that run without webpack. Perhaps we should just have in the `Builder::withWebpack` a boolean that should webpack be used or not instead of using given values. If there is a request for withWebpack then webpack should be used.

Artur-: Let's try like this then

caalador: This is better.

caalador:
Maven plugin test needs to have the setting of the 2 removed fields removed.

caalador (approved):

PR #12674, opened 2022-01-04
fix: remove <custom-style> and use registerStyles

### LINKED ISSUE #12106: @CssImport(themeFor="whatever") is broken in v22
This code https://github.com/vaadin/flow/blob/9fa4ed549eae0fac269d593b261a1c65fbe1ac6b/flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java#L74-L82 uses `<custom-style>` and is probably broken with the latest changes. The code does not include the legacy style adapter but still relies on those features. It is strange that no test is failing.

The code should be rewritten to use `registerStyles`

joheriks:
Do not wrap CSS added directly to head in `<custom-style>`.
For targeted CssImports, use `registerStyles` from `vaadin-themable-mixin`
instead of constructing a `dom-module` element.

Fix #12106.


flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
```diff
@@ -480,17 +478,34 @@ private boolean addCssLines(Collection<String> lines, CssData cssData,
...
+
+        Map<String, String> optionalsMap = new LinkedHashMap<>();
+        if (cssData.getInclude() != null) {
+            optionalsMap.put("include", cssData.getInclude());
+        }
+        if (cssData.getId() != null) {
+            optionalsMap.put("moduleId", cssData.getId());
+        } else if (cssData.getThemefor() != null) {
+            optionalsMap.put("moduleId", getThemeIdPrefix());
+        }
```
caalador, line 490: As this ordering is not defined anywhere should we perhaps throw an exception when id and themeFor are both defined in the same annotation?

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
```diff
@@ -480,17 +478,34 @@ private boolean addCssLines(Collection<String> lines, CssData cssData,
...
+        String optionals = optionalsMap.isEmpty() ? ""
+                : ", {" + optionalsMap.keySet().stream()
+                        .map(k -> k + ": '" + optionalsMap.get(k) + "'")
+                        .collect(Collectors.joining(", ")) + "}";
```
caalador, line 494: This would read better as
```
String optionals = "";
if(!optionalsMap.isEmpty()){
        String mapped = optionalsMap.keySet().stream()
                .map(k -> k + ": '" + optionalsMap.get(k) + "'")
                .collect(Collectors.joining(", "));
        optionals = ", {" + mapped + "}";
}
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
joheriks: Done.

caalador (approved):

PR #12695, opened 2022-01-11
fix: Always use defined node version

### LINKED ISSUE #12686: Running devMode doesn't respect nodeVersion given in plugin configuration
### Description of the bug

Starting up a project with `requireHomeNodeExec` and `nodeVersion` defined in for the plugin does install and use a node to `~/.vaadin`, but does not use the defined nodeVersion instead using the `DEFAULT_NODE_VERSION`

### Expected behavior

Defined version should be installed and not the default version

- Define to any v14 demo to the maven plugin the configuration:
```
<configuration>
  <requireHomeNodeExec>true</requireHomeNodeExec>
  <nodeVersion>v17.2.0</nodeVersion>
</configuration>
```
- Run `mvn`
- See that the installed version is not 17.2.0, but the default 14.15.4

### Versions

- Vaadin / Flow version: 14+

caalador:
If defined always use the node version
and download root url is defined.

Closes #12686

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -30,6 +30,7 @@ import java.util.Set;
 import java.util.concurrent.TimeoutException;
 
 import org.apache.commons.io.FileUtils;
+import org.slf4j.LoggerFactory;
 import org.zeroturnaround.exec.InvalidExitValueException;
 import org.zeroturnaround.exec.ProcessExecutor;
 
@@ -196,8 +197,7 @@ public class BuildFrontendUtil {
      *            - the PluginAdapterBase.
      * @return the Token {@link File}.
      */
-    public static File propagateBuildInfo(PluginAdapterBase adapter)
-            throws URISyntaxException {
+    public static File propagateBuildInfo(PluginAdapterBase adapter) {
 
         // For forked processes not accessing to System.properties we leave a
         // token file with the information about the build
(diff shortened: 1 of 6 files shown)


flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
```diff
@@ -208,6 +211,14 @@ public static File propagateBuildInfo(PluginAdapterBase adapter) {
         buildInfo.put(SERVLET_PARAMETER_INITIAL_UIDL,
                 adapter.eagerServerLoad());
         buildInfo.put(NPM_TOKEN, adapter.npmFolder().getAbsolutePath());
+        buildInfo.put(NODE_VERSION, adapter.nodeVersion());
+        try {
+            buildInfo.put(NODE_DOWNLOAD_ROOT,
+                    adapter.nodeDownloadRoot().toString());
+        } catch (URISyntaxException e) {
+            LoggerFactory.getLogger("BuildInfo")
+                    .error("NODE_DOWNLOAD_ROOT defined incorrectly", e);
```
joheriks, line 220: Should the logged error contain the literal value of `NODE_DOWNLOAD_ROOT` (`node.download.root`) instead

joheriks (changes requested):

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
caalador: clarified.

joheriks (approved):

joheriks (approved):

PR #12718, opened 2022-01-13
Fix statistic test failure

haijian-vaadin:
`UsageStatisticsExporterTest` is failing when running `mvn verify` because it assumes nonempty entries, but some other tests clear the entries.

### DIFF
--- a/flow-server/src/test/java/com/vaadin/experimental/FeatureFlagsTest.java
+++ b/flow-server/src/test/java/com/vaadin/experimental/FeatureFlagsTest.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.Collection;
+import java.util.List;
 import java.util.stream.Collectors;
 
 import com.vaadin.flow.internal.UsageStatistics;
@@ -154,13 +155,17 @@ public class FeatureFlagsTest {
     @Test
     public void disabledFeatureFlagsNotMarkedInStatsWhenLoading()
             throws IOException {
-        Collection<UsageEntry> originalEntries = UsageStatistics.getEntries().collect(Collectors.toList());
+        Collection<UsageEntry> originalEntries = getUsageStatisticsEntries();
         UsageStatistics.clearEntries();
         createFeatureFlagsFile("");
         featureFlags.loadProperties();
         Assert.assertFalse(
                 hasUsageStatsEntry("flow/featureflags/exampleFeatureFlag"));
         restoreUsageStatistics(originalEntries);
+    }
+
+    private List<UsageEntry> getUsageStatisticsEntries() {
+        return UsageStatistics.getEntries().collect(Collectors.toList());
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


flow-server/src/test/java/com/vaadin/experimental/FeatureFlagsTest.java
```diff
@@ -151,11 +155,17 @@ public void setEnabledOnlyInDevelopmentMode() throws IOException {
     @Test
     public void disabledFeatureFlagsNotMarkedInStatsWhenLoading()
             throws IOException {
+        Collection<UsageEntry> originalEntries = getUsageStatisticsEntries();
```
Artur-, line 158: If you want to take this approach then `restore`  should be in a `finally` block.

A less fragile version would be that a test ensures in its setup phase that its preconditions are met. In this case for instance you shouldn't need to touch the feature flag tests or any other but instead do `clear` + `markAsUsed` for a dummy value in the test that is broken, i.e. the stats test

Artur- (changes requested):

caalador:
I see a problem in that if the tests are run concurrently then they are clearing and restoring each other. Shouldn't the most correct fix be to have clearEntries add the java version mark as that's the default state of the statistics e.g.
```
    public static void clearEntries() {
        entries.clear();
        String version = System.getProperty("java.version");

        // Ignore pre, build and opt fields
        version = version.replaceAll("[-_+].*", "");

        markAsUsed("java", version);
    }
```
And any test running statistics with clear should probably be marked not thread safe

haijian-vaadin:
updated the code based on @caalador's suggestion, renamed clearEntries() to resetEntries() better reflect the behavior.

flow-server/src/test/java/com/vaadin/experimental/FeatureFlagsTest.java
caalador: The restore shouldn't be needed anymore with the clean change.

manolo:
StatisticsSenderTest is also failing in actions

flow-server/src/test/java/com/vaadin/experimental/FeatureFlagsTest.java
haijian-vaadin: oops, now fixed.

haijian-vaadin:
> StatisticsSenderTest

Do I need to fix sth there? Or would this PR fix that automatically?

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/internal/UsageStatistics.java
```diff
@@ -120,9 +115,19 @@ public static void removeEntry(String name) {
     }
 
     /**
-     * Clear the usage entries.
+     * Reset the usage entries.
      */
-    public static void clearEntries() {
+    public static void resetEntries() {
```
Artur-, line 120: This really shouldn't be public API as it is only needed for tests. 

But this PR doesn't make the current situation any worse as it already is public API

flow-server/src/main/java/com/vaadin/flow/internal/UsageStatistics.java
```diff
@@ -122,7 +117,17 @@ public static void removeEntry(String name) {
...
         entries.clear();
+        setupDefaultEntries();
+    }
+
+    private static void setupDefaultEntries() {
+        String version = System.getProperty("java.version");
+
+        // Ignore pre, build and opt fields
+        version = version.replaceAll("[-_+].*", "");
```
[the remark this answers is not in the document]
Artur-: This should be fixed so if `java.version` for some reason is null it is set to e.g. `unknown` instead of failing with an NPE. As can be seen it was already broken so not blocking...

Artur- (approved):

PR #12755, opened 2022-01-18
chore: add controller package name to spring.factories

haijian-vaadin:
Related to https://github.com/vaadin/flow/pull/12741, the `FusionController` package name was unintentionally removed by the IDE during refactoring.


vursen:
Shouldn't the PR be marked as a `fix:`? My Fusion application doesn't want to start without this fix.

haijian-vaadin:
> Shouldn't this change be marked as `fix:`? My Fusion application doesn't want to start without this fix.

As explained in the description, this should've been done inside #12741, just an error occurred during the IDE refactoring. 
It would be part of the release note if marked as a bug fix, but this fix alone doesn't make much to the user.

Lodin (approved):

PR #12766, opened 2022-01-19
feat: Fallback to alternative node when preferred is unsupported

### LINKED ISSUE #12494: Always fall back to `~/.vaadin/node` when preferred Node is unsupported
Flow can use Node from three different locations:

1. In project directory (installed with [frontend-maven-plugin](https://github.com/eirslett/frontend-maven-plugin))
2. Global installation (`node` on `PATH`)
3. Auto-downloaded and expanded into `~/.vaadin/node`

Locations 1. and 2. take priority over 3., even when the Node version is deemed incompatible with Vaadin (currently, older than `8.9` or `17.0.x`). In that case an `IllegalStateException` with a message stating that "Your Node is too old" is thrown. A friendlier approach would be to use `~/.vaadin/node` as a fallback in that case, since this version is managed by Vaadin to always be a compatible version.

The supported version lower range is currently `12.22` (`FrontendTools::SUPPORTED_NODE_VERSION`) and the automatically installed version is `16.13` (`FrontendTools::DEFAULT_NODE_VERSION`). With automatic fallback in place, there is no need to accept older versions (`FrontendTools::SHOULD_WORK_NODE_VERSION`), so this check and constant could be removed.

### Acceptance criteria
-  For project and global Node, the version is checked and if outside the supported range(s), `~/.vaadin/node` is populated with the same version of Node.JS as had been downloaded currently if no Node was pre-installed.
- A message is logged to clearly display the version and location of the used Node / NPM
- Messages admonishing the user about their Node being too old and upgrade nudges are removed:
  - https://github.com/vaadin/flow/blob/f6bf82215393e444cb93e0f12c35e68529d4daa0/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java#L73-L80
  - https://github.com/vaadin/flow/blob/f6bf82215393e444cb93e0f12c35e68529d4daa0/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java#L93-L101
- If downloading fails, failure message should suggest downloading Node or configuring the proxy settings (as this can be a reason for downloading failing).
- Remove checks against `SHOULD_WORK_NODE_VERSION`. If the preferred node is older than `SUPPORTED_NODE_VERSION`, always download the supported version.

mcollovati:
## Description

When project or global node version is detected but deemed incompatible with Vaadin, alternative node installation (~./vaadin/node) is used as fallback (and downloaded if not present).

Fixes #12494 

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -520,7 +520,7 @@ public class FrontendTools {
      *
      * Returns the input executable if version is supported, otherwise
      * {@literal null}.
-     * 
+     *
      * @param nodeExecutable
      *            node executable to be checked
      * @return input node executable if supported, otherwise {@literal null}.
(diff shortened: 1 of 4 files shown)


mcollovati:
Closing due to wrong `feature` prefix in branch name, that is reserved for other purposes

PR #12772, opened 2022-01-20
WIP: feat(Vite): load JS modules dynamically

### LINKED ISSUE #12497: Build generated-flow-imports.js incrementally in devevelopment mode
It seems like `target/frontend/generated-flow-imports.js` is populated with all JS modules annotations that can be found in the whole project. This makes both webpack and vite slow as all files need to be processed before a hello world view with a button and a text field can be shown.

We could instead only add the needed js modules to the file and update it whenever a js module is used for the first time. This should speed up startup time significantly.

### Versions:
    - Vaadin / Flow version: 22

platosha:
Fixes #12497

Instead of processing frontend dependencies upfront, Flow loads JS modules dynamically when components are added to the page.

This also disables Vite pre-bundling of dependencies in dev mode for faster dev server startup.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/ThemeHolder.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/ThemeHolder.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
+ *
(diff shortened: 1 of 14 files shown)


Artur-:
Should this be closed?

platosha:
Closing, we went with the bundles idea instead.

PR #12791, opened 2022-01-21
feat!: Update to Testbench 8

alvarezguille:
with Selenium 4
update imports for changes classes
add osgi.annotation dependency in test module

### DIFF
--- a/flow-tests/test-root-context/pom.xml
+++ b/flow-tests/test-root-context/pom.xml
@@ -62,6 +62,13 @@
 
         <dependency>
             <groupId>org.osgi</groupId>
+            <artifactId>osgi.annotation</artifactId>
+            <version>${osgi.annotation.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.osgi</groupId>
             <artifactId>osgi.core</artifactId>
             <version>${osgi.core.version}</version>
             <scope>provided</scope>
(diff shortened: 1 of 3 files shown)


flow-tests/test-root-context/pom.xml
```diff
@@ -60,6 +60,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>osgi.annotation</artifactId>
```
Artur-, line 65: Why is this needed? Should all projects that use TestBench add this?

pom.xml
```diff
@@ -102,7 +102,7 @@
         <maven.resources.plugin.version>3.0.2</maven.resources.plugin.version>
         <maven.clean.plugin.version>3.1.0</maven.clean.plugin.version>
         <maven.exec.plugin.version>1.6.0</maven.exec.plugin.version>
-        <testbench.version>7.0.3</testbench.version>
+        <testbench.version>8.0-SNAPSHOT</testbench.version>
```
Artur-, line 105: We should use an alpha here instead of a snapshot

alvarezguille: This is a draft PR validating that things work, once everything is green and the other changes are accepted I'll release an alpha, update the PR and mark it as ready for review.

flow-tests/test-root-context/pom.xml
Artur-: Seems like it is related to the other use of OSGi in this module. If you remove `src/main/java/com/vaadin/flow/osgi` then you do not need this dependency either

Artur- (approved):

PR #12793, opened 2022-01-21
fix: handle faulty imports

### LINKED ISSUE #12765: java.net.BindException: Cannot assign requested address: connect when trying to start a V23 Spring Boot app
### Description of the bug

When I start my application, it ends up on a whitelabel error page with 
```
This application has no explicit mapping for /error, so you are seeing this as a fallback.

Wed Jan 19 14:47:35 EET 2022
There was an unexpected error (type=Internal Server Error, status=500).
Cannot assign requested address: connect
java.net.BindException: Cannot assign requested address: connect
```

### Expected behavior

The application should start normally

[vaadin-base.zip](https://github.com/vaadin/flow/files/7897216/vaadin-base.zip)

### Versions

- Vaadin / Flow version: V23.0.0.alpha2
- Java version: Java 11 / 17
- OS version: Windows 10

caalador:
Do not throw silently for incompatible
import paths. Fix comment removal
for JS files to accept ' as string.

fixes #12765

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
@@ -127,7 +127,7 @@ public final class StringUtil {
                     result.append(character);
                     if (character.equals("\"")) {
                         state = State.IN_STRING;
-                    } else if (character.equals("\'")) {
+                    } else if (character.equals("'")) {
                         state = State.IN_STRING_APOSTROPHE;
                     }
                 }
@@ -142,7 +142,7 @@ public final class StringUtil {
                 break;
             case IN_STRING_APOSTROPHE:
                 result.append(character);
-                if (character.equals("\'")) {
+                if (character.equals("'")) {
                     state = State.NORMAL;
                 } else if (character.equals("\\") && scanner.hasNext()) {
                     result.append(scanner.next());
--- a/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
@@ -40,71 +40,26 @@ public final class StringUtil {
     /**
      * Removes comments (block comments and line comments) from the JS code.
      *
+     * @param code
+     *            code to clean comments from
      * @return the code with removed comments
      */
-    public final static String removeComments(String code) {
-        State state = State.NORMAL;
-        StringBuilder result = new StringBuilder();
-        Map<String, Character> replacements = new HashMap<>();
-        Scanner scanner = new Scanner(normalize(code, replacements));
-        scanner.useDelimiter("");
-        while (scanner.hasNext()) {
-            String character = scanner.next();
-            switch (state) {
-            case NORMAL:
-                if (character.equals("/") && scanner.hasNext()) {
-                    String nextCharacter = scanner.next();
-                    if (nextCharacter.equals("/")) {
-                        state = State.IN_LINE_COMMENT;
-                    } else if (nextCharacter.equals("*")) {
-                        state = State.IN_BLOCK_COMMENT;
-                    } else {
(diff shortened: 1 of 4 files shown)


joheriks:
The original `StringUtilTest::removeComments` is used only here and in Lit and Polymer template parsers. Is there a reason that `removeJsComments` couldn't replace the previous `removeComments`?

flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
```diff
@@ -99,6 +99,78 @@ public final static String removeComments(String code) {
...
+                    } else if (nextCharacter.equals("*")) {
+                        state = State.IN_BLOCK_COMMENT;
+                    } else {
+                        result.append(character).append(nextCharacter);
+                    }
+                } else {
+                    result.append(character);
+                    if (character.equals("\"")) {
+                        state = State.IN_STRING;
+                    } else if (character.equals("\'")) {
```
joheriks, line 130:
```suggestion
                    } else if (character.equals("'")) {
```

flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
```diff
@@ -99,6 +99,78 @@ public final static String removeComments(String code) {
...
+                result.append(character);
+                if (character.equals("\"")) {
+                    state = State.NORMAL;
+                } else if (character.equals("\\") && scanner.hasNext()) {
+                    result.append(scanner.next());
+                }
+                break;
+            case IN_STRING_APOSTROPHE:
+                result.append(character);
+                if (character.equals("\'")) {
```
joheriks, line 145:
```suggestion
                if (character.equals("'")) {
```

caalador:
The main reason is that it is used to clean the `return html'....'` html contents and with the new js version it would see the whole html as sting and not remove comments from there making it not work as wanted.

joheriks:
> The main reason is that it is used to clean the `return html'....'` html contents and with the new js version it would see the whole html as sting and not remove comments from there making it not work as wanted.

Is `html'...'` actually legal JS, doesn't this notation require backticks (``html`...` ``) so it should not be affected? If there is a way to quote the template with apostrophes, then I would suggest to parametrize the parser as the only difference between `removeComments` and `removeJsComments` is whether to transition to the `IN_STRING_APOSTROPHE` state.

caalador:
> > The main reason is that it is used to clean the `return html'....'` html contents and with the new js version it would see the whole html as sting and not remove comments from there making it not work as wanted.
> 
> Is `html'...'` actually legal JS, doesn't this notation require backticks (`` html`...` ``) so it should not be affected? If there is a way to quote the template with apostrophes, then I would suggest to parametrize the parser instead of code duplication, as the only difference between `removeComments` and `removeJsComments` is whether to transition to the `IN_STRING_APOSTROPHE` state.

Well it is for polymer template and there it is actually fine to have it with `` `|'|" ``. The original tests were at least testing that we can remove the comments from the template method.

joheriks:
> > > The main reason is that it is used to clean the `return html'....'` html contents and with the new js version it would see the whole html as sting and not remove comments from there making it not work as wanted.
> > 
> > 
> > Is `html'...'` actually legal JS, doesn't this notation require backticks (`` html`...` ``) so it should not be affected? If there is a way to quote the template with apostrophes, then I would suggest to parametrize the parser instead of code duplication, as the only difference between `removeComments` and `removeJsComments` is whether to transition to the `IN_STRING_APOSTROPHE` state.
> 
> Well it is for polymer template and there it is actually fine to have it with `` `|'|" ``. The original tests were at least testing that we can remove the comments from the template method.

Aight, my suggestion would be to refactor the method to reduce the code duplication and mention in the JavaDoc why the parameter is required.

caalador:
Done.

joheriks (approved):

PR #12798, opened 2022-01-23
test: Upgrade to Jetty 10

Artur-:


caalador (commented):
I guess this update would need some changes to the documentation samples also?

Artur-:
Do we have documentation samples about how to configure a specific version of Jetty? What are you referring to?

Artur-:
This one should be ok now

joheriks:
There is some Jetty configuration pertaining to previous Jetty version in the `vaadin-spring-tests` parent:

https://github.com/vaadin/flow/blob/f1903e2262483e33400b3e4c0f149fba01601a58/flow-tests/vaadin-spring-tests/pom.xml#L292-L311

It looks like this could be removed altogether as it gets it from the `flow-tests` parent.

joheriks (approved):

PR #12805, opened 2022-01-24
test: Ignore generated index.html files inside tests

Artur-:


.gitignore
```diff
@@ -50,6 +50,7 @@ flow-tests/**/package.json
 flow-tests/**/webpack*.js
 flow-tests/**/tsconfig.json
 flow-tests/**/pnpm-lock.yaml
+flow-tests/**/index.html
```
caalador, line 53: or should we just ignore `flow-tests/**fronted/generated` to not get any generated files?

Artur-: there is already `flow-tests/**/generated/**` there

caalador (approved):

Issue #12807, opened 2022-01-24
Cannot invoke "com.vaadin.flow.server.VaadinSession.getService()" because the return value of "com.vaadin.flow.component.UI.getSession()" is null


TatuLund:
### Description of the bug

In very rare conditions this kind of exception is thrown

java.lang.NullPointerException: Cannot invoke "com.vaadin.flow.server.VaadinSession.getService()" because the return value of "com.vaadin.flow.component.UI.getSession()" is null
at com.vaadin.flow.server.communication.UidlWriter.createUidl(UidlWriter.java:178)
at com.vaadin.flow.server.communication.UidlRequestHandler.writeUidl(UidlRequestHandler.java:124)
at com.vaadin.flow.server.communication.UidlRequestHandler.synchronizedHandleRequest(UidlRequestHandler.java:93)
at com.vaadin.flow.server.SynchronizedRequestHandler.handleRequest(SynchronizedRequestHandler.java:40)
at com.vaadin.flow.server.VaadinService.handleRequest(VaadinService.java:1578)
at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:252)
at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:111)
at javax.servlet.http.HttpServlet.service(HttpServlet.java:764)
at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:227)
at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162)
at org.apache.catalina.core.ApplicationDispatcher.invoke(ApplicationDispatcher.java:711)
at org.apache.catalina.core.ApplicationDispatcher.processRequest(ApplicationDispatcher.java:459)
at org.apache.catalina.core.ApplicationDispatcher.doForward(ApplicationDispatcher.java:353)
at org.apache.catalina.core.ApplicationDispatcher.forward(ApplicationDispatcher.java:313)
at org.springframework.web.servlet.mvc.ServletForwardingController.handleRequestInternal(ServletForwardingController.java:141)
at org.springframework.web.servlet.mvc.AbstractController.handleRequest(AbstractController.java:177)
at org.springframework.web.servlet.mvc.SimpleControllerHandlerAdapter.handle(SimpleControllerHandlerAdapter.java:51)
at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1067)
[...]

PR #12808, opened 2022-01-24
fix: Use service already assigned

TatuLund:
Code cleanup

fixes: https://github.com/vaadin/flow/issues/12807

Note, exception is thrown here: https://github.com/vaadin/flow/blob/2.8/flow-server/src/main/java/com/vaadin/flow/server/communication/UidlWriter.java#L178

But earlier session.getService() did not throw https://github.com/vaadin/flow/blob/2.8/flow-server/src/main/java/com/vaadin/flow/server/communication/UidlWriter.java#L157

Hence I assume using stored service reference will make code less prone to throw


Artur- (approved):
Nothing wrong with this change but if this fixes some issue, something else might be wrong

PR #12822, opened 2022-01-25
fix: reliable deletion of node_modules

### LINKED ISSUE #12810: PiT v23: Upgrading from v22, unable to delete file in node_module
### Description of the bug

When bumping from v22 to v23 (alpha3) flow is unable to remove certain files under node_modules folder, and fronted compilation fails

### Expected behavior

No errors should be shown and compilation succeeded 

```
git clone https://github.com/vaadin/base-starter-flow-osgi.git 
cd base-starter-flow-osgi/
git checkout v22 
mvn clean install
mvn versions:set-property -Dproperty=vaadin.version -DnewVersion=23.0.0.alpha3
mvn clean install
```

Note that it also happens when migrating other projects, I found the same problem in:
- Karaf example: https://github.com/vaadin/vaadin-flow-karaf-example (master to v23)
- Start LTS app with all views: https://start.vaadin.com/app?preset=all-lts (v24 to v23)

### workaround

Either:
- Use pnpm in v23
- Remove manually `node_modules`

### Versions

- Vaadin / Flow version: v23.0.0.alpha3
- Java version: 11
- OS version: mac and linux

joheriks:
Do not set permissions nor follow symlinks when 
deleting node_modules.

Fixes #12810

### DIFF
--- a/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/CleanFrontendMojo.java
+++ b/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/CleanFrontendMojo.java
@@ -26,6 +26,8 @@ import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 
+import com.vaadin.flow.server.frontend.FrontendUtils;
+
 import elemental.json.Json;
 import elemental.json.JsonObject;
 import elemental.json.impl.JsonUtil;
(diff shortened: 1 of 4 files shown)


Artur-:
This looks like a reversed approach i.e. fixing the issue after it has happened (files created without writable permission). What is causing the files to be created with the wrong permissions?

joheriks:
> This looks like a reversed approach i.e. fixing the issue after it has happened (files created without writable permission). What is causing the files to be created with the wrong permissions?

Actually, setting write permission in the initial pass is an additional measure. It works without it also, but I added it as I am not sure if `pnpm` / `npm` could create a file with no user write permission in `node_modules`. What caused the permissions to get set wrong was when the `FileUtils`-based deletion failed trying override the permissions of a symlink that pointed to a file that had been deleted.

Artur- (approved):
Can this be unit tested?

joheriks:
I will wing together a unit test that deletes a directory containing some internal symbolic links, though it may be hard to get it to actually fail with the previous impl as the file visiting order is non-deterministic.

Artur- (dismissed):

Artur- (approved):

Artur- (approved):

PR #12839, opened 2022-01-25
test: Try to find chromedriver also from PATH

Artur-:
The driver download plugin does not work in Java 17 nor does it support M1 macs


flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeDriverLocator.java
```diff
@@ -87,6 +103,28 @@ private static String getDriverLocation() {
 
     }
 
+    private static String getDriverFromPath() {
+        try {
+            Process p = new ProcessBuilder().command("which", "chromedriver")
```
caalador, line 108: This should take into account that windows has `where`
```suggestion
            Process p = new ProcessBuilder().command(FrontendUtils.isWindows() ? "where" : "which", "chromedriver")
```

caalador (approved):

Issue #12383, opened 2021-11-17
[Vite] Offline support


gilberto-torrezan:
A follow-up ticket to https://github.com/vaadin/flow/issues/12052

What is missing for offline support, according to @caalador :

sw-vite.ts only does `precacheAndRoute` so everything else from sw.ts
The biggest missing part is to  know what does the `manifestTransforms: [swManifestTransform]`
collect and populate.

PR #12857, opened 2022-01-26
fix: make Vite PWA apps work offline in prod mode

vursen:
## Description

The PR fixes several bugs that caused Vite PWA applications not to work offline in production mode.

Note, Vite PWAs don't work offline in dev mode at this point because Vite uses a different approach for loading modules in dev mode that needs special treatment. The dev mode will be considered separately.

**What is fixed:**
- [x] Fixed the `OFFLINE_PATH` constant in `sw.js` so that it points to `settings.offlinePath`.
- [x] Fixed injecting manifest to `sw.js` (was previously done in the `buildStart` hook though should've been in `closeBundle`).
- [x] Fixed the manifest glob pattern to include not only JS files but all the output resources, including icons.
- [x] Added missing logic that rewrites the URL of `index.html` in the manifest so that the application is available at the `/` URL.

**Test coverage:**
- [x] Added ITs for the case: "Basic Vite PWA app with a custom `offlinePath`".
- [x] Added ITs for the case: "Basic Vite PWA app".

Part of #12383

## Type of change

- [x] Bugfix

## Checklist


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -44,6 +46,69 @@ const themeOptions = {
...
+        await bundle.close()
+      }
+      // end of resolve and transpilation
+    },
+    async closeBundle() {
+      await injectManifest({
+        swSrc: path.resolve(frontendBundleFolder, 'sw.js'),
+        swDest: path.resolve(frontendBundleFolder, 'sw.js'),
+        globDirectory: frontendBundleFolder,
+        globPatterns: ['**/*'],
```
vursen, line 103: **question:** Do you know of any other files that need to be excluded as well?

vursen: It is probably important to copy these properties (from Webpack) to the Vite config as well:

https://github.com/vaadin/flow/blob/37827019b819cbba15457c310e270137949f2c8c/flow-server/src/main/resources/webpack.generated.js#L158-L159

flow-tests/test-frontend/vite-pwa-production-custom-offline-path/src/test/java/com/vaadin/viteapp/PwaTestIT.java
```diff
@@ -0,0 +1,69 @@
...
+public class PwaTestIT extends ChromeDeviceTest {
+    @Before
+    public void init() {
+        open();
+        waitForServiceWorkerReady();
+    }
+
+    @After
+    public void teardown() throws IOException {
+        setOnline();
```
vursen, line 40: **question:** Is it necessary to explicitly reset the network state?

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -44,6 +46,76 @@ const themeOptions = {
...
+    const indexEntry = manifest.find((entry) => entry.url === 'index.html');
+    if (indexEntry) {
+      indexEntry.url = appShellUrl;
+    }
+
+    return { manifest, warnings: [] };
+  }
+
+  return {
+    name: 'vaadin:inject-manifest-to-sw',
```
vursen, line 102: **note:** Adding `vaadin`-prefix will prevent possible interference Vaadin plugins with the user's custom plugins which might coincidently have the same name. We should consider this prefix for the other Vaadin plugins for the same reason.

vursen: Did it in a separate PR: https://github.com/vaadin/flow/pull/12895

flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeDeviceTest.java
```diff
@@ -30,6 +31,9 @@
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.chrome.ChromeDriver;
 import org.openqa.selenium.chrome.ChromeOptions;
+import org.openqa.selenium.devtools.DevTools;
+import org.openqa.selenium.devtools.HasDevTools;
+import org.openqa.selenium.devtools.v96.network.Network;
```
vursen, line 36: This was the only way to reach the `Network` namespace. Even though according to `drivers.xml`, we use ChromeDriver v97, the `v96` namespace was the latest version available in the autocomplete.

vursen: These warnings are shown when running tests:

```
WARNING: Unable to find an exact match for CDP version 97, so returning the closest version found: 96
Feb 01, 2022 2:21:55 PM org.openqa.selenium.devtools.CdpVersionFinder findNearestMatch
```

Artur-: https://github.com/SeleniumHQ/selenium/issues/10286

flow-server/src/main/resources/vite.generated.ts
platosha: `maximumFileSizeToCacheInBytes` is probably useful.

`.cache.js` URLs are created by webpack, not Vite indeed. Could be useful to support [Embedded applications](https://vaadin.com/docs/latest/flow/integrations/embedding) created with webpack, I guess?

platosha: Not aware of any files to exclude.

flow-tests/test-frontend/vite-pwa-production-custom-offline-path/src/test/java/com/vaadin/viteapp/PwaTestIT.java
platosha: Most likely not, browser sessions are not reused between test suites. I haven’t seen a case when it is reused.

I assume, for a clean session it would be using default online state anyway.

vursen: By "a test suite" you mean a Java class with tests, do you?

vursen: There is no need for this `teardown` reset logic, I checked it. Deleted.

flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeDeviceTest.java
vursen: Now that TestBench uses Selenium 4.1.2, the `v97` namespace is available.

flow-server/src/main/resources/vite.generated.ts
vursen: Specified the `maximumFileSizeToCacheInBytes` option.

platosha (approved):

PR #12863, opened 2022-01-27
fix: update css-loader

caalador:
Update css-loader to get a
newer version of postcss


joheriks (approved):

joheriks:
We'll want this for 2.8 as well, but it will require manual cherry-pick.

PR #12870, opened 2022-01-27
test: Rewrite tests to use Spring Boot + devtools for reload

Artur-:
Jetty 10 does not support using scan combined with jetty:start.

This replaces tests need redeployment during the test with Spring Boot using Spring Boot devtools


Artur-:
Now this should work...

flow-tests/test-redeployment/pom.xml
```diff
@@ -32,6 +32,34 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-devtools</artifactId>
+            <version>${spring.boot.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-spring</artifactId>
+            <version>20.0-SNAPSHOT</version>
```
knoobie, line 60: With this, the flow build depends on the spring repo resulting in a circular dependency.. is that worth it?

Artur-: We should probably do https://github.com/vaadin/spring/issues/922

Artur-:
Replacing this with #13010 as sonarcloud doesn't seem to understand what is not included in this PR anymore

PR #12889, opened 2022-01-31
fix: moved shortucts only initialized once

### LINKED ISSUE #12307: Shortcuts: keydown-Listener not removed on client-side
### Description of the bug
I have a dialog in which I register a few `Shortcut`s. For performance reasons when closing the dialog, I only hide it by setting the overlay to `display:none`, to be able to reuse it, instead of closing it.
When opening/showing the dialog, I remove all `ShortuctEventListener`s and register new ones, since the available actions could change in the meantime.
Now if I open and close the dialog mulitple times, the `ShortcutEventListener` gets triggered multiple times. While debugging, I found that the `ShortcutRegistration` instance is always the same, so it seems like there are multiple events coming from the client.
Inspecting the element on the client-side it indeed seems like the old listeners are not removed:
Dialog after opening once:
image
Dialog after opening twice:
image

1. Download a [vaadin flow starter](https://start.vaadin.com/app/) with an empty view
2. Add the following code to the `EmptyView.java`:
```
@PageTitle("Empty")
@Route(value = "empty", layout = MainLayout.class)
public class EmptyView extends VerticalLayout {
  private final Dialog dlg;
  private Registration shortcutRegistration = () -> {};

  public EmptyView() {
    dlg = new Dialog();
    dlg.setCloseOnEsc(false);
    dlg.setCloseOnOutsideClick(false);
    Shortcuts.setShortcutListenOnElement("this.$.overlay", dlg);
    VerticalLayout vlDlg = new VerticalLayout();
    vlDlg.getElement().setAttribute("tabindex", "-1");
    vlDlg.add(
      new Span("Press 'Del' to trigger shortcut"),
      new Span("Press 'Esc' to hide dialog")
    );
    dlg.add(vlDlg);
    add(new Button("Open Dialog", e -> {
      updateShortcuts();
      if (!dlg.isOpened())
        dlg.open();
      else {
        dlg.setModal(true);
        dlg.getElement().executeJs(  "this.$.overlay.style.display = \"\";");
        vlDlg.getElement().executeJs("this.focus();");
      }
    }));
  }
  private void updateShortcuts() {
    shortcutRegistration.remove();
    shortcutRegistration = Registration.combine(
      Shortcuts.addShortcutListener(this, e -> {
        dlg.setModal(false);
        dlg.getElement().executeJs("this.$.overlay.style.display = \"none\";");
      }, Key.ESCAPE).listenOn(dlg),
      Shortcuts.addShortcutListener(this, e -> System.out.println("Received del"), Key.DELETE)
        .listenOn(dlg)
    );
  }
}
```
3. Run the application and open the dialog by clicking the button
4. Close (hide) the dialog with `ESC`
5. Reopen the dialog by clicking the button
6. Press `DEL` to trigger the other shortcut and see the logs in the console.
   The `DEL` shortcut was logged multiple times

### Expected behavior
Removing the `ShortcutRegistration` should also remove the client-side listener

### Actual behavior
Removing the `ShortcutRegistration` does not remove the client-side listener and therefor the server-side listener is triggered multiple times.

### Versions:
    - Vaadin / Flow version: 21.0.3
    - Java version: 11.0.10 AdoptOpenJDK
    - OS version: Windows 10 64bit
    - Browser version (if applicable): Chrome 95.0.4638.69
    - IDE (if applicable): IntelliJ IDEA Community Edition

caalador:
Fix issuer with initializing shortcuts
for component delegation so the
shortcuts are not added with already
initialized actions.

Fixes #12307

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
@@ -15,6 +15,10 @@
  */
 package com.vaadin.flow.internal;
 
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.MessageDigestSpi;
+import java.security.NoSuchAlgorithmException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -149,5 +153,33 @@ public final class StringUtil {
             return generateReplacement(str);
         }
         return replacement;
+    }
+
+    /**
+     * Generate a hash for given content.
+     *
+     * @param content
+     *            content to generate hash for
+     * @return hash String for given content. In case content is null or empty
+     *         returns empty String.
+     */
+    public static String getHash(String content) {
+        if (content == null || content.isEmpty()) {
+            return "";
+        }
+
+        return bytesToHex(MessageDigestUtil.sha256(content));
+    }
+
+    private static String bytesToHex(byte[] hash) {
+        StringBuilder result = new StringBuilder();
+        for (byte bit : hash) {
+            String hex = Integer.toHexString(0xff & bit);
+            if (hex.length() == 1) {
+                result.append('0');
+            }
+            result.append(hex);
+        }
+        return result.toString();
     }
 }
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
```diff
@@ -150,4 +154,32 @@ private static String generateReplacement(String str) {
...
+        if (content == null || content.isEmpty()) {
+            return "";
+        }
+
+        return bytesToHex(MessageDigestUtil.sha256(content));
+    }
+
+    private static String bytesToHex(byte[] hash) {
+        StringBuilder result = new StringBuilder();
+        for (byte bit : hash) {
```
joheriks, line 176: It is not a bit, 
so perhaps  `hashByte`, 
and yes, very nit.

caalador: done

anssit (approved):

PR #12906, opened 2022-02-01
Fix: do not remove package-lock on platform update

### LINKED ISSUE #12796: Do not delete package-lock.json on version upgrades
### Description of the bug

package-lock.json defines versions used in my project. There should be no need to remove it completely and thus upgrade all dependencies to the latest version just because I upgrade Vaadin. For all other npm dependencies, it is enough to install a new version which in turns upgrades that version and maybe some dependency versions in the lock file.

A side effect of the current behavior is that it takes minutes to run `npm install` after upgrading Vaadin instead of seconds because of https://github.com/npm/cli/issues/3703

### Expected behavior

Upgrading Vaadin should only upgrade Vaadin package versions. If you today upgrade the Vaadin version you might get the newest version of `@babel/core` which triggers a Node segfault bug in certain conditions. If the version is force upgraded as a side effect of upgrading Vaadin, it looks like a problem in Vaadin. Which it is not.

```
npx @vaadin/cli init hello --latest
cd hello
npx upgrade-vaadin 22.0.0
```

This should not delete and recreate `pnpm-lock.yaml`

### Versions

Vaadin 22

caalador:
Leave package-lock even if platform version has updated.

Fixes #12796

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractNodeUpdatePackagesTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractNodeUpdatePackagesTest.java
@@ -227,7 +227,7 @@ public abstract class AbstractNodeUpdatePackagesTest
                 getScanner(classFinder), baseDir, generatedDir, resourcesDir,
                 false, true, TARGET, featureFlags);
         packageUpdater.execute();
-        Assert.assertFalse("NPM package-lock should be removed fro pnpm",
+        Assert.assertFalse("NPM package-lock should be removed for pnpm",
                 packageLock.exists());
     }
 
--- a/flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractNodeUpdatePackagesTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractNodeUpdatePackagesTest.java
@@ -227,7 +227,8 @@ public abstract class AbstractNodeUpdatePackagesTest
                 getScanner(classFinder), baseDir, generatedDir, resourcesDir,
                 false, true, TARGET, featureFlags);
         packageUpdater.execute();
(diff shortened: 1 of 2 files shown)


Artur-:
Does this have any non wanted side effects?

caalador:
After running some scenarios I couldn't find anything and downgraded packages should also update the package-lock correctly.

manolo:
I guess removing lock file still is needed when upgrading npm, at least when upgrading to 8.3 and there are overrides (see #12908).
But probably this case was not considered in flow, only platform update.

caalador:
Flow hasn't done any checks for npm version so that it would have had any cleanup implications.

joheriks:
We also delete `package-lock.json` in `flow:clean-frontend`. Not sure if that is consistent after this change, feels like we should either manage the lockfile or not manage it.

flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractNodeUpdatePackagesTest.java
```diff
@@ -227,7 +227,8 @@ public void pnpmIsInUse_packageLockExists_removePackageLock()
                 getScanner(classFinder), baseDir, generatedDir, resourcesDir,
                 false, true, TARGET, featureFlags);
         packageUpdater.execute();
-        Assert.assertFalse(packageLock.exists());
+        Assert.assertFalse("NPM package-lock should be removed fro pnpm",
```
joheriks, line 230:
```suggestion
        Assert.assertFalse("NPM package-lock should be removed for pnpm",
```

caalador:
The clean-frontend is not an automatic clean and is requested as a way to get the project to a clean state.
So as it's not an automatic cleaning done due to a change (update of platform version) I do not see any conflict there.

joheriks (approved):

PR #12911, opened 2022-02-02
fix: also consider package.json for lock versions

### LINKED ISSUE #12662: Generated versions.json should also take into account package.json
### Description of the bug

When running a project with `vaadin_versions.json` from platform only the versions defined there will be set to the generated `versions.json`, but this may leave out versions to be locked like for instance polymer that is only defined in `package.json` `dependencies`.

### Expected behavior

Also versions defined in `package.json` should be locked. so if package.json would contain `polymer: 3.2.0` then that should be enforced and not be able to get a mix of 3.2.0 and 3.4.1

This would seem to be reproduclible by adding for instance https://vaadin.com/directory/component/pdf-browser to the project and using it as: `PdfBrowserViewer viewer = new PdfBrowserViewer();` without even adding it to the layout.

### Versions

- Vaadin / Flow version: 22/9.0+

caalador:
Even if we have a vaadin_versions.json
we should also take into account versions
defined in package.json to be locked
for transitive dependencies.

fixes #12662

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -585,8 +585,15 @@ public abstract class NodeUpdater implements FallibleCommand {
         File versions = new File(generatedFolder, "versions.json");
 
         JsonObject versionsJson = getPlatformPinnedDependencies();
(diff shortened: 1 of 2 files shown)


joheriks (approved):

caalador:
Cherrypick should be done from the 2.8/2.7 Pr as the changes in master are not applicable.

PR #12916, opened 2022-02-02
fix!: no double decode query parameters

### LINKED ISSUE #12887: Problem with fix for #12354
### Description of the bug

This is a problem related to the fix of [this issue](https://github.com/vaadin/flow/issues/12354).
That fix only [solved the problem for the plus sign](https://github.com/vaadin/flow/pull/12432) "+" but there are similar issues with some other characters as "%".
Is it possible to avoid the double decoding for preventing a similar issue with other characters?

### Expected behavior

All of the double decoding related issues should be solved with the fix for the related issue

In this case with the "%" character, it's encoded as "%25" in the query parameter, but the double decoding in the Location constructor leads to an exception. As an example:
Location l = new Location("https://example.com?param=%25HF")
After decoding once, "%25HF" becomes "%HF". Since "%HF" is not a valid url-encoded pattern, it throws an exception.

### Versions

- Vaadin / Flow version: 14.8.0 (but probably also remains in the latest version)

joheriks:
Fix 1: Don't double-decode query parameters.

Fix 2: URI.getQuery is broken, making it impossible to distinguish `&` separating 
parameters from literal ampersands:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8214423'

Fix 3: Return a properly encoded query string from
QueryParamaters.getQueryString instead of justing joining the decoded
strings in the map. This is a potentially breaking change.

Fixes #12887

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -18,6 +18,7 @@ package com.vaadin.flow.router;
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
+import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -127,17 +128,11 @@ public class QueryParameters implements Serializable {
     private static List<String> makeQueryParamList(String paramAndValue) {
         int index = paramAndValue.indexOf('=');
         if (index == -1) {
-            return Collections.singletonList(paramAndValue);
+            return Collections.singletonList(decode(paramAndValue));
         }
         String param = paramAndValue.substring(0, index);
-        String value = paramAndValue.substring(index + 1).replace("+", "%2B");
-        try {
-            value = URLDecoder.decode(value, "utf-8");
-        } catch (UnsupportedEncodingException e) {
-            throw new RuntimeException(
-                    "Unable to decode parameter value: " + value, e);
-        }
-        return Arrays.asList(param, value);
+        String value = paramAndValue.substring(index + 1);
+        return Arrays.asList(decode(param), decode(value));
     }
 
     private static List<String> getParameterValues(List<String> paramAndValue) {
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -193,10 +192,28 @@ public String getQueryString() {
             Entry<String, List<String>> entry) {
         List<String> params = entry.getValue();
         if (params.size() == 1 && "".equals(params.get(0))) {
-            return Stream.of(entry.getKey());
+            return Stream.of(encode(entry.getKey()));
```
caalador, line 195: Use `UriUtil.encodeURIComponent` for encoding to get an equivalent for the original and not for instance get a + instead of %20 for space.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
joheriks: Updated to use `UrlUtil.encodeURIComponent`. Though none of the solutions can always give the original encoding as that is forgotten after decoding (as we reconstruct the query string from the decoded parameters in the map).

caalador: Well a more correct encoding of the uri.

flow-server/src/test/java/com/vaadin/flow/router/LocationTest.java
```diff
@@ -112,14 +112,35 @@ public void locationFromSegments() {
...
+        Assert.assertEquals("'%2B' should not be decoded in query param string",
+                "%25HF", queryParameters.getQueryString());
+
+        queryParameters = new Location("home?p=%26&q=%20").getQueryParameters();
+        Assert.assertEquals("'%26' should be decoded in map", "&",
+                queryParameters.getParameters().get("p").get(0));
+        Assert.assertEquals("'%20' should be decoded in map", " ",
+                queryParameters.getParameters().get("q").get(0));
+        Assert.assertEquals(
+                "'%26' and '%2B' should not be decoded in query param string",
```
caalador, line 142:
```suggestion
                "'%26' and '%20' should not be decoded in query param string",
```

flow-server/src/test/java/com/vaadin/flow/router/LocationTest.java
```diff
@@ -112,14 +112,35 @@ public void locationFromSegments() {
...
                 "someValue1+someValue2",
-                location.getQueryParameters().getQueryString());
+                queryParameters.getParameters().keySet().iterator().next());
+        Assert.assertEquals("'%2B' should not be decoded in query param string",
+                "someValue1%2BsomeValue2", queryParameters.getQueryString());
+
+        queryParameters = new Location("home?%25HF").getQueryParameters();
+        Assert.assertEquals("'%25' should be decoded in map", "%HF",
+                queryParameters.getParameters().keySet().iterator().next());
+        Assert.assertEquals("'%2B' should not be decoded in query param string",
```
caalador, line 133:
```suggestion
        Assert.assertEquals("'%25' should not be decoded in query param string",
```

caalador (commented):
2 Wrong values in assert messages.

flow-server/src/test/java/com/vaadin/flow/router/LocationTest.java
joheriks: Done

caalador (approved):

PR #12928, opened 2022-02-03
feat: make Vite PWA apps work offline in dev mode

vursen:
## Description

This PR focuses on making Vite PWA apps work offline in dev mode. The PR achieves it in a bit different manner than it is done for Webpack. The central idea here and difference from Webpack is that every `/VAADIN/` request ends up cached for offline only after it has been requested once online. This approach has been taken because we don't want to perform any pre-scaning / pre-caching of the whole project during Vite startup for the sake of performance. This approach also appears to be in line with the Vite general concept that is, everything is processed on demand.

A brief overview of the PR's changes:

- [x] Configured a `NetworkFirst` strategy handler (caches everything at runtime) for `/VAADIN/*` requests in the service worker. The runtime cache gets cleaned on every service worker `activate` event.
- [x] Added a `NetworkFirst` strategy handler for `/index.html` that activates when the offline fallback path points to `index.html`. In other cases, Flow defines the offline fallback path in the manifest which leads to its pre-caching).
- [x] `/VAADIN/__vite_ping` requests are refused when offline to prevent a recurring page reload.
- [x] ~~Configured a watcher that triggers re-building for the service worker for the case when the user has a custom service worker.~~ (Reverted)
- [x] Added tests on the basic PWA app with Vite in dev mode

Fixes https://github.com/vaadin/flow/issues/12383

## Type of change

- [x] Feature

## Checklist


flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw.ts
```diff
@@ -1,21 +1,33 @@
...
-  __WB_MANIFEST: Array<PrecacheEntry>;
-  additionalManifestEntries?: Array<PrecacheEntry>;
+  __WB_MANIFEST: PrecacheEntry[];
+  additionalManifestEntries?: PrecacheEntry[];
 };
 
-skipWaiting();
+declare var OFFLINE_PATH: string; // defined by Webpack/Vite
+
+self.skipWaiting();
```
vursen, line 17: https://developers.google.com/web/tools/workbox/modules/workbox-core#the_skipwaiting_wrapper_is_deprecated

> For these reasons, workbox-core's skipWaiting() is deprecated, and developers should switch to calling self.skipWaiting() directly. Unlike with self.clients.claim(), self.skipWaiting() will not throw an exception if called at the "wrong" time, so there is no need to wrap it in an event handler.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw.ts
```diff
@@ -24,61 +37,111 @@ declare var OFFLINE_PATH: string; // defined by webpack.generated.js
...
-    return precachedResponse || error;
-  }
+const networkOnly = new NetworkOnly({
+  plugins: [checkConnectionPlugin()]
 });
+const networkFirst = new NetworkFirst({
+  plugins: [checkConnectionPlugin()]
+});
+
+if (process.env.NODE_ENV === 'development') {
```
platosha, line 76: Tried this with webpack, seems that currently it alters the dev mode offline behaviour also for webpack.

How about adding some env flag, so that we only change the dev mode behaviour for Vite?

platosha (commented):
Missing source maps for `sw.js` in dev mode for some reason.

I tried enabling them by adding the `bundle.write({sourcemap: true})` option, but this resulted in an empty source map file in the output and served.

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -250,9 +270,9 @@ export const vaadinConfig: UserConfigFn = (env) => {
           },
         },
       },
-      checker({
-        typescript: true
-      })
+      // checker({
+      //   typescript: true
+      // })
```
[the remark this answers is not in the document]
vursen: It was only for testing purposes. Reverted it back.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw.ts
vursen: Good point, added a flag `VITE_ENABLED`. Not sure about the naming, let me know if you would like to use a different one.

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -67,7 +66,25 @@ function transpileSWPlugin(): PluginOption {
...
+        includedPluginNames.push('vite:client-inject');
+      }
+
+      const plugins = config.plugins.filter((p) => includedPluginNames.includes(p.name));
+      if (config.mode === 'development') {
+        plugins.push({
+          name: 'vaadin:inject-vite-env',
+          enforce: 'pre',
+          transform(code, id) {
+            if (id.endsWith('sw.ts')) {
```
vursen, line 79:
```suggestion
            if (id.endsWith('/sw.ts')) {
```

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -46,43 +48,78 @@ const themeOptions = {
...
+      if (config.command === 'serve') {
+        watcher = chokidar.watch(path.resolve(settings.clientServiceWorkerSource), {
+          ignoreInitial: true,
+        });
+        watcher.on('all', async () => {
+          try {
+            await build();
+          } catch (error) {
+            console.error(error);
+          }
```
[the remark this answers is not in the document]
vursen: ignore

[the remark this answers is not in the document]
vursen: @sonatype-lift ignore

flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeDeviceTest.java
```diff
@@ -180,13 +188,38 @@ protected void setConnectionType(
...
+            ImmutableMap.Builder<String, Object> serialized = ImmutableMap.builder();
+            serialized.put("id", "100");
+            serialized.put("sessionId", sessionId);
+            serialized.put("method", command.getMethod());
+            serialized.put("params", command.getParams());
+            StringBuilder json = new StringBuilder();
+            try (JsonOutput out = new Json().newOutput(json).writeClassName(false)) {
+                out.write(serialized.build());
+            }
+            String jsonString = json.toString();
```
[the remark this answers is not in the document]
vursen: @sonatype-lift ignore

flow-test-util/src/main/java/com/vaadin/flow/testutil/DevToolsWrapper.java
```diff
@@ -0,0 +1,159 @@
...
+                    .getDeclaredField("capabilities");
+            capabilitiesField.setAccessible(true);
+
+            String sessionId = driver.getSessionId().toString();
+            String devtoolsUrl = String.format("ws://%s:%s/devtools/%s/page",
+                    remoteURL.getHost(), remoteURL.getPort(), sessionId);
+
+            MutableCapabilities mutableCapabilities = (MutableCapabilities) capabilitiesField
+                    .get(driver);
+            mutableCapabilities.setCapability("se:cdp", devtoolsUrl);
```
Artur-, line 62: Can't these be set before creating the original driver instead without using any reflection?
E.g. 
```
chromeOptions.setCapability("se:cdp", devtoolsUrl);
```

flow-test-util/src/main/java/com/vaadin/flow/testutil/DevToolsWrapper.java
```diff
@@ -0,0 +1,159 @@
...
+    public void setCacheDisabled(Boolean isDisabled) {
+        sendToAllTargets(Network.enable(Optional.empty(), Optional.empty(),
+                Optional.empty()));
+        sendToAllTargets(Network.setCacheDisabled(isDisabled));
+    }
+
+    /**
+     * Creates a custom DevTools CDP connection if there is not one yet.
+     *
+     * Note, there is already a CDP connection provided by {@class DevTools} but
```
Artur-, line 101:
```suggestion
     * Note, there is already a CDP connection provided by {@link DevTools} but
```

Artur- (changes requested):

flow-test-util/src/main/java/com/vaadin/flow/testutil/DevToolsWrapper.java
vursen: It would be great. However, it doesn't seem possible since it requires `sessionId` that becomes only available after a session is created. I also looked up if there is a template format, but didn't find anything like that.

Anyway, I decided to move this workaround back to `ChromeDeviceTest.java` where the web driver initializes as this somewhat feels to be a more appropriate place for it.

flow-test-util/src/main/java/com/vaadin/flow/testutil/DevToolsWrapper.java
```diff
@@ -0,0 +1,159 @@
...
+            String devtoolsUrl = String.format("ws://%s:%s/devtools/%s/page",
+                    remoteURL.getHost(), remoteURL.getPort(), sessionId);
+
+            MutableCapabilities mutableCapabilities = (MutableCapabilities) capabilitiesField
+                    .get(driver);
+            mutableCapabilities.setCapability("se:cdp", devtoolsUrl);
+            mutableCapabilities.setCapability("se:cdpVersion",
+                    mutableCapabilities.getBrowserVersion());
+        } catch (Exception e) {
+            System.err.println(
```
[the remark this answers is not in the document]
vursen: Good point, throwing the runtime exception now.

flow-test-util/src/main/java/com/vaadin/flow/testutil/DevToolsWrapper.java
vursen: Fixed.

platosha (approved):

Artur- (dismissed):

Artur- (approved):

Artur-:
This should not be picked after all

PR #12929, opened 2022-02-03
spike: Spring add-on validation in Flow repo

joheriks:
DO NOT MERGE!


joheriks:
> The other way to remove this would be to remove `slf4j-simple` from the parent pom. Not sure if it matters which way you do it

Did it the same way in the actual PR for now: https://github.com/vaadin/flow/pull/13002/commits/9d8f9fa4274d0796f8fe47cffad8a4de5fcc2924

Closing this.

PR #12944, opened 2022-02-07
chore: set version to 23.1

joheriks:
Also updates README.


Artur- (approved):
There are some duplicate/redundant versions in some pom files. These could be removed in a separate PR

joheriks:
> There are some duplicate/redundant versions in some pom files. These could be removed in a separate PR

 #12946 for this

PR #11372, opened 2021-07-01
fix: don't create a new UI instance for service worker resource

denis-anisimov:
fixes #6454

## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #12963, opened 2022-02-09
fix: don't create a new UI instance for service worker resource

### LINKED ISSUE #12922: Service work serving up negative hits
Not certain if I know how to reproduce this but.

Vaadin 22.

I've just spent half a day trying to work out why I was getting an invalid copy of VAADIN/build/vaadin-2-398f0dc50fc60665f718.cache.js.

I don't think the actual file was the issue but rather the process.

I have a login route that then redirects to a login validator route. The login process was failing and the vaadin debug widget would say the .js file contained invalid content.
It seems that every second request to the above noted file would fail.
If I did a hard cache reload the file would be served correctly but the second time it was served it would have invalid content. Instead of the expected java script it contained the vaadin boot loader HTML page.

I eventually realised that the second request was being provided by the service worker.
I then cleared the service workers cache and the problem went away.

So here is my theory.

At some point, I had my system mis-configured so when vaadin was asked to serve the above noted js file it instead returned with vaadin boot page with an error (possibly the we didn't find a route, here are the available ones).
The service worker then cached this response because it return a 200 rather than something like a 404.
From then on, whenever the service work served the file from cache it would return this same invalid response.

caalador:
Also ignore Service-Worker request for
IndexHtmlRequestHandler

Fixes #12922

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
@@ -261,7 +261,7 @@ public class IndexHtmlRequestHandler extends JavaScriptBootstrapHandler {
     }
 
     /**
-     * Checks if the request is potentially a request for a HTML page.
+     * Checks if the request is potentially a request for an HTML page.
      * 
      * @param request
      *            the request to check
@@ -270,6 +270,9 @@ public class IndexHtmlRequestHandler extends JavaScriptBootstrapHandler {
      *         image or something else
      */
     protected boolean isRequestForHtml(VaadinRequest request) {
+        if (request.getHeader(SERVICE_WORKER_HEADER) != null) {
+            return false;
+        }
         String fetchDest = request.getHeader("Sec-Fetch-Dest");
         if (fetchDest == null) {
             // Old browsers do not send the header at all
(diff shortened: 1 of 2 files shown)


Artur-:
Where is actually the `Service-Worker` header set?

caalador:
I´m trusting the #11372 fix as the IndexHtmlRequestHandler overrides parent methods.
@joheriks might know what adds the header as I can't say.

vursen:
Couldn’t find any official mention of a `Service-Worker` header. I also checked this locally and I don’t see the service worker sending such headers.

Nevertheless, I'm not sure if the fix itself is correct. Even if such a header exists, it won't be correct to prevent any SW requests to `index.html`. There is at least one case that comes to my mind: the service worker will have to make a request to `index.html` to put it in the runtime cache in Vite dev mode.

joheriks:
The header is standard, but I believe only added when the browser requests the initial `sw.js` script:
https://w3c.github.io/ServiceWorker/#service-worker-script-request

A brief article about it here: https://medium.com/dev-channel/two-http-headers-related-to-service-workers-you-never-may-have-heard-of-c8862f76cc60

vursen:
Oh, well, I'm taking my words back. If it is only for the `sw.js` script, then the fix indeed makes sense. However, the issue linked to the PR describes a bit different situation. It is likely that Artur's PR will fix the original issue rather than this one.

caalador:
Needs to be checked again then. At least when working on the ViteJs PWA I initially got index.html instead of sw.js for the service worker request, but the fetch dest change is newer then the initial project I guess.

Artur- (approved):
Safari 15.3 sends `Service-Worker: script` but no `Sec-Fetch-Dest` header at all so let's merge this

PR #12967, opened 2022-02-09
chore: Actually read the user key from the file

Artur-:


samie (dismissed):
Looks good. I think slightly changes the reporting in the case when writing of userKey failed. Previously reported a null key, but is now reported as a valid transient key. No problem with that.

Artur-:
reverted to reporting `null` again

samie (approved):

PR #12980, opened 2022-02-10
chore: Support also alternative source id form

Artur-:


vaadin-dev-server/src/test/resources/stats-data/gradle-project-folder2/settings.gradle
```diff
@@ -1,3 +1,3 @@
-// Vaadin project from https://start.vaadin.com/test/4
+// Project from https://start.vaadin.com/test/4
```
samie, line 1: Do we lose the previous test case?

samie (approved):

vaadin-dev-server/src/test/resources/stats-data/gradle-project-folder2/settings.gradle
Artur-: There is `gradle-project-folder1` which uses the other format. Same for Maven

PR #12993, opened 2022-02-11
chore: do not import modules of featured-flag components

manolo:
This adds a new optional field to  the `Feature` object so as we can specify the class name of an experimental component.
Scanners will skip this classes if the feature is disabled.
This avoids experimental components being imported in dev-mode bundle, or in production fall-back one.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
+++ b/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
@@ -39,7 +39,7 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Tracks available feature flags and their status.
- * 
+ *
  * Enabled feature flags are stored in
  * <code>vaadin-featureflags.properties</code> inside the resources folder
  * (<code>src/main/resources</code>).
@@ -51,13 +51,15 @@ public class FeatureFlags implements Serializable {
     public static final Feature EXAMPLE = new Feature(
             "Example feature. Will be removed once the first real feature flag is added",
             "exampleFeatureFlag", "https://github.com/vaadin/flow/pull/12004",
-            false);
+            false,
+            "com.vaadin.flow.server.frontend.NodeTestComponents$ExampleExperimentalComponent");
     public static final Feature VITE = new Feature(
             "Use Vite for faster front-end builds", "viteForFrontendBuild",
-            "https://github.com/vaadin/platform/issues/2448", true);
+            "https://github.com/vaadin/platform/issues/2448", true, null);
     public static final Feature MAP_COMPONENT = new Feature(
             "Map component (Pro)", "mapComponent",
-            "https://github.com/vaadin/platform/issues/2611", true);
+            "https://github.com/vaadin/platform/issues/2611", true,
+            "com.vaadin.flow.component.map.Map");
 
     private List<Feature> features = new ArrayList<>();
 
@@ -148,7 +150,11 @@ public class FeatureFlags implements Serializable {
         loadProperties();
     }
 
-    void loadProperties() {
+    /**
+     * Read the feature flag properties files and updates the enable property of
+     * each feature object.
+     */
+    public void loadProperties() {
         final ResourceProvider resourceProvider = lookup
                 .lookup(ResourceProvider.class);
         if (resourceProvider != null) {
@@ -195,11 +201,9 @@ public class FeatureFlags implements Serializable {
                 props.load(propertiesStream);
             }
             for (Feature f : features) {
-                boolean enabled = "true".equals(
-                        props.getProperty(getPropertyName(f.getId()), "false"));
-                f.setEnabled(enabled);
+                f.setEnabled(Boolean.valueOf(
+                        props.getProperty(getPropertyName(f.getId()))));
             }
-
         } catch (IOException e) {
             getLogger().error("Unable to read feature flags", e);
         }
@@ -233,7 +237,7 @@ public class FeatureFlags implements Serializable {
 
     /**
      * Get a list of all available features and their status.
-     * 
+     *
      * @return a list of all features
      */
     public List<Feature> getFeatures() {
@@ -242,7 +246,7 @@ public class FeatureFlags implements Serializable {
 
     /**
      * Checks if the given feature is enabled.
-     * 
+     *
      * @param feature
      *            the feature to check
      * @return <code>true</code> if enabled, <code>false</code> otherwise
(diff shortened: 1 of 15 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/ClassFinder.java
```diff
@@ -120,6 +121,18 @@ public URL getResource(String name) {
...
         }
+
+        @Override
+        public void setExcludeClassNames(List<String> classNames) {
+            excludedClasses = classNames;
+        }
+
+        @Override
+        public List<String> getExcludedClassNames() {
+            return excludedClasses != null ? excludedClasses
```
joheriks, line 132: Maybe wrap this with `Collections.unmodifiableList`

flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
```diff
@@ -51,13 +51,15 @@
     public static final Feature EXAMPLE = new Feature(
             "Example feature. Will be removed once the first real feature flag is added",
             "exampleFeatureFlag", "https://github.com/vaadin/flow/pull/12004",
-            false);
+            false,
+            "com.vaadin.flow.server.frontend.NodeTestComponents$ExampleExperimentalComponent");
```
joheriks, line 55: We probably should really remove this EXAMPLE feature, but could be in a separate PR.

manolo: yes it needs more work because it's used in tests. So it's worth a separated PR for mocking correctly the Feature

manolo: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/ClassFinder.java
manolo: done: storing it in the class as unmodifialble.

flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
```diff
@@ -195,11 +201,9 @@ void loadProperties(InputStream propertiesStream) {
                 props.load(propertiesStream);
             }
             for (Feature f : features) {
-                boolean enabled = "true".equals(
-                        props.getProperty(getPropertyName(f.getId()), "false"));
-                f.setEnabled(enabled);
+                f.setEnabled(Boolean.valueOf(
```
Artur-, line 204: How is this related to the change? Won't it break if the value is "foo" and the old one did not?

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
```diff
@@ -259,12 +259,15 @@ private String adaptCssValue(Annotation cssImport, String method) {
...
 
-            annotatedClasses.stream().forEach(clazz -> annotationFinder
-                    .apply(clazz, loadedAnnotation).forEach(ann -> {
-                        String value = valueExtractor.apply(ann);
-                        valueHandler.accept(clazz, value);
-                        logs.add(value + " " + clazz);
-                    }));
+            List<String> experimental = getFinder().getExcludedClassNames();
+            annotatedClasses.stream()
+                    .filter(c -> !experimental.contains(c.getName()))
```
Artur-, line 264: Why should this be `contains` and not `equals`?

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/ClassFinder.java
```diff
@@ -247,4 +273,8 @@ public ClassLoader getClassLoader() {
         Class<T> parent = loadClass(name);
         return getSubTypesOf(parent);
     }
+
+    void setExcludedClassNames(List<String> classNames);
```
Artur-, line 277: Looks to me like you are completely breaking the architecture here. A ClassFinder is immutable and you are changing this only to be able to use ClassFinder as a way to pass data from one class to another class. ClassFinder does not even use the data. Surely there is a better way?

Maybe it can be done in `FrontendDependenciesScannerFactory.createScanner`

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
manolo: experimental is a list, then I want to know whether the current class to be visited is in the list, or I don't understand what it's your suggestion. What should the code look like then?

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/ClassFinder.java
manolo: well, probably it's not the best way,  but I found this way simpler because class-finder is a utility class that is passed always in all implementations, I didn't consider that it should be inmutable. Tough I'll give try to the way you propose.

flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
manolo: It's unrelated to this PR, it just cause surprise to me not use the Boolean api to know whether a string is actually true or TRUE, everything else is false, like the previous code. So the only change here is that we now also accept uppercase in the true string.
"foo" always return false in previous code and use the Boolean API.
But I can revert this change if you are not confident with, not a big deal

manolo: It's unrelated to this PR, it just cause surprise to me not use the Boolean api to know whether a string is actually true or TRUE, everything else is false, like the previous code. So the only change here is that we now also accept uppercase in the true string.
"foo" always return false in previous code and use the Boolean API.
But I can revert this change if you are not confident with, not a big deal

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
Artur-: Looks like I misread the code and interpreted it as comparing the name

flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
manolo: added a 'foo' test

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/ClassFinder.java
manolo: reverted classFinder and moved logic to scanners,

Artur- (approved):

PR #12998, opened 2022-02-11
fix!: Supported node is updated to 16.14

### LINKED ISSUE #12995: Require node 16.14.0 or newer
### Describe your motivation

Node 16.14 is the first node version that ships with npm 8.3. 
npm 8.3 is important because it contains the "overrides" feature. 
Without this feature people will constantly run into issues where multiple versions of the same custom element are imported, there is a JS exception and the app does not start.

Require Node 16.14 as the minimum version. Autoinstall and use that version if an older is installed globally.

We currently install Node 17 which is by many considered an "experimental version" you shouldn't use although nodejs.org does not really put it like that. We should install 16 instead now that it has npm 8.3.

caalador:
Due to requirement of npm 8.3+ the supported
node version is updated to 16.14 and
the autoinstalled node is set to the current lts v16.14.0.

Also autoinstall an use .vaadin if node version is less than
the supported version 8.3

Fixes #12995

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -61,11 +61,11 @@ import static com.vaadin.flow.server.InitParameters.NODE_VERSION;
  */
 public class FrontendTools {
 
-    public static final String DEFAULT_NODE_VERSION = "v17.3.0";
+    public static final String DEFAULT_NODE_VERSION = "v16.14.0";
     /**
      * This is the version shipped with the default Node version.
      */
-    public static final String DEFAULT_NPM_VERSION = "8.3.0";
+    public static final String DEFAULT_NPM_VERSION = "8.3.1";
 
     public static final String DEFAULT_PNPM_VERSION = "5.18.10";
 
@@ -104,10 +104,10 @@ public class FrontendTools {
     private static final FrontendVersion WHITESPACE_ACCEPTING_NPM_VERSION = new FrontendVersion(
             7, 0);
 
-    private static final int SUPPORTED_NODE_MAJOR_VERSION = 12;
-    private static final int SUPPORTED_NODE_MINOR_VERSION = 22;
-    private static final int SUPPORTED_NPM_MAJOR_VERSION = 6;
-    private static final int SUPPORTED_NPM_MINOR_VERSION = 14;
+    private static final int SUPPORTED_NODE_MAJOR_VERSION = 16;
+    private static final int SUPPORTED_NODE_MINOR_VERSION = 14;
+    private static final int SUPPORTED_NPM_MAJOR_VERSION = 8;
+    private static final int SUPPORTED_NPM_MINOR_VERSION = 3;
 
     private static final FrontendVersion SUPPORTED_NODE_VERSION = new FrontendVersion(
             SUPPORTED_NODE_MAJOR_VERSION, SUPPORTED_NODE_MINOR_VERSION);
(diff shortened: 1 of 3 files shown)


flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
```diff
@@ -200,7 +206,7 @@ public void nodeIsBeingLocated_unsupportedNodeInstalled_fallbackToNodeInstalledT
         FrontendStubs.ToolStubInfo npmStub = FrontendStubs.ToolStubInfo.none();
         createStubNode(nodeStub, npmStub, baseDir);
 
-        tools.installNode("v13.10.1", null);
+        tools.installNode("v16.14.10", null);
```
Artur-, line 209:
```suggestion
        tools.installNode("v16.14.0", null);
```

Artur-:
According to random internet posts, Node has not tested or supported Windows 7 since node 13.

Artur-:
If somebody is still on Windows 7 and developing a modern web app, now is a very good time to upgrade

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -1005,6 +1007,26 @@ private String getNonNull(String... valueArray) {
...
+                        if (npmVersion.isOlderThan(SUPPORTED_NPM_VERSION)) {
+                            getLogger().warn(
+                                    "Global npm is older than {}. Using npm form .vaadin.",
+                                    SUPPORTED_NPM_VERSION.getFullVersion());
+                            returnCommand = new ArrayList<>();
+                            // Force installation if not installed
+                            getNodeBinary();
+                        }
+                    } catch (UnknownVersionException uve) {
+                        getLogger().error("Could not determine npm version",
```
Artur-, line 1026: If version detection fails, does this then install and use a known working version or will it use the unknown version?

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendToolsTest.java
```diff
@@ -145,22 +148,24 @@ public void nodeIsBeingLocated_updateTooOldNode_NodeInstalledToTargetDirectoryIs
     }
 
     @Test
+    // Test will test correctly when supported node is newer than 16.14
```
Artur-, line 151: What does this comment really mean? It will hopefully be far in the future when we change the minimum supported Node version

caalador: It means that in this instance where the minimum supported version is 16.14 and the default install is the newest LTS which is 16.14 that the test doesn't validate anything until we have a install version > 16.14 as we can not determine if we stayed at the installed version if the version to autoupdate to is the same as the installed version. Same goes for the other autoupdate expectations.

Artur-: So it will make sense when there is a newer Node than 16.14 available and we install that, not when the supported node version changes?

caalador: Yes. So the test(s) will validate something smart when the installed version is for instance 16.15

Artur-: Alright. I would just remove the comments as suspect they will confuse somebody and nobody will remember to ever remove them. Otherwise it should say when "... the auto downloaded node is newer..." or something

caalador: Removed

Artur- (approved):

PR #12999, opened 2022-02-11
chore: fix how disabledPnpm flag was set for TaskUpdateImports

manolo:
When passing `enablePnpm` flag to `TaskUpdateImports` it must be negated since flag is the oposite in this class.


manolo:
@Artur- done

Artur- (dismissed):

Artur-:
Aren't a lot of tests incorrect now?

Artur-:
Seems like none of the tests actually care what the flag is set as and usage of the flag is not tested at all

Artur- (approved):

PR #13019, opened 2022-02-14
feat: add an offline property to the PWA annotation

### LINKED ISSUE #6548: Ability to control service worker installation in @PWA
It will be nice to have an option to opt out service worker installation:

```java
@PWA(installServiceWorker = false)
```

vursen:
## Description

This PR adds an `offline` property to the `PWA` annotation that defaults to `true`. The property specifies whether the application should be available offline, which technically comes down to whether the service worker should be enabled.

**`@PWA(offline = true)`:**
- The service worker build is enabled.
- The service worker will be registered on the user's first visit.

**`@PWA(offline = false)`:**
- The service worker build is disabled.
- The active service worker (if one is running in the browser) will be unregistered on the next user's visit.

Part of #12383

Fixes #6548

## Type of change

- [x] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/server/PWA.java
```diff
@@ -168,4 +168,14 @@
...
+
+    /**
+     * Whether to disable the service worker.
+     *
+     * Setting this property to {@code true} will cause the active service
+     * worker to unregister.
+     *
+     * @return whether to disable the service worker.
+     */
+    boolean serviceWorkerDisabled() default false;
```
Artur-, line 180: A non-negated option makes more sense, could it even be `serviceWorker` with a default value of `true`

vursen: > The other question would be if it really is the service worker you want to disable or a set of features, e.g. offline

Good point, `offline` sounds more appropriate as the service worker could be considered just an implementation detail of the whole offline feature. This way it would be also in line with the existing `offlinePath`, `offlineResources` properties.

vursen: Done.

platosha (dismissed):

platosha:
LGTM @vursen great job! 👍

platosha (approved):

Issue #12198, opened 2021-10-29
Nullpointer related to Shortcuts


probert94:
### Description of the bug / feature
In a project I am working on, I am facing an issue related to the `ShortcutRegistration` in vaadin.
Basically, when I execute a `Shortcut` while a new `ShortcutListener` is getting registered, I get the following exception:
```
java.lang.NullPointerException
		at elemental.json.impl.JreJsonObject.getBoolean(JreJsonObject.java:92)
		at com.vaadin.flow.internal.nodefeature.ElementListenerMap$DomEventListenerWrapper.matchesFilter(ElementListenerMap.java:210)
		at com.vaadin.flow.internal.nodefeature.ElementListenerMap.fireEvent(ElementListenerMap.java:437)
		at com.vaadin.flow.server.communication.rpc.EventRpcHandler.handleNode(EventRpcHandler.java:62)
		at com.vaadin.flow.server.communication.rpc.AbstractRpcInvocationHandler.handle(AbstractRpcInvocationHandler.java:66)
		at com.vaadin.flow.server.communication.ServerRpcHandler.handleInvocationData(ServerRpcHandler.java:412)
		at com.vaadin.flow.server.communication.ServerRpcHandler.lambda$handleInvocations$1(ServerRpcHandler.java:393)
		at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
		at com.vaadin.flow.server.communication.ServerRpcHandler.handleInvocations(ServerRpcHandler.java:393)
		at com.vaadin.flow.server.communication.ServerRpcHandler.handleRpc(ServerRpcHandler.java:320)
		at com.vaadin.flow.server.communication.UidlRequestHandler.synchronizedHandleRequest(UidlRequestHandler.java:115)
		at com.vaadin.flow.server.SynchronizedRequestHandler.handleRequest(SynchronizedRequestHandler.java:40)
		at com.vaadin.flow.server.VaadinService.handleRequest(VaadinService.java:1535)
		at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:299)
```

In my case, I have a `F2` shortcut, which opens the selected object in a `Dialog`. The `Dialog` itself registers a `ctrl+s` shortcut to save changes on the current object. When I double-press the `F2` key, I get the mentioned exception.
Basically `JreJsonObject.getBoolean` throws a `NullPointerException`, because the given key does not exist in the `JsonObject`.
[...]

PR #13021, opened 2022-02-14
fix: Get boolean only if eventData has the key

TatuLund:
Fixes: https://github.com/vaadin/flow/issues/12198


mcollovati (commented):
Does it make sense to add a test in `ElementListenersTest` for this change? Something like

```java
AtomicInteger eventCount = new AtomicInteger();
DomListenerRegistration registration = ns.add("foo", e -> eventCount.incrementAndGet());
registration.setFilter("filterKey");

ns.fireEvent(createEvent("foo"));
Assert.assertEquals(0, eventCount.get());

JsonObject eventData = Json.createObject();
eventData.put("filterKey", true);
ns.fireEvent(new DomEvent(new Element("element"), "foo", eventData));
Assert.assertEquals(1, eventCount.get());
```

mcollovati (approved):

PR #13022, opened 2022-02-15
fix: Update jaxb-runtime

caalador:
Update jaxb-runtime to a version
compatible with jdk17

Fixes vaadin/spring#981

### DIFF
--- a/flow-tests/vaadin-spring-tests/pom.xml
+++ b/flow-tests/vaadin-spring-tests/pom.xml
@@ -248,6 +248,28 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
+        <!-- JDK 9+ dependencies -->
+        <dependency>
+            <groupId>javax.activation</groupId>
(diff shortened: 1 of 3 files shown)


Artur- (approved):
Looks fine but are these now the same deps as used elsewhere in the project?

caalador:
Yes the versions match inside the project.

flow-tests/vaadin-spring-tests/pom.xml
```diff
@@ -502,7 +502,7 @@
         <profile>
             <id>java 11</id>
             <activation>
-                <jdk>11</jdk>
+                <jdk>[9,)</jdk>
```
Artur-, line 505: This profile could be removed next.. it will always be active as we only support Java 11+.

caalador: done

Artur- (approved):

Artur- (approved):

PR #13026, opened 2022-02-15
feat: Add modality helpers

### LINKED ISSUE #12992: Make it possible to check server-side modality status
### Describe your motivation

For better handling of server-side modality with components like Notification/dialog etc. it would be required to be able to check the current active modal component (if any) so that these components can be added to the modal tree so that they are active.

UI could have the method `Optional<Component> getModalComponent()` that would return the current top most modal component.

caalador:
Add methods to check if there is a
modal component set and to add
component to current modal component.

Fixes #12992

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -1291,4 +1291,38 @@ public class UI extends Component
         }
     }
 
+    /**
+     * Check if UI has a defined modal component.
+     * 
+     * @return {@code true} if a modal component has been set
+     */
+    public boolean hasModal() {
+        return getInternals().hasModalComponent();
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1291,4 +1291,38 @@ public void setChildComponentModal(Component childComponent,
         }
     }
 
+    /**
+     * Check if UI has a defined modal component.
+     * 
+     * @return {@code true} if a modal component has been set
+     */
+    public boolean hasModal() {
```
joheriks, line 1299: Maybe this could also be called `hasModalComponent` to give a bit more context as UI contains a lot of things, and to be consistent with `addToModalComponent`.

caalador: done

anssit (approved):

PR #13032, opened 2022-02-15
fix: prevent using @fs URLs in Vite

### LINKED ISSUE #12957: Files should not be loaded from '@fs' URLs when using Vite
### Description of the bug

A new default project from start.vaadin.com loads the following files on startup when using Vite

```
<script type="module" src="/VAADIN/generated/vite-devmode.ts"></script> 
<script type="module" src="/VAADIN/generated/vaadin.ts"></script> 
```
This is fine but then those files start loading other files using the `/@fs/` URL, e.g. 
```js
import '/VAADIN/@fs/.../vite-v23beta/node_modules/vite/dist/client/env.mjs';
```
This looks like some kind of misconfiguration of Vite.

### Expected behavior

`/@fs` URLs should not be used for normal operations

https://start.vaadin.com
Select Vaadin 23
Enable Vite
Run and look at the network tab in the browser dev tools

### Versions

- Vaadin / Flow version: 23 beta 1

platosha:
## Description

In Vite development mode, serves the application so that:
- `/VAADIN/frontend/` maps to the sources from the frontend directory
- `/VAADIN/node_modules/` maps to the `node_modules` installed

This configuration works better for Vite, as it needs to serve the `node_modules/` directory.

After this change, Vite does not use the `@fs` fallback URL mapping with absolute paths.

Fixes #12957

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -23,6 +23,8 @@ const themeFolder = path.resolve(frontendFolder, settings.themeFolder);
 const frontendBundleFolder = path.resolve(__dirname, settings.frontendBundleOutput);
 const addonFrontendFolder = path.resolve(__dirname, settings.addonFrontendFolder);
 
+const frontendPath = path.posix.relative(__dirname, settings.frontendFolder) + '/';
+
 const projectStaticAssetsFolders = [
   path.resolve(__dirname, 'src', 'main', 'resources', 'META-INF', 'resources'),
   path.resolve(__dirname, 'src', 'main', 'resources', 'static'),
@@ -162,9 +164,10 @@ export const vaadinConfig: UserConfigFn = (env) => {
     runWatchDog(process.env.watchDogPort);
   }
   return {
-    root: 'frontend',
+    root: devMode ? '' : frontendFolder,
     base: '',
     resolve: {
+      preserveSymlinks: true,
       alias: {
         themes: themeFolder,
         Frontend: frontendFolder
@@ -220,7 +223,7 @@ export const vaadinConfig: UserConfigFn = (env) => {
         transformIndexHtml: {
           enforce: 'pre',
           transform(_html, { path, server }) {
-            if (path !== '/index.html') {
+            if (path !== `/${devMode ? frontendPath : ''}index.html`) {
               return;
             }
 
@@ -229,12 +232,12 @@ export const vaadinConfig: UserConfigFn = (env) => {
               return [
                 {
                   tag: 'script',
-                  attrs: { type: 'module', src: `${basePath}generated/vite-devmode.ts` },
+                  attrs: { type: 'module', src: `${basePath}${frontendPath}generated/vite-devmode.ts` },
                   injectTo: 'head',
                 },
                 {
                   tag: 'script',
-                  attrs: { type: 'module', src: `${basePath}generated/vaadin.ts` },
+                  attrs: { type: 'module', src: `${basePath}${frontendPath}generated/vaadin.ts` },
                   injectTo: 'head',
                 }
               ]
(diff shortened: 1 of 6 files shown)


flow-tests/test-frontend/vite-basics/src/test/java/com/vaadin/viteapp/BasicsIT.java
```diff
@@ -49,4 +50,17 @@ public void canImportJson() {
         Assert.assertEquals("{\"hello\":\"World\"}",
                 $("*").id(MainView.JSON_CONTAINER).getText());
     }
+
+    @Test
+    public void hasNeatNodeModulesPath() {
+        final String processedJs = (String) ((JavascriptExecutor) getDriver())
+                .executeAsyncScript(
+                        "const done = arguments[arguments.length - 1];"
+                                + "fetch('frontend/deploader.js', {"
```
vursen, line 59: I wonder if the path needs a `VAADIN` prefix? Does it work even without that prefix?

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -23,6 +23,8 @@ const themeFolder = path.resolve(frontendFolder, settings.themeFolder);
 const frontendBundleFolder = path.resolve(__dirname, settings.frontendBundleOutput);
 const addonFrontendFolder = path.resolve(__dirname, settings.addonFrontendFolder);
 
+const frontendPath = path.posix.relative(__dirname, settings.frontendFolder) + '/';
```
Artur-, line 26: This could use a comment about what the difference is between `frontendFolder` and `frontendPath`

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/ViteHandler.java
```diff
@@ -67,6 +69,11 @@
     public ViteHandler(Lookup lookup, int runningPort, File npmFolder,
             CompletableFuture<Void> waitFor) {
         super(lookup, runningPort, npmFolder, waitFor);
+        frontendPath = FrontendUtils.getUnixRelativePath(npmFolder.toPath(),
+                new File(getApplicationConfiguration().getStringProperty(
+                        FrontendUtils.PARAM_FRONTEND_DIR,
+                        System.getProperty(FrontendUtils.PARAM_FRONTEND_DIR,
```
vursen, line 75: Is it necessary to default the application's property to the system's property? I thought it works the way that the application configuration already takes into account the system's property, doesn't it?

vursen: Otherwise, it can be so that setting the system's property will only take effect in this particular place which is not desired, I guess.

Artur-: Yeah, `System.getProperty` shouldn't be needed here

flow-tests/test-frontend/vite-basics/src/test/java/com/vaadin/viteapp/BasicsIT.java
platosha: no, added it.

flow-tests/test-frontend/vite-basics/frontend/deploader.js
```diff
@@ -0,0 +1 @@
+import('@vaadin/flow-frontend/package.json');
```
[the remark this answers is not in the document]
platosha: @sonatype-lift ignore

platosha:
@Artur- right, I understand keeping URLs the same this is more important than fixing the issue. Vite, as it seems, does not allow to have both easily at this point.

Closing this until some implementation better ideas come up.

PR #13039, opened 2022-02-16
feat: use Vaadin component container bundles in Vite dev mode

### LINKED ISSUE #12941: Use Vaadin components from bundles in Vite dev mode
Add a plugin in Vite development server generated config to use Vaadin components and other dependencies from bundles.

See also #12940 the bundles package ticket.

- detect if `@vaadin/bundles` npm package is installed
- load the list of packages / exposes from `vaadin-bundle.json`
- there is a loader for known modules matched from the list, that uses a stub that re-exports names from the bundle:

```js
import { init as VaadinBundleInit, get as VaadinBundleGet } from '/VAADIN/node_modules/@vaadin/bundles/vaadin.js?v=d001b352';
await VaadinBundleInit('default');
const { Button } = (await VaadinBundleGet('./node_modules/@vaadin/button/src/vaadin-button.js'))();
export { Button };
```

DoD:
- implementation
- IT test
- startup performance test

### LINKED ISSUE #12942: Opt-out from using bundles in Vite dev mode when a version mismatch is detected
The Vite plugin using the Vaadin component bundles should check if the bundled package versions (from the `vaadin-bundle.json` list) match the installed package versions.

When a mismatch is detected:
- the corresponding bundle are disabled
- print a warning in the development server output
- keep the Vite pre-bundling enabled for the packages from the disabled bundle

The check is done during startup of the development server.

This is intended to allow for dependency package overrides in the application.

DoD:
- implementation
- IT test

platosha:
Fixes #12941
Fixes #12942


Artur-:
Tested in Start with the following issue:
```
Uncaught SyntaxError: The requested module '/VAADIN/@fs/.../node_modules/@vaadin/vaadin-lumo-styles/all-imports.js?v=51a2d3b5' does not provide an export named 'badge'
```

originates from
```
import { badge } from '@vaadin/vaadin-lumo-styles';
```
in the generated theme js

and http://localhost:8080/VAADIN/@fs/.../node_modules/@vaadin/vaadin-lumo-styles/all-imports.js?v=51a2d3b5 contains 
```
import { init as VaadinBundleInit, get as VaadinBundleGet } from '/VAADIN/@fs/Users/artur/workspace/vaadin/start/node_modules/@vaadin/bundles/vaadin.js?v=51a2d3b5';
await VaadinBundleInit('default');
const {  } = (await VaadinBundleGet('./node_modules/@vaadin/vaadin-lumo-styles/all-imports.js'))();
export {  };
```

Likely a problem with the `@vaadin/bundles` `v23.0.0-beta2` as `vaadin-bundle.json` does not contain `badge.js` and it contains
```
        "./all-imports.js": {
          "exports": []
        },
```

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -115,6 +128,166 @@ function injectManifestToSWPlugin(): PluginOption {
...
+    apply(config, {command}) {
+      if (command !== "serve") return false;
+
+      resolveVaadinBundleJsonPath();
+      if (vaadinBundleJsonPath === undefined) {
+        console.info(`@vaadin/bundles npm package is not found, ${disabledMessage}`);
+        vaadinBundleJson = {packages: {}};
+        return false;
+      }
+
```
Artur-, line 221: Is there a reason that `vaadinBundleJson` is not read and parsed here? Makes it a bit more difficult to follow the code

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -115,6 +128,166 @@ function injectManifestToSWPlugin(): PluginOption {
...
+            // Happen to fail dedupe in Vite pre-bundling
+            '@polymer/iron-meta',
+            '@polymer/iron-meta/iron-meta.js',
+          ],
+          exclude: [
+            // Vaadin bundle
+            '@vaadin/bundles',
+            ...Object.keys(vaadinBundleJson.packages),
+            // Known small packages
+            '@vaadin/router',
```
Artur-, line 267: Why should this and nothing else be excluded? It looks very random here

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -115,6 +128,166 @@ function injectManifestToSWPlugin(): PluginOption {
...
+        console.info(`Packages with version mismatches: ${JSON.stringify(versionMismatches, undefined, 2)}`);
+        vaadinBundleJson = {packages: {}};
+        return config;
+      }
+
+      return mergeConfig(config, {
+        optimizeDeps: {
+          include: [
+            // Happen to fail dedupe in Vite pre-bundling
+            '@polymer/iron-meta',
```
Artur-, line 259: What does this really mean? Is it a Vite bug? Something else?

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -115,6 +128,166 @@ function injectManifestToSWPlugin(): PluginOption {
...
+            sourceExports.forEach((e) => exportsSet.add(e));
+          }
+        }
+      }
+    }
+    return Array.from(exportsSet);
+  }
+
+  return {
+    name: 'vaadin-bundle',
```
Artur-, line 210: If this would happen to not work in my project, is there a way to disable it until the problem has been fixed? Can I remove the bundles plugin in my vite.config.ts?

Artur- (changes requested):

flow-server/src/main/resources/vite.generated.ts
platosha: Doing so requires an ugly workaround to filter-out the plugin as of now.

Unfortunately, Vite config structure (and Rollup in general) does not seem to be friendly to passing options down to custom plugins defined elsewhere.

Some options on my mind:
- Add a custom Vaadin options argument in `overrideVaadinConfig` and add expose an option there. Would probably be most user-friendly, with TypeScript typings.
- Release a special version tag to use with the npm package version overrides: `npm install --save-dev @vaadin/bundles#optout` installs an empty package, then the plugin disables itself.

Artur-: For now I think an ugly workaround is enough

flow-server/src/main/resources/vite.generated.ts
platosha: This was fixed in the Vaadin bundles build. Removed those.

flow-server/src/main/resources/vite.generated.ts
platosha: Yes, however it’s a bit quirky to filter-out a plugin from the generated config list. We should probably provide a more friendly way than that. How about:

- An argument in the `overrideVaadinConfig` with a custom option exposed to disable Vaadin bundles
- A special version tag for the npm package, so that when you depend on `@vaadin/bundles#optout`, then it installs an empty stub and disables the plugin.

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -187,8 +363,15 @@ export const vaadinConfig: UserConfigFn = (env) => {
         }
       }
     },
+    optimizeDeps: {
+      entries: [
+        // Pre-scan entrypoints in Vite to avoid reloading on first open
+        'generated/vaadin.ts'
```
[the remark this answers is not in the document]
platosha: Does not make, I think. It has no dependencies, nothing to optimize there.

flow-server/src/main/resources/vite.generated.ts
platosha: I used async file IOs for that, but now thinking more, probably sync ones would work just fine. Done.

flow-server/src/main/resources/vite.generated.ts
platosha: Moved into the general config section, does not make sense in the plugin context indeed.

Still makes sense to exclude the router in general, I think, it is rolled up already.

Artur-:
Something broke `BundlesIT` in the last set of changes

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -4,17 +4,30 @@
  * This file has been generated by the `flow:prepare-frontend` maven goal.
  * This file will be overwritten on every run. Any custom changes should be made to vite.config.ts
  */
-import path from 'path';
-import * as net from 'net';
+import path
```
vursen, line 7: The formatting is broken.

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -115,6 +128,152 @@ function injectManifestToSWPlugin(): PluginOption {
...
+    packages: Record<string, PackageInfo>,
+  };
+
+  const disabledMessage = 'Vaadin component dependency bundles are disabled.'
+
+  const modulesDirectory = path.posix.resolve(__dirname, 'node_modules');
+
+  let vaadinBundleJson: BundleJson;
+
+  function resolveVaadinBundleJsonPath() {
```
vursen, line 156: **nit:** Empty function.

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -168,7 +327,8 @@ export const vaadinConfig: UserConfigFn = (env) => {
       alias: {
         themes: themeFolder,
         Frontend: frontendFolder
-      }
+      },
+      preserveSymlinks: true,
```
vursen, line 331: Is it necessary after all?

platosha: it simplifies the loader, currently it just matches the path after `./node_modules` against the bundle JSON.

Artur- (approved):

PR #13053, opened 2022-02-17
fix: respect PWA offline property in the Vite config

vursen:
## Description

This PR fixes the Vite config so that it would respect the `offline` property of the `PWA` annotation and disable the service worker build when `offline == false`.

Seems like these changes have got lost after a rebase in the original PR introducing the `offline` property.

A follow-up to #13019

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -189,8 +189,8 @@ export const vaadinConfig: UserConfigFn = (env) => {
     },
     plugins: [
       !devMode && brotli(),
-      settings.pwaEnabled && transpileSWPlugin(),
-      settings.pwaEnabled && injectManifestToSWPlugin(),
+      settings.offlineEnabled && transpileSWPlugin(),
+      settings.offlineEnabled && injectManifestToSWPlugin(),
       {
         name: 'vaadin:custom-theme',
         config() {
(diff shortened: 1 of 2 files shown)


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -189,8 +189,8 @@ export const vaadinConfig: UserConfigFn = (env) => {
     },
     plugins: [
       !devMode && brotli(),
-      settings.pwaEnabled && transpileSWPlugin(),
-      settings.pwaEnabled && injectManifestToSWPlugin(),
+      settings.offlineEnabled && transpileSWPlugin(),
```
Artur-, line 192: When is pwa disabled? Can offline be enabled somehow even though pwa is disabled?

vursen: PWA is disabled when there is no PWA annotation. When there is no PWA annotation, `offlineEnabled` is also `false`.

https://github.com/vaadin/flow/blob/1a859755f98683b9de21b462bc3e62e65f72977a/flow-server/src/main/java/com/vaadin/flow/server/PwaConfiguration.java#L69-L74

Artur- (approved):

Artur-:
Ah this one shouldn't go into 23.0 because the `offline` flag didn't go there, right?

PR #13055, opened 2022-02-17
chore: Record npm install time

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -249,6 +252,7 @@ private boolean isVaadinHashUpdated() {
      * `package.json` has been updated.
      */
     private void runNpmInstall() throws ExecutionFailedException {
+        long startTime = System.currentTimeMillis();
         // Do possible cleaning before generating any new files.
         cleanUp();
```
caalador, line 257: cleanUp might take some time if node_modules is deemed to be removed due to npm->pnpm or pnpm->npm change.
Should it really be taken into account for (p)npm install time?

Artur-: We should definitely record the full time that we need for getting the frontend in shape. But maybe we should record it separately then?

Artur-: Fixed

caalador: You forgot to push the StatisticsConstants with EVENT_PACKAGEMANAGER_CLEANUP_TIME_PREFIX

Artur-: I indeed did..

Issue #13008, opened 2022-02-14
Stop using shrinkwrap


Artur-:
### Describe your motivation

The shrinkwrap does not serve its intended purpose of locking dependency versions. Now we have a better mechanism through npm `overrides` and that one actually works. We should not include shrinkwrap in package.json anymore. 

Including the shrinkwrap actually breaks npm installation because of an npm problem: https://github.com/npm/cli/issues/4409

### Describe the solution you'd like

Do not install shrinkwrap. Do not use it for anything. This can be extended to: do not publish a shrinkwrap

joheriks:
### Acceptance criteria
1. Remove `@NpmPackage` annotation from https://github.com/vaadin/platform/blob/92d227cd26319b9b533cd9053391ee9982004a5d/vaadin-core/src/main/templates/com/vaadin/shrinkwrap/VaadinCoreShrinkWrap.java
2. We currently check the version of `vaadin-shrinkwrap` to figure out if the platform has been updated, this could check `vaadin-core` package version instead.
3. Verified that `package.json` does not contain `vaadin-shrinkwrap` nor `vaadin-core-shrinkwrap`.
4. Verified that when upgrading the platform then the cleanup step in `TaskUpdatePackages`(deleting node_modules, etc.) is executed.

manolo:
We need to remove/skip steps during platform release

ZheSun88:
except the vaadinCoreShrinkWrap.java, there is also the VaadinShrinkWrap.java https://github.com/vaadin/platform/blob/master/vaadin/src/main/templates/com/vaadin/shrinkwrap/VaadinShrinkWrap.java

Artur-:
https://github.com/vaadin/platform/pull/2751

anantjagania:
Can we have this removed from Vaadin 14 as well?

Thanks.

PR #13059, opened 2022-02-17
fix: save platform version in node_modules

joheriks:
To decide whether additional cleanup is needed, store the 
platform version in `node_modules/.vaadin/vaadin.json`, then 
compare with the platform published version. Was previously 
checking `vaadin-shrinkwrap` in `package.json`, but this 
package is removed in V23.

Part of #13008

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
@@ -46,7 +46,7 @@ public class Platform implements Serializable {
                 return Optional.of(vaadinVersions.get("platform").asText());
             } else {
                 if (!versionErrorLogged) {
-                    versionErrorLogged = true; // NOSONAR
+                    versionErrorLogged = true;
                     LoggerFactory.getLogger(Platform.class)
                             .info("Unable to determine version information. "
                                     + "No vaadin_versions.json found");
--- a/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.server;
+
+import java.io.InputStream;
+import java.io.Serializable;
+import java.util.Optional;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Provides information about the current version of Vaadin Platform.
+ *
+ * @since 23.0
+ */
+public class Platform implements Serializable {
+
+    private static boolean versionErrorLogged = false;
+
+    /**
+     * Returns the platform version string, e.g., {@code "23.0.0"}.
+     *
+     * @return the platform version or {@link Optional#empty()} if unavailable.
+     */
+    public static Optional<String> getVaadinVersion() {
+        try (InputStream vaadinVersionsStream = Platform.class.getClassLoader()
+                .getResourceAsStream(Constants.VAADIN_VERSIONS_JSON)) {
+            if (vaadinVersionsStream != null) {
+                ObjectMapper m = new ObjectMapper();
+                JsonNode vaadinVersions = m.readTree(vaadinVersionsStream);
+                return Optional.of(vaadinVersions.get("platform").asText());
+            } else {
+                if (!versionErrorLogged) {
+                    versionErrorLogged = true; // NOSONAR
+                    LoggerFactory.getLogger(Platform.class)
+                            .info("Unable to determine version information. "
+                                    + "No vaadin_versions.json found");
+                }
+            }
+        } catch (Exception e) {
+            LoggerFactory.getLogger(Platform.class)
+                    .error("Unable to determine version information", e);
+        }
+
+        return Optional.empty();
+    }
+}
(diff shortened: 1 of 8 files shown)


caalador (approved):

Artur-:
What is `@vaadin/vaadin-core` used for as you do not typically ever have that as a dependency?

Artur-:
If it is about knowing which version installed `node_modules`, why don't we write the Flow version into `node_modules/.vaadin/vaadin.json` e.g.
```
{
  "hash": "d4e92e3e993a8edd30596cf4d06dbc632cafe53573253c110f1c97fe26ce7120",
  "flowVersion": "23.0.1"
}
```
or the Vaadin version if that is the relevant one?

caalador:
`@vaadin/vaadin-core` is always available as it is brought in by `@vaadin/vaadin` or if you depend on vaadin-core for the free components 
Also the feature is explicitly for when the platform version has updated not flow so we shouldn't depend on flow as it may stay on the same patch version if platform releases for only components changes.

Artur-:
Nobody is depending on `@vaadin/vaadin` either

Artur-:
What is special about upgrading a platform version compared to updating anything by hand in package.json?

caalador:
Humm... What project was I looking at because I did always get the vaadin packages in that one.

The main reason for this has been that upgrading platform version broke the node_modules more often than acceptable.

joheriks:
Next attempt will be to add platform version to `node_modules/.vaadin/vaadin.json` and compare against that then.

joheriks:
Now doing it like that, ready for review again.

flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractNodeUpdatePackagesTest.java
```diff
@@ -372,41 +299,77 @@ public void oldVaadinDevDependency_missmatchWithDevDependency_vaadinDependencyIs
...
+            Assert.assertTrue(mainNodeModules.exists());
+            Assert.assertTrue(appNodeModules.exists());
+            // package-lock is removed
+            Assert.assertFalse(packageLock.exists());
+        }
     }
 
     @Test
     public void versionsMatch_noCleanUp() throws IOException {
+        // TODO: Fixme
```
caalador, line 372: I feel this test is not needed as it now stands

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/Platform.java
```diff
@@ -0,0 +1,62 @@
...
+    public static Optional<String> getVaadinVersion() {
+        try (InputStream vaadinVersionsStream = Platform.class.getClassLoader()
+                .getResourceAsStream(Constants.VAADIN_VERSIONS_JSON)) {
+            if (vaadinVersionsStream != null) {
+                ObjectMapper m = new ObjectMapper();
+                JsonNode vaadinVersions = m.readTree(vaadinVersionsStream);
+                return Optional.of(vaadinVersions.get("platform").asText());
+            } else {
+                if (!versionErrorLogged) {
+                    versionErrorLogged = true; // NOSONAR
```
Artur-, line 49: nit: NOSONAR should no longer needed here as the method is static

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -365,29 +364,30 @@ private int updateNpmLocalDependency(JsonObject dependenciesObject,
...
 
     /**
-     * Compares vaadin-core dependency version (which is the same as the
-     * platform version) from the {@code dependencies} object with the current
-     * vaadin-core version (retrieved from file system: package.json,
-     * package-lock.json). In case there was no existing vaadin-core version,
-     * then version is considered updated.
+     * Compares current platform version with the one last recorded as installed
+     * in node_modules/.vaadin/vaadin_version. In case there was no existing
+     * vaadin-core version, then version is considered updated.
```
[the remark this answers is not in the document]
joheriks: Good catch, the above code would actually delete `package-lock.json` on initial install which slows things down. Modified to return false when `node_modules` is absent.

flow-server/src/main/java/com/vaadin/flow/server/Platform.java
joheriks: Done.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -224,22 +221,16 @@ private boolean shouldRunNpmInstall() {
...
-                            .getPackageJson();
-                    return !content.getString(HASH_KEY).equals(packageJson
-                            .getObject(VAADIN_DEP_KEY).getString(HASH_KEY));
-                }
-            } catch (IOException e) {
-                packageUpdater.log()
-                        .warn("Failed to load hashes forcing npm execution", e);
+        try {
+            JsonObject content = packageUpdater.getVaadinJsonContents();
+            if (content != null && content.hasKey(HASH_KEY)) {
```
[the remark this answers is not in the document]
joheriks: Done.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -575,6 +583,32 @@ String writePackageFile(JsonObject json, File packageFile)
...
+    }
+
+    JsonObject getVaadinJsonContents() throws IOException {
+        File vaadinJsonFile = getVaadinJsonFile();
+        if (vaadinJsonFile.exists()) {
+            String fileContent = FileUtils.readFileToString(vaadinJsonFile,
+                    UTF_8.name());
+            return Json.parse(fileContent);
+        } else {
+            return null;
```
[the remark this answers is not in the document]
joheriks: Done.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -368,29 +363,31 @@ private int updateNpmLocalDependency(JsonObject dependenciesObject,
...
-            shrinkWrapVersion = dependencies.getString(SHRINK_WRAP);
+    private boolean isPlatformVersionUpdated() throws IOException {
+        // if no record of current version is present, version is not
+        // considered updated
+        Optional<String> platformVersion = Platform.getVaadinVersion();
+        if (platformVersion.isPresent() && nodeModulesFolder.exists()) {
+            JsonObject vaadinJsonContents = getVaadinJsonContents();
+            // If no record of previous version, version is considered updated;
+            if (vaadinJsonContents == null
+                    || !vaadinJsonContents.hasKey(NodeUpdater.VAADIN_VERSION)) {
```
caalador, line 383:
```suggestion
            if (!vaadinJsonContents.hasKey(NodeUpdater.VAADIN_VERSION)) {
```
Null check is not needed as we always get a JsonObject.

joheriks: Thanks, missed it. Done.

caalador (approved):

joheriks:
Picked in #13099

PR #13106, opened 2022-02-21
chore: Fix code smells

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -56,6 +56,45 @@
...
 public class TaskRunNpmInstall implements FallibleCommand {
 
+    /** Container for npm installation statistics. */
+    public static class Stats {
+        private long installTimeMs = 0;
+        private long cleanupTimeMs = 0;
+        private String packageManager = "";
+
+        /** Create an instance. */
+        public Stats() {
```
mcollovati, line 66: minor: maybe we can restrict constructor visibility, since we instantiate this class only in the enclosing on

Artur-: Yes, it can be private. Done

mcollovati (approved):

Issue #11282, opened 2021-06-18
Reduce the number of startup log messages


tarekoraby:
In v20+, running a Vaadin app produces a rather long list of startup log messages: more than 140 lines, most of which are attributed to the frontend compilation phase.  This IMO detracts from how a new evaluator might view Vaadin as a large number of log messages would leave the impression that Vaadin is a complex thing (and while it might indeed be a complex machine, the value in using it is that we are hiding the complexity under the hood so that the developer would focus on solving their use cases).

For a better DX, please consider reducing the number of default startup messages to the essentials (and allow enabling a verbose log only with a flag).

Related Slack discussion: https://vaadin.slack.com/archives/C6RAXJATF/p1623920841274400

pleku:
It would make sense to just utilize the existing log levels more to so that e.g. webpack compilation would be output for log level `DEBUG` but any error output would be always output. But it needs work.

tarekoraby:
#11783 introduced this [line](https://github.com/vaadin/flow/blob/e43db2a989407f6d8501f2b9e1664f9f0183a6b3/flow-server/src/main/resources/webpack.generated.js#L232) that improves the situation considerably by reducing the clutter.

PR #13129, opened 2022-02-23
Only log npm error messages by default

tarekoraby:
The typical Flow user doesn't need to see npm warnings in the log. Warnings like:

```
npm WARN deprecated source-map-url@0.4.1: See https://github.com/lydell/source-map-url#deprecated
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated source-map-resolve@0.5.3: See https://github.com/lydell/source-map-resolve#deprecated
npm WARN deprecated chokidar@2.1.8: Chokidar 2 does not receive security updates since 2019. Upgrade to chokidar 3 with 15x fewer dependencies
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
npm WARN deprecated core-js@2.6.12: core-js@<3.4 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Please, upgrade your dependencies to the actual version of core-js.
```

### DIFF
--- a/flow-server/src/main/resources/npmrc
+++ b/flow-server/src/main/resources/npmrc
@@ -4,3 +4,4 @@
 # This file sets the default parameters for manual `pnpm install`.
 #
 shamefully-hoist=true
+loglevel=error


tarekoraby:
Partly address https://github.com/vaadin/flow/issues/11282.

knoobie:
I'm not sure I like this change.. personally I would expect that flow fixes the warning instead of don't showing them.. in a corporation environment with dedicated security teams this is really bad for compliance.. once you build your application in your CI / DI pipeline and afterwards the CVE scanner complains about something but you can't find it in the logs where it comes from.

tarekoraby:
To my mind, the question is whether the typical Flow user can take some action based on the display of those warnings (which are only displayed once by the way in dev mode if one doesn't run `vaadin:dance`). My uneducated guess is that 99.3% of all Flow users will not do any action if they see these warnings. 

At the moment, the display of these warnings is only a disservice to the DX of the typical Flow user (especially the DX of new evaluators).

Artur-:
This sounds like hiding problems instead of resolving them

tarekoraby:
Ok, I stand corrected.

PR #13148, opened 2022-02-24
feat: add Vite support for exported web components

### LINKED ISSUE #12298: [Vite] Embedded Web Components do not init in development mode
After #12259 
When running embedded generic test in development mode the embedded component doesn't start requesting js files with Vite.

vursen:
## Description

This PR adds Vite support for exported web components making it possible to build and run them with Vite in development and production modes.

A brief overview of the differences between Webpack and Vite implementations: 

**Webpack:**

When there is any component marked as exported, in addition to `bundle-***.js`, Webpack will build an `export-***.js` chunk that contains `generated-flow-imports`. The Flow client will be generated by Flow as a separate script.

Once an exported web component is requested, `WebComponentBootstrapHandler` is the one that is in charge of generating a bootstrap script to initialize this web component.

The current Webpack implementation of `WebComponentBootstrapHandler` is based on `BootstrapHandler`. It dynamically builds an HTML bootstrap document injecting all the scripts required for the exported web component to work to the document's head. There is where it will inject the Flow client as a separate script and the `export-***.js` chunk, getting it from the chunk stats file.

Eventually, the handler serializes the head of the resulting HTML document into a set of javascript instructions and sends them to the browser as a result.

**Vite:**

When there is any component marked as exported, Vite will build `vaadin-web-component.ts` and inject it to `web-component.html`. It is worth noting that `vaadin-web-component.ts` includes both `generated-flow-imports` and the Flow client.

For Vite, the `WebComponentBootstrapHandler` logic was modified such that it instead takes `web-component.html` as an initial bootstrap document. The bootstrap document already contains the actual link to `vaadin-web-component.ts` which is  
 injected by Vite. All the other scripts required for the exported web component to work, such as the `@Push` annotation script, the handler injects dynamically. An advantage of this approach is that it doesn't require any chunk stats so that it works pretty much the same way as `IndexHtmlRequestHandler`.

Eventually, the handler serializes the head of the resulting HTML document into a set of javascript instructions and sends them to the browser as a result.

Fixes #12298

## Type of change

- [x] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateWebComponentHtml.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Create a task to generate <code>web-component.html</code> if necessary.
+     *
+     * @param frontendDirectory
+     *            frontend directory is to check if the file already exists
+     *            there.
+     */
+    TaskGenerateWebComponentHtml(File frontendDirectory) {
+        webComponentHtml = new File(frontendDirectory, WEB_COMPONENT_HTML);
```
Artur-, line 46: This would need a comment about the reasons why `WEB_COMPONENT_HTML` is in `frontend` and not `frontend/generated`

vursen: Done. https://github.com/vaadin/flow/pull/13148/commits/6b5f248b93911eb82b5de5cc04086a90b9b74478

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -476,6 +488,27 @@ public static String getIndexHtmlContent(VaadinService service)
         return getFileContent(service, INDEX_HTML);
     }
 
+    /**
+     * Gets the content of the <code>frontend/web-component.html</code> file
+     * which is served by webpack-dev-server in dev-mode and read from classpath
```
mcollovati, line 493: What's the meaning of `webpack-dev-server` in this context? Is it a specific reference to the webpack implementation or is it just a typo, and should be `vaadin-dev-server`?
I can see `web-component.html` used in `ViteHandler`, but can't find any reference on webpack stuff. Maybe I'm missing something.

mcollovati, line 493: What's the meaning of `webpack-dev-server` in this context? Is it a specific reference to the webpack implementation or is it just a typo, and should be `vaadin-dev-server`?
I can see `web-component.html` used in `ViteHandler`, but can't find any reference on webpack stuff. Maybe I'm missing something.

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -97,6 +104,48 @@ private WebComponentBootstrapContext(VaadinRequest request,
 
     private static class WebComponentBootstrapPageBuilder
             extends BootstrapPageBuilder {
+        @Override
+        public Document getBootstrapPage(BootstrapContext context) {
+            VaadinService service = context.getSession().getService();
+
+            if (FeatureFlags.get(service.getContext())
+                    .isEnabled(FeatureFlags.VITE)) {
```
caalador, line 112: Shouldn't webpack use the same bootstrapping as vite. There shouldn't really be any need to have different ways due to this.

vursen: Agree. There is actually a separate PR aligning the Webpack and Vite implementations: https://github.com/vaadin/flow/pull/13165. However, I didn't have enough time to finish it and some tests are still failing there.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
vursen: Right, looks like a typo that I missed when copy-pasting the JavaDoc block.

vursen: `webpack-dev-server` should become just `the dev server`, I guess.

Artur-: Updated javadoc

platosha (approved):

PR #13170, opened 2022-02-28
fix: clean overrides for removed dependencies

caalador:
Overrides should also automatically
be cleaned from the package.json
if there is no dependency that the
override would be targeting.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -188,8 +188,8 @@ public abstract class NodeUpdater implements FallibleCommand {
      * Gets the platform pinned versions that are not overridden by the user in
      * package.json.
      *
-     * @return JsonObject with the dependencies or empty JsonObject if file
-     *         doesn't exist
+     * @return {@code JsonObject} with the dependencies or empty
+     *         {@code JsonObject} if file doesn't exist
      * @throws IOException
      *             when versions file could not be read
      */
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -188,7 +188,8 @@ public abstract class NodeUpdater implements FallibleCommand {
      * Gets the platform pinned versions that are not overridden by the user in
      * package.json.
      *
-     * @return json object with the dependencies or {@code null}
+     * @return JsonObject with the dependencies or empty JsonObject if file
+     *         doesn't exist
      * @throws IOException
      *             when versions file could not be read
      */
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -137,19 +137,29 @@ boolean lockVersionForNpm(JsonObject packageJson, String versionsPath)
...
+                    && !isInternalPseudoDependency(
+                            versionsJson.getString(dependency))) {
+                overridesSection.put(dependency, "$" + dependency);
+                versionLockingUpdated = true;
+            }
+        }
+        for (String dependency : overridesSection.keys()) {
+            if (!(dependencies.hasKey(dependency)
+                    || devDependencies.hasKey(dependency))
+                    && overridesSection.getString(dependency).startsWith("$")) {
```
joheriks, line 157: Nit: could simplify conditional `!(a || b) && c` to `!a && !b && c` to reduce parentheses.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -188,7 +188,8 @@ protected File getPackageLockFile() {
      * Gets the platform pinned versions that are not overridden by the user in
      * package.json.
      *
-     * @return json object with the dependencies or {@code null}
+     * @return JsonObject with the dependencies or empty JsonObject if file
+     *         doesn't exist
```
joheriks, line 192: Nit: `{@code JsonObject}` (or {`@link ... }`)

caalador: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -137,19 +137,29 @@ boolean lockVersionForNpm(JsonObject packageJson, String versionsPath)
...
+                overridesSection.remove(dependency);
```
joheriks, line 158: Does it mean that any `$`-overrides will be removed, even some added by the user?

joheriks: (Note: certainly not blocking, as this may not be a real issue as this mechanism is unlikely to be used much in Flow projects, just asking for the record)

joheriks (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
Artur-: This checks the dependencies and devDependencies sections, right? So if I have a dependency on `foo` I can define an override for `foo` without this removing it?

joheriks: So it will remove overrides starting with `$` that are not in `dependencies` or `devDependencies`. As such a configuration would cause npm to raise an error anyway this is probably unlikely to be an issue ever.

PR #13172, opened 2022-02-28
fix: remove home node_modules before install

joheriks:
Files left by previous node+npm installation caused
corrupted installation when unpacking on top of an
existing installation. Fixed by removing 
~/.vaadin/node/node_modules prior to
installation of newer version.


mcollovati (approved):

joheriks:
The following scenario should be tested on Windows:
- Remove any global node installation
- Download and run V14 starter app to get Node 14 into `~/.vaadin`
- Download and run V23 starter app to get upgrade to Node 16

caalador (approved):

mcollovati (approved):

PR #13177, opened 2022-03-01
Fix @CssImport for Vite

### LINKED ISSUE #13176: CssImport causes Vite to fail
### Describe the bug

I use Vite and add CssImport like this:

    @CssImport(value = "./styles/text-field.css", themeFor = "vaadin-text-field")

After that, the application stops running

image

### Expected-behavior

_No response_

### Reproduction

Use Vite and add @CssImport(value = "./styles/text-field.css", themeFor = "vaadin-text-field")

### System Info

Windows 11, Vaadin 23.0.0.beta4, Java 11, Vite instead of Webpack

Artur-:
Fixes #13176


flow-tests/pom.xml
```diff
@@ -45,6 +44,9 @@
 
         <!-- make sure we do not leave webpack-dev-server running after IT -->
         <vaadin.reuseDevServer>false</vaadin.reuseDevServer>
+
+	<!-- Component version used for tests -->
+        <vaadin.version>23.0-SNAPSHOT</vaadin.version>
```
caalador, line 49: Should we really have components referenced from flow that reference flow?

flow-tests/test-frontend/vite-basics/frontend/cssimport-textfield.css
```diff
@@ -0,0 +1,3 @@
+:host {
+    background: rgb(173, 216, 230);
+}
```
caalador, line 3:
```suggestion
}

```

flow-tests/test-frontend/vite-basics/frontend/cssimport.css
```diff
@@ -0,0 +1,3 @@
+body {
+    background: rgb(211, 211, 211);
+}
```
caalador, line 3:
```suggestion
}

```

flow-tests/test-frontend/vite-basics/pom.xml
```diff
@@ -33,6 +33,11 @@
             <groupId>com.vaadin</groupId>
             <artifactId>flow-html-components-testbench</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-text-field-flow</artifactId>
+            <version>${vaadin.version}</version>
+        </dependency>
```
caalador, line 40: Seeing as we should avoid using things that depend on Flow repo/versions. We should use the same approach used in `flow-test-application-theme-component-live-reload` `TestThemedTextField` which is a vaadin-text-field, but doesn't depend on the Flow implementation

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
```diff
@@ -71,14 +71,15 @@
             + " document.head[before ? 'insertBefore' : 'appendChild'](tpl.content, document.head.firstChild);\n"
             + "};";
 
-    private static final String CSS_IMPORT = "import $css_%d from '%s';";
+    private static final String CSS_IMPORT = "import $cssFromFile_%d from '%s';%n" //
+            + "const $css_%d = typeof $cssFromFile_%d  === 'string' ? unsafeCSS($cssFromFile_%d) : $cssFromFile_%d;";
```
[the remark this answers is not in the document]
Artur-: Yes, that's much better. Done

flow-tests/test-frontend/vite-basics/pom.xml
Artur-: Is there any gain in depending on the npm component version instead of the Flow component version? Many of the Spring tests already depend on the Flow component versions

caalador: A min problem could become that Flow updates something that need components to change that then depends on the flow change that can't go in due to not having the changes in components that require the flow change.

Spring should also not use components inside the flow repo or at least the tests using components should run in another validation/build

Artur-: Fixed

caalador (approved):

PR #13179, opened 2022-03-01
feat: Add scrollIntoView method to Component

mstahv:
## Description

This change adds back a scrollIntoView method that we had in past framework versions.

The implementation is a trivial shorthand to the similarly named DOM method, and very handy in UI development.
It has been requested by multiple SoF and forum users, but apparently nobody wants to write this trivial enhancement issues as I couldn't find an existing issue for this.

The implementation leaves door open for overloading the method with a version with additional parameters, but already this method should solve most use cases.

## Type of change

- [ ] Bugfix
- [x] Feature

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -682,6 +682,6 @@ public abstract class Component
      * window.
      */
     public void scrollIntoView() {
-        getElement().callJsFunction("scrollIntoView");
+        getElement().scrollIntoView();
     }
 }
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -676,4 +676,11 @@ public abstract class Component
         }
         return locale;
     }
+
+    /**
+     * Scrolls the current component into the visible area of the browser window.
+     */
+    public void scrollIntoView() {
+        getElement().callJsFunction("scrollIntoView");
+    }
 }
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -678,7 +678,8 @@ public abstract class Component
     }
 
     /**
-     * Scrolls the current component into the visible area of the browser window.
+     * Scrolls the current component into the visible area of the browser
+     * window.
      */
     public void scrollIntoView() {
         getElement().callJsFunction("scrollIntoView");
--- a/flow-server/src/main/java/com/vaadin/flow/dom/Element.java
+++ b/flow-server/src/main/java/com/vaadin/flow/dom/Element.java
@@ -1603,4 +1603,17 @@ public class Element extends Node<Element> {
         }
     }
 
+    /**
+     * Executes the similarly named DOM method on the client side.
+     *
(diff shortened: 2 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/component/Component.java
```diff
@@ -676,4 +676,12 @@ protected Locale getLocale() {
         }
         return locale;
     }
+
+    /**
+     * Scrolls the current component into the visible area of the browser
+     * window.
+     */
+    public void scrollIntoView() {
+        getElement().callJsFunction("scrollIntoView");
```
joheriks, line 685: 1. Would it make sense to add `scrollIntoView` also to the `com.vaadin.flow.dom.Element` API and delegate to that?

2. What is the precondition for calling this method? When the component is not attached, e.g. in a constructor, the JS invocation is postponed until the component is attached but it does not scroll still. It also does not work when called from an attachlistener. It could either be documented that the caller must ensure the page is fully rendered before calling this method, or the scrolling action should be deferred somehow.

joheriks (changes requested):

joheriks:
Should be rebased.

flow-server/src/main/java/com/vaadin/flow/component/Component.java
mstahv: Don't really know why it didn't work on new instances, but setTimeout hack fixed it. Moved the implementation to Element.

caalador:
I guess something went wrong with the rebase `+3,451 −2,208 `

mstahv:
I guess so, and now it shows conflicts 🤷‍♂️

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/scroll/ScrollableView.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
```
joheriks, line 2:
```suggestion
 * Copyright 2000-2022 Vaadin Ltd.
```

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -1603,4 +1603,19 @@ private void informChildrenOfStateChange(boolean enabled, Element element) {
...
+     * @see <a href=
+     *      "https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView">Mozilla
+     *      docs</a>
+     * @return the element
+     */
+    public Element scrollIntoView() {
+        // for an unknown reason, needs to be called deferred to work on a newly
+        // created element
+        executeJs(
+                "var el = this; setTimeout(function() {el.scrollIntoView();}, 0);");
```
joheriks, line 1618: I guess this works because the JS is deferred long enough until the page rendering has stabilized. Looks fine to me, but should be tested on the major browsers. Which have you tested so far?  I will test the rest.

mstahv: Safari, Chrome, FF on Mac.

joheriks: Also tested Windows Edge and iOS Safari, seemed to scroll fine.

joheriks (approved):

PR #13189, opened 2022-03-02
fix: do not clean manual dev references

caalador:
If a devDependency reference has
been manually added to override
do not clean those.


Artur-:
Looks like it would need a test

caalador:
Added.

Artur- (approved):

PR #13218, opened 2022-03-05
test: Properly lock component versions in tests

Artur-:
To ensure that version locking works properly we must depend on the whole vaadin and not on individual Flow components. This also simplifies the setup

### DIFF
--- a/flow-tests/vaadin-spring-tests/pom.xml
+++ b/flow-tests/vaadin-spring-tests/pom.xml
@@ -59,148 +59,17 @@
 
             <dependency>
                 <groupId>com.vaadin</groupId>
-                <artifactId>vaadin-login-flow</artifactId>
-                <version>${vaadin.version}</version>
+                <artifactId>flow-bom</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
             </dependency>
             <dependency>
                 <groupId>com.vaadin</groupId>
-                <artifactId>vaadin-upload-flow</artifactId>
+                <artifactId>vaadin-bom</artifactId>
                 <version>${vaadin.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>com.vaadin</groupId>
-                <artifactId>vaadin-dialog-flow</artifactId>
-                <version>${vaadin.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>com.vaadin</groupId>
-                <artifactId>vaadin-notification-flow</artifactId>
-                <version>${vaadin.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>com.vaadin</groupId>
-                <artifactId>vaadin-ordered-layout-flow</artifactId>
-                <version>${vaadin.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>com.vaadin</groupId>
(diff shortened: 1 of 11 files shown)


flow-tests/vaadin-spring-tests/pom.xml
```diff
@@ -59,148 +59,17 @@
...
+                <artifactId>flow-bom</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
             </dependency>
             <dependency>
                 <groupId>com.vaadin</groupId>
-                <artifactId>vaadin-confirm-dialog-flow</artifactId>
+                <artifactId>vaadin-bom</artifactId>
                 <version>${vaadin.version}</version>
```
caalador, line 70: This should also be project.version as we should only be off by bugfix and the resolution is only to minor.
Probably should be put in the property so that it cam be overridden for new versions that do not have a vaadin snaphot available.

flow-tests/vaadin-spring-tests/test-spring-white-list/pom.xml
```diff
@@ -47,12 +47,7 @@
 
         <dependency>
             <groupId>com.vaadin</groupId>
-            <artifactId>vaadin-button-flow</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>com.vaadin</groupId>
-            <artifactId>vaadin-text-field-flow</artifactId>
+            <artifactId>vaadin-core</artifactId>
```
caalador, line 50: Should keep this light as we don't need all the pacakges to be installed.

caalador (changes requested):

flow-tests/vaadin-spring-tests/test-spring-white-list/pom.xml
Artur-: How? As long as `vaadin_versions.json` is part of `vaadin-core` it seems we need that for this to work

flow-tests/vaadin-spring-tests/pom.xml
Artur-: Updated the property

Artur-: Ah right, we cannot do that as the snapshot build sets the project version to 999.99-SNAPSHOT or similar. Reverted to 23.1-SNAPSHOT

caalador: =_=

flow-tests/vaadin-spring-tests/test-spring-white-list/pom.xml
caalador: Even without vaadin_versions.json we lock for versions in package.json shouldn't that be enough?

Artur-: Not if only `vaadin-button` and `vaadin-text-field` ends up in `package.json` and the problem is the version of some mixin

flow-tests/vaadin-spring-tests/pom.xml
caalador: Now this can't be autopicked to 23.0 as it will get the wrong version

caalador (approved):

PR #13223, opened 2022-03-06
fix: Always lock chokidar version to avoid chokidar 2

### LINKED ISSUE #13222: chokidar 2 with security alerts is installed when using npm
### Description of the bug

A new application has
```
npm ls --all|grep chokidar@
```

```
├─┬ chokidar@3.5.3
│ ├── chokidar@3.5.3 deduped
│ ├── chokidar@3.5.3 deduped
│ │ ├── chokidar@3.5.3 deduped
│ │   └─┬ chokidar@2.1.8
```

GitHub reports a "Regular expression denial of service" issue in `glob-parent` because of this

### Expected behavior

A new application should only use chokidar 3.

This already works for pnpm because it is forced in https://github.com/vaadin/flow/blob/master/flow-server/src/main/resources/pnpmfile.js#L38

What is needed is only
```
"chokidar": "$chokidar"
```

in the `"overrides"` section of `package.json`

-

### Versions

- Vaadin / Flow version: 23.0.0

Artur-:
chokidar 2 has known security issues that will show up when auditing.
These issues are probably impossible to exploit as dev tools do not
get input from a user but we should still fix them.

This is 1:1 with the pnpm locking in pnpmfile.js

Fixes #13222

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
@@ -141,9 +141,8 @@ public class TaskUpdatePackages extends NodeUpdater {
         final JsonObject dependencies = packageJson.getObject(DEPENDENCIES);
         for (String dependency : versionsJson.keys()) {
             if (!overridesSection.hasKey(dependency)
-                    && dependencies.hasKey(dependency)
-                    && !isInternalPseudoDependency(
-                            versionsJson.getString(dependency))) {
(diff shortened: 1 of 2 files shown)


caalador (dismissed):

caalador:
Could you rerun the `tests/test-frontend/test-npm/` for a new package lock also so we could get rid of the dependabot alert at the same time.

Artur-:
> Could you rerun the `tests/test-frontend/test-npm/` for a new package lock also so we could get rid of the dependabot alert at the same time.

✅

caalador (approved):

PR #13264, opened 2022-03-10
fix: allow server registered return calls

### LINKED ISSUE #13208: JavaScript call with return value does not work with Dialog
### Description of the bug

JavaScript call with return value does not work with Dialog in Vaadin 23, this used to work in Vaadin 22.0.7. See the attached example. 

### Expected behavior

JavaScript with return value, the response call back is executed.

@PageTitle("Dialog2")
@Route(value = "dialog2", layout = MainLayout.class)
public class DialogView2 extends VerticalLayout {

    public DialogView2() {
        add(new Button("Open Window", e -> new Window().open()));
    }

    private class Window extends Dialog {

        private Window() {
            add(new Button("Close", e -> close()));
        }

        @Override
            super.onAttach(attachEvent);
            UI ui = attachEvent.getUI();
            ui.getPage().fetchCurrentURL(url -> {
                ui.access(() -> Notification.show(url.toString()));
            });
        }
    }
}

### Versions

- Vaadin / Flow version: Vaadin 23.0.0
- Java version: 11

caalador:
Allow return calls for server-side
registered return channel on targeted
node.

Fixes #13208

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/ReturnChannelMap.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/ReturnChannelMap.java
@@ -164,4 +164,12 @@ public class ReturnChannelMap extends ServerSideFeature {
         return channels.get(Integer.valueOf(channelId));
     }
 
+    /**
+     * Return if map contains any registered channels.
+     *
+     * @return {@code true} if registered channels exist.
+     */
+    public boolean hasChannels() {
+        return !channels.isEmpty();
+    }
 }
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/ReturnChannelHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/ReturnChannelHandler.java
@@ -82,11 +82,8 @@ public class ReturnChannelHandler extends AbstractRpcInvocationHandler {
     @Override
     protected boolean allowInert(StateNode node) {
         // Allow calls if a return channel has been registered for the node.
-        final Optional<ReturnChannelMap> returnChannelMap = node
-                .getFeatureIfInitialized(ReturnChannelMap.class);
-        return returnChannelMap.isPresent()
-                ? returnChannelMap.get().hasChannels()
-                : false;
+        return node.getFeatureIfInitialized(ReturnChannelMap.class)
+                .map(ReturnChannelMap::hasChannels).orElse(false);
     }
 
     private static Logger getLogger() {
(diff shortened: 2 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/ReturnChannelHandler.java
```diff
@@ -79,6 +79,16 @@ public String getRpcType() {
...
     }
 
+    @Override
+    protected boolean allowInert(StateNode node) {
+        // Allow calls if a return channel has been registered for the node.
+        final Optional<ReturnChannelMap> returnChannelMap = node
+                .getFeatureIfInitialized(ReturnChannelMap.class);
+        return returnChannelMap.isPresent()
+                ? returnChannelMap.get().hasChannels()
+                : false;
```
joheriks, line 89:
```suggestion
        return node.getFeatureIfInitialized(ReturnChannelMap.class)
                .map(ReturnChannelMap::hasChannels).orElse(false);
```

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/ReturnChannelMap.java
```diff
@@ -164,4 +164,12 @@ public ReturnChannelRegistration get(int channelId) {
         return channels.get(Integer.valueOf(channelId));
     }
 
+    /**
+     * Return if map contains any registered channels.
+     *
+     * @return {@code true} if registered channels exist.
+     */
+    public boolean hasChannels() {
+        return !channels.isEmpty();
```
joheriks, line 173: So anything with a registered return channel will respond even if inert. Would it make sense to restrict this to the root (UI) node to not punch additional holes in the modality curtain?

joheriks: Update: let's keep the logic simple for now as we have yet to finalize what should be let through and what should be blocked.

caalador: As the registrations are only on the server side and get removed on response it shouldn't be a problem in opening up too much also this enables for instance:
```

    protected void execJS(String js,
            SerializableConsumer<String> resultHandler) {
        getElement().executeJs(js)
                .then(String.class, resultHandler);
    }

    private class Window extends Dialog {

        private Window() {
            add(new Button("show url", e-> {
                UI ui = getUI().get();
                execJS("return window.location.href", urlString ->
                        ui.access(() -> Notification.show(urlString))
                );
            }));
        }
    }
```
where the executing js is outside of the dialog, but not on the RootNode.

joheriks (approved):

Issue #13205, opened 2022-03-04
Click shortcut of a Button in a Dialog is not triggered


TatuLund:
### Description of the bug

Click shortcut of a Button in a Dialog is not triggering the click event of the button. 

Tested in Vaadin 23.0.0

Works in Vaadin 22.0.7

### Expected behavior

Shortcut triggers the Button click event (Notification shown in the attached example)

### Minimal reproducible example

```
@PageTitle("Dialog")
@Route(value = "dialog", layout = MainLayout.class)
public class DialogView extends VerticalLayout {

	public DialogView() {
		Dialog dialog = new Dialog();
		Button button = new Button("Click me");
		button.addClickListener(event -> {
			Notification.show("Hello");
		});
		button.addClickShortcut(Key.ENTER);
		dialog.add(button);
		dialog.open();		
	}
}
```

### Versions

- Vaadin / Flow version: 23.0.0
- Java version: 11
- OS version: Wind

joheriks:
The issue is because the state node root which handles the key events is inert (and hence [discards the event](https://github.com/vaadin/flow/blob/7ca947a658fe681766d3458e34a91460df4e7473/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java#L64-L69)). This will need some special handling to not block these. When fixing this, remember to test with key shortcut both behind the server-side modality (listener should not be triggered) and on top of it (the above case).

PR #13272, opened 2022-03-10
fix: set shortcut listener to modal only if owner is child

joheriks:
The previous fix for #13205 (PR #13263) moved all shortcut
registration event sources from UI to the modal component 
if one was active; this caused shortcuts owned by components
not under the modal tree to also be moved if they were registered
when a modal component was active, and hence they stopped
working when the modal component tree was removed. This
modifies the previous fix to only change the event source if the 
shortcut owner is under the an active modal tree.

### DIFF
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/ModalDialogIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/ModalDialogIT.java
@@ -58,10 +58,12 @@ public class ModalDialogIT extends ChromeBrowserTest {
         // shortcuts on view should not trigger while dialog is open
         pressShortcutKey(
                 $(NativeButtonElement.class).id(ModalDialogView.UI_BUTTON));
+        Assert.assertTrue("No event should be logged",
+                eventLog.$(DivElement.class).all().isEmpty());
 
         closeDialog();
 
-        // shortcuts on view should not trigger when dialog has been closed
+        // shortcuts on view should trigger when dialog has been closed
         pressShortcutKey(
                 $(NativeButtonElement.class).id(ModalDialogView.UI_BUTTON));
         validateLatestShortcutEvent(0, ModalDialogView.UI_BUTTON);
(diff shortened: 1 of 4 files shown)


flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/ModalDialogIT.java
```diff
@@ -49,8 +51,26 @@ public void modalDialogOpened_sameShortcutsListeningOnUi_noShortcutTriggered() {
...
+    public void modalDialogOpenInitially_dialogClosed_shortcutsInViewTrigger() {
+        open("open_dialog=modal");
+
+        // shortcuts on view should not trigger while dialog is open
+        pressShortcutKey(
+                $(NativeButtonElement.class).id(ModalDialogView.UI_BUTTON));
+
+        closeDialog();
+
+        // shortcuts on view should not trigger when dialog has been closed
```
knoobie, line 64:
```suggestion
        // shortcuts on view should trigger when dialog has been closed
```

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/ModalDialogIT.java
```diff
@@ -49,8 +51,26 @@ public void modalDialogOpened_sameShortcutsListeningOnUi_noShortcutTriggered() {
...
     }
 
+    @Test
+    public void modalDialogOpenInitially_dialogClosed_shortcutsInViewTrigger() {
+        open("open_dialog=modal");
+
+        // shortcuts on view should not trigger while dialog is open
+        pressShortcutKey(
+                $(NativeButtonElement.class).id(ModalDialogView.UI_BUTTON));
+
```
caalador, line 61: This could have a check that `div id="event-log"` has no child divs that would mean no added events
e.g. `assertTrue("No event should be logged", $(DivElement.class).id("event-log").$(DivElement.class).all().isEmpty())`

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/ModalDialogIT.java
caalador, line 64: nit comment indicates that event shouldn't have happened even though it has.

caalador (changes requested):

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/ModalDialogIT.java
joheriks: Thanks, copy-past gaffe. Fixed.

joheriks: Done.

caalador (approved):

Issue #11654, opened 2021-08-27
Google app on iOS shows "browser too old" for Vaadin apps


Artur-:
The user agent the Google App reports is
```
Mozilla/5.0 (iPhone; CPU iPhone OS 14_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) GSA/117.0.321844219 Mobile/15E148 Safari/604.1
```

For reference, Safari on the same iOS device reports
```
Mozilla/5.0 (iPhone; CPU iPhone OS 14_7_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.2 Mobile/15E148 Safari/604.1
```

The Google App user agent does not contain the version information at all. Therefore, the logic added in #6518 by @bennewi  looks at the webkit version. In this case it is 605.1.15 which according to the logic in the code is too old but in reality is Safari 14.1.

The user agent on latest Safari on desktop is
```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.2 Safari/605.1.15
```
which makes me think that either the wikipedia page with webkit versions is completely wrong or then Safari does not report the actual web kit version used.

PR #13282, opened 2022-03-11
fix: require Safari 14 minimum

### LINKED ISSUE #13271: Vaadin 22: Safari 13 should be detected as too old
### Description of the bug

Open https://start.vaadin.com/app on Safari 13. 
The app will appear, but doesn't work properly (there is a lot of js errors on the browser console):

image

### Expected behavior

Safari 13 should be checked to be too old in Vaadin "browser to old" handler.

Just open https://start.vaadin.com/app on Safari 13. 

### Versions

- Vaadin / Flow version: 22.0.8
- OS version: macOS
- Browser version (if applicable):  Safari 13

joheriks:
Safari 13 does not support the EventTarget constructor,
used in @vaadin/components-base:
https://caniuse.com/?search=EventTarget()

Fixes #13271

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
+++ b/flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
@@ -608,11 +608,10 @@ public class BrowserDetails implements Serializable {
         }
         // Safari 14+
         if (isSafari() && getBrowserMajorVersion() < 14) {
-            if (getOperatingSystemMajorVersion() > 14) {
-                return false;
-            }
-            if (getOperatingSystemMajorVersion() == 14
-                    && getOperatingSystemMinorVersion() >= 7) {
+            if (isIPhone() && (getOperatingSystemMajorVersion() > 14
+                    || (getOperatingSystemMajorVersion() == 14
+                            && getOperatingSystemMinorVersion() >= 7))) {
+                // #11654
                 return false;
             }
             return true;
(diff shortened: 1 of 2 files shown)


caalador (dismissed):

flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
```diff
@@ -606,8 +606,8 @@ public boolean isTooOldToFunctionProperly() {
         if (isEdge() && getBrowserMajorVersion() < 79) {
             return true;
         }
-        // Safari 13+
-        if (isSafari() && getBrowserMajorVersion() < 13) {
+        // Safari 14+
+        if (isSafari() && getBrowserMajorVersion() < 14) {
             if (getOperatingSystemMajorVersion() > 14) {
```
Artur-, line 611: What is the os check about?

joheriks: Looks like a fix to allow any Safari for iPhone: https://github.com/vaadin/flow/issues/11654
But without additional checks if seems we may be setting up for a regression once macOS hits 15...

Artur-: But why would any safari on iPhone work?

Artur-: Ah, I see... bugs and bugs

caalador (approved):

PR #13291, opened 2022-03-15
fix: Warn for duplicate npm classes

### LINKED ISSUE #13285: npm/pnpm install does not show version conflicts
### Description of the bug

We have a large application, that defines npm packages using the flow annotation. We had now the case, that (p)npm install had run all the time despite there have been no changes to the npm packages. After some research we found, that there have been two different versions of the same npm package. 

While in earlier Vaadin versions npm install had shown warnings for conflicting versions, now it does not anymore neither with npm nor with pnpm.

### Expected behavior

Using (p)npm install should show version conflicts.

1. Create a new app using the Vaadin starter.
2. Create two simple views, with an NpmPackage annotation, e.g. `@NpmPackage(value = "tippy.js", version = "6.3.5")` on one and `@NpmPackage(value = "tippy.js", version = "6.3.7")` on the second
3. Do the normal stuff to start the application. 
4. The package.json will now show one of the versions.
5. Now exchange the versions in the two views(so .7 to .5 and vice versa)
6. On the next startup the package json will have changed to the other version

Step 5 might seem a bit theoretical, but as Vaadin seem to use the "last" found version, this is necessary to show, how the version seems to be picked randomly from addons / other modules.

### Versions

- Vaadin / Flow version: 23
- Java version: 17
- OS version: Windows 10

caalador:
If encountering multiple NpmPackage
annotations with different versions log
a warning and inform of choice taken.

Fixes #13285

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
@@ -107,6 +107,8 @@ class FullDependenciesScanner extends AbstractDependenciesScanner {
      *            whether we are in V14 bootstrap mode
      * @param featureFlags
      *            available feature flags and their status
+     * @param fallback
+     *            whether dependency scanner is used as fallback
      */
     FullDependenciesScanner(ClassFinder finder, boolean useV14Bootstrap,
             FeatureFlags featureFlags, boolean fallback) {
@@ -126,6 +128,8 @@ class FullDependenciesScanner extends AbstractDependenciesScanner {
      *            whether we are in V14 bootstrap mode
      * @param featureFlags
      *            available feature flags and their status
+     * @param fallback
+     *            whether dependency scanner is used as fallback
      */
     FullDependenciesScanner(ClassFinder finder,
             SerializableBiFunction<Class<?>, Class<? extends Annotation>, List<? extends Annotation>> annotationFinder,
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
```diff
@@ -110,9 +129,11 @@
      */
```
mcollovati, line 129: It seems that `@param fallback` javadoc is missing

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
```diff
@@ -221,11 +242,28 @@ private String adaptCssValue(Annotation cssImport, String method) {
...
+                packageAnnotations.forEach(annotation -> {
+                    String value = getAnnotationValueAsString(annotation,
+                            VALUE);
+                    String version = getAnnotationValueAsString(annotation,
+                            VERSION);
+                    logs.add(value + " " + version + " " + clazz.getName());
+                    if (result.containsKey(value)
+                            && !result.get(value).equals(version)) {
+                        String foundVersions = "[" + version + ", "
+                                + result.get(value) + "]";
```
mcollovati, line 254: nitpicking: in `foundVersions` the `result.get(value)` element is at the last position, but the log below message says `First version found` will be considered; what about inverting versions order in `foundVersions`?

Also: can this statement be moved inside the below if statement?

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
```diff
@@ -92,7 +94,24 @@
...
+     * Creates a new scanner instance which discovers all dependencies in the
+     * classpath.
+     *
+     * @param finder
+     *            a class finder
+     * @param useV14Bootstrap
+     *            whether we are in V14 bootstrap mode
+     * @param featureFlags
+     *            available feature flags and their status
+     */
```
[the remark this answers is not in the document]
caalador: added

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
caalador: added

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
caalador: done

mcollovati (approved):

PR #13302, opened 2022-03-15
fix: Do not run any Maven goals automatically in Eclipse / VS Code

Artur-:
It seems like especially VS Code can start running `prepare-frontend` over and over again in some cases which will interfere with an externally running `mvn spring-boot:run` process that will continuously restart because of an updated `target/classes/META-INF/VAADIN/config/flow-build-info.json`

There should be no need nowadays to run `prepare-frontend` from Eclipse at all as it is run through the dev server if needed.


Artur-:
I could not find any problems in running a project in Eclipse with this

caalador (approved):

PR #13309, opened 2022-03-16
fix: Warn for duplicate npm classes (#13291)

### LINKED ISSUE #13285: npm/pnpm install does not show version conflicts

caalador:
If encountering multiple NpmPackage
annotations with different versions log
a warning and inform of choice taken.

Fixes #13285

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
@@ -57,6 +57,7 @@ class FullDependenciesScanner extends AbstractDependenciesScanner {
     private static final String COULD_NOT_LOAD_ERROR_MSG = "Could not load annotation class ";
 
     private static final String VALUE = "value";
+    private static final String VERSION = "version";
 
     private ThemeDefinition themeDefinition;
     private AbstractTheme themeInstance;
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
```diff
@@ -70,15 +71,21 @@
...
+
     /**
      * Creates a new scanner instance which discovers all dependencies in the
      * classpath.
      *
      * @param finder
      *            a class finder
+     * @param fallback
+     *            whether FullDependenciesScanner is used as fallback
+     * @param finder
```
mcollovati, line 84: Seems like `@param finder`  is duplicated

caalador: removed

mcollovati (approved):

PR #13312, opened 2022-03-18
Use Node 16.14.2

Artur-:


tarekoraby:
Will this make `16.4.2` the required minimum now? If so, why make the change?

knoobie:
This won't change the required minimum version in my understand - just the default version vaadin is installing if no supported version was found (which is a good change with the CVE in 16.14.0/1)

The code with the required minimum version is below:

```java

private static final int SUPPORTED_NODE_MAJOR_VERSION = 16;
private static final int SUPPORTED_NODE_MINOR_VERSION = 14;

```

Artur-:
Updated the javadoc

caalador (approved):

PR #13344, opened 2022-03-23
feat: Enable usage of pnpm >5

### LINKED ISSUE #12953: npx pnpm@5.18.10 executes pnpm 6 in some cases
### Description of the bug

Apparently when you have pnpm 6 installed globally then running
```
npx pnpm@5.18.10
```
on Windows will run pnpm 6 and not 5. So the output of
```
npx pnpm@5.18.10 --version
```
is e.g.
```
6.9.1
```

This will result in versions not being locked and e.g. today running a 22.0.4 project might result in
```
Failed to execute 'define' on 'CustomElementRegistry': the name "vaadin-lumo-styles" has already been used with this registry
```
as pnpm will pick some mixin versions that are from 22.0.5 and then there will be duplicates installed

### Expected behavior

Everything should work even if you have pnpm 6 globally installed

This was reproduced on Windows with the default https://start.vaadin.com project. Cannot reproduce on Mac

### Versions

- Vaadin / Flow version: 22.0.4

caalador:
Allow usage of pnpm 6 by
then generating `.pnpmfile.cjs`
instead of `pnpmfile.js`

Closes #12953

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
@@ -504,14 +504,14 @@ public class TaskRunNpmInstall implements FallibleCommand {
             return;
         }
 
-        String pnpmFileName = "pnpmfile.js";
+        String pnpmFileName = ".pnpmfile.cjs";
         final List<String> pnpmExecutable = tools.getSuitablePnpm();
         pnpmExecutable.add("--version");
         try {
             final FrontendVersion pnpmVersion = FrontendUtils.getVersion("pnpm",
                     pnpmExecutable);
-            if (pnpmVersion.isNewerThan(new FrontendVersion("6.0"))) {
-                pnpmFileName = ".pnpmfile.cjs";
+            if (pnpmVersion.isOlderThan(new FrontendVersion("6.0"))) {
+                pnpmFileName = "pnpmfile.js";
             }
         } catch (FrontendUtils.UnknownVersionException e) {
             packageUpdater.log().error("Failed to determine pnpm version", e);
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -496,13 +498,27 @@ private Process runNpmCommand(List<String> command, File workingDirectory)
...
             return;
         }
 
+        String pnpmFileName = "pnpmfile.js";
+        final List<String> pnpmExecutable = tools.getSuitablePnpm();
+        pnpmExecutable.add("--version");
+        try {
+            final FrontendVersion pnpmVersion = FrontendUtils.getVersion("pnpm",
+                    pnpmExecutable);
+            if (pnpmVersion.isNewerThan(new FrontendVersion("6.0"))) {
```
mcollovati, line 513: Just a question: is it meant that if global pnpm is at version 6.0.0 it should use `pnpmfile.js` instead of `pnpmfile.cjs`?

caalador: If the determined pnpm to use version is > 6.0 then we should use `.pnpmfile.cjs`.
At this point it doesn't matter where the pnpm is from, just the version matters.

mcollovati: Ok, so if I understand correctly, pnpm `6.0.0` should still use 'pnpmfile.js', while all later versions should use '.pnpmfile.cjs'?

caalador: Good catch. anything older than 6.0.0 should get js all the rest cjs. reversed the check as also 6.0.0 should be .pnpmfile.cjs

mcollovati (approved):

PR #13350, opened 2022-03-24
feat: Hilla push implementation behind a feature flag

Artur-:
This adds support for endpoints methods to return a Flux<T> which will be streamed to the client side to the subscribers that have registered using TheEndpoint.theMethod().onNext(value => ...);

The implementation is based on https://socket.io/ but this is not considered to be part of the public API. We should be free to change the underlying implementation if needed.

Fixes https://github.com/vaadin/hilla/issues/56


Artur-:
Known issues and limitations:
1. There are no tests 😅 
2. The code in connect-client should go into `@hilla/frontend`
3. There is an issue in socket.io that can cause problems if you use long polling: https://github.com/socketio/engine.io-server-java/issues/46
4. It uses org.json which cannot be replaced and only brings extra overhead https://github.com/socketio/engine.io-server-java/issues/47

flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
```diff
@@ -80,6 +83,7 @@ public FeatureFlags(Lookup lookup) {
         features.add(new Feature(EXAMPLE));
         features.add(new Feature(VITE));
         features.add(new Feature(MAP_COMPONENT));
+        features.add(new Feature(HILLA_PUSH));
```
knoobie, line 86: I'm wondering: Would it be good to allow for some kind of sanity check if the feature is "use-able" in the current application? Like only showing hilla if an endpoint is available or some other kind of check? Or different feature sets for flow, hilla and components like adding a category enum to show only "use-able" features to the developer depending on the type of application he starts?

Artur-: It would be great to have the feature flags so that they can be in separate modules and not all in `flow-server`: https://github.com/vaadin/flow/issues/13356

platosha (changes requested):

fusion-endpoint/src/main/resources/connect-client.default.template.ts
```diff
@@ -1,3 +1,218 @@
-import {ConnectClient} from '@hilla/frontend';
+import { ConnectClient } from '@hilla/frontend';
+import { DefaultEventsMap } from "@socket.io/component-emitter";
+import { io, Socket } from 'socket.io-client';
+
+/* Most of the code below should end up in @hilla/frontend but the first version is easier to do this way.*/
+/** The following is copied from @hilla/frontend as the methods are not exported in package.json */
+export function getCookie(name: string) {
```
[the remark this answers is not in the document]
Artur-: No, the whole beginning here is copied from `common-frontend` as the needed function was not exported. It will be removed once the code is moved to the correct place

platosha (changes requested):

platosha (approved):

Artur-:
To my understanding, all Lift dependency issues here are invalid:
* spring-beans@5.3.17 - not used
* spring-web@5.3.19 - says the issue is with 5.3.16 and older
* spring-security-crypto@5.6.3 - says the issues is with 5.3 and older

PR #13373, opened 2022-03-28
feat: Automatically use available converters in Binder 

### LINKED ISSUE #12881: Automatically use available converters in binder
### Describe your motivation

When you bind a field and a property which do not have the same value type, you get an exception, e.g.
```
Caused by: java.lang.IllegalStateException: Property type 'java.util.UUID' doesn't match the field type 'java.lang.String'. Binding should be configured manually using converter.
	at com.vaadin.flow.data.binder.Binder.bindProperty(Binder.java:2972) ~[flow-data-2.7.9.jar:2.7.9]
```

Instead of saying that the user should configure a converter, Binder should automatically pick a suitable converter if one is available. For the above case there is e.g. `StringToUuidConverter`. It is just not used by default

mcollovati:
## Description

When binding a field and a property which do not have the same value type
an exception is thrown. With this change, an attempt is made to
automatically pick a suitable converter from the ones provided by the framework.

Closes #12881

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

### DIFF
--- a/flow-data/src/test/java/com/vaadin/flow/data/binder/BeanBinderTest.java
+++ b/flow-data/src/test/java/com/vaadin/flow/data/binder/BeanBinderTest.java
@@ -15,10 +15,11 @@
  */
 package com.vaadin.flow.data.binder;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertSame;
-import static org.junit.Assert.assertTrue;
+import javax.validation.constraints.Digits;
+import javax.validation.constraints.Max;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
 
 import java.io.Serializable;
 import java.time.LocalDate;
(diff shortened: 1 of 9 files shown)


flow-data/src/test/java/com/vaadin/flow/data/binder/BeanBinderTest.java
```diff
@@ -35,6 +35,7 @@
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
```
mshabarov, line 38:
```suggestion
```

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderInstanceFieldTest.java
```diff
@@ -15,10 +15,15 @@
...
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.time.LocalDate;
 import java.util.Arrays;
+import java.util.Date;
+import java.util.UUID;
 
 import org.junit.Assert;
+import org.junit.Ignore;
```
mshabarov, line 26:
```suggestion
```

mshabarov (dismissed):
Well designed, implemented and tested, LGTM.

flow-data/src/test/java/com/vaadin/flow/data/binder/BeanBinderTest.java
mcollovati: Done

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderInstanceFieldTest.java
mcollovati: Done

mshabarov (approved):

PR #13374, opened 2022-03-28
fix: validate node and npm on creation

### LINKED ISSUE #13364: Global NPM may be used even if node version is unsupported
### Description of the bug

When an unsupported global node is detected, Flow installs a newer version in `$HOME/.vaadin` directory; however when it comes to run npm, the version from global node installation may be used if it is newer that the minimum one supported by Flow.

```
[INFO] Globally node version 16.13.1 is older than 16.14.0. Using node from /Users/xxxxx/.vaadin.
...
2022-03-25 14:21:14.513  INFO 2026 --- [onPool-worker-1] c.v.f.s.frontend.TaskUpdatePackages      : using '/Users/xxxxx/.nvm/versions/node/v16.13.1/bin/npm --no-update-notifier --no-audit --scripts-prepend-node-path=true --ignore-scripts install' for frontend package installation
```

### Expected behavior

NPM from `$HOME/.vaadin` node installation is used when global node version is unsupported

* Globally install node older than 16.14.0 (e.g. `nvm install v16.13.1`)
* Update global npm to a version greater than 8.3 (e.g. `npm update -g npm`)
* Create an application from start.vaadin.com and run it with `mvn spring:run`
* Check that Maven goal is using node from `$HOME/.vaadin` but `npm install` is performed by the globally installed npm

### Versions

- Vaadin / Flow version: 23
- Java version: 11
- OS version: Ubuntu 21.10

caalador:
Always when creating a FrontendTools
validate node and npm so that if the
global node is too old and we should
use the installed one we do not use
the global npm even if it was new enough.

Fixes #13364

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -370,6 +370,7 @@ public class BuildFrontendUtil {
 
         FrontendToolsSettings settings = getFrontendToolsSettings(adapter);
         FrontendTools tools = new FrontendTools(settings);
+        tools.validateNodeAndNpmVersion();
         if (featureFlags.isEnabled(FeatureFlags.VITE)) {
             BuildFrontendUtil.runVite(adapter, tools);
         } else {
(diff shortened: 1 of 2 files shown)


mcollovati:
Should we first pick to 23 PR vaadin/flow#13344 and then retry the pick of this PR?

PR #13377, opened 2022-03-28
fix: Register push related beans only when the feature flag is enabled

Artur-:
Followup for #13350


Artur-:
Rebased on latest master

platosha (approved):

PR #13400, opened 2022-03-31
chore: Change debug level to info as this logging is frequent in harmless cases

TatuLund:
Closes: https://github.com/vaadin/flow-components/issues/1157


Artur-:
Should this even be `info` or is it `debug` logging?

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -44,12 +44,12 @@
         StateNode node = ui.getInternals().getStateTree()
                 .getNodeById(getNodeId(invocationJson));
         if (node == null) {
-            getLogger().warn("Ignoring RPC for non-existent node: {}",
+            getLogger().info("Ignoring RPC for non-existent node: {}",
```
mshabarov, line 47: This looks like hiding the problem, but not fixing. 
I would start with figuring out why does clicking a button ends up with "node not found" errors on the server side.
Data filtering looks not related from the first glance.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
TatuLund: Removal of the component during the event process will result logging this warning, see the simplest case here https://github.com/vaadin/flow-components/issues/1157#issuecomment-1084318223

TatuLund: Logging is correct, but too vocal as I am assuming that it is allowed to remove the TextField in its value change listener. If you think otherwise, we should clearly document that removal of the textField in its value change listener or other event triggered by the component itself is not allowed.

tarekoraby: @TatuLund your [comment](https://github.com/vaadin/flow-components/issues/1157#issuecomment-1084318223) describes a situation in which the warning is a false positive. But are there other situations in which the warning would be a true positive, and hence would warrant the `warn` level?

Artur-: If you get "Ignoring RPC for non-existent node" in your production log. What should you do? It sure sounds to me like info you would need while debugging a problem and not particularly care about in production. If everything works in production, it is a "false positive". If something is broken in production, the error message still gives you no way of finding out what is wrong and how to fix it

mshabarov (changes requested):
@TatuLund I had an another and more deep look into `AbstractRpcInvocationHandler::handle` method. Changing `warn` to `debug` seems a fine solution, because:
1. When the `node` does not exist on the server state tree, the server basically has no information why, therefore it cannot decide whether log the message and what level to choose, and what do with it in the end.
2. When the `node` is detached, IMO it doesn't make sense to figure out why it is detached. We can probably check if the `node` corresponds to `HasValue` component, but it is a "private" case when such a warnings occur.
3. As Artur said, and I agree with him, these warn message doesn't give a clear understanding of what is broken or what's wrong with the caller's code. But it can be useful for development, debugging or anomaly investigation.
So in short I propose to set `debug` level and apply this patch.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -48,12 +48,12 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
```
mshabarov, line 51:
```suggestion
            getLogger().debug("Ignoring RPC for non-existent node: {}",
```

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -48,12 +48,12 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
                     getNodeId(invocationJson));
             return Optional.empty();
         }
         if (!node.isAttached()) {
-            getLogger().warn("Ignoring RPC for detached node: {}",
+            getLogger().info("Ignoring RPC for detached node: {}",
```
mshabarov, line 56:
```suggestion
            getLogger().debug("Ignoring RPC for detached node: {}",
```

mshabarov (approved):

PR #13401, opened 2022-03-31
Upgrade to Spring Boot 2.6.6 and Spring 5.3.18

Artur-:


pom.xml
```diff
@@ -83,8 +83,8 @@
         <failOnMissingWebXml>false</failOnMissingWebXml>
 
         <!-- Dependencies -->
-        <spring.version>5.3.17</spring.version>
-        <spring.boot.version>2.6.5</spring.boot.version>
+        <spring.version>5.3.18</spring.version>
+        <spring.boot.version>2.6.6</spring.boot.version>
```
tarekoraby, line 87:
```suggestion
        <spring.boot.version>2.6.5</spring.boot.version>
```

Let's change Spring version for now until Spring Boot is released.

knoobie: Boot is going to be released today

Artur-: There is no point in only upgrading Spring

tarekoraby: I was thinking just to see if the tests pass right away. But probably they will anyway.

caalador: Only fusion-endpoint uses the spring version as everything else is defined through the `<artifactId>spring-boot-dependencies</artifactId>` bom using the spring.boot.version

caalador (approved):

PR #13409, opened 2022-04-01
test: remove shadow root helpers

caalador:
update getting from shadowroot


mshabarov (changes requested):
Apparently missing ";" in the test codes.

mshabarov (approved):

PR #13411, opened 2022-04-01
Make components have a 'title', that is, a tool-tip.

### LINKED ISSUE #4169: Unified Tooltip handling needed in component Java API's
At the moment tooltips are implemented as per component.

In Vaadin 8 the tooltip API was coherent in whole framework. It would be desirable to have the same in Vaadin 10 as well. 

- Have setTooltip(..) (or something similar) API for each component

- Have setTooltipGenerator(..) (or something similar) for components with item list like ComboBox, Grid, ...

enver-haase:
## Description

A customer of mine needs setDescription(String) of Vaadin8 badly, and it's a good idea for us to finally implement https://github.com/vaadin/flow/issues/4169 .

Fixes #4169 .

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

I haven't done anything like that as flow/README.md does not have compilation instructions.
So the change is entirely 'should work' judging by eye.

Not even that.

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -2,7 +2,7 @@ name: Flow Validation
 on:
   push:
     branches: [master, '9.0']
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened]
   workflow_dispatch:
 jobs:
@@ -40,7 +40,7 @@ jobs:
           echo "::set-output name=matrix-unit::$(./scripts/computeMatrix.js unit-tests --parallel=1 current module args)"
       - name: Compile and Install Flow
         run: |
-          cmd="mvn install -B -DskipTests  -pl \!flow-plugins/flow-gradle-plugin"
+          cmd="mvn install -B -ntp -DskipTests  -pl \!flow-plugins/flow-gradle-plugin"
           eval $cmd -T 2C -q || eval $cmd
       - name: Save workspace
         run: |
@@ -84,7 +84,7 @@ jobs:
           [ -n "${{matrix.module}}" ] && \
             ARGS="-pl ${{matrix.module}} -Dtest=${{matrix.args}}" || \
             ARGS="-pl ${{matrix.args}}"
-          cmd="mvn -B -T 1C $ARGS"
+          cmd="mvn -B -ntp -T 1C $ARGS"
           set -x -e -o pipefail
           ($cmd -T 1C verify || $cmd clean verify -U) | tee mvn-unit-tests-${{matrix.current}}.out
       - name: Set build status flag
@@ -144,11 +144,11 @@ jobs:
       - name: Compile Shared modules
         run: |
           if [ ${{matrix.current}} -eq 2 -o ${{matrix.current}} -eq 3 ]; then
-            cmd="mvn install -B -DskipTests -Pit-shared-modules -amd -pl flow-tests"
+            cmd="mvn install -B -ntp -DskipTests -Pit-shared-modules -amd -pl flow-tests"
             $cmd -T 1C || $cmd
           fi
           if [ ${{matrix.current}} -eq 4 -o ${{matrix.current}} -eq 5 -o ${{matrix.current}} -eq 6 ]; then
-            cmd="mvn install -B -DskipTests -Pit-shared-spring-modules -amd -pl flow-tests"
+            cmd="mvn install -B -ntp -DskipTests -Pit-shared-spring-modules -amd -pl flow-tests"
             $cmd -T 1C || $cmd
           fi
       - name: Run ITs
@@ -156,7 +156,7 @@ jobs:
           [ -n "${{matrix.module}}" ] && \
             ARGS="-Dfailsafe.forkCount=4 -pl ${{matrix.module}} -Dit.test=${{matrix.args}}" || \
             ARGS="-pl ${{matrix.args}}"
-          cmd="mvn -V -B -e -fae -Dcom.vaadin.testbench.Parameters.testsInParallel=5 -Dfailsafe.rerunFailingTestsCount=2 -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3 $ARGS"
+          cmd="mvn -V -B -ntp -e -fae -Dcom.vaadin.testbench.Parameters.testsInParallel=5 -Dfailsafe.rerunFailingTestsCount=2 -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3 $ARGS"
           set -x -e -o pipefail
           $cmd verify | tee -a mvn-it-tests-${{matrix.current}}.out
       - name: Set build status flag
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ flow-tests/**/*.generated.js
 flow-tests/**/*.generated.ts
 flow-tests/**/generated/**
 flow-tests/**/pnpmfile.js
+flow-tests/**/.pnpmfile.cjs
 flow-tests/**/*.npmrc
 
 *.xml.versionsBackup
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@ Vaadin Flow
 ======
 *[Vaadin Flow](https://vaadin.com/flow) is the Java framework of Vaadin Platform for building modern web sites that look great, perform well and make you and your users happy.*
 
-**For instructions about developing web applications with Vaadin Flow**, please refer to [the starter packs for Vaadin 14 with Flow](https://vaadin.com/start) or [the documentation](https://vaadin.com/docs/flow/Overview.html).
+**For instructions about developing web applications with Vaadin Flow**, please refer to [the documentation](https://vaadin.com/docs/latest/flow/overview).
 
 **To contribute, first refer to [Contribution Guide](/CONTRIBUTING.md) for general instructions and requirements for contributing code to Flow.**
 
-**For issues related to [Vaadin Fusion](https://vaadin.com/fusion), please refer to the https://github.com/vaadin/fusion repository.**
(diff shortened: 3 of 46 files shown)


flow-server/src/main/java/com/vaadin/flow/component/Component.java
```diff
@@ -49,7 +49,7 @@
  * @since 1.0
  */
 public abstract class Component
-        implements HasElement, AttachNotifier, DetachNotifier {
+        implements HasTitle, AttachNotifier, DetachNotifier {
```
mshabarov, line 52: I would prefer to have `HasTitle` interface apart from `HasElement`, because having a title is a more specific functionality.
Maybe it's better to include `setTitle` API into another existing mixin interface.
Additionally, not all the components can have a tooltip.

Why not have just one method in `Component` without interface?

mshabarov: Why https://github.com/vaadin-component-factory/tooltip doesn't fit your goals? Is it too complex?

flow-server/src/main/java/com/vaadin/flow/component/HasTitle.java
```diff
@@ -0,0 +1,5 @@
+public interface HasTitle extends HasElement {
+    default setTitle(String title) {
```
mshabarov, line 2: It needs a javadoc and especially, the explanation that it will work with desktop browsers, but not with touchscreen devices.

mshabarov (changes requested):
Needs design

flow-server/src/main/java/com/vaadin/flow/component/Component.java
enver-haase: > Additionally, not all the components can have a tooltip.
> 
> Why not have just one method in `Component` without interface?

Are you saying you suggest one method in 'Component', knowing however that not all the components can have a tooltip?   Which ones cannot, by the way...?

mshabarov: > Which ones cannot, by the way...?

Technically, I believe all the components can have a title. But I'm not sure that semantically the components like layouts or overlays (dialog or notification) are fine to have a tooltip/title.

> Are you saying you suggest one method in 'Component', knowing however that not all the components can have a tooltip?

Yes, you're right, that it is non-sense. It was a random thought.
What I would prefer to see is a mixin interface like you introduced but: 1. With different name, IMO `HasDescription` would be better. 2. It is not implemented by `Component`, but by a certain flow components.

mshabarov:
@enver-haase we discussed this internally, so what I propose:
1. Change the name of mixin to `HasTooltip` and method name to `setTooltip `
2. Since the `title` attribute is global (https://html.spec.whatwg.org/multipage/dom.html#global-attributes), let `Component` implements `HasTooltip`, but in addition to `HasElement`: `Component implements HasElement, HasTooltip`.
3. Add javadocs and license header to `HasTooltip`.

Please let me know if you have any comments or need help for moving forward this PR.

enver-haase:
There is HtmlComponent which already derives from Component and it does implement setTitle/getTitle. Having said that, it would be better to not let HtmlComponent inherit setTooltip(String) and and setTitle(String) doing the same.

More design work is needed.
https://github.com/vaadin/flow/issues/4169

flow-server/src/main/java/com/vaadin/flow/component/Component.java
enver-haase: There's a reworked version of the PR now, so I'll call this part of the conversation resolved.

flow-server/src/main/java/com/vaadin/flow/component/HasTitle.java
enver-haase: ok

flow-server/src/main/java/com/vaadin/flow/component/HtmlComponent.java
```diff
@@ -49,30 +49,4 @@ protected HtmlComponent() {
...
-     * Sets the title of this component. Browsers typically use the title to
-     * show a tooltip when hovering an element or any descendant without a title
-     * value of its own.
-     *
-     * @param title
-     *            the title value to set, or <code>""</code> to remove any
-     *            previously set title
-     */
-    public void setTitle(String title) {
-        set(titleDescriptor, title);
```
mshabarov, line 63: `titleDescriptor` is not used anymore in this class and can be removed as well.

rolfsmeds:
Is it really wise to introduce a `HasTitle` interface and `setTitle` API, rather than `HasTooltip` / `setTooltip`?

1. Developers want to set a _tooltip_, not a _title_. Even though the html attribute currently used for this is for some reason called `title`, it is neither intuitive or descriptive.
2. We probably want to use the same API for an accessible (and possibly touch-friendly) tooltip implementation in the future, rather than ending up with _both_ `HasTitle` and `HasTooltip`
3. Once we improve the implementation to something that is actually accessible, it may not necessarily have anything to do with the `title` attribute anymore (or at least that attribute will be only part of the solution), so the name will make even less sense.
5. We want to preserve the term "title" for component features that are generally understood as "titles", such as the title of a Dialog, the title of a Card, etc.

enver-haase:
Hi Rolf,
the HtmlComponent already had setTitle/getTitle as published API, so in consequence this is why all components now got this.
I don't like the name too much myself, as it is mirroring the HTML misnomer.
As the ticket about a "real" (ARIA compatible, mobile compatible) tool-tip is still open that probably needs a few more years of design and implementation I feel we have a sensible stop-gap solution now and when we end up having the real deal, then the naming with "setTitle" will make clear what the simple outdated HTML variant is and what is the perfect solution that took us years to provide.

enver-haase:
"title" is a global attribute which in fact can be set on every component there is.
Should we find it is nonsensical on one or some, there we can always introduce an override which throws an exception.

mshabarov:
This PR is being reverting because we have now a clash between the new `HasTitle` mixin interface and `LoginOverlay` component (and probably other flow-components), so it was a premature change.

We can rework it later as follows:

1. Create a `HasTooltip` interface with `set/getTooltip` API in Flow with default implementation based on "title".
2. Implement `HasTooltip` interface for Flow components, where needed.
3. Keep `HtmlContainer` with `setTitle` method as it was in the past.
4.  Change implementation of `set/getTooltip` in flow components once we have a more "sophisticated" way for tooltip, including not only the plain text, but styling and other features.

@rolfsmeds @enver-haase

enver-haase:
@mshabarov How does the setTitle in HtmlComponent come into play with this plan? Change its name to setTooltip/getTooltip with incompatible API, for Vaadin 24?

mshabarov:
`HtmlContainer` API can be kept as is, because it represents a built-in HTML components on server-side. Thus, having an API  for mirroring common HTML attribute is fine IMO.

rolfsmeds:
We're not sure if the "advanced" tooltip feature will be implemented for `HtmlComponent`s, so they may well continue to have the plain `title` attribute implementation. Not sure which is better in that case:

1. Use the setTitle name for it, to reflect the difference
2. Use the setTooltip name for consistency with Vaadin components, despite the difference

I also don't see a problem with having `setTooltip` and `setTitle` methods in parallel on those in the future: one for the advanced tooltip and one for the html attrib, since they are, by definition, html elements. The `setTitle` / `getTitle` method names also don't conflict with any other feature in those classes, as opposed to other Vaadin components (like LoginOverlay and, from 23.1, Dialog).

mshabarov:
One additional note: having `HasTooltip` only in Flow and implement it in `Component` seems less flexible than having it implemented for concrete flow components: flow components API can be used as part of implementation of tooltip, other components can be used as a parts of tooltip, styles etc.

enver-haase:
I am very unhappy about reverting this. Also, I am not convinced about having setTitle for HtmlComponent, LoginOverlay, Dialog with at least two different meanings. The clash in LoginOverlay / Dialog could have been documented (a work-around is known) and everyone could have lived with a simple solution they know and love from Vaadin7/8. 

Now we're back heading towards a messy solution with at least two names with at least three meanings, and certainly to title-attribute based common API across HtmlComponent and other Component. What is more, we know this and we postpone this to 'we `can` rework it `later`' knowing very well we have no solution since the introduction of Vaadin 10 from Jund 25th, 2018.

rolfsmeds:
Just to clarify, I am not against adding API for setting the `title` attrib. I just dont' want a `setTitle()` method on Vaadin components because of the above reasons. If it's called `setTooltip()` (has the interface `HasTooltip`) I'm totally on board.

I'm also a bit skeptical about adding that interface to the `Component` class (again for the reason mentioned above), but I'm not against adding it to individual component classes.

enver-haase:
And you reckon we should use the `setTooltip()` with the same API for any later implementation which is ARIA and mobile compatible, or will this then be another new method with a new name?  I wouldn't know of any component that should not carry `HasTooltip` -- I do oppose being able to add the tooltip to only some select components, given that `title` is a global attribute, and given that the explicit request of my customer is to have a common API across the components, like what they are used to from Vaadin 8.

rolfsmeds:
Yes, the improved implementation would use the same API, as I have said above.

I haven't looked through everything that extends `Component`, but at the very least in these components it simply will not work however "global" an attribute `title` may be: Dialog, ConfirmDialog, CookieConsent, LoginOverlay.

Having the `HasTitle` / `HasTooltip` interface on those is a promise that it works. A PR that makes it work them as well is of course welcome.

mshabarov:
I agree with Rolf, having the mixin `HasTooltip` interface implemented by `Component` does not only make no sense for some components, but also can potentially break the existing codes for custom components in Vaadin based applications where a similar signature of methods `set/getTooltip` are already applied, but with other semantics behind it. In this case IMO it's better to explicitly add the mixin interface to the class, if it's really needed and not break the component design. Custom components might have a various semantics and purposes which doesn't make sense to have a tooltip on them.

PR #13412, opened 2022-04-02
feat: add experimental flag for the spreadsheet component

manolo:


yuriy-fix (dismissed):

flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
```diff
@@ -60,7 +60,10 @@
             "Map component (Pro)", "mapComponent",
             "https://github.com/vaadin/platform/issues/2611", true,
             "com.vaadin.flow.component.map.Map");
-
+    public static final Feature SPREADSHEET_COMPONENT = new Feature(
```
mshabarov, line 63: Where is it added to use? This is only a declaration.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
manolo: done

mshabarov (approved):

PR #13427, opened 2022-04-06
chore: Upgrade to Jetty 10.0.9

Artur-:


caalador (approved):

Artur-:
Seems like 9.0 is still on Jetty 9 so let's leave it like that

PR #13429, opened 2022-04-06
chore: Make JSON valid JSON

Artur-:


flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -1,8 +1,3 @@
-// This TypeScript configuration file is generated by vaadin-maven-plugin.
-// This is needed for TypeScript compiler to compile your TypeScript code in the project.
-// It is recommended to commit this file to the VCS.
-// You might want to change the configurations to fit your preferences
-// For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
```
caalador, line 5: Should we keep the information as a `details` or `comment` object as I doubt this information is available anywhere else.

Artur-:
Seems like TypeScript actually considers tsconfig.json as "JSON with comments" https://github.com/sarod/TypeScript/commit/b60d88fa80bdd09029610584603734419ef89ae9 so maybe we should not do this at all

PR #13438, opened 2022-04-07
test: Use better port detection for dev server

caalador:
When binding to all interfaces it seems that the http server can choose a port that is in use for localhost but free for other interfaces.
As was done for the statistics sender test.


vaadin-dev-server/src/test/java/com/vaadin/base/devserver/WebpackHandlerTest.java
```diff
@@ -607,7 +607,7 @@ public static HttpServer createStubWebpackTcpListener(int port, int status,
             String response) throws Exception {
         try {
             HttpServer httpServer = HttpServer
-                    .create(new InetSocketAddress(port), 0);
+                    .create(new InetSocketAddress("localhost", 0), 0);
```
Artur-, line 610: Maybe we should take https://github.com/vaadin/flow/blob/8be2b8aa3f793d74e1914e999c6c3a74bdd7e71e/flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeBrowserTest.java#L136 , copy the Selenium code used to find a port and then reuse it in all places instead of having different implementations here and there

Artur-:
Can it be somewhere else than `vaadin-dev-server` so that it could be used also in `ChromeBrowserTest`?

caalador:
From a usage point of view we only need it with the dev server. Any good suggestions for where it should be to be able to use with test modules also?

Artur-:
> From a usage point of view we only need it with the dev server. Any good suggestions for where it should be to be able to use with test modules also?

Isn't `flow-test-generic` available in `vaadin-dev-server` and tests?

Artur-:
Sorry, forgot about this PR. If it is used everywhere then it sounds like `flow-server` would be the place...

caalador:
flow-server is not the place as it would have been only in use with the dev-server. But having it only for tests seems more stable.

flow-test-util/src/main/java/com/vaadin/flow/testutil/net/EphemeralPortRangeDetector.java
```diff
@@ -0,0 +1,47 @@
...
+     * Returns the first port in the ephemeral range
+     *
+     * @return The first ephemeral port
+     */
+    int getLowestEphemeralPort();
+
+    /**
+     * Returns the last port that could be searched for free ports
+     *
+     * @return The first port that may be free
```
mcollovati, line 43: Should it be the "last" port?
```suggestion
     * @return The last port that may be free
```

flow-test-util/src/main/java/com/vaadin/flow/testutil/net/FixedIANAPortRange.java
```diff
@@ -0,0 +1,36 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+package com.vaadin.flow.testutil.net;
+
+/**
+ * Fixed port range for when a defined can not be calculated used.
```
mcollovati, line 20: Maybe "used" can be removed? 
```suggestion
 * Fixed port range for when a defined can not be calculated.
```

flow-test-util/src/main/java/com/vaadin/flow/testutil/net/PortProber.java
```diff
@@ -0,0 +1,161 @@
...
+        if (current.isLinux()) {
+            ephemeralRangeDetector = LinuxEphemeralPortRangeDetector
+                    .getInstance();
+        } else {
+            ephemeralRangeDetector = new FixedIANAPortRange();
+        }
+    }
+
+    public static final int HIGHEST_PORT = 65535;
+    public static final int START_OF_USER_PORTS = 1024;
```
mcollovati, line 67: Does it make sense to add javadocs for those public fields? Class is meant for internal use only, so maybe they can also be omitted

mcollovati (approved):

Artur- (approved):

PR #13472, opened 2022-04-08
test: Add lock file to speed up pnpm install and wait for dev server

Artur-:
Hopefully makes the test more reliable

### DIFF
--- a/flow-tests/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/ScopesIT.java
+++ b/flow-tests/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/ScopesIT.java
@@ -55,6 +55,7 @@ public class ScopesIT extends AbstractSpringTest {
     @Test
     public void checkUiScope() throws Exception {
         getDriver().get(getTestURL() + "ui-scope");
+        waitForDevServer();
 
         String mainId = findElement(By.id("main")).getText();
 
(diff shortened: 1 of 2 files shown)


flow-tests/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/ScopesIT.java
```diff
@@ -33,6 +33,7 @@ protected String getTestPath() {
     @Test
     public void checkSessionScope() throws Exception {
         open();
+        waitForDevServer();
```
mcollovati, line 36: Doesn't `open()` already do a `waitForDevServer()`?

Artur-: It does indeed. Then I don't know why the current test does not work because that wait should wait forever and not only for 20s like waitForVaadin

Artur-: But if this makes the test more stable, maybe it is still good to merge... Compared to having the PR build fail 50% of the time

mcollovati (approved):

flow-tests/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/ScopesIT.java
Artur-: So this one is actually unnecessary... but the test below does not use `open` and therefore never calls `waitForDevServer`

mcollovati (approved):

PR #13473, opened 2022-04-08
fix: Do not append a nonsense slot to exported web components

### LINKED ISSUE #9899: Confusing <slot> in exported web component
As seen in the screenshot in https://github.com/vaadin/flow/issues/9880, there's a `<slot>` element in the exported web component which assumingly has no purpose since all the content is supposed to reside inside the shadow root.

This is most likely a leftover from before https://github.com/vaadin/flow/issues/5363 was fixed. It seems like it doesn't cause any practical problems, but it is still confusing since it gives the impression that content is expected to exist in the light DOM.

Artur-:
Fixes #9899

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/webcomponent/webcomponent-script-template.js
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/webcomponent/webcomponent-script-template.js
@@ -13,7 +13,6 @@ _ThemeImport_class _TagCamel_ extends HTMLElement {
       }
     `;
     _ApplyTheme_shadow.appendChild(style);
-    shadow.appendChild(document.createElement("slot"));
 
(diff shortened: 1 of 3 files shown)


caalador (dismissed):

Artur-:
The failing tests assumes that you can do
```<client-select show='true'>Web Component</client-select>```

and the text `Web Component` shows up inside the shadow root of the exported component.

Is there any reason this test exists?

caalador:
I at least do not know what the use case is for.

Artur-:
To me it sounds plain wrong. If I wanted `<client-select>` to support child content, I should be the one who decides where in the component the content goes... Let's delete the test then and hope for the best

caalador (approved):

PR #13520, opened 2022-04-13
test: Upgrade test Vaadin version to 23.0.5

Artur-:


flow-tests/test-resources/src/main/java/com/vaadin/flow/uitest/ui/dependencies/TestVersion.java
```diff
@@ -2,7 +2,7 @@
 
 public class TestVersion {
 
-    public static final String VAADIN = "22.0.0-alpha9";
+    public static final String VAADIN = "23.0.4";
```
caalador, line 5:
```suggestion
    public static final String VAADIN = "23.0.5";
```
As now we are getting mixed 23.0.4 and 23.0.5 in the package lock file.

Artur-: We can change that but it will break the same way when 23.0.6 is out

caalador: Should we then just have an empty package.json files that get populated instead to not get snyk pr's for internal testing things?

Artur-: Let's upgrade to 23.0.5 and see how it goes

Artur-:
Should the Spring tests use the same version or is there a reason they are using 
```
        <vaadin.version>23.1-SNAPSHOT</vaadin.version>
```
?

Artur-:
Let's do a separate PR for that..

caalador:
> Should the Spring tests use the same version or is there a reason they are using
> 
> ```
>         <vaadin.version>23.1-SNAPSHOT</vaadin.version>
> ```
> 
> ?
The flow tests only need a component node package, but spring uses the java component also.

Spring should test the latest against the latest, but as spring is now in flow I'm not convinced that it should use the platform at all for testing things.

taefi (approved):

PR #13526, opened 2022-04-13
chore: Use proper version for jackson-databind

mshabarov:


Artur- (changes requested):
Why would you do that? What does "proper" mean?

Artur- (approved):
Ah crap, I read it the wrong way

Artur-:
Was a bit confused there for a second..

mshabarov:
`jackson-databind` used other version which is for other jackson dependencies. This is the only place where the common version used, other places uses `jackson.databind.version` as well.

PR #13535, opened 2022-04-13
feat: Add Binder.refreshFields()

tarekoraby:
In situations when a bean `foo` that is bound to a binder has one its properties changed on the server-side, one currently needs to reset the binding to that bean in order for the new property value to be reflected on the fields.

So if I have
```java
binder.setBean(foo);
foo.setBar(new Bar());
```

Then I would need to call `binder.setBean(foo)` again for the value to be reflected on the field associated with the property `bar`.

This usage is, however, not obvious to the user as what they want to do is only update the fields associated with the binder in order for them to reflect the updated state of the bean. (plus there is a very minor wastefulness in calling `binder.setBean(foo)` again as that removes the same bean only to add it again).

This PR adds a `Binder.refreshFields()` API to match the intention of the user in those use cases.


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2128,6 +2128,23 @@ public void readBean(BEAN bean) {
...
+     * If no bean is currently associated with this binder (using
+     * {@link #setBean(Object)}), the bound fields will be cleared.
+     *
+     * @see #setBean(Object)
+     * @see #readBean(Object)
+     * @see #writeBean(Object)
+     * @see #writeBeanIfValid(Object)
+     *
+     */
+    public void updateFields() {
```
knoobie, line 2144: Instead of update fields, I would suggest refresh fields :) nevertheless huge +1.. I have a custom binder in all my applications with a similar method as well

tarekoraby: I'm neutral between using "update" and "refresh". I can change it if the majority thinks "refresh" is better.

taefi: I'm also in favor of `refreshFields` or even `refreshFieldsValues` ;)

tarekoraby: `refreshFields` it is 👍

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2128,6 +2129,23 @@ public void readBean(BEAN bean) {
...
+     * <p>
+     * If no bean is currently associated with this binder (using
+     * {@link #setBean(Object)}), the bound fields will be cleared.
+     *
+     * @see #setBean(Object)
+     * @see #readBean(Object)
+     * @see #writeBean(Object)
+     * @see #writeBeanIfValid(Object)
+     *
+     */
```
taefi, line 2144:
```suggestion
    /**
     * Refreshes the fields values by reading them again from
     * the currently associated bean via invoking their 
     * corresponding value provider methods. 
     * <p>
     * If no bean is currently associated with this binder ( 
     * {@link #setBean(Object)} has not been called before 
     * invoking this method), the bound fields will be cleared.
     * <p>
     * @see #setBean(Object)
     * @see #readBean(Object)
     * @see #writeBean(Object)
     * @see #writeBeanIfValid(Object)
     */
```

taefi (changes requested):

mshabarov:
@tarekoraby please merge current master to your branch

mshabarov (approved):

taefi (approved):

PR #13543, opened 2022-04-14
fix: reusable themes with vite

### LINKED ISSUE #13538: Vite doesn't work with a custom theme that packaged as jar
### Description of the bug

Vite cannot find and serve a custom theme that packaged as jar.

[extend-existing-app-theme-demo](https://github.com/vaadin/custom-theme-demo/tree/main/extend-existing-app-theme-demo) with `vaadin.version=23.0.3` and  `com.vaadin.experimental.viteForFrontendBuild=true`. 

### Expected behavior

It should work just like in webpack.

### Actual behavior

`application-theme-plugin/theme-loader` cannot find the packaged theme because the `themeResourceFolder` value on `vaadin-dev-server-settings.json` is not correct .

The application starts in dev mode but vite dev server cannot find the resources. It's looking for them in frontend/themes but they are in target/flow-frontend/themes

[custom-theme-demo](https://github.com/vaadin/custom-theme-demo) with `vaadin.version=23.0.3` and webpack works properly.

### Versions:
    - Vaadin / Flow version: 23.0.3
    - Java version: 17.0.1

caalador:
Fix resolution of reusable themes
when using vite

Fixes #13538


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -18,12 +18,13 @@ import brotli from 'rollup-plugin-brotli';
 import replace from '@rollup/plugin-replace';
 import checker from 'vite-plugin-checker';
 
-const appShellUrl = '.';
+const appShellUrl = '.'
```
taefi, line 21: I suggest keeping the semicolon :)
```suggestion
const appShellUrl = '.';
```

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -327,10 +328,16 @@ export const vaadinConfig: UserConfigFn = (env) => {
     root: 'frontend',
     base: '',
     resolve: {
-      alias: {
-        themes: themeFolder,
-        Frontend: frontendFolder
-      },
+      alias: [
+        { find: 'themes', replacement: (importee: string) => {
+            if(existsSync(path.resolve(themeResourceFolder, importee))){
```
taefi, line 333:
```suggestion
            if (existsSync(path.resolve(themeResourceFolder, importee))) {
```

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -327,10 +328,16 @@ export const vaadinConfig: UserConfigFn = (env) => {
...
+              return path.resolve(themeResourceFolder, settings.themeFolder);
+            }
+            return themeFolder;
+          }
+        },
+        { find: 'Frontend', replacement: frontendFolder }
```
taefi, line 339: Where is this `'Frontend'` referring to?

taefi (changes requested):

flow-server/src/main/resources/vite.generated.ts
caalador: * A special prefix used by webpack to map imports placed in the
     * {@link FrontendUtils#DEFAULT_FRONTEND_DIR}. e.g.
     * <code>import 'Frontend/foo.js';</code> references the
     * file<code>frontend/foo.js</code>.

flow-server/src/main/resources/vite.generated.ts
caalador: done

flow-server/src/main/resources/vite.generated.ts
caalador: done

flow-server/src/main/resources/vite.generated.ts
caalador: So basically when you have the js annotation `@JsModule("./my-component.js")` it will be added to `generated-flow-imports.js` as `import 'Frontend/my-component.js';` which will then be modified by webpack/vite to target the correct directory so for instance `C:/Users/user1/flow/flow-tests/test-frontend/vite-basics/frontend/my-component.js`

taefi (approved):

PR #13547, opened 2022-04-14
fix: Force async version 3.2.2

caalador:
force async version 3.2.2 to get fix for
Prototype Pollution in async

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -420,6 +420,7 @@ public abstract class NodeUpdater implements FallibleCommand {
         defaults.put("workbox-core", WORKBOX_VERSION);
         defaults.put("workbox-precaching", WORKBOX_VERSION);
         defaults.put("glob", "7.1.6");
+        defaults.put("async", "3.2.2");
 
         return defaults;
     }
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -420,6 +420,7 @@ Map<String, String> getDefaultDevDependencies() {
         defaults.put("workbox-core", WORKBOX_VERSION);
         defaults.put("workbox-precaching", WORKBOX_VERSION);
         defaults.put("glob", "7.1.6");
+        defaults.put("async", "3.2.2");
```
Artur-, line 423: Do we really need to add it here? Where does it come from?

caalador: async is a transitive dependency which comes in from `portfinder` and `jake` where `portfinder` comes in from `webpack-dev-server` and `jake` from `@surma/rollup-plugin-off-main-thread` that comes from `workbox-build`

caalador: We still get only version `2.6.4` for async even if I update `workbox-webpack-plugin` to `6.5.3` that was published 2 days ago.

Artur-:
Looks ok. When https://github.com/webpack/webpack-dev-server/pull/4384 has been released, we can remove the override and upgrade webpack-dev-server instead. Jake was fixed in [v10.8.5](https://github.com/jakejs/jake/releases/tag/v10.8.5)

Artur- (approved):

mshabarov (approved):

PR #13597, opened 2022-04-22
fix: Optimize Binder#doWriteIfValid to avoid duplicate field level validations

TatuLund:
This is useful optimization for the case where you have e.g. validator that checks that input is not in the database, hence avoiding duplicate DB query.

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -2258,12 +2258,12 @@ public class Binder<BEAN> implements Serializable {
                     currentBindings);
 
             // Field level validation can be skipped as it was done already
-            boolean validatorsInUse = isValidatorsDisabled();
+            boolean validatorsNotInUse = isValidatorsDisabled();
             setValidatorsDisabled(true);
             currentBindings
                     .forEach(binding -> ((BindingImpl<BEAN, ?, ?>) binding)
                             .writeFieldValue(bean));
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2257,9 +2257,14 @@ private BinderValidationStatus<BEAN> doWriteIfValid(BEAN bean,
             Map<Binding<BEAN, ?>, Object> oldValues = getBeanState(bean,
                     currentBindings);
 
+            // Field level validation can be skipped as it was done already
+            boolean validatorsInUse = isValidatorsDisabled();
```
OlliTietavainenVaadin, line 2261: Variable name is inconsistent here; if `isValidatorsDisabled` returns true, I would expect validators not to be in use.

TatuLund: Good catch

simasch:
When will this change be relased?

taefi (changes requested):
@TatuLund Is there a chance that a field can get an invalid value between the first time bindings are checked and when the `writeFieldValue` is being called on all the bindings? Possibly it wouldn't be a case that an end-user can cause it, but some race condition, maybe!?

taefi (approved):

simasch:
If this fix is really in 23.1.0.alpha4 then it's not fixed!

simasch:
@TatuLund 
I was wrong. The fix works in a simple example. Sorry.

PR #13603, opened 2022-04-22
fix: do not select from session if parent has a better match

### LINKED ISSUE #13348: dynamic registration of route-template in session-scope might break existing routes of application-scope
### Description of the bug

Following situation:
* annotated route "other/view" (OtherView.class)
* annotated route-template ":first/:second" (DefaultView.class)

=> at this point navigation to "other/view" with view OtherView.class works!

* override ":first/:second"  by RouteConfiguratio.forSessionScope().... with class OverrideView.class.

=> at this point navigation to "other/view" does not work anymore. OverrideView.class is shown instead of OtherView.class

* execute RouteConfiguration.forSessionScope().getHandledRegistry().clean() 

=> now navigation to "other/view" with view OtherView.class works again!

### Expected behavior

Navigation to path "other/view" uses OtherView.class at any time.

See https://github.com/olafj/vaadin-router-problem. Start the application, default-view displays an intro text to reproduce the problem. 

### Versions

- Vaadin / Flow version: 23
- Java version: 11
- OS version: Ubuntu 21.10
- Browser version (if applicable): does not matter 
- Application Server (if applicable): Tomcat 9
- IDE (if applicable): does not matter

caalador:
Check if parent registry contains a better match
than the session registry.

Fixes #13348

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/SessionRouteRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/SessionRouteRegistry.java
@@ -172,11 +172,72 @@ public class SessionRouteRegistry extends AbstractRouteRegistry {
         Objects.requireNonNull(url, "pathString must not be null.");
         final Optional<Class<? extends Component>> target = getConfiguration()
                 .getTarget(url);
-        if (target.isPresent()) {
+        if (target.isPresent() && !parentContainsExactMatch(url)) {
             return target;
         }
 
         return getParentRegistry().getNavigationTarget(url);
+    }
+
+    /**
+     * When session registry contains a matching route, check if it is not an
+     * exact match and parenmt registry contains an exact or closer match.
+     *
+     * @param url
+     *            url to check for exact match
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/SessionRouteRegistry.java
```diff
@@ -172,13 +172,74 @@ public Optional<Class<? extends Component>> getNavigationTarget(
...
+     *
+     * @param urlParts
+     *            url segment parts list
+     * @param target
+     *            route target segments list
+     * @return amount of matching segments from the start
+     */
+    private int equalParts(List<String> urlParts, List<String> target) {
+        int maxSize = urlParts.size() > target.size() ? target.size()
+                : urlParts.size();
```
mcollovati, line 231: does it make sense to use `Math.min(...)` here, to improve readability? 
```suggestion
        int maxSize = Math.min(urlParts.size(), target.size());
```

mcollovati (approved):

PR #13604, opened 2022-04-22
fix: ignore server-modality for poll RPC invocations

### LINKED ISSUE #13337: Poll listeners are not invoked when a modal dialog is shown
### Description

When you open a modal dialog, poll listeners are not invoked anymore

### Expected outcome

The poll listeners should be called, even when a modal dialog is shown.

### Actual outcome

The poll listeners are not called anymore

Main `Application` class:
```
@SpringBootApplication
@Push(value = PushMode.MANUAL, transport = Transport.LONG_POLLING)
public class Application extends SpringBootServletInitializer implements AppShellConfigurator, VaadinServiceInitListener
{
        SpringApplication.run(Application.class, args);
    }

    @Override
    public void serviceInit(ServiceInitEvent serviceInitEvent)
    {
        serviceInitEvent.getSource().addUIInitListener(uiInitEvent -> {
            UI ui = uiInitEvent.getUI();
            ui.setPollInterval(2000);
            ui.addPollListener(pollEvent -> {
                System.out.println("In poll listener");
            });
        });

    }
}
```

View with a button to show a modal dialog
```
@PageTitle("Modal Dialogs")
@Route(value = "")
public class ModalDialogs extends VerticalLayout
{
    public ModalDialogs()
    {
        add(new Button("Start operation", clicked -> {
            Dialog modalDialog = new Dialog(new Div(new Span("Modal dialog")));
            modalDialog.setCloseOnEsc(false);
            modalDialog.setCloseOnOutsideClick(false);
            modalDialog.setModal(true);

            modalDialog.add(new Button("Close", clickEvent -> modalDialog.close()));
            modalDialog.setWidth("40ch");

            modalDialog.open();
        }));
    }
}
```

When the modal dialog is open, the code in the poll listener attached in the `Application` class is no longer invoked.

### Steps to reproduce

Clear steps describing how to reproduce the issue.

1. Add a poll listener on UI initialization
2. Show a modal dialog anywhere in the application

### Environment

- Vaadin v23.0.1
- OS: Linux

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE 11
- [ ] iOS Safari
- [ ] Android Chrome

taefi:
## Description
The inert status of the UI as the result of
server-modality should be ignored for
some RPC calls such as polling events. 

Fixes #13337 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
@@ -369,8 +369,8 @@ public class ComponentUtil {
     }
 
     /**
-     * Check if the component has at least one registered listener
-     * of the given event type.
+     * Check if the component has at least one registered listener of the given
+     * event type.
      *
      * @param component
      *            the component to which the listener(s) are registered.
--- a/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
@@ -369,10 +369,8 @@ public class ComponentUtil {
     }
 
     /**
-     * Queries the given component whether a listener has been registered for
-     * the given event type or not. This method delegates the query to the
-     * component and simply returns the result of
-     * {@link com.vaadin.flow.component.Component#hasListener(Class)}.
+     * Check if the component has at least one registered listener
+     * of the given event type.
      *
      * @param component
      *            the component to which the listener(s) are registered.
@@ -383,7 +381,7 @@ public class ComponentUtil {
      */
     public static <T extends ComponentEvent<?>> boolean hasEventListener(
             Component component, Class<? extends T> eventType) {
-        return component.getEventBus().hasListener(eventType);
+        return component.hasListener(eventType);
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
@@ -369,6 +369,24 @@ public class ComponentUtil {
     }
 
     /**
+     * Queries the given component whether a listener has been registered for
+     * the given event type or not. This method delegates the query to the
+     * component and simply returns the result of
+     * {@link com.vaadin.flow.component.Component#hasListener(Class)}.
+     *
+     * @param component
+     *            the component to which the listener(s) are registered.
+     * @param eventType
+     *            the event type for which the listener(s) are registered.
+     * @return a boolean indicating whether at least one listener registered to
+     *         the component for the given event type.
+     */
+    public static <T extends ComponentEvent<?>> boolean hasEventListener(
+            Component component, Class<? extends T> eventType) {
+        return component.getEventBus().hasListener(eventType);
+    }
+
+    /**
      * Dispatches the event to all listeners registered for the event type.
      *
      * @see Component#fireEvent(ComponentEvent)
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
@@ -20,6 +20,7 @@ import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.vaadin.flow.component.PollEvent;
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.internal.StateNode;
 import com.vaadin.flow.shared.JsonConstants;
@@ -61,7 +62,8 @@ public abstract class AbstractRpcInvocationHandler
                     + "the client side for an inactive (disabled or invisible) node id='{}'",
                     getClass().getName(), node.getId());
             return Optional.empty();
-        } else if (!allowInert(node) && node.isInert()) {
+        } else if (!ignoreInertForRpcInvocation(ui, invocationJson)
+                && !allowInert(node) && node.isInert()) {
             getLogger().trace(
                     "Ignored RPC for invocation handler '{}' from "
                             + "the client side for an inert node id='{}'",
@@ -70,6 +72,29 @@ public abstract class AbstractRpcInvocationHandler
         } else {
             return handleNode(node, invocationJson);
         }
+    }
+
+    /**
+     * Specifies whether inert status should be ignored for an RPC invocation or
+     * not. For example, if push is enabled, polling events should still be
+     * handled, while ignoring other requests.
+     *
+     * @param ui
+     *            the current UI instance
+     * @param invocationJson
+     *            the JsonObject containing invocation properties
+     * @return a boolean indicating that the inert status should be ignored for
+     *         the current invocation or not.
+     */
+    protected boolean ignoreInertForRpcInvocation(UI ui,
+            JsonObject invocationJson) {
+        return isPollEventInvocation(invocationJson);
+    }
+
+    private boolean isPollEventInvocation(JsonObject invocationJson) {
+        return invocationJson.hasKey(JsonConstants.RPC_EVENT_TYPE)
+                && PollEvent.DOM_EVENT_NAME.equalsIgnoreCase(
+                        invocationJson.getString(JsonConstants.RPC_EVENT_TYPE));
     }
 
     protected boolean allowInert(StateNode node) {
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
@@ -15,6 +15,8 @@
  */
 package com.vaadin.flow.server.communication.rpc;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 
 import com.vaadin.flow.component.ComponentUtil;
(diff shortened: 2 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -72,6 +77,110 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
+
+        if (!isPollingEnabledForUI(ui)) {
+            StringBuilder warning = new StringBuilder(
+                    "Ignoring Poll RPC for UI that does not have polling enabled.");
+            if (!VaadinService.getCurrent().getDeploymentConfiguration()
+                    .isProductionMode() && ui.getUIId() > 0) {
+                warning.append(" UI id: '{}'");
+            }
+            getLogger().warn(warning.toString(), ui.getUIId());
+            return false;
```
caalador, line 107:
```suggestion
            getLogger().warn("Ignoring Poll RPC for UI that does not have polling enabled.");
            getLogger().debug("Ignored payload\n{}", invocationJson);
            return false;
```
I doubt you can figure anything out with the UI.id so that is a bit extra. But for debugging the whole invocation could be added.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -72,6 +77,110 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
+     *         accordance with the UI instance.
+     */
+    private boolean isLegitimatePollEventInvocation(UI ui,
+            JsonObject invocationJson) {
+        int allowedKeysSize = 3;
+        String[] invocationKeys = invocationJson.keys();
+        if (invocationKeys == null
+                || invocationKeys.length != allowedKeysSize) {
+            return false;
+        }
```
caalador, line 165: I would rather check that we have only the allowed amount and types of keys:
```suggestion
        List<String> allowedKeys = Arrays.asList(new String[]{JsonConstants.RPC_TYPE, JsonConstants.RPC_NODE, JsonConstants.RPC_EVENT_TYPE});
        List<String> invocationKeys = Arrays.asList(invocationJson.keys());
        if (!allowedKeys.containsAll(invocationKeys)) {
            return false;
        }
```

flow-server/src/test/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandlerTest.java
```diff
@@ -49,10 +56,45 @@ protected Optional<Runnable> handleNode(StateNode node,
...
+            });
+        }
+
+        @Override
+        public String getRpcType() {
+            return null;
+        }
+    }
+
+    private AbstractRpcInvocationHandler nonEmptyHandler = new NonEmptyRpcInvocationHandler();
```
caalador, line 79: This is not really needed as you can just have the TestRpcInvocationHandler return an `Optional.of(()->{})` as the return value there is not used for anything.

flow-server/src/test/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandlerTest.java
```diff
@@ -49,10 +56,45 @@ protected Optional<Runnable> handleNode(StateNode node,
...
+        deploymentConfiguration = Mockito.mock(DeploymentConfiguration.class);
+        Mockito.when(deploymentConfiguration.isProductionMode())
+                .thenReturn(false);
+
+        vaadinService = Mockito.mock(VaadinServletService.class);
+        Mockito.when(vaadinService.getDeploymentConfiguration())
+                .thenReturn(deploymentConfiguration);
+
+        VaadinService.setCurrent(vaadinService);
+    }
```
caalador, line 96: We shouldn't need to init any deploymentconfiguration as there is no need to know if production mode or dev mode for the message.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -72,6 +77,110 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
+
+        if (!invocationJson.hasKey(JsonConstants.RPC_NODE)) {
+            return false;
+        }
+        StateNode node = ui.getInternals().getStateTree()
+                .getNodeById(getNodeId(invocationJson));
+        // Polling events should target only the root component in a UI:
+        return node.getParent() == null;
```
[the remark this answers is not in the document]
taefi: Good catch! Done.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
taefi: Done.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -72,6 +77,110 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
+     * ignoring other requests.
+     *
+     * @param ui
+     *            the current UI instance
+     * @param invocationJson
+     *            the JsonObject containing invocation properties
+     * @return a boolean indicating that the inert status should be ignored for
+     *         the current invocation or not.
+     */
+    protected boolean ignoreInertForRpcInvocation(UI ui,
```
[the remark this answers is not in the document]
taefi: Good point, I'll change to this.

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
```diff
@@ -368,6 +368,24 @@ public static <T extends ComponentEvent<?>> Registration addListener(
...
+     * @param component
+     *            the component to which the listener(s) are registered.
+     * @param eventType
+     *            the event type for which the listener(s) are registered.
+     * @return a boolean indicating whether at least one listener registered to
+     *         the component for the given event type.
+     */
+    public static <T extends ComponentEvent<?>> boolean hasEventListener(
+            Component component, Class<? extends T> eventType) {
+        return component.getEventBus().hasListener(eventType);
```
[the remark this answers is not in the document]
taefi: Done.

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
```diff
@@ -368,6 +368,24 @@ public static <T extends ComponentEvent<?>> Registration addListener(
                 domListenerConsumer);
     }
 
+    /**
+     * Queries the given component whether a listener has been registered for
+     * the given event type or not. This method delegates the query to the
+     * component and simply returns the result of
+     * {@link com.vaadin.flow.component.Component#hasListener(Class)}.
```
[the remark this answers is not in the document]
taefi: Done.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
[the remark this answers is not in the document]
taefi: However, if I want to call this from `allowInert` I'll miss the UI reference that exists in `handle` unless I change the `allowInert`'s signature which is not a big hassle but a change to a long-existing API. I think `UI.getCurrent()` cannot be used instead as it does not always return the same UI instance (navigation is possible even while having a modal component on a UI).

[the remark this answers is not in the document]
caalador: allowInert is only 4 months old and has been updated last 1 month ago to support features that can be counted as bugs like this one.
Also as it's an internal feature it's fine to update it.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -72,6 +77,110 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
+            return false;
+        }
+
+        if (!invocationJson.hasKey(JsonConstants.RPC_TYPE)) {
+            return false;
+        }
+        String type = invocationJson.getString(JsonConstants.RPC_TYPE);
+        if (type == null || !type.equals(JsonConstants.RPC_TYPE_EVENT)) {
+            return false;
+        }
```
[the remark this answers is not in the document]
taefi: To be honest, I'm personally in favor of having separate "simple and easier to read" `if` statements, so I wouldn't compress them into one if with a `!(...)`, but I inverted the condition of the second `if` to get rid of the unnecessary `type` variable definition and the `null` check.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
taefi: Although the `debug` level of logging should not be enabled in Production mode normally, I wrapped the `getLogger().debug` within the condition of being in "Development" mode. Also, added tests to verify the level of logging in Production and Development modes.

flow-server/src/test/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandlerTest.java
taefi: Done.

flow-server/src/test/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandlerTest.java
taefi: Added tests to make sure `getLogger.debug()` is only logging the ignored Payload in development mode.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
caalador: I would in this case say that the payload is much more interesting in production mode than dev mode as in production mode it means we are probably running a live server which someone is trying to send an unaccepted payload to and this shouldn't contain any information at all that would be non loggable for production mode.

Just to keep the feature clean I would drop anything you deem would need cheking execution mode for.

taefi: You are right, logging the ignored payload would not create a security risk as even if the attacker can access the server logs they would see nothing but their own malicious payload and no sensitive data.

Removed everything related to the execution mode. Changed the tests as well.

flow-server/src/test/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandlerTest.java
taefi: Done.

caalador (approved):

PR #13615, opened 2022-04-25
feat: Support @NonNullApi to define @Nonnull for all methods in a package

Artur-:
`@NonNullApi` can be used as a package level annotation, on `package-info.java`, to define that all method return types and parameter types should be considered `@Nonnull`. Individual parameter or return types can be overridden to allow null using`@Nullable`.

For https://github.com/vaadin/hilla/issues/107


fusion-endpoint/src/main/java/dev/hilla/ExplicitNullableTypeChecker.java
```diff
@@ -84,23 +111,28 @@ public static boolean isRequired(List<AnnotationExpr> annotations) {
...
+     * @param requiredByContext
+     *            {@code true} if the context defines that the node is required
+     *
      * @return error message when the value is null while the expected type does
      *         not explicitly allow null, or null meaning the value is OK.
      */
     public String checkValueForAnnotatedElement(Object value,
-            AnnotatedElement annotatedElement) {
-        if (!isRequired(annotatedElement)) {
+            AnnotatedElement annotatedElement, boolean requiredByContext) {
```
Artur-, line 121: This might need a test or two for the case where `requiredByContext` is `true` both for a `@Nullable` annotatedElement and a plain one

Artur-: Added

platosha (approved):

PR #13629, opened 2022-04-27
chore: use pull_request_target event and check-user-permissions to allow validations from external contributions

manolo:
Validations triggered from users with `read-only` permissions will fail sooner, then project owners can re-run the build once they verify that changes in the PR are not dangerous for the organisation.

Fixes https://github.com/vaadin/releases-team-tasks/issues/75

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -1,10 +1,12 @@
 name: Flow Validation
 on:
   push:
-    branches: [master, '9.0']
+    branches: [master, '23.1', '23.0', '9.0']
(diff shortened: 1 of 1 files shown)


Artur-:
I have no idea if the implementation is good or bad but the idea is good

ZheSun88:
as we have tested that it works... so let us get it in.

manolo:
For some reason GH actions does not work in this PR. 
Let's close and re-open to see if GH runs it.

ZheSun88 (approved):

PR #13634, opened 2022-04-27
fix: ignore server-modality for poll RPC invocations (CP: #13604)

taefi:
Manual CP to 23.0 branch to exclude
the breaking changes of #13604

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
@@ -15,16 +15,20 @@
  */
 package com.vaadin.flow.server.communication.rpc;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
+import com.vaadin.flow.component.ComponentUtil;
+import com.vaadin.flow.component.PollEvent;
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.internal.StateNode;
 import com.vaadin.flow.shared.JsonConstants;
 
 import elemental.json.JsonObject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Abstract invocation handler implementation with common methods.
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -61,7 +65,8 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
                     + "the client side for an inactive (disabled or invisible) node id='{}'",
                     getClass().getName(), node.getId());
             return Optional.empty();
-        } else if (!allowInert(node) && node.isInert()) {
+        } else if (!allowInert(node) && node.isInert()
+                && !ignoreInertForPollInvocations(ui, invocationJson)) {
```
mshabarov, line 69: Method `ignoreInertForPollInvocations` should return `false` to pass into "ignoring RPC for inert" code block.
This is misleading.
I would expect something like below for conditional:

```
} else if (!allowInert(node) && node.isInert()
                && ignoreInertForPollInvocations(ui, invocationJson)) {
```

or

```
} else if ((!allowInert(node) && node.isInert()) ||
                ignoreInertForPollInvocations(ui, invocationJson)) {
```

(from the methods naming point of view and semantic)

mshabarov: Maybe to reverse the returning values of `ignoreInertForPollInvocations`?

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
taefi: I think the previously existing block of code that logs "Ignored RPC for invocation ..." (regardless of the text being logged) is more of a "blocking" rather than an "ignoring", but I get with your point about the naming weirdness ;)

However, I would change the method name to something like `allowInvocationWhileInert` instead and not the return values nor the `if` conditions. I think this naming change should make the awkwardness go away :) Let me know if this is still not helping. 

Considering that having another method like `allowInert` is making the code confusing, I would mark it as deprecated if you show a thumbs up, as it doesn't clearly say what it does and has no javadoc. Its input parameters has already changed in the `master` (while keeping the same name, unfortunately). That also can/should be renamed to a more generic and self-descriptive name like `allowInvocationWhileInert` (but that is a separate issue).

mshabarov (approved):

PR #13666, opened 2022-05-03
fix: Only run new license checker when flag is enabled

Artur-:
The checker is invoked in three places:
1. After production build with the plugin
2. For fallback chunk in production build
3. Development time through dev mode gizmo

With this addition, all the three places are guarded by the feature flag

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
@@ -478,7 +478,8 @@ public class TaskUpdateImports extends NodeUpdater {
     }
 
     private Stream<String> filter(Stream<String> modules) {
-        if (!productionMode) {
+        if (!productionMode
+                || !featureFlags.isEnabled(FeatureFlags.NEW_LICENSE_CHECKER)) {
             return modules;
         }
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
```diff
@@ -478,7 +478,8 @@ private JsonArray makeFallbackModules(AbstractUpdateImports updater) {
     }
 
     private Stream<String> filter(Stream<String> modules) {
-        if (!productionMode) {
+        if (!productionMode
+                || !featureFlags.isEnabled(FeatureFlags.NEW_LICENSE_CHECKER)) {
```
mcollovati, line 482: If I correctly understand with this change we filter modules only if we are in production mode and the feature flag is enabled. If so, maybe inverting the condition may be easier to understand (at least for me)

Artur-: Aren't you not a not-fan of weird logic? Fixed

mcollovati: When I see more that one not, I need pen and paper to figure out how the condition works

mcollovati (approved):

PR #13670, opened 2022-05-03
Upgrade to license checker 1.4 for offline support

Artur-:


flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
```diff
@@ -499,7 +500,7 @@ private static void validateLicenses(PluginAdapterBase adapter) {
         for (Product component : commercialComponents) {
             try {
                 LicenseChecker.checkLicense(component.getName(),
-                        component.getVersion());
+                        component.getVersion(), BuildType.PRODUCTION);
```
manolo, line 503: It means that the bundle in dev-mode always contains the component? 
When developer realises that they don't have a valid license?

Artur-: When the license is checked by the logic added in https://github.com/vaadin/flow/pull/13601. In short: dev mode is checked in the browser

manolo: Gotcha, thanks for the clarification

manolo (approved):

mcollovati (approved):

PR #13674, opened 2022-05-04
refactor: Make ViewAccessControl more extensible

Artur-:
Makes it possible to fetch the principal and roles from other sources than the HTTP request

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
@@ -68,14 +68,6 @@ public class ViewAccessCheckerTest {
     }
 
     @Test
-    public void cannotUseWithoutServlet() {
-        Result request = setupRequest(AnonymousAllowedView.class, null, true);
-        CurrentInstance.clearAll();
-        this.viewAccessChecker.beforeEnter(request.event);
-        Assert.assertEquals(NotFoundException.class, request.getRerouteError());
-    }
-
-    @Test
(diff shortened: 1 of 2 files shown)


flow-server/src/test/java/com/vaadin/flow/server/auth/ViewAccessCheckerTest.java
```diff
@@ -67,14 +67,6 @@ public void init() {
         this.viewAccessChecker.setLoginView(TestLoginView.class);
     }
 
-    @Test
-    public void cannotUseWithoutServlet() {
```
mshabarov, line 71: Was this test deleted intentionally? I'd say it is still valid because when the user/principal is empty, the view access checker reroutes to `NotFoundException`.

Artur-: This test tests that the access checker does not work without a servlet. Now it works without a servlet

Artur-: What do you want to test? E.g. `anonymousAccessToPermitAllViewDenied` checks what happens when principal is empty and you go to a `PermitAll` view

mshabarov: Indeed, `HttpServletRequest` is not used for access check anymore. This test's codes has no explicit "servlet" related setups, so I could judge only based on the test name. But okay, now I see that it's not needed.

mshabarov (approved):

PR #13677, opened 2022-05-04
fix: Ensure @HandlesTypes types are not included in the collection

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/startup/ErrorNavigationTargetInitializer.java
```diff
@@ -40,9 +40,8 @@
...
     @Override
     public void initialize(Set<Class<?>> classSet, VaadinContext context) {
-        if (classSet == null) {
-            classSet = new HashSet<>();
-        }
+        classSet = AbstractAnnotationValidator
+                .removeHandleTypesSelfReferences(classSet, this);
         Set<Class<? extends Component>> routes = classSet.stream()
                 // Liberty 18 also includes the interface itself in the set...
                 .filter(clazz -> clazz != HasErrorParameter.class)
```
caalador, line 47: This can be removed now as is done for VaadinAppShellInitializer interface filters.

caalador (changes requested):

caalador (approved):

PR #13679, opened 2022-05-04
chore!: Rename dev mode gizmo to dev tools

Artur-:


mshabarov (approved):
@Artur- is there a good reason to pick this change to 23.1 ? Since it's just a naming, I'd not introduce breaking changes into beta version.

Artur-:
The only reason would be to reduce backporting issues by keeping the branches as similar as possible

PR #13711, opened 2022-05-09
feat: add flag for CE backend support

heruan:
This is to add a feature-flag for Collaboration Engine backend support (aka clustering support).


mshabarov:
@heruan does the feature need to be included into `V23.1`? If so, we have to add a `target/23.1` label.

heruan:
It is, added!

mshabarov (approved):

PR #13724, opened 2022-05-10
Allow dev mode startup spinner to animate fully

jouni:
Instead of 1/3 of the animation, allow it to run fully before reloading the page.

This change makes it spin really fast, but at the moment it doesn't look too nice either, when the animation is cut short and then it starts again.


Artur-:
If the page is reloaded at all, then something else is wrong

jouni:
Well, it indeed has. Nice 👍 

https://github.com/vaadin/flow/commit/cd8b9bef5499184bd8428a05b91326a74fa03d6e

PR #13729, opened 2022-05-11
fix: Clean up flux subscription on browser disconnect

Artur-:
When the browser disconnects, we need to close all subscriptions. With socket.io, the browser will not reuse the same socket when it reconnects so on reconnect there will be new subscriptions created

Test based on #13728

### DIFF
--- a/fusion-endpoint/src/main/java/dev/hilla/push/PushMessageHandler.java
+++ b/fusion-endpoint/src/main/java/dev/hilla/push/PushMessageHandler.java
@@ -1,7 +1,7 @@
 package dev.hilla.push;
 
 import java.security.Principal;
-import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


fusion-endpoint/src/main/java/dev/hilla/push/PushMessageHandler.java
```diff
@@ -149,14 +165,62 @@ private void send(Consumer<AbstractClientMessage> sender,
...
+     *            the connection id
+     * @param subscriptionId
+     *            the subscription id
+     */
+    private void disposeSubscriptionInfo(String connectionId,
+            String subscriptionId) {
+        ConcurrentHashMap<String, Disposable> fluxMap = fluxSubscriptionDisposables
+                .get(connectionId);
+        if (fluxMap != null) {
+            Disposable subscriptionInfo = fluxMap.remove(subscriptionId);
```
cromoteca, line 209: Concurrency is not handled between the `remove` and `isEmpty` calls. Anyway, I can't see a path leading to issues, as one of the simultaneous calls to this method should still be returned `true` by `isEmpty`, and if more than one get `true`, executing `fluxSubscriptionDisposables.remove` twice is harmless.

cromoteca (dismissed):

fusion-endpoint/src/main/java/dev/hilla/push/PushMessageHandler.java
Artur-: Could a potential problem come between `isEmpty` and `remove` below? So `isEmpty` returns `true` and then another thread adds a subscription to the map. After that, we remove the whole map from `fluxSubscriptionDisposables` in the first thread and the added subscription info is lost. What the code tries to guard against by passing `fluxMap` to `remove` but wouldn't the `fluxMap` instance still be the same even if another subscription has been added...

Artur-: Let me change it to not use a nested map, then it should be ok

Artur-: Ok so using a nested map would mean that we need to iterate through the whole map when the browser disconnects. Instead I changed the connection info to be created when the browser connects and disposed only when the browser disconnects. It will use slightly more memory if a browser is connected and has no open subscriptions but that should not be an issue at all.

cromoteca (approved):

PR #13735, opened 2022-05-11
Improve dev mode startup message and styling

jouni:
Simplify the message. Avoid "please", just state what is happening.

Avoid the ellipsis character as it might imply truncated text. The spinner indicates ongoing progress.

Center the message within the viewport. Adapt to light and dark OS theme automatically.

Before:

Light | Dark
---|---
Screen Shot 2022-05-11 at 12 38 45|Screen Shot 2022-05-11 at 12 38 45 |

After:
Light | Dark
---|---
Screen Shot 2022-05-11 at 12 35 44 | Screen Shot 2022-05-11 at 12 36 00

### DIFF
--- a/vaadin-dev-server/src/main/resources/com/vaadin/base/devserver/dev-mode-not-ready.html
+++ b/vaadin-dev-server/src/main/resources/com/vaadin/base/devserver/dev-mode-not-ready.html
@@ -9,7 +9,7 @@
     justify-content: center;
     margin: 0;
     padding: 1rem;
-    min-height: 100%;
+    min-height: 100vh;
     box-sizing: border-box;
   }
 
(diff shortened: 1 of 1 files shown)


mshabarov (approved):

knoobie:
I was hoping for this to come to 23.1 as well - it looks way better (especially with the dark mode). Totally missed in after my rc.1 migration

mshabarov:
@knoobie flow was in beta while this was implemented, and it wasn't that critical (however indeed improves the DX) to be picked to 23.1. Now we have an RC and it's even more late to pick it.

knoobie:
Sadly true - so that's something to look forward to the first 23.2.alpha ;)

PR #13746, opened 2022-05-12
fix: Avoid potential slow Set#removeAll call in HierarchicalCommunicator

### LINKED ISSUE #13745: Slow UIDL generation when scrolling through large grids
### Description of the bug

`HierarchicalCommunicationController#passivateInactiveKeys` calls `removeAll` on an `AbstractSet` implementation passing in a a `List`. This implementation of `removeAll` compares the size of the set with the size of the collection parameter. If the set is larger than the collection, it loops over the collection and calls `remove` in a loop. In the other case, it iterates over the set, calls `contains` on the collection for each item, and then calls `Iterator.remove`. This second code path leads to very poor performance when `passivateInactiveKeys` is called with fairly large collections since `List#contains` has to do a linear search for ever item in the Set. Worst case scenario this is `O(n * m)` in time complexity. It's significant enough to show up as a 25% hotspot during profiling.

### Expected behavior

`O(n * m)` code paths are avoided.

I noticed this while testing our application with a TreeGrid contains about 10k rows and scrolling up and down rapidly.

### Versions

- Vaadin / Flow version: 23.0.7
- Java version: 11.0.15
- OS version: macOS
- Browser version (if applicable): Safari 15.4
- Application Server (if applicable): Jetty 9.4.36
- IDE (if applicable): N/A

pepijnve:
## Description

The performance of calling Set#removeAll(List) is dependent on the relative sizes of the Set and List parameter. Replace Set#removeAll with List#forEach(Set::remove) to avoid this.

Fixes #13745

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
@@ -287,7 +287,7 @@ public class HierarchicalCommunicationController<T> implements Serializable {
             }
 
             // Finally clear any passivated items that have now been confirmed
-            oldActive.removeAll(newActiveKeyOrder);
+            newActiveKeyOrder.forEach(oldActive::remove);
             if (!oldActive.isEmpty()) {
                 passivatedByUpdate.put(Integer.valueOf(updateId), oldActive);
             }
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
@@ -287,6 +287,13 @@ public class HierarchicalCommunicationController<T> implements Serializable {
             }
 
             // Finally clear any passivated items that have now been confirmed
+            /*
+             * Due to the implementation of Set#removeAll, if the size of 
+             * newActiveKeyOrder is bigger than oldActive's size, then 
+             * calling oldActive.removeAll(newActiveKeyOrder) would end 
+             * up calling contains for all of newActiveKeyOrder items 
+             * which is a slow operation on lists. The following avoids that:
+             */
             newActiveKeyOrder.forEach(oldActive::remove);
             if (!oldActive.isEmpty()) {
                 passivatedByUpdate.put(Integer.valueOf(updateId), oldActive);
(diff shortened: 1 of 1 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
```diff
@@ -287,7 +287,7 @@ private void passivateInactiveKeys(Set<String> oldActive,
             }
 
             // Finally clear any passivated items that have now been confirmed
-            oldActive.removeAll(newActiveKeyOrder);
+            newActiveKeyOrder.forEach(oldActive::remove);
```
taefi, line 290:
```suggestion
            /*
             * Due to the implementation of Set#removeAll, if the size of 
             * newActiveKeyOrder is bigger than oldActive's size, then 
             * calling oldActive.removeAll(newActiveKeyOrder) would end 
             * up calling contains for all of newActiveKeyOrder items 
             * which is a slow operation on lists. The following avoids that:
             */
            newActiveKeyOrder.forEach(oldActive::remove);
```

taefi (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
```diff
@@ -287,7 +287,14 @@ private void passivateInactiveKeys(Set<String> oldActive,
...
 
             // Finally clear any passivated items that have now been confirmed
-            oldActive.removeAll(newActiveKeyOrder);
+            /*
+             * Due to the implementation of Set#removeAll, if the size of 
+             * newActiveKeyOrder is bigger than oldActive's size, then 
+             * calling oldActive.removeAll(newActiveKeyOrder) would end 
+             * up calling contains for all of newActiveKeyOrder items 
+             * which is a slow operation on lists. The following avoids that:
+             */
```
taefi, line 296:
```suggestion
            /*
             * Due to the implementation of AbstractSet#removeAll, if the size
             * of newActiveKeyOrder is bigger than oldActive's size, then
             * calling oldActive.removeAll(newActiveKeyOrder) would end up
             * calling contains method for all of newActiveKeyOrder items which
             * is a slow operation on lists. The following avoids that:
             */
```

taefi (changes requested):

taefi (approved):

PR #13773, opened 2022-05-13
fix: Do not serve static resources from dev server if dev server startup failed

### LINKED ISSUE #13772: Exception from waitFor future not handled in AbstractDevServerRunner
### Description of the bug

If there is a problem when running the future passed to the AbstractDevServerRunner constructor (in practice `DevModeInitializer.runNodeTasks`) then `runOnFutureComplete` is never invoked and the dev server is never started.

You see this in practice as
```
java.net.ConnectException: Can't assign requested address (connect failed)
	java.base/java.net.PlainSocketImpl.socketConnect(Native Method)
	java.base/java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:399)
	java.base/java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:242)
	java.base/java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:224)
	java.base/java.net.Socket.connect(Socket.java:609)
	java.base/sun.net.NetworkClient.doConnect(NetworkClient.java:177)
```

because the webpack or vite server port remains at the initial value of `0`

### Expected behavior

The actual exception should be reported instead of lost. In my case it was

```
java.io.IOException: Unable to delete files: /.../hello/node_modules/.pnpm/workbox-strategies@6.5.3/node_modules/workbox-strategies/utils, ...
```

-

### Versions

Vaadin 23

Artur-:
This happens when the dev server startup fails, i.e. DevModeInitializer.runNodeTasks throws an exception,
then "action" in the constructor throws an exception and devServerStartFuture completes with an exception.
After this when the browser reloads, it needs to go to AbstractDevModeHandler.handleRequest which checks
if an exception happened and sets isDevServerFailedToStart accordingly.

However before getting to AbstractDevModeHandler.handleRequest the static file server checks if the requested
path is a static resource by calling AbstractDevModeHandler.serveDevModeRequest. This throws an exception
because it only checks that the dev server is not still starting and then it tries to access it using port 0.

Fixes #13772

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/AbstractDevServerRunner.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/AbstractDevServerRunner.java
@@ -671,7 +671,8 @@ public abstract class AbstractDevServerRunner implements DevModeHandler {
     public boolean serveDevModeRequest(HttpServletRequest request,
             HttpServletResponse response) throws IOException {
         // Do not serve requests if dev server starting or failed to start.
-        if (isDevServerFailedToStart.get() || !devServerStartFuture.isDone()) {
+        if (isDevServerFailedToStart.get() || !devServerStartFuture.isDone()
+                || devServerStartFuture.isCompletedExceptionally()) {
             return false;
         }
         // Since we have 'publicPath=/VAADIN/' in the dev server config,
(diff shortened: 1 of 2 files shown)


mshabarov (approved):

Artur-:
Fixed formatting...

mshabarov (dismissed):

Artur-:
Fixed test so that the dummy server seemingly starts correctly

mshabarov (approved):

PR #13786, opened 2022-05-16
fix: Read url mapping using the correct property name

Artur-:
The property in the application.properties file is vaadin.url-mapping while the old code read a non-existant vaadin.urlMapping property


tarekoraby:
Do the docs need to change to `url-mapping` as well: https://vaadin.com/docs/latest/flow/integrations/spring/configuration/#using-spring-boot-properties?

PR #13795, opened 2022-05-18
fix!: Always load static files from context root

### LINKED ISSUE #13769: Always serve static files from VAADIN in the context root
### Description of the bug

In most cases, you have only one servlet deployed as `/*` in your application.

In cases when you have other web frameworks in use, REST endpoints etc you might map your application to `/myapp/*` instead.

In the rare cases where you have multiple Vaadin servlets in the same application, you might have `/myapp/*` and `/admin/*` mappings.

Right now the bundle files are loaded from 
- `/VAADIN` in the `/*` mapping case
- `/myapp/VAADIN` in the `/myapp/*` mapping case
- `/myapp/VAADIN` AND `/admin/VAADIN` in the multi mapping case

### Expected behavior

In all cases, the static files (JS bundle files, vaadinPush script etc) should be loaded from `VAADIN` inside the context root. Thus is you have multiple servlet mappings, you still load e.g. web components only once from `/VAADIN` and not from both `/myapp/VAADIN` and `/admin/VAADIN`.

Any dynamic content related to the servlet should still be served through `/myapp/VAADIN/dynamic`.

-

### Versions

Vaadin: 23.0.9
Flow: 23.0.7
Java: Homebrew 17.0.1
OS: aarch64 Mac OS X 12.3.1
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/101.0.4951.54 Safari/537.36

Artur-:
Fixes #13769

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/communication/AtmospherePushConnection.java
+++ b/flow-client/src/main/java/com/vaadin/client/communication/AtmospherePushConnection.java
@@ -727,7 +727,7 @@ public class AtmospherePushConnection implements PushConnection {
             Console.log("Loading " + pushJs);
             ResourceLoader loader = registry.getResourceLoader();
             String pushScriptUrl = registry.getApplicationConfiguration()
-                    .getServiceUrl() + pushJs;
+                    .getContextRootUrl() + pushJs;
             ResourceLoadListener loadListener = new ResourceLoadListener() {
                 @Override
                 public void onLoad(ResourceLoadEvent event) {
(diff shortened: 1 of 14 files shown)


mshabarov (changes requested):
The starter project with activated Vite now works, when this patch is applied.
However, web sockets doesn't work, I see this error in the browser console:

`WebSocket connection to 'ws://localhost:8080/vaadin/?v-r=push&debug_window' failed:` 

Apparently something is missing for proper web sockets handling.

mshabarov:
Works well without web socket error with any mapping except `/vaadin/*`.
Looks like a case sensitive issue: `/vaadin/*` is handles as `/VAADIN/*` with no difference.

mshabarov (approved):
The WS issue will be addressed via https://github.com/vaadin/flow/issues/13839

PR #13859, opened 2022-05-25
Move license checker dependency to vaadin-dev-server and plugin

Artur-:
This makes it clearer that Flow (flow-server) does not depend on the license checker nor is it needed in production


mshabarov (approved):

Artur-:
in #13861

PR #13862, opened 2022-05-25
fix: Fix the theme files resolution for Vite apps

vursen:
## Description

The PR fixes the theme files resolution for Vite apps by adding a custom resolver for the paths starting with `themes`. The custom resolver tries to sequentially resolve the path inside the folders (1) `flow-frontend` and (2) `frontend`, returning the first successful resolution.

Fixes https://github.com/vaadin/flow/issues/13538

A follow-up to https://github.com/vaadin/flow/pull/13543

## Type of change

- [x] Bugfix

## Checklist


flow-tests/test-application-theme/test-reusable-as-parent-vite/vite.config.ts
```diff
@@ -0,0 +1,12 @@
+import { UserConfigFn } from 'vite';
+import { overrideVaadinConfig } from './vite.generated';
+
+const customConfig: UserConfigFn = (env) => ({
+  // Here you can add custom Vite parameters
+  // https://vitejs.dev/config/
+  build: {
+    assetsInlineLimit: 0
```
Artur-, line 8: What does this one do? Should it be in all projects or for some weird reason just in this test project?

vursen: We could consider disabling this limit for the Vite config in order to keep it as close as possible to the Webpack config.

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -354,14 +354,21 @@ export const vaadinConfig: UserConfigFn = (env) => {
     base: '',
     resolve: {
       alias: [
-        { find: 'themes', replacement: (importee: string) => {
-            if (existsSync(path.resolve(themeResourceFolder, importee))) {
-              return path.resolve(themeResourceFolder, settings.themeFolder);
+        {
+          find: 'themes',
+          replacement: settings.themeFolder,
```
Artur-, line 359: settings.themeFolder is always(?) "themes" so does this replace `themes` with `themes`? What's the relationship between `replacement` and `customResolver`?

vursen: Good point. There seems to be no practical sense in using aliases for the `themes` case. Moved the resolution logic to a separate `themePlugin`.

Artur- (approved):

PR #13879, opened 2022-05-31
fix: Fix Vite to use vaadin/bundles on Windows

### LINKED ISSUE #13525: Vite does not use the component bundle on Windows
### Description of the bug

When you start any app with Vite on Windows, the bundle is ignored

The reason is that the path matching in vite.generated.ts fails:
```
    load(rawId) {
      const [path, params] = rawId.split('?');
      if (!path.startsWith(modulesDirectory)) return;
```

On Windows, the path is like `C:/Users/Foo/Downloads/my-app/node_modules/@vaadin/list-box/src/vaadin-list-box.js` and `modulesDirectory` like `/Users/Foo/Downloads/my-app/C:\Users\Foo\Downloads\my-app/node_modules` (yes, the path is kind of duplicated and uses a mix of `/` and `\`)

### Expected behavior

Components should be loaded from `@vaadin/bundles`

```
vaadin init --hilla hello # Using Hilla only because it uses Vite by default
cd hello
mvn
```

### Versions

- Vaadin / Flow version: 23.0.5
- Java version: 11
- OS version: Windows

taefi:
## Description

Fix resolving the `node_modules` directory path to 
enable the `@vaadin/bundles` usage while using Vite 
in Windows as well.

Fixes #13525

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [x] I have performed a self-review and corrected misspellings.


mshabarov (dismissed):

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -180,7 +180,9 @@ function vaadinBundlesPlugin(): PluginOption {
 
   const disabledMessage = 'Vaadin component dependency bundles are disabled.';
 
-  const modulesDirectory = path.posix.resolve(__dirname, 'node_modules');
+  const modulesDirectory = path.resolve(__dirname, 'node_modules');
+  // always refine path to comply with URI path separator (forward slash) for requested resources:
+  const modulesDirPathSepRefined = path.sep === '/' ? modulesDirectory : modulesDirectory.replace(/\\/g, '/');
```
caalador, line 185: This is a bit extra as it could just be 
```suggestion
  const modulesDirectory = path.resolve(__dirname, 'node_modules').replace(/\\/g, '/');
```

mshabarov (approved):

mshabarov (approved):

PR #13902, opened 2022-06-03
feat: notify component about identifier provider changes

sissbruecker:
## Description

Components like `Grid` or `MultiSelectComboBox` are supposed to reuse the identifier provider of the Flow data classes to identify items, for example to manage a selection state. However retrieving the identifier provider from a component is problematic at the moment as developers can set an identifier provider on any data view, and there is no way to react to that apart from extending individual implementations of data view classes to implement a custom notification mechanism that the `setIdentifierProvider` method was called.

This change introduces a `IdentifierProviderChangeEvent` that is fired when the developer sets an identifier provider on a data view. Components like `Grid` or `MultiSelectComboBox` can listen for this event, and retrieve the new identifier provider from it.

## Type of change

- [x] Feature

## Checklist

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
@@ -164,6 +164,28 @@ public abstract class AbstractDataView<T> implements DataView<T> {
         }
     }
 
+    /**
+     * Add an identifier provider change listener that is fired when a custom
+     * identifier provider is set with
+     * {@link #setIdentifierProvider(IdentifierProvider)}.
+     * <p>
+     * Can be used by components to get notified that a new identifier provider
+     * has been set through the data view.
+     *
+     * @param listener
+     *            identifier provider change listener to register
+     * @return registration for removing the listener
+     */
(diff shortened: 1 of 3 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
```diff
@@ -143,6 +143,8 @@ public void setIdentifierProvider(
                 "Item identity provider cannot be null");
         ComponentUtil.setData(component, IdentifierProvider.class,
                 identifierProvider);
+        ComponentUtil.fireEvent(component, new IdentifierProviderChangeEvent<>(
```
mshabarov, line 146: It makes sense to add `addIdentityProviderChangeListener` new method to this class and call `ComponentUtil.addListener` in it, so you can subscribe your component to listen these events through the DataView API.

mshabarov (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
sissbruecker: Done

mshabarov (approved):

PR #13903, opened 2022-06-03
Annotate OptionalParameter with @Nullable

tarekoraby:
To make it clear that `OptionalParameter` can be null (with the benefit of providing better hints from IDEs and code analyzers).


manolo:
Probably an unit test for checking the change works would be useful

tarekoraby:
Not sure how to test this since it's something that is mainly useful for IDEs.

tarekoraby:
Basically, there is no changed application behavior involved in this PR (only a change in IDE behavior). Hence, I think no testing is needed.

mshabarov (approved):

PR #13906, opened 2022-06-03
feat: Add API to modify a view after UI.navigate

mstahv:
Allows users to configure the target view, with a proper Java API provided by the target view itself, instead of forcing to use query/path parameters (which ruins the power of having a language like Java in use).

A playground project that uses this change and demonstrates the usage/power of this change:
https://github.com/mstahv/typed-navigation-playground

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


knoobie:
I really like this addition; especially if this could be combined with the overloaded navigate methods to pass "security" related things without having to worry about URL params or already available data from one view to another.

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -845,9 +845,34 @@ public Page getPage() {
...
+        return Optional.empty();
+    }
+
+    /**
+     * Alternative version without Optional (that I'm allergic to :-) )
+     * @param navigationTarget
+     * @return
+     * @param <T>
+     */
+    public <T extends Component> T navigateTo(Class<T> navigationTarget) {
```
mshabarov, line 867: I see no worth to have this override if the "optional" version already exists. IMO it should be excluded.

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -845,9 +845,34 @@ public Page getPage() {
      *             navigationTarget.
      * @see #navigate(Class, Object)
      * @see #navigate(Class, RouteParameters)
+     * @return the view target instance, if navigation actaully happened
```
mshabarov, line 848:
```suggestion
     * @return the navigation target instance, if navigation actually happened without errors, otherwise {@link Optional#empty()}
```

mshabarov (changes requested):
This is a draft, but could you also make a test case showing the usage of the returned navigation target class? Why is not just a `boolean`?

mshabarov:
> This is a draft, but could you also make a test case showing the usage of the returned navigation target class? Why is not just a boolean?

Ok, I checked your example project and it gives a clue how that can be used.

mshabarov:
> Edit: it probably has to be documented that those methods called on the view are executed before? enter listener are called?

Good point, the documentation should explain the order of calling navigation target's methods VS navigation lifecycle.

flow-server/src/main/java/com/vaadin/flow/component/UI.java
mstahv: The actual body of the docs talks about views, which I like more than "navigation targets", if it is ok, I'd leave it like this here.  Best would be to unify the terminology in docs/parameter names and some point in single change.

I don't think we should discuss about Optional at all here, even though it is used. I think the documentation just becomes more complex if we start using it. We should be able to think of it as a core JDK feature that users know (or then not use it at all).

mstahv:
I'm pretty much ready with the PR to Flow itself. I think both @caalador & @mshabarov should review, but up to the team of course.

Are we confident enough about the change so that I can start drafting changes to the vaadin.com/docs ?

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -843,11 +848,25 @@ public Page getPage() {
      * @throws NotFoundException
      *             in case there is no route defined for the given
      *             navigationTarget.
+     * @return the view instance, if navigation actually happened
      * @see #navigate(Class, Object)
      * @see #navigate(Class, RouteParameters)
      */
-    public void navigate(Class<? extends Component> navigationTarget) {
-        navigate(navigationTarget, RouteParameters.empty());
+    public <T extends Component> Optional<T> navigate(
```
mshabarov, line 855: Instead of removing `void` methods, I propose to deprecate them.

flow-server/src/test/java/com/vaadin/flow/component/UITest.java
```diff
@@ -328,9 +328,11 @@ public void navigateWithParameters_afterServerNavigation()
         UI ui = new UI();
         initUI(ui, "", null);
 
-        ui.navigate(FooBarParamNavigationTarget.class,
+        Optional<FooBarParamNavigationTarget> newView = ui.navigate(FooBarParamNavigationTarget.class,
```
mshabarov, line 331: I still don't see a test showing the use case.
I propose to create an IT test similar to that one you have in your project, `AnotherView `, set the object to the view and check that the corresponding UI component appears on the page.
Without it, IMO almost no chance to understand why this feature is needed.

mshabarov (changes requested):
I agree to apply this feature.
Have a couple of comments.

Also `mvn formatter:format` is needed.

flow-server/src/test/java/com/vaadin/flow/component/UITest.java
mstahv: OK. Which directory/project should I use for the example use case? I was thinking to add something similar to docs and add the use case there.

flow-server/src/main/java/com/vaadin/flow/component/UI.java
mshabarov: Ah, sorry, ignore me. I was proposing to preserve the old methods, but since they have the same signature, we cannot do that.

But this give me a clue that we probably should postpone it to V24, because of binary incompatibility. This method is broadly used, so there can be an issues.

mstahv: Eh, this is why I had drafted with different name earlier. I think I had "navigateTo" as an option. Might be an option to add as such to 23.2. 23 series apps will be "alive" for a long time...

I wonder how many things would actually break? Can UI be easily extended these days 🤔

mstahv:
@knoobie You mentioned you'd prefer the version without Optional. The patch is now derived into direction that uses optional, based on offline discussions with @mshabarov. 

It would be last moments to convince the @vaadin/flow team to leave the Optional away 😉 After that discussion, I'd attack on documentations improvements.

knoobie:
@mstahv like you, I really don't like having optional everywhere. For example that `button.getUI()` returns an optional feels clunky, even tho I understand the idea and the semantic meaning behind it. 

I thought a little about this case of optional as return value as well and after some back and forth, I think here is it worth it (instead of the button example from above). 

Example use cases where optional makes sense is when the operation can fail. 
A navigation can fail, if e.g. the user gets redirected inside the before/after view navigation event based on the roles. 

```java

navigate(SecureView.class).ifPresentOrElse(securedView -> {
   securedView.setParamThatShouldNotBeSeenInTheURL("123");
   }, () -> {
      // user has no right to see the secure view and was redirected to another view or redirect aborted
      log.info("User could not be navigated...");
    })

```

This addition could also be used inside the testbench uitest project, see https://github.com/vaadin/testbench/pull/1460

flow-server/pom.xml
```diff
@@ -242,6 +242,27 @@
                     </excludedArtifactIds>
                 </configuration>
             </plugin>
+			<plugin>
```
mshabarov, line 245: Would be better to define this plugin in the root pom.xml in plugin management section and when add it to flow-server without version and executions? I believe other modules of Flow potentially can also need this.

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -843,11 +848,35 @@ public Page getPage() {
...
+            Class<T> navigationTarget) {
+        return navigate(navigationTarget, RouteParameters.empty());
+    }
+
+    /**
+     * For backwards compatibility, to be removed in V24
+     *
+     * @hidden
+     */
+    public <T extends Component> void navigate$$bridge(
```
mshabarov, line 865: I'm quite sure that in the byte code it would resolve into the `Component` anyway, but the previous signature was:
`public void navigate(Class<? extends Component> navigationTarget)`, so should we keep it the same, just in case?

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -907,11 +958,22 @@ public <T, C extends Component & HasUrlParameter<T>> void navigate(
...
+        return findCurrentNavigationTarget(navigationTarget);
+    }
+
+    /**
+     * For backwards compatibility, to be removed in V24
+     *
+     * @hidden
+     */
+    public <T extends Component> void navigate$$bridge(
+            Class<T> navigationTarget, RouteParameters parameters) {
```
mshabarov, line 975: Same, signature was `public void navigate(Class<? extends Component> navigationTarget,
            RouteParameters parameters) {`

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/UI.java
mstahv: In bytecode it is just Class. Changed to old one anyways.

flow-server/pom.xml
mstahv: Also updated to the latest version that don't need the ASM version declared.

I'd aim to remove this plugin in 24, unless we end up having some similar case.

mstahv:
Started to think about how/where to in the current navigation docs this should be added. In the current documentation, "postponing" navigation is in a rather central role (mostly because the whole chapter is  "sub-optimally" organised/ordered, cc @tarekoraby). This new functionality is not compatible with "postpone functionality", which made me give yet another look at an alternative implementation that I thought of originally and which I also discussed with @taefi offline. In this solution developer would pass SerializableConsumer<TargetView> as an additional parameter to the navigate method (which would be called after the navigation has happened).

Documenting here why I'm not pushing to that direction:

 * I have never used "postponing" myself and I assume it is fairly rarely used feature
 * Using UI.navigate(), happens most likely from the current view that would do the postponing as well, so developers do have other measures to prevent navigation if needed
 * Implementation would become larger and more complex (SerializableConsumer would need to be passed around and saved in "navigation context"
 * There would be a higher risk of memory leaks developers wouldn't necessary figure out
 * The code end users write would be less of a "story"

mshabarov (approved):

mshabarov (changes requested):
> This new functionality is not compatible with "postpone functionality"

This one sounds scary. I see no direct connection between "postpone" and modifying view. So if this change was about improving "postponing", it would be fine. But in practice it seems we sacrifice one feature in favour of another. Not sure it's a proper way. By the way, how all tests passing if "postpone" is not compatible?

mstahv:
It does not break anything, works as expected. If somebody calls "BeforeLeaveEvent.postpone()", the return value is empty optional -> the new API does nothing. And it does nothing if the navigation is continued later on.

mshabarov:
> It does not break anything, works as expected. If somebody calls "BeforeLeaveEvent.postpone()", the return value is empty optional -> the new API does nothing. And it does nothing if the navigation is continued later on.

Thanks for clarifications. Well, then it's reasonable: if the navigation postponed, the view is empty. And javadocs mention that.

mshabarov (approved):
LGTM, feel free to merge if you won't find any new obstacles during documentation development.

tarekoraby:
@mstahv the structure of the Routing sections in the docs is an important topic, and I don't want it to be hidden here. I suggest starting the documentation-related discussion in a separate ticket/PR in the [docs repo](https://github.com/vaadin/docs).

PR #13924, opened 2022-06-07
refactor: make vaadin-spring not depend on fusion-endpoint

platosha:
Connected to: vaadin/hilla#208

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/hilla/EndpointRequestUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/hilla/EndpointRequestUtil.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/hilla/EndpointRequestUtil.java
```diff
@@ -0,0 +1,53 @@
...
+import java.io.Serializable;
+import java.lang.reflect.Method;
+import java.util.Optional;
+
+/**
+ * A container for utility methods related with Hilla endpoints.
+ *
+ * @since 23.2
+ */
+public interface EndpointRequestUtil extends Serializable {
```
platosha, line 29: I chose to put this interface to `flow-server` mainly because it’s a common base dependency for both `fusion-endpoint` and `vaadin-spring`. Feel free to suggest better places for it, if any.

platosha, line 29: I chose to put this interface to `flow-server` mainly because it’s a common base dependency for both `fusion-endpoint` and `vaadin-spring`. Feel free to suggest better places for it, if any.

Artur-: Endpoints have no relevance for Flow so it should be in a Hilla module, e.g. fusion-endpoint

platosha: That would be exactly the opposite of making `vaadin-spring` not depend on `fusion-endpoint`...

mshabarov:
@platosha sorry for keeping this PR for a while without feedback, we'll make the review asap, hopefully tomorrow.

mshabarov (approved):

PR #13932, opened 2022-06-08
chore: Change license checker feature flag to concern only offline

Artur-:


Artur-:
This should be only done in master

mshabarov (approved):
Let's see how the validation of Flow and add-ons pass.

mshabarov (approved):

PR #13936, opened 2022-06-09
refactor: make vaadin-dev-server not depend on fusion-endpoint

platosha:
Connected to: vaadin/hilla#208

Removes dependency on `EnpointGeneratorTaskLibraryImpl` in `vaadin-dev-server` tests.

### DIFF
--- a/vaadin-dev-server/pom.xml
+++ b/vaadin-dev-server/pom.xml
@@ -47,13 +47,6 @@
         </dependency>
 
         <dependency>
-            <groupId>com.vaadin</groupId>
-            <artifactId>fusion-endpoint</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>annotations</artifactId>
             <version>3.0.1</version>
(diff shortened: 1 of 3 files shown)


Artur- (commented):
I am not sure what the relevance of DevModeEndpointTest is after this. What does it test?

Artur-:
Does it actually want to test only that this line is in place https://github.com/vaadin/flow/blob/master/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java#L289 and it does it through 3 different test methods and using a custom task?

platosha:
> Does it actually want to test only that this line is in place https://github.com/vaadin/flow/blob/master/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java#L289 and it does it through 3 different test methods and using a custom task?

Yes.

platosha:
Should we just assert the invocation using mocks? Or move the test to the integration module?

mshabarov:
@platosha sorry for keeping this PR for a while without feedback, we'll make the review asap, hopefully tomorrow.

mshabarov:
> Should we just assert the invocation using mocks? Or move the test to the integration module?

IMO it can be a unit test, because it checks that the API is generated after dev server start. I doesn't need to be a end-to-end having browser, user interaction or things like that.

mshabarov (approved):

PR #13940, opened 2022-06-09
feat: enable components to notify binder when validation state changes

### LINKED ISSUE #8242: Binder should always validate after field's internal validation
Currently `Binder` validates fields on `ValueChangeEvent`, because the valid/invalid -state is based only on the field's server-side value.

However, some fields should be also validated when the field is blurred and only the presentation value has changed. For example, when entering "asdf" to an empty `DatePicker`, the server-side value remains `null`, but the validation-state should change from valid to invalid. Same applies to eg. non-parseable numbers entered in number fields. This kind of validation should be implemented by the components themselves, but `Binder` should also re-validate, so that all validators are effective.

For more details, see: https://github.com/vaadin/vaadin-date-picker-flow/issues/223#issuecomment-623386532

Possible solutions:
- Make `Binder` validate also on `blur`, if field implements `BlurNotifier` (although academically incorrect)
- Add an event fired by `HasValidator` fields to notify `Binder` that it should validate

This issue would be also good to get fixed: https://github.com/vaadin/flow/issues/8241

One more thing to consider is the inconsistency between Java and web component users:
The UX for web components has been designed so that "visiting" (focus and blur even without changing value) is considered "enough" to trigger validation (https://github.com/vaadin/vaadin-text-field-flow/issues/193#issuecomment-502631352).
Java forms and fields on the other hand validate only on value changes, which can mean many situations based on the `ValueChangeMode`, but it never means visiting a field without changing anything.
This inconsistency might become more disturbing with Vaadin 15+, where a single application may contain views written in Java, and others created with TypeScript.

taefi:
## Description
Adds the needed API to notify Binder about validation status changes that happen in field components.

Fixes #8242 
Related to https://github.com/vaadin/flow-components/issues/1158

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-data/src/main/java/com/vaadin/flow/data/binder/ValidationStatusChangeListener.java
```diff
@@ -0,0 +1,44 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.data.binder;
+
+import java.io.Serializable;
+
+/**
+ * The listener interface for receiving {@link ValidationStatusChangeEvent}
```
mshabarov, line 21: Do you expect if any other classes, rather than a `Binder`, would use this API to react to validation status change listener? For example, if someone would want to modify the view/page according to a validation status of the field, e.g. put a banner, disable buttons etc. ? If we expect such a use cases, then maybe worth to describe them here.

mshabarov (changes requested):

.github/workflows/validation.yml
```diff
@@ -7,7 +7,7 @@ on:
     types: [opened, synchronize, reopened]
 jobs:
   build:
-    timeout-minutes: 15
+    timeout-minutes: 20
```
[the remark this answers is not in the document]
taefi: This is moved to: https://github.com/vaadin/flow/pull/13973

flow-data/src/main/java/com/vaadin/flow/data/binder/ValidationStatusChangeListener.java
taefi: More descriptions were added to announce the primary purpose and other possibilities through the newly added API.

flow-data/src/main/java/com/vaadin/flow/data/binder/HasValidator.java
```diff
@@ -41,4 +44,89 @@
     default Validator<V> getDefaultValidator() {
         return Validator.alwaysPass();
     }
+
+    /**
+     * Enables the implementing components to notify changes in their validation
+     * status to the observers. This method is
```
mshabarov, line 50:
```suggestion
     * status to the observers.
```

flow-data/src/main/java/com/vaadin/flow/data/binder/HasValidator.java
```diff
@@ -41,4 +44,89 @@
...
+     * <p>
+     * <strong>Note:</strong> This method should be overridden by the
```
mshabarov, line 52:
```suggestion
     * <strong>Note:</strong> This method can be overridden by the
```

mshabarov (approved):

mshabarov (approved):

knoobie:
@mshabarov I would like to work on a solution do deactivate this feature; I was wondering if you guys would prefer it on Binder-level or Binding-level. 

This could later be enhanced from `private boolean enableFieldValidation = false / true` as default to `private boolean enableFieldValidation = FeatureFlags.get(BINDER_FIELD_VALIDATION);` if you guys wanna put it behind a feature flag and swap the default value later on.

PR #13942, opened 2022-06-10
fix: Scan Hilla packages by default

Artur-:
The removes the need to manually specify dev.hilla in the whitelist, something that should be considered an internal detail


platosha:
Added the target labels to get this fix to the `23.0` and `23.1` maintenance releases also.

platosha (approved):

Issue #9077, opened 2020-09-29
pushURL configuration is not working


tarekoraby:
### Description of the bug
setting pushURL configuration property is not working 

### Minimal reproducible example
Using a Vaadin 14.3.7 Spring Boot starter project, add a simple view that uses @Push
```
@Route("")
@Push
public class PushView extends VerticalLayout {
    public PushView() {

    }
}
```

And in `application.properties`, add the following line: `vaadin.pushURL = customUrl`
Then open `http://localhost:8080`

### Expected behavior
The websocket communication would work using the specified Url

### Actual behavior
An error is obtained in the client-side:
![image](https://user-images.githubusercontent.com/42799254/94551153-95995480-025d-11eb-804f-58e5627dfa5e.png)

### Versions:
    - Vaadin / Flow version: 14.3.7
    - Java version: 8
    - OS version: Windows
    - Browser version (if applicable): Chrome and Firefox

joheriks:
Workaround until fixed is to define `vaadin.pushURL=/vaadinServlet/customUrl` instead.

PR #13960, opened 2022-06-13
fix: Do not expose /vaadinServlet/* publicly

Artur-:
Fixes https://github.com/vaadin/spring/issues/789

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
@@ -80,8 +80,13 @@ public class SpringBootAutoConfiguration {
         Map<String, String> initParameters = new HashMap<>();
         boolean rootMapping = RootMappedCondition.isRootMapping(mapping);
         String[] urlMappings;
+        String pushRegistrationPath;
+
         if (rootMapping) {
             mapping = VaadinServletConfiguration.VAADIN_SERVLET_MAPPING;
+            pushRegistrationPath = "";
+        } else {
+            pushRegistrationPath = mapping.replace("/*", "");
         }
 
         urlMappings = new String[] { mapping, "/VAADIN/*" };
@@ -92,7 +97,7 @@ public class SpringBootAutoConfiguration {
          * and websockets will fail.
          */
         initParameters.put(ApplicationConfig.JSR356_MAPPING_PATH,
-                mapping.replace("/*", ""));
+                pushRegistrationPath);
 
         ServletRegistrationBean<SpringServlet> registration = new ServletRegistrationBean<>(
                 new SpringServlet(context, rootMapping), urlMappings);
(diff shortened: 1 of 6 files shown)


vaadin-spring/src/test/java/com/vaadin/flow/spring/service/SpringServletTest.java
```diff
@@ -52,59 +52,6 @@ public void readUniformNameProperty_propertyNameContainsDash_propertyNameIsConve
         Assert.assertEquals(PushMode.MANUAL, pushMode);
     }
 
-    @Test
-    public void pushURL_rootMapping_isPrefixedWithContextVaadinMapping()
```
mshabarov, line 56: Should we keep these tests instead, but exclude `/vaadinServlet` from the expected push URL?
Otherwise how we can prove that the https://github.com/vaadin/flow/issues/9077 won't be back and removing the bypass in `SpringServlet` is safe?

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
```diff
@@ -80,8 +80,13 @@ public ServletRegistrationBean<SpringServlet> servletRegistrationBean(
         Map<String, String> initParameters = new HashMap<>();
         boolean rootMapping = RootMappedCondition.isRootMapping(mapping);
         String[] urlMappings;
+        String pushRegistrationPath;
+
         if (rootMapping) {
             mapping = VaadinServletConfiguration.VAADIN_SERVLET_MAPPING;
+            pushRegistrationPath = "";
```
mshabarov, line 87: Common question, just for clarification: is this an "internal" improvement aiming to clean up our code and get rid of extra `vaadinServlet` segment? From a developers point of view this is implementation details, right?

vaadin-spring/src/test/java/com/vaadin/flow/spring/service/SpringServletTest.java
Artur-: As discussed outside this issue, I do not think there is a use case for #9077 and manually changing the push URL. It used to be a workaround for old weblogic servers that required a certain URL to be used but that is no longer relevant after the JSR-356 websocket standard became commonplace. We should instead remove support for setting a push URL: #14027

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
Artur-: This PR changes `/vaadinServlet` to be an internal implementation detail. Earlier it leaked out to the Spring Security configuration which needed to allow the push requests to `/vaadinServlet` and it might have leaked out also in some other places. 

Nothing should break by this though as you only need to remove rules allowing connections from the security configuration and if you extend the Vaadin security config they are removed automatically for you.

mshabarov (approved):

knoobie:
That was not a fix - it's an undocumented **breaking** change - see https://github.com/vaadin/flow/issues/14641

PR #14002, opened 2022-06-16
feat: Find a parent component with a given type

### LINKED ISSUE #13984: Add a method to find the active view (or parent view/layout)
### Describe your motivation

When building re-usable components they often need to interact with other parts of the UI, like the current view, navigation menu or some shared parts in some parent layout. Those are most often components in the current component tree, but in Vaadin provides no easy way to search them up. Instead we then inject them to other components using Spring/CDI or build application specific hacks to get references to them.

Providing a way for developers to seek views from the active UI might make it easier to create re-usable code.

A code example of a method (that could be placed to the UI class with minor changes):

```
    /**
     * Finds a reference to the current view or parent layout of given type.
     * 
     * @param viewType the class of the view or parent layout to search for
     * @return the view or empty optional if given view is not active
     * @param <T> the type of the view or parent layout to look for
     */
    public <T extends Component> Optional<T> findView(Class<T> viewType) {
        List<HasElement> activeRouterTargetsChain = UI.getCurrent().getInternals().getActiveRouterTargetsChain();
        for (HasElement element : activeRouterTargetsChain) {
            if (viewType.isAssignableFrom(element.getClass())) {
                return Optional.of((T) element);
            }
        }
        return Optional.empty();
   }

```

The implementation would be pretty much the same as in a private helper introduced in https://github.com/vaadin/flow/pull/13906

Then e.g. a single component could call a method in main layout like this:

```
ui.findView(MainLayout.class).ifPresent( main -> main.makeNewMessagesIndicatorBlink());
```

Write all code into single class and all problems are solved 🤓

Alternatively, Component could have a following helper:

```
/**
 * Finds a parent component of given type.
 *
 * @param componentType the class of the component to search for
 * @return the component
 * @param <T> the type of the component
 */
public <T extends Component> T findParent(Class<T> componentType) {
    Optional<Component> parent = getParent(); // see, parent here is Optional as well :-)
    while (parent.isPresent()) {
        Component component = parent.get();
        if (componentType.isAssignableFrom(component.getClass())) {
            return (T) component; // or Optional.of((T) element);
        } else {
            parent = component.getParent();
        }
    }
    return null; // or Optional.empty() or throw
}
```

### Additional context

I have been missing this kind of functionality in my own apps couple of time, but got the idea to bake this into actual proposal while working on improving the routing & navigation docs related to https://github.com/vaadin/flow/pull/13906 & https://github.com/vaadin/docs/issues/1482

mstahv:
## Description

Add a helper method to seek the parent component of given type from the component tree. 

Fixes #13984

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -684,16 +684,21 @@ public abstract class Component
     public void scrollIntoView() {
         getElement().scrollIntoView();
     }
-    
+
     /**
-     * Finds a component of given type from the parents of this component.
+     * Traverses the component tree up and returns the first component that
+     * matches the given type.
      *
-     * @param componentType the class of the component to search for
-     * @return the component as Optional or empty Optional if a parent of given type is not found
-     * @param <T> the type of the component to return
+     * @param componentType
+     *            the class of the component to search for
+     * @return the component as Optional or empty Optional if a parent of given
+     *         type is not found
+     * @param <T>
+     *            the type of the component to return
      */
     @SuppressWarnings("unchecked")
-	public <T extends Component> Optional<T> findParent(Class<T> componentType) {
+    public <T extends Component> Optional<T> findParent(
+            Class<T> componentType) {
         Optional<Component> parent = getParent();
         while (parent.isPresent()) {
             Component component = parent.get();
@@ -705,5 +710,5 @@ public abstract class Component
         }
         return Optional.empty();
     }
-    
+
 }
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -684,4 +684,26 @@ public abstract class Component
     public void scrollIntoView() {
         getElement().scrollIntoView();
     }
+    
+    /**
+     * Finds a component of given type from the parents of this component.
+     *
+     * @param componentType the class of the component to search for
+     * @return the component as Optional or empty Optional if a parent of given type is not found
+     * @param <T> the type of the component to return
+     */
+    @SuppressWarnings("unchecked")
+	public <T extends Component> Optional<T> findParent(Class<T> componentType) {
+        Optional<Component> parent = getParent();
+        while (parent.isPresent()) {
+            Component component = parent.get();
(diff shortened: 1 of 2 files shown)


mcollovati:
Personally I don't like that much the idea of traversing the tree and directly interact with an ancestor component, but I can get the points expressed in the related issue.
In addition we do not prevent this in any way and I'm pretty sure many developer may have already implemented such a helper on their own projects.

Some cosiderations on the proposed implementation:

- I would maybe move the implementation to `ComponentUtil` class since we already have some traversal helpers in that class.
- If instead we want to have it in `Component`, I think the API should not be public. I see it more as an internal helper for the component, rather than a way to traverse tree from outside.
- Do not restrict the component type to `Component` make sense to me; it is implicit that we will find a `Component`, but it may be useful to search it for example by a mixin or a custom interface.
- I would maybe call the method `findAncestor` rather than `findParent`

mstahv:
ComponentUtil feels like a set of internal helpers to me and it is not documented at all (other than JavaDocs), so I think far less developers would find it from there.

Renamed. Much better!

Also make protected. This is bit controversial. I guess most often one uses that from component, so protected *should* be fine, but somehow I have a feeling that someday I find my self opening that method directly as public in some Vaadin view 🤷‍♂️

mcollovati (commented):
I would still move the implementation to `ComponentUtil` and then have the method in `Component` that delegates to it.
But it is up to you.
Other than that, LGTM

mshabarov (changes requested):
> I would still move the implementation to ComponentUtil and then have the method in Component that delegates to it.

This sounds better for me as well.
By the way, I don't get the idea why you made it `protected`? The use case is to get the layout/menu or other high level component to modify it somehow. This method is invoked for some child component, so should the application actually extend the component to use this API, since is a protected, not public.
I would make it public, but transfer the implementation to the utils.

mshabarov:
Well, I would say it can be reused when moved to utils. But it can also be reused if placed to component directly.

mshabarov (dismissed):

mstahv:
👌 I'll write some basic unit test for this and convert to actual PR.

mstahv:
@mshabarov I'm ready with this!

mshabarov (approved):

PR #14005, opened 2022-06-17
chore: Simplify Gradle build so no dependency copying is needed

Artur-:


mshabarov:
Seems this approach doesn't work, let's close this PR.

Artur-:
Tested in #14084 that this fails without `mavenLocal()`

taefi (approved):

taefi:
I can remember at the time of adopting the Gradle plugin in flow, although the Gradle runnable was being executed via maven, it wasn't able to resolve the flow-plugin-base or any other dependency from the maven sandbox during a clean build (when no dependency pre-exists in the local Maven cache). The reason to copy the dependencies to `/target` was that Gradle could resolve them from a file system location. Currently, I can confirm that this is working, and possibly the use of the newer Gradle version resolved the problem.

PR #14012, opened 2022-06-21
feat: Make the Router lookup method available for others to use

Artur-:
When building a navigation menu component which is not based on RouterLink, there is the same need as RouterLink has: to create links to certain views. That requires a Router reference.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
@@ -39,7 +39,9 @@ import com.vaadin.flow.i18n.LocaleChangeEvent;
 import com.vaadin.flow.i18n.LocaleChangeObserver;
 import com.vaadin.flow.internal.ReflectionCache;
 import com.vaadin.flow.internal.StateNode;
+import com.vaadin.flow.internal.StateTree;
 import com.vaadin.flow.internal.nodefeature.VirtualChildrenList;
+import com.vaadin.flow.router.Router;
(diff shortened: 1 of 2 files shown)


mshabarov:
Could you please elaboration more why we need to expose the `Router`. I'm not against that, but just want to clarify why `UI::navigate` is not enough. Is there an issue/request for it?

Artur-:
When building a navigation menu component which is not based on `RouterLink`, there is the same need as `RouterLink` has: to create links to certain views. That requires a `Router` reference

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
```diff
@@ -602,4 +604,31 @@ public static <T> T getData(Component component, Class<T> type) {
                 (attributes, ignore) -> attributes.getAttribute(type), type);
     }
 
+    /**
+     * Gets the router instance for the given component.
+     *
+     * Falls back to the router for the currently active VaadinService if the
+     * component is not attached.
+     *
+     * @return a router instance
```
mshabarov, line 613:
```suggestion
     * @param component component for which the requested router instance serves navigation
     * @return a router instance
```

mshabarov (approved):

mshabarov:
> When building a navigation menu component which is not based on RouterLink, there is the same need as RouterLink has: to create links to certain views. That requires a Router reference

Thanks! I've updated the description.

mshabarov (approved):

Issue #13749, opened 2022-05-12
Commercial product leakage to package.json


knoobie:
### Description of the bug

Currently commercial projects like `vaadin-grid-pro` are added to the package.json and package-lock.json even tho only `vaadin-core` is used as dependency.

### Expected behavior

Not used dependencies / commercial products are added into the package.json or package-lock.json.

### Minimal reproducible example

- Create a V23 app in start.vaadin.com
- Replace `vaadin` dependency with `vaadin-core`
- Delete package.json & package-lock.json 
- mvn
- Look at the nearly created package.json containing commercial products 

### Versions

- Vaadin / Flow version: V23 + V23.1.0.beta1 tested
- Java version: 11

Artur-:
Yes, there should be separate version jsons in the two jars. Earlier it did not matter that vaadin-core listed all the versions as only what was found by scanning for `@NpmPackage` was installed but now I think Flow installs everything it finds in versions.json.

Technically it might be easier to name the files differently in the two jars so you can just fetch from the classpath using a unique name

haijian-ccc:
This is actually a critical bug. It's not uncommon that companies have license check in the build script, upon detecting commercial licenses, the build fails which makes Vaadin 23 not usable. And there is no workaround, as Vaadin always forces the commercial products, even the user only intends to use the free ones with vaadin-core.

ZheSun88:
I have created this [PR](https://github.com/vaadin/platform/pull/3060) to generate different json file for vaadin and vaadin-core

and the json will be included into different jars. 

```
Platform includes free components into vaadin-versions-core.json and includes it into vaadin-core JAR
Platform includes commertial components into vaadin-versions.json and includes it into vaadin JAR
```
i think flow will need to do corresponding changes to make it include the correct json into the generated package.json files.

PR #14014, opened 2022-06-21
fix: separate core and commercial components version files

taefi:
## Description

Previously, the vaadin_versions.json contained
both core and commercial components listed
which led to including them all in package.json
even if the user only used vaadin-core. Now
platform generates vaadin-core-version.json in
vaadin-core and vaadin-version.json to include
commercial components in vaadin artifact.
This commit will adopt having separate files
while generating package.json and pinning the
platform dependencies.

Fixes: #13749

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -230,8 +230,8 @@ public abstract class NodeUpdater implements FallibleCommand {
                     .parse(IOUtils.toString(content, StandardCharsets.UTF_8)));
             versionsJson = convert.getConvertedJson();
             versionsJson = new VersionsJsonFilter(getPackageJson(),
-                    DEPENDENCIES).getFilteredVersions(versionsJson,
-                            versionsOrigin);
+                    DEPENDENCIES)
+                    .getFilteredVersions(versionsJson, versionsOrigin);
         }
         return versionsJson;
     }
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -193,22 +193,50 @@ public abstract class NodeUpdater implements FallibleCommand {
      *             when versions file could not be read
      */
     JsonObject getPlatformPinnedDependencies() throws IOException {
-        URL resource = finder.getResource(Constants.VAADIN_VERSIONS_JSON);
-        if (resource == null) {
-            log().info("Couldn't find {} file to pin dependency versions."
-                    + " Transitive dependencies won't be pinned for pnpm.",
-                    Constants.VAADIN_VERSIONS_JSON);
+        URL coreVersionsResource = finder
+                .getResource(Constants.VAADIN_CORE_VERSIONS_JSON);
+        if (coreVersionsResource == null) {
+            log().info(
+                    "Couldn't find {} file to pin dependency versions for core components."
+                            + " Transitive dependencies won't be pinned for npm/pnpm.",
+                    Constants.VAADIN_CORE_VERSIONS_JSON);
             return Json.createObject();
         }
 
         JsonObject versionsJson;
-        try (InputStream content = resource.openStream()) {
+        try (InputStream content = coreVersionsResource.openStream()) {
             VersionsJsonConverter convert = new VersionsJsonConverter(Json
                     .parse(IOUtils.toString(content, StandardCharsets.UTF_8)));
             versionsJson = convert.getConvertedJson();
             versionsJson = new VersionsJsonFilter(getPackageJson(),
-                    DEPENDENCIES).getFilteredVersions(versionsJson);
+                    DEPENDENCIES).getFilteredVersions(versionsJson,
+                            Constants.VAADIN_CORE_VERSIONS_JSON);
         }
+
+        URL commVersionsResource = finder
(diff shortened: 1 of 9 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -193,22 +193,50 @@ protected File getPackageLockFile() {
...
+                            Constants.VAADIN_CORE_VERSIONS_JSON);
+        }
+
+        URL commVersionsResource = finder
+                .getResource(Constants.VAADIN_VERSIONS_JSON);
+        if (commVersionsResource == null) {
+            log().info(
+                    "Couldn't find {} file to pin dependency versions for commercial components."
+                            + " Transitive dependencies won't be pinned for npm/pnpm.",
+                    Constants.VAADIN_VERSIONS_JSON);
```
knoobie, line 222: Is there a way to determine if a commercial component is used? Otherwise this would be printed always and kinda looks like a bad advertisement every-time the app is build.

taefi: Thanks, @knoobie. I must admit This PR is not a draft but probably episode 1 of 2, since tests would be added in a separate PR. And of course, good hints like this are welcomed, and if for some reason they might not apply to this PR, they would be appreciated and applied to the next one.

taefi: For now, I changed it to `trace` instead of `info` to determine if it is actually needed there or not later.

caalador:
This sounds all wrong as the vaadin-versions should just be used to generate the versions.json used to lock versions in pnpm and nothing from it should be added to the package.json
I guess this is an issue with npm which for some reason adds everything also as dependency when it should only use it for the overrides package if the package exists in dependencies.

So I would see that the issue is that with an empty package.json (and no NpmPackage annotations) running the generation should leave you with only the default packages from NodeUpdater.

caalador:
Right. Npm became the default in [10.0.0.alpha1](https://github.com/vaadin/flow/releases/tag/10.0.0.alpha1) e.g. 23 so the issue should be there since v20 (?) as it was picked to 7.0, but was not noticed as pnpm was used instead.

Then this way is fine as I thought it was a side effect from the `overrides` change that was introduced to v23

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -193,22 +193,50 @@ protected File getPackageLockFile() {
...
+
+        JsonObject commVersionsJson;
+        try (InputStream content = commVersionsResource.openStream()) {
+            VersionsJsonConverter convert = new VersionsJsonConverter(Json
+                    .parse(IOUtils.toString(content, StandardCharsets.UTF_8)));
+            commVersionsJson = convert.getConvertedJson();
+            commVersionsJson = new VersionsJsonFilter(getPackageJson(),
+                    DEPENDENCIES).getFilteredVersions(commVersionsJson,
+                            Constants.VAADIN_VERSIONS_JSON);
+        }
```
caalador, line 234: As the same lines are duplicated I would perhaps add a `resourceToJson` method that is called instead. e.g. `private JsonObject resourceToJson(URL resource) { ... }`

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -193,22 +193,50 @@ protected File getPackageLockFile() {
...
             VersionsJsonConverter convert = new VersionsJsonConverter(Json
                     .parse(IOUtils.toString(content, StandardCharsets.UTF_8)));
             versionsJson = convert.getConvertedJson();
             versionsJson = new VersionsJsonFilter(getPackageJson(),
-                    DEPENDENCIES).getFilteredVersions(versionsJson);
+                    DEPENDENCIES).getFilteredVersions(versionsJson,
+                            Constants.VAADIN_CORE_VERSIONS_JSON);
+        }
+
+        URL commVersionsResource = finder
```
caalador, line 216:
```suggestion
        URL vaadinVersionsResource = finder
```
commVersionsResource sounds like some communication package or other. Would call it vaadinVersionsResource or just versionsResource (to be the same as core) to reflect better what it is.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
caalador: I feel this is probably extra log, also it should note that non core components are not pinned. 
Also if coreVersionsResource logged that no pinning is done then probably this shouldn't log at all.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
taefi: Done.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
taefi: Done.

caalador (approved):

caalador:
You need to do a `mvn formatter:format`

mshabarov (approved):

mshabarov (approved):

PR #14028, opened 2022-06-22
fix: Do not rely on web components calling the license checker

Artur-:


flow-client/src/main/frontend/vaadin-dev-tools.ts
```diff
@@ -1119,8 +1119,11 @@ export class VaadinDevTools extends LitElement {
     );
 
     if ((window as any).Vaadin) {
+      const existing = (window as any).Vaadin.devTools;
       (window as any).Vaadin.devTools = this;
+      Object.assign((window as any).Vaadin.devTools, existing);
```
vursen, line 1124:
```suggestion
      (window as any).Vaadin.devTools = Object.assign(this, (window as any).Vaadin.devTools);
```

And probably `window as any` is not required inside the `if` branch. I guess TypeScript should infer that?

Artur-: Fixed. TypeScript does not infer that

flow-client/src/main/frontend/License.ts
```diff
@@ -60,40 +60,45 @@ const showNoLicenseFallback = (element: Element, productAndMessage: ProductAndMe
...
+  if (cvdlName && version) {
+    const product: Product = { name: cvdlName, version };
+    productTagNames[cvdlName] = productTagNames[cvdlName] ?? [];
+    productTagNames[cvdlName].push(tagName);
+
+    const failedLicenseCheck = productMissingLicense[cvdlName];
+    if (failedLicenseCheck) {
+      // Has been checked and the check failed
+      const { connectedCallback } = constructor.prototype;
+      constructor.prototype.connectedCallback = function () {
```
vursen, line 83: The `connectedCallback` should be also overridden when the license has not been checked yet. Otherwise, there will be no license fallback shown in the case the license check failed but the chart is still not in the DOM.

Artur-: Fixed so it is always overwritten

vursen (approved):

PR #14034, opened 2022-06-22
Improving UX of Query parameters

### LINKED ISSUE #14013: Improve navigation with query parameters
### Describe your motivation

Vaadin 14.8.11

Currently, if I want to navigate to a view by java class name and I want to add a query parameter, I need to write the following code:

```java
RouteConfiguration configuration = RouteConfiguration
        .forRegistry(UI.getCurrent().getInternals().getRouter().getRegistry());
String url = configuration.getUrl(BookingFreightView.class, HasUrlParameterFormat.getParameters(bookingReference));
Map<String, List<String>> queryParameters = Collections.singletonMap("prev_view", Collections.singletonList(prevView.name()));
UI.getCurrent().navigate(url, new QueryParameters(queryParameters));
```

Please provide an additional overload for navigation: `UI.navigate(Class, Object, QueryParameters)` and also a static factory `QueryParameters.of(String, String)` so that the code can read:

```java
UI.getCUrrent().navigate(BookingFreightView.class, bookingReference, QueryParameters.of("prev_view", prevView.name()));
```

Also provide `QueryParameters.toString()` and `equals()/hashCode()`, similar to `RouteParameters`.

### Alternatives considered

`QueryParameter.simple()` still needs a map even though it's a bit better; `QueryParameter.fromString()` could work but will suffer from escaping issues.

For the massive `UI.navigate() overload` - perhaps a builder pattern could help? Not sure though, they all require a trailing `go()` call, which, when omitted, will simply run no navigation - that will be a frequent source of bugs. Not recommended:

```java
UI.getCurrent().navigateTo(it -> it.view(Foo.class, param).queryParam("foo", "bar"))
```
or
```java
new Navigation().route(Foo.class, param).queryParam("foo", "bar").go();
```

mstahv:
* made it easier to use both query and url parameters
 * shorthand to create URL parameters from single key-value pair

Fixes #14013

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -213,4 +226,28 @@ private static String decode(String parameter) {
                     "Unable to decode parameter: " + parameter, e);
         }
     }
+
+    @Override
+    public String toString() {
+        return getQueryString();
```
knoobie, line 232: nit: I'm personally not a big fan of toString methods without object identifier like class names, e.g. `QueryParameters(foo=bar)` instead of `foo=bar`.

The other part of the implementation looks good to me! (Not sure if you need to add unit-test for the new methods in UI tho)

mstahv: ✅

There is an integration test so I think that is even better covered.

mstahv: Oh, but it covers just the other new navigate method...

mstahv: Now both are covered.

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -975,6 +975,103 @@ public <T extends Component> Optional<T> navigate(Class<T> navigationTarget,
...
+     *             parameter is not annotated with @OptionalParameter
+     *             or @WildcardParameter.
+     * @throws NotFoundException
+     *             in case there is no route defined for the given
+     *             navigationTarget matching the parameters.
+     */
+    @SuppressWarnings("unchecked")
+    public <T extends Component> Optional<T> navigate(
+            Class<? extends T> navigationTarget,
+            QueryParameters queryParameters) {
```
mcollovati, line 1063: Are the parameter mentions in javadocs valid for this method?

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/UI.java
mstahv: As far as I know. Do you see something wrong here 🤔

I think they could be better, but should be improved all (overloaded methods) at once. IIRC, I copied the formulations from existing javadocs.

mcollovati: I'm asking because I do not see any parameter in this specific method, nor the need for the view to be a `HasUrlParameter`.
I supposed it was a copy/paste from the other method, but not clean up

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -975,6 +975,103 @@ public <T extends Component> Optional<T> navigate(Class<T> navigationTarget,
...
+        getInternals().getRouter().navigate(this,
+                new Location(url, queryParameters),
+                NavigationTrigger.UI_NAVIGATE);
+        return (Optional<C>) findCurrentNavigationTarget(navigationTarget);
+    }
+
+    /**
+     * Updates this UI to show the view corresponding to the given navigation
+     * target with the specified parameter. The parameter needs to be the same
+     * as defined in the route target HasUrlParameter.
```
mcollovati, line 1032:
```suggestion
     * Updates this UI to show the view corresponding to the given navigation
     * target and query parameters.
```

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -975,6 +975,103 @@ public <T extends Component> Optional<T> navigate(Class<T> navigationTarget,
...
+     * <p>
+     * Besides the navigation to the {@code location} this method also updates
+     * the browser location (and page history).
+     * <p>
+     * Note! A {@code null} parameter will be handled the same as
+     * navigate(navigationTarget) and will throw an exception if HasUrlParameter
+     * is not @OptionalParameter or @WildcardParameter.
+     * <p>
```
mcollovati, line 1040:
```suggestion
```

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -975,6 +975,103 @@ public <T extends Component> Optional<T> navigate(Class<T> navigationTarget,
...
+     *            navigation target to navigate to
+     * @param queryParameters
+     *            additional query parameters to pass to view
+     * @param <T>
+     *            navigation target type
+     * @return the view instance, if navigation actually happened
+     * @throws IllegalArgumentException
+     *             if a {@code null} parameter is given while navigationTarget's
+     *             parameter is not annotated with @OptionalParameter
+     *             or @WildcardParameter.
```
mcollovati, line 1055:
```suggestion
```

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/UI.java
mstahv: Oh, not type parameter but route parameter :-)

mcollovati (approved):

mcollovati:
Perhaps a temporary network error

PR #14058, opened 2022-06-24
fix: rename global js constant conflicting with require.js

### LINKED ISSUE #14057: PushWithRequireJSIT.pushWithRequireJS fails with Identifier 'define' has already been declared
`PushWithRequireJSIT,pushWithRequireJS` seems to fail frequently because of the following unexpected console message

```
SyntaxError: Identifier 'define' has already been declared
```

mcollovati:
## Description

In IndexHtmlRequestHandler the customElements.define function is
redefined to apply a custom feature after invoking the original
function. The reference to the the original function is stored in a
`define` constant and unfortunately this conflicts with require.js.
This change renames the constant in order to avoid the conflict.

In addition, it fixes an NPE in ApplicationRunnerServlet, when getting
DeploymentConfiguration in a debug window websocket request.

Fixes #14057

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist


mcollovati:
Perhaps a better solution may be wrapping the function substitution in an IIFE, so that the reference to the original function is not in global scope, but for now let see if the change fixes the build

taefi (approved):

Artur- (approved):

PR #14068, opened 2022-06-27
fix: Let DataCommunicator#flush keep track of which StateTree is supposed to execute its flushRequest

### LINKED ISSUE #14067: DataCommunicator#flush can get in a stuck state when the @PreserveOnRefresh annotation is present
### Description of the bug

In routes annotated with `@PreserveOnRefresh` , we sometimes see empty grids after a page reload, with the following obfuscated client-side stack trace popping up:

```
10:58:01.754 Exception is thrown during JavaScript execution. Stacktrace will be dumped separately. shares:6:224
10:58:01.754 The error has occurred in the JS code: '$0, $1, return $0.$connector.setSorterDirections($1)' shares:6:224
    error shares:6
    oC FlowClient.js:209
    lt FlowClient.js:997
    kt FlowClient.js:949
    it FlowClient.js:572
    Pq FlowClient.js:500
    mr FlowClient.js:1007
    VA FlowClient.js:919
    Rq FlowClient.js:1003
    gr FlowClient.js:1007
    lk FlowClient.js:415
    Jq FlowClient.js:1005
    Kq FlowClient.js:986
    $s FlowClient.js:1007
    zB FlowClient.js:1007
    onreadystatechange FlowClient.js:582
    tb FlowClient.js:434
    wb FlowClient.js:870
    vb FlowClient.js:620
10:58:02.536 Uncaught TypeError: $0.$connector is undefined
    anonymous vaadin-3-c3811f66fa2805358b94.cache.js line 11 > eval line 999 > Function:3
    lt FlowClient.js:997
    kt FlowClient.js:949
    it FlowClient.js:572
    Pq FlowClient.js:500
    mr FlowClient.js:1007
    VA FlowClient.js:919
    Rq FlowClient.js:1003
    gr FlowClient.js:1007
    lk FlowClient.js:415
    Jq FlowClient.js:1005
    Kq FlowClient.js:986
    $s FlowClient.js:1007
    zB FlowClient.js:1007
    onreadystatechange FlowClient.js:582
    tb FlowClient.js:434
    wb FlowClient.js:870
    vb FlowClient.js:620
vaadin-3-c3811f66fa2805358b94.cache.js line 11 > eval line 999 > Function:3:1
```

This indicates that the initialization of the client-side grid connector from `com.vaadin.flow.component.grid.Grid.GridArrayUpdaterImpl` has not run at the appropriate time.

Investigating this further, it looks like the root cause is an issue in `com.vaadin.flow.data.provider.DataCommunicator`
* in `requestFlush` a `flushRequest` callback is created
* the callback is scheduled on the `StateTree` beforeClientResponse task queue
* subsequent invocations of `requestFlush` will not schedule a new flush as long as the callback from the previous step is not cleared
* the callback itself is responsible for clearing out the `flushRequest` callback field. I.e. the pending flush is only cleared when the flush runs
* the `com.vaadin.flow.internal.StateTree#runExecutionsBeforeClientResponse`
  *  gathers all tasks from that beforeClientResponse task queue **and clears that queue** (see the `flushCallbacks` implementation)
  * executes those tasks, but **only those that `canExecute`**. The ones that cannot be executed are simply skipped, and never rescheduled

When there is a `@PreserveOnRefresh` annotation present (and with some unlucky timing on the page refresh), it is possible that the `canExecute` for the flush of the `DataCommunicator` returns `false`, because it was moved from one UI to another (i.e. from one `StateTree` to another)

As a result, the `flushRequest` callback in `DataCommunicator` does not run. That results in the `flushRequest` field never being cleared.

This means that there will never be another flush for that data communicator. The initialization of `com.vaadin.flow.component.grid.Grid.GridArrayUpdaterImpl` happens during these flushes, so that is why saw the client-side error mentioned in the beginning.

### Expected behavior

When refreshing, the grid should always properly execute its initialization logic and should not remain empty.

I haven't taken the time to make an MRE yet, since it takes quite a few moving parts.

In our application, we have
* A main layout annotated with `@PreserveOnRefresh` meaning all routes are preserved
* A grid with an in-memory data provider
* An asynchronous operation that updates the items in that data provider when it completes (*with* the appropriate `VaadinSession#access` calls)

### Versions

- Vaadin / Flow version: 23.0.7
- Java version: 11
- OS version: Linux
- IDE (if applicable): N/A

willemv:
## Description

As described in #14067 the `DataCommunicator#flush` can get stuck when its `StateNode` is moved between `StateTree` instances.

This change keeps track of the `StateTree` that will execute its `flushRequest` and forces a new `flushRequest` when the current `StateTree` and the one associated with the existing `flushRequest` don't match.

Fixes #14067

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
@@ -110,11 +110,8 @@ public class DataCommunicator<T> implements Serializable {
 
     private Registration dataProviderUpdateRegistration;
     private HashSet<T> updatedData = new HashSet<>();
-
-    private SerializableConsumer<ExecutionContext> flushRequest;
-    private StateTree.ExecutionRegistration flushRequestRegistration;
-    private SerializableConsumer<ExecutionContext> flushUpdatedDataRequest;
-    private StateTree.ExecutionRegistration flushUpdatedDataRequestRegistration;
+    private FlushRequest flushRequest;
+    private FlushRequest flushUpdatedDataRequest;
 
     private CallbackDataProvider.CountCallback<T, ?> countCallback;
     private int itemCountEstimate = -1;
@@ -1090,41 +1087,27 @@ public class DataCommunicator<T> implements Serializable {
     }
 
     private void requestFlush(boolean forced) {
-        if (flushRequestRegistration != null
-                && !flushRequestRegistration.canExecute()) {
-            flushRequest = null;
-        }
-        if ((flushRequest == null || forced) && fetchEnabled) {
-            flushRequestRegistration = null;
-            flushRequest = context -> {
+        if ((flushRequest == null || !flushRequest.canExecute(stateNode)
+                || forced) && fetchEnabled) {
+            flushRequest = FlushRequest.register(stateNode, context -> {
                 if (!context.isClientSideInitialized()) {
                     reset();
                     arrayUpdater.initialize();
                 }
                 flush();
                 flushRequest = null;
-            };
-            stateNode.runWhenAttached(ui -> flushRequestRegistration = ui
-                    .getInternals().getStateTree()
-                    .beforeClientResponse(stateNode, flushRequest));
+            });
         }
     }
 
     private void requestFlushUpdatedData() {
-        if (flushUpdatedDataRequestRegistration != null
-                && !flushUpdatedDataRequestRegistration.canExecute()) {
-            flushUpdatedDataRequest = null;
-        }
-        if (flushUpdatedDataRequest == null) {
-            flushUpdatedDataRequestRegistration = null;
-            flushUpdatedDataRequest = context -> {
-                flushUpdatedData();
-                flushUpdatedDataRequest = null;
-            };
-            stateNode.runWhenAttached(
-                    ui -> flushUpdatedDataRequestRegistration = ui
-                            .getInternals().getStateTree().beforeClientResponse(
-                                    stateNode, flushUpdatedDataRequest));
+        if (flushUpdatedDataRequest == null
+                || !flushUpdatedDataRequest.canExecute(stateNode)) {
+            flushUpdatedDataRequest = FlushRequest.register(stateNode,
+                    context -> {
+                        flushUpdatedData();
+                        flushUpdatedDataRequest = null;
+                    });
         }
     }
 
@@ -1483,6 +1466,27 @@ public class DataCommunicator<T> implements Serializable {
 
         public static Activation empty() {
             return new Activation(Collections.emptyList(), false);
+        }
+    }
+
+    private static class FlushRequest implements Serializable {
+
+        private NodeOwner owner;
+
+        static FlushRequest register(StateNode stateNode,
+                SerializableConsumer<ExecutionContext> action) {
+            FlushRequest request = new FlushRequest();
+            request.owner = stateNode.getOwner();
+            stateNode.runWhenAttached(ui -> {
+                request.owner = stateNode.getOwner();
+                ui.getInternals().getStateTree().beforeClientResponse(stateNode,
+                        action);
+            });
+            return request;
+        }
+
+        boolean canExecute(StateNode stateNode) {
+            return owner instanceof NullOwner || owner == stateNode.getOwner();
         }
     }
 
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
@@ -42,12 +42,7 @@ import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.function.SerializableComparator;
 import com.vaadin.flow.function.SerializableConsumer;
 import com.vaadin.flow.function.SerializableSupplier;
-import com.vaadin.flow.internal.ExecutionContext;
-import com.vaadin.flow.internal.JsonUtils;
-import com.vaadin.flow.internal.NodeOwner;
-import com.vaadin.flow.internal.Range;
-import com.vaadin.flow.internal.StateNode;
-import com.vaadin.flow.internal.StateTree;
+import com.vaadin.flow.internal.*;
 import com.vaadin.flow.shared.Registration;
 import com.vaadin.flow.shared.communication.PushMode;
 
@@ -117,7 +112,9 @@ public class DataCommunicator<T> implements Serializable {
     private HashSet<T> updatedData = new HashSet<>();
 
     private SerializableConsumer<ExecutionContext> flushRequest;
+    private StateTree.ExecutionRegistration flushRequestRegistration;
     private SerializableConsumer<ExecutionContext> flushUpdatedDataRequest;
+    private StateTree.ExecutionRegistration flushUpdatedDataRequestRegistration;
 
     private CallbackDataProvider.CountCallback<T, ?> countCallback;
     private int itemCountEstimate = -1;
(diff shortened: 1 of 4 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -42,12 +42,7 @@
 import com.vaadin.flow.function.SerializableComparator;
 import com.vaadin.flow.function.SerializableConsumer;
 import com.vaadin.flow.function.SerializableSupplier;
-import com.vaadin.flow.internal.ExecutionContext;
-import com.vaadin.flow.internal.JsonUtils;
-import com.vaadin.flow.internal.NodeOwner;
-import com.vaadin.flow.internal.Range;
-import com.vaadin.flow.internal.StateNode;
-import com.vaadin.flow.internal.StateTree;
+import com.vaadin.flow.internal.*;
```
czp13, line 45: I like better single imports for visibility, but in that case where no name clashing can happen, probably it is not a big thing, so it is just preference.

czp13 (commented):
Thanks for contribution.

I think I understand what we would like to achieve:
- storing flushUpdateDataRequestRegistrations and flushRequestRegistrations anytime we flush, and if they are not null and cannot execute for the next time, then we try the flush again on the stateNode.

But I am not sure about the potential (side) effects of this change, and it is doing what we would like to so, **would be very good to cover it with some tests, if possible...**

mcollovati:
Hi, thanks for the contribution!

I tried to setup some unit tests that should mimic the problem with `@PreserveOnRefresh`. You can add them to `com.vaadin.flow.data.provider.DataCommunicatorTest`

Unfortunately I'm still not able to setup a proper integration test

```
    // Simulates a flush request enqueued during a page reload with
    // @PreserveOnRefresh
    // see https://github.com/vaadin/flow/issues/14067
    @Test
    public void reattach_differentUI_requestFlushExecuted() {
        dataCommunicator.setDataProvider(createDataProvider(), null);
        dataCommunicator.setRequestedRange(0, 50);

        MockUI newUI = new MockUI();
        // simulates preserve on refresh
        // DataCommunicator has a flushRequest pending
        // that should be rescheduled on the new state tree
        newUI.getInternals().moveElementsFrom(ui);
        ui = newUI;
        fakeClientCommunication();

        Assert.assertEquals("Expected initial full reset.",
                Range.withLength(0, 50), lastSet);
    }

    // Simulates a flush request enqueued during a page reload with
    // @PreserveOnRefresh
    // see https://github.com/vaadin/flow/issues/14067
    @Test
    public void reattach_differentUI_requestFlushUpdatedDataExecuted() {
        List<Item> items = new ArrayList<>();
        for (int i = 0; i < 2; i++) {
            items.add(new Item(i));
        }
        // Abusing the fact that ListDataProvider doesn't copy the backing store
        ListDataProvider<Item> dataProvider = new ListDataProvider<>(items);
        dataCommunicator.setDataProvider(dataProvider, null);

        dataCommunicator.setRequestedRange(0, 50);
        fakeClientCommunication();

        Item originalItem = items.get(0);
        String key = dataCommunicator.getKeyMapper().key(originalItem);

        Assert.assertSame(originalItem,
                dataCommunicator.getKeyMapper().get(key));

        Item updatedItem = new Item(originalItem.id, "Updated");
        items.set(0, updatedItem);
        dataCommunicator.refresh(updatedItem);

        MockUI newUI = new MockUI();
        // simulates preserve on refresh
        // DataCommunicator has a flushRequest pending
        // that should be rescheduled on the new state tree
        newUI.getInternals().moveElementsFrom(ui);
        ui = newUI;
        fakeClientCommunication();

        Assert.assertSame(updatedItem,
                dataCommunicator.getKeyMapper().get(key));
    }

```

mcollovati (commented):
I would avoid changing the `ExecutionRegistration` interface, because it is  a public API (even if we provide implementation only in `StateNode`).

I would like to restrict changes only to the `DataCommunicator` class if possible.
A way could be to introduce an internal class that stores the `StateNode` owner at the moment of the `beforeClientResponse` registration and then provides the `canExecute` method that check the stored owner with the one from current NodeState

Something like this

```java
    private FlushRequest flushRequest;
    private FlushRequest flushUpdatedDataRequest;

    private static class FlushRequest implements Serializable {

        private NodeOwner owner;

        static FlushRequest register(StateNode stateNode,
                SerializableConsumer<ExecutionContext> action) {
            FlushRequest request = new FlushRequest();
            request.owner = stateNode.getOwner();
            stateNode.runWhenAttached(ui -> {
                request.owner = stateNode.getOwner();
                ui.getInternals().getStateTree().beforeClientResponse(stateNode,
                        action);
            });
            return request;
        }

        boolean canExecute(StateNode stateNode) {
            return owner instanceof NullOwner || owner == stateNode.getOwner();
        }
    }


    private void requestFlush(boolean forced) {
        if ((flushRequest == null || !flushRequest.canExecute(stateNode)
                || forced) && fetchEnabled) {
            // flushRequest = context -> {
            flushRequest = FlushRequest.register(stateNode, context -> {
                if (!context.isClientSideInitialized()) {
                    reset();
                    arrayUpdater.initialize();
                }
                flush();
                flushRequest = null;
            });
        }

    private void requestFlushUpdatedData() {
        if (flushUpdatedDataRequest == null
                || !flushUpdatedDataRequest.canExecute(stateNode)) {
            // flushUpdatedDataRequest = context -> {
            flushUpdatedDataRequest = FlushRequest.register(stateNode, context -> {
                        flushUpdatedData();
                        flushUpdatedDataRequest = null;
                    });
        }
    }

```

Let me know what do you think about this approach

mcollovati:
Hi @willemv do you have some time to check the proposed changes?
If you are busy we can create a new PR and mention you as co-author

willemv:
@mcollovati regarding the tests : one of the testcases does not fail before the fix.

Specifically, `reattach_differentUI_requestFlushUpdatedDataExecuted` succeeds even without the changes in `DataCommunicator`

The other test (`reattach_differentUI_requestFlushExecuted`) _does_ fail as expected before the fix, and passes after the fix, so that seems to be in order.

mcollovati:
@willemv seems like `flushUpdatedData()` is not called on attach and, furthermore, the callback registered by `requestFlush()` resets updated data.
So changes to `flushUpdatedData()` and the `reattach_differentUI_requestFlushUpdatedDataExecuted` test can be removed, or we can leave it there to prevent future regressions.

willemv:
> So I propose to remove the `reattach_differentUI_requestFlushUpdatedDataExecuted` test. We can leave the changes on the `flushUpdatedData()` method to make it consistent with `requestFlush()`

Agreed. I will adjust the PR

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mcollovati: @willemv May I ask you to use single class imports, as this is a Vaadin convention?

https://vaadin.com/docs/latest/contributing/editor-settings-intellij-idea/#imports

mcollovati (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
mcollovati:
```suggestion
import com.vaadin.flow.internal.ExecutionContext;
import com.vaadin.flow.internal.JsonUtils;
import com.vaadin.flow.internal.NodeOwner;
import com.vaadin.flow.internal.NullOwner;
import com.vaadin.flow.internal.Range;
import com.vaadin.flow.internal.StateNode;
import com.vaadin.flow.internal.StateTree;
```

willemv:
Should I rebase this branch to the latest `master` already, or are you still reviewing and would that interrupt your workflow?

Similarly, do you want me to squash these commits? The intermediary commits are not that interesting, I just kept them like this to facilitate your incremental reviews.

mcollovati:
> Should I rebase this branch to the latest `master` already, or are you still reviewing and would that interrupt your workflow?

There are no merge conflicts right now, so rebase is not needed

> Similarly, do you want me to squash these commits? The intermediary commits are not that interesting, I just kept them like this to facilitate your incremental reviews.

We will squash on merge, so no need for you to do it

mcollovati (approved):

taefi (approved):

PR #14100, opened 2022-06-29
fix: Set Vite hot reload port to match the server port

Artur-:
This is needed in Vite 3 to avoid an initial attempt to connect Vite's websocket channel to localhost:8080 which logs a browser warning


vursen:
As a result of this change, the app with Vite 2 is unable to connect to the HMR server for some reason:

image

Artur-:
Odd, it worked in my test. I need to recheck it then. Does it try to connect to ws://localhost:<viteport>/VAADIN/ or something different?

vursen:
Hmm, now that I updated `sw.ts` manually, everything works.

vursen (approved):

Issue #14092, opened 2022-06-28
Vaadin 14 npm mode pulls commercial products


haijian-ccc:
### Description of the bug

Probably the same problem as in https://github.com/vaadin/flow/issues/13749. In **Vaadin 14 with npm mode**, commercial projects like vaadin-grid-pro are added to the package.json and package-lock.json even though only vaadin-core is used as dependency. **pnpm** mode works fine.

### Expected behavior

No commercial products are added into the package.json or package-lock.json when **vaadin-core** is used.

### Minimal reproducible example

1. Download the Vaadin 14 starter app https://github.com/vaadin/skeleton-starter-flow/archive/v14.zip
2. Replace vaadin dependency with vaadin-core
3. Delete package.json & package-lock.json
4. mvn
5. See the newly created package.json containing commercial products

### Versions

- Vaadin / Flow version: 14.8.13
- Java version: 11

taefi:
Thanks @haijian-ccc for reporting this, it will be fixed by the CP for #14014 into 2.7 soon.

mcollovati:
Closing, as the fix has been released in [2.7.17](https://github.com/vaadin/flow/releases/tag/2.7.17)

PR #14123, opened 2022-07-01
 fix: separate core and commercial components version files (#14014) (CP: 9.0)

taefi:
This is a manual cherry-pick to collect
all changes that are done in #14014,
#14079, and #14083.

Related to #14092

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -212,8 +212,8 @@ public abstract class NodeUpdater implements FallibleCommand {
                     .parse(IOUtils.toString(content, StandardCharsets.UTF_8)));
             versionsJson = convert.getConvertedJson();
             versionsJson = new VersionsJsonFilter(getPackageJson(),
-                    DEPENDENCIES)
-                    .getFilteredVersions(versionsJson, versionsOrigin);
+                    DEPENDENCIES).getFilteredVersions(versionsJson,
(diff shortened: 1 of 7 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -175,25 +175,45 @@ private File getPackageJsonFile() {
...
-                    + " Transitive dependencies won't be pinned for pnpm.",
-                    Constants.VAADIN_VERSIONS_JSON);
+        URL coreVersionsResource = finder
+                .getResource(Constants.VAADIN_CORE_VERSIONS_JSON);
+        if (coreVersionsResource == null) {
+            log().info(
+                    "Couldn't find {} file to pin dependency versions for core components."
+                            + " Transitive dependencies won't be pinned for npm/pnpm.",
+                    Constants.VAADIN_CORE_VERSIONS_JSON);
+            return null;
```
mcollovati, line 185: should return an empty JSON?

taefi: Actually modified it to preserve the return null behavior as before in case it could not find the artifact in the classpath.

mcollovati: Ok, got it. So it is just the test `testGetPlatformPinnedDependencies_vaadinCoreVersionIsNotPresent_outputIsEmptyJson` that should be fixed

mcollovati (approved):

PR #14125, opened 2022-07-01
fix: Override connectedCallback in customElements.define

Artur-:
This is needed for the override to take effect, at least in Chrome


flow-client/src/main/frontend/License.ts
```diff
@@ -55,43 +55,39 @@ const showNoLicenseFallback = (element: Element, productAndMessage: ProductAndMe
...
 };
 
-const checkLicenseIfNeeded = (tagName: string) => {
-  const constructor: CustomElementConstructor & { cvdlName?: string } & { version?: string } = customElements.get(
-    tagName
-  )!;
-  const { cvdlName, version } = constructor;
+const checkLicenseIfNeeded = (element: Element) => {
+  const { cvdlName, version } = element.constructor as any;
   if (cvdlName && version) {
```
vursen, line 74: This check is no longer needed, I guess.

flow-client/src/main/frontend/License.ts
```diff
@@ -55,43 +55,39 @@ const showNoLicenseFallback = (element: Element, productAndMessage: ProductAndMe
...
   return `${product.name}_${product.version}`;
 };
 
-const checkLicenseIfNeeded = (tagName: string) => {
-  const constructor: CustomElementConstructor & { cvdlName?: string } & { version?: string } = customElements.get(
-    tagName
-  )!;
-  const { cvdlName, version } = constructor;
+const checkLicenseIfNeeded = (element: Element) => {
+  const { cvdlName, version } = element.constructor as any;
```
vursen, line 73:
```suggestion
  const { cvdlName, version } = element.constructor as CustomElementConstructor & { cvdlName: string, version: string };
```

flow-client/src/main/frontend/License.ts
Artur-: Removed

flow-client/src/main/frontend/License.ts
Artur-: Done

vursen (approved):

PR #14158, opened 2022-07-11
feat: allow disabling validation status change listener registration on binder

knoobie:
## Description

Add binder-level flag to allow disabling of 
validation status change listener registration 
for HasValidator fields.

Related to https://github.com/vaadin/flow/pull/13940#issuecomment-1174668400 and https://github.com/vaadin/flow/pull/13940#issuecomment-1174668400

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

- [X] I have performed a self-review and corrected misspellings.

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -1227,7 +1227,8 @@ public class Binder<BEAN> implements Serializable {
             onValueChange = getField().addValueChangeListener(
                     event -> handleFieldValueChange(event));
 
-            if (getField() instanceof HasValidator) {
+            if (getBinder().isFieldsValidationStatusChangeListenerEnabled()
+                    && getField() instanceof HasValidator) {
                 HasValidator<FIELDVALUE> hasValidatorField = (HasValidator<FIELDVALUE>) getField();
                 onValidationStatusChange = hasValidatorField
                         .addValidationStatusChangeListener(
@@ -1684,6 +1685,8 @@ public class Binder<BEAN> implements Serializable {
     private Set<Binding<BEAN, ?>> changedBindings = new LinkedHashSet<>();
 
     private boolean validatorsDisabled = false;
+
+    private boolean fieldsValidationStatusChangeListenerEnabled = true;
 
     /**
      * Creates a binder using a custom {@link PropertySet} implementation for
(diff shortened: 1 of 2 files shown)


knoobie:
@mcollovati Based on my discussion here: https://github.com/vaadin/flow-components/pull/3406#discussion_r908427652 and my comment here: https://github.com/vaadin/flow/pull/13940#issuecomment-1174668400

Idea behind this is: first enhancement to allow this change to be toggled by a FeatureFlag. I would love to gradually test this change with multiple applications before I do a big flip on all our forms and applications in a minor release (23.2) and going crazy with problems regarding "not meaningful error messages" and "user has no idea why the field is red".

mcollovati:
LGTM, but I would like for one additional review

I wonder if it would be good to be able to enable/disable the feature also for a single Binding

knoobie:
> LGTM, but I would like for one additional review

You okay with the name? It was really hard to come up with something.. and even that name feels clumsy.. 

> I wonder if it would be good to be able to enable/disable the feature also for a single Binding 

I had the feeling as well, but wanted to get something out first ;) I had multiple questions as well, but neglected them all in my first throw to make it simple. Should this not stop the listener registration and instead stop the execution of the validate() call.. allowing this to be changed on the fly, instead of only once per binder and so on..

mcollovati:
> You okay with the name? It was really hard to come up with something.. and even that name feels clumsy..

I agree that the name feels clumsy. I tried to find an alternative but was unsuccessfully. That's why I would like for another review.

mcollovati:
> Should this not stop the listener registration and instead stop the execution of the validate() call.. allowing this to be changed on the fly, instead of only once per binder and so on

Interesting point

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -1227,7 +1227,9 @@ public BindingImpl(BindingBuilderImpl<BEAN, FIELDVALUE, TARGET> builder,
             onValueChange = getField().addValueChangeListener(
                     event -> handleFieldValueChange(event));
 
-            if (getField() instanceof HasValidator) {
+            if (!builder.getBinder()
```
taefi, line 1230:
```suggestion
            if (!getBinder()
```

taefi (changes requested):
@knoobie the whole thing sounds like a good feature! Thanks for the contribution. Just suggested some changes.

mcollovati (approved):

knoobie:
Reminder: that the PR description probably has to be updated to something more meaningful, before a release is pushed with this encrypted message

taefi (approved):

PR #14165, opened 2022-07-12
test: fix minor issues related to macos m1 tests failing

MarcinVaadin:
## Description

Fixes issues that have been encountered during local build of master branch on MacOS M1 (arm) using OpenJDK 11.

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-client/src/test/java/com/vaadin/client/DomApiAbstractionUsageTest.java
+++ b/flow-client/src/test/java/com/vaadin/client/DomApiAbstractionUsageTest.java
@@ -47,7 +47,8 @@ public class DomApiAbstractionUsageTest {
 
     private static final Set<String> ignoredElementMethods = Stream
             .of("getTagName", "addEventListener", "getOwnerDocument",
-                    "hasAttribute", "getStyle", "getLocalName", "getAttribute")
+                    "hasAttribute", "getStyle", "getLocalName", "getAttribute",
+                    "equals", "getClass")
             .collect(Collectors.toSet());
 
     private final ClassVisitor classVisitor = new ClassVisitor(Opcodes.ASM5) {
--- a/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/ComplexDialogShortcutView.java
+++ b/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/ComplexDialogShortcutView.java
@@ -31,8 +31,6 @@ public class ComplexDialogShortcutView extends DialogShortcutView {
         // transport the dialog contents to overlay element
         dialog.getElement().executeJs(
                 overlayFetchJS + ".appendChild(this.firstElementChild);");
-        dialog.content.getStyle().set("position", "fixed").set("inset",
-                "calc(50% + " + (dialog.index * 100) + "px)");
     }
 
     @Override
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/scroll/MultipleAnchorsIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/scroll/MultipleAnchorsIT.java
@@ -28,7 +28,7 @@ public class MultipleAnchorsIT extends AbstractScrollIT {
 
     @Test
     public void numerousDifferentAnchorsShouldWorkAndHistoryShouldBePreserved() {
-        testBench().resizeViewPortTo(700, 800);
+        testBench().resizeViewPortTo(700, 600);
         open();
 
         final Long initialHistoryLength = getBrowserHistoryLength();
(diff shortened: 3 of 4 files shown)


flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/scroll/MultipleAnchorsIT.java
```diff
@@ -28,7 +28,7 @@ public class MultipleAnchorsIT extends AbstractScrollIT {
 
     @Test
     public void numerousDifferentAnchorsShouldWorkAndHistoryShouldBePreserved() {
-        testBench().resizeViewPortTo(700, 800);
+        testBench().resizeViewPortTo(700, 600);
```
mcollovati, line 31: Why it is needed to decrease viewport height?

flow-client/src/test/java/com/vaadin/client/DomApiAbstractionUsageTest.java
```diff
@@ -47,7 +47,8 @@ public class DomApiAbstractionUsageTest {
 
     private static final Set<String> ignoredElementMethods = Stream
             .of("getTagName", "addEventListener", "getOwnerDocument",
-                    "hasAttribute", "getStyle", "getLocalName", "getAttribute")
+                    "hasAttribute", "getStyle", "getLocalName", "getAttribute",
+                    "equals", "getClass")
```
mcollovati, line 51: Seems like "equals" and "getClass" exclusions are needed only when building with Java 18.
But it seems reasonable to me to add them to the list

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/ComplexDialogShortcutView.java
```diff
@@ -31,8 +31,6 @@ public void open(Dialog dialog) {
         // transport the dialog contents to overlay element
         dialog.getElement().executeJs(
                 overlayFetchJS + ".appendChild(this.firstElementChild);");
-        dialog.content.getStyle().set("position", "fixed").set("inset",
-                "calc(50% + " + (dialog.index * 100) + "px)");
```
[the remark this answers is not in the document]
MarcinVaadin: Fixed position of dialog out of viewport was causing ElementNotFound selenium exception.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/scroll/MultipleAnchorsIT.java
MarcinVaadin: In general setting fixed viewport size was causing issues in case if real viewport has different size - test fails. I've decreased it because of scaling issues on my environment. It has no impact on tests. Currently there is no option to resize viewport without validation using Testbench.

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/ComplexDialogShortcutView.java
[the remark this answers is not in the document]
mcollovati: There's a comment some line before 
```
// transport the dialog contents to overlay element
```
so I'm not that sure it is correct to remove the line. We may have false positive results

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/scroll/MultipleAnchorsIT.java
mcollovati: Unfortunately I couldn't find the reason for the original value.
However seems like that the important fix was adding the `resizeViewPortTo`.
So since test are green, I think the change is fine.
I also run the validation on teamcity and there was no failures

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/ComplexDialogShortcutView.java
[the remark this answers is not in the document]
MarcinVaadin: The fixed position with offset was introduced for each new dialog not to overlap previous dialogs:
image
But in case of "reusable-dialog" (which has id=999) the offset is 99900px! (because dialog.index * 100) which makes it move out of viewport and is causing:
`org.openqa.selenium.ElementNotInteractableException: element not interactable`

After removing fixed offset dialogs are rendered as standard block elements. DOM structure is intact and all of them are accessible by selenium:
Screenshot 2022-07-13 at 08 25 06

flow-client/src/test/java/com/vaadin/client/DomApiAbstractionUsageTest.java
MarcinVaadin: Indeed, not an issue when switched to OpenJDK 17

mcollovati (approved):

Issue #13115, opened 2022-02-22
[Vite] Re-build the service worker when it changes


vursen:
### Describe your motivation

When developing a custom service worker, it would be useful to have Vite re-building the service worker on source code change, preventing the user from having to restart the server.

### Describe the solution you'd like

Set up a watcher in `buildSWPlugin` that will cause re-building of the service worker once the user changes its source code.

vursen:
At this moment, Vite has an issue that causes generating an error when you try to re-build the service worker after the deps have been pre-bundled.

> [vite] Internal server error: Could not load /Users/vursen/dev/vaadin/flow/flow-tests/test-frontend/vite-pwa/node_modules/.pnpm/workbox-core@6.4.2/node_modules/workbox-core/index.mjs?v=5e1803bc (imported by frontend/sw.ts): ENOENT: no such file or directory, open '/Users/vursen/dev/vaadin/flow/flow-tests/test-frontend/vite-pwa/node_modules/.pnpm/workbox-core@6.4.2/node_modules/workbox-core/index.mjs?v=5e1803bc'

PR #14182, opened 2022-07-19
fix: build the service worker without writing it to FS

### LINKED ISSUE #13123: [Vite] A server restart occurs on every successful service worker build
### Describe your motivation

A follow-up to #12928, #12383

There is an issue in the current implementation which is that the Flow server restarts on every successful service worker build. This happens because the output `sw.js` gets written to the `/webapp` folder as a result of the build whereas Flow watches that folder and restarts the server when a file changes.

One of the consequences is that the "The frontend development build has not yet finished. Please wait..." loader, which you see when starting up the app, sometimes gets stuck when it coincides with a server restart. You'll have to manually reload the loader page when the server is available to continue the process.

It also slows down the initial app start-up.

A prototype: https://github.com/vaadin/flow/commit/4c8d5b5e7c2db84fdf0ce2a0a952ea578b06dae6

### LINKED ISSUE #14022: Application restarts immediately on startup when using Vite
### Description of the bug

When starting a 23.1 application using Vite, the startup sequence ends in
```
2022-06-22 10:55:56.268 DEBUG 4669 --- [   File Watcher] o.s.boot.devtools.restart.Restarter      : Application restart required due to the following changes: [/.../target/classes [/.../target/classes/META-INF/VAADIN/webapp/sw.js.map (ADD), /.../target/classes/META-INF/VAADIN/webapp/sw.js (ADD)]]
```

This causes two problems:
1. Longer startup time as the Spring Boot application is restarted before it is available
2. The browser shows a white page instead of the application because of 
image

Apparently the browser reloads when the Vite server is ready but this is exactly when the Spring Boot server is restarting. 

Because of this it all looks broken

### Expected behavior

The application should not restart on startup. The service worker should not be written to that directory in development mode

```
vaadin init hello
cd hello
echo "com.vaadin.experimental.viteForFrontendBuild=true" > src/main/resources/vaadin-featureflags.properties
mvn
```

### Versions

Vaadin: 23.1.1
Flow: 23.1.1
Java: Homebrew 17.0.1
OS: aarch64 Mac OS X 12.4
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36

mcollovati:
Prevents restart of application during startup.

Fixes #13123
Fixes #14022

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -52,8 +52,50 @@ const hasExportedWebComponents = existsSync(path.resolve(frontendFolder, 'web-co
 console.trace = () => {};
 console.debug = () => {};
 
-function buildSWPlugin(): PluginOption {
+function buildSWPlugin(opts): PluginOption {
   let config: ResolvedConfig;
+  const devMode = opts.devMode;
+
+  const swObj = {}
+
+  async function build(action: 'generate' | 'write') {
+    const includedPluginNames = [
+      'alias',
+      'vite:resolve',
+      'vite:esbuild',
+      'rollup-plugin-dynamic-import-variables',
+      'vite:esbuild-transpile',
+      'vite:terser',
+    ]
+    const plugins: rollup.Plugin[] = config.plugins.filter((p) => {
+      return includedPluginNames.includes(p.name)
+    });
+    plugins.push(
+      replace({
+        values: {
+          'process.env.NODE_ENV': JSON.stringify(config.mode),
+          ...config.define,
+        },
+        preventAssignment: true
+      })
+    );
+    const bundle = await rollup.rollup({
+      input: path.resolve(settings.clientServiceWorkerSource),
+      plugins
+    });
+
+    try {
+      return await bundle[action]({
+        file: path.resolve(frontendBundleFolder, 'sw.js'),
+        format: 'es',
+        exports: 'none',
+        sourcemap: config.command === 'serve' || config.build.sourcemap,
+        inlineDynamicImports: true,
+      });
+    } finally {
+      await bundle.close();
+    }
+  }
 
   return {
     name: 'vaadin:build-sw',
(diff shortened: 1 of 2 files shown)


mcollovati:
I took the prototype mentioned in #13123 and applied some minor changes.
To prevent application restart I had to avoid `bundle.write` on dev mode, replacing it with `bundle.generate` and storing the results into constant used later in `transform` to serve contents.
But I don't know if this is a good approach

mcollovati:
Additionally, if using `bundle.generate` in `transform` leads to the same error reported [here](https://github.com/vaadin/flow/issues/13115#issuecomment-1047692328)

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -62,49 +104,25 @@ function buildSWPlugin(): PluginOption {
...
-  };
+    },
+    async load(id) {
+      if (id.endsWith('sw.js')) {
+        return '';
+      }
+    },
+    async transform(_code, id) {
+      if (id.endsWith('sw.js')) {
+        return swObj;
```
vursen, line 122: Vite caches transform results so the bundle can be confidently generated here:

```suggestion
        const { output } = await build('generate');
        return output[0]; 
```

This way the bundle generation will be delayed until the first `sw.js` request and furthermore no need to maintain `swObj`.

vursen: Ah, didn't notice the comment that generating `sw.js` in `transform` actually causes an error.

mcollovati: Calling `build('generate')` in the `transform` hook give me the same error mentioned here
https://github.com/vaadin/flow/issues/13115#issuecomment-1047692328

That's the reason why i moved it in a different hook

mcollovati: > Ah, sorry, didn't notice the comment that generating `sw.js` in `transform` actually causes an error.

No problem, the comment was somehow "hidden" by bot comments :)

mshabarov (approved):

PR #14209, opened 2022-07-22
feat: Introduce getListeners(eventType) method on com.vaadin.flow.component.ComponentEventBus

### LINKED ISSUE #9766: Introduce getListeners(eventType) method on com.vaadin.flow.component.ComponentEventBus
With Vaadin 8.x it was possible to use com.vaadin.server.AbstractClientConnector#getListeners

It is not possible to access the listeners with Vaadin Flow without reflection.

Please re-introduce a getListeners(eventType) method on com.vaadin.flow.component.ComponentEventBus
in Vaadin Flow 14.x

In our case, we need it to record user interaction with a component.
For example a click via click listener.
For this, every component will add an "internal" click listener.
On layouts (like VerticalLayout) the interaction is only recorded if another click listener (!= internal) is registered.
So basically the internal click listener checks for "getListeners(ClickEvent.class) > 1" on click event.
And only if more than one click listener is added, the click is recorded.

What we already have tested:
Temporary remove our internal click listener and then call "hasListener".
But the problem is that this will change the listener order.
Our internal click listener must be always fired as the first.

Another possible solution would be to override all add(Click)Listener methods and track it by our self.
But that feels a bit poor

MarcinVaadin:
## Description

Introduced `com.vaadin.flow.component.ComponentEventBus.getListeners(Class<? extends ComponentEvent>)` together with corresponding shorthand method (on `Component`) and static helper (on `ComponentUtil`).

Fixes #9766 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/component/ComponentEventBus.java
```diff
@@ -176,6 +178,31 @@ public boolean hasListener(Class<? extends ComponentEvent> eventType) {
...
 
+    /**
+     * Returns all listeners that match or extend the given event type.
+     *
+     * @param eventType
+     *            the component event type
+     * @return A collection with all registered listeners. Empty if no listeners
+     *         are found.
+     */
+    protected Collection<?> getListeners(
```
mshabarov, line 189: I suggest to make this method public, since from the component event bus perspective it's a public API

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
```diff
@@ -386,6 +386,19 @@ public static <T extends ComponentEvent<?>> boolean hasEventListener(
         return component.hasListener(eventType);
     }
 
+    /**
+     * Returns all listeners that match or extend the given event type.
+     *
+     * @param eventType
+     *            the component event type
+     * @return A collection with all registered listeners. Empty if no listeners
```
mshabarov, line 394: -> "A collection with all registered listeners for a given event type"

flow-server/src/test/java/com/vaadin/flow/component/ComponentUtilTest.java
```diff
@@ -88,4 +90,21 @@ public void addListenerToComponent_hasListener_returnsTrue() {
...
+        Registration listener = ComponentUtil.addListener(component,
+                PollEvent.class, event -> {
+                });
+        Collection<?> listeners = ComponentUtil.getListeners(component,
+                PollEvent.class);
+        Assert.assertEquals(1, listeners.size());
+
+        listener.remove();
+        Assert.assertFalse(
+                ComponentUtil.hasEventListener(component, PollEvent.class));
```
mshabarov, line 108: Should this line check that `ComponentUtil.getListeners` returns empty collection, not `ComponentUtil.hasEventListener` (since you want to test exactly `getListeners`)?

mshabarov (changes requested):
Have a minor suggestions about javadoc and access modifier, otherwise LGTM.

mshabarov (approved):

Issue #13987, opened 2022-06-16
[vite] The offline path is requested regardless of the original path when online in dev mode


vursen:
### Description of the bug

The service worker always requests the offline path in dev mode even if the original request path is different. Although this makes sense when the app is offline, it is absolutely incorrect when the app is online.

The related code fragment:

https://github.com/vaadin/flow/blob/1bc5ff8606942cf611ed612f2f8a7bff2adc23a8/flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw.ts#L92-L101

### Expected behavior

When the app is online, the service worker should not modify HTTP requests.

When the app is offline, the service worker should redirect HTTP requests to the offline path.

### Minimal reproducible example

1. Clone https://github.com/vaadin/cookbook
2. Run `mvn`
3. Open http://localhost:8080/css-from-java

**Expected:**
```html
<meta name="description" content="This recipe shows how to inject small CSS class definitions into the current page from Java.">
```

**Actual:**
```html
<meta name="description" content="Find ready to use solutions and code snippets for common development problems and UX patterns. Includes Vaadin examples in Java and TypeScript.">
```

### Versions

- Vaadin / Flow version: 23
- Java version: 17
- OS version: Mac OS

mshabarov:
Thanks for the issue @vursen !
What is the impact on users? Does this issue make the app not working or working incorrectly?

vursen:
Not sure about the impact but it is presumably High: your app routes might not open correctly when you request them in the browser in dev mode and the service worker is enabled.

Artur-:
This issue makes it impossible to develop applications as it returns the wrong data in development mode in certain cases and you have no idea why

vursen:
### Background

There are two cases:

#### `OFFLINE_PATH` points to a custom offline path when it is specified by the user via the `PWA` annotation (e.g. `offline.html`).

In this case, Flow includes the offline path in the runtime manifest (`sw-runtime-resources-precache.js`) which causes it to get pre-cached in both dev and prod modes. Flow also maintains the revision hash, updating it when the content of the offline fallback file is changed. In other words, no workarounds are needed, and this case works exactly the same as in Webpack.

#### `OFFLINE_PATH` points to the root path (= `.`).

[...]

PR #14210, opened 2022-07-22
fix: Fix ServiceWorker path requests in dev mode (#14151) (CP: 23.0)

taefi:
* fix: Fix ServiceWorker path requests in dev mode

Fixes: #13987

Co-authored-by: Marco Collovati <marco@vaadin.com>

(cherry picked from commit 4e62c0400df594d6df2a3d11010d7d848d6259e7)

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/PwaRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/PwaRegistry.java
@@ -40,6 +40,7 @@ import java.util.stream.Collectors;
 
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


taefi:
While backporting the changes to 23.0, there was a big conflict between 23.0 and master in `sw.ts` and since it was hard to distinguish which parts should be backported and which parts should not, I ported the sw.ts from master to 23.0 in this PR. Please DO NOT MERGE without verifying that this was the right call.

mshabarov:
I propose to close this one, since V23.0 will be out of support soon (3 months after V23.1 is out -> September 2022), let's not put many efforts on that.

PR #14214, opened 2022-07-25
fix!: Make Component::findAncestor compatible with Classic components

mshabarov:
## Description

Renames Component::findAncestor to be compatible with the existing method in Classic Components https://github.com/vaadin/classic-components-internal/blob/3cc3033934a9bb24744bdbe6ce4b50b013be2883/vaadin-classic-components-flow/src/main/java/com/vaadin/classic/v8/ui/AbstractComponent.java#L158 .

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -697,7 +697,7 @@ public abstract class Component
      *            the type of the component to return
      */
     @SuppressWarnings("unchecked")
-    public <T> Optional<T> findAncestor(Class<T> componentType) {
+    public <T> Optional<T> findOptionalAncestor(Class<T> componentType) {
         Optional<Component> parent = getParent();
         while (parent.isPresent()) {
             Component component = parent.get();
--- a/flow-server/src/test/java/com/vaadin/flow/component/ComponentTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/ComponentTest.java
@@ -1664,7 +1664,7 @@ public class ComponentTest {
     }
 
     @Test
-    public void findAncestorTest() {
+    public void findOptionalAncestorTest() {
         UI ui = new UI();
         TestComponentContainer componentContainer = new TestComponentContainer();
         TestComponent component = new TestComponent();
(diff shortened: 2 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/component/Component.java
```diff
@@ -697,7 +697,7 @@ public void scrollIntoView() {
      *            the type of the component to return
      */
     @SuppressWarnings("unchecked")
-    public <T> Optional<T> findAncestor(Class<T> componentType) {
```
knoobie, line 700: What would you think of `findAncestorBy` or `findAncestorByClass` instead of `Optional`

mshabarov: These are a good names, but having `findAncestor` and `findAncestorBy` in Classic component plugin with the same method argument IMO looks confusing. At least with `Optional` in the middle it explains the difference.
`findAncestor` method should have been included into `Component` in the very beginning and plugin would then just use this API.

knoobie: Imo, I think the usage of Classic Components is far lower.. and it's just "a way to quickly migrate".. so in the long run.. classic components shouldn't be part of the app anyway (JM2C) I'm not familiar with the Classic Components source code after it got privatized ;) But I would guess that changing this method to return something without Optional<?> could be an option as well, or? But my comment can be neglected - something has to be done anyway :)

mcollovati: I wonder if we can apply the same "pattern" used for `UI.navigate()`

https://github.com/vaadin/flow/blob/1cacea3e2e3e8ffd66dabe433c8a22d2528c1884/flow-server/src/main/java/com/vaadin/flow/component/UI.java#L968-L976

If we add the bridge method for the non `Optional` version in addition to the new `findAncestor` signature, will it be compatible with classic components?
If it works we can deprecate the method without `Optional` and update classic components to use the new one in next releases, but preserving backward compatibility.

mshabarov: I absolutely agree about the usage thing. However having `By`/`ByClass`/`ByType` may seem for developer that the API contains other "find by" methods with different arguments and be also confusing. I don't see a big issue with "Optional" in the middle, but don't like it either.
For the long run, I would just remove this method from Classic components and don't touch the Flow. Since this method is not used internally by Classic components and its external usage most likely is low, I would consider remove it from there instead. Need to discuss this option.

mshabarov:
Closed in favour of https://github.com/vaadin/classic-components-internal/pull/164

PR #14221, opened 2022-07-26
feat: add configuration to `TaskGenerateHilla`

### LINKED ISSUE #14218: `TaskGenerateHilla` is not aware of the Maven context it's run in
### Describe your motivation

Implementations of the `TaskGenerateHilla` interface have no information about the Maven module they are run in. Basically, I'm not able to execute the code I need to.

Passing the `NodeTasks.Builder` instance might be a solution, but I still need to make sure that this is the case.

cromoteca:
Adds a method to `TaskGenerateHilla` that allows implementations to get some configuration, notably the path of the current Maven project.

Closes #14218

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -796,7 +796,8 @@ public class NodeTasks implements FallibleCommand {
             // use the new Hilla generator if available, otherwise the old
             // generator.
             if (hillaTask != null) {
-                commands.add(hillaTask);
+                hillaTask.configure(builder.getNpmFolder(),
+                        builder.getBuildDirectory());
             } else {
                 if (builder.endpointSourceFolder != null
                         && builder.endpointSourceFolder.exists()
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -798,6 +798,7 @@ public class NodeTasks implements FallibleCommand {
             if (hillaTask != null) {
                 hillaTask.configure(builder.getNpmFolder(),
                         builder.getBuildDirectory());
+                commands.add(hillaTask);
             } else {
                 if (builder.endpointSourceFolder != null
                         && builder.endpointSourceFolder.exists()
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateHilla.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateHilla.java
@@ -24,6 +24,16 @@ import java.io.File;
  * For internal use only. May be renamed or removed in a future release.
  */
 public interface TaskGenerateHilla extends FallibleCommand {
+    /**
+     * Configures the task by passing it some parameters.
+     *
+     * @param projectDirectory
+     *            the project root directory. In a Maven multi-module project,
+     *            this is the module root, not the main project one.
+     * @param buildDirectoryName
+     *            the name of the build directory (i.e.&nbsp;"build" or
+     *            "target").
+     */
     default void configure(File projectDirectory, String buildDirectoryName) {
     }
 }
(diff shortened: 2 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -796,7 +796,8 @@ private NodeTasks(Builder builder) {
             // use the new Hilla generator if available, otherwise the old
             // generator.
             if (hillaTask != null) {
-                commands.add(hillaTask);
+                hillaTask.configure(builder.getNpmFolder(),
+                        builder.getBuildDirectory());
```
mcollovati, line 800: Shouldn't 'hillaTask` be added to commands? I think task are not meant to be execute eagerly

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateHilla.java
```diff
@@ -15,12 +15,14 @@
...
+
 /**
  * Generate the Vaadin TS files for endpoints, and the Client API file. It uses
  * the new Maven/Gradle plugin based generator.
  * <p>
  * For internal use only. May be renamed or removed in a future release.
  */
 public interface TaskGenerateHilla extends FallibleCommand {
-
+    void configure(File projectDirectory, String buildDirectoryName);
```
mcollovati, line 27: Could it be a default method with empty body if the logic is in `execute` method?

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
cromoteca: Sure, it was not meant to replace the other line, but added to it. I don't know how I managed to create the commit this way as I absolutely need the command to be added.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateHilla.java
cromoteca: Yes, that can be done of course.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateHilla.java
```diff
@@ -15,12 +15,15 @@
...
+
 /**
  * Generate the Vaadin TS files for endpoints, and the Client API file. It uses
  * the new Maven/Gradle plugin based generator.
  * <p>
  * For internal use only. May be renamed or removed in a future release.
  */
 public interface TaskGenerateHilla extends FallibleCommand {
-
+    default void configure(File projectDirectory, String buildDirectoryName) {
```
mcollovati, line 27: Being a public API we should perhaps add Javadocs. For example we can make it explicit that this method will always be invoked once and before `execute`

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateHilla.java
```diff
@@ -24,6 +24,16 @@
...
  */
 public interface TaskGenerateHilla extends FallibleCommand {
+    /**
+     * Configures the task by passing it some parameters.
+     *
+     * @param projectDirectory
+     *            the project root directory. In a Maven multi-module project,
+     *            this is the module root, not the main project one.
+     * @param buildDirectoryName
+     *            the name of the build directory (i.e.&nbsp;"build" or
```
mcollovati, line 34: I think HTML space entity is not needed
```suggestion
     *            the name of the build directory (i.e. "build" or
```

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateHilla.java
cromoteca: Actually, I don't remember if @param description are shown in short format somewhere, so I put that there just in case.

mcollovati: Sorry, I don't understand why using a single `&nbsp;`. Can you please elaborate your previous answer?

cromoteca: I don't think it's needed as only method descriptions are truncated to the first sentence in Javadoc Summaries. I think that parameter description are never truncated.

mcollovati: Thanks for explanation

mcollovati (dismissed):

mcollovati (approved):

PR #14233, opened 2022-07-29
feat: swallow exceptions on unparseable NPM package versions

### LINKED ISSUE #14217: `build-frontend` goal fails on `npm` package versions which are not regular version numbers
### Describe your motivation

When developing frontend applications in JS/TS, like in Hilla, it is common to use relative paths as package versions to point to current code or local `npm pack`-generated packages.

The `build-frontend` goal is unable to parse those `package.json` files and fails the build.

The exception is thrown [here](https://github.com/vaadin/flow/blob/1cacea3e2e3e8ffd66dabe433c8a22d2528c1884/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java#L521).

[Extract of the build stacktrace](https://github.com/vaadin/flow/files/9187801/stacktrace.log)

Maybe the Node updater could simply ignore package versions it doesn't understand.

cromoteca:
When checking for new NPM package versions, let's skip all versions that we do not understand.

Closes #14217

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
@@ -263,14 +263,13 @@ public class NodeUpdaterTest {
 
         dependencies.put(formPackage, existingVersion);
 
-        try {
-            nodeUpdater.addDependency(packageJson, NodeUpdater.DEPENDENCIES,
-                    formPackage, newVersion);
-            Assert.fail("Expected NumberFormatException");
-        } catch (NumberFormatException e) {
-            Assert.assertTrue(
-                    e.getMessage().contains("is not a valid version"));
-        }
+        NumberFormatException expectedException = Assert
+                .assertThrows(NumberFormatException.class,
+                        () -> nodeUpdater.addDependency(packageJson,
+                                NodeUpdater.DEPENDENCIES, formPackage,
+                                newVersion));
+        Assert.assertTrue(expectedException.getMessage()
+                .contains("is not a valid version"));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


mcollovati:
We discussed internally and choosed to handle two try/catch statements.
* for `newVersion` we will throw the exception if the version is not parsable, since the value is originated in Flow code and should always be correct
* for `existingVersion` we will catch the exception, log the error and return `false`, so that the value from package.json is not replaced

I'll take care of this changes as soon of possible

cromoteca:
> We discussed internally and choose to handle two try/catch statements.

OK. In the meanwhile, I got the `newVersion` calculation out of the `try/catch` block, so the exception can propagate.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
```diff
@@ -226,6 +226,53 @@ public void shouldUpdateExistingLocalFormPackageToNpmPackage()
...
+        dependencies.put(formPackage, existingVersion);
+
+        try {
+            nodeUpdater.addDependency(packageJson, NodeUpdater.DEPENDENCIES,
+                    formPackage, newVersion);
+            Assert.fail("Expected NumberFormatException");
+        } catch (NumberFormatException e) {
+            Assert.assertTrue(
+                    e.getMessage().contains("is not a valid version"));
+        }
```
mcollovati, line 273: nit: try/catch can be replaced with `Assert.assertThrows`
```suggestion
        NumberFormatException expectedException = Assert
                .assertThrows(NumberFormatException.class,
                        () -> nodeUpdater.addDependency(packageJson,
                                NodeUpdater.DEPENDENCIES, formPackage,
                                newVersion));
        Assert.assertTrue(expectedException.getMessage()
                .contains("is not a valid version"));
```

mcollovati (approved):

PR #14234, opened 2022-07-29
fix: support import without url keyword in theme generator

mcollovati:
With Vite it may happend that '@import url(...)' in CSS are rewritten
as '@import ...', that is a valid CSS statement.
Currently the theme generator does not handle the case and this prevents
the resource to be added to the head section. A warning is also
logged on browser console, stating import rules not allowed when
creating stylesheets.
This change updates the importMatcher regex to handle @import without url
In addition it also handles media query declared in import statements.


mshabarov (approved):

mcollovati:
Need to do an additional check for Vite

mshabarov (approved):

PR #14174, opened 2022-07-15
test: workaround for internal selenium grid

mcollovati:
Test using dev tools are not working with internal selenium grid
because it is not up to date. Adding dependency to support chrome 100

PR #14252, opened 2022-08-02
Revert "test: workaround for internal selenium grid (#14174)"

alvarezguille:
Selenium grid now has chrome 103
This reverts commit 7f6155a19e39ee7fce06d4be4c6580253ae74912.


mcollovati (approved):
Tested on Bender and the build completes with success

PR #14254, opened 2022-08-03
chore: use a patch user permission checking

ZheSun88:
based on a [recent change](https://github.blog/changelog/2022-07-19-differentiating-triggering-actor-from-executing-actor/) the addon that we have used cannot return the user info for GHA rerun any more. 

i have patch-fix the addon, so let us use it for now and updated it later when they have fixed the issue.

**after merging this, please rebase the PRs from external contributors.**


ZheSun88:
guess this is not needed to cherry-pick back to the other branches, as normally the CP are done by vaadin-bot.

mcollovati (approved):

PR #14257, opened 2022-08-03
fix: let Vite process theme also on theme.js changes

mcollovati:
Flow theme plugin for Vite processes themes only on changes on files
in active theme directory.
This prevents regeneration with java live reload when changing value of
Theme annotation.
This change also triggers theme processing for updates on theme.js file.

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -334,6 +334,8 @@ function themePlugin(opts): PluginOption {
     handleHotUpdate(context) {
       const contextPath = path.resolve(context.file);
       const themePath = path.resolve(themeFolder);
+      // Track also updates on fronted/generated/theme.js to regenerate theme
+      // upon a java live reload when value of @Theme annotation changes.
       const isThemeJS = contextPath === path.resolve(themeOptions.frontendGeneratedFolder, "theme.js");
       if (isThemeJS || contextPath.startsWith(themePath)) {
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -334,12 +334,13 @@ function themePlugin(opts): PluginOption {
     handleHotUpdate(context) {
       const contextPath = path.resolve(context.file);
       const themePath = path.resolve(themeFolder);
-      if (contextPath.startsWith(themePath)) {
+      const isThemeJS = contextPath === path.resolve(themeOptions.frontendGeneratedFolder, "theme.js");
```
mshabarov, line 337: Do I understand it right that when the theme name in `@Theme` annotation is changed, this file is regenerated because of java live reload, but changing of theme.js by developer doesn't make sense since it contains only theme import/export?

mcollovati: Yes, `theme.js` in generated folder is not supposed to be changed by the developer

mshabarov: Ok, thanks. Could you then add a comment describing that the theme.js monitoring is needed for live reload upon theme name change in `@Theme`

mshabarov: You can add "[skip ci]" to the commit name to skip the validation

mshabarov (approved):

PR #14261, opened 2022-08-03
test: use npm instead of pnpm in ITs

taefi:
## Description

As the default package manager is npm
ITs should be also validated by npm 
instead of pnpm. 
Using pnpm can still be considered as 
an option to improve the build and test 
time on CI.


mshabarov (approved):

Artur-:
What is the performance impact of this?

taefi:
> What is the performance impact of this?

This was more of a test to see whether the repo is npm friendly ;)
I expect to see some timeouts and failures in the tests, and the performance impact can be measured better after a successful build.

mshabarov:
@taefi I propose to close this PR, some tests are failing and probably there is no big profit for us to change the build tool in tests.

taefi:
I Agree, the original intention was to see the impact of changing it to npm, but we can actually try this in practice later.

PR #14276, opened 2022-08-05
fix: Ensure internal parent set to null when node is detached

TatuLund:


mshabarov (changes requested):
@TatuLund could you please provide more context on this patch, why is it needed?

TatuLund:
This is a pure side observation found when investigating other thing, where I first thought I would need this, but eventually not related. This is however something that may pop up later. At this point of time we do not have known logic that would suffer from this bug. But it is somewhat a bug nevertheless.

mshabarov:
Thanks. Yes, it makes sense to set parent to `null` when detached, since the javadoc also says that parent is supposed to be null when the node is detached:
```
the new parent of this node; or <code>null</code> if this node is not attached to another node
```
I also propose to slightly change the title:

`fix: Ensure internal parent set to null when node is detached`.

mshabarov (approved):

PR #14285, opened 2022-08-08
feat: add a feature flag for new field validation behavior

vursen:
## Description

This PR adds a feature flag for the new field validation behavior. The flag will be used by Flow field components. For more information about the new behavior, please refer to https://github.com/vaadin/platform/issues/3066.

Part of https://github.com/vaadin/platform/issues/3066

## Type of change

- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
+++ b/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
@@ -76,7 +76,8 @@ public class FeatureFlags implements Serializable {
             "collaborationEngineBackend",
             "https://github.com/vaadin/platform/issues/1988", true, null);
     public static final Feature ENFORCE_FIELD_VALIDATION = new Feature(
-            "Enforce client / constraint / binder validation", "enforceFieldValidation",
+            "Enforce client / constraint / binder validation",
+            "enforceFieldValidation",
             "https://github.com/vaadin/platform/issues/3066", false, null);
     private List<Feature> features = new ArrayList<>();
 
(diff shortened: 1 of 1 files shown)


mshabarov (changes requested):
Conflicts with master should be resolved.

By the way, where is this flag supposed to be checked? Does the Flow's `Binder` need to check it and behave accordingly?

vursen:
> By the way, where is this flag supposed to be checked? Does the Flow's Binder need to check it and behave accordingly?

The flag is planned to be used only in Flow field components.

mshabarov (dismissed):

vursen:
@yuriy-fix suggested an alternative naming for the flag: `synchronizedValidation`

knoobie:
Another name could be `(enforce) Field / Client Validation` because it enables the field validation to run before the binder. I don't think a lot of people understand the concept of chaining and validation ordering.

vursen:
@knoobie I agree that the current naming is not descriptive and that we need to find a better alternative. 

It is noteworthy the field constraint validation has already been somewhat integrated into the binder validation by implementing `getDefaultValidator` which is already available without feature flags, so the field validation is already enforced in a way. This is probably not a problem and your suggested naming would still work. I wonder how we should name the feature flag itself then: `enforceClientFieldValidation`? `enforceClientValidation`?

knoobie:
> and this is already available without feature flags, so the field validation is already enforced in a way.

Oh! I was expecting the whole epic (including the field changes) would be behind the feature flag and not just the client side. (Don't mind, was just unexpected) 

> enforceClientFieldValidation

I would think this describes it the best

vursen:
After an internal discussion, we came to the agreement that 

- We will hide everything related to the new validation behavior behind a feature flag, **including the recent `getDefaultValidator` changes** 
- We will name the feature flag `enforceFieldValidation`.

yuriy-fix (approved):

mshabarov (approved):

PR #14287, opened 2022-08-08
feat: Support Object type keys in Component#getTranslation

### LINKED ISSUE #14286: Support Object type keys in Component#getTranslation
In some systems the key to localization bundle is not necessarily a String, but could be e.g. Enum. Hence it would be convenient to have overloading of the getTranslation method so that Object type keys could be used.

TatuLund:
Fixes #14286

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -622,6 +622,26 @@ public abstract class Component
     }
 
     /**
+     * Get the translation for the component locale.
+     * <p>
+     * The method never returns a null. If there is no {@link I18NProvider}
+     * available or no translation for the {@code key} it returns an exception
+     * string e.g. '!{key}!'.
+     *
+     * @see #getLocale()
+     *
+     * @param key
+     *            translation key
+     * @param params
+     *            parameters used in translation string
+     * @return translation for key if found (implementation should not return
+     *         null)
+     */
+    public String getTranslation(Object key, Object... params) {
+        return getTranslation(key, getLocale(), params);
+    }
+
+    /**
      * Get the translation for key with given locale.
      * <p>
      * The method never returns a null. If there is no {@link I18NProvider}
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/component/Component.java
```diff
@@ -646,6 +666,31 @@ public String getTranslation(String key, Locale locale, Object... params) {
...
+     * @param locale
+     *            locale to use
+     * @param params
+     *            parameters used in translation string
+     * @return translation for key if found
+     * @deprecated Use {@link #getTranslation(Locale, String, Object...)}
+     *             instead
+     */
+    @Deprecated
+    public String getTranslation(Object key, Locale locale, Object... params) {
```
mshabarov, line 687: I propose to not include this method into master - looks weird when you add a new method and tell that it is deprecated :)
If it is still needed for V14, let's deprecate it there only.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/Component.java
TatuLund: It will not be deprecated in V14.

TatuLund: See https://github.com/vaadin/flow/pull/14312

mshabarov: This is not clear for me still...You add a new method and immediately recommend to not use it.
Can we remove it and use another one in the overloaded method (needs to swap first two parameters):

```java
    public String getTranslation(Object key, Object... params) {
        return getTranslation(getLocale(), key, params);
    }
```

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/Component.java
mshabarov: Discussed offline that we keep the deprecated method for the sake of consistency between V14 and V23, because the same API has been added into V14.

mshabarov (approved):

PR #14289, opened 2022-08-08
feat: add feature flag for Hilla multi-module engine

### LINKED ISSUE #14266: Hilla: add a feature flag for multi-module engine

Currently, the new multi-module engine in Hilla is an add-on that requires adding both a project dependency with the runner task and the engine maven plugin to `pom.xml` in a user application.

A simpler way to enable multi-module engine in new and existing projects could be achieved using a feature flag. This allows ship the multi-module engine dependencies with existing Hilla platform artifacts (`dev.hilla:hilla`, `hilla-maven-plugin`), removing the need to add new dependencies to `pom.xml`, and enabling easier migration, as well as better discoverability for the feature.

- `FeatureFlags` defines the `FeatureFlags.HILLA_ENGINE` feature.
- `vaadin-featureflags.properties` supports enabling `FeatureFlags.HILLA_ENGINE` with `com.vaadin.experimental.hillaEngine=true`

### Expected behavior

`NodeTasks` only uses `TaskGenerateHilla` found by lookup when the feature flag is enabled, otherwise uses the old `EndpointGeneratorTaskFactory` / `TaskGenerateOpenAPI` / `TaskGenerateEndpoint`.

### Actual behavior

`NodeTasks` uses `TaskGenerateHilla` implementation when found by lookup.

### Versions:
    - Vaadin / Flow version: 23.2
    - Java version:
    - OS version:
    - IDE (if applicable):

platosha:
Fixes #14266

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksWebpackTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksWebpackTest.java
@@ -56,10 +56,7 @@ import static com.vaadin.flow.server.frontend.FrontendUtils.WEBPACK_GENERATED;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
-import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.times;
 
 public class NodeTasksWebpackTest {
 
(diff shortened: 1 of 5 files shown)


flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTasksWebpackTest.java
```diff
@@ -302,70 +299,6 @@ public void should_GenerateTsConfigAndTsDefinitions_When_Vaadin14BootstrapMode()
         Assert.assertTrue(new File(userDir, "types.d.ts").exists());
     }
 
-    @Test
-    public void should_useHillaGeneartor_whenAvailable()
```
mshabarov, line 306: Removing these tests from NodeTasksWebpackTest would effectively mean that Hilla generator would be tested only agains Vite. Is this enough for you? I would suggest make NodeTasksHillaWebpackTest and move them there. However if it is mostly about covering the feature flag triggering, then let's keep it as is.

platosha: It does not matter if Webpack or Vite is used here, I think. This is about having the Hilla tasks invoked, and yes, covering the feature flag also. Those are separate from Webpack / Vite tasks, so it’s arguably better to unit-test them separately.

mshabarov (approved):

PR #14296, opened 2022-08-08
test: reduce flakyness of ccdm flow navigation tests

mcollovati:
Decreases heartbeat interval to make test more stable
Uses system property to detect development mode, since getting 'Vaadin.developmentMode' from browser returns wrong value in Bender validation job


mcollovati:
Validation job passes also on Bender

mshabarov (approved):

PR #14299, opened 2022-08-09
test: Explicitly set dev mode for Vite basic tests

mshabarov:
`BasicsIT::debugWindowShown` is flaky: dev tool popup is not shown at all in CI, probably because the app is running in prod mode erroneously.


mcollovati (dismissed):
It may not be the main cause of the test failure, but it is good to have

mcollovati (approved):

PR #14312, opened 2022-08-10
feat: Support Object type keys in Component#getTranslation (#14287) (CP 2.8)

### LINKED ISSUE #14286: Support Object type keys in Component#getTranslation

TatuLund:
Fixes #14286 for Vaadin 14

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -588,6 +588,26 @@ public abstract class Component
     }
 
     /**
+     * Get the translation for the component locale.
+     * <p>
+     * The method never returns a null. If there is no {@link I18NProvider}
+     * available or no translation for the {@code key} it returns an exception
+     * string e.g. '!{key}!'.
+     *
+     * @see #getLocale()
+     *
+     * @param key
+     *            translation key
+     * @param params
+     *            parameters used in translation string
+     * @return translation for key if found (implementation should not return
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/component/Component.java
```diff
@@ -609,6 +629,30 @@ public String getTranslation(String key, Locale locale, Object... params) {
...
+     * string e.g. '!{key}!'.
+     *
+     * @param key
+     *            translation key
+     * @param locale
+     *            locale to use
+     * @param params
+     *            parameters used in translation string
+     * @return translation for key if found
+     * @deprecated Use {@link #getTranslation(Locale, String, Object...)}
```
mcollovati, line 646:
```suggestion
     * @deprecated Use {@link #getTranslation(String, Locale, Object...)}
```

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/Component.java
TatuLund: Good catch, the @depracatd should not be there at all, it was copy paste left over from V23.

mshabarov (approved):

mcollovati (approved):

PR #14341, opened 2022-08-12
fix: Strip servlet path from path when serving static files (#14239)

### LINKED ISSUE #14239: Vaadin23: index.html rewritten causing vaadin-bundle-xxx.cache.js URL to be 404 not found
### Description of the bug

The situation is described in [this stackoverflow.com question](https://stackoverflow.com/questions/73170520):

I'm working on a Vaadin application running under tomcat in a WAR file.

I'm trying to simply upgrade it from Vaadin 22 to Vaadin 23.

It was loading reliably with Vaadin 22 but now it fails to load with Vaadin 23.

The WAR file is named `pcom.war` and within it `web.xml` contains:
```xml
<servlet-mapping>
    <servlet-name>PCOM</servlet-name>
    <url-pattern>/pcom/*</url-pattern>
</servlet-mapping>
```

Therefore, the URL I trying to load the web application from, and from which Vaadin is failing to load successfully, is: `http://localhost:8080/pcom/pcom/`.

The `index.html` generated by Vaadin in `META-INF/VAADIN/webapp/index.html` in `pcom.war` contains (line breaks added for clarity):
```
<!doctype html>
<html lang="en">
<head>
<meta charset="UTF-8"/>
<meta name="viewport" content="width=device-width,initial-scale=1"/>
<style>body, #outlet {
      height: 100vh;
      width: 100%;
      margin: 0;
    }</style>
<script defer="defer" src="VAADIN/build/vaadin-bundle-67fde5fb08b0f134e867.cache.js"></script>
</head>
<body>
<div id="outlet"></div>
</body>
</html>
```

Note the `src="VAADIN/build/vaadin-bundle-...` relative URL. This is correct, relative to `index.html` as it sits within the `META-INF` contents of `pcom.war`:

```
        0  07-29-2022 14:33   META-INF/
       79  07-29-2022 14:33   META-INF/MANIFEST.MF
        0  07-29-2022 14:33   META-INF/VAADIN/
        0  07-29-2022 14:33   META-INF/VAADIN/config/
     1844  07-29-2022 14:33   META-INF/VAADIN/config/flow-build-info.json
     1658  07-29-2022 14:33   META-INF/VAADIN/config/stats.json
        0  07-29-2022 14:33   META-INF/VAADIN/webapp/
        0  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/
        0  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/
  1051904  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-1-cfc74d1f13c51748c7ba.cache.js
   259593  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-1-cfc74d1f13c51748c7ba.cache.js.gz
   607425  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-2-737c0b4498b03051df61.cache.js
   158040  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-2-737c0b4498b03051df61.cache.js.gz
    67025  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-3-460383d2eeb67dbda1eb.cache.js
    23477  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-3-460383d2eeb67dbda1eb.cache.js.gz
    27991  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-4-9ba06ea1fd5a6da385f4.cache.js
     8040  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-4-9ba06ea1fd5a6da385f4.cache.js.gz
     3518  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-5-a21e874ef30c99e3575f.cache.js
     1480  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-5-a21e874ef30c99e3575f.cache.js.gz
   138893  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-6-5710b84bf57453285225.cache.js
    46654  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-6-5710b84bf57453285225.cache.js.gz
    86955  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-bundle-67fde5fb08b0f134e867.cache.js
    26514  07-29-2022 14:33   META-INF/VAADIN/webapp/VAADIN/build/vaadin-bundle-67fde5fb08b0f134e867.cache.js.gz
      365  07-29-2022 14:33   META-INF/VAADIN/webapp/index.html
      273  07-29-2022 14:33   META-INF/VAADIN/webapp/index.html.gz
```

But when I load the application, FireFox inspector shows this error:

[![enter image description here][1]][1]

Note the reported error loading `http://localhost:8080/pcom/VAADIN/build/vaadin-bundle-67fde5fb08b0f134e867.cache.js`.

That URL is incorrect and generates a 404 error - I can verify this using `curl(1)`.

However I can use `curl(1)` to successfully load the correct URL, which is `http://localhost:8080/pcom/pcom/VAADIN/build/vaadin-bundle-67fde5fb08b0f134e867.cache.js`.

**Why does Vaadin 23 rewrite the URL in the `index.html` file so as to make it unloadable??**

  [1]: https://i.stack.imgur.com/Hux0j.png

### Expected behavior

Application should load.

None available yet. See description above for the basic setup.

### Versions

- Vaadin / Flow version: 23.1.3
- Java version: 11.0.16
- OS version: Mac OS X 12.5
- Browser version (if applicable): Firefox 103.0
- Application Server (if applicable): Tomcat 9.0.65
- IDE (if applicable): vim

> you can copy this from the debug window

(How do I show this?)

archiecobbs:
## Description

When the Vaadin servlet tries to serve static files from classpath resources, it needs to strip off the servlet path, if any, associated with the request before trying to find them, because the servlet is located relative to the servlet path in the HTTP URI request space. But the `StaticFileServer` is not doing that.

When the Vaadin servlet is mapped to `/*` then this bug has no impact, because the servlet path is the empty string.

But in any other scenario, e.g., servlet mapped to `/myapp`, then classpath resources will not be found because Vaadin will try to find classpath resource `/foo/bar.gif` at classpath location `/myapp/foo/bar.gif`, which doesn't exist.

For example, this bug causes any add-on that includes JS or CSS files (which is most of them) to fail.

Fixes #14239

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

NOTE: A couple of existing unit tests were wrong (in my opinion) and had to be corrected. They now serve as unit tests for this fix.

Obviously, I'm unsure about "correcting" an existing unit test. But basically they seemed obviously wrong to me because they were trying to access a static resource from the servlet, but without using the servlet's servlet path, which makes no sense.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
@@ -259,6 +259,11 @@ public class StaticFileServer implements StaticFileHandler {
                             + filenameWithPath.replaceFirst("^/", ""));
         }
 
+        // Strip off servlet path (if any) from file path so it corresponds to our resources as we know them
+        final String servletPath = request.getServletPath();
+        if (servletPath != null && !servletPath.isEmpty() && !isSpecialRequestFilename(request))
+            filenameWithPath = filenameWithPath.substring(servletPath.length());
+
         if (resourceUrl == null) {
             resourceUrl = getStaticResource(filenameWithPath);
         }
@@ -434,6 +439,28 @@ public class StaticFileServer implements StaticFileHandler {
      * @return the requested file name, starting with a {@literal /}
      */
     String getRequestFilename(HttpServletRequest request) {
+
+        // In special cases, do not prepend the servlet path
+        if (isSpecialRequestFilename(request))
+            return request.getPathInfo();
+
+        // Prepend servlet path to path info (if any)
+        String requestFilename = request.getServletPath();
+        if (request.getPathInfo() != null)
+            requestFilename += request.getPathInfo();
+
+        // Done
+        return requestFilename;
+    }
+
+    /**
+     * Normally we prepend the servlet path to the request filename, but for some special paths we do not.
+     * This method makes that determination.
+     *
+     * @param request HTTP request
+     * @return true if request filename does not get the servlet path prepended
+     */
+    boolean isSpecialRequestFilename(HttpServletRequest request) {
         // http://localhost:8888/context/servlet/folder/file.js
         // ->
         // /servlet/folder/file.js
@@ -445,14 +472,11 @@ public class StaticFileServer implements StaticFileHandler {
         // http://localhost:8888/context/servlet/sw.js
         // ->
         // /sw.js
-        if (request.getPathInfo() == null) {
-            return request.getServletPath();
-        } else if (request.getPathInfo().startsWith("/" + VAADIN_MAPPING)
-                || APP_THEME_PATTERN.matcher(request.getPathInfo()).find()
-                || request.getPathInfo().startsWith("/sw.js")) {
-            return request.getPathInfo();
-        }
-        return request.getServletPath() + request.getPathInfo();
+        final String pathInfo = request.getPathInfo();
+        return pathInfo != null
+          && (pathInfo.startsWith("/" + VAADIN_MAPPING)
+                || APP_THEME_PATTERN.matcher(pathInfo).find()
+                || pathInfo.startsWith("/sw.js"));
     }
 
     /**
--- a/flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/StaticFileServerTest.java
@@ -277,7 +277,7 @@ public class StaticFileServerTest implements Serializable {
     @Test
     public void isResourceRequest() throws Exception {
         fileServer.writeResponse = false;
-        setupRequestURI("", "/static", "/file.png");
+        setupRequestURI("", "/static", "/static/file.png");
         Mockito.when(servletService.getStaticResource("/static/file.png"))
                 .thenReturn(new URL("file:///static/file.png"));
         Assert.assertTrue(fileServer.serveStaticResource(request, response));
@@ -286,7 +286,7 @@ public class StaticFileServerTest implements Serializable {
     @Test
     public void isResourceRequestWithContextPath() throws Exception {
         fileServer.writeResponse = false;
-        setupRequestURI("/foo", "/static", "/file.png");
+        setupRequestURI("/foo", "/static", "/static/file.png");
         Mockito.when(servletService.getStaticResource("/static/file.png"))
                 .thenReturn(new URL("file:///static/file.png"));
         Assert.assertTrue(fileServer.serveStaticResource(request, response));
@@ -818,7 +818,7 @@ public class StaticFileServerTest implements Serializable {
 
     @Test
     public void serveStaticResource() throws IOException {
-        setupRequestURI("", "/some", "/file.js");
+        setupRequestURI("", "/some", "/some/file.js");
         String fileData = "function() {eval('foo');};";
 
         Mockito.when(servletService.getStaticResource("/some/file.js"))
@@ -1103,7 +1103,7 @@ public class StaticFileServerTest implements Serializable {
         long browserLatest = 123L;
         long fileModified = 123L;
 
-        setupRequestURI("", "/some", "/file.js");
+        setupRequestURI("", "/some", "/some/file.js");
         Mockito.when(request.getDateHeader("If-Modified-Since"))
                 .thenReturn(browserLatest);
 


Legioth:
I don't think this is a proper solution. The purpose of `StaticFileServer` is to emulate the servlet container's "default" servlet for serving static resources in the cases where the Vaadin servlet mapping "gets in the way".

Without any specific servlet mapped to a path, a request to `contextPath/foo/bar.txt` would be served by the file from `src/main/webapp/foo/bar.txt` if built as a WAR file with standard Maven conventions, from `src/main/resources/META-INF/resources/foo/bar.txt` from a JAR file on the classpath based on the Web Fragment feature defined by Servlet 3, and some additional locations with Spring Boot conventions.

This shouldn't change just because a Vaadin servlet happens to be mapped to `/foo/*` (as long as there isn't a `@Route` that matches `bar.txt`). `StaticFileServer` should therefore retain the servlet path when looking up static resources so that a request to `contextPath/foo/bar.txt` serves the file from `src/main/webapp/foo/bar.txt` regardless of how Vaadin servlets are deployed.

archiecobbs:
Hi @Legioth,

> Without any specific servlet mapped to a path, a request to contextPath/foo/bar.txt would be served by the file from src/main/webapp/foo/bar.txt if built as a WAR file with standard Maven conventions

> This shouldn't change just because a Vaadin servlet happens to be mapped to `/foo/*` 

Well the Tomcat developers disagree with you. Tomcat's default servlet strips off the servlet path. See for example https://bz.apache.org/bugzilla/show_bug.cgi?id=42411

The root cause here is the current functionality that was changed by the "fix" for #13769.

That bug is just wrongheaded. It completely breaks the modularity of servlet mapping.

The bug reports states:

> In all cases, the static files (JS bundle files, vaadinPush script etc) should be loaded from VAADIN inside the context root.

What?? According to whom?? There is no justification given here, and this completely breaks a bunch of stuff.

> Thus if you have multiple servlet mappings, you still load e.g. web components only once from /VAADIN and not from both /myapp/VAADIN and /admin/VAADIN.

So at best, this is all for a completely trivial optimization. Not worth it!!

Did you read the part where **almost every add-on is broken now?**

Legioth:
If you explicitly deploy the `default` servlet to some specific path, then it's very clear that you do that explicitly for the purpose of making static resources available below that path instead _instead_ of the default.

When it comes to Vaadin's servlet path, it does so much more than just serve static resources, thus putting it in a different situation than in the Tomcat case. The path also affects static resources not related to Vaadin, and the path served by Vaadin would be _in addition_ to the default one (unless there's also some other servlet that overrides the default path). Can we for instance assume that an application with a custom access control setting for the static resource `foo.txt` would realize that when they deploy a Vaadin servlet to `/bar`, then they would need to apply the same security configuration not only for `/foo.txt` but also for `/bar/foo.txt`?

I realize that https://github.com/vaadin/flow/issues/13769 causes problems with add-ons, but the reason it was fixed in the first place was that the other way also caused problems. Another potential fix might be to change the default for resource annotations such as  `@StyleSheet` to interpret relative URLs relative to the context root rather than the current default that is relative to the base URI (which is effectively the same as the servlet path). It's also always possible to change add-ons to  explicitly use that mode by prefixing resource annotation values with `context://`, but that does of course require a separate fix to each affected add-on.

Legioth:
Or maybe the proper solution would simply be to revert https://github.com/vaadin/flow/issues/13769?

archiecobbs:
> I realize that https://github.com/vaadin/flow/issues/13769 causes problems with add-ons, but the reason it was fixed in the first place was that the other way also caused problems.

I would like to better understand this (the motivation for 13769)... can you describe the problems that 13769 was supposed to fix? No proper motivation was provided in the bug description itself (as I've already pointed out).

> The path also affects static resources not related to Vaadin, and the path served by Vaadin would be in addition to the default one

Good point! This bug is a big **security hole** because it causes every resource to be unexpectedly exposed at two different URL paths, when most security setups are designed to protect secure resources at the exactly one URL path - the one (and only one) they are expected to be accessible at.

> Or maybe the proper solution would simply be to revert https://github.com/vaadin/flow/issues/13769?

**Yes!** I believe this is the correct answer here... at least until someone can provide a valid counter-argument.

The change in #13769 has created a giant mess, and I've yet to hear any valid explanation of what was the supposed upside. Whatever it was, it's hard to imagine that it was worth the mess.

Legioth:
https://github.com/vaadin/flow/issues/13769 was introduced to deal with issues related to the Vite integration when using a custom servlet path.

Based on an internal discussion earlier today, we are leaning towards reverting and finding another solution to the Vite issues.

PR #14356, opened 2022-08-15
fix: Fix asset handling in Vite production mode

Artur-:


flow-server/src/main/resources/plugins/webpack-plugins.json
```diff
@@ -3,6 +3,7 @@
     "build-status-plugin",
     "application-theme-plugin",
     "theme-loader",
-    "theme-live-reload-plugin"
+    "theme-live-reload-plugin",
+    "rollup-plugin-postcss-lit-custom"
```
mshabarov, line 7: Webpack doesn't need this plugin.

mshabarov (changes requested):

flow-server/src/main/resources/plugins/webpack-plugins.json
Artur-: Indeed but there is no mechanism for installing Vite plugins as far as I know? There is just a mechanism for installing plugins for both and it is called "webpack plugins"

mshabarov: Oh, okay, this is an another story then. This is something to improve in the future.

mshabarov (approved):

Issue #14227, opened 2022-07-28
Service worker does not work correctly in combination with frontend views and servlet mapping


mcollovati:
### Description of the bug

If a Vaadin application is served on a nested path (e.g. '/nested/*` custom servlet mapping with also `/VAADIN/*` mapped), service worker is not able to handle frontend view when browser is offline.

Service worker javascript resources is served at `/contextPath/nested/sw.js` and it seems like it is configured to handle request under `/contextPath/nested/`.
When switching to offline mode and clicking on a link handled by Vaadin Router the requested path is like `/contextPath/VAADIN/views/myview/myview` and it is not served by service worker.

### Minimal reproducible example

Run `ServiceWorkerOnNestedMappingIT` from `flow-tests/test-ccdm-flow-navigation` with Vite enabled.
Tests works fine with Webpack.

### Expected behavior

Frontend view is accessible even offline.
Test defined in `ServiceWorkerOnNestedMappingIT` pass.

### Actual behavior

When offline the request for the frontend view is not handled by service worker.
Test defined in `ServiceWorkerOnNestedMappingIT` fail.

### Versions:
    - Vaadin / Flow version: 23.2
    - Java version: 11
    - Development mode:
    - Existing test in Flow code base

mshabarov:
I put the severity of this ticket to `low`, since the configuration affected is:
- development mode
- offline mode
- client-side views
- custom context and servlet url
In my opinion the combination of dev mode and offline mode don't make many sense and big impact. Production mode works fine.

Issue #13826, opened 2022-05-20
[vite] Update the Vite ping route in `sw.ts` when upgrading to Vite 3.0


vursen:
The ping route has changed in Vite 3.0 (https://github.com/vitejs/vite/pull/6819), so we'll need to update it also in `sw.ts` once we get to upgrading. Otherwise, the URL can end up accidentally cached.

**Before:** `/VAADIN/__vite_ping`. **After:** `/VAADIN/`.

Artur-:
Do you mean it now pings the base URL, i.e. `/VAADIN/` and we should. make sure that this very path is never cached?

Artur-:
i.e. in sw.ts we have
```
  registerRoute(
    ({ url }) => url.pathname.startsWith(`${scopePath}VAADIN/__vite_ping`),
    networkOnly
  );

  registerRoute(
    ({ url }) => url.pathname.startsWith(`${scopePath}VAADIN/`),
    networkFirst
  );

```

So we should delete the first and make the second `networkOnly`?

vursen:
> So we should delete the first and make the second networkOnly?

Not exactly. I guess we should add a new `networkOnly` route specifically for the base path and then we can remove the `__vite_ping` route.

Artur-:
Ah, sorry the second one was `startsWith` so we should just change the first one to be an exact match

PR #14357, opened 2022-08-16
fix!: Revert always load static files from context root (#13795)

### LINKED ISSUE #13190: absolute /VAADIN path for scripts rerefenced by index.html with custom servlet mapping
### Description of the bug
In development mode, `<head>` contains scripts with absolute `/VAADIN/` prefix, e.g.:

```html
<script type="module" src="/VAADIN/@vite/client"></script>
<script type="module" src="/VAADIN/generated/vaadin.ts"></script>
<script type="module" src="/VAADIN/generated/vite-devmode.ts"></script>
```

This only works when Vaadin servlet is mapped to root (`/*`), but fails to load script when a custom servlet mapping (eg `/vaadin/*`) is used, resulting in a blank page.
Problem appears in both plain java (`@WebServlet(urlPatterns=...)`) and spring boot based projects (`vaadin.urlMapping` setting)

To be noted that trying to access resources prepending url mapping does not works for all of them. For example, giving a `/vaadin/*` mapping,  `/vaadin//VAADIN/generated/vaadin.ts` and `/vaadin/VAADIN/generated/vite-devmode.ts` work, but `/vaadin/VAADIN/@vite/client` returns 404.

Similar ticket related to context-path: #12388

To reproduce 
- create a new Vaadin 23 project with start.vaadin.com, activating Vite for frontend build
- update `application.properties`, setting a custom url mapping (e.g. `vaadin.urlMapping=/vaadin/*`)
- start application, go http://localhost:8888/vaadin/ and observe that requests starting with /VAADIN results in a 404.

Alternative way to reproduce
- Download plain java starter from https://github.com/vaadin/skeleton-starter-flow
- Add a custom servlet to the projet
   ```java
   @WebServlet(urlPatterns = {"/vaadin/*" }, asyncSupported=true)
   public class MyServlet extends VaadinServlet {
   }
   ```
- Activate Vite support by adding the file `src/main/resources/vaadin-featureflags.properties` with the following content
  ```
  com.vaadin.experimental.viteForFrontendBuild=true
  ```
- start application with `mvn jetty:run`, go http://localhost:8888/vaadin/ and observe that requests starting with /VAADIN results in a 404. 

### Expected behavior

An application configured to use Vite for frontend build should work also with custom servlet mapping.

### Actual behavior

Scripts referenced in index.html are not loaded (HTTP 404) and a blank page is shown.

### Versions:
    - Vaadin / Flow version: 23
    - Java version: 11.0.13
    - OS version: Ubuntu 21.10

mshabarov:
Changes static files loading so that:
1. The overall behavior is reverted to that in 23.0, i.e. static files are loaded from /context/servlet/VAADIN instead of /context/VAADIN
2. When using Vite in dev mode, uses the first deployed `VaadinServlet` to serve static resources. If you have both a `/foo/*` and a `/bar/*`mapped servlet, then if `/foo` is the servlet deployed first, also `/bar/` requests will use `/foo/VAADIN` for static resources.

This reverts commit af975389 

Fixes https://github.com/vaadin/flow/issues/14239
Fixes #13190

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
@@ -598,9 +598,9 @@ public class VaadinServlet extends HttpServlet {
     /**
      * For internal use only.
      *
-     * @return the first vaadin servlet that was reqistered
+     * @return the vaadin servlet used for frontend files in development mode
      */
-    public static String getFirstMapping() {
+    public static String getFrontendMapping() {
         return frontendMapping;
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
@@ -140,10 +140,14 @@ public class VaadinServlet extends HttpServlet {
 
             synchronized (VaadinServlet.class) {
                 if (frontendMapping == null) {
-                    String definedPath = getService()
-                            .getDeploymentConfiguration().getInitParameters()
-                            .getProperty(
-                                    INTERNAL_VAADIN_SERVLET_VITE_DEV_MODE_FRONTEND_PATH);
+                    String definedPath = null;
+                    DeploymentConfiguration deploymentConfiguration = getService()
+                            .getDeploymentConfiguration();
+                    if (deploymentConfiguration != null) {
+                        definedPath = deploymentConfiguration
+                                .getInitParameters().getProperty(
+                                        INTERNAL_VAADIN_SERVLET_VITE_DEV_MODE_FRONTEND_PATH);
+                    }
                     if (definedPath != null) {
                         frontendMapping = definedPath;
                     } else {
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
@@ -18,6 +18,7 @@ package com.vaadin.flow.server;
 import javax.servlet.ServletConfig;
 import javax.servlet.ServletContext;
 import javax.servlet.ServletException;
+import javax.servlet.ServletRegistration;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -28,6 +29,7 @@ import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
 import com.vaadin.flow.component.UI;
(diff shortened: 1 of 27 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -126,12 +138,47 @@ public void init(ServletConfig servletConfig) throws ServletException {
...
+                        List<String> mappings = new ArrayList<>();
+                        Map<String, ? extends ServletRegistration> servletRegistrations = this
+                                .getServletContext().getServletRegistrations();
+                        if (servletRegistrations != null
+                                && !servletRegistrations.isEmpty()) {
+                            // This should only happen in unit tests
+                            mappings.addAll(servletRegistrations
+                                    .get(this.getServletName()).getMappings());
+                            frontendMapping = mappings.get(0);
+                            Collections.sort(mappings);
```
taefi, line 163: Is there a need for this `mapping` local variable to be sorted?

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -548,4 +595,13 @@ private VaadinServletContext initializeContext() {
         return vaadinServletContext;
     }
 
+    /**
+     * For internal use only.
+     *
+     * @return the first vaadin servlet that was reqistered
+     */
+    public static String getFirstMapping() {
+        return frontendMapping;
```
taefi, line 604: Just a minor suggestion: would it make sense to create uniformity between the field name and its getter (to rename either of them, most likely the field)? It seems we just pick the first mapping from the list of mappings, and then we call it `frontendMapping`, but having a field named this way somehow implies that you could specifically set the `frontendMapping` unless you read the `public void init(ServletConfig servletConfig)` method implementation.

taefi (approved):

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
taefi, line 163: Is there a need for this `mappings` local variable to be sorted?

taefi (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
Artur-: Well ... the mapping should obviously be sorted before `mappings.get(0)`... It is a way to ensure that `frontendMapping` is deterministic as `getMappings()` return a set.

Artur-: Changed

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
Artur-: Renamed to `frontendMapping`. Spring sets this through a property and being able to set "the first deployed servlet mapping" makes no sense

taefi (approved):

mshabarov:
Got these errors when running 23.2.0.beta2 starter with this patch:

```
NetworkOnly.js:93 Uncaught (in promise) no-response: The strategy could not generate a response for 'http://localhost:54524/VAADIN/'. The underlying error is TypeError: Failed to fetch.
    at NetworkOnly._handle (http://localhost:8080/sw.js:3617:19)
    at async NetworkOnly._getResponse (http://localhost:8080/sw.js:1642:24)
```

So it requests `http://localhost:54524/VAADIN/` for some reason and fails.

taefi:
Tested #14239 with the changes in this PR, and it works as expected.

Artur-:
> Got these errors when running 23.2.0.beta2 starter with this patch:
> 
> ```
> NetworkOnly.js:93 Uncaught (in promise) no-response: The strategy could not generate a response for 'http://localhost:54524/VAADIN/'. The underlying error is TypeError: Failed to fetch.
>     at NetworkOnly._handle (http://localhost:8080/sw.js:3617:19)
>     at async NetworkOnly._getResponse (http://localhost:8080/sw.js:1642:24)
> ```
> 
> So it requests `http://localhost:54524/VAADIN/` for some reason and fails.

Vite 3 uses `/VAADIN` and the Vite port as its ping URL (see #13826) so you see this error at least if you shut down the server while keeping the app open. Not sure why you would see it otherwise but also it should have nothing at all to do with this PR

mshabarov:
https://github.com/vaadin/flow/issues/14227 is not solved, but let's fix it in a separate PR.

taefi (approved):

PR #14359, opened 2022-08-16
fix: force connectionLost when Atmosphere resource is resumed

### LINKED ISSUE #13660: First request to vaadin application fails
### Description of the bug

The first Request to my Application always ends up printing errors to the console and stops working. After a refresh everything works as intended.

Example messages:
```
Gave up waiting for message 3 from the server
Received resync message with id 4 while waiting for 3
Assertion error: undefined
Node id is not registered with this tree
```

The same issue appears when i restart my application and try to use it after it automaticlly reconnected. When i refresh the page everything works fine again.

I remember back in Vaadin 8 (from which i am migrating) there was a "Session expired" message where you can click on to refresh the page. Back there everything was working fine. Is there something like this in the current vaadin version or does it just autoreconnect? 

### Expected behavior

The application should not fail and even work in the first request.

I can't find and good minimal reproducible example.

### Versions

- Vaadin / Flow version: 23.0.5
- Java version: 11.0.11
- OS version: Windows 10
- Application Server (if applicable): Spring Boot 2.6.6

mcollovati:
## Description

When Push is configured with long polling transport, sometimes after server restart
Atmosphere resource is in a 'resumed' state. In this situation the UIDL message
is not set to the client, but the internal serverId counter is incremented, causing
an endless resynchronization loop because of inconsistency with client stored serverId.
This change forces the Push connection to be considered lost when pushing in resumed state,
so that a fresh connection can be established by the client.

Fixes #13660

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
@@ -191,6 +191,15 @@ public class AtmospherePushConnection implements PushConnection {
             } else {
                 state = State.RESPONSE_PENDING;
             }
+        } else if (resource != null && resource.isResumed()) {
+            // This can happen for long polling and will prevent the message to
+            // reach the client
+            // In this case we should not create UIDL to prevent an infinite
+            // resynchronization loop because of serverId never sent to the
+            // client
+            getLogger().debug(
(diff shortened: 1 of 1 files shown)


mcollovati:
This change reduces the error rate but does not fix the issue

PR #14360, opened 2022-08-16
feat: Backporting automatically use available converters in Binder 

melihkurtaran:
## Description

When binding a field and a property which do not have the same value type
an exception is thrown. With this change, an attempt is made to
automatically pick a suitable converter from the ones provided by the framework.

It was implemented for Vaadin 23  (#13373) now backporting to version 14

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist


mcollovati:
@melihkurtaran Some tests are failing, and I think it is because they also consider the changes from this PR
https://github.com/vaadin/flow/pull/13340

Perhaps it should also be back ported. Otherwise, the tests should be rewritten

PR #14389, opened 2022-08-19
fix: No client updates when Element::setText does nothing

### LINKED ISSUE #14388: Element.setText() causes unnecessary updates on client
### Description of the bug

When java code invokes Element.setText("someText") this always results in dom changes being sent to the browser, even if the text is not changing. 
While not a bug, it creates unnecessary network traffic and javascript code active, especially if there are many labels etc. on the screen.
It is cumbersome for application code to keep track of the current state of each element.
For other things like 'setAttribute()' or 'setProperty()', flow already detects that no change happens and therefore does not send updates. So the same logic should be in effect for text nodes in elements.

At the moment, even when nothing changes, the json sent to browser after a roundtrip contains the statements to remove all children of the element, and then again append the (already existing, unmodified) text node to the children list of that element.

### Expected behavior

when nothing changes the json returned in a roundtrp should not contain any changes.

Label lab = new Label();

init() {
   add(lab);
   lab.setText("hello");
   add(new Button("update", e -> lab.setText("hello")));
}

open this view, press button -> the json returned should not contain any changes related to the dom elements of the label.

### Versions

- Vaadin / Flow version: 23
- Java version: any
- OS version: n/a

Manfred-on-github:
## Description
when invoking setText(sometext) on an Element, using the same text as before, this still triggers some dom changes being sent to browser, even though nothing changes. This causes unnecessary network traffic and browser code to run.
The same situation is handled correctly for setProperty(prop,value) etc., so it should work the same way also for setText().

Fixes #14388

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/dom/Element.java
+++ b/flow-server/src/main/java/com/vaadin/flow/dom/Element.java
@@ -1071,7 +1071,11 @@ public class Element extends Node<Element> {
     private void setTextContent(String textContent) {
         Element child;
         if (getChildCount() == 1 && getChild(0).isTextNode()) {
-            child = getChild(0).setText(textContent);
+            Element oneChild = getChild(0);
+            child = oneChild.setText(textContent);
+            if (oneChild == child) {
+                return;
+            }
         } else {
             child = createText(textContent);
         }
--- a/flow-server/src/main/java/com/vaadin/flow/dom/Element.java
+++ b/flow-server/src/main/java/com/vaadin/flow/dom/Element.java
@@ -1069,18 +1069,13 @@ public class Element extends Node<Element> {
     }
 
     private void setTextContent(String textContent) {
-        Element child;
         if (getChildCount() == 1 && getChild(0).isTextNode()) {
-            Element oneChild = getChild(0);
-            child = oneChild.setText(textContent);
-            if (oneChild == child) {
-                return;
-            }
+            getChild(0).setText(textContent);
         } else {
-            child = createText(textContent);
+            Element child = createText(textContent);
+            removeAllChildren();
+            appendChild(child);
         }
-        removeAllChildren();
-        appendChild(child);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -1071,7 +1071,11 @@ public Element setText(String textContent) {
     private void setTextContent(String textContent) {
         Element child;
         if (getChildCount() == 1 && getChild(0).isTextNode()) {
-            child = getChild(0).setText(textContent);
+            Element oneChild = getChild(0);
+            child = oneChild.setText(textContent);
+            if (oneChild == child) {
```
mshabarov, line 1076: `Element.setText` returns `this`, so `child` always points to the same object as `oneChild` does.
Thus, the condition is always `true` and it never removes all children and append children, if it's a text node.
So the following patch does the same.

```java
    private void setTextContent(String textContent) {
        Element child;
        if (getChildCount() == 1 && getChild(0).isTextNode()) {
            getChild(0).setText(textContent);
            return;
        } else {
            child = createText(textContent);
            removeAllChildren();
            appendChild(child);
        }
    }
```

Could you elaborate more why you're comparing the child element objects references?
Thanks

mshabarov (changes requested):
@Manfred-on-github thank you for contribution!
The motivation for this fix is clear, but I'm not sure I get the implementation details right.

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
Manfred-on-github: you are right, this code does the same -  
however I did not want to assume that setText() will always return the same object, therefore, being extra careful, have added this if.
So your choice what to accept as final version...

mshabarov: The contract of `setText` methods says the returned object is `this` element and we have many tests relying on that, so in my opinion this reference check is not needed.

I would propose to make this method look like:
```java
    private void setTextContent(String textContent) {
        Element child;
        if (getChildCount() == 1 && getChild(0).isTextNode()) {
            getChild(0).setText(textContent);
        } else {
            child = createText(textContent);
            removeAllChildren();
            appendChild(child);
        }
    }
```
Unit test looks good.
Also I've tested this patch with the started project, works as expected.

mshabarov (changes requested):

mshabarov:
@Manfred-on-github please also run `mvn formatter:format`.

mshabarov (approved):

PR #14390, opened 2022-08-19
fix: initialize PwaHandler only once

mcollovati:
PwaHandler has an isInitialized flag that is checked on handleRequest method,
but never initialized, causing init method to be invoked on every request.
This change sets the flag within init method.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/PwaHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/PwaHandler.java
@@ -123,7 +123,6 @@ public class PwaHandler implements RequestHandler {
                     }
                     return true;
                 });
-        isInitialized = true;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/server/communication/PwaHandler.java
```diff
@@ -123,6 +123,7 @@ private void init(PwaRegistry pwaRegistry) {
                     }
                     return true;
                 });
+        isInitialized = true;
```
mshabarov, line 126: Would it be semantically more correct to place it right after ` init(pwaRegistry);`?

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/PwaHandler.java
mcollovati: I would say that the handler `isInitialized` only when all stuff is done, rather than when entering the method.
But it should not make a big difference since init is executed with session locking.

mshabarov (approved):

taefi (approved):

PR #14397, opened 2022-08-19
fix: Not modify browser history if location has no hash

mshabarov:
Fixes the hashed segment removal logic:
if no segment with hash is present in the location,
while navigating between Flow and legacy views/components,
the update of the browser history is not needed.

Fixes https://github.com/vaadin/multiplatform-runtime/issues/115


taefi (approved):

mshabarov:
`2.7` and `2.8` doesn't need this fix, they were added by mistake.

PR #14402, opened 2022-08-22
feat: Give an access to ViewAccessChecker in subclasses

mshabarov:
Makes ViewAccessChecker be accessible in the VaadinWebSecurity subclasses,
for example, to be able to override the security configuration for SSO.

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
@@ -437,6 +437,23 @@ public abstract class VaadinWebSecurity {
         return requestUtil.applyUrlMapping(path);
     }
 
+    /**
+     * Vaadin views access checker bean.
+     * <p>
+     * This getter can be used in implementing class to override logic of
+     * <code>VaadinWebSecurity.setLoginView</code> methods and call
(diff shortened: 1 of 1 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -437,6 +437,23 @@ protected String applyUrlMapping(String path) {
...
+     * <code>VaadinWebSecurity.setLoginView</code> methods and call
+     * {@link ViewAccessChecker} methods explicitly.
+     * <p>
+     * Note that this bean is a field-autowired, thus this getter returns
+     * <code>null</code> when called from the constructor of implementing class.
+     *
+     * @return {@link ViewAccessChecker} bean used by this VaadinWebSecurity
+     *         configuration.
+     */
+    protected ViewAccessChecker getViewAccessChecker() {
```
Artur-, line 453: Why wouldn't you just `@Autowire` it?

mshabarov: Can be an option as well. But with having a getter, you can avoid an extra code in your implementation class.

taefi (approved):

PR #14403, opened 2022-08-22
test: Redirect test output to test-output.log instead of stderr

Artur-:
This reduces log pollution and makes it easier to see what actually fails. 
Reduces `mvn test` output in `flow-server` from 2621 to 708 rows.

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -113,6 +113,7 @@ jobs:
           name: saved-outputs
           path: |
             **/target/*-reports/*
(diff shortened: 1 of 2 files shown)


mshabarov (approved):

mshabarov:
@Artur- are we waiting for any CI configuration before proceed with this change?

Artur-:
1. The GitHub build should be changed to collect the log files
2. The failsafe plugin should also be updated to log instead of output

PR #14408, opened 2022-08-22
feat: allow to set a feature flag using a system property

### LINKED ISSUE #14407: Allow to set a feature flag using a system property
### Describe your motivation

While searching of how to run some integration tests twice with different values for a feature flag, we thought that being able to override its value with a system property would be a good solution, usable from a GitHub action.

I'd like to provide a PR with the code needed to set a feature flag using a Java system property.

I tried to load the flag value from the classpath, but I couldn't, since we're running some `ChromeBrowserTest`s and the feature flag loading is hardwired. I should rewrite the `properties` file in the sources, which is not easy when other flags are present.

cromoteca:
Accepts a value for a feature flag passed through a system property and overrides existing value.

Closes #14407

### DIFF
--- a/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
+++ b/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
@@ -201,17 +201,25 @@ public class FeatureFlags implements Serializable {
             for (Feature f : features) {
                 f.setEnabled(false);
             }
-            return;
+        } else {
+            try (FileInputStream propertiesStream = new FileInputStream(
+                    featureFlagFile)) {
+                getLogger().debug("Loading properties from file '{}'",
+                        featureFlagFile);
+                loadProperties(propertiesStream);
+            } catch (IOException e) {
+                throw new UncheckedIOException(
+                        "Failed to read properties file from filesystem", e);
+            }
         }
 
-        try (FileInputStream propertiesStream = new FileInputStream(
-                featureFlagFile)) {
-            getLogger().debug("Loading properties from file '{}'",
-                    featureFlagFile);
-            loadProperties(propertiesStream);
-        } catch (IOException e) {
-            throw new UncheckedIOException(
-                    "Failed to read properties file from filesystem", e);
+        // Allow users to override a feature flag with a system property
+        for (Feature f : features) {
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
```diff
@@ -201,17 +201,25 @@ public void loadProperties() {
...
-                featureFlagFile)) {
-            getLogger().debug("Loading properties from file '{}'",
-                    featureFlagFile);
-            loadProperties(propertiesStream);
-        } catch (IOException e) {
-            throw new UncheckedIOException(
-                    "Failed to read properties file from filesystem", e);
+        // Allow users to override a feature flag with a system property
+        for (Feature f : features) {
+            var prop = System.getProperty("featureFlag-" + f.getId());
```
knoobie, line 218: Should the system property be prefixed with `vaadin-` like the name of the property file?

flow-server/src/test/java/com/vaadin/experimental/FeatureFlagsTest.java
```diff
@@ -193,6 +193,16 @@ public void enabledFeatureFlagsMarkedInStatsWhenToggled()
                 hasUsageStatsEntry("flow/featureflags/exampleFeatureFlag"));
     }
 
+    @Test
+    public void featureFlagShouldBeOverridableWithSystemProperty()
+            throws IOException {
+        System.setProperty("featureFlag-exampleFeatureFlag", "true");
```
taefi, line 199: In general, if we set some system properties during a test, it is highly recommended to hold their pre-change states and restore them after the test ends. The impact of setting this specific property might not be very obvious but in the future, this might mess up some other test results.

flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
taefi: Thanks, @cromoteca for the contribution, I also prefer to see a `vaadin-` prefix with the property name to prevent any possible clash with some unwanted properties, even though it might be very rare.

taefi (changes requested):

taefi (approved):

Issue #11188, opened 2021-06-09
Provide helper for keeping state while a component is attached


Legioth:
It's a relatively common pattern to initialize some state when a component is attached and then do the reverse when the component is detached. One typical example of this is to register the UI for background updates to be pushed out. Another example is with using `executeJs` to initialize the client-side of a component but then wanting to cancel the registration on detach to avoid double init in case the  component is attached and detached again during the same round trip.

The most straightforward way of implementing this is by overriding `onAttach` and `onDetach`.
```java
public class MyComponent extends Div {
  private Registration subscribeRegistration;

  @Override
  protected void onAttach(AttachEvent attachEvent) {
    subscribeRegistration = registerForPush(this, attachEvent.getUI());
    super.onAttach(attachEvent);
  }

  @Override
  protected void onDetach(DetachEvent detachEvent) {
    super.onDetach(detachEvent);
    subscribeRegistration.remove();
    subscribeRegistration = null;
  }
}
```

There's a whole bunch of boilerplate code associated with this technique, and the logic does also suffer from lack of locality since the concerns are spread out over two separate instance methods and one instance field.

Another alternative is to use `addAttachListener` and `addDetachListener` to keep all logic in one method. Furthermore, `unregisterListener` can be used to create a one-time detach listener which in turn allows keeping the state local instead of relying on an instance field.

```java
public class MyComponent extends Div {
  public MyComponent() {
    addAttachListener(attachEvent -> {
      Registration subscribeRegistration = registerForPush(this, attachEvent.getUI());

      addDetachListener(detachEvent -> {
        detachEvent.unregisterListener();
        subscribeRegistration.remove();
      });
    });
  }
}
```

There's still some boilerplate in this pattern, and it's also a pattern that isn't immediately obvious.

One additional challenge is in case the logic is outside the component itself. In this case, the same technique with `addAttachListener` can be used, but there's also the need for additional boilerplate to run the attach logic immediately in case the target component is already attached.

As an alternative, there could be a helper method either as an instance method in `Component` or as a static helper in `ComponentUtil` that takes an attach handler that is supposed to return a matching detach handler. The attach handler immediately receives the `UI` instance rather than an event object that needs unpacking. The helper would also take care of running the attach handler immediately in case the method is currently attached and it would finally also take care of running the detach handler if one is present in case the registration of the attach handler itself is removed.

The method signature could thus look like this if implemented as an instance method:
```
[...]

PR #14414, opened 2022-08-23
feat: added a proxy of UI.access to the component level

mstahv:
## Description

I was looking at the greatly improved docs (by @tarekoraby  💪) for handling long running processes and started to think we are doing something wrong. The concept of UI has pretty much vanished from Vaadin since v10, but then we are doing some rather complex looking activities with it in rather simple use cases.

I’d thus propose to bring down the UI.access helper to the level of Component. I don’t think the actual implementation needs to change at all, it can simply proxy the call to the UI the component is attached to. And as discussed earlier, it is better to have the component instead of the thread local the way to get UI, this would guide people to better habits in that sense as well.

Opened a draft PR for discussion as the improvement looks so obvious to me. No tests yet, but otherwise I think it could be pretty much ready.
## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -15,15 +15,6 @@
  */
 package com.vaadin.flow.component;
 
-import java.io.Serializable;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Locale;
-import java.util.Optional;
-import java.util.stream.Stream;
-import java.util.stream.Stream.Builder;
-
 import com.vaadin.flow.component.internal.ComponentMetaData;
 import com.vaadin.flow.component.template.Id;
 import com.vaadin.flow.dom.Element;
@@ -34,8 +25,19 @@ import com.vaadin.flow.i18n.I18NProvider;
 import com.vaadin.flow.internal.AnnotationReader;
 import com.vaadin.flow.internal.nodefeature.ElementData;
 import com.vaadin.flow.server.Attributes;
+import com.vaadin.flow.server.Command;
 import com.vaadin.flow.server.VaadinService;
 import com.vaadin.flow.shared.Registration;
+
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+import java.util.Optional;
+import java.util.concurrent.Future;
+import java.util.stream.Stream;
+import java.util.stream.Stream.Builder;
 
 /**
  * A Component is a higher level abstraction of an {@link Element} or a
@@ -787,6 +789,51 @@ public abstract class Component
             }
         }
         return null;
+    }
+
+    /**
+     * Provides exclusive access to this component and the UI it is attached to
+     * from outside a request handling thread.
+     * <p>
+     * The given command is executed while holding the session lock to ensure
+     * exclusive access to this UI. If the session is not locked, the lock will
+     * be acquired and the command is run right away. If the session is
+     * currently locked, the command will be run before that lock is released.
+     * </p>
+     * <p>
+     * RPC handlers for components inside this UI do not need to use this method
+     * as the session is automatically locked by the framework during RPC
+     * handling.
+     * </p>
+     * <p>
+     * Please note that the command might be invoked on a different thread or
+     * later on the current thread, which means that custom thread locals might
+     * not have the expected values when the command is executed.
+     * {@link UI#getCurrent()}, {@link VaadinSession#getCurrent()} and
+     * {@link VaadinService#getCurrent()} are set according to this UI before
+     * executing the command. Other standard CurrentInstance values such as
+     * {@link VaadinService#getCurrentRequest()} and
+     * {@link VaadinService#getCurrentResponse()} will not be defined.
+     * </p>
+     * <p>
+     * The returned future can be used to check for task completion and to
+     * cancel the task.
+     * </p>
+     *
+     * @see UI#getCurrent()
+     *
+     * @param command
+     *            the command which accesses the UI
+     * @throws UIDetachedException
+     *             if the UI is not attached to a session (and locking can
+     *             therefore not be done)
+     * @return a future that can be used to check for task completion and to
+     *         cancel the task
+     * @since 23.3.0
+     */
+    public Future<Void> access(final Command command) {
+        UI ui = getUI().orElseThrow(() -> new UIDetachedException());
+        return ui.access(command);
     }
 
 }


mvysny:
The problem is that `Component.getUI()` (as called from the proposed `Component.access()` function) is not thread-safe and should not be called from the background thread.

One solution would be to make `Component.getUI()` thread-safe; however I believe that the simplest rule to follow is that "The Component API must not be accessed from background threads". Instead I would propose some sort of VaadinThreadContext interface which would only offer `access()` function and will be "given" to background threads. I don't have anything concrete, it's just an idea.

Legioth:
As an alternative, I would suggest a more high-level approach rather than low-level callbacks. E.g. something along the lines of https://github.com/Legioth/accessor.

mstahv:
Is that thread safety with getUI() an academical or real problem 🤔 When does it happen? When the component has been removed?

mstahv:
I remember the Accessor add-on you Leif put together (after some of our discussions 🤔), but not sure if that makes things easier for Vaadin developers as such 🤷‍♂️

mvysny:
> Is that thread safety with getUI() an academical or real problem thinking

:-D The worst class of errors are threading errors: they happen randomly, lead to unpredictable behavior, require intimate knowledge of Java Memory Model and are very hard to fix afterwards. I assure you such errors are very real.

> When does it happen? When the component has been removed?

I would have to analyze the `getUI()` code flow to answer that. I took a brief look and there could be a call path such as getUI() -> ComponentUtil.findParentComponent() -> getInnermostComponent(Component) -> getInnermostComponent(Composite) -> Composite.getContent() which could lead to a potential composite lazy content initialization which is a lot of (both Vaadin and cuustomer) code that was never meant to be thread-safe. To put it simply, that is a big NO-NO to call from a background thread.

In short, thread-safe methods can not call thread-unsafe methods without taking proper precautions. `Component.getUI()` is not thread-safe, therefore `Component.access()` is not thread-safe, which means that you can only call that method from the UI thread, which renders the function useless (or very dangerous since it would trick the customers to think it's okay to call it from the background thread).

The simplest solution would probably be something like:
```java
new Button("Foo", e -> {
  UIAccessor ui = new UIAccessor();
  executor.submit(() -> {
    ui.access(() -> {});
  });
});
```
UIAccessor would simply store UI as its private final field and would expose the sole function `submit()` which would be explicitly marked thread-safe.

mstahv:
Oh, now that I opened the implementation, I see that getUI isn't simply traversing upwards the component tree. I can imagine now that Composite.getContent could really cause harm in some cases. And cause weird exceptions, because UI is initialised in non-UI thread... I wonder if it could be implemented otherwise easily 🤔 Not only to making the problem here somewhat academical, but that could even fix random errors in many current Vaadin apps (which we currently though can just claim that are developed in a wrong way...).

Legioth:
With the specific example in the documentation, the trick is to move `getUI()` into the phase that runs in the click listener while still keeping things simple. We already have `UI::accessLater` for that purpose, and that is something that can also be scoped to `Component` if we want. With that, the example could be something like this:
```java
Consumer<String> showNotification = accessLater(Notification::show, () -> logger.log("Detached"));
ListenableFuture<String> future = backendService.longRunningTask();
future.addCallback(
    successResult -> showNotification("Task finished: " + successResult), 
    failureException -> showNotification("Task failed: " + failureException.getMessage()) 
);
```

mstahv:
That sounds like a quick and easy win 🥇, although I'd love to have that original callback version as well. 

Maybe also overload that accessLater with a version where no detachHandler is given🤔 Based on the JavaDocs, the error handler can be null as well, but then a UIDetachedException is thrown. Not sure what the default behaviour should be. Intuitively I'd say we should just eat the exception. If the UI is detached, then the UI update is just obsolete, even though some task has completed. It would be easy to implement a silent default detach handler in the overloaded method.

Legioth:
There's another approach that might be even better, which is to change the `@Async` method to return a `CompletableFuture` and then use an adaptation of this idea: https://twitter.com/leifastrand/status/1391766544124891147?t=SqDCYAdpP34IC8XKoIQ-2w&s=19

With an executor provider in `Component`, the example could thus be almost a oneliner:
```java
service.longRunning().handleAsync(
  (result, error) -> {
    if (error == null) {
      Notification.show("Finished: " + result);
    } else {
      Notification.show(”Failed: " + error.getMessage());
  }, getAccessExecutor());
```

Legioth:
I was also thinking about the Accessor add-on and realized it's for a related use case for which it might be good to have a symmetric solution.

What we have is really two fundamental reasons for updating components from a background thread: persistent subscriptions and one-off tasks. A persistent subscription can trigger an arbitrary number of UI updates and needs to be explicitly closed when it's no longer used to avoid leaking memory. Closing can be explicitly triggered or more conveniently attached to the detach event of a specific component, which also implies that attaching the component again should reopen the subscription (and get the UI to `access` in that way). The one-off case is a special variant of this where exactly one update is expected (either a success value or an error (possibly a timeout)) which also means that it's typically not necessary to explicitly clean up anything (though there are also tasks that are meaningful to cancel eagerly to avoid wasting resources).

I was originally thinking that it would be meaningful to have a high level abstraction on top of `access` to support both of the cases. There's however also another alternative which is to reuse the idea from https://github.com/vaadin/flow/issues/11188 for handling attaching and detaching. The typical Accessor example could then be put together like this
```java
Consumer<String> handleMessage = accessLater(this::addMessage);
whileAttached(ui -> DemoSubscription.subscribe(handleMessage));
```

Legioth:
One more aspect to consider here is whether this case is important enough to bloat the sacred `Component` API rather than giving this a dedicated class (e.g. the suggested `UIAccessor`) or dumping it in `ComponentUtil`. Either of those approaches would be functionally equivalent except for the need for passing an explicit `this` reference instead of the implicit reference in an instance method invocation. The drawback is of course that discoverability suffers when the entry point isn't on a type that is already in context for auto competition, even though that has also been the case previously with the `access` method only in `UI`.

MarcinVaadin:
As that PR dived deeper into thread safe UI access I would not merge the change as proposed by author. 

@mstahv - please comment if you would apply @Legioth proposed changes or would you like to drop the idea / create separate PR?

Legioth:
Making `getUI()` safe to use from background threads will introduce some overhead and complexity that I would want to avoid unless we have no other choice. As an alternative, we could also have a per-component list of pending access tasks that are flushed when the component is attached.

On the other hand, I feel like `Component::access` would just be lipstick on the pig - we would make our bespoke low-level mechanism slightly more convenient to use rather than taking this though as an opportunity to introduce a high-level mechanism. Sure, it's a little bit more effort (especially on the design side) to raise the abstraction level but I still believe that would be more worthwhile as long as there's enough momentum to actually carry it through.

mvysny:
My two cents. For the record: I am not an academic by any means - I am always coming from the place of practicality and usability.

The current PR of `Component.access()` API will only allow the customer to shoot itself in the foot easier, in the worst possible way: the customer will get unreproducible random behavior in production under heavy load - the worst kinds of bugs - and he will not have any idea what he did wrong. This will even be worsened when running on ARM. The reason is that x86 flushes memory caches fairly aggressively which causes thread-unsafe code to "mostly work"; ARM is much lazier when it comes to memory flushes.

The proposal would be to make `getUI()` thread-safe. That is a very bad idea. As we know, all attempts to make UI frameworks multi-threaded ended up being complete failures, and it is generally agreed that an UI framework should be thread-unsafe and expected to be run from the UI event loop (or, with Vaadin, with the session lock held). Any threading code+locking mechanism should be outside of the common UI code, as part of a very specific class (say `UIAccess`) with just one intent: accessing UI from bg threads in a safe way. Which is a direct contradiction to making `getUI()` thread-safe.

Even today, it's really hard to prove that the current `UI.access()` is actually thread-safe. The reading of `UIInternals.session` happens-before `VaadinSession.service`, true, but there's no obvious happens-before relationship such that write to `VaadinSession.service` happens-before write to `UIInternals.session`, therefore the `VaadinSession.service` may randomly be null during the call to `UI.access()` with some unhappy timing. There's no `@GuardedBy` documentation for `VaadinSession.service` or anything like that, therefore the correctness of UI.access() is not known. In a perfect world, there would be no UI.access(), no Session.access() and no Session.lock, and everything would be handled by the `UIAccess` class, which would be meticulously documented with respect to thread-safety, according to the patterns presented in the "Java Concurrency In Practice" book.

If we let some random guy figure how this could be done, such attempt will almost surely end up with more thread-unsafe code in the Vaadin framework, making it a maintenance nightmare in the future.

mshabarov:
I wouldn't apply this method with the proposed implementation, because:
1. Excellent comment about [thread-safety](https://github.com/vaadin/flow/pull/14414#issuecomment-1223722940)
2. Low profit in API usage comparing to what we have now for async UI access, see [example](https://vaadin.com/docs/latest/advanced/long-running-tasks/#handling-long-running-tasks-asynchronously): it gives a shorthand method, yes, but it hides what happens actually inside this method. I'd prefer use a bit more low level API here, but let developer realise what he is doing and from where the UI instance is obtained.

Sorry @mstahv , but I don't see a profit of putting extra effort to make this API thread-safe against what developers achieve in the end: one line of code instead of two lines of code.

PR #14449, opened 2022-08-31
fix: Print a warning instead of failing when importing from lit/decorators

Artur-:


mshabarov (approved):
Tested with starter project containing frontend views - works as expected.

PR #14460, opened 2022-09-02
fix: Do not handle Vite special URLs in the service worker

### LINKED ISSUE #14451: Vite keeps pinging the server producing error logs in the browser console
### Description of the bug
Vite keeps pinging the old port of the server, even after restart of the application, producing error messages in the logs, like follows:
```
The FetchEvent for "http://localhost:53620/VAADIN/" resulted in a network error response: the promise was rejected.
NetworkOnly.js:93 Uncaught (in promise) no-response: The strategy could not generate a response for 'http://localhost:53620/VAADIN/'. The underlying error is TypeError: Failed to fetch.
    at NetworkOnly._handle (http://localhost:8080/sw.js:3617:19)
    at async NetworkOnly._getResponse (http://localhost:8080/sw.js:1642:24)
```
These errors are present in the console even for a new tab after app restart.
The only way to get rid of them is to close all tabs and restart the server.

1. Download starter project from start.vaadin.com for 23.2 prerelease.
2. Start the app
3. Stop the app
4. Notice the errors in the browser's console
5. Start app again, open the console of a new tab.
6. See the errors.

### Expected behavior
No errors in a new tabs after server restart.

### Actual behavior
Errors are present even after server restart.

### Versions:
    - Vaadin / Flow version: Vaadin 23.2.0.rc1
    - Java version: 17
    - OS version: Mac OS
    - Application Server (if applicable): Spring Boot
    - IDE (if applicable): Intellij Idea
    - Development or production mode: dev mode only

Artur-:
Fixes #14451

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw.ts
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/sw.ts
@@ -32,7 +32,7 @@ const offlinePath = OFFLINE_PATH;
 
 // Compute the registration scope path.
 // Example: http://localhost:8888/scope-path/sw.js => /scope-path/
-const scopePath = new URL(self.registration.scope).pathname;
+const scope = new URL(self.registration.scope);
 
 /**
  * Replaces <base href> in pre-cached response HTML with the service worker’s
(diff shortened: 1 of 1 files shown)


vursen (approved):

taefi:
Checked it locally, and this fixes the issue with the newly opened tab after server restart.

Artur-:
>However, the stale tabs are getting a net::ERR_CONNECTION_REFUSED as they keep pinging the old port

That would be expected as they are trying to fetch a URL which does not respond. That is now only 1 row in the console instead of 6 as it does not go through workbox / the service worker

taefi:
> > However, the stale tabs are getting a net::ERR_CONNECTION_REFUSED as they keep pinging the old port
> 
> That would be expected as they are trying to fetch a URL which does not respond. That is now only 1 row in the console instead of 6 as it does not go through workbox / the service worker

Does it make sense to swallow the error from the service worker like this: 
```
self.addEventListener('fetch', function (event) {
    var req = fetch(event.request).then(resolve).catch(() => {});
    event.respondWith(req);
});
```
and possibly log a WARN message asking the user to hard-refresh the tab if they insist to keep the old tabs open?

taefi (approved):

Artur-:
I am not sure what you are trying to solve now. Who should hard refresh what and why?

PR #14461, opened 2022-09-02
test: remove spring-data-jpa and h2 db from ITs

taefi:
H2 db instance sometimes fails to stop
and it leads to the failure of spring
tests. The data services mocked
and replaced jpa repositories. This also
improves the validation build time as it
removes the necessity of restarting the
in-memory db instance between tests.


flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/service/AccountService.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.math.BigDecimal;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+@Service
+public class AccountService {
+
+    private static final Map<String, Account> ACCOUNT_CACHE = new HashMap<>(2);
```
Artur-, line 15: This should be a concurrent map

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/service/UserInfoService.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+@Service
+public class UserInfoService {
+
+    public static String ROLE_USER = "user";
+    public static String ROLE_ADMIN = "admin";
+
+    private static final Map<String, UserInfo> USER_INFO_CACHE = new HashMap<>(
```
Artur-, line 16: This should be a concurrent map

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/test/java/com/vaadin/flow/spring/flowsecurity/UIAccessContextIT.java
```diff
@@ -23,15 +23,14 @@
...
 import org.openqa.selenium.WebDriver;
 
 public class UIAccessContextIT extends AbstractIT {
 
     @Test
     public void securityContextSetForUIAccess() throws Exception {
-        String expectedUserBalance = "Hello John the User, your bank account balance is $10000.00.";
-        String expectedAdminBalance = "Hello Emma the Admin, your bank account balance is $200000.00.";
+        String expectedUserBalance = "Hello John the User, your bank account balance is $10000.";
+        String expectedAdminBalance = "Hello Emma the Admin, your bank account balance is $200000.";
```
Artur-, line 33: Why does this change when you remove the database?

Artur- (changes requested):

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/test/java/com/vaadin/flow/spring/flowsecurity/UIAccessContextIT.java
taefi: This doesn't have anything to do with the DB. The locale settings on the CI server somehow forced this format of the BigDecimal outputs. First I thought maybe the CI server instance locale settings might be messed up, but the flow-data module also has some unit tests with checkings for the formatted outputs and they are passing without any change. Needs more investigation in case we insist on keeping the previous format.

Artur-: Doesn't seem super relevant although I don't get why the CI server would change when you drop the DB. Maybe the test app should explicitly set the locale to avoid surprises in the future

taefi: Maybe it was related to how precision was being set on a BigDecimal object which is being instantiated by JPA when reading from the h2 DB, even though, there wasn't any precision and scale information defined in the bean before. Anyway, I reverted the changes to the expected output and changed the way that the balance property was being initialized. As these tests are related to the security aspects, I guess they can be considered sufficient for the sake of this refactoring.

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/service/AccountService.java
taefi: Done.

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/service/UserInfoService.java
taefi: Done.

Artur- (approved):

Issue #13851, opened 2022-05-10
Memory leak on adding and removing components


nSharifA:
### Description of the bug

The following snippet results a memory leak:
```
@Push
@Route("test")
@Viewport("width=device-width, initial-scale=1")
public class Test extends Div {

  FlexLayout testLayout;

  public Test() {
    testLayout = new FlexLayout();
    add(new Button("Click me", e -> {
      testLayout.removeAll();
      IntStream.range(0, 10000).forEach(i -> testLayout.add(new Button("B")));
    }));
  }
}
```
If you click on the button it will remove the testLayout content and add 10 000 new Buttons to the testLayout. 

### Expected behavior

After clicking multiple times on the button Java VisualVM should show 10 000 + 1 button instance in the memory after pressing the perform GC button.
Instead it will show (n * 10000) + 1 buttons (where n == click count).

### Minimal reproducible example

see the code.

### Versions

- Vaadin / Flow version: 14.8.4
- Java version: 1.8.0_271
- OS version: Win10
- Browser version (if applicable): Chrome 101.---
- Application Server (if applicable): Jetty
- IDE (if applicable): Eclipse

PR #14478, opened 2022-09-05
fix: Be more sensible when adding changes to list of pending changes

TatuLund:
Fixes: https://github.com/vaadin/flow/issues/13851


MarcinVaadin (approved):
Changes reviewed and tested, memory usage stays on same level.

mshabarov (approved):
Looks good to me.
This patch re-indexes changes once any removal change is added, this gives a memory consumption improvement: removed items are eligible for garbage collection early, not only after sending a response to client.
Additionally, the logic and codes in `addChange` are simpler, than were before in `collectChanges` in my opinion.

I made a quick performance test using the example project from the issue's author (modified it a bit to include removals):

```java
        add(new Button("Click me", e -> {
            IntStream.range(0, 10000).forEach(i -> {
                testLayout.add(new Button("B"));

                if (i > 5000 && random.nextBoolean()) {
                    int componentCount = testLayout.getComponentCount();
                    testLayout.remove(testLayout.getComponentAt(
                            random.nextInt(componentCount - 1)));
                }

            });
        }));
```

New algorithm has more traverses/loops and these traverses happen on each removal, which makes the overall algorithm complexity higher. However, I believe that it shouldn't affect the performance, until one has a millions of elements in the change list - not sure it's a case for real applications. 
The test gives me approximately the same response processing time as before the patch.

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/NodeList.java
```diff
@@ -277,59 +275,88 @@ protected List<AbstractListChange<T>> getChangeTracker() {
...
-                        : "AbstractListChange has only three subtypes: add, remove and clear";
-            }
-            index++;
-        }
-
-        List<AbstractListChange<T>> changes;
-
-        changes = allChanges.stream().filter(this::acceptChange)
-                .collect(Collectors.toList());
+        List<AbstractListChange<T>> changes = getChangeTracker().stream()
```
knoobie, line 358: If you worried about performance of the added loops, you can refactor this stream usage to a loop to improve performance :) 

See e.g. https://medium.com/levi-niners-crafts/java-performance-improvement-java-8-streams-vs-loops-and-lists-vs-arrays-e824136832d6 and hundred other threads about loops > sequential streams performance vise

mshabarov: Thanks for the tip @knoobie . I made a quick test stream + filter vs loop and indeed it gives me around 3-4 times faster results for loops while testing with N ~1000 of items.

mshabarov (approved):

ZheSun88:
2.8 CP -> https://github.com/vaadin/flow/pull/14586

PR #14487, opened 2022-09-09
feat: Allow setting content of the Html component

TatuLund:
When using Html component as replacement for Vaadin 8's Label /w html content allowed, it is quite common use case to alter the content programmatically.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Html.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Html.java
@@ -147,7 +147,7 @@ public class Html extends Component {
 
         doc.outputSettings().prettyPrint(false);
         setInnerHtml(root.html());
-    }    
+    }
 
     private void setOuterHtml(String outerHtml) {
         Document doc = Jsoup.parseBodyFragment(outerHtml);
--- a/flow-server/src/main/java/com/vaadin/flow/component/Html.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Html.java
@@ -123,7 +123,7 @@ public class Html extends Component {
      * Any heading or trailing whitespace is removed while parsing but any
      * whitespace inside the root tag is preserved.
      *
-     * @param outerHtml
+     * @param html
      *            the HTML to wrap
      */
     public void setHtmlContent(String html) {
--- a/flow-server/src/main/java/com/vaadin/flow/component/Html.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Html.java
@@ -113,6 +113,42 @@ public class Html extends Component {
         setOuterHtml(outerHtml);
     }
 
+    /**
+     * Sets the content based on the given HTML fragment. The fragment must have
+     * exactly one root element.
+     * <p>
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/component/Html.java
```diff
@@ -113,6 +113,42 @@ public Html(String outerHtml) {
...
+     * whitespace inside the root tag is preserved.
+     *
+     * @param html
+     *            the HTML to wrap
+     */
+    public void setHtmlContent(String html) {
+        Document doc = Jsoup.parseBodyFragment(html);
+        int nrChildren = doc.body().children().size();
+        if (nrChildren != 1) {
+            String message = "HTML must contain exactly one top level element (ignoring text nodes). Found "
```
Legioth, line 133: There's a whole bunch of mostly duplicated code between this method and the internal `setOuterHtml` method. Could that somehow be eliminated by having one method delegate to the other or by extracting a helper method?

flow-server/src/main/java/com/vaadin/flow/component/Html.java
```diff
@@ -113,6 +113,42 @@ public Html(String outerHtml) {
...
+            if (nrChildren > 1) {
+                String tagNames = doc.body().children().stream()
+                        .map(org.jsoup.nodes.Element::tagName)
+                        .collect(Collectors.joining(", "));
+                message += " elements with the tag names " + tagNames;
+            }
+            throw new IllegalArgumentException(message);
+        }
+
+        org.jsoup.nodes.Element root = doc.body().child(0);
```
Legioth, line 144: I suspect the client-side implementation doesn't support changing the root element's tag name from what it originally was. I would like to either see a test that shows that the tag name can actually change, or alternatively a check here to throw an `IllegalStateException` in case it would be changing.

Legioth (commented):
(Cursory review. Someone else should still look at the change in-depth)

flow-server/src/main/java/com/vaadin/flow/component/Html.java
TatuLund: True, I chose to throw.

flow-server/src/main/java/com/vaadin/flow/component/Html.java
TatuLund: With the changes to below the redundancy is less obvious.

Legioth: First 14 lines still seem to be identical based on a quick look

TatuLund: Better now?

MarcinVaadin: After 10fa6741ade8cca70b2bbec313566b889762e9b4 there are no more duplicated code fragments. @Legioth would you like to add something?

Legioth: LGTM

PR #14496, opened 2022-09-12
fix: Hide elements with 'hidden' attribute via CSS

### LINKED ISSUE #14453: setVisible(false) does not work in production mode
### Description of the bug

I have created a component `Composite<Div>`. (here is the code: https://github.com/jcgueriaud1/googlemap-v23-issue/blob/visibility/src/main/java/com/example/application/components/XPanel.java)
`setVisible(false)` set correctly the attribute hidden on the div.
It's not visible in development mode.
In production mode, it's visible.
I can reproduce it when I'm using a subset of Vaadin components so I suppose the css rule is added by one of the component I'm not using. Apparently it's coming from this component: https://github.com/PolymerElements/iron-flex-layout/blob/a8b8e91b261ca835aa8b457cc571d55ac4973fdc/iron-flex-layout.html#L62-L72

### Expected behavior

A component not visible should be hidden.

https://github.com/jcgueriaud1/googlemap-v23-issue/tree/visibility
Start the application and go http://localhost:9080/visibility
In development mode, 
image

In production mode:
image

### Versions

- Vaadin / Flow version: 23.2.0.rc1
- Java version: 11 and 17
- OS version: Mac
- Browser version (if applicable): Chrome (probably not relevant)

mshabarov:
## Description

Fixes #14453

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -151,6 +151,9 @@ public class UI extends Component
         getNode().getFeature(ElementData.class).setTag("body");
         Component.setElement(this, Element.get(getNode()));
         pushConfiguration = new PushConfigurationImpl(this);
+        // Component::setVisible relies on hidden attribute.
+        // Adds a global display:none style to elements with hidden attribute
+        page.addStyleSheet("./frontend/styles/hidden.css");
     }
 
     /**
--- a/flow-server/src/main/resources/META-INF/resources/frontend/styles/hidden.css
+++ b/flow-server/src/main/resources/META-INF/resources/frontend/styles/hidden.css
@@ -1,0 +1,3 @@
+[hidden] {
+    display: none !important;
+}


flow-server/src/main/resources/META-INF/resources/frontend/styles/hidden.css
```diff
@@ -0,0 +1,3 @@
+[hidden] {
+    display: none !important;
+}
```
caalador: missing new line

caalador (commented):
If this works in dev mode, but not in production mode is the issue actually that something is missing when doing the FrontendDependencies scanning vs FullDependencies scanning?
Is there something that should be added to the theme so that this would work?

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -151,6 +151,9 @@ protected UI(UIInternalUpdater internalsHandler) {
         getNode().getFeature(ElementData.class).setTag("body");
         Component.setElement(this, Element.get(getNode()));
         pushConfiguration = new PushConfigurationImpl(this);
+        // Component::setVisible relies on hidden attribute.
+        // Adds a global display:none style to elements with hidden attribute
+        page.addStyleSheet("./frontend/styles/hidden.css");
```
Artur-, line 156: Doesn't this cause a separate request for this file? It should be part of the index page or bundle

Artur-:
> If this works in dev mode, but not in production mode is the issue actually that something is missing when doing the FrontendDependencies scanning vs FullDependencies scanning? Is there something that should be added to the theme so that this would work?

It works in dev mode because this CSS rule is present in some Polymer components and in dev mode everything is included in the bundle, so you accidentally get this global rule in use

Artur-:
Here for instance https://github.com/PolymerElements/iron-flex-layout/blob/24279a26a9ce42e4be741d2ba6124b7e221ae90c/iron-flex-layout.js#L431-L433

flow-server/src/main/java/com/vaadin/flow/component/UI.java
mshabarov: Yes, it does, you're right. I rework it to be a part of index page.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1179,6 +1179,10 @@ protected void setupCss(Element head, BootstrapContext context) {
             // Basic reconnect and system error dialog styles just to make them
             // visible and outside of normal flow
             setupErrorDialogs(styles);
+
+            styles = head.appendElement("style").attr("type",
```
Artur-, line 1183: Is there a reason we need another style element instead of adding it to the existing one, i.e. the one called "setupErrorDialogs"?

mshabarov: No special reason, thanks!

Artur- (approved):

Issue #14479, opened 2022-09-06
PiT 23.2: CDI and Quarkus starters need an HTTP hit for start compiling the frontend


manolo:
### Description of the bug

Quarkus and CDI starters got stuck at some point before compiling the frontend.
By sending an HTTP request, the process continues compiling the frontend.
It din't happen with webpack or in other starters like spring flow

### Expected behavior

Frontend compilation in dev mode does not need any http trigger for compiling the frontend in dev mode

#### base-starter-flow-quarkus

- First double check that you dont have any browser with an open tab pointing to `http://localhost:8080` 
- second download and run the application
```
git clone https://github.com/vaadin/base-starter-flow-quarkus.git
cd base-starter-flow-quarkus
mvn
```
- check that It gets stuck (for ever) at this point:

```
2022-09-06 14:57:15,043 INFO  [com.vaa.flo.ser.fro.TaskUpdatePackages] (ForkJoinPool.commonPool-worker-11) using '/usr/local/bin/npm --no-update-notifier --no-audit --scripts-prepend-node-path=true --ignore-scripts install' for frontend package installation
2022-09-06 14:57:21,391 INFO  [com.vaa.flo.ser.fro.TaskUpdatePackages] (ForkJoinPool.commonPool-worker-11) Frontend dependencies resolved successfully.
2022-09-06 14:57:22,323 INFO  [com.vaa.flo.ser.fro.TaskCopyFrontendFiles] (ForkJoinPool.commonPool-worker-11) Copying frontend resources from jar files ...
2022-09-06 14:57:22,353 INFO  [com.vaa.flo.ser.fro.TaskCopyFrontendFiles] (ForkJoinPool.commonPool-worker-11) Visited 20 resources. Took 29 ms.
```
- finally visit http://localhost:8080, and it will continue frontend compilation

#### skeleton-starter-flow-cdi

- Close all tabs in your browser
- download and run the starter

```
git clone https://github.com/vaadin/skeleton-starter-flow-cdi.git 
cd skeleton-starter-flow-cdi
mvn
``` 
- It is stuck for ever
```
15:00:53,498 INFO  [org.wildfly.extension.undertow] (ServerService Thread Pool -- 8) WFLYUT0021: Registered web context: '/' for server 'default-server'
[...]

PR #14501, opened 2022-09-12
fix: set load-on-startup for automatically registered Vaadin servlet

mcollovati:
## Description

When using Vite, DevModeInitializer blocks dev-server startup until a VaadinServlet is deployed because it needs to get the servlet path to use. If the container lazily loads servlets, Vite will not start until the first HTTP request for the Vaadin servlet is received.
This change sets load-on-startup feature for automatically deployed Vaadin servlet, to ensure that the servlet and Vite are loaded on the startup of the Web application

Part of #14479

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/startup/ServletDeployer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/startup/ServletDeployer.java
@@ -336,6 +336,7 @@ public class ServletDeployer implements ServletContextListener {
 
         registration.setAsyncSupported(true);
         registration.addMapping(path);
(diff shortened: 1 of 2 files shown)


MarcinVaadin (approved):

MarcinVaadin:
Shouldn't be a feature?

mcollovati:
> Shouldn't be a feature instead of bugfix?

Good question. I would say it is a bug fix because if we need a servlet to be initialized to start Vite, we should provide an eager initialization out of the box.

PR #14510, opened 2022-09-13
fix: Do not overwrite generated files with the same content

### LINKED ISSUE #14490: Extra Vite recompiles because generated files are always rewritten
### Description of the bug

If you make a change to the classpath for a running application, e.g. `touch target/classes/application.properties` then you will see that Vite recompiles several times during the Spring Boot restart cycle. This is because the files in `frontend/generated` are written again although there are no changes. Vite cannot know they have not changed, so it will recompile.

### Expected behavior

The files are only written if they have changed, so Vite does not have to do anything

vaadin init hello
cd hello
mvn
touch classes/application.properties

### Versions

Artur-:
Overwriting the files will cause the frontend tooling to recompile when there is no need to do that

Fixes #14490

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractTaskClientGenerator.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractTaskClientGenerator.java
@@ -17,6 +17,7 @@ package com.vaadin.flow.server.frontend;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
@@ -66,10 +67,7 @@ public abstract class AbstractTaskClientGenerator implements FallibleCommand {
         File generatedFile = getGeneratedFile();
         try {
             String fileContent = getFileContent();
-            log().debug("writing file '{}'", generatedFile);
-
-            FileUtils.forceMkdirParent(generatedFile);
-            FileUtils.writeStringToFile(generatedFile, fileContent, UTF_8);
+            writeIfChanged(generatedFile, fileContent);
         } catch (IOException exception) {
             String errorMessage = String.format("Error writing '%s'",
                     generatedFile);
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractTaskClientGenerator.java
```diff
@@ -66,18 +67,37 @@ public void execute() throws ExecutionFailedException {
...
+        FileUtils.writeStringToFile(file, content, UTF_8);
+    }
+
+    private static String getExistingFileContent(File generatedFile)
+            throws IOException {
+        if (!generatedFile.exists()) {
+            return null;
+        }
+        return FileUtils.readFileToString(generatedFile,
+                StandardCharsets.UTF_8);
```
mcollovati, line 101: nit: we can maybe use a static import for `UTF_8`, as we do in `writeIfChanged`
```suggestion
        return FileUtils.readFileToString(generatedFile, UTF_8);
```

mcollovati (commented):
May we add some tests for `writeIfChanged`?

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractTaskClientGenerator.java
```diff
@@ -66,18 +67,37 @@ public void execute() throws ExecutionFailedException {
...
         }
     }
 
-    Logger log() {
-        return LoggerFactory.getLogger(getClass());
+    static Logger log() {
+        return LoggerFactory.getLogger(AbstractTaskClientGenerator.class);
+    }
+
+    static void writeIfChanged(File file, String content) throws IOException {
```
mshabarov, line 82: We need at least a unit test for this.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractTaskClientGenerator.java
Artur-: Added

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractTaskClientGenerator.java
Artur-: Done

flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractTaskClientGeneratorTest.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.nio.file.attribute.FileTime;
+
+import org.apache.commons.io.FileUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.nimbusds.jose.util.StandardCharset;
```
mshabarov, line 13: Should be `java.nio.charset.StandardCharsets` ?

Artur-: Fixed

mshabarov (dismissed):

mshabarov (approved):

PR #14516, opened 2022-09-13
fix: Remove presence check on item refresh

### LINKED ISSUE #14170: [Grid] lazy loading refreshItem will pass PageSize = Integer.MAX_VALUE
### Description of the bug

in grid using lazy data loading, when we call method `GridLazyDataView.refreshItem(data)`. it will call the callback method `fetchCallback` of `public V setItems(     com.vaadin.flow.data.provider.CallbackDataProvider.FetchCallback<T, F> fetchCallback )` with q.getPageSize = Integer.MAX_VALUE -> will load huge data.

This causes program very slow. see below code for detailed bug.

image

### Expected behavior

getPageSize = configurable. Default = 50.
Also ensure when call `refreshItem` will not change current position of the edited item in the grid.

N/A

### Versions

- Vaadin / Flow version: 23
- Java version: 17
- OS version: N/A
- Browser version (if applicable):N/A
- Application Server (if applicable):N/A
- IDE (if applicable):N/A

MarcinVaadin:
## Description

Moved item presence check to dataProvider to prevent fetching all items

Fixes #14170 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
@@ -21,7 +21,6 @@ import java.util.stream.Stream;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentEventListener;
 import com.vaadin.flow.component.ComponentUtil;
-import com.vaadin.flow.function.SerializablePredicate;
 import com.vaadin.flow.function.SerializableSupplier;
 import com.vaadin.flow.shared.Registration;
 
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java
```diff
@@ -123,12 +124,11 @@ public Stream<T> getItems() {
...
     public void refreshItem(T item) {
         Objects.requireNonNull(item, NULL_ITEM_ERROR_MESSAGE);
-        //@formatter:off
-        getItems()
-                .filter(i -> equals(item, i))
-                .findFirst()
-                .ifPresent(i -> dataProviderSupplier.get().refreshItem(item));
-        //@formatter:on
+        SerializablePredicate<T> filter = i -> equals(item, i);
+        if (dataProviderSupplier.get().fetch(new Query(filter)).findAny()
```
mshabarov, line 128: `SerializablePredicate` is used only for in memory data as far as I remember. So for callback data providers it wouldn't be used and the query in fact would fetch from 0 to int max.

taefi (approved):

PR #14523, opened 2022-09-14
fix: Fix navigation to secured view with oauth2 without router-ignore

### LINKED ISSUE #14253: Cannot navigate to secured View when not logged for OAuth2
### Description of the bug

When using RouterLink and Anchor to a secured @Route with @PermitAll an error page is displayed when not logged in.

Could not navigate to '<page>'
Available routes:

### Expected behavior

A login page should be show. Azure call it user flow: https://docs.microsoft.com/en-us/azure/active-directory-b2c/user-flow-overview.

Adding router-ignore makes it work. I have added router-ignore to the /logout link which is not a Vaadin route.

It is the same behavior for  com.azure.spring:spring-cloud-azure-starter-active-directory-b2c:4.3.0 and org.springframework.boot:spring-boot-starter-oauth2-client:2.6.7. For azure b2c the config is:
```
@EnableWebSecurity
public class SecurityConfig extends VaadinWebSecurityConfigurerAdapter {

    private final AadB2cOidcLoginConfigurer configurer;

    public SecurityConfig(AadB2cOidcLoginConfigurer configurer) {
        this.configurer = configurer;
    }

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        super.configure(http);
        http.apply(configurer);
    }
}
```

```
@Route("vaadin-hello")
@PermitAll()
public class VaadinHelloView extends VerticalLayout {
    public VaadinHelloView() {
        add(new H1("Hello from VAADIN"));
```

### Versions

Vaadin: 23.2.0.alpha3
Flow: 23.2.0.alpha2
Java: JetBrains s.r.o. 17.0.2
OS: amd64 Windows 10 10.0
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36
Live reload: Java active (HotswapAgent): Front end active
IntelliJ
Tomcat

### LINKED ISSUE #14520: ViewAccessChecker causes the NotFoundException to flash before redirecting to the login view
### Description of the bug

When an application uses a non-Flow view as the login view the "not found" error view is briefly shown before the user is redirected to the actual login view. The reason is the following logic in the view access checker
```
// Prevent the view from being created

if (loginUrl != null) {
}
```

Rerouting takes place because the target view instance should not be created nor rendered when the user does not have access. Then the `beforeEnterEvent.getUI().getPage().setLocation(loginUrl);` redirect is only executed after the error view has been rendered, causing it to be briefly shown.

### Expected behavior

The error view should not at all be visible during redirection to the login view.

A potential solution would be #11029 and implementing a `forwardToUrl` in the before enter event so the code could be like
```
if (loginUrl != null) {
    beforeEnterEvent.forwardToUrl(loginUrl); // This must also prevent the view from being created
} else {
}
```

See https://github.com/vaadin/flow/issues/14253#issuecomment-1212577437 for an oauth app where this is clearly visible

### Versions

taefi:
## Description
This is to fix the problem of navigating to a secured view
from a public view before login in while using the OAuth2 external
login page. Prior to this, it was needed to set "router-ignore" 
attribute on the links from public pages to skip vaadin-router and 
let the spring-security forward the user to the external login page. 

Fixes #14253

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSavedRequestAwareAuthenticationSuccessHandler.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSavedRequestAwareAuthenticationSuccessHandler.java
@@ -24,6 +24,7 @@ import javax.servlet.http.HttpSession;
 
 import com.vaadin.flow.server.auth.ViewAccessChecker;
 
+import org.springframework.core.log.LogMessage;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.web.DefaultRedirectStrategy;
 import org.springframework.security.web.authentication.SavedRequestAwareAuthenticationSuccessHandler;
@@ -31,6 +32,7 @@ import org.springframework.security.web.csrf.CsrfToken;
 import org.springframework.security.web.savedrequest.HttpSessionRequestCache;
 import org.springframework.security.web.savedrequest.RequestCache;
 import org.springframework.security.web.savedrequest.SavedRequest;
+import org.springframework.util.StringUtils;
 
 /**
  * A version of {@link SavedRequestAwareAuthenticationSuccessHandler} that
@@ -89,16 +91,9 @@ public class VaadinSavedRequestAwareAuthenticationSuccessHandler
         @Override
         public void sendRedirect(HttpServletRequest request,
                 HttpServletResponse response, String url) throws IOException {
-            String redirectUrl;
-            String savedRedirectUrl = response.getHeader(SAVED_URL_HEADER);
-            if (savedRedirectUrl != null) {
-                redirectUrl = savedRedirectUrl;
-            } else {
-                redirectUrl = url;
-            }
 
             if (!isTypescriptLogin(request)) {
-                super.sendRedirect(request, response, redirectUrl);
+                super.sendRedirect(request, response, url);
                 return;
             }
 
@@ -126,6 +121,7 @@ public class VaadinSavedRequestAwareAuthenticationSuccessHandler
      */
     public VaadinSavedRequestAwareAuthenticationSuccessHandler() {
         setRedirectStrategy(new RedirectStrategy());
+        setTargetUrlParameter(SAVED_URL_HEADER);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


mshabarov:
@Artur- do you have a memories in which use case the code execution here (`VaadinSavedRequestAwareAuthenticationSuccessHandler` class) goes into the second code block, i.e. when it takes the the URL from saved request? For the first case, it takes the target URL from session, and it is being put into session by view access checker - it's clear. Does the ordering of those blocks matter and why? 
```java
        String storedServerNavigation = getStoredServerNavigation(request);
        if (storedServerNavigation != null) {
            // The saved server navigation URL is relative to the context path
            if (!"".equals(request.getContextPath())) {
                storedServerNavigation = "/" + storedServerNavigation;
            }
            response.setHeader(SAVED_URL_HEADER, storedServerNavigation);
        } else if (savedRequest != null) {
            /*
             * This is here instead of in sendRedirect as we do not want to
             * fallback to the default URL but instead send that separately.
             */
            response.setHeader(SAVED_URL_HEADER, savedRequest.getRedirectUrl());
        }
```

Artur-:
> do you have a memories in which use case the code execution here (VaadinSavedRequestAwareAuthenticationSuccessHandler class) goes into the second code block, i.e. when it takes the the URL from saved request? For the first case, it takes the target URL from session, and it is being put into session by view access checker - it's clear. Does the ordering of those blocks matter and why?

Probably if you are not using `ViewAccessChecker` so e.g. with Hilla or some custom solution.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSavedRequestAwareAuthenticationSuccessHandler.java
```diff
@@ -126,33 +121,66 @@ public void sendRedirect(HttpServletRequest request,
...
+                response.setHeader(SAVED_URL_HEADER, targetUrl);
+                this.getRedirectStrategy().sendRedirect(request, response,
+                        targetUrl);
+                return;
+            } else {
+                this.requestCache.removeRequest(request, response);
+            }
+        } else if (fullySavedRequestUrl != null) {
+            response.setHeader(SAVED_URL_HEADER, fullySavedRequestUrl);
+        }
```
mshabarov, line 156: This code block needs explanatory comments as discussed offline.

taefi: Done.

mcollovati (approved):

mshabarov (approved):

PR #14529, opened 2022-09-14
chore: remove not needed dependency since it comes from flow-server

manolo:
Having this dependency here brings an old and vulnerable version of `commons-codec`

### DIFF
--- a/flow-plugins/flow-maven-plugin/pom.xml
+++ b/flow-plugins/flow-maven-plugin/pom.xml
@@ -66,12 +66,6 @@
             <version>1.21</version>
         </dependency>
         <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>httpclient</artifactId>
-            <version>4.5.13</version>
-        </dependency>
-
-        <dependency>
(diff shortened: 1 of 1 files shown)


knoobie:
Removing the httpclient from flow-server sounds more sexy :) 

The only class that needs it in flow-server [DefaultFileDownloader](https://github.com/vaadin/flow/blob/bf2b84fed9071d399248b3fedfc0c8922c2282c3/flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/DefaultFileDownloader.java) which is only used on build time.

I haven't dug through all sources, but I think it should be reasonable to move DefaultFileDownloader or e.g. caller NodeInstaller to the maven and gradle base plugin to declutter flow.

manolo:
@knoobie sounds reasonable both: removing definitively the httpclient dependency by using java utility classes, and moving the downloader to the plugins.
This should be performed separately to this PR, do you mind to open a ticket with the request?

knoobie:
Created https://github.com/vaadin/flow/issues/14538 for it.

mshabarov (approved):

PR #14560, opened 2022-09-19
chore: Read default dependencies from a package.json

### LINKED ISSUE #13640: Define default dependencies using a package.json file
### Describe your motivation

The dependencies currently defined in [NodeUpdater.java](https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java#L366) and the development dependencies defined in https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java#L384 need manually updating and are often forgotten. For example https://github.com/vaadin/flow/pull/13638 looks like it updates Lit but in practice it only changes one test package lock file and lit remains the old version

The dependencies should instead be defined in a `package.json` file in e.g. 
- https://github.com/vaadin/flow/tree/master/flow-server/src/main/resources/com/vaadin/flow/server/dependencies/webpack - for webpack
- https://github.com/vaadin/flow/tree/master/flow-server/src/main/resources/com/vaadin/flow/server/dependencies/vite - for Vite

By using a package.json file it is easy to see what dependencies are but also it allows tools like dependabot or snyk to upgrade the versions.

Artur-:
This should allow tools to create automated PRs for dependency updates

Fixes #13640

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -44,6 +44,7 @@ import com.vaadin.flow.server.frontend.scanner.FrontendDependencies;
 import com.vaadin.flow.server.frontend.scanner.FrontendDependenciesScanner;
 
 import elemental.json.Json;
+import elemental.json.JsonArray;
 import elemental.json.JsonException;
 import elemental.json.JsonObject;
 import elemental.json.JsonValue;
@@ -98,8 +99,6 @@ public abstract class NodeUpdater implements FallibleCommand {
     private static final String DEP_MAIN_VALUE = "index";
     private static final String DEP_VERSION_KEY = "version";
     private static final String DEP_VERSION_DEFAULT = "1.0.0";
-    private static final String ROUTER_VERSION = "1.7.4";
-    protected static final String POLYMER_VERSION = "3.5.1";
 
     static final String VAADIN_VERSION = "vaadinVersion";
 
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -387,21 +386,29 @@ private static <T extends JsonValue> T computeIfAbsent(
...
     Map<String, String> getDefaultDependencies() {
-        Map<String, String> defaults = new HashMap<>();
-
-        defaults.put("@vaadin/router", ROUTER_VERSION);
-
-        defaults.put("@polymer/polymer", POLYMER_VERSION);
-
-        defaults.put("lit", "2.3.0");
+        try {
+            Map<String, String> map = new HashMap<>();
```
knoobie, line 390: Should this be LinkedHashMap to make sure it stays the same order for testability?

Artur-: Hm I am pretty sure that there is sorting of all package.json dependencies taking place somewhere to deal with this also for `@NpmPackage` from addons

knoobie: You are right, there is e.g. this PR to fix the behavior https://github.com/vaadin/flow/pull/14556 - was just wondering if you wanna add a test that all provided dependencies from the files are loaded with e.g assertThat(map).containsKey(..) so that nothin get accidentally removed. Having proper order on this low level would have been helpful for that test tho

Artur-: Well the idea to test is good :) I added one. Intentionally without version numbers so it won't be broken everytime there is a version update

knoobie: 👍 (no thumbup on mobile 🤦🏼‍♂️)

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -387,70 +385,46 @@ private static <T extends JsonValue> T computeIfAbsent(
...
+            JsonObject dependencies = readPackageJson(id)
+                    .getObject(packageJsonKey);
+            for (String key : dependencies.keys()) {
+                map.put(key, dependencies.getString(key));
+            }
 
-        defaults.put("lit", "2.3.0");
+            return map;
+        } catch (IOException e) {
+            log().error("Unable to read default dependencies", e);
```
mcollovati, line 403: Adding `id` value to the error message may simplify troubleshooting, in case of exception

Artur-: Added

mcollovati (approved):

PR #14571, opened 2022-09-19
feat: Allow rerouting to an external URL

### LINKED ISSUE #11029: BeforeEnterEvent.forwardTo(String) should forward to the given URL
With this before enter event handler
```
@Route(value = "hello")
public class HelloWorldView extends HorizontalLayout implements BeforeEnterObserver {

    @Override
        event.forwardTo("/hello/world/");
    }
```
and a servlet deployed using context path `/context`

You would expect that opening `/context/hello` would redirect to `/hello/world/`. Instead it redirects to `/context/hello/world`.

The  URL is wrong (relative to `/context`) and the trailing slash is missing.

```
git clone https://github.com/Artur-/test-forward-to-url
cd test-forward-to-url
open http://localhost:8080/context/hello
```

### Expected behavior
http://localhost:8080/hello/world/ is shown

### Actual behavior
http://localhost:8080/context/hello/world is shown

### Versions:
    - Vaadin / Flow version: 19
    - Java version: 15

Artur-:
Fixes #11029

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/BeforeEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/BeforeEvent.java
@@ -53,8 +53,10 @@ public abstract class BeforeEvent extends EventObject {
     private NavigationState rerouteTargetState;
     private ErrorParameter<?> errorParameter;
 
-    private String forwardToUrl = null;
-    private String rerouteToUrl = null;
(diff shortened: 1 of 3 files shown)


Artur-:
Not sure what exactly would need to be tested here as the logic of the navigator is not really straight forward

mcollovati (approved):

PR #14576, opened 2022-09-19
fix: Redirect to login page without showing the 404 page

### LINKED ISSUE #14520: ViewAccessChecker causes the NotFoundException to flash before redirecting to the login view

Artur-:
Fixes #14520

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
@@ -188,11 +188,10 @@ public class ViewAccessChecker implements BeforeEnterListener {
             if (loginView != null) {
                 beforeEnterEvent.forwardTo(loginView);
             } else {
-                // Prevent the view from being created
-                beforeEnterEvent.rerouteToError(NotFoundException.class);
-
                 if (loginUrl != null) {
-                    beforeEnterEvent.getUI().getPage().setLocation(loginUrl);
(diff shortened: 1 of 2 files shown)


Artur-:
Based on #14571

Artur-:
Why does it reload infinitely? Sounds like access is denied to the login page?

mcollovati:
I have the same issue also with 23.2.1, so it seems not related to the change in this PR

mcollovati:
As @Artur- pointed out, reload was because of my configuration mistake on test application.
Redirect works as expected and "not found" page is not shown

mcollovati (approved):

PR #14578, opened 2022-09-20
test: Make test more stable

Artur-:


flow-tests/test-root-ui-context/src/test/java/com/vaadin/flow/uitest/ui/push/AbstractClientServerCounterIT.java
```diff
@@ -45,9 +47,13 @@ protected WebElement getServerCounterStartButton() {
...
     public static int getServerCounter(TestBenchTestCase t) {
-        WebElement serverCounterElem = t
-                .findElement(By.id(ClientServerCounterUI.SERVER_COUNTER_ID));
-        return Integer.parseInt(serverCounterElem.getText());
+        try {
+            WebElement serverCounterElem = t.findElement(
+                    By.id(ClientServerCounterUI.SERVER_COUNTER_ID));
+            return Integer.parseInt(serverCounterElem.getText());
+        } catch (StaleElementReferenceException | NoSuchElementException e) {
+            return getServerCounter(t);
```
mshabarov, line 55: Does it make sense to limit the number of retries to some value, say 10 tries and then throw?

Artur-: We can do that. Changed to a for loop

mshabarov (approved):

PR #14597, opened 2022-09-21
fix: update css-loader(#12863)(CP:2.8)

ZheSun88:
this is a Cherry pick of https://github.com/vaadin/flow/pull/12863
Update css-loader to get a newer version of postcss


mshabarov:
I have no memories what did we fix with this update. This is a major version update, may bring breaking changes to 14.8 and 14.9 users.

ZheSun88:
if it is not needed, feel free to close this one.

mshabarov:
@caalador do you recall what we fixed by updating this version? Can we skip it for 14.8/14.9 (since it's a major update)

caalador:
As it says  to `get a newer version of postcss` I would guess it is to get a vulnerability patch in postcss

mshabarov (approved):

PR #14614, opened 2022-09-22
fix: Redirect to the correct server view also when coming from a non-root URL

### LINKED ISSUE #14528: VaadinSavedRequestAwareAuthenticationSuccessHandler redirects to the wrong URL if login is not in the application root
### Description of the bug

If you have a public view at /hello, a private view at /foo and a login view at /my/login where the public view has a link to the private view, then when you click the link you will end up on the login form and after login you will end up on /my/foo instead of /foo

The reason is that `ViewAccessChecker` stores the target view it gets from `beforeEnterEvent.getLocation().getPathWithQueryParameters()`. This is relative path so in this case `foo`.

The login POST goes to /my/login and the response to that is a redirect to "foo", which the browser then interprets as /my/foo

If you instead open /foo directly, then it is not `ViewAccessChecker` that saves the target URL but instead `VaadinDefaultRequestCache` using `HttpSessionRequestCache`. This in turn stores the full request and provides the redirect URL as a full URL, e.g. http://localhost:8080/foo

### Expected behavior

You should end up on the page you tried to navigate to.

As described above

### Versions

- Vaadin / Flow version: 23.2.1

Artur-:
Fixes #14528

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
@@ -43,6 +43,8 @@ public class ViewAccessChecker implements BeforeEnterListener {
 
     public static final String SESSION_STORED_REDIRECT = ViewAccessChecker.class
             .getName() + ".redirect";
+    public static final String SESSION_STORED_REDIRECT_ABSOLUTE = ViewAccessChecker.class
+            .getName() + ".redirectAbsolute";
     private final AccessAnnotationChecker accessAnnotationChecker;
     private Class<? extends Component> loginView;
     private String loginUrl;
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
```diff
@@ -173,8 +175,15 @@ public void beforeEnter(BeforeEnterEvent beforeEnterEvent) {
...
             if (session != null) {
-                session.setAttribute(SESSION_STORED_REDIRECT, beforeEnterEvent
-                        .getLocation().getPathWithQueryParameters());
+                VaadinServletRequest servletRequest = (VaadinServletRequest) request;
+                String servletHostAndPath = servletRequest.getRequestURL()
+                        .toString();
+                String viewPathAndParameters = beforeEnterEvent.getLocation()
+                        .getPathWithQueryParameters();
+                session.setAttribute(SESSION_STORED_REDIRECT,
+                        viewPathAndParameters);
```
mcollovati, line 184: Do we still need to store this information?
It seems that now only `SESSION_STORED_REDIRECT_ABSOLUTE`, at least in Flow code base.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSavedRequestAwareAuthenticationSuccessHandler.java
```diff
@@ -171,8 +167,8 @@ private static String getStoredServerNavigation(
         if (session == null) {
             return null;
         }
-        return (String) session
-                .getAttribute(ViewAccessChecker.SESSION_STORED_REDIRECT);
+        return (String) session.getAttribute(
+                ViewAccessChecker.SESSION_STORED_REDIRECT_ABSOLUTE);
```
mcollovati, line 171: Does it make sense to also remove the attribute from the session here? Or should it be preserved for other uses?

flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java
Artur-: I left it there only to stay compatible if people have implemented a redirect for something else than Spring Security

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSavedRequestAwareAuthenticationSuccessHandler.java
Artur-: I guess it should be removed to avoid a potential wrong redirect later on

mshabarov (approved):

PR #14638, opened 2022-09-26
fix: Only map websockets to servlet path and not any sub paths

Artur-:
Allows other websockets to work, especially when Vaadin is root mapped


HawkSK:
Is this fix going to be released also in `+0.0.1` (the label was removed)? Because it should be, the problem did not happen in 23.1.1, so this is a fix for 23.2.x. Also, it's a blocker in using 23.2.0.

MarcinVaadin (approved):

flow-tests/vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/CustomWebSocket.java
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            @Override
+            public boolean supportsPartialMessages() {
+                return false;
+            }
+        }, WEBSOCKET_URL);
+    }
+
+}
```
mcollovati, line 54: newline missing
```suggestion
}

```

flow-tests/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/CustomWebSocketIT.java
```diff
@@ -0,0 +1,71 @@
...
+            session.close();
+        }
+
+        @OnClose
+        public void onClose(CloseReason reason) {
+            closureLatch.countDown();
+        }
+
+        public String waitForMessage() throws InterruptedException {
+            closureLatch.await();
```
mcollovati, line 41: May we use `await` with timeout to avoid the test wait indefinitely if something goes wrong?

mcollovati (approved):

MarcinVaadin:
PR could be merged but please have in mind that is not the final solution - see https://github.com/vaadin/flow/issues/14641

PR #14618, opened 2022-09-23
test: Test Spring Boot behind a reverse proxy

Artur-:

PR #14656, opened 2022-09-27
test: Test Spring Security behind a reverse proxy

Artur-:
The test is not enabled by default as it does not start/stop a reverse proxy

It can be run if you configure nginx like

```
    server {
        listen       1234;

location /public/path/ {
    proxy_pass http://localhost:8888/;
}
```

or if you configure an apache 2 proxy like

```
Listen 1234

<Location /public/path/>
    ProxyPass "http://localhost:8888/"
    ProxyPassReverse /
</Location>
```


Artur-:
Depends on #14618

mcollovati:
LGTM. Let's wait for 14618 to be merged

mcollovati (approved):

PR #14672, opened 2022-09-28
chore: update component version to 23.3.0-alpha2

caalador:
Update component version used for testing.
Update all package.json component versions
to the tested component version.


mcollovati (commented):
We should perhaps also update `VaadinSmokeTest.kt`

https://github.com/vaadin/flow/blob/672d29ce7ebbddc5e22574bd99677707103145ad/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/VaadinSmokeTest.kt#L322

caalador:
Updated gradle test version

mcollovati (approved):

PR #14681, opened 2022-09-29
fix: make system property feature flag value always override stored one

### LINKED ISSUE #14679: System property for Feature Flag does not override value when loading features file with ResourceProvider
In #14408 we introduced the possibility to override feature flags value by setting a system property.
This works fine unless a `ResourceProvider` is used to load the feature flags file; in this case the system property is ignored.

mcollovati:
## Description

Enable or disable a feature flag by providing a system property is not working when feature flags file is loaded by a ResourceProvider. This change makes system properties always take precedence over stored values

Fixes #14679

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
+++ b/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
@@ -200,7 +200,8 @@ public class FeatureFlags implements Serializable {
         if (featureFlagFile == null || !featureFlagFile.exists()) {
             // Disable all features if no file exists
             for (Feature f : features) {
-                f.setEnabled(false);
+                f.setEnabled(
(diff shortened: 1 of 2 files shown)


caalador (commented):
Only thing is could the tests use a static mock of system instead of changing the actual system property?

https://www.baeldung.com/mockito-mock-static-methods#mocking-a-static-method-with-arguments

mcollovati:
Good point. Will try

mcollovati:
Seems like `System` cannot be mocked
```
MockitoException: It is not possible to mock static methods of java.lang.System to avoid interfering with class loading what leads to infinite loops
```

caalador (approved):

Issue #14602, opened 2022-09-21
Atmosphere endpoint is intercepting all Web Socket requests


MarcinVaadin:
### Description of the bug

After #13960 Atmosphere endpoint is mapped to `/` which causes that all Web Socket requests are handled by Atmosphere handlers. It became impossible to register own Web Socket handlers.

Description and reproducible example is present on https://stackoverflow.com/questions/73791155/upgrading-to-vaadinwebsecurity-is-breaking-custom-websocket/73798818#73798818

Workaround - exclusion of Atmosphere:
```
<dependency>
    <groupId>com.vaadin</groupId>
    <!-- Replace artifactId with vaadin-core to use only free components -->
    <artifactId>vaadin</artifactId>
    <exclusions>
        <exclusion>
            <groupId>com.vaadin.external.atmosphere</groupId>
            <artifactId>atmosphere-runtime</artifactId>
        </exclusion>
    </exclusions>
</dependency>
```

### Expected behavior

It should be possible to register custom Web Socket handlers.

### Minimal reproducible example

Available on https://stackoverflow.com/questions/73791155/upgrading-to-vaadinwebsecurity-is-breaking-custom-websocket/73798818#73798818

### Versions

- Vaadin / Flow version: 23.2
- Java version: 17

ggecy:
+1 
This has very high priority for our project since we cannot deploy vaadin 23.2.0 to our production systems because of it. We have vaadin push enabled and using it extensively (so the workaround is not applicable) while at the same time we are using spring websockets which now don't work.

Artur-:
I think the summary is slightly wrong as the websocket endpoint was SUPPOSED to be mapped to `/` only but as the path mapping length of Atmosphere is set to 5 by default it is also mapped to urls like `/{path}` and `/{path}/{path}`. Making it mapped to only the servlet path should fix this, drafted in #14638

Tonne-TM:
@Artur-  The issue still exists. If you update the example provided at StackOverflow to Vaadin 23.2.3 the custom WebSocket is still overwritten by the Atmosphere framework returning

```
for(;;);[{"meta":{"async":true,"sessionExpired":true}}]
``

**Comment by Artur-** [comment]
If I take the app, change to 23.2.3 and run `new WebSocket("ws://localhost:8080/websocket")` in the browser I see 
```
connected
```
in the server log

PR #14694, opened 2022-09-29
feat: Bring back vaadin servlet mapping behind feature flag

### LINKED ISSUE #14641: Websocket in combination with AJP broken behind reverse-proxy in 23.2.
### Description of the bug

Previously websocket connections could be forwarded by using a separate location in the httpd configuration like so `<Location /web/vaadinServlet/>` allowing this connection to be used with http / websockets - after 23.2 the websockets connection is not routed tho `/web/vaadinServlet/` anymore, instead it goes straigth to the `/web/?v-r=push...` where the connection can't be established because `/web/**` forwards to AJP, which DOES not support websockets.

Example exception:

```log

'HTTP upgrade is not supported by this protocol' Requested resource was: '/web/?v-r=push&v-uiId=1&v-pushId=0ce8ff89-eae1-4612-9a12-3383167a2d5b&X-Atmosphere-tracking-id=0&X-Atmosphere-Framework=3.1.2-javascript&X-Atmosphere-Transport=websocket&X-Atmosphere-TrackMessageSize=true&Content-Type=application/json;%20charset=UTF-8&X-atmo-protocol=true'

```

### Expected behavior

Websockets should be handled by a separat endpoint `/web/vaadinServlet/` like done since V8.

.

### Versions

- Vaadin / Flow version: 23.2
- Java version: 11
- OS version: xx
- IDE (if applicable):

MarcinVaadin:
## Description

As there is a need to support custom servlet mapping after it has been removed (#14602) - it has been bring back behind feature flag.

Fixes #14641

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
```diff
@@ -79,6 +79,11 @@ public class FeatureFlags implements Serializable {
             "Enforce client / constraint / binder validation",
             "enforceFieldValidation",
             "https://github.com/vaadin/platform/issues/3066", false, null);
+
+    public static final Feature SERVLET_MAPPING = new Feature(
```
Artur-, line 83:
```suggestion
    public static final Feature VAADINSERVLET_PUSH_MAPPING = new Feature(
```

flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
```diff
@@ -79,6 +79,11 @@ public class FeatureFlags implements Serializable {
...
+            "Use non-root servlet mapping (/vaadinServlet)", "vaadinServlet",
```
Artur-, line 84:
```suggestion
            "Use /vaadinServlet mapping for push (deprecated)", "vaadinServletForPush",
```

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurerAdapter.java
```diff
@@ -186,6 +188,18 @@ public static RequestMatcher getDefaultHttpSecurityPermitMatcher(
                 .map(path -> RequestUtil.applyUrlMapping(urlMapping, path))
                 .forEach(paths::add);
 
+        if (FeatureFlags.get(VaadinService.getCurrent().getContext())
```
Artur-, line 191: Is there really a current `VaadinService` available at this point?

Artur- (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurerAdapter.java
MarcinVaadin: Yes, here it is present, there are few issues still that I’m working on:
- Lookup is not available during SpringBootAutoconfiguration so I cannot use FeatureFlags.get( )
- VaadinWebSecurity requires access to Feature flag in static methods
I’m testing solution that for Spring feature is defined within Application Properties - they are available almost from the startup of spring-boot (edited)

flow-tests/vaadin-spring-tests/test-spring-boot-push-mapping/src/main/resources/application.properties
```diff
@@ -0,0 +1,4 @@
+server.port=8888
+vaadin.exclude-urls=/swagger-ui/**
+vaadin.pushServletMapping=true
```
mshabarov, line 3: I'd propose "nonRootPushServletMapping" (or vice versa - "pushRootServletMapping", true by default) as a name for this parameter, because "pushServletMapping" doesn't tell much about its meaning. Also, not "pushServletMapping=false" doesn't make sense, since mapping is always there.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
```diff
@@ -84,7 +84,7 @@ public ServletRegistrationBean<SpringServlet> servletRegistrationBean(
         String pushRegistrationPath;
 
         if (rootMapping) {
-            mapping = VaadinServletConfiguration.VAADIN_SERVLET_MAPPING;
+            mapping = VaadinServletConfiguration.VAADIN_SERVLET_PUSH_MAPPING;
```
mshabarov, line 87: `mapping` is used here not for push endpoint, but for SpringServlet registration, which is fine.
Apart from Push, it is fine to have `/vaadinServlet` mapped spring servlet.
We only talk about mapping path for `JSR356_MAPPING_PATH`, it should be changed, not the spring servlet.
Or I terribly missed something.

mshabarov: So in my opinion you should change `pushRegistrationPath` to include `/VAADIN/push` in case of root mapping, but not `mapping` itself.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
```diff
@@ -176,4 +197,34 @@ private String upperCaseToDashSeparated(String value) {
...
+     * Spring Boot.
+     *
+     * @param environment
+     *            the application environment
+     * @return the url mapping or null if none is defined
+     */
+    protected boolean isPushServletMappingEnabled(Environment environment) {
+        if (SpringUtil.isSpringBoot()) {
+            try {
+                return (boolean) Class.forName(
```
mshabarov, line 211: Why we have to do this? Isn't `VaadinConfigurationProperties` instance available in the context where the method `isPushServletMappingEnabled` is needed?

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinMVCWebAppInitializer.java
```diff
@@ -66,11 +67,19 @@ public void onStartup(ServletContext servletContext)
...
                     .addServlet("dispatcher", new DispatcherServlet(context));
             dispatcherRegistration.addMapping("/*");
-            mapping = VaadinServletConfiguration.VAADIN_SERVLET_MAPPING;
+
+            if (pushServletMappingEnabled) {
+                initParameters.put(InitParameters.SERVLET_PARAMETER_PUSH_URL,
+                        makeContextRelative(mapping.replace("*", "")));
+            }
+
+            mapping = VaadinServletConfiguration.VAADIN_SERVLET_PUSH_MAPPING;
```
mshabarov, line 82: `mapping` should be "/vaadinServlet", because it corresponds to the servlet, not to the push registration.
`pushRegistrationPath` should be modified.

mshabarov (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
```diff
@@ -99,9 +99,14 @@ public ServletRegistrationBean<SpringServlet> servletRegistrationBean(
...
          * and websockets will fail.
          */
-        initParameters.put(ApplicationConfig.JSR356_MAPPING_PATH,
-                pushRegistrationPath);
-        initParameters.put(ApplicationConfig.JSR356_PATH_MAPPING_LENGTH, "0");
+
+        if (configurationProperties.isPushServletMapping()) {
+            initParameters.put(ApplicationConfig.JSR356_MAPPING_PATH,
+                    mapping.replace("/*", ""));
+        } else {
```
mcollovati, line 106: Did I understand correctly that if I've set a custom vaadin url mapping (e.g. `/ui/*`), the mapping for PUSH will be the on `/ui/*` and not on `/ui/VAADIN/push`?
If so, the AJP reverse proxy configuration will still have the same problem of missing a dedicate path for websockets

mshabarov: Yes, it would be `/ui/*` and it was so in the past event with `/vaadinServlet`. So seems like if someone uses AJP with non root url mapping, this fix wouldn't work.

mshabarov: What if we add `VAADIN/push` always when this new parameter is turned on?

mcollovati: I agree, it seems more consistent to me

MarcinVaadin:
Closing a PR due to approach change. New PR will be created.

PR #14715, opened 2022-10-03
test: fix BootstrapHandlerTest on Windows

MarcinVaadin:
Added Windows check assumption to make TeamCity branches green again.


caalador (changes requested):
The tests should work on windows, but the check is wrong in that it looks for `\n` when it should look for `\r\n`

So the actual fix should be to have contains string be `.contains(String.format("...%n"));` so that the new line gets the correct format

caalador (approved):

PR #14749, opened 2022-10-06
fix: ignore nonexistent nodes during resync

### LINKED ISSUE #14232: Resynchronizing UI by client's request
### Description of the bug

I am often getting the message in the log of Resynchronizing UI by client's request.

The full message is:
Resynchronizing UI by client's request. The network message was lost before reaching the client and the client is reloading the full UI state. This typically happens because of a bad network connection with packet loss or because of some part of the network infrastructure (load balancer, proxy) terminating a push (websocket or long-polling) connection. If you are using push with a proxy, make sure the push timeout is set to be smaller than the proxy connection timeout

It's not connection, apparently it's session related.

I have on the same linux server running another Vaadin 8 application.
They have different names, different ports, in different folders and are NGINX-mapped with different subdomains.

I have already evaluated the other problems related to the topic:
#12640
There was no clear solution, someone raised some possibilities:
- Browser, it's not in my case, it wouldn't happen so often.
- HTTP proxy, I use NGINX for the subdomains, but I use other services like NODE, static HTML and never had a problem. NGINX was configured by default, without any additional configuration, it just throws the subdomain to port X.
- Possibility of mixing sessions: it makes no sense, since I have very little load, it happens even with only 1 user logged in.

#12173
In this problem the user uses long duration push. Not my case, I use simple, default @Push.

#11645
In this problem as I understand it was the slow connection. It's not my case, everything is flying here.

#12173
In this problem the user uses long duration push. Not my case, I use simple, default @Push.

#10096
This is a very similar scenario. But there was no conclusion, the user closed without informing how the issue was resolved and if it was resolved.

#9399
This one he solved by changing the server, it is difficult to assess what the problem was

Anyway, this problem is quite recurrent and should be better explained in the documentation. I downloaded the example available from the site, nothing out of the ordinary, little or no extra configuration.

### Expected behavior

The expectation is that it doesn't lock the user's screen, it's terrible to have to ask him to refresh the page, because after it's broken it doesn't come back.

It's hard to simulate, because it doesn't always happen.
The impression is that it happens after a while without changes on the page, but sometimes it happens right after logging in or during some slower operation.

### Versions

- Vaadin / Flow version: 23.2.0.alpha1
- Java version: 
openjdk version "11.0.3" 2019-04-16
OpenJDK Runtime Environment (build 11.0.3+7-Ubuntu-1ubuntu218.10.1)
OpenJDK 64-Bit Server VM (build 11.0.3+7-Ubuntu-1ubuntu218.10.1, mixed mode, sharing)

- OS version:
-Ubuntu

Chrome

### LINKED ISSUE #14470: Resynchronization can fail if it happens more than once and there is action between
### Description of the bug

Resynchronization can fail if it happens more than once and there is action between like clicking other buttons, showing notification. Showing the Notification could be necessary in order to replicate the issue. This may make the pattern feel very random.

### Expected behavior

Resync should allways recover

Test view

```
@Route("resync")
public class ResyncListenerView extends Div {

    public ResyncListenerView() {
        NativeButton resynch = new NativeButton("Resynchronize",
                event -> UI.getCurrent().getInternals().incrementServerId());
        resynch.setId("resync");

        NativeButton button = new NativeButton("Click me");
        button.addClickListener(event -> {
            Span works = new Span("Works");
            Notification.show("Works");
            works.setId("works");
            add(works);
        });
        button.setId("button");

        add(resynch, button);
    }
}
```

Test procedure
```
@TestPath("resync")
public class ResyncListenerIT extends AbstractComponentIT {

    @Test
    public void listenerWorksAfterResync() {
        open();
        $(TestBenchElement.class).id("resync").click();
        sleep(8);
        $(TestBenchElement.class).id("button").click();
        int count = $(SpanElement.class).all().size();
        Assert.assertEquals(2, count);
        $(TestBenchElement.class).id("resync").click();
        sleep(8);
        $(TestBenchElement.class).id("button").click();
        count = $(SpanElement.class).all().size();
        Assert.assertEquals(3, count);
    }

    private void sleep(int seconds) {
        try {
            Thread.sleep(1000 * seconds);
        } catch (InterruptedException e) {
        }
    }
}
```

### Versions

- Vaadin / Flow version: Vaadin 23.1.7

caalador:
Ignore changes for nodes that are not
available in the tree when a resync is underway.

This will get the page to a working
condition after a resync as all changes for
existing nodes are executed.

Fixes #14232
Fixes #14470

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/flow/StateTree.java
+++ b/flow-client/src/main/java/com/vaadin/client/flow/StateTree.java
@@ -158,6 +158,7 @@ public class StateTree {
 
     /**
      * Check if tree is resynchronizing after a {@link #prepareForResync}
+     *
      * @return true if resync called
      */
     public boolean isResync() {
@@ -166,7 +167,9 @@ public class StateTree {
 
     /**
      * Set the resynchronization state for the StateTree.
-     * @param resync resynchronization state to set
+     *
+     * @param resync
+     *            resynchronization state to set
      */
     public void setResync(boolean resync) {
         this.resync = resync;
--- a/flow-client/src/main/java/com/vaadin/client/flow/StateTree.java
+++ b/flow-client/src/main/java/com/vaadin/client/flow/StateTree.java
@@ -50,6 +50,7 @@ public class StateTree {
     private JsMap<Integer, String> nodeFeatureDebugName;
 
     private boolean updateInProgress;
+    private boolean resync;
 
     /**
      * Creates a new instance connected to the given registry.
@@ -152,6 +153,23 @@ public class StateTree {
                 node.setParent(null);
             }
         });
+        setResync(true);
+    }
+
+    /**
+     * Check if tree is resynchronizing after a {@link #prepareForResync}
+     * @return true if resync called
+     */
+    public boolean isResync() {
+        return resync;
+    }
+
+    /**
+     * Set the resynchronization state for the StateTree.
+     * @param resync resynchronization state to set
+     */
+    public void setResync(boolean resync) {
+        this.resync = resync;
     }
 
     /**
@@ -482,5 +500,4 @@ public class StateTree {
             return "Unknown node feature: " + id;
         }
     }
-
 }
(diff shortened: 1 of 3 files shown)


caalador:
Another issue is that the client is compiled with `<enableAssertions>true</enableAssertions>` which means that any assert used in the java code becomes an `if() throw` is javascript

mcollovati (approved):

PR #14750, opened 2022-10-06
PoC for /VAADIN/push set by default (to be removed)

MarcinVaadin:
Will be removed after test, related to #14694


MarcinVaadin:
Not needed anymore

PR #14752, opened 2022-10-07
fix: Watch flow-frontend for changes

### LINKED ISSUE #14695: Vite doesn't detect changes in files in target/flow-frontend
### Description of the bug

We have a maven multimodule project where js/ts files from different maven modules end up in `target/flow-frontend` directory. Once the server is started for the first time (in development mode), vite creates the `node_modules/.vite` folder and it works. However any time I make a change in other modules and the updated file gets to `target/flow-frontend`, vite doesn't detect the change until I manually delete the `node_modules/.vite` directory and restart the server.
We even have a custom file watcher implemented which copies the modified files from `src/main/resources/META-INF/resources/frontend` from other modules to target/flow-frontend even when the server is running - this caused webpack to reload the page and the changes were effective even without server restart. However this doesn't work with vite, not even with server restart.

### Versions:
    - Vaadin / Flow version: 23.2.2
    - Java version: 17
    - OS version: Mac OS 12.6
    - Browser version (if applicable): Chrome 105.0.5195.125
    - Application Server (if applicable): Spring boot 2.7.3 with embedded tomcat
    - IDE (if applicable): Intellij IDEA

Artur-:
Especially in multi module projects, the folder is updated with new contents quite often. Also if you change an addon version, the changes need to be picked up.

Fixes #14695

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/PopStateHandler.java
+++ b/flow-client/src/main/java/com/vaadin/client/PopStateHandler.java
@@ -17,6 +17,7 @@ package com.vaadin.client;
 
 import java.util.Objects;
 
+import com.google.gwt.core.client.Scheduler;
 import com.vaadin.client.flow.RouterLinkHandler;
 
 import elemental.client.Browser;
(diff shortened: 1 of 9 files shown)


Artur-:
This fails because of https://github.com/vitejs/vite/issues/10373

PR #14799, opened 2022-10-11
chore: Set version for nexus stage plugin in the root pom

mshabarov:


pom.xml
```diff
@@ -654,6 +654,12 @@
 					<artifactId>bridger</artifactId>
```
caalador, line 654: Why is the bridger plugin formatted all which ever way. As this is a chore, mind fixing this format also at the same time.

mshabarov: Done

caalador (approved):

PR #14815, opened 2022-10-13
chore: Update version to 24.0

mshabarov:


flow-plugins/flow-gradle-plugin/build.gradle
```diff
@@ -132,7 +132,8 @@ task functionalTest(type: Test) {
         showStandardStreams = true
     }
 }
-check.dependsOn functionalTest
```
Artur-, line 135: Why do tests need to be disabled when changing the version number?

pom.xml
```diff
@@ -67,14 +67,14 @@
             https://maven.vaadin.com/vaadin-addons
         </flow.addons.repo.url>
 
-        <maven.compiler.release>11</maven.compiler.release>
```
Artur-, line 70: Can't this be a separate PR for more clarity

Artur- (changes requested):

flow-plugins/flow-gradle-plugin/build.gradle
mshabarov: Because PR validation won't pass otherwise.

pom.xml
mshabarov: ok, I'll make a separate PR.

flow-plugins/flow-gradle-plugin/build.gradle
mshabarov: Making Gradle plugin support JDK 17 is a separate work, that we start soon.

Artur- (approved):

PR #14827, opened 2022-10-13
fix: use access and not lock

caalador:
Use access so that we do not
block with the lock

Fixes vaadin/multiplatform-runtime#101


tepi (approved):

Artur-:
This causes https://github.com/vaadin/sso-kit/issues/50 because `VaadinSession.getSession()` does not return `null` even though the HTTP session has been invalidated

PR #14834, opened 2022-10-14
chore(deps-dev): bump jaxb-impl from 4.0.0 to 4.0.1 (#14600)

Artur-:
Reapplies the old commit which was reverted as part of the Servlet 5 upgrade

Bumps jaxb-impl from 4.0.0 to 4.0.1.

---
updated-dependencies:
- dependency-name: com.sun.xml.bind:jaxb-impl dependency-type: direct:development update-type: version-update:semver-patch ...


mshabarov (approved):

Artur-:
Was reverted in the earlier PR because it caused parse errors in validation.xml in the flow-data tests. Apparently it was not the real cause

PR #14840, opened 2022-10-14
fix: old node installation

### LINKED ISSUE #14839: npm taken from global installation instead of .vaadin installed
### Description of the bug

```
[INFO] The globally installed Node.js version 14.19.0 is older than the required minimum version 16.14.0. Using Node.js from /root/.vaadin.
[DEBUG] Using node 16.17.1 located at /root/.vaadin/node/node
[DEBUG] Using npm 6.14.16 located at /usr/bin/npm
```

### Expected behavior

As node is used as vaadin installed, then also npm is used from there and not the global one.

globall install old node, run application with 23.2-SNAPSHOT+

### Versions

- Vaadin / Flow version: 23.2+
- OS: Unix

caalador:
Fix so that if we have a usable old
node installation we check the correc
path for npm script.

Fixes #14839


mshabarov (approved):
The fix looks good. Please add a unit test, if it takes reasonable efforts.

PR #14842, opened 2022-10-14
feat: Throw a meaningful exception if deployed into a Servlet 5 container

Artur-:


flow-server/pom.xml
```diff
@@ -21,6 +21,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.vaadin.servletdetector</groupId>
+            <artifactId>throw-if-servlet5</artifactId>
```
mshabarov, line 26: What is this dependency, is it a part of Flow?

Artur-: https://github.com/vaadin/servlet-detector 

It is intentionally not a part of Flow as it should never change and it requires servlet 5 API to compile

mshabarov (approved):

flow-server/pom.xml
mshabarov: Just noticed: shouldn't it be excluded from flow-client, as it done in another PR?

Artur-: It could be but does not have to. The flow-client tests run Jetty 9 which is servlet 3 so the servlet 3 check fails the tests but having a servlet 5 check there is not a problem

PR #14853, opened 2022-10-14
fix: Make Jwt security work with new Spring Security

Artur-:
Please ensure this actually does the same thing. I cannot tell but Hilla tests pass...

With Spring Security 6 the response in `loadContext(HttpRequestResponseHolder requestResponseHolder)` is null so the old code does not work at all

Related to https://github.com/vaadin/hilla/pull/599

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/JwtSecurityContextRepository.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/JwtSecurityContextRepository.java
@@ -20,7 +20,6 @@ import jakarta.servlet.http.HttpServletResponse;
 import java.util.Date;
 import java.util.List;
 import java.util.Objects;
-import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import com.nimbusds.jose.JOSEException;
@@ -52,7 +51,6 @@ import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.authentication.JwtGrantedAuthoritiesConverter;
 import org.springframework.security.web.context.HttpRequestResponseHolder;
-import org.springframework.security.web.context.SaveContextOnUpdateOrErrorResponseWrapper;
 import org.springframework.security.web.context.SecurityContextRepository;
 
 /**
@@ -181,25 +179,20 @@ class JwtSecurityContextRepository implements SecurityContextRepository {
     }
 
     @Override
-    public SecurityContext loadContext(HttpRequestResponseHolder holder) {
-        return loadContext(holder.getRequest()).get();
-    }
+    public SecurityContext loadContext(
+            HttpRequestResponseHolder requestResponseHolder) {
+        SecurityContext context = SecurityContextHolder.createEmptyContext();
 
-    @Override
-    public Supplier<SecurityContext> loadContext(HttpServletRequest request) {
-        return () -> {
-            SecurityContext context = SecurityContextHolder
-                    .createEmptyContext();
+        HttpServletRequest request = requestResponseHolder.getRequest();
 
-            Jwt jwt = decodeJwt(request);
-            if (jwt != null) {
-                Authentication authentication = jwtAuthenticationConverter
-                        .convert(jwt);
-                context.setAuthentication(authentication);
-            }
+        Jwt jwt = decodeJwt(request);
+        if (jwt != null) {
+            Authentication authentication = jwtAuthenticationConverter
+                    .convert(jwt);
+            context.setAuthentication(authentication);
+        }
 
-            return context;
-        };
+        return context;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/JwtSecurityContextRepository.java
```diff
@@ -180,22 +181,25 @@ private Jwt decodeJwt(HttpServletRequest request) {
...
+    }
 
-        Jwt jwt = decodeJwt(request);
-        if (jwt != null) {
-            Authentication authentication = jwtAuthenticationConverter
-                    .convert(jwt);
-            context.setAuthentication(authentication);
-        }
+    @Override
+    public Supplier<SecurityContext> loadContext(HttpServletRequest request) {
```
vursen, line 189: - `loadContext(HttpServletRequest request)`
- `loadContext(HttpRequestResponseHolder holder)`

Note, both have been deprecated since 5.8. Using `loadDeferredContext` is recommended instead.

However, if we don't plan to move to `loadDeferredContext` right now, it probably doesn't make much sense to include any method structure changes. From my testing, the only thing that matters in this fix is actually this part:

```java
requestResponseHolder.setResponse(new UpdateJwtResponseWrapper(request,
                requestResponseHolder.getResponse()));
```

When I remove it, the tests start passing. 

It was introduced in https://github.com/vaadin/spring/pull/900 by @platosha but I'm not sure I understand its original purpose since there is not any comment about that. Based on my manual testing, the basic scenario works correctly: log in / open some private views / log out.

vursen: Based on [this](https://github.com/spring-projects/spring-security/blob/55877c5f328acde963b8fd0037c831c277852ee4/web/src/main/java/org/springframework/security/web/context/SecurityContextRepository.java#L57-L61), `SaveContextOnUpdateOrErrorResponseWrapper` can be used to guarantee that the security context is saved in the case of redirects or HTTP errors. It is also deprecated now, although it is not clear what is a replacement for it.

vursen: Digging deeper into the Spring Security's source code, I believe I've come to understand the way everything works.

There are two approaches to maintaining the context and one of those is claimed to be deprecated:

### [SecurityContextPersistenceFilter](https://github.com/spring-projects/spring-security/blob/9cb668aec2ad14f91c122c66b7d7d4a8b6e133f7/web/src/main/java/org/springframework/security/web/context/SecurityContextPersistenceFilter) (deprecated)

This is an old approach where `loadContext()` is called on every request and `saveContext()` is called after all the consequent filters are executed in this [`finally` block](https://github.com/spring-projects/spring-security/blob/9cb668aec2ad14f91c122c66b7d7d4a8b6e133f7/web/src/main/java/org/springframework/security/web/context/SecurityContextPersistenceFilter.java#L119-L126).

However, as far as I understand, there was a chance that the context won't be saved in the case of an error or redirect (sounds like `finally` could not be executed in some cases). If you wanted to get the context saved in those cases too, it was recommended to wrap the response into a class extending `SaveContextOnUpdateOrErrorResponseWrapper` to guarantee that the context is saved on the response commit.

### [SecurityContextHolderFilter](https://github.com/spring-projects/spring-security/blob/9cb668aec2ad14f91c122c66b7d7d4a8b6e133f7/web/src/main/java/org/springframework/security/web/context/SecurityContextHolderFilter.java) (enabled by default since 6.0)

This is a new approach where the context is loaded lazily with `loadDeferredContext()` and `saveContext()` is called explicitly in all places it is necessary (see [AuthenticationFilter](https://github.com/spring-projects/spring-security/blob/9cb668aec2ad14f91c122c66b7d7d4a8b6e133f7/web/src/main/java/org/springframework/security/web/authentication/AuthenticationFilter.java#L207) for example).

Wrapping the response into `SaveContextOnUpdateOrErrorResponseWrapper` is no longer necessary because of explicit `saveContext()`.

### Conclusion

`UpdateJwtResponseWrapper` can be safely dropped.

vursen: Well, It is currently not possible to use `loadDeferredContext()` because `DeferredSecurityContext` and `SupplierDeferredSecurityContext` – which are required to specify the method's type – are not available publicly for some reason.

vursen (approved):

vursen (approved):

Artur-:
The latest commit changes from using the non-deprecated `loadDeferredContext` to the deprecated `loadContext`. Why is that preferable?

vursen:
> The latest commit changes from using the non-deprecated loadDeferredContext to the deprecated loadContext. Why is that preferable?

The main reason is that Spring Security 6.0.0-M6 doesn't contain `loadDeferredContext`.

Artur-:
Ok so the default implementation of `loadDeferredContext` uses the deprecated `loadContext`... Let's keep it like this for now then and fix when needed

Making `SupplierDeferredSecurityContext` `final` and package private looks quite deliberate to avoid creating more public API so we probably need to implement our own variant in another PR at some point

PR #14911, opened 2022-10-25
chore: Upgrade to Spring Boot 3 RC1

Artur-:


vursen:
I wonder if it should be something other than a chore taking into account that the PR drops `VaadinWebSecurityConfigurerAdapter.java`.

Artur-:
Updated to slf4j 2.0.3 which Spring Boot 3 RC1 uses

Artur-:
> I wonder if it should be something other than a chore taking into account that this PR drops `VaadinWebSecurityConfigurerAdapter.java`.

I can extract that into a separate PR that would be `chore!` and which would make sense in release notes

Artur-:
This probably cannot work before https://github.com/spring-projects/spring-authorization-server/milestone/19 is released

knoobie:
> [chore(deps): bump slf4j.version from 1.7.36 to 2.0.3](https://github.com/vaadin/flow/pull/14911/commits/3517c94715c049ddba94dbe52e8539c33a8dec91)

Should this be done seperatly as well as? See discussion here https://github.com/vaadin/flow/pull/14409

Artur-:
Well Spring Boot 3 RC1 depends on slf4j 2.0.3 so at least this project does not work when mixing the versions

knoobie:
Yeah :) I'm not against the updated, would just recommend to do it first as `chore!` so that people know about this once their update to V24

Artur-:
Yep, probably does not work with Spring Boot M5 though, that uses slf4j 1.7...

Maybe we can make the "Upgrade to Spring Boot 3" PR with `!`and list the relevant info there

knoobie:
It should work 🙂 it was already in master some weeks ago and build successfully on M4 or so and was reverted by @mshabarov after my comment

Artur-:
Let's see then

Artur-:
#14916 indeed looks pretty 🍏

Artur-:
Depends on #14916 , otherwise should be ok

mshabarov (approved):

PR #14902, opened 2022-10-24
feat: Move new license checker feature behind config parameter

mshabarov:
## Description

Related-to https://github.com/vaadin/flow/issues/14633

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #14921, opened 2022-10-26
test: Add tests for new license checker config parameter

mshabarov:
Related-to https://github.com/vaadin/flow/pull/14902

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
@@ -1,5 +1,6 @@
 package com.vaadin.flow.server.startup;
 
+import javax.servlet.ServletContext;
 import javax.servlet.ServletException;
 
 import java.io.File;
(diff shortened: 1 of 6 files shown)


flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
```diff
@@ -453,6 +458,46 @@ public void onStartup_fallbackBaseDirIsGradleProjectDirectory_isAccepted()
...
+        DeploymentConfiguration deploymentConfiguration =
+                Mockito.mock(DeploymentConfiguration.class);
+        Mockito.when(deploymentConfiguration.enableDevServer())
+                .thenReturn(true);
+
+        try (MockedStatic<LicenseChecker> licenseChecker = Mockito.mockStatic(LicenseChecker.class)) {
+            try {
+                DevModeInitializer.initDevModeHandler(new HashSet<>(), servletContext, deploymentConfiguration);
+            } catch (NullPointerException e) {
+                // expected exception within dev server init
```
Artur-, line 473: If dev server init fails with NPE, does this test actually check anything?

flow-server/src/test/java/com/vaadin/flow/server/startup/DevModeInitializerTest.java
```diff
@@ -453,6 +458,46 @@ public void onStartup_fallbackBaseDirIsGradleProjectDirectory_isAccepted()
...
+        try (MockedStatic<LicenseChecker> licenseChecker = Mockito.mockStatic(LicenseChecker.class)) {
+            Mockito.when(deploymentConfiguration.isOldLicenseCheckerEnabled())
+                    .thenReturn(true);
+            try {
+                DevModeInitializer.initDevModeHandler(new HashSet<>(), servletContext, deploymentConfiguration);
+            } catch (NullPointerException e) {
+                // expected exception within dev server init
+            }
+
+            licenseChecker.verify(() -> LicenseChecker.setStrictOffline(true), times(0));
```
Artur-, line 497: If dev mode initializer fails with NPE, does this actually check anything?

mshabarov: Here I check that the `DevModeInitializer` skips `setStrictOffline` call, because old license checker is enabled https://github.com/vaadin/flow/blob/3e129cfaa72e4caefb1cab02816d0c2bbf020e64/flow-server/src/main/java/com/vaadin/flow/server/startup/DevModeInitializer.java#L261. It throws the NPE later on for another reason (something is not mocked). So it should check what it is supposed to.

PR #14924, opened 2022-10-26
fix: clean commented css before handling

### LINKED ISSUE #14770: Commenting @import in styles.css gives net::ERR_ABORTED 404
### Description of the bug

image

(File exists on disk)

image

### Expected behavior

Parser should ignore everything in comments

`/*@import "my-style.css";*/`

### Versions

Hilla: 23.2.3
Flow: 23.2.3
Java: JetBrains s.r.o. 17.0.2
OS: amd64 Windows 10 10.0
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/106.0.0.0 Safari/537.36
Live reload: Java active (HotswapAgent): Front end active

caalador:
Remove any comment blocks in CSS
before creating link references.

Fixes #14770

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/default/package.json
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/default/package.json
@@ -20,6 +20,7 @@
     "glob": "7.2.3",
     "typescript": "4.7.4",
     "workbox-core": "6.5.0",
-    "workbox-precaching": "6.5.0"
+    "workbox-precaching": "6.5.0",
+    "strip-comments": "2.0.1"
   }
 }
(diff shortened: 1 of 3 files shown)


Artur- (commented):
Would it make sense to use e.g. https://www.npmjs.com/package/strip-comments instead?

caalador:
Does it note that http://... the // is not a comment and to only handle block comments?
Need to test.

Artur-:
"Strip line and/or block comments from a string. Blazing fast, and works with JavaScript, Sass, CSS, Less.js, and a number of other languages."

"block {Boolean}: if false strip only line comments, default true"

caalador:
Testing say seems to work.
Drops away upkeep from us, even though it is 90 lines more than our own with strange regex handling.

Artur- (dismissed):

caalador:
Seems it removes something extra. I'll try to investigate what and from where exactly.

caalador:
Couldn't get strip-comments to not remove line comment for `url(http://` even with the setting to ignore line comments.
Changed to strip-css-comments which is more css specific and doesn't handle css as javascript

Artur- (approved):

PR #14942, opened 2022-10-28
test: Use the correct Spring Boot version for gradle tests

Artur-:


Artur-:
This should not be merged but #14937 should be backported selectively instead

PR #14978, opened 2022-11-02
fix: use full scan for ts task

### LINKED ISSUE #14733: Frontend compilation fails if TS files are only in the fallback bundle
### Description of the bug

In Vaadin 14, tsconfig.json is only generated if the project contains `.ts` files. However, if the TS files end up in the fallback bundle, for instance because of #14732, then no tsconfig.json is generated. The build then fails with errors like
```
ERROR in [tsl] ERROR
      TS18002: The 'files' list in config file 'tsconfig.json' is empty.
 @ ../target/frontend/generated-flow-imports-fallback.js 21:0-46
 @ ../target/frontend/generated-flow-imports.js?babel-target=es5
```

I can imagine this would happen also if you add an addon with TS files and you do not use the addon but I have not tested that.

### Expected behavior

Compilation works when TS files are included only in the fallback bundle

https://github.com/samie/specialbuttons/commit/571832b121b86653f87c02740a0d3e7e6134eede#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8

### Versions

- Vaadin / Flow version: 14.8.18

caalador:
Always use the allDependencies scan
for checking for creation of ts files.
This means that for a production build
we check the fallback dependencies for
ts files also.

Closes #14733

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -453,8 +453,15 @@ public class NodeTasks implements FallibleCommand {
                         frontendDependencies.getThemeDefinition());
             }
 
-            commands.add(new TaskGenerateTsFiles(builder.npmFolder,
-                    frontendDependencies.getModules()));
+            final FrontendDependenciesScanner fallbackScanner = getFallbackScanner(
+                    builder, classFinder);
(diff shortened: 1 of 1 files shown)


caalador:
Closing PR as issue is fixed in 2.8 and 2.7 is now deprecated as the new Vaadin 14 version was released.

PR #14979, opened 2022-11-02
fix!: Extract addon files to frontend/generated/jarResources instead of a fake npm package

### LINKED ISSUE #14695: Vite doesn't detect changes in files in target/flow-frontend

Artur-:
No longer creates target/flow-frontend so any tool or script relying on that needs to be updated

Fixes #14695


caalador:
Add to `NodeUpdater:getPackageJson` the same update fix as `removeWebpackPlugins` that removes `"@vaadin/flow-frontend": "./target/flow-frontend",` from the `package.json` if there. Else running the `clean-frontend` plugin goal would fix it.

Artur-:
> Add to `NodeUpdater:getPackageJson` the same update fix as `removeWebpackPlugins` that removes `"@vaadin/flow-frontend": "./target/flow-frontend",` from the `package.json` if there. Else running the `clean-frontend` plugin goal would fix it.

Is that the correct place for cleanup? There is cleanup for `@vaadin/flow-deps` in `TaskUpdatePackages.java`

caalador:
`removeLegacyProperties` sounds like a good place to have it.

Artur-:
Cosmetic question, would `frontend/generated/jar-resources` actually look better than `jarResources`?

caalador:
I would choose `jar-resources` as camelcase on folders feel a bit off for some reason.

Gradle build probably fails on the `VaadinSmokeTest` as there is the expectation:
```
        val cssFile = File(testProject.dir, "build/flow-frontend/mystyle.css")
        expect(true, cssFile.toString()) { cssFile.exists() }
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateBootstrap.java
```diff
@@ -39,8 +39,8 @@
  */
 public class TaskGenerateBootstrap extends AbstractTaskClientGenerator {
 
-    static final String DEV_TOOLS_IMPORT = String
-            .format("import '@vaadin/flow-frontend/vaadin-dev-tools.js';%n");
+    static final String DEV_TOOLS_IMPORT = String.format("import '"
+            + FrontendUtils.JAR_RESOURCES_IMPORT + "vaadin-dev-tools.js';%n");
```
caalador, line 43:
```suggestion
    static final String DEV_TOOLS_IMPORT = String.format("import '%svaadin-dev-tools.js';%n", FrontendUtils.JAR_RESOURCES_IMPORT );
```
As it's string format we should just use the format format.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -152,30 +152,16 @@ public class FrontendUtils {
...
-            + "/";
-
-    /**
-     * Default folder where front-end resources present in the classpath jars
-     * are copied to. Relative to the
-     * {@link com.vaadin.flow.server.InitParameters#BUILD_FOLDER}.
-     */
-    public static final String DEFAULT_FLOW_RESOURCES_FOLDER = "flow-frontend";
-
+    public static final String JAR_RESOURCES_FOLDER = "jarResources";
```
caalador, line 158: This could probably be better as `jar-resources` even though it doesn't make a difference on how it will work. Same for the scripts referencing the folder as a string.

flow-server/src/main/resources/webpack.generated.js
```diff
@@ -174,7 +174,7 @@ const createServiceWorkerPlugin = function () {
 const flowFrontendThemesFolder = path.resolve(flowFrontendFolder, 'themes');
 const themeOptions = {
   devMode: devMode,
-  // The following matches folder 'target/flow-frontend/themes/'
+  // The following matches folder 'frontend/generated/jarResources/themes/'
```
caalador, line 177:
```suggestion
  // The following matches folder 'frontend/generated/jar-resources/themes/'
```
So we don't confuse with the comment not matching the actual

caalador (commented):
Looks ok, but `VaadinSmokeTest:vaadinBuildFrontendShouldScanForResourcesOnClasspath` needs a fix for the tested path.

Also formatting

Artur-:
Seems like gradle is somehow broken still because it does not fail on the asserted path but before that when trying to find the file in `frontend`

PR #14984, opened 2022-11-02
Increase initial delay to show loading indicator

mstahv:
The current initial delay to show loading indicator seems quite short. If one deploys an app to another continent, even no op visits can make the indicator "flash". This may kind of just increases the feeling of "broken app". Especially as the indicator in latest versions is bold visually.

Still, at least mine app feels perfectly usable and don't even realise there is a server visit. I don't have the skills to me the indication look less bolder, so I'd just propose to increase the delay a bit. With this level (450ms) all seems to work fine for my app, but maybe it could be even bit longer still 🤷‍♂️

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/LoadingIndicatorConfigurationMap.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/LoadingIndicatorConfigurationMap.java
@@ -29,7 +29,7 @@ import com.vaadin.flow.internal.StateNode;
 public class LoadingIndicatorConfigurationMap extends NodeMap
         implements LoadingIndicatorConfiguration {
     public static final String FIRST_DELAY_KEY = "first";
-    public static final int FIRST_DELAY_DEFAULT = 300;
+    public static final int FIRST_DELAY_DEFAULT = 450;
     public static final String SECOND_DELAY_KEY = "second";
     public static final int SECOND_DELAY_DEFAULT = 1500;
     public static final String THIRD_DELAY_KEY = "third";


caalador:
What might have changed in the theme as the first delay has been the same 300 since V10 and this issue has not come up before.
Which versions show the indicator too eagerly?

jouni:
I believe the CSS and delays have been the same since the beginning. But I have a similar feeling as Matti, that it seems more annoying in recent versions than in Vaadin 7/8.

jouni:
Where is the default theme for the loading indicator defined? I can easily find the [loading-indicator.css](https://github.com/vaadin/flow/blob/master/flow-tests/test-root-context/src/main/webapp/loading-indicator.css) file, but that’s not it, at least not all of it.

mstahv:
I'd expect that to be on the one line that is changed in this PR 😎 In my app that I was playing around with, the default value was changed like instructed in the docs to 450 and then all felt much smoother.

mstahv:
For a perfect fix also the looks good be changed, but it might be that we have more important things on the table...

mstahv:
Äh, never mind, you were asking for default theme, not default value 😀

jouni:
Yeah. I would want to refresh my memory how the CSS works, if there’s something that I should take into consideration. Otherwise, I feel increasing the delay is okay to do.

caalador:
@jouni Did you see anything else that could be changed/fixed or should we just increase the delay?

caalador:
I guess these are the default styles. Not easy to find when parts are a bit here and there.
https://github.com/vaadin/flow-hilla-common/blob/04f7eea1a5f243078a1596d767d465716d862b95/frontend/packages/common-frontend/src/ConnectionIndicator.ts#L282

jouni:
Thanks, those are the styles I was looking for.

There are delays defined for the CSS animations as well, and I’m having difficulty figuring out how they play together with the these  delays that set the class name on the element.

https://github.com/vaadin/flow-hilla-common/blob/04f7eea1a5f243078a1596d767d465716d862b95/frontend/packages/common-frontend/src/ConnectionIndicator.ts#L335-L356

Since this has been like this for ages, suppose "this is fine", and we can just update one of these values 🤷‍♂️ 

Is it okay that the default value is different on the server than in the client? https://github.com/vaadin/flow-hilla-common/blob/04f7eea1a5f243078a1596d767d465716d862b95/frontend/packages/common-frontend/src/ConnectionIndicator.ts#L57

jouni:
Yeah, it does. I don't have push access to that repo, I would’ve done a companion-PR otherwise. Can you or @mstahv do that instead?

jouni (approved):

caalador (approved):

PR #14992, opened 2022-11-03
chore: Bump commons-fileupload version to 1.4.0

TatuLund:
The same change has been done in Vaadin 23 earlier.

### DIFF
--- a/flow-server/pom.xml
+++ b/flow-server/pom.xml
@@ -57,8 +57,8 @@
         <dependency>
             <groupId>commons-fileupload</groupId>
             <artifactId>commons-fileupload</artifactId>
-            <version>1.3.3</version>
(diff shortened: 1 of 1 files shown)


flow-server/pom.xml
```diff
@@ -57,8 +57,8 @@
         <dependency>
             <groupId>commons-fileupload</groupId>
             <artifactId>commons-fileupload</artifactId>
-            <version>1.3.3</version>
-            <!-- commons-fileupload:1.3.3 depends on commons-io:2.2, override
+            <version>1.4.0</version>
+            <!-- commons-fileupload depends on commons-io, override
```
Artur-, line 61: Is the override still needed?

TatuLund: I am not 100%, but you have it in Vaadin 23

mshabarov (approved):

PR #15014, opened 2022-11-07
feat: Basics for Spring native support

Artur-:
This is the foundation to make Spring native work. It does not cover dynamic use like @Route classes.


caalador:
Actually it fails for all builds now. I'll make a PR for updating gretty so we get it fixed for all branches.

PR #15022, opened 2022-11-08
chore: Use open util package to open a browser

Artur-:


vaadin-spring/pom.xml
```diff
@@ -37,6 +37,11 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>open</artifactId>
+            <version>8.4.0.rc1</version>
+        </dependency>
```
knoobie, line 44: Should this be added to spring and bloating the dependency tree? Shouldn't this be moved to the plugin base or vaadin-dev-server instead?

Artur-: It could be moved to vaadin-dev-server is there would be a good way of communicating the server port to the dev server from a spring startup listener

Artur-: Moved to vaadin-dev-server so it will be excluded if you exclude vaadin-dev-server from the production build

mshabarov (approved):

PR #15029, opened 2022-11-08
chore: Update Node requirement to the recently released version

Artur-:
The version contains security patches and no older versions should be used

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -114,6 +114,11 @@ public class FrontendTools {
             + FrontendUtils.DISABLE_CHECK //
             + MSG_SUFFIX;
 
+    private static final List<FrontendVersion> NPM_BLACKLISTED_VERSIONS = Arrays
+            .asList(new FrontendVersion("6.11.0"),
+                    new FrontendVersion("6.11.1"),
+                    new FrontendVersion("6.11.2"));
+
     private static final FrontendVersion WHITESPACE_ACCEPTING_NPM_VERSION = new FrontendVersion(
             7, 0);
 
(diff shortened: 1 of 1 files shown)


caalador (approved):

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -75,6 +75,12 @@ public class FrontendTools {
 
     public static final String DEFAULT_PNPM_VERSION = "5.18.10";
 
+    static final FrontendVersion SUPPORTED_NODE_VERSION = new FrontendVersion(
+            18, 12, 1);
```
knoobie, line 79: Isn't the fix also backported to v16.18.1 - why the enforcement to v18.12.1 for 23.x?

Artur-: With additional code, it could in theory enforce either 16.x >= 16.18.1 or 18.x >= 18.12.1 but as 18 is the LTS version of Node, it does not seem like a big deal?

Artur-: Thanks for the sanity check. We should indeed stop confusing ourselves and everybody else by mixing "support" aka "This is what is needed for stuff to work" with "recommended" aka "this is what you should use to be safe".

So instead of this we could do:
1. Keep Node 16 support for all V23
2. Bump requirement to Node 18.0 for V24
3. Turn on autoupdates for the auto installed node in 23.3 and 24

That would mean that:
1. Everything keeps working
2. If you install your own node, you are responsible for that (obviously)
3. If you let Flow install node, you will have a safe version as long as you keep up with Flow versions

knoobie: Thanks for reconsidering! Bumping Nodejs to 18.x in 24 sounds like a good way

PR #15043, opened 2022-11-08
chore!: Require Node 18 (npm 8.6)

Artur-:
Node 18 is the latest LTS version and what everbody should at least have going forward

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -33,7 +33,7 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
       - uses: actions/setup-node@v3
         with:
-          node-version: '16'
+          node-version: '18'
       - name: Set up JDK 17
         uses: actions/setup-java@v3
         with:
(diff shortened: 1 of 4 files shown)


caalador (approved):

caalador (changes requested):
Slow tests build shows that the autoUpdate tests need to be fixed:
https://teamcity.vaadin.com/buildConfiguration/Flow_FlowPullRequestsValidationSet_ValidationSlow/150996?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&expandBuildTestsSection=true

caalador:
There are 2 places in `.github/workflows/validation.yml` that have
```
      - uses: actions/setup-node@v3
        with:
          node-version: '16'
```
I guess this could be updated to 18.12.1 so that 2 threads do not try to install the local node version breaking each other.

Artur-:
Updated GH actions. I think the autoupdate tests were fixed by the autoupdate PR, let's see

caalador (approved):

PR #15055, opened 2022-11-09
Revert "test: update gretty version (#15024) (#15028)"

caalador:
The new gretty version required jdk11 and is thus not compatible.
3.0.1 is again found online so there must have been an outage somewhere previously.

### DIFF
--- a/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/MiscSingleModuleTest.kt
+++ b/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/MiscSingleModuleTest.kt
@@ -65,7 +65,7 @@ class MiscSingleModuleTest : AbstractGradleTest() {
             """
             plugins {
                 id 'war'
-                id 'org.gretty' version '4.0.3'
+                id 'org.gretty' version '3.0.1'
                 id("com.vaadin")
             }
             repositories {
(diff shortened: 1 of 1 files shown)


ZheSun88:
yes, there was a gradle issue yesterday. 
they suggested to use artifacts from maven-central instead of gradle portal during that time. 

the gretty-core prior to 3.0.5 is only available in gradle, so that our build failed during that outage, 
updating to 3.0.9 can be one choice here. https://repo.maven.apache.org/maven2/org/gretty/gretty-core/

caalador:
The issue with gretty > 3.0.1 is that then we need to update gradle version as we start getting
`Could not set unknown property 'archiveBaseName' for task ':archiveProduct' of type org.gradle.api.tasks.bundling.Zip.` for 3.0.2+ and the devsoap plugin doesn't work with newer than gradle 5 so that would drop testVaadin8VaadinPlatformMPRProject from gradle

ZheSun88:
it looks platform has been using 3.0.4 since the beginning. 
do we need to change that https://github.com/vaadin/platform/blob/22.0/vaadin-platform-gradle-test/build.gradle#L15

caalador:
If it works I guess you have a new enough gradle and I don't think you need to change it as you are not running mpr tests with the devsoap plugin

tepi (approved):

caalador:
I just noted that the build seems to run with jdk 17
```
[13:35:01 ](https://teamcity.vaadin.com/buildConfiguration/Flow_FlowPullRequestsRunGradleTests/151262?buildTab=log&focusLine=105&logView=flowAware&linesState=105)    openjdk version "17.0.2" 2022-01-18
```
But the version of gradle fails with jdk14+ and should run jdk8 instead. Can you check the script that selects jdk version @ZheSun88 ?

PR #15056, opened 2022-11-09
chore: Make @vaadin/flow-frontend jar-resources change backwards compatible

Artur-:
Related to #14695

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -501,7 +501,7 @@ export const vaadinConfig: UserConfigFn = (env) => {
     base: '',
     resolve: {
       alias: {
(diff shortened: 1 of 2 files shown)


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -501,6 +501,7 @@ export const vaadinConfig: UserConfigFn = (env) => {
     base: '',
     resolve: {
       alias: {
+        '@vaadin/flow-frontend': path.resolve(frontendFolder, 'generated', 'jar-resources'),
```
caalador, line 504: Shouldn't it be ok to use `jarResourcesFolder` here directly?

Artur-: Yes :)

caalador (approved):

PR #15084, opened 2022-11-10
feat(hilla): make tsconfig.json React compatible

platosha:
Summary of changes:

- Added a good default for `jsx` setting
- Relaxed file extensions for TypeScript, so that `.tsx` and such are considered

This should not interfere with existing TypeScript / Lit-based code.

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
@@ -7,6 +7,7 @@
   "flow_version": "23.3.0",
   "compilerOptions": {
     "sourceMap": true,
+    "jsx": "react-jsx",
     "inlineSources": true,
     "module": "esNext",
     "target": "es2020",
@@ -29,8 +30,7 @@
     }
   },
   "include": [
-    "frontend/**/*.ts",
-    "frontend/index.js",
+    "frontend/**/*",
     "types.d.ts"
   ],
   "exclude": []
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
@@ -4,7 +4,7 @@
 // You might want to change the configurations to fit your preferences
 // For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
 {
-  "flow_version": "23.3.0",
+  "flow_version": "23.3.1",
   "compilerOptions": {
     "sourceMap": true,
     "jsx": "react-jsx",
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
@@ -4,7 +4,7 @@
 // You might want to change the configurations to fit your preferences
 // For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
 {
-  "flow_version": "23.3.1",
+  "flow_version": "23.3.0.1",
   "compilerOptions": {
     "sourceMap": true,
     "jsx": "react-jsx",
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
@@ -48,8 +48,8 @@ public class TaskGenerateTsConfig extends AbstractTaskClientGenerator {
     private static final String VERSION = "flow_version";
     private static final String ES_TARGET_VERSION = "target";
     private static final String TSCONFIG_JSON_OLDER_VERSIONS_TEMPLATE = "tsconfig-%s.json";
-    private static final String[] vaadinVersions = { "latest", "v23.2", "v23.1",
-            "v22", "v14", "osgi" };
+    private static final String[] vaadinVersions = { "latest", "v23.3.0",
+            "v23.2", "v23.1", "v22", "v14", "osgi" };
 
     //@formatter:off
     static final String ERROR_MESSAGE =
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig-v23.3.0.json
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig-v23.3.0.json
@@ -4,10 +4,9 @@
 // You might want to change the configurations to fit your preferences
 // For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
 {
-  "flow_version": "23.3.1",
+  "flow_version": "23.3.0",
   "compilerOptions": {
     "sourceMap": true,
-    "jsx": "react-jsx",
     "inlineSources": true,
     "module": "esNext",
     "target": "es2020",
@@ -30,7 +29,8 @@
     }
   },
   "include": [
-    "frontend/**/*",
+    "frontend/**/*.ts",
+    "frontend/index.js",
     "types.d.ts"
   ],
   "exclude": []
--- a/flow-server/src/main/resources/tsconfig.json
+++ b/flow-server/src/main/resources/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "sourceMap": true,
+    "jsx": "react-jsx",
     "inlineSources": true,
     "module": "esNext",
     "target": "es2020",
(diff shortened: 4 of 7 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -5,11 +5,13 @@
...
 {
   "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
     "sourceMap": true,
+    "jsx": "react-jsx",
     "inlineSources": true,
     "module": "esNext",
     "target": "es2020",
-    "moduleResolution": "node",
+    "moduleResolution": "node16",
```
Artur-, line 14: What is the difference between "node" and "node16"? We already require node 16

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -5,11 +5,13 @@
 // For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
 {
   "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
     "sourceMap": true,
+    "jsx": "react-jsx",
```
Artur-, line 10: Does this have any impact when you are not using jsx?

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -5,11 +5,13 @@
 // For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
 {
   "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
```
Artur-, line 8: Is this needed for React or just handy?

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -19,6 +21,7 @@
     "noUnusedLocals": false,
     "noUnusedParameters": false,
     "experimentalDecorators": true,
+    "importsNotUsedAsValues": "error",
```
Artur-, line 24: How is this related to React?

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -27,8 +30,7 @@
     }
   },
   "include": [
-    "frontend/**/*.ts",
-    "frontend/index.js",
+    "frontend/**/*",
```
Artur-, line 33: This will now include images, json, everything. Is that the intent or are we mainly looking to add `*.tsx` here?

Artur- (commented):
Please clarify the changes so we know in the future why they were made

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
Artur-: Error:  frontend/generated/index.ts(20,25): error TS2835: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean '../../target/frontend/generated-flow-imports.js'?

At least it is not 100% compatible

platosha: https://github.com/microsoft/TypeScript/issues/35290

enables a very handy `exports` definition in `package.json` in dependencies, which is already supported by Vite but confusingly not by TypeScript... We use it in `@hilla/react-components`

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
platosha: yes

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
platosha, line 24:
```suggestion
    "experimentalDecorators": true,
```

platosha: not related it seems,

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -5,11 +5,13 @@
 // For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
 {
   "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
     "sourceMap": true,
```
platosha, line 9:
```suggestion
    "sourceMap": true,
```

seems not needed, as `import * as React from 'react';` is preferred

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
Lodin: It also defaults to the Node 16+ ESM/CJS resolution algorithm which allows importing CJS from ESM without any specific interop.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
platosha: It’s a nice tweak, the default is "react" option, which requires a full `React` import. See https://www.typescriptlang.org/tsconfig#jsx

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
Lodin: > At least it is not 100% compatible

Unfortunately, yes. Still, I believe this is very important change; it looks like to become the standard de facto, so we need to push it

Artur-: Yeah but how much does it break? Can we do it for 23.3 or only for 24?

Artur-: I get 38 errors from the default Hilla project when enabling this

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
Lodin: I would suggest to keep it this way. Otherwise, if the user wants to include image/css/other file to the bundle (as `base64` string or as an automatic link), the Typescript will show an error even if we have a module declaration in the `types.d.ts` file (as far as I remember).

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
Lodin: Yeah, probably, it's better to wait for v24.

Artur- (approved):

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -7,6 +7,7 @@
   "flow_version": "23.3.0",
```
Artur-, line 7: This needs to be updated

mshabarov:
@Artur- @platosha yes, we should update some parts to make autoupdate works fine:
1. Increment version in tsconfig.json ("flow_version" property): can be `"flow_version" : "23.3.1"`.
2. Create a new tsconfig-v23.3.0.json template in flow-server resources and put the old content there (before your patch)
3. Update `vaadinVersion` array in `TaskGenerateTsConfig` with `v23.3.0`
4. Put a new content in `tsconfig-reference.json` in tests.

P.S. yes, I know this manual work looks ugly...maybe I can remove (2) - (4) someday.

platosha:
@Artur- @mshabarov OK done, even though `23.3.0` was not even released yet...

platosha:
Whoa, a test failed.

@mshabarov FYI the instructions were missing a step: https://github.com/vaadin/flow/pull/15084/commits/9af2f3cb6229178c422a1197103cfbe7e5488930

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -4,9 +4,10 @@
 // You might want to change the configurations to fit your preferences
 // For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
 {
-  "flow_version": "23.3.0",
+  "flow_version": "23.3.1",
```
Artur-, line 7: Won't this cause confusion forever as it is not part of 23.3.1 but 23.30? Can it be 23.3.0.1 instead?

Artur- (approved):

mshabarov (approved):

PR #15085, opened 2022-11-10
chore: Use env instead of custom parameter to webpack (#9934) (CP: 2.8)

Artur-:
This is compatible with Webpack 5

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendUtilsTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendUtilsTest.java
@@ -244,14 +244,13 @@ public class FrontendUtilsTest {
         List<String> command = Arrays.asList("./node/node",
                 "./node_modules/webpack-dev-server/bin/webpack-dev-server.js",
                 "--config", "./webpack.config.js", "--port 57799",
-"--env watchDogPort=57798", "-d", "--inline=false", "--progress",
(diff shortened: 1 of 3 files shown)


caalador:
`liveReloadOnTouchedFrontendFile` is failing and I can't see why this change would have made a difference there.

Artur-:
For the record: Live reload depends on the watchdog and using `--env` does not actually pass an environment variable to the webpack process. Not sure why that has ever worked (or maybe it has not...)

flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendUtilsTest.java
```diff
@@ -244,14 +244,14 @@ public void commandToString_longCommand_resultIsWrapped() {
         List<String> command = Arrays.asList("./node/node",
                 "./node_modules/webpack-dev-server/bin/webpack-dev-server.js",
                 "--config", "./webpack.config.js", "--port 57799",
-                "--watchDogPort=57798", "-d", "--inline=false", "--progress",
+"--env watchDogPort=57798", "-d", "--inline=false", "--progress",
```
caalador, line 247: The test could just drop the watchDogPort all in all as it doesn't work.

Artur-: Done

caalador (approved):

PR #15087, opened 2022-11-10
chore: Import files using extension

Artur-:
There is no gain in dropping the extension and it does not work if we want to use 'node16' module resolution


Artur-:
There might be a `.ts`/`.js` expectation mismatch here

caalador (approved):

PR #15095, opened 2022-11-10
chore: Import require to be future proof

Artur-:
Using node16 module resolution does not work without this


Lodin:
We don't have `"type": "module"` enabled in `package.json`, so either the file should be `.mts`, or the `"type": "module"` should be enabled. I assume that tests are passing because Vite does some black magic which I would prefer not to rely on.

Also, since we're not introducing Node 16 support yet (according to vaadin/start#2102), I would suggest to postpone this PR.

PR #15098, opened 2022-11-11
chore: Upgrade to Spring Boot 3 RC2

Artur-:
As Spring Boot now uses Tomcat 10.1, we cannot mix Jetty 11 (servlet 5) and Tomcat 10.1 (servlet 6) in the same test modules so they are all run with Jetty 11 for now.

### DIFF
--- a/flow-tests/vaadin-spring-tests/test-spring-boot-only-prepare/pom.xml
+++ b/flow-tests/vaadin-spring-tests/test-spring-boot-only-prepare/pom.xml
@@ -37,6 +37,16 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.springframework.boot</groupId>
+                    <artifactId>spring-boot-starter-tomcat</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-jetty</artifactId>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -235,7 +235,7 @@
             <dependency>
                 <groupId>jakarta.servlet</groupId>
                 <artifactId>jakarta.servlet-api</artifactId>
-                <version>5.0.0</version>
```
knoobie, line 238: FYI: This comment is probably really important for you guys as well: https://github.com/spring-projects/spring-framework/issues/29435#issuecomment-1309127414

Artur-: Do you see any need for servlet 5 support instead of going to servlet 6?

flow-tests/vaadin-spring-tests/test-spring-boot-only-prepare/pom.xml
```diff
@@ -37,6 +37,16 @@
...
+            <exclusions>
+                <exclusion>
+                    <groupId>org.springframework.boot</groupId>
+                    <artifactId>spring-boot-starter-tomcat</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-jetty</artifactId>
```
mshabarov, line 49: Just want to clarify for myself: this is needed to run all the tests within the same test module with either tomcat or jetty, right? But jetty 11 itself seems to be compatible with the Servlet 6.0, as you bumped jakarta.servlet-api version in the root pom?

mshabarov (approved):

flow-tests/vaadin-spring-tests/test-spring-boot-only-prepare/pom.xml
Artur-: By default Spring Boot runs on Tomcat. This makes it run on Jetty instead. Spring Boot 3 is still compatible with servlet 5, i.e. Jetty 11 as there is nothing that uses the new Servlet 6 API in Flow or in Spring Boot 3

PR #15104, opened 2022-11-11
fix: Improve polymer2lit error handling

vursen:
## Description

- Changed the implementation of `FrontendConverter` so that it is based on `FrontendUtils.executeCommand` which provides better error handling over `FrontendUtils.createProcessBuilder` which was used before.
- Added more logging when the converter skips a file.

A follow-up to #14972 

Depends on #15108


flow-plugins/flow-plugin-base/src/test/java/com/vaadin/flow/plugin/base/ConvertPolymerCommandTest.java
```diff
@@ -5,6 +5,7 @@
 import java.net.URISyntaxException;
 import java.nio.file.Path;
 
+import org.apache.commons.lang3.Conversion;
```
mshabarov, line 8: Leftover? Remove if not used.

flow-polymer2lit/src/main/java/com/vaadin/flow/polymer2lit/FrontendConverter.java
```diff
@@ -77,6 +72,15 @@ public int convertFile(Path filePath, boolean useLit1,
         ProcessBuilder builder = FrontendUtils.createProcessBuilder(command);
         builder.inheritIO();
         Process process = builder.start();
-        return process.waitFor();
+        if (process.waitFor() != 0) {
+            throw new ConversionFailedException(
+                    "An error occurred while the conversion. See logs for more details.");
```
mshabarov, line 77: Would the root cause of the frontend converter's failure be visible in the output for the developer? I see no code which takes the output and logs it.

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
```diff
@@ -101,8 +101,7 @@ public void execute() throws IOException, InterruptedException {
             try {
                 serverConverter.convertFile(filePath);
             } catch (Exception e) {
-                adapter.logError("An error occured while processing "
-                        + filePath.toString(), e);
+                adapter.logError("An error occurred while processing.", e);
```
mshabarov, line 104: Would the information about the file being converted be logged in another place. Still good to have this information in the logs.

mshabarov (changes requested):

flow-polymer2lit/src/main/java/com/vaadin/flow/polymer2lit/FrontendConverter.java
vursen: There is a `build.inheritIO()` instruction in the code that forwards any Node process output to the Java process stdout, so it should work.

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
vursen: Yeah, good point. Let me add something.

flow-plugins/flow-plugin-base/src/test/java/com/vaadin/flow/plugin/base/ConvertPolymerCommandTest.java
vursen: Removed.

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
```diff
@@ -97,23 +97,33 @@ public void execute() throws IOException, InterruptedException {
         Path lookupPath = getLookupPath();
 
         for (Path filePath : getFilePathsByGlob(lookupPath, SERVER_GLOB)) {
-            adapter.logInfo("Processing " + filePath.toString() + "...");
             try {
-                serverConverter.convertFile(filePath);
+                adapter.logInfo("Processing " + filePath.toString() + "...");
```
mshabarov, line 101: nit: since this is used inside a loop, would it better to use `String.format` instead to not produce a redundant string objects?

flow-tests/test-frontend/test-npm/package.json
```diff
@@ -2,79 +2,68 @@
   "name": "no-name",
   "license": "UNLICENSED",
   "dependencies": {
-    "@polymer/polymer": "3.5.1",
     "@vaadin/common-frontend": "0.0.17",
-    "@vaadin/router": "1.7.4",
-    "@vaadin/vaadin-text-field": "23.3.0-alpha2",
     "construct-style-sheets-polyfill": "3.1.0",
-    "lit": "2.4.1"
+    "lit": "2.4.1",
```
mshabarov, line 7: Do we really need these changes in package.json of the test module?

mshabarov (changes requested):

flow-tests/test-frontend/test-npm/package.json
vursen: Oh, it isn't intentional, let me remove that.

mshabarov (approved):

PR #8519, opened 2020-06-08
Chunk module submodules

caalador:
Also take into account that
chunk modules may have
submodules.

Fixes #8493

PR #15124, opened 2022-11-15
fix: Filter chunks to minimize stats.json

TatuLund:
https://github.com/vaadin/flow/pull/6919 added optimization of stats.json

However due https://github.com/vaadin/flow/pull/8519 there is regression as chunks are not filtered in similar fashion and stats.json is overgrown again.

### DIFF
--- a/flow-server/src/main/resources/plugins/stats-plugin/stats-plugin.js
+++ b/flow-server/src/main/resources/plugins/stats-plugin/stats-plugin.js
@@ -112,8 +112,8 @@ function collectChunks(statsJson, acceptedChunks) {
             if (module.modules) {
               slimModule.modules = collectSubModules(module);
             }
+            modules.push(slimModule);
           }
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/plugins/stats-plugin/stats-plugin.js
```diff
@@ -103,24 +103,28 @@ function collectChunks(statsJson, acceptedChunks) {
...
+              id: module.id,
+              name: module.name,
+              source: module.source
+            };
+            if (module.modules) {
+              slimModule.modules = collectSubModules(module);
+            }
           }
           modules.push(slimModule);
         });
```
caalador, line 117: The push slim module needs to be inside the if else it will push undefined to modules.
```suggestion
            }
            modules.push(slimModule);
          }
        });
```

caalador (approved):

PR #15125, opened 2022-11-15
chore: Do not check enableDevServer() where it is not needed

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -521,8 +521,7 @@ private static String getFileContent(VaadinService service, String path)
         try {
             Optional<DevModeHandler> devModeHandler = DevModeHandlerManager
                     .getDevModeHandler(service);
-            if (!config.isProductionMode() && config.enableDevServer()
```
caalador, line 524: Wont this fail if you run <goal>prepare-frontend</goal><goal>build-frontend</goal> with productionMode set to false which is what the feature is for.

Artur-: Is not the dev mode handler absent in the case that you run without enableDevServer?

caalador: One would hope so yes.

caalador (approved):

Issue #14958, opened 2022-10-31
Authentication API to get the current user and perform logout


heruan:
### Describe your motivation

Start projects have the `AuthenticatedUser` bean which provides the current user and the logout feature. SSO Kit provides the [`AuthenticationContext`](https://github.com/vaadin/sso-kit/blob/main/sso-kit-starter/src/main/java/com/vaadin/sso/starter/AuthenticationContext.java) which is very similar.

If Flow would provide an API for this, it could be used both in Start projects and in SSO Kit.

### Describe the solution you'd like

This is the default implementation for SSO Kit: [SingleSignOnConfiguration.java#L199-L254](https://github.com/vaadin/sso-kit/blob/f0ccef4ce5a63d998be748318807602e233d285d/sso-kit-starter/src/main/java/com/vaadin/sso/starter/SingleSignOnConfiguration.java#L199-L254) which can be used as a starting point.

### Additional context

The API should be extendable/configurable enough to be able for other projects (e.g. SSO Kit) to add logout handlers.

PR #15129, opened 2022-11-15
feat: Add the authentication context API to the Vaadin Web Security

mshabarov:
AuthenticationContext bean provides an API for getting authenticated user details and logout capabilities.

Fixes https://github.com/vaadin/flow/issues/14958

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
@@ -15,24 +15,54 @@
  */
 package com.vaadin.flow.spring.security;
 
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.List;
 import java.util.Optional;
 
-import org.springframework.security.core.AuthenticatedPrincipal;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.security.authentication.AnonymousAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.web.authentication.logout.CompositeLogoutHandler;
+import org.springframework.security.web.authentication.logout.LogoutHandler;
+import org.springframework.security.web.authentication.logout.LogoutSuccessHandler;
+
+import com.vaadin.flow.component.UI;
+import com.vaadin.flow.server.VaadinServletRequest;
+import com.vaadin.flow.server.VaadinServletResponse;
 
 /**
- * An interface to access the authentication context of the application.
+ * The authentication context of the application.
  * <p>
- * An instance of this interface is available for injection as bean in view and
+ * An instance of this class is available for injection as bean in view and
  * layout classes.
+ *
+ * It allows to access authenticated user information and to initiate the logout
+ * process.
  *
  * @author Vaadin Ltd
  * @since 23.3
  */
-public interface AuthenticationContext {
+public class AuthenticationContext implements Serializable {
+
+    private static final Logger LOGGER = LoggerFactory
+            .getLogger(AuthenticationContext.class);
+
+    private transient LogoutSuccessHandler logoutSuccessHandler;
+
+    private transient CompositeLogoutHandler logoutHandler;
 
     /**
      * Gets an {@link Optional} with an instance of the current user if it has
      * been authenticated, or empty if the user is not authenticated.
+     *
+     * Anonymous users are considered not authenticated.
      *
      * @param <U>
      *            the type parameter of the expected user instance
@@ -40,19 +70,26 @@ public interface AuthenticationContext {
      *            the type of the expected user instance
      * @return an {@link Optional} with the current authenticated user, or empty
      *         if none available
+     * @throws ClassCastException
+     *             if the current user instance does not match the given
+     *             {@code userType}.
      */
-    <U> Optional<U> getAuthenticatedUser(Class<U> userType);
+    public <U> Optional<U> getAuthenticatedUser(Class<U> userType) {
+        return getAuthentication().map(Authentication::getPrincipal)
+                .map(userType::cast);
+    }
 
     /**
-     * Gets an {@link Optional} subclass of {@link AuthenticatedPrincipal} with
-     * an instance of the current user if it has been authenticated, or empty if
-     * the user is not authenticated.
+     * Indicates whether a user is currently authenticated.
      *
-     * @return an {@link Optional} with the current authenticated user, or empty
-     *         if none available
+     * Anonymous users are considered not authenticated.
+     *
+     * @return {@literal true} if a user is currently authenticated, otherwise
+     *         {@literal false}
      */
-    default Optional<AuthenticatedPrincipal> getAuthenticatedUser() {
-        return getAuthenticatedUser(AuthenticatedPrincipal.class);
+    public boolean isAuthenticated() {
+        return getAuthentication().map(Authentication::isAuthenticated)
+                .orElse(false);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @param <U>
+     *            the type parameter of the expected user instance
+     * @param userType
+     *            the type of the expected user instance, a subclass of
+     *            {@link UserDetails}
+     * @return an {@link Optional} with the current authenticated user, or empty
+     *         if none available
+     */
+    <U extends UserDetails> Optional<U> getAuthenticatedUser(Class<U> userType);
```
heruan, line 45: Bounding the generic type to `UserDetails` is problematic since that interface is e.g. not implemented by the `DefaultOidcUser` which is the user principal provided by the Spring Security OpenID Connect helpers.

Either we use a less restrictive interface, such as `AuthenticatedPrincipal` or we just unbind the generic type and let the user specify any.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -121,6 +135,22 @@ public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
...
+    @Bean(name = "VaadinAuthenticationContext")
+    public AuthenticationContext getAuthenticationContext(HttpSecurity http)
+            throws Exception {
+        LogoutConfigurer<HttpSecurity> logoutConfigurer = http.logout();
+        DefaultAuthenticationContext authenticationContext = new DefaultAuthenticationContext();
+        authenticationContext.setLogoutHandlers(
+                logoutConfigurer.getLogoutSuccessHandler(),
+                logoutConfigurer.getLogoutHandlers());
+        return authenticationContext;
+    }
```
heruan, line 152: We must be sure this happens after the `filterChain` method has been invoked, since we need to be sure that the `HttpSecurity` instance has all the logout handlers set (i.e. the chain is already built).

Adding `@DependsOn("VaadinSecurityFilterChainBean")` might be enough, but this needs to be tested 🙂

heruan (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
mshabarov: Good point, I've just noticed that tests fail because of empty logout handlers lists.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -105,6 +118,9 @@ public abstract class VaadinWebSecurity {
     @Value("#{servletContext.contextPath}")
     private String servletContextPath;
 
+    private final DefaultAuthenticationContext authenticationContext =
+            new DefaultAuthenticationContext();
```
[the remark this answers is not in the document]
mshabarov: I'm really not sure whether we should make this bean extendable on the client's codes. My vision is that it's like a service bean providing logout and authenticated user common services and should be configured by Vaadin. If the developer wants to configure the logout handlers or to, for example, invalidate session on logout, he better uses `LogoutConfigurer` in the overriden `configure(HttpSecurity http)` method. Or I missed something?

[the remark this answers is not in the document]
knoobie: Understandable! If it's by design not overrideable, you could think about the removal of the Interface all together making it implicit that this can't be overwritten and see how people gonna use / complain about it. 

Your example was exactly what I had in mind :) we normally invalide sessions on logout and I was confused why the default implementation doesn't do it and was wondering how I can override it.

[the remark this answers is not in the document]
mshabarov: Yes, good point, I also thought to add invalidation by default.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
```diff
@@ -0,0 +1,64 @@
...
+ * layout classes.
+ *
+ * @author Vaadin Ltd
+ * @since 23.3
+ */
+public interface AuthenticationContext {
+
+    /**
+     * Gets an {@link Optional} with an instance of the current user if it has
+     * been authenticated, or empty if the user is not authenticated.
```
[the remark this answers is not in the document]
mcollovati: @heruan what do you think? Does it make sense to throw an exception (ClassCastException) if the current user object does not match the expected one?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Gets an {@link Optional} subclass of {@link AuthenticatedPrincipal} with
+     * an instance of the current user if it has been authenticated, or empty if
+     * the user is not authenticated.
+     *
+     * @return an {@link Optional} with the current authenticated user, or empty
+     *         if none available
+     */
+    default Optional<AuthenticatedPrincipal> getAuthenticatedUser() {
```
mcollovati, line 54: Having `AuthenticatedPrincipal` as potential return type may cause problems because, for example, `org.springframework.security.core.userdetails.User` does not implement it and `UserDetails` does not extend it.
Does it make sense to you to simply return an `Optional<Object>`, since also `Authentication::getPrincipal` return type is `Object`?
If a developer needs a specific type, the `getAuthenticatedUser(Class<U> userType)` method can be used.

mcollovati: Or even remove the method at all and replace it with a `boolean isAuthenticated()`, since I think the only use of this signature may be something like `if (authContext.getAuthenticatedUser().isPresent()) { ... }`

mshabarov: Yes, having `Optional<Object> getAuthenticatedUser()` means that `<U> Optional<U> getAuthenticatedUser(Class<U> userType);` would be used almost always. Having `boolean isAuthenticated()` instead sounds good!

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/UidlRedirectStrategy.java
```diff
@@ -0,0 +1,30 @@
...
+ * @since 1.0
+ */
+public class UidlRedirectStrategy extends DefaultRedirectStrategy {
+
+    @Override
+    public void sendRedirect(HttpServletRequest request,
+            HttpServletResponse response, String url) throws IOException {
+        final var servletMapping = request.getHttpServletMapping().getPattern();
+        if (HandlerHelper.isFrameworkInternalRequest(servletMapping, request)) {
+            UI.getCurrent().getPage().setLocation(url);
```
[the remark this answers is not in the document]
mcollovati: In theory, the `Page.setLocation()` should only be engaged as a result of a user interaction (e.g. pressing a button that invokes `AuthenticationContext.logout()` or the same function invoked through `UI.access()`), but it makes sense to add a check and a WARN or DEBUG log message to be able to debug potentially unexpected situations

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -504,6 +531,25 @@ protected ViewAccessChecker getViewAccessChecker() {
...
     }
 
+    /**
+     * Sets additional {@link LogoutHandler}s that will participate in logout
+     * process.
+     *
+     * @param registry
+     *            used to add custom handlers.
+     */
+    protected void addLogoutHandlers(Consumer<LogoutHandler> registry) {
```
mshabarov, line 541: Should it be `SerializableConsumer` ?

mcollovati: It should not be needed, since the Consumer should not be referenced anywere in the VaadinSession

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
heruan: Has this been addressed?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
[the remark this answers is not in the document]
heruan: Yes, it makes sense to me: let's throw an exception if the casting fails.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
mcollovati: Sorry, I missed the comment. I'll check it

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
heruan: I'm in favor of adding a `boolean isAuthenticated()` and also unbinding the user type, but then I'd also remove the `Optional<Object> getAuthenticatedUser()` in favor of the `Optional<T> getAuthenticatedUser(Class<? extends T> type)` overload.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
mcollovati: The bean definition method does not take the `HttpSecurity` parameter anymore.
It returns a field initialized instance, that is configured within `filterChain`

mcollovati:
A summary of the change:

*  `AuthenticationContext` is a concrete class that gets access to the authenticated user and allows performing logout integrated with Spring Security
* Developers can plug additional `LogoutHandler`s in `VaadinWebSecurity.addLogoutHandlers()`
* `VaadinWebSecurity` configures a LogoutSuccessHandler with a redirect strategy that can handle also XHR UIDL requests (`UidlRedirectStrategy`)

heruan (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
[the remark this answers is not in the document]
mcollovati: Session is now invalidated by default.  Additional `LogoutHandler`s can be configured with `VaadinWebSecurity.addLogoutHandlers()`

PR #15136, opened 2022-11-15
test: Delete webpack specific tests

Artur-:


caalador:
Was webpack being dropped in 24 or why are we cleaning the tests?

Artur-:
Webpack was deprecated in V23 so it will be removed in V24. This is a first step to speed up the build

caalador (approved):
Just confirming that this was the case.

PR #15152, opened 2022-11-16
fix: fix faulty cleanup of stats.json

caalador:
fixes cleanup of stats.json contents
so that the needed templates are
not dropped.


caalador:
Test with and without in `vaadin-spring/vaadin-spring-tests/test-spring-boot-only-prepare` for vaadin-spring 12.4 in the spring repository

PR #15188, opened 2022-11-17
fix!: Use dedicated mapping for Vaadin push endpoint

### LINKED ISSUE #14641: Websocket in combination with AJP broken behind reverse-proxy in 23.2.

tepi:
Removes pushUrl configuration option, and instead maps push always to a dedicated path (/VAADIN/push).

Related-to #14641

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
@@ -21,6 +21,7 @@ import jakarta.servlet.ServletException;
 
 import java.io.IOException;
 
+import jakarta.servlet.ServletRegistration;
 import org.atmosphere.cache.UUIDBroadcasterCache;
 import org.atmosphere.client.TrackMessageSizeInterceptor;
 import org.atmosphere.cpr.ApplicationConfig;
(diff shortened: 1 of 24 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
```diff
@@ -236,6 +245,17 @@ public AtmosphereFramework addInitParameter(String name,
         return atmosphere;
     }
 
+    private static String findFirstUrlMapping(ServletConfig config) {
+        String name = config.getServletName();
+        String firstMapping = "/";
+        if (name != null) {
+            ServletRegistration reg = config.getServletContext()
+                    .getServletRegistrations().get(name);
+            firstMapping = reg.getMappings().stream().findFirst().orElse("/");
```
mshabarov, line 254: I propose to sort the mappings before taking the first one to make it deterministic and return the same mapping always.

mshabarov (changes requested):

tepi:
Note: if this is merged, we also need the change the push configuration documentation accordingly

mshabarov (approved):

PR #15216, opened 2022-11-22
refactor: Record project folder inside node_modules instead

### LINKED ISSUE #14908: Frontend build breaks if folder name changed
### Description of the bug

When having a project, changing the project folder name with no other changes after having run the project once breaks the build with missing frontend modules.

This is due to npm creating a symlink to `./target/flow-frontend` and the symlink is absolute, so changing the project folder name doesn't break the hash check so no `npm install` is run to fix the symlink.

### Expected behavior

Changing folder name should not break build with missing frontend modules.

Get app from start.
Run application and shutdown.
change folder name
Run application again.

### Versions

- Vaadin / Flow version: 14+

Artur-:
The package.json hash is now again a hash of the contents and can be used by other tools to check if the content (installed packages) has changed

Fixes #14908 in a different way


caalador (approved):

cdir:
We need a backport for 23.2 because package.json now updates with every build job depending on the location of the npm installation.

PR #15222, opened 2022-11-22
fix: Serve index.html directly from the project frontend folder when a dev server is not used or available

Artur-:
While this does not directly solve any issue it will make https://github.com/vaadin/hilla/issues/225 clearer when there is no longer any error about a missing index.html but instead you see in the logs:

```
[qtp1200171710-60] INFO com.vaadin.flow.server.DefaultDeploymentConfiguration - Vaadin is running in DEVELOPMENT mode - do not use for production deployments. [qtp1200171710-60] WARN com.vaadin.flow.server.VaadinServletService - no DevModeHandlerManager implementation found but dev server enabled. Include the com.vaadin.vaadin-dev-server dependency.
```


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -521,19 +521,44 @@ private static String getFileContent(VaadinService service, String path)
...
+
+    public static File getFileFromProjectDir(DeploymentConfiguration config,
+            String path) {
+        File f = new File(getProjectDir(config), path);
+        if (f.exists()) {
+            return f;
+        }
+        return null;
+    }
+
```
caalador, line 561: This is not used from anywhere so it is a bit of extra.

caalador (changes requested):

caalador:
Quite many IT tests fail with 
`Unable to find index.html. It should be available in XYZ`

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -521,19 +521,44 @@ private static String getFileContent(VaadinService service, String path)
...
                     .getDevModeHandler(service);
-            if (!config.isProductionMode() && devModeHandler.isPresent()) {
+            if (config.isProductionMode()) {
+                // In production mode, this is on the class path
+                content = getFileFromClassPath(service, path);
+            } else if (devModeHandler.isPresent()) {
                 content = getFileFromDevModeHandler(devModeHandler.get(), path);
+            } else {
+                // Get directly from the frontend folder in the project
+                content = getFileFromProjectFrontendDir(config, path);
```
caalador, line 531: Should this be an `if(content==null) {` to fallback on actual file if not found from classpath/devModeHandler

Artur-:
Yes so the "not production mode but not using a dev server" mode is broken. In this case, the files in frontend: `index.html` and `web-component.html` are just the templates if the html files and the real files can be found from the classpath where the frontend build has placed them

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
caalador: True. Was just thinking about when developing the app as else it should get it in production mode.

Issue #15238, opened 2022-11-23
PiT 23.3: implement a mechanism for updating `tsconfig.json`


manolo:
### Description of the bug

Any upgrade of an app having `tsconfig.json` file in git, will fail with an error.

#### Workaround

Remove the file manually

### Expected behavior

User does not have to do any action, flow updates the file appropriatelly

### Minimal reproducible example

Checkout any project with the `tsconfig.json` and run it. 

### Versions

- Vaadin / Flow version:  23.3.0.alpha3 / 23.3.0.beta1

Artur-:
Would this work?

1. If tsconfig.json matches a "known" old version, we replace it with the new one without asking. We must match all variants back to V14 as somebody might still have an old tsconfig that has happened to work with newer versions, or she is upgrading from V14
2. Otherwise tsconfig.json has presumably been customized. We overwrite tsconfig.json anyway (if you have customized it, it is in the repository) and then throw an exception saying that the user should verify the changes as we just overwrote it. If the user just starts the app again, the new one will be used and no exception is thrown as it now matches the 23.3 version

Issue #15237, opened 2022-11-23
PiT 23.3: frontend error when compiling karaf example


manolo:
### Description of the bug

Upgrading `vaadin-flow-karaf-example` or `base-starter-flow-quarkus` to 23.3.0.alpha3 causes an error during frontend compilation

```
ERROR] node_modules/@types/node/globals.d.ts(72,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; abort(reason?: any): AbortSignal; timeout(milliseconds: number): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; }'.
```

### Expected behavior

Application works

### Minimal reproducible example

- Checkout, upgrade and compile the demo
```
git clone https://github.com/vaadin/vaadin-flow-karaf-example.git
cd vaadin-flow-karaf-example
mvn -B -q versions:set-property -Dproperty=vaadin.version -DnewVersion=23.3.0.alpha3
mvn clean install
```
- There is the error above in terminal

- The same happens with the quarkus demo
```
git clone https://github.com/vaadin/base-starter-flow-quarkus.git 
cd base-starter-flow-quarkus
mvn -B -q versions:set-property -Dproperty=vaadin.version -DnewVersion=23.3.0.alpha3
mvn
```
- The error in console:
```
ERROR(TypeScript)  Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; abort(reason?: any): AbortSignal; timeout(milliseconds: number): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; }'.
```

### Versions

- Vaadin / Flow version: 23.3.0.alpha3 / 23.3.0.beta1

PR #15247, opened 2022-11-23
feat: Auto-update TS config file to the latest version

mshabarov:
Simplifies the upgrade to newer Vaadin versions with breaking changes in tsconfig.json.
Adds a version property to the default tsconfig.json and:
- Overrides tsconfig.json content if the version is obsolete
- Throws an exception with explanatory message if custom content is detected in tsconfig.json
- Doesn't override if the version is up-to-date.

Fixes https://github.com/vaadin/flow/issues/15238


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
```diff
@@ -111,19 +143,98 @@ private String getDefaultEsTargetVersion() throws ExecutionFailedException {
...
+                    TSCONFIG_JSON);
+            String projectTsConfigAsString = FileUtils
+                    .readFileToString(projectTsConfigFile, UTF_8);
+
+            JsonObject projectTsConfigContent;
+            try {
+                projectTsConfigContent = parseTsConfig(projectTsConfigAsString);
+            } catch (Exception e) {
+                // This could be a malformed tsconfig, leave it alone
+                log().debug("Unable to parse tsconfig.json", e);
```
Artur-, line 179: Shouldn't this be an error if tsconfig cannot be parsed? Won't the app be broken also? At least logged as an error

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -4,6 +4,7 @@
 // You might want to change the configurations to fit your preferences
 // For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
 {
+  "version": "23.3.0",
```
Artur-, line 7: If we add a custom attribute to tsconfig.json we should probably prefix it with `flow_` to avoid problems later on. Should the version here be the Flow version where tsconfig.json was last changed or should it just be a running number starting from 1?

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
mshabarov: Yes, `error` is a correct level here. I assume this kind of error is very unlikely in a projects.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
mshabarov: Good point, will add flow prefix. I vote for the flow version where it was last changed. Would be more helpful to track the history at least.

mshabarov: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
mshabarov: done

manolo:
It works for me and fix the PiT issues:
Fixes https://github.com/vaadin/hilla/issues/673
Fixes https://github.com/vaadin/flow/issues/15237

manolo (approved):

mshabarov:
Let's merge this patch now to speed up the release, hope @Artur- doesn't mind :)

Artur- (approved):

PR #15278, opened 2022-11-25
fix: stats in development mode

caalador:
In development mode the local
frontend stat files got left out
from the cleaned stats that
webpack served


caalador:
Can be tested against Spring test module `test-spring-white-list`

mshabarov:
Tested with `test-spring-white-list`: fails without this patch, and work well with it.

mshabarov (approved):

Issue #14797, opened 2022-10-11
Page constantly reloads by itself after a websocket connection fails


echarlus:
### Description of the bug

On some occasions, the browser will start reloading the current page forever. 
Looking at Chrome console one can see js exceptions. 
Looking at the network activity, one can also see that there is a recurrent POST  being made to the server and it results in an exception on the client. 
The only way to get out of this behavior is to force reload the whole page.

### Expected behavior

Page would not be reloaded forever ....

### Minimal reproducible example

Unfortunately  as this is fairly random I do not have a sample to provide. 
Here's a screenshot of the console (the image load which results in a 404 is "normal", it's probably when the page is being reloaded that it tries to refetch this resource that does not exist for this particular case).

![image](https://user-images.githubusercontent.com/7509875/195004108-600ec29e-73df-4fea-8764-a239a3788020.png)

[logs.zip](https://github.com/vaadin/flow/files/9751396/logs.zip)

Here are logs from the client (chrome console + chrome network activity) and the corresponding tomcat logs on the server side taken when the issue occurred.

### Versions

- Vaadin 23.2.3 with flow 23.2.4 (as I had also faced issue https://github.com/vaadin/flow/issues/14232)
- Java version: 11
- OS version:  Mac OS 12.6  
- Browser version (if applicable): Chrome 106.0.5249.103
- Application Server (if applicable): Linux with Tomcat 9.0.65 on Java openjdk version "11.0.16.1" 2022-08-12
- IDE (if applicable): IntelliJ IDEA 2022.2.2

caalador:
Would it be possible to strip away everything from the app so that it still fails with the issue?
From this it would seem like something goes wrong after the 404 for the image.

echarlus:
The 404 for the image is not from the app itself since it's a direct URL to the resource embedded in the war file : /img/areas/logo_21.png for example. How could this affect the behavior of the app ? I was assuming this type of query is handled directly by the browser and the app server, without any vaadin interaction. Am I wrong ?

caalador:
Mainly because I heard of another project where a resync happens and there I heard it also seems to happen after a failing image download. Though that seems to do a `Gave up waiting for message {} from the server` then resync.

But without a project to test this is a guessing game with not good guesses.

echarlus:
[...]

PR #15294, opened 2022-11-28
chore: add call thread debug logging

caalador:
Add debug logging for incrementing
server id to be able to see what is
calling it during runtime.

Targets #14797

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -18,6 +18,7 @@ package com.vaadin.flow.component.internal;
 import java.io.Serializable;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -316,6 +318,10 @@ public int getServerSyncId() {
      */
     public void incrementServerId() {
         serverSyncId++;
+        getLogger().debug("Increment syncId:\n{}",
```
knoobie, line 321: Should this be guarded with a isDebugEnabled call? This looks kinda intensive logging

mcollovati: If I'm not mistaken, SLF4J already has the guard internally.
However, we can use the `debug` overload that takes a `Supplier` for message arguments, so the stack computation will be performed only if needed

knoobie: That is correct but the object / stack / stream has to be created anyway if I remember correctly. Sadly the overloaded method with Supply is only available in 2.0 or was this backported to 1.0 as well? Otherwise this would also work!

mcollovati: @knoobie You're right, my mistake. It is a 2.x feature

caalador: :+1:

mcollovati (approved):

PR #15315, opened 2022-11-30
fix: Prefix web component client app with wc-

### LINKED ISSUE #15263: Dynamic loading of a web component fails with 'Error Application ROOT-2521314 already initialized'
### Description of the bug

I've implemented dynamic loading of a Vaadin exported web component inside a Vaadin 23.2 view, the respective commit is here: https://gitlab.com/mrts/vaadin-embedding-example-guestbook/-/commit/b091d08c1038264f25b6f2c725652297852494ce.

The custom element is loaded and registered successfully, but rendering fails with the error `Application ROOT-2521314 already initialized`:
```
Loading custom element guest-book   dynamic-webcomponent-loader.ts:37:16
registering <guest-book> using id GuestBook-0   guest-book.js:112:13
Processing time was 2ms   FlowClient.js:205:19
Uncaught Error: Application ROOT-2521314 already initialized
    initApplication   FlowBootstrap.js:93
```

### Expected behavior

When I open http://localhost:8080/dynamic-loader and click the _Load web component_ button, then the web component is loaded and displayed on the page, the same way as in http://localhost:8080/basic.html.

See https://gitlab.com/mrts/vaadin-embedding-example-guestbook.

Run the project using `mvn jetty:run` and open http://localhost:8080/dynamic-loader in the browser to test.

### Versions

- Vaadin / Flow version: 23.2.9
- Java version: 11
- OS version: Ubuntu 22.04
- Browser version (if applicable): Firefox 107
- Application Server (if applicable): Jetty

Artur-:
Avoids conflicts with the application

Fixes #15263

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -223,6 +223,8 @@ public class UI extends Component
      *            the initialization request
      * @param uiId
      *            the id of the new ui
+     * @param appId
+     *            the application id
      *
      * @see #getUIId()
      */
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -206,8 +206,27 @@ public class UI extends Component
      *            the id of the new ui
      *
      * @see #getUIId()
+     * @deprecated Use {@link #doInit(VaadinRequest, int, String)} instead
      */
+    @Deprecated
     public void doInit(VaadinRequest request, int uiId) {
+        doInit(request, uiId,
+                getSession().getService().getMainDivId(getSession(), request));
+    }
+
+    /**
+     * Internal initialization method, should not be overridden. This method is
+     * not declared as final because that would break compatibility with e.g.
+     * CDI.
+     *
+     * @param request
+     *            the initialization request
+     * @param uiId
+     *            the id of the new ui
+     *
+     * @see #getUIId()
+     */
+    public void doInit(VaadinRequest request, int uiId, String appId) {
         if (this.uiId != -1) {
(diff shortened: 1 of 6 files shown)


mcollovati (commented):
LGTM.
I just wonder if we need to add tests for the `isBootstrapRequest` method since it is public

Artur-:
I would rather have wanted to change app id in WebComponentBootstrapContext and create no API but it seems that the context method is used in one place and the servlet app id method is used directly in doInit for the UI. I wonder if passing the context or appId to doInit would be better..

mcollovati:
Compute the appId in the bootstrap context makes sense indeed.
Refactoring `doInit` will be a breaking change, but I suppose it is not a common pattern to subclass UI and override the `doInit` method, so the impact of the change may be low.

Artur-:
Added a second `doInit` instead

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -206,8 +206,27 @@ public int getUIId() {
...
+    /**
+     * Internal initialization method, should not be overridden. This method is
+     * not declared as final because that would break compatibility with e.g.
+     * CDI.
+     *
+     * @param request
+     *            the initialization request
+     * @param uiId
+     *            the id of the new ui
+     *
```
mcollovati, line 226: missing javadoc for `appId` param

mcollovati:
Perhaps also this test method in `flow-server/src/test/java/com/vaadin/flow/component/UITest.java` should be refactored

```java
    private static UI createTestUI() {
        MockLogger mockLogger = new MockLogger();
        UI ui = new UI() {
            @Override
            public void doInit(VaadinRequest request, int uiId) {

            }
        };
        return ui;
    }
```

mcollovati (approved):

PR #15324, opened 2022-11-30
feat: add refresh flag to navigation event

### LINKED ISSUE #14999: PiT 23.3: PreserveOnRefresh does not work in master-detail views
### Description of the bug

When adding the `@PreserveOnRefresh` to a view that loads an entity based on the route path, unsaved changes are not correctly restored.

Check the following animation to see that it works with new entities, but it does not work when the entity is edited.

preserve-on-refressh-1

### Workaround

Use some mechanism to prevent loading the entity e.g.

```java
    @Override
    	if (binder.hasChanges()) {
    		return;
    	}
```

### Solution

When `@PreserveOnRefresh` is set, `BeforeEnterEvent` should not be called on reload. 
Based on the [documentation](https://vaadin.com/docs/latest/advanced/preserving-state-on-refresh/#preconditions-and-limitations) only `AttachEvent` and `DetachEvent` should only be generated

### Expected behavior

`PreserveOnRefresh` should work in both cases like in this screenshot:

preserve-on-refressh-2

- Download a project with the master-detail view from start website, and set current version
```bash
curl -s -f 'https://start.vaadin.com/dl?&preset=master-detail-view&projectName=test-app' -o test-app.zip 
unzip test-app.zip 
cd test-app
mvn -q versions:set-property -Dproperty=vaadin.version -DnewVersion=23.3.0.alpha1
```
- Add `@PreserveOnRefresh` annotation to the `MasterDetailJavaView.java` view
```bash
cat <<EOF |patch -p1
--- a/src/main/java/com/example/application/views/masterdetailjava/MasterDetailJavaView.java
+++ b/src/main/java/com/example/application/views/masterdetailjava/MasterDetailJavaView.java
@@ -31,11 +31,13 @@ import java.util.Optional;
 import java.util.UUID;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.domain.PageRequest;
+import com.vaadin.flow.router.PreserveOnRefresh;

 @PageTitle("Master-Detail (Java)")
 @Route(value = "master-detail-view/:samplePersonID?/:action?(edit)", layout = MainLayout.class)
 @Uses(Icon.class)
+@PreserveOnRefresh
 public class MasterDetailJavaView extends Div implements BeforeEnterObserver {

     private final String SAMPLEPERSON_ID = "samplePersonID";
EOF
```

### Versions

- Vaadin / Flow version: `23.3.0.alpha1` and previous
- Java version: 11
- OS version: mac
- Browser version: chrome

caalador:
Add the isRefreshEvent to BeforeEnterEvent
and AfterNavigationEvent to be make it possible
to diustinguish if the event is for a refresh of
a preserve on refresh view.

Fixes #14999

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/AfterNavigationEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/AfterNavigationEvent.java
@@ -30,7 +30,6 @@ import com.vaadin.flow.component.HasElement;
 public class AfterNavigationEvent extends EventObject {
 
     private final LocationChangeEvent event;
-    private final boolean refreshEvent;
 
     /**
      * Construct event from a NavigationEvent.
@@ -41,22 +40,6 @@ public class AfterNavigationEvent extends EventObject {
     public AfterNavigationEvent(LocationChangeEvent event) {
         super(event.getSource());
         this.event = event;
-        this.refreshEvent = false;
-    }
-
-    /**
-     * Construct event from a NavigationEvent.
-     *
-     * @param event
-     *            NavigationEvent that is on going
-     * @param refreshEvent
-     *            if event for refresh of a preserve on refresh view
-     */
-    public AfterNavigationEvent(LocationChangeEvent event,
-            boolean refreshEvent) {
-        super(event.getSource());
-        this.event = event;
-        this.refreshEvent = refreshEvent;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/router/AfterNavigationEvent.java
```diff
@@ -30,6 +30,7 @@
 public class AfterNavigationEvent extends EventObject {
 
     private final LocationChangeEvent event;
+    private final boolean refreshEvent;
```
knoobie, line 33: Instead of adding another type, could / should the NavigationTrigge be extended with a refresh enum and used?

caalador: done.

mcollovati (approved):

PR #15327, opened 2022-11-30
refactor: Merge constructors and use setters instead

Artur-:


caalador (approved):
Only question is with this can you mess things up so that something that should have been set is not set?

caalador (approved):
Yes. This way it feels a bit less error prone as the rest should be generatable with the project root directory set.

PR #15343, opened 2022-12-02
test: Always run postinstall for esbuild in flow-client

Artur-:
Makes the build work even when you are using --ignore-scripts

### DIFF
--- a/flow-client/pom.xml
+++ b/flow-client/pom.xml
@@ -349,6 +349,19 @@
                         </configuration>
                     </execution>
                     <execution>
+                        <id>install-esbuild</id>
+                        <phase>validate</phase>
+                        <goals>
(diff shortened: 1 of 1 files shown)


flow-client/pom.xml
```diff
@@ -348,6 +348,19 @@
...
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>install-esbuild</id>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <executable>node</executable>
```
caalador, line 358: Does this work without a global node installed?

Artur-: I would imagine it works just as well as the "npm install" exec above

PR #15351, opened 2022-12-02
fix: Use root servlet mapping for push url in OSGi case

mshabarov:
Fallback to a root mapping for push in case if not servlet registration is available (OSGi runtime).

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
@@ -231,7 +231,7 @@ public class PushRequestHandler
                     findFirstUrlMapping(servletRegistration.get())
                             + Constants.PUSH_MAPPING);
         } else {
-            getLogger().warn(
+            getLogger().debug(
                     "Unable to determine servlet registration for {}. "
                             + "Using root mapping for push",
                     vaadinServletConfig.getServletName());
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/PushRequestHandler.java
```diff
@@ -223,9 +224,18 @@ public AtmosphereFramework addInitParameter(String name,
...
-                findFirstUrlMapping(vaadinServletConfig)
-                        + Constants.PUSH_MAPPING);
+        Optional<ServletRegistration> servletRegistration = getServletRegistration(
+                vaadinServletConfig);
+        if (servletRegistration.isPresent()) {
+            atmosphere.addInitParameter(ApplicationConfig.JSR356_MAPPING_PATH,
+                    findFirstUrlMapping(servletRegistration.get())
+                            + Constants.PUSH_MAPPING);
+        } else {
+            getLogger().warn(
```
knoobie, line 234: Should this be info? Now every start of an OSGI deployment would log this warning and the developer has no way fix 😅

mshabarov: Good point! Indeed, info or even a debug level would be better. Thanks!

mcollovati (approved):

PR #15353, opened 2022-12-04
chore: Write package.json hash, npm module versions and entry points to stats.json

Artur-:


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -5,7 +5,7 @@
  * This file will be overwritten on every run. Any custom changes should be made to vite.config.ts
  */
 import path from 'path';
-import { readFileSync, existsSync, writeFileSync, mkdirSync } from 'fs';
+import { readFileSync, existsSync, writeFileSync, mkdirSync, fstat } from 'fs';
```
caalador, line 8: `fstat` is not used anywhere.

I guess the idea is to use it to check file age in a later step?

Artur-: Removed, it was probably auto imported for some reason

mshabarov (approved):

PR #15381, opened 2022-12-06
test: ensure ComponentTest preconditions (CP: 23.3)

mcollovati:
ComponentTest.getComponentLocale_noCurrentUI_returnsDefaultLocale assumes that I18nProvider is null but it may have been initialized by another test. This change makes the Instantiator instance a mockito spy so that it can be instructed by the tests to return a null I18nProvider instance.


caalador (approved):
This should have auto cherrypicked fine to the older branches, as the original change did also.

PR #15392, opened 2022-12-07
feat: Add the Authentication Context bean to the Vaadin Web Security (#15129) (CP: 23.3)

### LINKED ISSUE #14958: Authentication API to get the current user and perform logout

mcollovati:
- AuthenticationContext is a concrete class that gets access to the authenticated user and allows performing logout integrated with Spring Security
- AuthenticationContext is an injectable managed bean
- Developers can plug additional LogoutHandlers in VaadinWebSecurity.addLogoutHandlers()
- VaadinWebSecurity configures a LogoutSuccessHandler with a redirect strategy that can handle also XHR UIDL requests (UidlRedirectStrategy)

Fixes #14958

### DIFF
--- a/flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/SecurityConfig.java
+++ b/flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/SecurityConfig.java
@@ -15,6 +15,9 @@ import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
 import org.springframework.security.provisioning.InMemoryUserDetailsManager;
 
+import com.vaadin.flow.component.UI;
+import com.vaadin.flow.internal.UrlUtil;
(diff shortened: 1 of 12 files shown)


mcollovati:
`UidlRedirectStrategy` required changes because in 23.3 we have a dependency on Servlet 3.1 so the `HttpServletRequest.getHttpServletMapping()` method is not available. It has been replaced with the internal `RequestUtil` component.

PR #15393, opened 2022-12-07
fix: npm version collection on windows

caalador:
Fix collecting versions for npm
modules in windows environment.


Artur- (approved):
oh joy

Issue #15375, opened 2022-12-05
Add a feature flag for Express Build


mshabarov:
Flow needs a feature flag that would be enabled just to be able to set the `enableDevMode` config parameter to `false`. Just to make sure the user knows that the Express Build feature is not stable yet.

PR #15446, opened 2022-12-12
feat: Add a feature flag for Express Build

mshabarov:
Adds a feature flag for Express Build and disables dev server forcibly if this feature flag is set.

Fixes https://github.com/vaadin/flow/issues/15375


flow-server/src/main/java/com/vaadin/flow/server/DefaultDeploymentConfiguration.java
```diff
@@ -404,4 +411,18 @@ private void checkSendUrlsAsParameters() {
                 DEFAULT_SEND_URLS_AS_PARAMETERS);
     }
 
+    private void checkEnableDevServer(boolean log) {
```
Artur-, line 414: Do we need special handling for dev server in two classes? Can't we just make it in the parent class?

Artur-: Shouldn't `enableDevServer`  always have the value of !featureFlagEnabled for now?

Artur-: Overriding the value makes no sense. It is always global, never per servlet

mshabarov: Yes, makes sense. Fixed.

PR #15466, opened 2022-12-13
test: Change push tests to not be UI classes

### LINKED ISSUE #15176: Tests in root-context-ui should not depend on V14 bootstrap mode
### Description of the bug

There is a comment in https://github.com/vaadin/flow/blob/f49ca8024d9efd73b902051600bf41e69ebed486/flow-tests/test-root-ui-context/pom.xml#L14-L15 indicating that the tests are running in v14 bootstrap mode because of #7559, which does not seem to be accurate.

When running the tests without the flag in #15171 at least the following tests fail
- LockingIT .testLockingTheUIFor4HeartBeats
- BasicPollIT .testServerInitiatedCommunication
- NoRouterIT. applicationShouldStart

At least some more tests fail like `PushWithRequireJSIT` which does not actually test that it tests anything (it checks the 404 page for javascript errors in the PR)

### Expected behavior

All tests should be run without the deprecated mode. The actual test for the deprecated mode are removed in #15172

-

### Versions

-

Artur-:
Closes #15176


flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/push/BasicPushWebsocketXhrView.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.vaadin.flow.component.AttachEvent;
+import com.vaadin.flow.internal.nodefeature.PushConfigurationMap;
+import com.vaadin.flow.router.Route;
+import com.vaadin.flow.shared.ui.Transport;
+import com.vaadin.flow.uitest.servlet.ViewTestLayout;
+
+/*
+ * Note that @Push is generally not supported in this location, but instead
+ * explicitly picked up by logic in the BasicPushUI constructor.
```
caalador, line 11: Comment is talking about `@Push` and `BasicPushUI`. 
Could be good to update this so it's not that confusing in the future when making changes.

This comment is on all BasicPush* classes.

caalador (dismissed):
Looks good, apart from the comments

caalador:
Also the whole `test-root-ui-context` could be removed after mering master as nothing will be left in that module.

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/push/BasicPushWebsocketXhrView.java
Artur-: Removed

Artur-:
> Also the whole `test-root-ui-context` could be removed after mering master as nothing will be left in that module.

This is done in a separate PR as it dependend on another change also

caalador (approved):

PR #15487, opened 2022-12-14
feat: added method to get authenticate principal name

### LINKED ISSUE #15484: Provide the user name through AuthenticationContext
### Describe your motivation

Fetching the details of a user is most often based on the username. When logging in, the username passed to a `UserDetailsService` implementation is defined as
```
	private String determineUsername(Authentication authentication) {
		return (authentication.getPrincipal() == null) ? "NONE_PROVIDED" : authentication.getName();
	}
```

In `AuthenticationContext` there is only a private `getAuthentication` and then a `getAuthenticatedUser` which assumes you want the principal from the authentication and assumes that you know the type of the principal.

However when you want to fetch some kind of `User` object from the database, like the `UserDetailsService` implementation, you need the user name and do not need to know the details about the principal.

`AuthenticationContext.getUsername()` if we can assume a user name is always available. Otherwise making `Optional<Authentication> getAuthentication` public.

### LINKED ISSUE #15485: Stricter TS checks in Vaadin 23.3.0
### Description of the bug

Vaadin 23.3.0 seems to introduce stricter TS checks by default.

I had to configure the following in `tsconfig.json` in order to run applications that were working fine with Vaadin 23.2:
```
"noImplicitAny": false,
"noImplicitThis": false,
"strictPropertyInitialization": false
```

(`noImplicitAny` and `noImplicitThis` were already `true` with Vaadin 23.2, but now they flag additional errors.)

See some sample error logs in https://github.com/FlowingCode/FontAwesomeIronIconset/issues/56

The new stricter defaults are problematic, since they also affect third-party components (addons), which may rely on loose checks.

### Expected behavior

Since it's not documented as a breaking change, I would expect no additional TS errors when upgrading from Vaadin 23.2 to Vaadin 23.3.0.

See https://github.com/FlowingCode/FontAwesomeIronIconset/tree/font-awesome-iron-iconset-4.2.0 (fixed in https://github.com/FlowingCode/FontAwesomeIronIconset/pull/58)
and https://github.com/FlowingCode/XTermConsoleAddon/commits/xterm-console-2.1.0 (fixed in https://github.com/FlowingCode/XTermConsoleAddon/pull/66)

The application runs with no issues up to vaadin 23.2 (`mvn jetty:run`), but it fails with TS errors with vaadin 23.3 and 24.

### Versions

- Vaadin / Flow version: 23.3.0, 24.0.0.alpha5
- Java version: 11, 17
- OS version: Windows 10

mcollovati:
## Description

Principal name usually refers to a user identifier that can be used to retrieve additional details.

Fixes #15484 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
@@ -17,6 +17,7 @@ package com.vaadin.flow.spring.security;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.security.Principal;
 import java.util.List;
 import java.util.Optional;
 
@@ -77,6 +78,23 @@ public class AuthenticationContext implements Serializable {
     public <U> Optional<U> getAuthenticatedUser(Class<U> userType) {
         return getAuthentication().map(Authentication::getPrincipal)
                 .map(userType::cast);
+    }
+
+    /**
+     * Gets an {@link Optional} containing of the authenticated principal name
+     * as defined in {@link Principal#getName()}, or empty if the user is not
+     * authenticated.
+     *
+     * The principal name usually refers to a username or an identifier that can
+     * be used to retrieve additional information for the authenticated user.
+     *
+     * Anonymous users are considered not authenticated.
+     *
+     * @return an {@link Optional} containing of the authenticated principal
+     *         name or empty if not available.
+     */
+    public Optional<String> getPrincipalName() {
+        return getAuthentication().map(Principal::getName);
     }
 
     /**
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
@@ -89,8 +89,8 @@ public class AuthenticationContext implements Serializable {
      *
      * Anonymous users are considered not authenticated.
      *
-     * @return an {@link Optional} containing of the authenticated principal
-     *         name or empty if not available.
+     * @return an {@link Optional} containing the authenticated principal name
+     *         or an empty optional if not available.
      */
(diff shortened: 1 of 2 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
```diff
@@ -79,6 +80,23 @@ public <U> Optional<U> getAuthenticatedUser(Class<U> userType) {
                 .map(userType::cast);
     }
 
+    /**
+     * Gets an {@link Optional} containing of the authenticated principal name
+     * as defined in {@link Principal#getName()}, or empty if the user is not
+     * authenticated.
```
heruan, line 86: Sounds links there's a "name" or something missing here. Also, I'm not sure about mentioning the `Principal` interface at all, since it's not something we return anywhere 🤔

Artur- (dismissed):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
```diff
@@ -79,6 +80,23 @@ public <U> Optional<U> getAuthenticatedUser(Class<U> userType) {
                 .map(userType::cast);
     }
 
+    /**
+     * Gets an {@link Optional} containing of the authenticated principal name
+     * as defined in {@link Principal#getName()}, or empty if the user is not
```
Artur-, line 85:
```suggestion
     * as defined in {@link Principal#getName()}, or an empty optional if the user is not
```

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
mcollovati: It is not a "name" missing, but an additional useless "of" :smile:

Artur- (approved):

PR #15497, opened 2022-12-15
fix: remove Serializable interface from AuthenticationContext

mcollovati:
The AuthenticationContext component is Serializable because the idea was to potentially inject it into Vaadin views, but this may cause several issues:

* a single instance of that class per application is expected, but if deserializating views we may end up with many of them
* logoutSuccessHandler and logoutHandler are transient, but they are not Spring beans, and we do not provide a simple way to inject them upon deserialization
* for the above reasons, AuthenticationContext may not work correctly with Kubernetes Kit

This change removes the implementation of Serializable interface and updates javadocs.

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
@@ -46,7 +46,7 @@ import com.vaadin.flow.server.VaadinServletResponse;
  *
  * An instance of this class is available for injection as bean in view and
  * layout classes. The class is not {@link java.io.Serializable}, so potential
- * referencing fields in Vaadin view should be defined {@literal transient}.
+ * referencing fields in Vaadin views should be defined {@literal transient}.
  *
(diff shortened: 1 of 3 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
```diff
@@ -40,23 +40,25 @@
...
  * <p>
- * An instance of this class is available for injection as bean in view and
- * layout classes.
  *
  * It allows to access authenticated user information and to initiate the logout
  * process.
  *
+ * An instance of this class is available for injection as bean in view and
+ * layout classes. The class is not {@link java.io.Serializable}, so potential
+ * referencing fields in Vaadin view should be defined {@literal transient}.
```
heruan, line 49:
```suggestion
 * referencing fields in Vaadin views should be defined {@literal transient}.
```

heruan (dismissed):
Just a small typo to fix, otherwise good for me!

mshabarov (approved):

Issue #15376, opened 2022-12-05
Improve the logic that decides whether to re-compile / re-build the bundle


mshabarov:
The logic that checks if a default bundle or a new custom bundle is needed is not good enough. It only checks the `package.json` hash https://github.com/vaadin/flow/blob/feature/nonodemode/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java#L131. 

It should check:
- project frontend files through hashes or similar
- npm dependencies and their versions
- ~bundle's entry points~
- package.json hash

caalador:
Done is npm dependencies with versions and package hash.
Next would be frontend files and after that changed frontend files.

caalador:
Validation against in-memory package.json and flow-generated-imports.js imports added.

caalador:
Closing as done. Any missing things or updates to logic should get a new issue.

PR #15504, opened 2022-12-15
feat: Compare bundle and package.json contents

caalador:
Compare the stored hash of the bundle and package.json.
Compare the dependency modules in bundle and package.json.
Compate NpmPackage annotations and bundle modules.

part of #15376


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -102,23 +102,24 @@ public NodeTasks(Options options) {
...
-        }
-
         if (options.enablePackagesUpdate || options.enableImportsUpdate
                 || options.enableWebpackConfigUpdate) {
             frontendDependencies = new FrontendDependenciesScanner.FrontendDependenciesScannerFactory()
                     .createScanner(!options.useByteCodeScanner, classFinder,
                             options.generateEmbeddableWebComponents,
                             options.useLegacyV14Bootstrap, featureFlags);
 
+            if (!options.productionMode && options.isDevBundleBuild()) {
```
Artur-, line 112: This could have a comment why it is exaclty in this position

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -82,10 +89,159 @@ public static boolean needsBuild(File npmFolder) {
...
 
-        return !new File(npmFolder, "dev-bundle").exists();
+        if (!FrontendUtils.getDevBundleFolder(npmFolder).exists()) {
+            // TODO: check for jar dev-bundle
+            return true;
+        }
+
+        String statsJsonContent = FrontendUtils.findBundleStatsJson(npmFolder);
+        if (statsJsonContent == null) {
+            // without stats.json in bundle we can not say if it is up to date
```
Artur-, line 103: You can never get here as the method throws an exception. Maybe it should not

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -82,10 +89,159 @@ public static boolean needsBuild(File npmFolder) {
...
+     * @return {@code true} if everything matches
+     */
+    private static boolean hasAllNpmModules(JsonObject statsJson,
+            JsonObject packageJson, Map<String, String> npmPackages) {
+
+        JsonObject bundleNpmModules = statsJson.getObject("npmModules");
+        JsonObject dependencies = packageJson.getObject("dependencies");
+
+        for (String dependency : dependencies.keys()) {
+            if (!bundleNpmModules.hasKey(dependency)) {
```
Artur-, line 214: Is version comparison for another PR?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -82,10 +89,159 @@ public static boolean needsBuild(File npmFolder) {
...
+                .getPackages();
+
+        if (!hashAndBundleModulesEqual(statsJson, packageJson, npmPackages)) {
+            // Hash in the project doesn't matche the bundle hash or NpmPackages
+            // are
+            // missing found.
+            return true;
+        }
+
+        if (!hasAllNpmModules(statsJson, packageJson, npmPackages)) {
```
Artur-, line 121: Cannot `hasAllNpmModules` and `bundleModulesEqual` be combined into one?

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
caalador: Combined checks

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -82,10 +89,154 @@ public static boolean needsBuild(File npmFolder) {
...
+            if (missingNpmPackages(npmPackages, bundleModules)) {
+                return false;
+            }
+        }
+
+        JsonObject dependencies = packageJson.getObject("dependencies");
+
+        List<String> collection = Arrays.stream(dependencies.keys())
+                .filter(pkg -> !dependencies.hasKey(pkg))
+                .collect(Collectors.toList());
```
mshabarov, line 159: This checks `dependencies` against `dependencies`, which gives an empty collection always. Should it check `bundleModules` agains `dependencies`?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -82,10 +89,154 @@ public static boolean needsBuild(File npmFolder) {
...
+        if (!collection.isEmpty()) {
+            for (String dependency : collection) {
+                getLogger().info("Dependency " + dependency
+                        + " is missing from the bundle");
+            }
+            return false;
+        }
+
+        // We know here that all dependencies exist
+        collection = Arrays.stream(dependencies.keys())
```
mshabarov, line 170: Can we merge these two checks (dependencies exists and version match) into one loop, so it would take less time?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -82,10 +89,154 @@ public static boolean needsBuild(File npmFolder) {
...
+     *            npm packages map
+     * @return {@code true} if up to date
+     */
+    private static boolean hashAndBundleModulesEqual(JsonObject statsJson,
+            JsonObject packageJson, Map<String, String> npmPackages) {
+
+        String packageJsonHash = getPackageJsonHash(packageJson);
+        String bundlePackageJsonHash = getStatsHash(statsJson);
+
+        if (packageJsonHash == null || packageJsonHash.isEmpty()) {
```
mshabarov, line 142: We can put a warning message here, because this is something unexpected if package.json doesn't have a hash.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
caalador: Well this choice was more of a being able to give a good exception message. Either it is missing or then the version is wrong. Also how big are the sets normally?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
caalador: fixed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
caalador: added

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov: Hundreds of items I guess, which is nothing basically. Ok, let's keep it as is, ignore me :)

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov: Do the tests need a tweak because of that ? :)

caalador: The test: `hashesMatch_statsMissingPackageJsonPackage_compilationRequired` tests that and the original mistake should have failed the test because the next part throws `java.lang.NullPointerException: Cannot invoke "elemental.json.JsonString.getString()" because the return value of "elemental.json.impl.JreJsonObject.get(String)" is null` Why the test passes even though something is thrown...

caalador: Oh. needsBuild eats exceptions and only logs them and says do rebuild

caalador: change test to use the internal method that doesn't catch

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -82,10 +89,155 @@ public static boolean needsBuild(File npmFolder) {
...
+        // Check that bundle modules contains all package dependencies
+        if (packageJsonHash.equals(bundlePackageJsonHash)) {
+            if (missingNpmPackages(npmPackages, bundleModules)) {
+                return false;
+            }
+        }
+
+        JsonObject dependencies = packageJson.getObject("dependencies");
+
+        List<String> collection = Arrays.stream(dependencies.keys())
```
Artur-, line 158:
```suggestion
        List<String> missingFromBundle = Arrays.stream(dependencies.keys())
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -82,10 +89,155 @@ public static boolean needsBuild(File npmFolder) {
...
+                getLogger().info("Dependency " + dependency
+                        + " is missing from the bundle");
+            }
+            return false;
+        }
+
+        // We know here that all dependencies exist
+        collection = Arrays.stream(dependencies.keys())
+                .filter(pkg -> !dependencies.getString(pkg)
+                        .equals(bundleModules.getString(pkg)))
```
Artur-, line 173: Won't an equals check always fail for range versions like `^1` or `~2.3.4` ?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -82,10 +89,155 @@ public static boolean needsBuild(File npmFolder) {
...
+    /**
+     * Check that all npmPackages are in the given dependency object.
+     *
+     * @param npmPackages
+     *            {@code @NpmPackage} key-value map
+     * @param dependencies
+     *            json object containing dependencies to check against
+     * @return {@code false} if all packages are found
+     */
+    private static boolean missingNpmPackages(Map<String, String> npmPackages,
```
Artur-, line 229:
```suggestion
    private static boolean dependenciesContainsAllPackages(Map<String, String> npmPackages,
```
+ inverse logic to be more readable

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
caalador: Yes. Didn't think about that as we always use specific versions. I'll update it to use FrontendVersion.

caalador: Now this is a can of worms for the acceptance.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
caalador: done,

mshabarov (approved):

PR #15511, opened 2022-12-15
feat: Deprecate getElement() and update javadoc

mstahv:
It has been seen many times new Vaadin users end up calling this method in their apps, even though they most probably should not do that (unless writing extensions/components).

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


mstahv:
@knoobie proposed in Discord to introduced getInternalElement or similar, that by naming convention accomplishes the same thing I'm trying in this PR with actual deprecation. Investigated that a bit, but doing that properly would need a HUGE PR to flow, flow-components etc.

flow-server/src/main/java/com/vaadin/flow/component/Component.java
```diff
@@ -194,14 +194,28 @@ private void mapToElement(String tagName) {
...
      * attached to a parent component, this element is attached to the parent
      * component's element hierarchy.
      *
      * @return the root element of this component
+     * @deprecated This is deprecated as you should never call this method in
+     * your Vaadin application code. This not planned to be removed and calling
+     * this method is the way to go when creating components/extension to be
+     * used in your application or to be shared as add-on components.
      */
+    @Deprecated
```
knoobie, line 218: Like discussed on discord: I'm against adding the Deprecation annotation to this. I'm on the other hand totally fine with a big notice in the javadoc that questions the usage of this method in the developer's code. Especially advanced use-cases or not officially supported Html attributes has to be added like this and there is not other way to add them. Like for example adding aria-* attributes on components. 

Having a Deprecation warning on all of those legitimate usages would result in hundreds of even thousands of Deprecation warnings in logs if the maven compiler is configured to show them or would even fail the build with strict policies. Let alone the noise this would create in the flow-components where sonarqube would literally flag 95% of the code with https://rules.sonarsource.com/java/RSPEC-1874. 

Personally I would only approve the Deprecation if there is a way for the developer to get rid of the warnings by for example migrating to other APIs (which isn't possible always) or add a new method "getInternalElement()" or something similar that suggests that this is advanved, improving the Usability of junior developer that stumble upon this method and using it without thinking about alternatives.

knoobie: How I'm using it for: 
- customizing my lit templates from java with properties and attributes 
- Using it to create low level components like a native Login Form
- adding attributes that aren't first-level important like title, aria-*, role and other advanced or newer html attributes 
- Calling executeJs on it 
- Accessing the theme list 
- Accessing the class list 
- Working with non-components that are just elements on the server side 
- Working with virtual childrens

knoobie: Another example would be the https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/router/RouterLayout.java which forces the developer to work on Element level of they wanna add their own RouterLayout, which is quite often the case if no AppLayout is used.

mstahv: There is probably going to be a lot of comments in this PR :-)

One thing I want to note, before going to other details: deprecation does not mean the API should not ever be used nor it will be removed. Also for example if "its usage is likely to lead to errors". Also there is `@SuppressWarnings("deprecation")` annotation that one can and should use if they have know what they are doing and have read the disclaimers. https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Deprecated.html

Some of your points in the list look very valid usage of getElement(), even if it was deprecated. Some might be signs of missing framework/component features. But we need to analyse your list well (and hope to get more details e.g. via Discord tomorrow), but couple of quick comments before I call this a day:

 * Seeing so many times that even experienced Vaadin developers call executeJs directly from their UI logic is one of the reasons I'm advocating for this change. I may be a bit purist, but I think all those should be hidden to to components internals (preferably to a separate class, but at least to a private method at the end of the class). *The abstraction is broken if you see JS in Vaadin UI logic*.
 * I'd claim that RouterLayout should use Component instead of HasElement. In fact, I think the whole HasElement should be burnt with 🔥Don't know when something else but Component would be HasElement and that already has getElement overridden (because apparently nobody came up with anything reasonable to put into the actual interface javadoc).

knoobie: > SuppressWarnings

Sadly I know about this, I'm already quite heavily using it for some not-so-nice stuff. But would like to reduce it's usage, instead of e.g. adding it to the whole class using multiple getElement calls. 😅 

Another alternative could be to add something like https://github.com/apiguardian-team/apiguardian but I'm not sure if it's worth it just for one place. Junit5 for example is using this quite excessive, making really good in-code documentation about the usage of methods. 


> The abstraction is broken if you see JS in Vaadin UI logic.

I'm hating Javascript myself as well, but sadly there are things like focus management for accessibility I have to do like this because for example H1 or other html components are not focusable by default but with tabindex=-1 I can get them focused via JS and improve the navigation for screen reader user drastically in "SPAs" like Vaadin Flow.

samie: I also see this more from a "pedagogic PoV": Abstractions serve a purpose of not needing to know/learn everything at once. Still we should not make anything impossible to access, but I'm still advocating the abstraction levels. In this case: if something feels like a "component internal implementation" it should be like that - also in the application code base. 

We have three logical levels of control we can leverage:
- For core team: Component implementations (and framework itself)
- For heavy users: Inherited custom components (application frameworks and add-ons)
- For everyone: Application code 

Generally speaking we want less application code without losing the low-level control. Now the element API is all the control. Handy, but also all over in the application codebase. And we all know that will just make the app maintenance harder. 

Not sure if this is technically doable like this, but removing getElement completely from public API would force inheritance, if you need to access it, but isn't that also closer to the component abstraction idea and separating JS hacks from maintained application UI logic.

Legioth: I would say that if it's problematic that developers unfamiliar with Flow try to use `getElement()` for the wrong purpose, then we should discourage accidental use by making it slightly less discoverable rather than through deprecating without replacement. For lower discoverability, I would suggest something like `Element.of(component)` as the main API.

samie: I was first thinking simply a protected method in Vaadin classes available for subclasses (also inline inheritance). `Element.of` sounds like a good candidate too. API consistency is one thing and if that is technically easier to implement. And as long as it is easy to spot that _"now you are doing something you should not do in your normal application UI code"_

How about making a quick prototype how would the application code look like in some of the use cases @knoobie presented above? I think some of them should be done on the client-side modules instead, but maybe e.g. these: 
- Calling executeJs on it (last fallback on many integration things, should be inside components or utility)
- Accessing the theme list (should be a proper component API?)
- Accessing the class list (logical states in application code, component API?)

mvysny: Deprecating `getElement()` and providing no replacement API will cause all reusable components fiddling with client-side to have inevitable deprecation warnings during compilation. That doesn't sound right to me. However, it's okay if there's `Element.of` replacement API.

Other than that, I agree that `HasElement` should be used as sparingly as possible: we should use `Component` whenever possible.

* Deprecating `getElement()` without replacement API - not okay for me
* Deprecating `getElement()` with `Element.of()` - okay for me
* Marking `getElement()` as INTERNAL in javadoc - okay for me

simasch: **Why I don't like the change**

IMO @Deprecated is incorrect in this situation. 

> A program element annotated @Deprecated is one that programmers are discouraged from using. An element may be deprecated for any of several reasons, for example, its usage is likely to lead to errors; it may be changed incompatibly or removed in a future version; it has been superseded by a newer, usually preferable alternative; or it is obsolete.
Source: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Deprecated.html

Usually in the JDK, frameworks or libraries and as well in the code I write, @Deprecated indicates that this method shouldn't be used anymore because it is replaced by another method. And this is not the case with getElement().

Many of my customers are corporate enterprises, and they have code-checking rules. One is that using deprecated methods is forbidden. This would mean explaining to some enterprise architects why we are using getElement because otherwise, the build will fail, and they have to define an exception for us. That's a long and unnecessary process.

**My usages of getElement or executeJs**

- Setting attributes
  - icon.getElement().setAttribute("title", "");
  - download.getElement().setAttribute("download", true);
  - c.getElement().setAttribute("hidden", "")
- icon.getElement().addEventListener("click", event -> {}).addEventData("event.stopPropagation()");
- getElement().executeJs("return this.querySelector('input').value")
  used on a ComboBox to get the value when Enter is pressed
- Adding prefix components
- component.getElement().addEventListener("vaadin-context-menu-before-open",
- getElement().executeJs("document.activeElement.blur()")
- upload.getElement().executeJs("this.files=[]");
- getElement().getThemeList().add("tosca-dialog");
- Adding cllick listeners in general
- Moving selection in Grid up and down with PageUp/PageDown
- Scroll to an item in TreeGrid
- Getting the position of the splitter in a SplitLayout
- Getting the actual size of a component
- Adding an unload observer

getElement can be deprecated if methods in the components cover these use cases.

OlliTietavainenVaadin: I think deprecation is not a good choice even if it is used in the technically correct sense because (despite what Futurama memes tell us) being technically correct is in fact not the best kind of correct. People have attached a certain meaning to @Deprecated and we shouldn't need to tell "well ACTUALLY it means something else" to the 90% of people who are accustomed to deprecation meaning "old, obsolete, and likely insecure". And I do agree with Matti that needing to use JavaScript at all is a sign of something being wrong at the abstraction level, but people have been using that API for 4.5 years by now so it's an established thing now. Renaming the method is going to cause a ton of migration pain

mvysny: My usages of getElement():
* `Select.getElement().setAttribute("theme", "small")`, `MultiselectComboBox.getElement().setAttribute("theme")` (the old one which Goran implemented, with Vaadin 14 it doesn't implement the `HasTheme` interface).
* `EventUtil.inspectHierarchy()` for custom observers (e.g. components observing changes to the currently selected account - very business domain specific though)
* `SplitLayout.getElement().addEventListener("splitter-dragend")` because SplitterDragendEvent doesn't contain the new sizes (`element._primaryChild.offsetWidth` and `element._secondaryChild.offsetWidth`)
* `CustomField` doesn't implement `HasStyle` yet I need to set a style to it.
* `executeJs()` for workarounds around Vaadin bugs (e.g. https://github.com/vaadin/flow-components/issues/2176 )
* `Component` doesn't implement `HasStyle` yet my layout needs to add a style to it.

mstahv:
The more I get a list of things why people use getElement() in their apps, the more I hate it. I start to feel like it is the reason why we are still lagging behind V8 in some areas. Looks like it is too easy door for even for more experienced Vaadin users to make a hack or workaround, instead of pushing us the fix the actual components themselves or share re-usable add-ons with the rest of the community.

BTW. Without seeing the full details, some of the usage list points look like hacks that shouldn't be needed anymore. We are progressing, but still a lot of things to do to make our component API a true pleasure to use...

mstahv:
Dropping here a proposal from an internal discussions (that should not happen!) as an alternative replacement:

    Component.getInternals().getElement();

simasch:
@mstahv most of the workarounds I listed came from the Vaadin Support team. So there should also be a process to create feature requests from support requests.

mstahv:
@simasch I couldn't agree more. I'll discuss this with our support team.

Legioth:
> the more I hate it

What's there to hate about a mechanism that allows developers to proceed immediately rather than waiting for a non-trivial period of time until the next feature release?

There will always be more potential use cases than what we have time to provide explicit support for. Long tail theory implies that it's a net win to support rather than suppress that dynamic.

Deprecating or renaming `getElement()` wouldn't change anything in that regard. Making it `protected` would just turn those oneliners into lots of application-specific subclasses.

This was also common in the V8 era. Any "real" application was inevitably building up its own custom widgetset with workarounds. The difference was that each workaround introduced 50 lines of custom code to implement and maintain instead of only one. V8 never got built-in support for custom ARIA attributes, prefix components or download links despite how difficult it was to create workarounds.

All else being equal, it would obviously be better to have direct support for more common use cases. But that's not a reason to throw out the baby with the bathwater.

mstahv:
@Legioth, I'm not proposing to deprecate Element API. I really like the powers it gives to developers. That is great API and absolutely best we have ever had to build components at browser/DOM/JS level. The problem is just that it is too prominently visible. Newbies use use it for example to iterate child elements that they then cast into components or are wondering why the "value" property of a select is a number instead of their DTO.

Also (as discussed in the ticket), it makes more advanced Vaadin users code worse quality, because it makes it too easy to inline JS haxies to UI logic (for bugs/missing features), instead of hiding them into their component internals (not to mention about filing feature requests or wrapping their utilities into re-usable add-ons for other community members).

Getting a deprecation warning from getElement() (or other proposed measures like seeing a keyword like `internal` in the call chain) would for sure help new users understand that they are probably doing something wrong. And I'm quite sure many of the "one line haxies" would be hidden at least to private component methods. That would already document them and make the actual UI logic more readable, which would most likely save money in the long run.

I don't think V8 not having a perfect ARIA support has nothing to do this this issue. The reason for that was that we hadn't yet invested a million into ARIA support. More fundamentally, there was not yet that strong need for it through customer requests. And technically, in V8 you were already able to execute JS and this way accomplish pretty much all the same as with the current Element API, but just in bit uglier way. And for V7 I made similar helper HtmlProperteSetter or something to Viritin. In many cases those "50+ line GWT extensions" were easier to implement with those. *The difference with the current setup was that developers for sore knew that they had engine room door open when they accessed the next abstraction layer.*

TatuLund:
I agree with @knoobie that we should not deprecate and rename this method, but just add the JavaDoc note clarifying its purpose. The method is fundamental piece of our own internal component implementations, custom components and most of the add-ons. Changing the name and forcing users to update their code is expensive operation (even if it is mostly search / replace), which does not add value to their end products. I agree, that we could have originally named it better, but this is legacy what we have to live with.

Legioth (dismissed):
I agree with the JavaDoc changes but the deprecation is incomplete.

I don't have a strong opinion on whether deprecation is the right course of action but I'm absolutely certain that _if_ `getElement()` is to be deprecated, then three conditions MUST be fulfilled:
1. There must exist a documented non-deprecated alternative for all current and future uses of `getElement()`.
2. There's a plan for ensuring official documentation and examples are updated to not use the deprecated `getElement()` method.
3. It's made absolutely clear to current users and future framework maintainers that the deprecation doesn't imply that the `getElement()` method is not intended to be removed. This is almost fulfilled except that the `forRemoval` attribute of the annotation could be explicitly set to `false` even though that's also the default value.


The current PR satisfies none of those conditions.

mstahv:
Introduced an alternative and added explicit forRemoval=false, although it is default I think.

I'm not quite sure if our official docs/examples should use the alternative. As discussed, deprecation does not mean the API is broken or to be removed. The whole point of the deprecation is kind of lost if we use an alternative API to instead. The new alternative now does from its inception have the warnings in the JavaDocs, but when has people copy pasting code snippets read the docs 🤷‍♂️

Arguably the Element.of(Component) might already look a bit more "scary" than public non-deprecated Component.getElement, but I think we should still consider some alternative naming. Dropping some here, but somehow I feel the best is still not found:

```
Element.internalOf
Element.rootOf
Element.unwrap
```

mshabarov:
Element API was designed to have a control for HTML DOM in the browser from the server-side. 

It isn't supposed to be internal API by design (at least according to current docs and javadoc).
The docs describe clearly the list of use cases: https://vaadin.com/docs/latest/create-ui/element-api .

If I understand the concept of `Component` and `Element` right, the difference is that the `Component` API deals with the server-side Vaadin components, abstracting out from the browser and DOM, providing the base methods for traversing the components tree, enabling/disabling component, providing attach/detach hooks and so on; whereas the `Element` API corresponds exactly to the DOM counterpart on the server.
These two APIs equally deserve to be a part of Flow public API and shouldn't be anyhow discouraged for usage in the applications in my opinion, because they chase slightly different goals.
However, I admit that probably some of the methods in Element API should be available on the Component abstraction level, so the developers can use them in a first place.

Deprecating `getElement()` makes me think that the Element API is a purely internal thing, which is not the case, at least now.
If we deprecate it:
- existing experienced developers got no value, but would need to change their codes to use another method - extra work with no value.
- newbies would use a new recommended method (Element.of for instance) for the same purposes they used `getElement()` - the problem remains: the API is still used incorrectly, but with the new non-deprecated method.
So, unfortunately, I don't see how the deprecation could solve the original problem.

I propose to solve the original problem in a two steps:
1. improve javadoc only, but not saying that using it is a sign of doing something wrong, but tell when to use the Element methods and when not to: 
   -   use for: get/set DOM element properties/attributes, accessing data from browser's events, calling JS from server, dynamic styling and basically for the documented use cases from here https://vaadin.com/docs/latest/create-ui/element-api.
   - not use for: getting parent / ancestor / children component(s).
2. figure out what API is missing in the Component class to let developers use that directly, and not to dig into Element class. 

(1) can be made within this PR. (2) can be made separately and iteratively based on the issues reported by developers.

mshabarov (changes requested):
https://github.com/vaadin/flow/pull/15511#issuecomment-1368953152

mstahv:
Yes, Element API is not internal API. It is public API and a key concept when building components (our core components, add-on components and project specific components), that are then used to build the actual application UI. The Element API itself is great and by no means a problem, but it is not meant to be used in your application code. As written in the description (in a too prompt way, shame on me), the problem is getElement() method in Component, that makes it is too easy (by accident or by laziness) to take a shortcut to the lower abstraction level. This severely degrades the quality of Vaadin applications and makes it harder to learn the best practices.

@mshabarov What are the use case you mean with :

> The docs describe clearly the list of use cases: https://vaadin.com/docs/latest/create-ui/element-api .

Is any of those somehow in conflict in deprecating getElement in Component and especially with the Javadocs written in this patch? If our documentation says Element API should be used to build actual applications, it should be fixed.

Now that you pointed out the docs section discussing Element API, I can see that we have a BIG issue(s) with our documentation as well. And that probably also contributes to the original problem we have with Element API (that it is too leaks too easily to other abstraction layer). By the structure one could think Element API should be used within normal process of "Creating UI", but it should in fact be a sub-section in creating components. Also, the content in "Creating components" and "Element API" are already both discussing the same things. Element API is already discussed a lot in "Creating components" and then re-introduced. And probably Integrating Web Components should be part of "Creating components" and "Creating Components" should be a top level topic 🤔

mvysny:
I see element API akin to Java reflection API. Many libraries depend on it and therefore it's not deprecated. Yet it's niche and if you overuse it in your app, you're doing something wrong. It's intended for advanced use-case, and no beginner should be forced to use it.

Legioth:
> Arguably the Element.of(Component) might already look a bit more "scary" than public non-deprecated Component.getElement, but I think we should still consider some alternative naming.

In my mind, the main benefit with having it as a static method in `Element` (regardless of the name) is that it won't show up as an option for auto completion based on a component instance. This means that you won't stumble upon it by accident but you can still easily access it as long as you know what you're looking for.

Out of the suggested naming options, I think I prefer `unwrap` since that implies that you're escaping to a lower abstraction level without being too verbose. `internalOf` is also in the same direction, but it raises the question of whether there's also a non-internal element for a given component.

On the other hand, showing up in auto completion for component instances might still be fine as long as the name is clearly indicating that it's a low level concept. Clear naming is also important for cases where code is copied from some example. There we could again have `someComponent.getInternalElement()` with the same question about non-internal elements. What about `someComponent.unwrap()` or `someComponent.getInternals().getElement()`?

If we introduce `Component::getInternals`, then we could also (separately) consider parts of `ComponentUtil` into the same class, e.g. `someComponent.getInternals().fireEvent(someEvent)` and `someComponent.getInternals().setData("foo", "bar")`

mstahv:
After a couple of weeks break and reading Leif's comment yesterday, I think unwrap is a decent name the replacing method in Element interface.

With Leif's comment about discoverability with autocomplete, I think the current solution is a fair compromise. Now this fairly often needed method gets found pretty easily (deprecated methods appear at least with my IDEs), but new users immediately see the disclaimers, even if the end up using the non-deprecated alternative. 

In our own code and examples we should probably keep using the getElement version (people will copy paste code without even trying to understand it). Also, we should re-visit code examples when updating the documentation, so that they follow best practices and guide people on right track.

mshabarov:
@mstahv I'd propose a small enhancement to the javadoc changes you made (replace Element.of with the Element.unwrap and provide some examples when to use each API):
```java
/**
 * Gets the low level root element of this component.
 * <p>
 * <b>Note!</b> Element API is designed for building components at a lower
 * abstraction level than normal Vaadin UI development. If you see a direct
 * call to this method in your applications UI code, you should consider
 * that as a sign that you are probably doing something wrong and check the 
 * methods in the Component class first, e.g. for getting children, parent or 
 * ancestor component. This method is breaking the Component's abstraction layer and its
 * implementations provided. You should only call this method and use the Element API when
 * creating or extending components (e.g. setting the attributes and properties, 
 * execute JavaScript codes), or when you otherwise need to break
 * through the abstraction layer. If it is a hack or a workaround, it is
 * also better to hide that into an extension, helper class, separate add-on
 * module or at least into private helper method documenting it.
 * <p>
 * Each component must have exactly one root element. When the component is
 * attached to a parent component, this element is attached to the parent
 * component's element hierarchy.
 *
 * @return the root element of this component
 * @deprecated This is deprecated as a warning as you should never call this
 *             method directly from your UI logic. This not planned to be
 *             removed and calling this method is perfectly fine when
 *             creating components/extension to be used in your application
 *             or to be shared as add-on components. If your project
 *             settings don't allow you to call deprecated methods
 *             suppressing and suppressing warnings, use
 *             {@link Element#unwrap(com.vaadin.flow.component.Component) }
 *             instead.
 */
@Deprecated(forRemoval = false)
@Override
public Element getElement() {
}

/**
 * Gets the underlying internal root Element of the given component. An
 * alternative to {@link Component#getElement() }.
 * <p>
 * <b>Note!</b> Element API is designed for building components at a lower
 * abstraction level than normal Vaadin UI development. If you see a direct
 * call to this method in your applications UI code, you should consider
 * that as a sign that you are probably doing something wrong and check the 
 * methods in the Component class first, e.g. for getting children, parent or 
 * ancestor component. This method is breaking the Component's abstraction layer and its
 * implementations provided. You should only call this method and use the Element API when
 * creating or extending components (e.g. setting the attributes and properties, 
 * execute JavaScript codes), or when you otherwise need to break
 * through the abstraction layer. If it is a hack or a workaround, it is also better to hide that
 * into an extension, helper class, separate add-on module or at least into
 * a private method documenting the usage.
 *
 * @param component
 *            the component whose element should be obtained
 * @return root element of the component
 */
@SuppressWarnings("deprecation")
public static Element unwrap(Component component) {
    return component.getElement();
}
```
Do you think those are reasonable?

It seems to me that we don't have many options to make the javadoc visible for users. Although I don't much like the deprecation, but it does worth it for the problem we're trying to resolve here. The replacement `Element.unwrap` looks good to me.

I'm taking the action to make an issue in the docs about revisiting Component API and Element API explanations and examples.

mshabarov:
Just noted that the component class has `Component::from(Element, Class)`, would it be better to have the same in Element: `Element::from(Component)` for consistency?

mstahv:
Committed some changes based on suggestions by @mshabarov

mstahv:
I'd stick to unwrap. Feels more right. I have never used Component.from method actually. And still don't understand what it is needed for when I read its javadocs. Seems very framework internal thing based on its usage.

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -1618,4 +1618,32 @@ public Element scrollIntoView() {
...
+     * class, separate add-on module or at least into a private method
+     * documenting the usage.
+     *
+     * @param component
+     *            the component whose element should be obtained
+     * @return root element of the component
+     */
+    @SuppressWarnings("deprecation")
+    public static Element unwrap(Component component) {
+        return component.getElement();
```
mshabarov, line 1647: A null check can be added here.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
mstahv: It could, but why? I guess we don't want the legendary `if(v == null) throw new NullPointerException()` 😂 Do you mean assertion? Don't know how that would help the world either though.

mshabarov: `Objects.requireNonNull` can be used. It would help to better spot the problem during development and its a good practice to check the input parameters. However the body of this method is trivial and it's very easy to figure out the source of NPE. Up to you to add it.

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
mstahv: Which is just doing the above ;-) Especially with JDK 17 (where this code snippet is only going) I find doing that obsolete, unless the parameter can cause "weird things" later on in the code. Here, a "natural NPE" comes from the only line of the method.

Legioth:
I got some additional thoughts on this topic. To summarize, I don't think we should deprecate `getElement()` based on the signals we have received so far because they are biased: those who would stand to lose from the change are much less likely to send signals in the first place.

Discoverability is important mostly for new users that don't know what they are looking for. Discoverability might also be useful for seasoned users who just don't remember exactly where to find something, but I don't think that's a significant concern in this case.

What makes this interesting is that new users aren't a homogenous group. Specifically for this case, they fall somewhere on the spectrum on how compatible their existing mental models are with the way Flow works (with regards to the Element API).

* Some new users start with a mental model where encountering the Element API early in their discovery journey leads to the types of complications that was the trigger for doing this PR. They struggle to understand Flow because of the Element API.
* Other new users are on the other extreme of the spectrum with an initial mental model where the Element API is essential to their understanding of how Flow works. They struggle to understand Flow unless they they discover the Element API.

Most users fall somewhere in the middle of the spectrum for which the discoverability of `getElement()` has no significant impact.

The thing to realize is that some subset of the first group is visible by showing their struggles and asking for help. The second group is on the other hand much more likely to just silently dismiss Flow because it doesn't "click" for them. If we assume that both scenarios are equally likely, then we would still see much more direct indicators for the first scenario. This in turn means that we cannot rely on those indicators.

Based on this line of reasoning, I'm now strongly recommending that we stay closer to the status quo by only clarifying documentation (both JavaDocs and /docs) while not making any API changes. We can then evaluate how well the documentation clarifications help reduce the symptoms before we would reopen the discussion about more radical changes.

Legioth:
IDEs prefer to show auto complete suggestions for non-deprecated APIs. Discoverability is therefore significantly reduced for API that is deprecated.

mstahv:
Just to repeat my old thinking: just adding javadocs is not going to solve the original problem. Especially now with IntelliJ becoming the most used IDE, people are not finding the disclaimers, even if we'd write them in CAPITALS.

Legioth:
You are right that deprecation won't move `getElement()` out of view because of the way IDEs also take things like historical use into account when determining how to rank suggestions. The underlying point still stands: deprecation directs new users away from considering `getElement()` which in turn is beneficial for those with a mental model that would lead to "misuse" while it's harmful for those with a mental model where familiarity with the Element API is essential for things to "click" enough for them to proceed on their journey. I don't see how we could assume that deprecation only affects the behavior of one group without also affecting the other.

I think we should assume that new users seeing a deprecated `getElement()` in an autosuggest list wouldn't be intrigued to start looking into why it's deprecated. A deprecation message for  `getElement()` might still have an impact when someone copies a code example that uses `getElement()` but the deprecation might also lead them to believe that they have found an outdated example that they should ignore.

Finally, I should clarify that if we would be designing the core Flow APIs from scratch based on what we know today, then I might lean towards API like `Element.unwrap(component)` rather than `component.getElement()`. But we're not in that situation today since there's a big existing community centered around `getElement()`. The realization that there might also be positive outcomes from new developers encountering `getElement()` has shifted the arithmetics in my mind from 1 - 1 to 0 - 1. But this is only my opinion rather than a definite decision. It's up to the PO to take all the arguments into account and determine which way we will go and I won't judge the decision regardless of which way it goes.

mstahv:
I think it would be very good if people would find Element API through documentation or good examples that utilise it in good manners.

Legioth:
>Just to repeat my old thinking: just adding javadocs is not going to solve the original problem.

I don't think we can _fully_ solve this with anything short of of completely removing the Element API. The question is just how far we would have to go to reach something that is sufficient?

>I think it would be very good if people would find Element API through documentation or good examples that utilise it in good manners.

That's why I've started leaning towards the idea that we would start with only updating documentation. Not only JavaDocs but also how things are described in the documentation site.

In that way, we might approximate getting 80% of the benefits with 20% of the "cost". The benefits might not be fully 80%, but I'm also sure that the costs would be much less than 20% when also taking into account how deprecation would affect the community (since no matter how we communicate, there will still be lots of confusion and lots of developers who feel they must update their code to get rid of their deprecation warnings).

mshabarov (changes requested):
Here is the issue for documentation improvement https://github.com/vaadin/docs/issues/2029, I've put there a general suggestions, the concrete steps can be added when we start working on that.

@mstahv the problem you are trying to solve with the deprecation is clear and I trust you in that regards, that's why I approved the changes despite I wasn't happy with the deprecation itself.
I asked internally in the Flow team about deprecation and nobody would want to go with deprecation. Based on the discussion we have so far in this pull request, looks like there are no strong "pro" opinions to deprecate `getElement`. Thus, I expect a huge negative feedback from experienced Vaadin users if we go this way. The explanations that we are trying to make it less happening to erroneously use Element API by some group of Vaadin newcomers sounds like a bad argument that I can give them. It feels for me so far that we sacrify the consistence for one group of users in favour to other group of users, and I'm not really sure which group is bigger or smaller and whether this is a right way of evolving the framework. Personally, being not a "thoughtful" guy and if I would start to evaluate any new framework, I would maybe just replace the deprecated method with a recommended one, I cannot guarantee that I would really change my mind and trying to do the things in another (proper) way, if my solution works. That's a bad thing, but I bet some group of people would do the same. Thus, I'm not sure the deprecation would really improve the situation, at least pay off the efforts we invest in that (replace `getElement()` everywhere).

Instead of deprecation, I would propose to focus on the docs and probably on the actual goals of the developers: what they are trying to find in the element API which is not available in the component API. Based on such a feedback we could extend the Component API further.

mstahv:
I can't express how dissapointed I am, especially as it happend like this that Leif stepped in to turn the boat that had already chosen the direction ;-)

With the initial blunt deprecation, I knew this would raise a lot of discussion, and was prepared to drop that part, but with the Element.unwrap alternative I think we had a decent compromise of fixing old design mistakes and keeping backards compatibility. To me this now feels like the current western support for Ukraine; we don't dare to do enough, to make a difference.

At least the ones following this thread will probably now understand the difference of Element and Component API, and hopefully write better code in the future.

mshabarov (approved):

knoobie:
I'm going to keep my comment short - we already had enough lengthy discussions here. 

I personally think it's good the method wasn't deprecated at this stage / now. 

But I was wondering, why the unwrap method was also removed again? Wouldn't it make sense (if all / most agree) to slowly use Element.unwrap to remove the usage of getElement() by e.g. using the new unwrap way of things
- when writing new code 
- when helping / promoting fixes for people 
- when updating docs 
- when explaining the layer in a blog post 😉
- you probably have even more ideas 

Allowing for a deprecation / removal hopefully in the loooooong run, once unwrap has wider use by people and the concepts are better known / used by all the available example / docs / cookbooks.

mstahv:
There was this new theory about "a group of evaluators" who might figure out why Vaadin is good because of the non-deprecated getElement method in Component, that turned the boat. So, as we will not be getting rid of this method in foreseeable future, I see there is no reason to have this separate (and more cumbersome) method to do the same.

PR #15514, opened 2022-12-16
chore: Upgrade to Vite 4.0.2

Artur-:


Artur-:
Is that useful to have?

caalador:
Update versions if old, but do not downgrade versions is ofc a good test to have, but could we have that somehow so that we would depend on the default version and add on top of that version a +0.1.0 or something

Artur-:
Made the number slightly bigger

caalador (approved):

Issue #15478, opened 2022-12-14
Custom theme should be processed separately from Express Build bundle


mshabarov:
The Flow's custom theme is currently processed as part of the frontend bundle. Custom theme should be processed separately from the bundle and modified/loaded without a frontend build, so no bundle re-compilation is needed: instead of injecting the CSS as JS into component's shadow DOM, we can handle CSS as static resources separately from bundle.

Acceptance criteria:
- [x] When working in the Express Bundle mode, it should be possible to modify the styles in `styles.css` without re-compiling the bundle and see the results by refreshing the page.
- [x] Vaadin Component styling should be possible with using [::part() selector](https://github.com/vaadin/platform/issues/3149) in `styles.css`.
- [x] Should cover the cases when the stylesheets are imported from `node_modules` to the `styles.css`.
- [x] Should serve assets/styles from project's dev-bundle folder (in case of a customised bundle) and from the classpath (in case of the default bundle).
- [ ] Should support `scss`.

mshabarov:
Support for `sccs` will be tested separately.

PR #15518, opened 2022-12-16
feat: Serve theme files separately from dev bundle

mshabarov:
In the Express Build mode:
- serves theme files (`styles.css` and other project's stylesheets imported to `styles.css`) separately from the `dev-bundle` directly from custom theme (`frontend/themes/my-theme`) folder
- copies the assets referenced in `theme.json` to the dev bundle within bundle re-compilation
- serves assets/styles from `dev-bundle`, if can't find them in the `frontend` folder

Does not yet take into account serving styles from default bundle (from classpath).

Related-to https://github.com/vaadin/flow/issues/15478


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -539,7 +540,7 @@ export const vaadinConfig: UserConfigFn = (env) => {
     },
     build: {
       outDir: buildOutputFolder,
-      emptyOutDir: devBundle,
```
Artur-, line 542: Why?

mshabarov: I've commented it when was developing the feature, because if it's set to `true` the assets copied from node_modules to the dev-bundle was removed immediately. So it was copied and removed. I'll double check it.

mshabarov: Seems like `themePlugin({devMode})`, which copies the assets to the `dev-bundle` is executed before the Vite empties the bundle output folder. An easiest way to fix it is probably to copy assets into `dev-bundle/assets`, not inside the bundle output dir `dev-bundle/webapp/assets`.

mshabarov: Fixed. Thanks for noticing this 👍

caalador (approved):

PR #15540, opened 2022-12-21
feat: add Svg component

mstahv:
A very trivial implementation capable only do display static SVG at this point.

Compared to using Html, this component don't need any hacks to wrap svg into some html element.

Compared to using SVG as Image source, the SVG content displayed using this component can be styled with CSS.

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/component/Svg.java
```diff
@@ -0,0 +1,113 @@
...
+     * Creates an instance based on the given SVG string. The the string must
+     * have exactly one root element.
+     *
+     * @param svg
+     *            the SVG to display
+     */
+    public Svg(String svg) {
+        super(null);
+        if (svg == null || svg.isEmpty()) {
+            throw new IllegalArgumentException("HTML cannot be null or empty");
```
OlliTietavainenVaadin, line 79: SVG, not HTML

mstahv:
If any of @vaadin/flow team member or other core hackers would have good idea how to quickly implement a way to create to root element with createElementNS or another hack how to get rid of the wrapper div, it would be cleaner and good for forward compatibility (in case we ever get SVG support to Element API). Otherwise I believe this component would already help people as such.

taefi:
Thanks @mstahv for the contribution. Just a question, why isn't using the `Image` helping? For example, I'm representing an SVG logo like this: 

```
Image logo = new Image("images/some-logo.svg", "logo"); // where it is located under META-INF.resources
```
and add it directly to the layout.

mstahv:
@taefi : "Compared to using SVG as Image source, the SVG content displayed using this component can be styled with CSS."

With CSS you can for example change colors of icons without actually touching the SVG itself. @anezthes Can provide you more details.

I don't think dynamic creation of SVG itself has nothing to do with this case, that can be done with Image already if needed. This is more about styling and discoverability/documentation.

flow-server/src/main/java/com/vaadin/flow/component/Svg.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Creates an instance based on the given SVG input. The string must have
+     * exactly one root element.
+     *
+     * @param stream
+     *            the SVG to display
+     */
+    public Svg(InputStream stream) {
+        super(new Element("div"));
```
taefi, line 51:
```suggestion
        this();
```

flow-server/src/main/java/com/vaadin/flow/component/Svg.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Creates an instance based on the given SVG string. The the string must
+     * have exactly one root element.
+     *
+     * @param svg
+     *            the SVG to display
+     */
+    public Svg(String svg) {
+        super(new Element("div"));
```
taefi, line 63:
```suggestion
        this();
```

flow-server/src/main/java/com/vaadin/flow/component/Svg.java
```diff
@@ -0,0 +1,137 @@
...
+     * @param svg
+     *            the SVG string
+     */
+    public void setSvg(String svg) {
+        if (svg == null || svg.isEmpty()) {
+            setInnerHtml("");
+        } else {
+            validateAndSet(svg);
+        }
+    }
```
taefi, line 89: IMO, there is a behavioural inconsistency between this `setSvg(String svg)` method and the constructor that accepts `String svg` as input argument. If a `null` or empty svg content is valid (as it is assumed so in this method), that constructor should not check the input for a `null` or empty content and should just call this setter method. Otherwise, let's move that logic into this method and again call this setter within the constructor.

flow-server/src/test/java/com/vaadin/flow/component/SvgTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @Test(expected = IllegalArgumentException.class)
+    public void text() {
+        new Svg("hello");
+    }
+
+    static String TRIVIAL_SVG = """
+            <svg>
+                <circle cx="50" cy="50" r="40" stroke="green" stroke-width="4" fill="yellow" />
+            </svg>
+                                    """;
```
taefi, line 57: (Not really that important)
Is the extra line and empty string in the end intentional? Otherwise, we can remove it:
```suggestion
            </svg>""";
```

taefi: Same for the next SVG static content definitions.

flow-server/src/test/java/com/vaadin/flow/component/SvgTest.java
```diff
@@ -0,0 +1,111 @@
...
+                 <circle cx="125" cy="125" r="75" fill="orange" />
+                 <polyline points="50,150 50,200 200,200 200,100" stroke="red" stroke-width="4" fill="none" />
+                 <line x1="50" y1="50" x2="200" y2="200" stroke="blue" stroke-width="4" />
+            </g>
+            </svg>
+                                               """;
+
+    @Test
+    public void simpleSvg() {
+        Svg html = new Svg(TRIVIAL_SVG);
```
taefi, line 82: I suggest to improve the naming of local variable `html`, for example to something like `svg`. (Same for the next tests)

flow-server/src/test/java/com/vaadin/flow/component/SvgTest.java
```diff
@@ -0,0 +1,111 @@
...
+        Assert.assertEquals(TRIVIAL_SVG,
+                html.getElement().getProperty("innerHTML"));
+    }
+
+    @Test
+    public void withDocType() {
+        Svg html = new Svg(SVG_WITH_DOCTYPE_ET_AL);
+        Assert.assertTrue(
+                html.getElement().getProperty("innerHTML").startsWith("<svg"));
```
taefi, line 91: I would suggest to add an additional method like `getSvgContent` to the `Svg` component, so that the user don't have to deal with low-level API to get the content anymore.

taefi (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/Svg.java
```diff
@@ -0,0 +1,137 @@
...
+     * @param stream
+     *            the SVG to display
+     */
+    public Svg(InputStream stream) {
+        super(new Element("div"));
+        setSvg(stream);
+    }
+
+    /**
+     * Creates an instance based on the given SVG string. The the string must
```
taefi, line 56:
```suggestion
     * Creates an instance based on the given SVG string. The string must
```

taefi (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/Svg.java
```diff
@@ -0,0 +1,137 @@
...
+        super(new Element("div"));
+    }
+
+    /**
+     * Sets the graphics shown in this component.
+     *
+     * @param svg
+     *            the SVG string
+     */
+    public void setSvg(String svg) {
```
taefi, line 83: I suggest to rename the setters for the content of the Svg component to `setContent`, since having a `svg` variable and calling a `setSvg` method on it is verbose and even a little confusing.
```suggestion
    public void setContent(String svgContent) {
```

flow-server/src/main/java/com/vaadin/flow/component/Svg.java
mstahv: Yeah, remains of a copy pasted initial poc from Html...

flow-server/src/main/java/com/vaadin/flow/component/Svg.java
mstahv: I don't agree. That method name would confuse people to pass in only content of the top level svg element.

mstahv: In Image we have setSrc. Don't like that either, especially in that shorthand form.

mstahv:
@taefi This is because Vaadin is creating the svg element on the client side with createElement, should use createElmentNS (or something to create it with the different namespace).

mstahv:
Do I remember wrong or was there some "virtual child" concept? That with some JS hack maybe (if we don't want to touch the client side code and add an actual framework feature)?

flow-server/src/main/java/com/vaadin/flow/component/Svg.java
taefi: I get your concern about a situation that someone might pass children contents of a `svg` instead of including the `<svg ...> </svg>` itself as well, but imagining myself reading/reviewing a code/documentation/tutorial (having a `Svg svg = new Svg();` in the context), I still prefer:
```
svg.setContent("<svg ...></svg>");
```
over
```
svg.setSvg("<svg ...></svg>");
```
Maybe this is only my personal taste.

Btw, we can address that issue by a good documentation. Also, given the fact that the implementation of the `validateAndSet` is taking care of the presence of additional contents such as `<?xml ...>` or `<!DOCTYPE svg... >`, or not having the `<svg ...>` in the input, I don't see a reason to add the verbosity.

taefi:
> @taefi This is because Vaadin is creating the svg element on the client side with createElement, should use createElmentNS (or something to create it with the different namespace).

I'm not sure if this is somehow related as well, but there seems to be an open issue for rendering `<svg>` in shadow dom:
https://github.com/WICG/webcomponents/issues/179

taefi:
@mstahv please `format` and push one last time to make it OK to merge. 

Another issue for could be created labeled as improvement. It is good to be merged IMO.

mstahv:
Formatted✅

Enhancement issue: https://github.com/vaadin/flow/issues/15601

taefi (approved):

PR #15552, opened 2022-12-22
fix: Replace hardcoded frontend folder with the configured one

mshabarov:
Extends the application configuration with the new methods for getting frontend and generated folder paths and replaces hardcoded paths with these config parameters.


flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
```diff
@@ -179,4 +179,37 @@ default File getJavaResourceFolder() {
...
                 "src/main/resources"));
     }
+
+    /**
+     * Return the folder containing application frontend source files.
+     *
+     * @return the folder containing frontend source files, by default it is
+     *         <code>/frontend</code> in the project folder.
+     */
+    default File getFrontendDirectory() {
```
Artur-, line 189: These utils are already in FrontendUtils nowadays, I think

mshabarov:
Need a rebase to main. Better to create a new pull request when it's ready, rather than having this PR hanging.

PR #15554, opened 2022-12-22
feat: Validate bundle frontend files

### LINKED ISSUE #15553: Validate bundle against in memory generated-flow-imports.js
Generate in memory generated-flow-imports for project to check that all requires JsImports arre available in the bundle.

caalador:
Validate frontenfiles in the bundle
to what generated-flow-imports.js contains
for the project.

Fixes #15553

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
@@ -38,6 +38,7 @@ import com.vaadin.flow.server.frontend.scanner.FrontendDependenciesScanner;
 import com.vaadin.flow.shared.util.SharedUtil;
 
 import elemental.json.Json;
+import elemental.json.JsonArray;
 import elemental.json.JsonObject;
 import static com.vaadin.flow.server.Constants.DEV_BUNDLE_JAR_PATH;
 
@@ -97,6 +98,7 @@ public class TaskRunDevBundleBuild implements FallibleCommand {
             FrontendDependenciesScanner frontendDependencies,
             ClassFinder finder) throws IOException {
         File npmFolder = options.getNpmFolder();
+
         if (!FrontendUtils.getDevBundleFolder(npmFolder).exists()
                 && !hasJarBundle()) {
             return true;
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -121,7 +123,51 @@ protected static boolean needsBuildInternal(Options options,
...
+            FrontendDependenciesScanner frontendDependencies) {
+
+        // Validate frontend requirements in flow-generated-imports.js
+        final GenerateMainImports generateMainImports = new GenerateMainImports(
+                finder, frontendDependencies, options);
+        generateMainImports.run();
+        final List<String> imports = generateMainImports.getLines().stream()
+                .filter(line -> line.startsWith("import"))
+                .map(line -> line.substring(line.indexOf('\'') + 1,
+                        line.lastIndexOf('\'')))
```
mshabarov, line 145: This transformation cuts of the file name if I understand it right.
But for some imports we can have more than one file in the folder, e.g.:
```
import '@vaadin/app-layout/theme/lumo/vaadin-app-layout.js';
import '@vaadin/app-layout/theme/lumo/vaadin-drawer-toggle.js';
```
So how they would be taken into account?
Also, probably `/` should be used instead of `\`.

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -199,10 +199,15 @@ function statsExtracterPlugin(): PluginOption {
       const projectPackageJson = JSON.parse(readFileSync(projectPackageJsonFile, { encoding: 'utf-8' }));
 
       const entryScripts = Object.values(bundle).filter(bundle => bundle.isEntry).map(bundle => bundle.fileName);
+      const generatedImports = readFileSync(path.resolve(generatedFlowImportsFolder, "generated-flow-imports.js"), {encoding: 'utf-8'})
```
mshabarov, line 202: Not sure I've got the idea: so we take `generated-flow-imports.js` and write the imports to the stats.json. Then we compare `generated-flow-imports.js` with the stats.json again. At what point `generated-flow-imports.js` is updated, so it wouldn't be the same as stats.json? When `@NpmPackage` or `@JsModule` or `@CssImport` are added?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
caalador: The substring takes the part from `'` to `'` so in the example case:
```
@vaadin/app-layout/theme/lumo/vaadin-app-layout.js
@vaadin/app-layout/theme/lumo/vaadin-drawer-toggle.js
```
the dropped part is `import '` and `';`

flow-server/src/main/resources/vite.generated.ts
caalador: So when we make a dev-bundle we collect the imports used to the stats.json so that when starting an application we can check that all necessary imports `@JsModule` and `@CssImport` have been used in the bundle. if there are missing ones we need to rebuild the bundle. The contents do not match 100% (nor does it need to) if it is not an application-dev-bundle.
Also we do not compare in vite as vite just generates the file if the app-dev-bundle build is needed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov: Ah, I overlooked quotation mark, I thought it was `'\'`, not `'\''`. Thanks!

caalador: Updated string to use " instead of ' to make the checked char clearer

flow-server/src/main/resources/vite.generated.ts
caalador: added comment

mshabarov (approved):

PR #15580, opened 2022-12-29
refactor: Move dev tools TS to vaadin-dev-server

Artur-:
It has nothing to do with the client engine and development becomes more difficult when it is there


flow-tests/test-dev-bundle/pom.xml
```diff
@@ -28,6 +28,12 @@
             <type>pom</type>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-dev-server</artifactId>
+            <version>${project.version}</version>
+            <optional>true</optional>
+        </dependency>
```
caalador, line 36: I guess this will require updating the platform `vaadin-dev-bundle` pom template to add the optional vaadin-dev-server also.

caalador (approved):

flow-tests/test-dev-bundle/pom.xml
Artur-: What is the purpose of the flow artifact? Is it even intentional that the dev server is not there?

caalador: I think the idea is to have a simple flow dependencies/version override... 
It might not be intentional that it's not there. It is simple to forget if adding a new module.

PR #15582, opened 2022-12-29
fix: Memory generated should use theme

caalador:
The in-memory generated-flow-imports needs
to take into account the possible theme used.

part of #15376

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/GenerateMainImports.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/GenerateMainImports.java
@@ -15,11 +15,14 @@
  */
 package com.vaadin.flow.server.frontend;
 
+import java.io.File;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/GenerateMainImports.java
```diff
@@ -84,8 +95,11 @@ protected Collection<String> getGeneratedModules() {
         if (options.getGeneratedFolder() == null) {
             return Collections.emptySet();
         }
+        // Exclude generated-flow-imports.js as it's not a generated module.
         return NodeUpdater.getGeneratedModules(options.getGeneratedFolder(),
-                Collections.emptySet());
+                Collections.singleton(
+                        new File(options.getGeneratedFolder(), IMPORTS_NAME)
+                                .getName()));
```
mshabarov, line 102: Can it be simplified to
```java
Collections.singleton(IMPORTS_NAME));
```
?

mshabarov (approved):
Tested with the starter project. Theme imports are properly handled and no bundle re-compilation is observed, works as expected.

flow-server/src/main/java/com/vaadin/flow/server/frontend/GenerateMainImports.java
mshabarov: Can be discussed later on and fixed, if necessary.

PR #15583, opened 2022-12-29
fix: prevent flush changes when synchronizing properties

### LINKED ISSUE #15615: Regression: property synchronization events cause a value flush
### Description of the bug

A regression introduced in https://github.com/vaadin/flow/issues/14090

When changing the value of NumberField from "" to "1" back and forth, the validation starts behaving weirdly (see the recording). As follows from the server logs, changing the value that way sometimes causes a value flush, but the flushed value is always the previous one.

In this specific case, the reason is the `has-input-value-changed` event that _results from the `@Synchronize` annotation_ and that _causes a value flush_. This event is fired before `value-changed` so an old value is flushed. 

### Recording

https://user-images.githubusercontent.com/5039436/210231161-171018ce-1833-4fb0-8239-d76f175bd9d2.mov

### Expected behavior

Property synchronization events should not be treated as triggered by the user and hence should not cause a value flush.

```
NumberField numberField = new NumberField();
numberField.setValueChangeMode(ValueChangeMode.LAZY);
numberField.setValueChangeTimeout(5000);
numberField.addValueChangeListener(event -> {
    System.out.println("New value: " + event.getValue());
});
add(numberField);
```

### Versions

- Vaadin / Flow version: 24
- Java version: 17
- OS version: Mac OS

mcollovati:
DOM events related to property synchronization must not flush changes
to avoid wrong execution order of events on server side.
For example, an 'has-input-value-changed' event may occur before the
actual value of a field has been changed and a flush will trigger
listeners with an old value.

Fixes #15615

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
+++ b/flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
@@ -15,7 +15,6 @@
  */
 package com.vaadin.client.flow.binding;
 
-import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
@@ -1316,11 +1315,17 @@ public class SimpleElementBindingStrategy implements BindingStrategy<Element> {
         if (sendNow) {
             // Send if there were not filters or at least one matched
 
-            // Flush all debounced events so that they don't happen
-            // in wrong order in the server-side
-            List<Consumer<String>> executedCommands = Debouncer.flushAll();
+            boolean commandAlreadyExecuted = false;
+            boolean flushPendingChanges = synchronizeProperties.isEmpty();
 
-            if (!executedCommands.contains(sendCommand)) {
+            if (flushPendingChanges) {
+                // Flush all debounced events so that they don't happen
+                // in wrong order in the server-side
+                commandAlreadyExecuted = Debouncer.flushAll()
+                        .contains(sendCommand);
+            }
+
+            if (!commandAlreadyExecuted) {
                 sendCommand.accept(null);
             }
         }
--- a/flow-client/src/test-gwt/java/com/vaadin/client/flow/GwtPropertyElementBinderTest.java
+++ b/flow-client/src/test-gwt/java/com/vaadin/client/flow/GwtPropertyElementBinderTest.java
@@ -17,6 +17,7 @@ package com.vaadin.client.flow;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import com.vaadin.client.ApplicationConfiguration;
 import com.vaadin.client.ClientEngineTestBase;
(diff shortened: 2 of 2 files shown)


flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
```diff
@@ -1316,16 +1317,25 @@ private void handleDomEvent(Event event, BindingContext context) {
...
                 sendCommand.accept(null);
             }
         }
     }
 
+    private static native void nativeConsoleLog(String message, Object... args)
+    /*-{
+        console.log(message, args);
+     }-*/;
+
```
vursen, line 1338:
```suggestion
    private static native void nativeConsoleLog(String message, Object... args)
    /*-{
        console.log(message, args);
     }-*/;

```

flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
```diff
@@ -1316,16 +1317,25 @@ private void handleDomEvent(Event event, BindingContext context) {
         if (sendNow) {
             // Send if there were not filters or at least one matched
 
-            // Flush all debounced events so that they don't happen
-            // in wrong order in the server-side
-            List<Consumer<String>> executedCommands = Debouncer.flushAll();
+            boolean commandAlreadyExecuted = false;
+            if (flushPendingChanges) {
```
vursen, line 1321:
```suggestion
            boolean flushPendingChanges = synchronizeProperties.isEmpty();
            if (flushPendingChanges) {
```

vursen:
Shall we add some tests?

flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
mcollovati: Good point. I defined it at top because I was uncertain if other checks were needed, other than synchronizeProperties emptiness.

flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
mcollovati: done

flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
mcollovati: done

knoobie:
Shouldn't this fix be prioritized because the feature it fixes was also backported and is released in 23.3 and below?

mcollovati:
@vursen do you have time to take another look into this one?

ping also @manolo, do you mind looking at GWT tests?

vursen:
@mcollovati The solution looks good to me. However, as for tests, I would leave them to @manolo to review as I'm not familiar with GWT.

manolo (approved):

PR #15586, opened 2022-12-29
fix: Use current folder when prepare-frontend has not run

Artur-:
When you launch a Spring Boot Application class directly from the IDE, there is no PROJECT_BASEDIR info available

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
@@ -17,6 +17,8 @@ package com.vaadin.flow.server;
 
 import java.io.File;
 import java.io.Serializable;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 
 import com.vaadin.flow.server.frontend.FrontendUtils;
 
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
```diff
@@ -162,7 +162,7 @@ default String getBuildFolder() {
     default File getProjectFolder() {
         String folder = getStringProperty(FrontendUtils.PROJECT_BASEDIR, null);
         if (folder == null) {
-            return null;
+            return new File("").getAbsoluteFile();
```
caalador, line 165: DevModeInitializer uses:
```
        String baseDir = config.getStringProperty(FrontendUtils.PROJECT_BASEDIR,
                null);
        if (baseDir == null) {
            baseDir = getBaseDirectoryFallback();
        }
```
Where getBaseDirectoryFallback does

```
     * Accept user.dir or cwd as a fallback only if the directory seems to be a
     * Maven or Gradle project. Check to avoid cluttering server directories
     * (see tickets #8249, #8403).
```
Perhaps this should also take into account that current dir might be inside a server instead of the project proper.

Artur-: Maybe the dev server logic should be moved to AbstractConfiguration then?

caalador: Probably should. The `getProjectFolder` is a new addition and I didn't connect the dots 3 weeks ago.

Artur-: Updated

caalador (approved):

PR #15590, opened 2022-12-29
feat: Find where component is created in code by clicking in UI

Artur-:


flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Checks if the application is running in production mode.
+     *
+     * When unsure, reports that production mode is true so tracking does not
+     * take place in production.
+     *
+     * @return true if in production mode or the mode is unclear, false if in
+     *         development mode
+     **/
+    private static boolean isProductionMode() {
```
knoobie, line 142: Should the result of this method be cached as field to make sure all the lookups / code below is only executed once in production mode? Otherwise all component instantiations with "ComponentTracker.trackCreate(this);" could really slow down the whole app in the long run

Artur-:
Will try to split this a bit more into smaller, hopefully testable, pieces

Artur-:
This is how it works

https://user-images.githubusercontent.com/260340/213159928-256ae566-9b06-4caf-8984-f15062a7c4af.mov

mshabarov (approved):
Tested with a fresh starter from start.vaadin.com and was very exciting to see magic of this feature. Super cool! It opened my Intellij IDE and lead me through the codes. I haven't noticed any weirdness, everything worked as I expected. LGTM!

PR #15593, opened 2022-12-30
feat: Add convenience method to close all Dialogs

TatuLund:


mcollovati:
IMO, Flow should not depend on Flow Components
Couldn't this be a utility in `vaadin-dialog-flow`?

TatuLund:
> IMO, Flow should not depend on Flow Components

Ok.

PR #15594, opened 2022-12-30
feat: Track where components are created and attached

Artur-:


flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
```diff
@@ -0,0 +1,174 @@
+package com.vaadin.flow.component.internal;
```
mcollovati, line 1: License header missing

flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public static void trackAttach(Component component) {
+        if (isProductionMode()) {
+            return;
+        }
+        StackTraceElement[] stack = Thread.currentThread().getStackTrace();
+
+        // In most cases the interesting attach call is found in the same class
+        // where
+        // the component was created and not e.g. in a generic layout class
```
mcollovati, line 97: nit: can be reformatted to take a single line?

flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * Checks if the application is running in production mode.
+     *
+     * When unsure, reports that production mode is true so tracking does not
+     * take place in production.
+     *
+     * @return true if in production mode or the mode is unclear, false if in
+     *         development mode
+     **/
+    private static boolean isProductionMode() {
```
mcollovati, line 148: Do we need to synchronize here? Or use an AtomicBoolean?

flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
```diff
@@ -0,0 +1,121 @@
+package com.vaadin.flow;
```
mcollovati, line 1: License header missing

flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    @Before
+    public void setup() throws Exception {
+        Field prodMode = ComponentTracker.class
+                .getDeclaredField("productionMode");
+        prodMode.setAccessible(true);
+        prodMode.set(null, false);
+    }
```
mcollovati, line 41: It would be good to add an `@After` method to restore the `productionMode` value, so that other tests running in the same JVM will not be affected by component tracking

flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
```diff
@@ -0,0 +1,121 @@
...
+import com.vaadin.flow.component.HasComponents;
+import com.vaadin.flow.component.Tag;
+import com.vaadin.flow.component.internal.ComponentTracker;
+
+/**
+ * Note that this is intentionally in the "wrong" package as internal packages
+ * are excluded in the tracker.
+ *
+ * Note that if you reformat this file, the tests will need to be adjusted as
+ * they track line numbers.
```
mcollovati, line 20: What about adding a `@formatter:off` around the whole class definition? It may help to avoid failures in case of accidental formatting inside the IDE

mcollovati (changes requested):
A question: is it intended to let do all the cleanups by the GC and to avoid removing tracked data on components detach?

Artur-:
> A question: is it intended to let do all the cleanups by the GC and to avoid removing tracked data on components detach?

The create location cannot be cleaned eagerly as we do not know when the component will no longer be used (except when GCed). It is not clear that the attach location becomes uninteresting once you do detach either. Do you see a problem with letting GC handling it?

flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
Artur-: I don’t think so. If two threads would happen to run it at once, they should reach the same result and just overwrite the same value?

mcollovati: Yes, makes sense.

mcollovati:
> > A question: is it intended to let do all the cleanups by the GC and to avoid removing tracked data on components detach?
> 
> The create location cannot be cleaned eagerly as we do not know when the component will no longer be used (except when GCed). It is not clear that the attach location becomes uninteresting once you do detach either. Do you see a problem with letting GC handling it?

No, no evident problems. Just curiosity :)

flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
Artur-: I would rather not as `@formatter:off` tend to cause problems in the long run. As there is a formatter checker, this should really only break if the test class is modified or if a new version of the formatter does something differently (then we can update a number or two)

flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
Artur-: Yes, done

flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
Artur-: Done

flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
Artur-: Done

flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
Artur-: Done

mcollovati (approved):

PR #15598, opened 2023-01-01
feat: add API to fetch page direction

### LINKED ISSUE #15597: Consider adding API to get the page direction (LTR/RTL)
### Describe your motivation

In the applications that need to support RTL languages, sometimes it is needed to get the page direction programmatically to decide about representation of some controls or manipulate styles according to the RTL or LTR direction of the page. 

The API to set the Direction is obvious:
```
UI.getCurrent().setDirection(Direction.RIGHT_TO_LEFT);
```
But apparently, there is no official API to get/fetch the page direction.

Proper solution for this requirement is to have an API for getting the current direction of the page at anytime, preferably, as an API in the UI or the Page class. 

Users can execute the `return document.dir` Js code against the current UI and fetch the direction:
```
UI.getCurrent().getPage()
            .executeJs("return document.dir")
            .then(String.class, dir -> doSomethingWith(dir));
```
But, it would be much nicer to officially support this. 

### Additional context

N/A

taefi:
## Description

Add API to fetch the document.dir from browser
and pass it to the provided callback.

Fixes #15597 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
@@ -675,11 +675,10 @@ public class Page implements Serializable {
      *            to be notified when the direction is resolved.
      */
     public void fetchPageDirection(SerializableConsumer<Direction> callback) {
-        UI.getCurrent().getPage().executeJs("return document.dir")
-                .then(String.class, dir -> {
-                    Direction direction = getDirectionByClientName(dir);
-                    callback.accept(direction);
-                });
+        executeJs("return document.dir").then(String.class, dir -> {
+            Direction direction = getDirectionByClientName(dir);
+            callback.accept(direction);
+        });
     }
 
     private Direction getDirectionByClientName(String directionClientName) {
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
@@ -24,12 +24,14 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.function.Function;
 
 import com.vaadin.flow.component.ClientCallable;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentEvent;
+import com.vaadin.flow.component.Direction;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.dependency.JavaScript;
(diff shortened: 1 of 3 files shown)


taefi:
This change has no dependency on the new features of the platform, and it can be back-ported to all the versions.

flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
```diff
@@ -655,4 +657,35 @@ public void fetchCurrentURL(SerializableConsumer<URL> callback) {
...
+     * <p>
+     * In case you need more control over the execution you can use
+     * {@link #executeJs(String, Serializable...)} by passing
+     * {@code return document.dir}.
+     *
+     * @param callback
+     *            to be notified when the direction is resolved.
+     */
+    public void fetchPageDirection(SerializableConsumer<Direction> callback) {
+        UI.getCurrent().getPage().executeJs("return document.dir")
```
mcollovati, line 678: Do we need `UI.getCurrent().getPage()` here, since the method is not static? Could it be simply a call to `executeJs`?
```suggestion
        executeJs("return document.dir")
```

mcollovati (changes requested):
Just a thought: it seems a bit odd to me that we set the direction on `UI`, but fetch it from `Page`.

Would it make sense to store direction value in `UI` or in `UIInternals` (as we do for title, for example) and have also a `UI.getDirection()` method?

taefi:
> Would it make sense to store direction value in `UI` or in `UIInternals` (as we do for title, for example) and have also a `UI.getDirection()` method?

IMO, keeping the state for those kind of stuff that can be changed on the client-side (for instance with a custom Js script)  makes it hard to keep it in sync with the corresponding value on the server side, and would make it complicated to cover corner cases, and it becomes unreliable as the result. Reading it from the browser on demand would be much simpler. But, I'm open to new ideas. 

> Just a thought: it seems a bit odd to me that we set the direction on `UI`, but fetch it from `Page`.

It is a little hard to decide where to put an API when it comes to decide Page and UI, and to be honest, I also struggled over this a little, but in the end I decided to put into the Page class since it was technically close to `fetchCurrentUrl` method. But, there's no heart feelings ;) I can move it to the UI class, as the added method and the correspondent tests aren't complicated and dependent on something in the Page class specifically.

flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
taefi: This was a left over from my draft codes, good catch!

mcollovati:
If we don't hold state and always fetch the information from the client, I think it is fine to have the method on Page class

mcollovati (approved):

ZheSun88:
this is  a new feature, which should not be picked to maintenance branch.  23.2 and 23.3

PR #15600, opened 2023-01-02
fix: Add tsconfig to vaadin-dev-server to avoid IDE errors

Artur-:
Experimental decorators are off by default for tsc

### DIFF
--- a/vaadin-dev-server/tsconfig.json
+++ b/vaadin-dev-server/tsconfig.json
@@ -1,0 +1,22 @@
+// This is included to allow decorators in TS files
+{
+  "compilerOptions": {
+    "sourceMap": true,
+    "jsx": "react-jsx",
+    "inlineSources": true,
+    "module": "esNext",
+    "target": "es2020",
+    "moduleResolution": "node",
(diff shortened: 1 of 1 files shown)


vaadin-dev-server/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "target": "es2020",
+    "moduleResolution": "node",
+    "strict": true,
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitReturns": true,
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "experimentalDecorators": true,
```
MarcinVaadin, line 17: And why do we need those decorators in vaadin-dev-server?

Artur-: Used e.g. here https://github.com/vaadin/flow/blob/master/vaadin-dev-server/src/main/resources/META-INF/frontend/vaadin-dev-tools.ts#L929

PR #15602, opened 2023-01-02
feat: Allow finding a server side element based on nodeid/appid from JS

Artur-:
Useful if you need to send a component reference outside the GWT code


flow-server/src/main/java/com/vaadin/flow/component/ComponentReference.java
```diff
@@ -0,0 +1,55 @@
+package com.vaadin.flow.component;
+
+import com.vaadin.flow.server.VaadinSession;
+
+/**
+ * A component reference, typically from the browser.
+ */
+public class ComponentReference {
```
Legioth, line 8: I'm not sure about the benefits of wrapping individual values inside a value object instead of just passing each of them as a separate parameter to `findComponent` (especially since I also put into question the explicit session reference in a separate comment)

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
```diff
@@ -647,4 +651,39 @@ public static Router getRouter(HasElement component) {
...
+     * @throws IllegalArgumentException
+     *             if the component was not found
+     */
+    public static Component findComponent(ComponentReference componentReference)
+            throws IllegalArgumentException {
+        assert componentReference.getSession().hasLock();
+
+        String appId = componentReference.getAppId();
+        int nodeId = componentReference.getNodeId();
+        Optional<UI> ui = componentReference.getSession().getUIs().stream()
```
Legioth, line 669: There's some thread safety concerns with searching through an arbitrary session that might not be locked by the current thread. I'm wondering if it would be clearer to move this API to `VaadinSession` instead to make it more explicit that you're supposed to only operate on a session that you already have a lock for (and also to assert the lock in the beginning of the method).

flow-server/src/main/java/com/vaadin/flow/component/ComponentReference.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private int nodeId;
+    private String appId;
+    private VaadinSession session;
+
+    /**
+     * Creates a new reference.
```
Legioth, line 15: Could somewhere document how users are typically supposed to find node id and app id values

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
Artur-: Moved to session

flow-server/src/main/java/com/vaadin/flow/component/ComponentReference.java
Artur-: Removed now that it had only two values

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -1160,4 +1164,42 @@ public long getLastLocked() {
...
+        StateNode node = ui.get().getInternals().getStateTree()
+                .getNodeById(nodeId);
+        if (node == null) {
+            throw new IllegalArgumentException(
+                    "Unable to find the component for node " + nodeId
+                            + " in app " + appId);
+        }
+
+        return ComponentMapping.getComponent(node).orElseThrow(
+                () -> new IllegalArgumentException("No component found"));
```
caalador, line 1202: As the node might be for a pure Element also should there be a `findElement(appId, nodeId)` as the client side doesn't know or care what the client-element is on the server?

Legioth: Good point. I would almost take it one step further - only offer `findElement` and then have the user perform the last step with `ComponentMapping` separately (which depending on the case might also use e.g. `ComponentUtil::findParentComponent` or `ComponentUtil::getInnermostComponent` instead).

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -1160,4 +1164,42 @@ public long getLastLocked() {
     public long getLastUnlocked() {
         return lastUnlocked;
     }
+
+    /**
+     * Finds the given component in the session.
+     * <p>
+     * The ids are typically acquired in the browser where appId refers to the
+     * UI instance and nodeId to the component inside that UI.
```
Legioth, line 1172: Could somewhere document how users are typically supposed to find node id and app id values, i.e. referencing the newly added `getNodeId` JS function and pointing out that app id is a key in `window.Vaadin.Flow.clients`.

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
Artur-: All of the use cases I have in my mind are interested in components and not at all in elements. The problem is rather then that you do not know on the client side if an element is a component or if you should keep going upwards

Artur-: Changed to find elements, maybe the info about components can be sent separately to the browser (in dev mode only) if needed

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
Artur-: Done

flow-test-generic/src/main/java/com/vaadin/flow/testutil/ClassesSerializableTest.java
```diff
@@ -161,6 +161,7 @@ protected Stream<String> getExcludedPatterns() {
                 "com\\.vaadin\\.flow\\.router\\.RoutePathProvider",
                 "com\\.vaadin\\.flow\\.router\\.RouteNotFoundError\\$LazyInit",
                 "com\\.vaadin\\.flow\\.router\\.internal\\.RouteSegment\\$RouteSegmentValue",
+                "com\\.vaadin\\.flow\\.component\\.ComponentReference",
```
caalador, line 164: This should be removed as we do not have the ComponentReference class anymore after all the changes.

caalador (approved):

Issue #15529, opened 2022-12-19
Add support index.tsx in addition to index.ts/js


marcushellberg:
### Describe your motivation

In TypeScript-based React apps, the convention is to use an `index.tsx` file. We want to keep Hilla as standard as possible and therefore want to support using JSX in the index file.

### Describe the solution you'd like
Update the bootstrap code to also consider `index.tsx` valid. 

### Describe alternatives you've considered
There are two alternatives: 
- Using an `index.ts` file and using `React.createElement`. This is what we do today. It looks outdated and is difficult to expand if you need more providers etc
- Loading a secondary file like `App.tsx` from the `index.ts` file where the actual app bootstrapping happens. This works, but does not follow React conventions.

### Additional context
-

MarcinVaadin:
Maybe @cromoteca or other Hilla member could give some more information / suggestions on that?

cromoteca:
It could be useful to provide a test example and add it to `TaskGenerateBootstrapTest`: it might be a bit complicated to test this as the whole generated code would only make sense in a Hilla+React application, so it's important to clearly define which output we expect from Flow. Is it enough to check for the existence of `index.tsx` and import it as `../index`?

MarcinVaadin:
I've created quick PR for @cromoteca suggestion - https://github.com/vaadin/flow/pull/15621 - but without test. There is also `TaskGenerateIndexTs` which generates JS code - how does it compares to .tsx files?

cromoteca:
I don't understand all the details there, but it seems to generate a default `index.ts` if no custom file is available. So `index.tsx` should be added to the files that to check for.

PR #15621, opened 2023-01-03
feat: support for index.tsx

MarcinVaadin:
See #15529 for details

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateIndexTs.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateIndexTs.java
@@ -23,6 +23,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
 
 import org.apache.commons.io.IOUtils;
 
@@ -64,11 +65,11 @@ public class TaskGenerateIndexTs extends AbstractTaskClientGenerator {
 
     @Override
     protected boolean shouldGenerate() {
-        File indexTs = new File(frontendDirectory, INDEX_TS);
-        File indexJs = new File(frontendDirectory, INDEX_JS);
-        File indexTsx = new File(frontendDirectory, INDEX_TSX);
-        compareActualIndexTsOrJsWithIndexTemplate(indexTs, indexJs);
-        return !indexTs.exists() && !indexJs.exists() && !indexTsx.exists();
+        return Arrays.asList(INDEX_TSX, INDEX_TS, INDEX_JS).stream()
+                .map(type -> new File(frontendDirectory, type))
+                .filter(File::exists)
+                .peek(this::compareActualIndexWithIndexTemplate).findAny()
+                .isEmpty();
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


Artur- (dismissed):

cromoteca (changes requested):
As @MarcinVaadin pointed out in the issue, `TaskGenerateIndexTs` should also probably be taken into account.

MarcinVaadin:
What is the purpose of
```
private void compareActualIndexTsOrJsWithIndexTempalate(File indexTs,
            File indexJs) {
        if (indexTs.exists() || indexJs.exists()) {
            File indexFileExist = indexTs.exists() ? indexTs : indexJs;
            String indexContent = null;
            String indexTemplate = null;
            try {
                indexContent = IOUtils.toString(indexFileExist.toURI(), UTF_8);
                indexTemplate = getFileContent();
            } catch (IOException e) {
                log().warn("Failed to read file content", e);
            }
            if (indexContent != null && !indexContent.equals(indexTemplate)) {
                UsageStatistics.markAsUsed(Constants.STATISTIC_ROUTING_CLIENT,
                        Version.getFullVersion());
            }
        }
    }
```
it was supposed to compare file with some template but on failure it just reports some statistic?

Artur-:
It detects that a custom index.ts is used and marks this in the stats as "client side routing"

cromoteca:
> What is the purpose of `TaskGenerateIndexTs.compareActualIndexTsOrJsWithIndexTempalate` ?

You can also fix the typo in the method name in the process :)

MarcinVaadin:
Ok... and what's the point of those stats? Or should .tsx be also compared with that template?

Artur-:
To know if client routing is used.. :)

Artur-:
It allows distinguishing between Flow and Hilla usage. I think. Kind of. I don't think it is relevants for tsx/jsx

Artur-:
> As @MarcinVaadin pointed out in the issue, TaskGenerateIndexTs should also probably be taken into account.

That only generates the default index.ts if no index.js or index.ts exists. Nothing to do with this I think

MarcinVaadin:
> > As @MarcinVaadin pointed out in the issue, TaskGenerateIndexTs should also probably be taken into account.
> 
> That only generates the default index.ts if no index.js or index.ts exists. Nothing to do with this I think

But if .tsx only exists .ts will be autogenerated I guess. So maybe I will just add checking if .tsx also does not exist?

Artur-:
Right. It should not do anything if one of index.js/index.ts/index.tsx exist

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateIndexTs.java
```diff
@@ -108,7 +110,7 @@ static String ensureValidRelativePath(String relativePath) {
         return relativePath;
     }
 
-    private void compareActualIndexTsOrJsWithIndexTempalate(File indexTs,
+    private void compareActualIndexTsOrJsWithIndexTemplate(File indexTs,
             File indexJs) {
         if (indexTs.exists() || indexJs.exists()) {
```
cromoteca, line 115: If `index.tsx` is used, it will not be detected in stats. Is that the intended behavior?

MarcinVaadin: Probably not, but if there is no "template" should I treat any .tsx as custom?

cromoteca (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateIndexTs.java
```diff
@@ -63,10 +65,11 @@ protected File getGeneratedFile() {
 
     @Override
     protected boolean shouldGenerate() {
-        File indexTs = new File(frontendDirectory, INDEX_TS);
-        File indexJs = new File(frontendDirectory, INDEX_JS);
-        compareActualIndexTsOrJsWithIndexTempalate(indexTs, indexJs);
-        return !indexTs.exists() && !indexJs.exists();
+        return Arrays.asList(INDEX_TSX, INDEX_TS, INDEX_JS).stream()
+                .map(type -> new File(frontendDirectory, type))
```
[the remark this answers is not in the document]
MarcinVaadin: @sonatype-lift ignore

PR #15624, opened 2023-01-05
feat: Improve Spring Native support

Artur-:
Registers project @Routes, component classes and more

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinBeanFactoryInitializationAotProcessor.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinBeanFactoryInitializationAotProcessor.java
@@ -1,0 +1,69 @@
+package com.vaadin.flow.spring;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.reflections.Reflections;
+import org.springframework.aot.hint.MemberCategory;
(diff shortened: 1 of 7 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinBeanFactoryInitializationAotProcessor.java
```diff
@@ -0,0 +1,69 @@
...
+                        .addAll(reflections.getTypesAnnotatedWith(Route.class));
+                routeTypes.addAll(
+                        reflections.getTypesAnnotatedWith(RouteAlias.class));
+                for (var c : routeTypes) {
+                    reflection.registerType(c, memberCategories);
+                    resources.registerType(c);
+                }
+                for (var c : reflections.getSubTypesOf(Component.class))
+                    reflection.registerType(c, memberCategories);
+                for (var c : reflections.getSubTypesOf(HasErrorParameter.class))
```
caalador, line 54: Guess we would also need at least `HasUrlParameter` as it is called through `ReflectTools.getGenericInterfaceType(target, HasUrlParameter.class);` and `Converter` `ReflectTools.getGenericInterfaceTypes(converterType, Converter.class);`

Artur-: Added

caalador (approved):

PR #15629, opened 2023-01-05
fix: replace any 'Frontend/' with './'

caalador:
Stats.json records some handled files as `Frontend/...` where as the requested file may be in the format `./...`
Use each `Frontend/` always as `./`


mshabarov (approved):

Artur-:
Does this fix something? All tests pass before and after

mshabarov (approved):

PR #15631, opened 2023-01-05
feat: Load theme using Vite

Artur-:
This changes theme loading so that theme CSS files are imported using
```
import 'theme/mytheme/styles.css';
```

Using this instead of 
```
import someCss from 'theme/mytheme/styles.css?inline';
// inject the css somehow into the DOM
```

allows the default hot replace functionality of Vite to be used.

Note that a followup PR is still needed to actually enable hot replace because the styles are still also loaded for shadow root injection, which currently prevents hot replace from being used.

In production mode, Vite will now bundle the theme `styles.css` and put it in `VAADIN/build/indexcss-something.css`, which means that URLs that refer to files in `src/main/webapp` or similar need to be rewritten to use `../..` which cancel out `VAADIN/build`.

Paths inside packaged themes also need to be rewritten the same way


mcollovati:
@Artur- is this PR still valid, but currently in progress?

Artur-:
Still needs some work with loading parent themes the same way so that overrides work properly

Artur-:
This should break exported web components by accidentally applying the theme globally, wonder why no tests break

Artur-:
Test was added in #16439 so now this should work

PR #15633, opened 2023-01-05
fix: Do not register anonymous classes with Spring aot

Artur-:
It is not needed, and causes a null pointer exception. See https://github.com/spring-projects/spring-framework/issues/29774


Artur-:
This really needs one test project that does a native compilation for any kind of app

caalador (approved):

PR #15646, opened 2023-01-09
fix: Don't generate frontend files with devBundle

### LINKED ISSUE #15640: If using express bundle, do not generate frontend files
When running with a dev bundle we should not generate
the frontend files 
- package.json
- vite.*.ts
- types.d.ts
- tsconfig.json

if they are not available, as they are not needed.

caalador:
If running application using devBundle
do not generate frontend files like,
package.json, vite.*.ts etc.

Closes #15640

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -153,7 +153,11 @@ public class NodeTasks implements FallibleCommand {
         if (frontendDependencies != null) {
             addGenerateServiceWorkerTask(options,
                     frontendDependencies.getPwaConfiguration());
-            addGenerateTsConfigTask(options);
+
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -153,7 +153,11 @@ public NodeTasks(Options options) {
         if (frontendDependencies != null) {
             addGenerateServiceWorkerTask(options,
                     frontendDependencies.getPwaConfiguration());
-            addGenerateTsConfigTask(options);
+
+            if (options.productionMode || options.isEnableDevServer()
+                    || options.isDevBundleBuild()) {
```
mshabarov, line 158: `isDevBundleBuild()` method name makes me doubt if it signal the express build mode is ON or the Flow is going to build a new dev bundle. I know the right answer, but maybe we could change the name in the future :)

mshabarov (approved):

PR #15661, opened 2023-01-10
chore: Remove flow-polymer-template from flow

Artur-:
flow-polymer-template is supposed to be opt-in


caalador (approved):

caalador:
Why didn't gradle fail as it should expect `"flow-polymer-template-", ` in `testIncludeExclude`

PR #15684, opened 2023-01-13
feat: Add support for reusable theme in Express Build mode

mshabarov:
- Serves styles and assets also for the packaged custom theme available from frontend/generated/jar-resources/themes
- Calculates the hash of the theme.json for the packaged theme and records it to the bundle's stats.json
- Checks if the hashes from the packaged themes match the corresponding hashes in the stats.json
- Rebuilds the bundle if the packaged themes are added/updated (hashes don't match).

Fixes https://github.com/vaadin/flow/issues/15691


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyFrontendFiles.java
```diff
@@ -108,12 +124,46 @@ public void execute() {
         long ms = (System.nanoTime() - start) / 1000000;
         log().info("Visited {} resources. Took {} ms.",
                 resourceLocations.size(), ms);
+
+        calculateHashForPackagedThemes(targetDirectory);
+    }
+
+    private void calculateHashForPackagedThemes(File targetDirectory) {
```
caalador, line 131:
```suggestion
    private void calculateHashesForThemeJson(File pacakgedThemesDirectory) {
```
Update name to reflect what is hashed and from where.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyFrontendFiles.java
```diff
@@ -108,12 +124,46 @@ public void execute() {
...
+                    try {
+                        String themeJsonContent = FileUtils.readFileToString(
+                                themeJson, StandardCharsets.UTF_8);
+                        JsonObject json = Json.parse(themeJsonContent);
+                        if (json.hasKey(ASSETS)) {
+                            String contentHash = StringUtil.getHash(
+                                    themeJsonContent, StandardCharsets.UTF_8);
+                            json.put("hash", contentHash);
+                            FileUtils.write(themeJson, json.toJson(),
+                                    StandardCharsets.UTF_8);
```
caalador, line 149: If we put the hash into theme.json here doesn't that mean that when vite is running it will generate the hash from the theme.json containing the hash?
Except we don't seem to calculate the hash in vite for theme.json only grab the hash if available.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -131,6 +134,63 @@ protected static boolean needsBuildInternal(Options options,
...
+        return false;
+    }
+
+    private static boolean packagedThemeAddedOrUpdated(Options options,
+            JsonObject statsJson) {
+        File packagedThemesFolder = new File(
+                options.getJarFrontendResourcesFolder(), "themes");
+        if (!packagedThemesFolder.exists()) {
+            return false;
+        }
```
caalador, line 150: Have you checked that we would have this initially at all?
As the needsBuild is executed before any tasks are run I would expect that for the initial execution no frontend resources are copied yet and this will then always return false.

Also the subsequent executions will be one step off, so if you had a jar that is update it wouldn't be noted until the second time the application is executed.

So this should probably check against the actual jar files and calculate the hash for those instead of reading from disk.

flow-tests/test-express-build/test-parent-theme-express-build/src/main/resources/vaadin-featureflags.properties
```diff
@@ -0,0 +1 @@
+com.vaadin.experimental.expressBuild=true
```
caalador, line 1: This could come from the test-dev-bundle as then it would be easier to clean up.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyLocalFrontendFiles.java
```diff
@@ -100,6 +117,22 @@ static void createTargetFolder(File target) {
         }
     }
 
+    static boolean notExcluded(File source, String[] relativePathExclusions,
```
caalador, line 120: `excludeFile` would perhaps sound clearer or if having the check this way name it `keepFile`

caalador (commented):
I think there is a timing issue with the bundle check at the moment.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyFrontendFiles.java
mshabarov: This should be changed as we discussed offline: hashes should be generated in `statsExtracterPlugin` and written to the `stats.json`, then `TaskRunDevBundleBuild` should looks into JAR resources for themes, calculate the hash and compare it with the one in `stats.json`.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -643,7 +645,15 @@ public static String getJarResourceString(String jarImport) {
         }
     }
 
-    private static File getJarResourcesFolder(File frontendDirectory) {
+    /**
+     * Get the folder the front-end resources from JAR dependencies are copied
+     * into.
```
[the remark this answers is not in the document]
mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyFrontendFiles.java
mshabarov: Not needed anymore, because this method has been removed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov: Done. It checks the themes from JARs now.

flow-tests/test-express-build/test-parent-theme-express-build/src/main/resources/vaadin-featureflags.properties
mshabarov: Removed

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyLocalFrontendFiles.java
mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -131,6 +140,75 @@ protected static boolean needsBuildInternal(Options options,
...
+                                    throw new IllegalStateException(
+                                            "Packaged theme folders structure is incorrect, should have META-INF/resources/themes/[theme-name]/");
+                                }
+                                String themeName = matcher.group(1);
+                                String hash = StringUtil.getHash(content,
+                                        StandardCharsets.UTF_8);
+                                packagedThemeHashes.put(themeName, hash);
+                            }
+                        });
+                    }
```
caalador, line 187: This could be it's own method just for readability

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -131,6 +140,75 @@ protected static boolean needsBuildInternal(Options options,
...
+                                if (!matcher.find()) {
+                                    throw new IllegalStateException(
+                                            "Packaged theme folders structure is incorrect, should have META-INF/resources/themes/[theme-name]/");
+                                }
+                                String themeName = matcher.group(1);
+                                String hash = StringUtil.getHash(content,
+                                        StandardCharsets.UTF_8);
+                                packagedThemeHashes.put(themeName, hash);
+                            }
+                        });
```
caalador, line 186:
```suggestion
                        for (String themeJson : themeJsons) {
                            byte[] byteContent = jarContentsManager
                                    .getFileContents(jarFile, themeJson);
                            String content = IOUtils.toString(byteContent,
                                    "UTF-8");
                            content = content.replaceAll("\\r\\n", "\n");

                            JsonObject themeJsonContent = Json.parse(content);
                            if (themeJsonContent.hasKey(Constants.ASSETS)) {
                                Matcher matcher = THEME_PATH_PATTERN
                                        .matcher(themeJson);
                                if (!matcher.find()) {
                                    throw new IllegalStateException(
                                            "Packaged theme folders structure is incorrect, should have META-INF/resources/themes/[theme-name]/");
                                }
                                String themeName = matcher.group(1);
                                String hash = StringUtil.getHash(content,
                                        StandardCharsets.UTF_8);
                                packagedThemeHashes.put(themeName, hash);
                            }
                        }
```
Explicit for loop makes it easier for reading than the for each lambda.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -131,6 +140,75 @@ protected static boolean needsBuildInternal(Options options,
...
+        JsonObject hashesInStats = statsJson.getObject("themeJsonHashes");
+        if (hashesInStats == null && !packagedThemeHashes.isEmpty()) {
+            getLogger().info("Found newly added packaged custom themes.");
+            return true;
+        }
+        for (Map.Entry<String, String> themeHash : packagedThemeHashes
+                .entrySet()) {
+            if (hashesInStats.hasKey(themeHash.getKey())) {
+                if (!hashesInStats.getString(themeHash.getKey())
+                        .equals(themeHash.getValue())) {
```
caalador, line 199: The if clauses could be combined or then changed to if { } else if { } e.g.
```
if (!hashesInStats.hasKey(themeHash.getKey())) {
    getLogger().info("Found newly added packaged custom theme '{}'.", themeHash.getKey());
    return true;
} else if (!hashesInStats.getString(themeHash.getKey()).equals(themeHash.getValue())) {
    getLogger().info("Found updated package custom theme '{}'.", themeHash.getKey());
    return true;        
}
```

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuildTest.java
```diff
@@ -853,4 +862,220 @@ public void frontendFileHashMissmatch_bundleRebuild() throws IOException {
...
+        }
+    }
+
+    @Test
+    public void reusedTheme_previouslyAddedThemes_justAddedNewTheme_bundleRebuild()
+            throws IOException {
+        createPackageJsonStub(BLANK_PACKAGE_JSON_WITH_HASH);
+
+        File jarWithTheme = TestUtils
+                .getTestJar("jar-with-theme-json-and-assets.jar");
```
caalador, line 985: One simplification that could be made later on would be to not have the prepackaged jars, but to generate a test jar with wanted content so that it would be easy to update the content when it changes.
See `StaticFileServerTest:generateZipArchive` for inspiration on the idea. But this can be its own ticket for later.

caalador (approved):

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuildTest.java
mshabarov: https://github.com/vaadin/flow/issues/15758

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov: Done

flow-tests/test-express-build/test-reusable-theme-express-build/src/main/resources/META-INF/resources/themes/reusable-theme/theme.json
```diff
@@ -0,0 +1,8 @@
+{
+  "importCss": ["@fortawesome/fontawesome-free/css/all.css"],
```
knoobie, line 2: Note: all projects that I currently have with this kind of config: throws vite warnings if the dev server recompiles due to changes. With a message like "Default and named Imports from CSS files are deprecated."

mshabarov: Thanks for noticing this, we are probably missing `?inline` here https://github.com/vaadin/flow/blob/5734f850c60b9535f9fa1406085c7791aa19b9ab/flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js#L221

Issue #13646, opened 2022-04-28
Do not connect directly to the Spring Boot livereload server


Artur-:
### Describe your motivation

When you start a Spring Boot application using Spring boot dev tools, it automatically starts a livereload server on port 35729. This port is hard coded so if you start two Spring Boot applications, you will have only one livereload server listening. 

Flow connects directly to the port from the browser, so if you run two applications locally then both will connect to the same livereload server. Live reloading will work for one and not for the other.

Apparently also if you configure the server to use ssl then we try to use a secure websocket connection (`wss://`) for port 35729 but this does not work (#13632).

### Describe the solution you'd like

Instead of connecting to the livereload server from the browser, we should use the same debug window connection we already have. On the server, we should listen to reload events from Spring Boot like the LiveReloadServer does: https://github.com/spring-projects/spring-boot/blob/fdb1010cbc75517f511d4ab82de7d8f0ee058849/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/LocalDevToolsAutoConfiguration.java#L193-L198

When an event arrives, we send the "reload" command using the debug window connection and react as we do today.

### Additional context

This will remove the need for one websocket connection and make livereload work even if you run multiple applications locally

PR #15689, opened 2023-01-15
fix: Use the servlet host and port for the Vite websocket connection

### LINKED ISSUE #14744: Dev mode with Vite doesn't support https
### Describe your motivation

Currently, the Vaadin dev mode doesn't support https.

Vite can be configured to start a TLS endpoint like this:

```javascript
// vite.config.ts

const customConfig: UserConfigFn = (env) => ({
    // Here you can add custom Vite parameters
    // https://vitejs.dev/config/
    server: {
        host: "service-localdev.example.com",  // resolves to localhost
        https: {
            key: fs.readFileSync("../assets/tls/localdev/service-localdev.example.com.key"),
            cert: fs.readFileSync("../assets/tls/localdev/service-localdev.example.com.crt")
        }
    }
});
```

which is also recognized by Vaadin:

```
2022-10-05 15:06:49.856  INFO [                ] c.v.b.devserver.AbstractDevServerRunner  : Started Vite. Time: 2092ms
2022-10-05 15:06:49.857  INFO [                ] c.v.b.devserver.DevServerOutputTracker   :   ➜  Local:   https://service-localdev.example.com:57132/ui/VAADIN/
2022-10-05 15:06:49.857  INFO [                ] c.v.b.devserver.DevServerOutputTracker   :   ➜  Network: https://10.67.0.214:57132/ui/VAADIN/
```

But unfortunately, Vaadin uses a hardcoded `DEV_SERVER_HOST` of "http://127.0.0.1" (see [`AbstractDevServerRunner`](https://github.com/vaadin/flow/blob/23.2.3/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/AbstractDevServerRunner.java#L606-L615)), which leads to a failing connection establishment:

```
java.net.SocketException: Unexpected end of file from server
	at java.base/sun.net.www.http.HttpClient.parseHTTPHeader(HttpClient.java:866) ~[na:na]
	at java.base/sun.net.www.http.HttpClient.parseHTTP(HttpClient.java:689) ~[na:na]
	at java.base/sun.net.www.http.HttpClient.parseHTTPHeader(HttpClient.java:863) ~[na:na]
	at java.base/sun.net.www.http.HttpClient.parseHTTP(HttpClient.java:689) ~[na:na]
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1615) ~[na:na]
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1520) ~[na:na]
	at java.base/java.net.HttpURLConnection.getResponseCode(HttpURLConnection.java:527) ~[na:na]
	at com.vaadin.base.devserver.AbstractDevServerRunner.serveDevModeRequest(AbstractDevServerRunner.java:723) ~[vaadin-dev-server-23.2.3.jar:na]
	at com.vaadin.flow.server.StaticFileServer.serveStaticResource(StaticFileServer.java:243) ~[flow-server-23.2.3.jar:23.2.3]
	at com.vaadin.flow.server.VaadinServlet.serveStaticOrWebJarRequest(VaadinServlet.java:401) ~[flow-server-23.2.3.jar:23.2.3]
	at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:356) ~[flow-server-23.2.3.jar:23.2.3]
	at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:106) ~[vaadin-spring-23.2.3.jar:na]
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:764) ~[tomcat-embed-core-9.0.65.jar:4.0.FR]
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:227) ~[tomcat-embed-core-9.0.65.jar:9.0.65]
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162) ~[tomcat-embed-core-9.0.65.jar:9.0.65]
	at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53) ~[tomcat-embed-websocket-9.0.65.jar:9.0.65]
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:189) ~[tomcat-embed-core-9.0.65.jar:9.0.65]
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162) ~[tomcat-embed-core-9.0.65.jar:9.0.65]
```
 
It's a common use case and best practice to use https everywhere, even on localdev. In my case, I'm further required to use https, as I'm using SPNEGO (Kerberos) as an authentication scheme.

Vaadin should use the scheme and host of the Vite dev server.

I'm unsure if this is needed at all. If the vite dev server is used internally only, using http://localhost is fine. But unfortunately, the client itself is trying to connect to the dev server directly, which fails if the Spring Boot application is using https and the Vite dev server is only available via http://localhost:

grafik

So probably a cleaner approach would be to proxy each request to the Vite dev server through the Vaadin application.

### Additional context

In my example, the Spring Boot application is providing the TLS endpoint. But this is not a necessity: Another valid way to develop https-only locally is to use a local reverse proxy. Even then, the dev mode should be usable.

### LINKED ISSUE #14773: Vaadin with Vite does not respect custom host names in vite.config.ts
### Description of the bug
We tried to migrate to 23.2.x this week, but were having trouble because of vite. One reason was that we tend to run multiple interdependent dev environments on our machines. they make use of the local 127.x.x.x range, but are not necessarily running on 127.0.0.1. Also we use domain names for these ip's which reside in our (C:\Windows\System32\Drivers\network\)/etc/hosts file. We tried to use the domain name, or the specific ip address (webportfolio.local and 127.0.16.1 respectively). But although the browser is able to get at the resources, the DevModeHandler implementation called AbstractDevServerRunner in package com.vaadin.base.devserver uses a hardcoded reference to 127.0.0.1 on line 81:
```java
   private static final String DEV_SERVER_HOST = "http://127.0.0.1";
```
take any project and change customConfig in vit.config.ts to:
```typescript
const customConfig: UserConfigFn = (env) => ({
  // Here you can add custom Vite parameters
  // https://vitejs.dev/config/
    server: {
      host: '127.0.16.1', 
      https: true
    }
});
```
### Expected behavior
Either the AbstractDevServerRunner should make requests to the correct ip address. Or if the host should not be changed in vite.config.ts, then The browser should request resources from the correct localhost address. it is now requesting from the server we host the app from (wildfly is listening on 127.0.16.1).
### Actual behavior
AbstractDevServerRunner is requesting from another ip address, which results in connection issues, which result in exceptions and a not-working app. Or if vite serves from 127.0.0.1, The app starts, but there are ERR_CONNECTION_REFUSED errors because it's trying to access 127.0.16.1:<someport>/VAADIN/ instead of 127.0.0.1:<someport>/VAADIN/
### Versions:
    - Vaadin / Flow version: 23.2.3
    - Java version: 17.0.4
    - OS version: Windows 11 22H2
    - Browser version (if applicable): Chrome 105.0.5195.127 (Official Build) (64-bit)
    - Application Server (if applicable): Wildfly 26.1.1
    - IDE (if applicable):Eclipse  2022-09 (4.25.0) Build id: 20220908-1902
    - Development or production mode: development mode
    - Project from start.vaadin.com or activated in application: activated in application

Artur-:
This proxies the Vite websocket connection the same way we proxy http requests

Fixes #14744 
Fixes #14773


Artur-:
The checker plugin still connects to the wrong URL because it has its own websocket logic. When https://github.com/fi3ework/vite-plugin-checker/issues/200 is fixed it should also work correctly

Artur-:
Upgrading the plugin to latest fixed the ws url issue

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketConnection.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
```
mcollovati, line 2:
```suggestion
 * Copyright 2000-2023 Vaadin Ltd.
```

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketEndpoint.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
```
mcollovati, line 2:
```suggestion
 * Copyright 2000-2023 Vaadin Ltd.
```

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketProxy.java
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
```
mcollovati, line 2:
```suggestion
 * Copyright 2000-2023 Vaadin Ltd.
```

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketEndpointInitializer.java
```diff
@@ -0,0 +1,78 @@
...
+import jakarta.servlet.ServletContextListener;
+import jakarta.websocket.DeploymentException;
+import jakarta.websocket.server.ServerContainer;
+import jakarta.websocket.server.ServerEndpointConfig;
+
+/**
+ * Creates the websocket endpoint that the Vite client JS connects to.
+ */
+public class ViteWebsocketEndpointInitializer
+        implements ServletContextListener {
```
mcollovati, line 37: It looks like that the only usage of this class is in `ViteHandler` and it is instantiated manually.
What is the purpose of making it implement `ServletContextListener`?
Is it meant for manual registrations in `ServeltContext`?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketEndpoint.java
```diff
@@ -0,0 +1,60 @@
...
+    public void onOpen(Session session, EndpointConfig config) {
+        getLogger().debug("Browser connected to Vite proxy");
+        ViteWebsocketProxy proxy;
+        try {
+            Integer vitePort = (Integer) config.getUserProperties()
+                    .get(VITE_PORT);
+            proxy = new ViteWebsocketProxy(session, vitePort);
+            session.addMessageHandler(proxy);
+        } catch (Exception e) {
+            getLogger().debug("Error creating Vite proxy connection", e);
```
mcollovati, line 47: Should we log at `error` level to make it more evident why it is not working as expected?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketConnection.java
```diff
@@ -0,0 +1,87 @@
...
+            Consumer<String> onMessage)
+            throws InterruptedException, ExecutionException {
+        this.onMessage = onMessage;
+        String wsHost = ViteHandler.DEV_SERVER_HOST.replace("http://", "ws://");
+        URI uri = URI.create(wsHost + ":" + port + "/VAADIN/");
+        WebSocket clientWebSocket = HttpClient.newHttpClient()
+                .newWebSocketBuilder().subprotocols(subProtocol)
+                .buildAsync(uri, this).get();
+        getLogger().debug("Connecting to " + uri + " using the "
+                + clientWebSocket.getSubprotocol() + " protocol");
```
mcollovati, line 64: nit:
```suggestion
        getLogger().debug("Connecting to {} using the {} protocol", uri,
                clientWebSocket.getSubprotocol());
```

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketConnection.java
```diff
@@ -0,0 +1,87 @@
...
+                + " protocol");
+        Listener.super.onOpen(webSocket);
+    }
+
+    @Override
+    public CompletionStage<?> onText(WebSocket webSocket, CharSequence data,
+            boolean last) {
+        // Message from Vite
+        String msg = data.toString();
+        getLogger().debug("Message from Vite: " + msg);
```
mcollovati, line 79: nit:
```suggestion
        getLogger().debug("Message from Vite: {}", msg);
````

File: vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketConnection.java [hunk 8]
```diff
@@ -0,0 +1,87 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.vaadin.base.devserver.ViteHandler;
+
+/**
+ * Communicates with a Vite server through a websocket connection.
+ */
+public class ViteWebsocketConnection implements Listener {
+
+    private Consumer<String> onMessage;
```
mcollovati, line 36: nit: can be final
```suggestion
    private final Consumer<String> onMessage;
```

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketProxy.java
```diff
@@ -0,0 +1,86 @@
...
+
+    protected Logger getLogger() {
+        return LoggerFactory.getLogger(getClass());
+    }
+
+    @Override
+    public void onMessage(String message) {
+        getLogger().debug("Got message from browser: " + message);
+        try {
+            CompletableFuture<WebSocket> send = clientWebSocket.get()
```
mcollovati, line 75: Where is `clientWebSocket` supposed to be initialized?
When I get a `ping` message, this listener always throws a NPE

mcollovati (changes requested):

mcollovati:
I did a try with the plain java flow starter and got this error when loading the web page

```
Caused by: java.lang.RuntimeException: Cannot load platform configurator
	at jakarta.websocket.server.ServerEndpointConfig$Configurator.fetchContainerDefaultConfigurator(ServerEndpointConfig.java:96)
	at jakarta.websocket.server.DefaultServerEndpointConfig.<init>(DefaultServerEndpointConfig.java:57)
	at jakarta.websocket.server.ServerEndpointConfig$Builder.build(ServerEndpointConfig.java:262)
	at com.vaadin.base.devserver.viteproxy.ViteWebsocketEndpointInitializer.<init>(ViteWebsocketEndpointInitializer.java:64)
	at com.vaadin.base.devserver.ViteHandler.doStartDevModeServer(ViteHandler.java:179)
	at com.vaadin.base.devserver.AbstractDevServerRunner.runOnFutureComplete(AbstractDevServerRunner.java:171)
	at com.vaadin.base.devserver.AbstractDevServerRunner.lambda$new$0(AbstractDevServerRunner.java:162)
```

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketConnection.java
Artur-: Shouldn't we use the same header in all files?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketEndpoint.java
Artur-: Let's do that and revert if it turns out that it logs a lot of intermittent failures during restarts or similar

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketEndpointInitializer.java
Artur-: It is from an earlier version and no longer needed. Removed.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketProxy.java
Artur-: Uh, looks like it was deleted during some refactoring..

Artur-:
> I did a try with the plain java flow starter and got this error when loading the web page
>
> Caused by: java.lang.RuntimeException: Cannot load platform configurator

It seems like `jetty:run` has a problem with finding its own server configurator (https://github.com/eclipse/jetty.project/issues/9180). Added a custom one as a workaround

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketConnection.java
mcollovati: You're right. I thought the headers were already up to date, but they aren't up to date yet

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketConnection.java
```diff
@@ -0,0 +1,107 @@
...
+
+import com.vaadin.base.devserver.ViteHandler;
+
+/**
+ * Communicates with a Vite server through a websocket connection.
+ */
+public class ViteWebsocketConnection implements Listener {
+
+    private final Consumer<String> onMessage;
+    private WebSocket clientWebSocket;
```
mcollovati, line 38: nit: can be final

mcollovati (changes requested):
I'm testing the latest changes on plain Java starter, and it seems that after a while from the first connection (~30 secs) the websocket channels are closed.

For example, if after 30 secs I try to type something wrong in a typescript source I get this error

```
[HttpClient-2-Worker-3] DEBUG com.vaadin.base.devserver.viteproxy.ViteWebsocketProxy - Error sending message to browser
java.io.IOException: java.nio.channels.ClosedChannelException
	at org.eclipse.jetty.util.FutureCallback.block(FutureCallback.java:163)
	at org.eclipse.jetty.websocket.jakarta.common.JakartaWebSocketBasicRemote.sendText(JakartaWebSocketBasicRemote.java:121)
	at com.vaadin.base.devserver.viteproxy.ViteWebsocketProxy.lambda$new$0(ViteWebsocketProxy.java:56)
	at com.vaadin.base.devserver.viteproxy.ViteWebsocketConnection.onText(ViteWebsocketConnection.java:82)
	at java.net.http/jdk.internal.net.http.websocket.WebSocketImpl$ReceiveTask.processText(WebSocketImpl.java:635)
	at java.net.http/jdk.internal.net.http.websocket.WebSocketImpl$ReceiveTask.run(WebSocketImpl.java:443)
	at java.net.http/jdk.internal.net.http.common.SequentialScheduler$CompleteRestartableTask.run(SequentialScheduler.java:149)
	at java.net.http/jdk.internal.net.http.common.SequentialScheduler$SchedulableTask.run(SequentialScheduler.java:230)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.nio.channels.ClosedChannelException
	at org.eclipse.jetty.websocket.core.internal.WebSocketSessionState.onOutgoingFrame(WebSocketSessionState.java:184)
	at org.eclipse.jetty.websocket.core.internal.WebSocketCoreSession.sendFrame(WebSocketCoreSession.java:514)
	at org.eclipse.jetty.websocket.jakarta.common.JakartaWebSocketRemoteEndpoint.sendFrame(JakartaWebSocketRemoteEndpoint.java:141)
	at org.eclipse.jetty.websocket.jakarta.common.JakartaWebSocketBasicRemote.sendText(JakartaWebSocketBasicRemote.java:120)
	... 9 more
````

**Comment by mcollovati** [comment]
Works fine with Spring.

However, an error is logged when reloading the page

```
2023-01-17T11:26:34.198+01:00 DEBUG 65602 --- [ient-9-Worker-2] c.v.b.d.viteproxy.ViteWebsocketProxy     : Error sending message to browser

java.lang.IllegalStateException: The WebSocket session [f] has been closed and no method (apart from close()) may be called on a closed session
	at org.apache.tomcat.websocket.WsSession.checkState(WsSession.java:989) ~[tomcat-embed-websocket-10.1.1.jar:10.1.1]
	at org.apache.tomcat.websocket.WsSession.getBasicRemote(WsSession.java:546) ~[tomcat-embed-websocket-10.1.1.jar:10.1.1]
	at com.vaadin.base.devserver.viteproxy.ViteWebsocketProxy.lambda$new$0(ViteWebsocketProxy.java:56) ~[vaadin-dev-server-24.0-SNAPSHOT.jar:na]
	at com.vaadin.base.devserver.viteproxy.ViteWebsocketConnection.onText(ViteWebsocketConnection.java:82) ~[vaadin-dev-server-24.0-SNAPSHOT.jar:na]
	at java.net.http/jdk.internal.net.http.websocket.WebSocketImpl$ReceiveTask.processText(WebSocketImpl.java:635) ~[java.net.http:na]
	at java.net.http/jdk.internal.net.http.websocket.WebSocketImpl$ReceiveTask.run(WebSocketImpl.java:443) ~[java.net.http:na]
	at java.net.http/jdk.internal.net.http.common.SequentialScheduler$CompleteRestartableTask.run(SequentialScheduler.java:149) ~[java.net.http:na]
	at java.net.http/jdk.internal.net.http.common.SequentialScheduler$SchedulableTask.run(SequentialScheduler.java:230) ~[java.net.http:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:833) ~[na:na]
```

**Comment by Artur-** [comment]
The default idle timeout seems to be 30s and Vite was pinging with a 30s interval so likely it had time to close before the ping arrived. I increased it to 60s but it seems that reopening of connections still need to be handled somehow as the same situation will probably occur if you close your laptop for more than a minute now

**Comment by Artur-** [comment]
Ah there is the separate vite checker plugin connection which does not send any pings..

**Comment by mcollovati** [comment]
> Ah there is the separate vite checker plugin connection which does not send any pings..

Yes, can confirm that the websockets connection opend by the vite checker plugin gets still closed after the max idle timeout

**Comment by Artur-** [comment]
Refactored so endpoint initialization takes place in the servlet context listener and not in a separate thread, which was what was causing the Jetty issues. Now also closes the other connection if one of them is closed

File: vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketConnection.java [hunk 11]
```diff
@@ -0,0 +1,134 @@
...
+import jakarta.websocket.CloseReason.CloseCodes;
+
+/**
+ * Communicates with a Vite server through a websocket connection.
+ */
+public class ViteWebsocketConnection implements Listener {
+
+    private final Consumer<String> onMessage;
+    private WebSocket clientWebSocket;
+    private Runnable onClose;
```
mcollovati, line 41: nit: can be final

Artur-: You are starting to sound like a static code analyzer

mcollovati:
I have not found the cause, but it seems Vite ws proxying does not work when setting a `context-path` (e.g. `server.servlet.context-path=/ui`)
I have tested this only on spring-based project right now

Artur-:
Yes the context path should not be included in the endpoint mapping but it must be included in the Vite base path, as Vite has no idea about any context path. Fixed and now it should work both with context and/or servlet path mappings

mcollovati:
Tested again with context path and url mapping and it works fine.
I also did a test with an HTTPS reverse proxy in front of the application, and it works, except for the `vaadin-dev-tools` connection that connects directly to Vite port.

@Artur- Do you think we can proxy also this connection through Vaadin Flow?

Artur-:
Which vaadin-dev-tools connection would that be? Right now there should be:
1. vaadin-dev-tools connects to `ws://localhost:8080/VAADIN/push?v-r=push&debug_window`
2. If you use push, it connects to `ws://localhost:8080/VAADIN/push?v-r=push`
3. Vite connects to `ws://localhost:8080/VAADIN` with this PR
4. The checker plugin in Vite connects to the same URL
5. Live reload connects to ws://localhost:35729/

mcollovati:
My mistake, the failing connection is not to Vite, but to Spring Boot livereload port

Artur-:
For live reload there is #13646 which would use the dev tools ws instead of a separate connection and additionally work if you have multiple apps running

mcollovati (approved):

mcollovati:
Just for the record, the checker plugin disconnects after 60 seconds, but it seems not to be a problem because if there are changes to typescript files, the page is reloaded because of the other Vite WS connection and all connections are established again

Artur-:
I think the checker plugin related issues will be solved once https://github.com/fi3ework/vite-plugin-checker/issues/200 is implemented

PR #15700, opened 2023-01-17
fix: Do not perform type checking for TS files from jar files

### LINKED ISSUE #15485: Stricter TS checks in Vaadin 23.3.0

Artur-:
The TS files that come from JAR files are not under control of the project developer so enforcing project TS rules on them does not make sense. This is similar to "skipLibCheck" used for TS files that are in node_modules.

Fixes #15485


Artur-:
Seems like this does not actually fix the issue. Files which are included in other files are checked even with the exclude section

mshabarov (approved):

PR #15709, opened 2023-01-17
feat: added a commonly needed helper method to get current route target

### LINKED ISSUE #4565: Add API for getting the current view
Currently, in order to get the active view path, you will need to go through `UI.getCurrent().getInternals().getActiveViewLocation()`. I don't feel comfortable relying on the UIInternals `which are intended for internal use`, so there should be a public method in UI to get this information instead. This way, if the internal implementation changes, the UI class just needs to keep up and the change will be transparent to the consumers.

mstahv:
## Description

In layouts and menus it is a common need to detect the active view/route/navigationtarget/whatever-we-call-it. Currently we offer no good ways how to get that. The basically undocumented and cumbersome getActiveChain() method in AfterNavigationEvent is the closest available method.

Fixes #4565

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


mstahv:
Note for @vaadin/flow team: if you are ok with it, I (and @knoobie) would like to drop the Optional usage still 🤓

This method is not randomly returning null, only in certain known states (when not attached to the dom and when not using routing) and developer should be able to know those cases.

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1195,6 +1195,30 @@ public boolean isNavigationSupported() {
...
+     *
+     * @return the currently active route instance if available
+     */
+    public Optional<Component> getCurrentView() {
+        try {
+            return Optional.ofNullable((Component) getInternals()
+                    .getActiveRouterTargetsChain().get(0));
+        } catch (Exception e) {
+            // Current route is not always available
+            return Optional.empty();
```
knoobie, line 1218: We could even throw a meaningful exception if the route is null / log a meaningful message for the developer why the route could be null - with the normal culprits: used before attached and other variants we've seen by developer, making it easier for them to understand the problem and fix it

MarcinVaadin:
All other similar helper methods are within `UIInternals`, IMO same should be in that case. As for the idea itself - LGTM.

MarcinVaadin (changes requested):
All other similar helper methods are within UIInternals, IMO same should be in that case. As for the idea itself - LGTM.

mstahv:
That UIInternals is not the place to add API for framework users. Both the naming and API docs say that quite clearly:

_Holds UI-specific methods and data which are intended for internal use by the framework.
For internal use only. May be renamed or removed in a future release._

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1195,6 +1195,30 @@ public boolean isNavigationSupported() {
...
+     * layouts. Thus, consider postponing the usage of this method to for
+     * example AfterNavigationEvent.
+     *
+     * @return the currently active route instance if available
+     */
+    public Optional<Component> getCurrentView() {
+        try {
+            return Optional.ofNullable((Component) getInternals()
+                    .getActiveRouterTargetsChain().get(0));
+        } catch (Exception e) {
```
mcollovati, line 1216: What exceptions may be raised by this code?
If it is only to catch `IndexOutOfBoundsException`, I would prefer a check on `List.isEmpty()`.

mcollovati (commented):
Since `getCurrentView()` is a public API, it would be good to have dedicated tests for it

flow-server/src/main/java/com/vaadin/flow/component/UI.java
mstahv: NPE (router can be null for example with embedded web components, didn't check, but read it from some javadocs).

mstahv: Actually, it doesn't seem to go to Router at all, but only ArrayList<HasElement> routerTargetChain that is always initialized. So there shouldn't be NPEs.

mstahv: So this should do it as well, but somebody could comment on the communitys (or at least mine and knoobies) wish about returning component directly, instead of Optional.

       if(getInternals().getActiveRouterTargetsChain().isEmpty()) {
            return Optional.empty();
        }
        return Optional.of((Component) getInternals()
                    .getActiveRouterTargetsChain().get(0));

mcollovati:
> Note for @vaadin/flow team: if you are ok with it, I (and @knoobie) would like to drop the Optional usage still nerd_face
> 
> This method is not randomly returning null, only in certain known states (when not attached to the dom and when not using routing) and developer should be able to know those cases.

I'm fine with dropping the Optional, but I would throw an `IllegalStateException` if the routerTargetChain is empty, to make it clear that the method is invoked from a wrong path

mstahv:
Pushed a version without Optional 👍

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1195,6 +1195,28 @@ public boolean isNavigationSupported() {
         return true;
     }
 
+    /**
+     * Returns the currently active route aka navigation target shown in this
+     * UI.
+     * <p>
+     * Note, that certain UIs, like embedded apps, don't support routing and for
+     * those an empty optional is returned.
```
mcollovati, line 1203: Javadoc needs to be updated, as now we throw an exception instead of return an Optional

mcollovati (changes requested):

mcollovati (approved):

PR #15713, opened 2023-01-18
fix: DevBundle used after clean-frontend goal

### LINKED ISSUE #15707: Running the clean-frontend maven target package.json hash is cleaned making devBundle build required

caalador:
Dev bundle should be used if possible
even after `clean-frontend` goal has been
executed.

Fixes #15707

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
@@ -316,8 +316,16 @@ public class TaskRunDevBundleBuild implements FallibleCommand {
 
         if (packageJsonFile.exists()) {
             try {
-                return Json.parse(FileUtils.readFileToString(packageJsonFile,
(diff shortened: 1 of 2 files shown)


mshabarov:
Tested with a custom-theme-demo project and got `No hash found for 'package.json'` after `mvn vaadin:clean-frontend` which led to bundle re-build. Will check with the fresh starter project later on.

caalador:
Did you also build the maven-plugin and plugin-base for the branch with flow-server? Else the plugin will run against the flow-server.

mshabarov:
Yes, I've built the whole Flow project. I also tried to use `mvn flow:clean-frontend` to be sure that the flow codes are executed, but the result is the same. Let me debug the app a bit.

mshabarov:
The problem was in the project's configuration. I had a `prepare-frontend` goal in my pom.xml, so it seems it impacted the build so that the generated package.json was removed and the Flow said that there is no hash. I removed this goal and everything works now.

mshabarov (approved):

PR #15719, opened 2023-01-19
fix: Use a random port for Spring Boot livereload if the default is taken

### LINKED ISSUE #7970: Live-reload does not recognize Spring Boot DevTools using non-default live-reload server port or not running at all
The live-reload client-side currently always attempts to connect to port 35729 to poll for a potential Spring Boot live-reload server, and if failing, attempts to connect to the dedicated push channel. This approach has two problems:

1. ~For non-Spring applications and Spring not running devtools, the browser console contains an error like the following: `WebSocket connection to 'ws://localhost:35729/' failed: Error in connection establishment: net::ERR_CONNECTION_REFUSED`. The message cannot be supressed and is confusing since it looks like something is wrong.~

2. If a Spring project has changed the port via configuration `spring.devtools.livereload.port` in `application.properties`, it will fall back to the dedicated push channel while the developer probably want to use the custom port. 

3. If Spring Boot Devtools are not running (e.g. `spring-boot-maven-plugin` is configured with `<fork>false</fork>`, this is not detected and the error in 1. is shown in the broswer console. Should detect this state on the server side and initialize the gizmo accordingly (live reload unavailable).
 
Acceptance criteria:
* No error in browser console during normal startup of both Spring and non-Spring applications.
* We identify on the server side that Spring Boot devtools is runing and communicate the correct port to the client side.

Artur-:
Fixes #7970

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/NetworkUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/NetworkUtil.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.internal;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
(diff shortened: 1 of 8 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/NetworkUtil.java
```diff
@@ -0,0 +1,62 @@
...
+    private NetworkUtil() {
+        // Utils only
+    }
+
+    /**
+     * Returns an available tcp port in the system.
+     *
+     * @return a port number which is not busy
+     */
+    public static int getFreePort() {
```
mcollovati, line 37: In `PortProber` class used by integration tests, we check if the port is free on all interfaces.
Does it make sense to do the same here, or even move the `PortProber` class from `flow-test-utils` module in `flow-server`?

Artur-: I thought about that but PortProber is broken in some cases where you have only ipv4 or only ipv6. It will then report every port as busy. That would need to be fixed first

mcollovati (approved):

PR #15721, opened 2023-01-19
feat: add usage tracking for when expressMode

caalador:
Add tracking of express mode usage when
dev-bundle  is checked.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -27,6 +27,7 @@ import org.slf4j.LoggerFactory;
 
 import com.vaadin.experimental.FeatureFlags;
 import com.vaadin.flow.di.Lookup;
+import com.vaadin.flow.internal.UsageStatistics;
 import com.vaadin.flow.server.ExecutionFailedException;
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -102,6 +103,7 @@ public NodeTasks(Options options) {
             // determine if we need a rebuild as the check happens immediately
             // and no update tasks are executed before it.
             if (!options.productionMode && options.isDevBundleBuild()) {
+                UsageStatistics.markAsUsed("flow/expressBuild", null);
```
Artur-, line 106: Is this a better place than in `DevModeInitializer` 
```
        if (!config.enableDevServer()) {
            nodeTasksFuture.join();
            return null;
        } else {
            return new ViteHandler(devServerLookup, port,
                    options.getNpmFolder(), nodeTasksFuture);
        }
```
?

caalador: This will mark it any time we check for expressBuild e.g. have it in use. Also you don't need devServer package if using the express build.

Artur- (approved):

PR #15735, opened 2023-01-20
feat: Allow smooth scrolling a component into view

Artur-:


jouni:
I assume you left room in the API for the other options, `block` and `inline`, as well: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView, if someone would ever want those added?

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -1546,10 +1546,26 @@ private void informChildrenOfStateChange(boolean enabled, Element element) {
...
         // for an unknown reason, needs to be called deferred to work on a newly
         // created element
-        executeJs(
-                "var el = this; setTimeout(function() {el.scrollIntoView();}, 0);");
+        String options = "";
+        if (smooth) {
+            options = "{ behavior: 'smooth' }";
+        }
+        executeJs("var el = this; setTimeout(function() {el.scrollIntoView("
+                + options + ");}, 0);");
```
knoobie, line 1568: Doesn't this break on Safari? https://caniuse.com/mdn-api_element_scrollintoview_options_parameter

Artur-: By ”break” do you mean ”does not scroll smoothly”?

knoobie: At least in the past it did nothing and just logged an error in the console (I created a similar utility in the past).. does it scroll now?

jouni: Works nicely in Safari 16.3 (beta) at least, smooth scrolling even. caniuse.com tells me that it’s been there since Safari 16.0.

Artur-:
Changed the API so you can give all the options available

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
jouni: I suppose we need to test how it works in Safari 15, and make sure to fallback to the older browser API in case it just throws an error.

Artur-: Safari 15.1 scrolls also when giving an option

mcollovati (approved):

PR #15740, opened 2023-01-22
fix: Do not fail component picking if Flow is undefined

Artur-:


MarcinVaadin (commented):
How it is possible to not have Flow defined while using Vaadin app with vaadin dev server enabled? Hilla?

Artur-:
Using Hilla

MarcinVaadin (approved):

PR #15742, opened 2023-01-24
feat: Allow every component to have styles

Artur-:
All components are capable of having class names yet you must remember to implement `HasStyle` every time you create a new component or the class API is missing. The only known exception is a text node which cannot have classes. It is fine if this edge case, which is almost never used, throws an exception from the API

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -52,7 +52,7 @@ import com.vaadin.flow.shared.Registration;
  * @since 1.0
  */
 public abstract class Component
-        implements HasElement, AttachNotifier, DetachNotifier {
+        implements HasStyle, AttachNotifier, DetachNotifier {
 
     /**
      * Encapsulates data required for mapping a new component instance to an


Artur-:
Updated the description

mstahv:
Was there a reason why the inline styles where ditched into a separate interface (and not implemented by Component)?

Artur-:
The only I can remember was the academic "A text node cannot have styles"

mstahv:
But it cannot have class names either 🤔 I definitely want this in, but just suggesting we should get bit further.

mstahv:
But it cannot have class names either 🤔 I definitely want this in, but just suggesting we should get bit further.

Artur-:
Ah I think I misread your question. You were suggesting we do this that I now updated the PR to instead?

mstahv:
👍 That's exactly what I meant.

mstahv:
And then override Text to throw UnsupportedOperationException 🤔

Artur-:
There will be `UnsupportedOperationException` thrown already although they originate in `AbstractTextElementStateProvider`

mstahv:
Then this looks very good to me.

mstahv (approved):
Excellent!

mstahv:
Some actual team member might wanna rubberstamp as well ;-)

mshabarov:
Looks good to me either. What would be the plan for Vaadin Flow components? Should we remove `HasStyles` from our components to make codes cleaner? We can keep though.

mshabarov (approved):

PR #15752, opened 2023-01-25
fix: Do not bring junit4 into all TestBench projects

Artur-:
Use the same dependency as flow-components

### DIFF
--- a/flow-html-components-testbench/pom.xml
+++ b/flow-html-components-testbench/pom.xml
@@ -17,8 +17,9 @@
     <dependencies>
         <dependency>
             <groupId>com.vaadin</groupId>
-            <artifactId>vaadin-testbench-core</artifactId>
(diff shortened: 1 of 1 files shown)


flow-html-components-testbench/pom.xml
```diff
@@ -17,8 +17,9 @@
     <dependencies>
         <dependency>
             <groupId>com.vaadin</groupId>
-            <artifactId>vaadin-testbench-core</artifactId>
+            <artifactId>vaadin-testbench-shared</artifactId>
```
MarcinVaadin, line 20: So no junit at all? Where to decide which junit version use?

Artur-: The user decides. We cannot decide in element classes which junit the user will use

MarcinVaadin (approved):

PR #15757, opened 2023-01-25
fix: Always set image alt text the same way

### LINKED ISSUE #15756: Image.getAlt() not working if alt text set by constructor
### Description of the bug

Create a new Image with alt text:
`Image image = new Image( "http://www.something.com" , "Some Alt Text" )`

Then test getAlt():
`assertEquals( "Some Alt Text" , image.getAlt().get() )`
throws a java.util.NoSuchElementException: No value present

Related to the change for #14880

If setAlt() is used, then getAlt() works correctly.
I would expect setting the Alt text to be the same whether the constructor is used or the setAlt() is used.

### Expected behavior

getAlt() should return the alt text, regardless of whether the constructor or the setAlt() was used.

See description of the bug

### Versions

- Vaadin / Flow version: 23.3
- Java version: 11
- OS version: Windows 10
- IDE (if applicable):

Artur-:
Fixes #15756

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Image.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Image.java
@@ -36,11 +36,9 @@ import com.vaadin.flow.server.StreamResource;
 public class Image extends HtmlContainer
         implements ClickNotifier<Image>, HasAriaLabel {
 
+    private static final String ALT_ATTRIBUTE = "alt";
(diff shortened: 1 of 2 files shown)


caalador (dismissed):

caalador:
Some image tests need updated expectations
```
Error:  setAndResetValues(com.vaadin.flow.component.html.ImageTest)  Time elapsed: 0.008 s  <<< FAILURE!
java.lang.AssertionError: Default value for alt should be an empty Optional expected:<Optional.empty> but was:<Optional[]>
```

PR #15760, opened 2023-01-27
chore: Update copyright year

caalador:
Update copyright year to 2023


MarcinVaadin:
Wasn't so that we should update the year in files that have been added or modified, not all of them?

caalador:
I doubt anyone will update a modified file.
Also the header templates have the wrong content so any new files would get in many cases the wrong year.

MarcinVaadin (approved):

Issue #15625, opened 2023-01-05
Flow 24 alpha does not work on payara-micro due to ASM incompatibility


MarcinVaadin:
### Description of the bug

From [StackOverflow](https://stackoverflow.com/questions/75013503/vaadin-flow-24-0-0-alpha7-doesnt-work-with-payara-micro):

> I tried to run Vaadin Flow 24.0.0.alpha7 (because it supports Jakarta EE 9) on Payara Micro but Vaadin servlet didn't deploy.
> 
> I believe Payara Micro 6.2022.2 should support it theoretically however when I tried to run it on Java 17 I got multiple errors of this sort:
> 
> Exception while visiting com/vaadin/flow/data/provider/SortDirection.class of size 1425
> java.lang.UnsupportedOperationException: PermittedSubclasses requires ASM9
> When I switched to Java 19 I got the same ASM9 errors plus a bunch of new ones like this:
> 
> Exception while visiting WEB-INF/classes/one/esng/ang/ui/views/MainLayout.class of size 3967
> java.lang.IllegalArgumentException: Unsupported class file major version 63
> Is there any way to make it work?

### Expected behavior

Should work

### Minimal reproducible example

we should add payara-micro example to one of skeleton starter modules

### Versions

- Vaadin / Flow version: 24 alpha8
- Java version: 17
- Application Server (if applicable): payara-micro 6.xx

Artur-:
Payara issue: https://github.com/payara/Payara/issues/6127

knoobie:
How about changing the enum to something like, to remove the need of the kinda fancy stuff ;) 

```java
public enum SortDirection {

  /**
   * Ascending (e.g. A-Z, 1..9) sort order
   */
  ASCENDING(Constants.ASCENDING_OPPOSITE),

  /**
   * Descending (e.g. Z-A, 9..1) sort order
   */
  DESCENDING(Constants.DESCENDING_OPPOSITE);

  private final SortDirection opposite;

  SortDirection(SortDirection opposite) {
    this.opposite = opposite;
  }

  /**
   * Get the sort direction that is the direct opposite to this one.
   *
   * @return a sort direction value
   */
  public SortDirection getOpposite() {
    return opposite;
  }

  private static class Constants {

    public static final SortDirection ASCENDING_OPPOSITE = DESCENDING;
    public static final SortDirection DESCENDING_OPPOSITE = ASCENDING;
  }
}

```

MarcinVaadin:
[...]

PR #15770, opened 2023-01-30
fix: Make Flow compatible with ASM versions older than 9

mshabarov:
Solution for https://github.com/vaadin/flow/issues/15625, until https://github.com/payara/Payara/issues/6127 is fixed.

Thanks @knoobie for inspiration!

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
@@ -53,10 +53,10 @@ public class ComponentTracker {
      * Represents a location in the source code.
      */
     public static class Location implements Serializable {
-        private String className;
-        private String filename;
-        private String methodName;
-        private int lineNumber;
+        private final String className;
+        private final String filename;
+        private final String methodName;
+        private final int lineNumber;
 
         public Location(String className, String filename, String methodName,
                 int lineNumber) {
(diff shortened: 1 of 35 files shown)


MarcinVaadin (dismissed):

MarcinVaadin:
Why not to simplify `SortDirection` to:

```
public enum SortDirection {

    /**
     * Ascending (e.g. A-Z, 1..9) sort order
     */
    ASCENDING,

    /**
     * Descending (e.g. Z-A, 9..1) sort order
     */
    DESCENDING;

    /**
     * Get the sort direction that is the direct opposite to this one.
     *
     * @return a sort direction value
     */
    public SortDirection getOpposite() {
        return ASCENDING.equals(this) ? DESCENDING : ASCENDING;
    }
}
```
?

flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentTracker.java
```diff
@@ -51,8 +52,79 @@ public class ComponentTracker {
     /**
      * Represents a location in the source code.
      */
-    public record Location(String className, String filename, String methodName, int lineNumber)
-            implements Serializable {
+    public static class Location implements Serializable {
```
mcollovati, line 55: As we are replacing a record, we can declare fields as final and remove setters

mcollovati: Also, making accessors without `get`  (e.g. `className` instead of `getClassName`) may simplify a future migration back to record

mcollovati (changes requested):

Artur-:
The summary makes no sense. Is about making it compatible with an old ASM version?

mshabarov:
> The summary makes no sense. Is about making it compatible with an old ASM version?

Yes, for older than 9. Changed the summary.

mcollovati (approved):

MarcinVaadin (approved):

Artur-:
Without a test, I give this one week before it is broken

mshabarov:
> Without a test, I give this one week before it is broken

We have a tests for CDI add-on vs Payara server validation (which will be enabled soon I hope), in my opinion only these can reliably check that Flow is fully compatible with the ASM. We could have a tests in Flow checking that there is no record classes, but IMO Flow can break by other reasons, like we seen with abstract classes in enum.

PR #15785, opened 2023-01-31
fix: use .ts instead of .js

caalador:
vaadin-dev-tools is now a ts file
and not a js file.
reflect this in the generated bootstrap.


Artur-:
At least if you use "node16" module resolution, this is wrong. You are supposed to import `.js` for `.ts` files

caalador:
I'll trust you on this one, even though it is a bit confusing with files.
Would it work without the .js/.ts ending?

Artur-:
At least if you use "node16" module resolution, it will not work without the suffix. You are supposed to import `foo.js` for `foo.ts` files. Not `foo.ts` nor `foo`. 

I am not sure we actually want to move to using `node16` resolution, but if the current version works with all module resolutions in node and nothing is broken, maybe we should not change it?

PR #15804, opened 2023-02-02
fix: Use correct SecurityContextRepository for stateless authentication

Artur-:
The shared objects need to be available when other configurers are run and because of https://github.com/spring-projects/spring-security/issues/12579 a workaround is needed to actually apply the correct SecurityContextRepository

Fixes https://github.com/vaadin/hilla/issues/681


cromoteca (approved):

platosha (approved):

RainerGanss:
[This ](https://github.com/spring-projects/spring-security/issues/12579) was merged into 6.0.2. Does this mean the workaround in VaadinWebSecurity can be removed from the 24er branch?

PR #15807, opened 2023-02-02
chore: Upgrade Vite to 4.1.1

Artur-:


caalador (dismissed):

caalador:
it-test 8 has a too long name failure. Feels like this should have been handled by another handler. Is static file before push handler?
```
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker - [vite] Internal server error: ENAMETOOLONG: name too long, stat '/home/runner/work/flow/flow/flow-tests/test-router-custom-context/frontend/push?v-r=push&v-uiId=0&v-pushId=a0f70b2d-1042-416e-bf43-25250d45f37a&X-Atmosphere-tracking-id=0&X-Atmosphere-Framework=3.1.2-javascript&X-Atmosphere-Transport=long-polling&X-Atmosphere-TrackMessageSize=true&Content-Type=application%2Fjson%3B charset%3DUTF-8&X-atmo-protocol=true&_=1675338554215'
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -       at Object.statSync (node:fs:1596:3)
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -       at tryResolveFile
```

Artur-:
Vite is consulted first as we do not know what it can serve

caalador:
But that was clearly a push request `...frontend/push?v-r=push&...` which we could skip and just hand all the way up to PushRequestHandler.

caalador:
Oh. It reverses the list later so it's bootstrap then push then rest... so push handler should be there to catch this first as long as atmosphere was available on Service init.

Artur-:
Vite is consulted through the static file server, before all request handlers. Still this feels like a bug in Vite where it throws an exception instead of returning 404 for a file it cannot find

Artur-:
Let's see what the outcome of https://github.com/vitejs/vite/issues/11904 is

caalador (approved):

PR #15813, opened 2023-02-02
fix: Use feature detection to show the unsupported browser page

### LINKED ISSUE #14366: Use feature detection instead of browser version detection for the "too old" page
### Describe your motivation

Right now we do browser version detection for the "too old" page which means that there is a hard coded number for instance for Edge 79 here https://github.com/vaadin/flow/blob/7345836c3a891af35599bae020d77af6054e490b/flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java#L642. 

However in the release notes we state that we support the evergreen aka latest version of Edge so there as a real possibility that we will include code that works in the evergreen Edge but not in Edge 79. This is fine but then the app will be broken in Edge 79 and probably show a blank screen instead of showing that "This app probably won't work because the browser version is outdated". That does not help anybody.

The "too old" page should be based on feature detection so that it is shown when we detect that the Vaadin version in use uses such features that are not available in the browser. This makes it quite likely that the "too old" page is correct in most cases and does not disturb users with browsers that would actually work nor show blank pages to people who really have too old browers.

### Additional context

We do not have to list all features we are using for this to work. It is most likely enough if we start by listing the latest features such as `ResizeObserver`. This should be the newest feature and browsers supporting resize observers should also support other features we need. 

The list does not have to be perfect either as it it merely there to help people and you can always get around it with the "use the app anyway" link if the detection is wrong

Artur-:
Fixes #14366


flow-server/src/main/java/com/vaadin/flow/server/UnsupportedBrowserHandler.java
```diff
@@ -17,124 +17,38 @@
...
-                                : "")
-                        + " <a href=\"http://www.opera.com/browser\">Opera</a>"
-                        + " and <a href=\"http://www.apple.com/safari\">Safari</a>.</b><br/>"
-                        + "Upgrading to the latest version of one of these <b>will make the web safer, faster and better looking.</b></p>"
-                        + "<p><sub><a onclick=\"document.cookie='"
-                        + FORCE_LOAD_COOKIE
-                        + "';window.location.reload();return false;\" href=\"#\">Continue without updating</a> (not recommended)</sub></p>"
-                        + "</body>\n" + "</html>");
-        // @formatter:on
+        IOUtils.copy(getClass().getResourceAsStream("too-old.html"), page,
```
knoobie, line 46: Would it make sense to allow people to customize this?

Artur-: You mean in a less fragile way than adding your own `src/main/resources/com/vaadin/flow/server/too-old.html` ?

knoobie: If you say it like this: yes 😅

Artur-: Ok now you can use `src/main/resources/[browser-too-old.html` instead :)

knoobie: Thanks 🙇‍♂️ This is great! Having a way to customize this page is really good. We had requests in the past especially for translation and corporate styling needs. Could be something worth to mention as new feature for v24 - some people are really looking forward to it. Customizing it atm is just hacky

Artur-:
Note that if this is backported (it probably should be) then the feature set to detect is different (ResizeObserver should be ok)

mcollovati (approved):
LGTM.
I just wonder if we need to mark this as a "breaking change" since we removed a public API.
Perhaps nobody is using the `WebBrowser.isTooOldToFunctionProperly()` method, but it would be good to mention the removal in the release notes

Artur-:
Yes, we could mention that even though usage is likely below 0.1%
It is almost like we could deprecate the whole `WebBrowser` class and say that you should not do browser detection..

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -155,6 +157,14 @@ public boolean synchronizedHandleRequest(VaadinSession session,
         return true;
     }
 
+    private void redirectToOldBrowserPageWhenNeeded(Document indexDocument) {
+        addScript(indexDocument, """
+                if (!('CSSLayerBlockRule' in window)) {
+                    window.location.search='v-r=oldbrowser';
+                }
+                """);
```
mshabarov, line 165: So is it a place where we put the checks for used features?

Artur-: This is the place yes

PR #15824, opened 2023-02-04
fix: Serve text/css/xml/javascript resources correctly outside the context of a page

### LINKED ISSUE #15817: Javascript, HTML and CSS should be sent as utf-8
### Description of the bug

If I create a new application and create `a.js` inside `src/main/resouces/META-INF/resources` with the content
```
console.log("füü");
```

and then open http://localhost:8080/a.js in the browser I see

```
console.log("fÃ¼Ã¼");
```

### Expected behavior

I expect UTF-8 to be used by default

As described above

### Versions

- Vaadin / Flow version: 24.0 alpha 9

Artur-:
When the resource is served without a charset, it seems the encoding of the page is used to interpret the resource correctly as utf-8. If you load a file outside the context of a page however, the content is wrong when the charset is not specified.

Fixes #15817


knoobie:
Sounds like another file type to include here :) https://github.com/vaadin/flow/issues/15834

mcollovati (approved):

mcollovati:
> Sounds like another file type to include here :) #15834

This has been fixed by #15884, since the response is written directly by request handlers defined in `PwaHandler`

PR #15865, opened 2023-02-08
test: Use --headless=new as ChromeOptions recommends for Chrome 109+

Artur-:


flow-tests/vaadin-spring-tests/pom.xml
```diff
@@ -35,6 +35,18 @@
...
+                <artifactId>vaadin-testbench-bom</artifactId>
+                <version>${testbench.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.seleniumhq.selenium</groupId>
+                <artifactId>selenium-chromium-driver</artifactId>
+                <version>4.8.0</version>
+            </dependency>
```
mshabarov, line 49: Let's see if the validation passes. If so, we should release TB new version with the selenium-chromium-driver included and remove this.

Artur-: It does not but we should still have the new TB

flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeBrowserTest.java
```diff
@@ -154,7 +154,7 @@ protected List<DesiredCapabilities> customizeCapabilities(
 
     static ChromeOptions createHeadlessChromeOptions() {
         final ChromeOptions options = new ChromeOptions();
-        options.addArguments("--headless", "--disable-gpu");
+        options.addArguments("--headless=new", "--disable-gpu");
```
knoobie, line 157: FYI: I played around with this new flag two weeks ago as well. Total nightmare, couldn't get it to work and postponed it for now.. just saw that chrome just published a blog post about this flag, mentioned 112 instead of 109.. so well.. https://developer.chrome.com/articles/new-headless/

Artur-:
This is the problem here #16371. The new headless mode works fine and reports a real error that was not reported by the old headless version

tepi (approved):

PR #15870, opened 2023-02-08
fix: Custom theme for embedded components in express build mode

mshabarov:
Fixes https://github.com/vaadin/flow/issues/15873


flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1595,6 +1597,72 @@ protected static void addJavaScriptEntryPoints(
...
+     * @param fileName
+     *            the stylesheet file name to add a reference to
+     * @return the collection of link tag elements to be added to the page
+     * @throws IOException
+     *             if theme name cannot be extracted from file
+     */
+    protected static Collection<Element> getTagForTheme(
+            DeploymentConfiguration config, String fileName)
+            throws IOException {
+        Collection<Element> tags = new ArrayList<>();
```
caalador, line 1616: I would guess calling this List here would perhaps not confuse sonatype to think we return mutable and immutable collections?

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1595,6 +1597,72 @@ protected static void addJavaScriptEntryPoints(
...
+            for (File themeFolder : Objects.requireNonNull(
+                    packagedThemesFolder.listFiles(File::isDirectory),
+                    "Expected at least one theme in the front-end generated themes folder")) {
+                String packagedThemeName = themeFolder.getName();
+                packagedThemeNames.add(packagedThemeName);
+                Element tag = getLinkTagWithStyleRef(packagedThemeName,
+                        fileName);
+                tags.add(tag);
+            }
+        }
```
caalador, line 1646: The packagedThemeNames could just be a `boolean customThemeFound = false;` with the check in the for loop `customThemeFound = customThemeFound && packagedThemeName.equals(themeName.get));` as that is the only thing the list is used for.

flow-server/src/test/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandlerTest.java
```diff
@@ -56,6 +59,7 @@
 import com.vaadin.flow.internal.UsageStatistics;
 import com.vaadin.flow.server.AppShellRegistry;
 import com.vaadin.flow.server.BootstrapHandler;
+import com.vaadin.flow.server.Constants;
```
caalador, line 62: Extra import as the constants are already as static imports?

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
mshabarov: Replaced by `List`

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
mshabarov: Changed

flow-server/src/test/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandlerTest.java
mshabarov: I added Constant class explicitly for better readability.

caalador (dismissed):

caalador (approved):

PR #15872, opened 2023-02-09
fix: rename stats dependencies

### LINKED ISSUE #15871: Express build introduces a regression in license checks
### Description of the bug

The license checker is invoked for Vaadin 24 projects even when no commercial components are used. This should not happen. The reason is that the meaning of `npmModules` in stats.json has been changed without taking into consideration that other tools use it and rely on the old meaning. 

### Expected behavior

The change should be reverted and the old meaning restored.

```
rm ~/.vaadin/proKey
vaadin init --next hello-world
mvn install -Pproduction
```

### Versions

- Vaadin / Flow version: 24 beta1

caalador:
package dependencies have
new name and vite handled
dependencies go to npmModules

Closes #15871

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
@@ -357,7 +357,8 @@ public class TaskRunDevBundleBuild implements FallibleCommand {
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


Artur- (approved):

Artur-:
I will still test if this actually solves the issue...

Artur-:
Yes, seems to work as expected once you remember that the snapshot version is now 24.1-SNAPSHOT

mshabarov (approved):

PR #15914, opened 2023-02-15
fix: Do not process *.css.js as CSS files

Artur-:
The include/exclude patterns for postcssLit are glob patterns and `*.css\?*` matches e.g. "foo.css.js`


web-padawan (approved):

Artur-:
The patterns should be changed back from regexps once https://github.com/rollup/plugins/issues/1430 has been fixed

mshabarov (approved):

PR #15919, opened 2023-02-15
fix: Do not warn about eval usage that the user cannot affect

Artur-:


mshabarov (commented):
Is this for avoiding messages like these in the log?
```
Use of eval in "frontend/generated/jar-resources/FlowClient.js" is strongly discouraged as it poses security risks and may cause issues with minification.
Use of eval in "frontend/generated/jar-resources/FlowClient.js" is strongly discouraged as it poses security risks and may cause issues with minification.
```

mshabarov (approved):

Artur-:
Yes, exactly for those

PR #15944, opened 2023-02-17
fix: Client bootstrap should send navigationTrigger

### LINKED ISSUE #12770: NavigationTrigger is always CLIENT_SIDE for any user interactions with the app
### Description of the bug

For any user interactions with the app (page reload, using history, clicking router link) `AfterNavigationEvent#getLocationChangeEvent().getTrigger()` returns `CLIENT_SIDE` even though the `NavigationTrigger` enum has dedicated values for different navigation cases.

### Expected behavior

According to the JavaDoc, `NavigationTrigger` can be:
* `PAGE_LOAD` - Navigation was triggered by the user opening the application or reloading in the browser.
* `ROUTER_LINK`- Navigation was triggered by the user following a router link.
* `HISTORY`- Navigation was triggered by the user going forward or back in the browser's history.

1. Create a stub app suing: https://start.vaadin.com/app/
2. Alter the `MainLayout` as follows:

```
public MainLayout() {
    ...

    UI.getCurrent().addAfterNavigationListener(this::afterNavigation);
}

private void afterNavigation(AfterNavigationEvent event) {
    Notification.show(
            event.getLocationChangeEvent().getTrigger().name(),
            3000,
            Notification.Position.BOTTOM_END
    );
}
```

### Versions

- Vaadin / Flow version: 22.0.2
- Java version: 11
- OS version: Mac OS Monterey 12.1
  - Chrome 97.0.4692.71 (Official Build) (x86_64)
  - Safari 15.2 (17612.3.6.1.6)
  - FireFox 96.0.1 (64-bit)
- IDE (if applicable):

caalador:
When using client side bootstrapping
we should send info on what the navigation
trigger was so that we can give the
correct info on the server and not
only CLIENT_SIDE

Fixes #12770


flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
```diff
@@ -258,7 +259,9 @@ private void pushHistoryStateIfNeeded(NavigationEvent event, UI ui) {
                         "this.scrollPositionHandlerAfterServerNavigation($0);",
                         s));
             }
-        } else if (NavigationTrigger.ROUTER_LINK.equals(event.getTrigger())) {
+        } else if (NavigationTrigger.ROUTER_LINK.equals(event.getTrigger())
+                && Boolean.TRUE
```
mshabarov, line 263: I didn't get why is it needed to check server routing here, could you please explain?

mshabarov (commented):
Tested with steps described in the ticket and navigation trigger has expected value.
Though I didn't see tests covering this fix/feature, do you mind to add such tests?

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
caalador: Client routing through `connectClient` doesn't have any special handling (or event state) for the router_link trigger and should handle history state like all other navigations as it has done so until now.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/NavigationTriggerIT.java
```diff
@@ -30,23 +30,19 @@ public class NavigationTriggerIT extends ChromeBrowserTest {
     @Test
     public void testNavigationTriggers() {
         String url = getTestURL() + "/abc/";
+        System.err.println(url);
```
czp13, line 33: Is this part needed?
```suggestion
```

czp13 (dismissed):
Tiny/minor comment about logging into the error console, otherwise seems good.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/NavigationTriggerIT.java
tepi: Removed

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1676,7 +1676,7 @@ public String getForwardToClientUrl() {
     @AllowInert
     public void connectClient(String clientElementTag, String clientElementId,
```
knoobie, line 1677: Just wondering: is it possible to overload ClientCallable methods to also add the old signature for e.g. the case when clients cached the js stuff? There where multiple reports about this kind of error in the last update where Artur changed the signature. 

Suggestion (if possible): Add the same method with the old signature (deprecated) forwarding the call with trigger==empty for e.g. 23.3, 24.0 and 24.1 and removal of the deprecated in 24.2 to reduce the noise that this could create again

tepi: I don't see what harm it could do. Added it back, with fixed "client" trigger in order for it to behave like before this change.

knoobie: Looks like it's not supported 😅

tepi: D'oh. Should've known that.. Well, looks like not doable after all.

knoobie: Good argument against it as well tho

czp13 (approved):

PR #15960, opened 2023-02-20
chore: add frontend dev setup to vaadin-dev-tools

sissbruecker:
## Description

Adds a frontend dev setup to the `vaadin-dev-tools` module in order to be able to work iteratively on the dev tools UI.

The setup includes a dev server that hosts the Typescript files for the dev tool UI, and provides hot module replacement functionality so that changes to the dev tool don't require a page reload. In order to use the local dev tool files in a Flow app, the setup includes a script that patches a Flow app to load the dev tool from the dev server instead of from the classpath.

Due to how the dev tool Lit component is constructed, changes to tab contents currently don't work correctly, but changes to the outer HTML or CSS should be visible. If this is merged I'll open a separate PR to refactor the component to make this work as well.


MarcinVaadin (dismissed):
LGTM, tested and working

MarcinVaadin (approved):

PR #15980, opened 2023-02-21
feat: ThemeModifier support for DebugWindowConnection command

MarcinVaadin:
## Description
Added `ThemeModifier` that inserts or updates CSS rules in `theme-editor.css` of given theme. `@import "theme-editor.css";` is added automatically to `styles.css`.

Added support for `themeEditorSetRule` command of json object data structure:
```
var data = {
  rules: [
    {
      selector: 'vaadin-button::part(label)',
      property: 'color',
      value: 'red'
    },
    {
      selector: 'vaadin-text-field::part(label)',
      property: 'color',
      value: 'green'
    }
  ]
};
```

To send command use `Connection.send('themeEditorSetRule', data);`.

In general due to selectors complexity (normal part _vaadin-button_ and function part _::part(label)_) rule is parsed again by ph-css library using `selector {property: value}` string.

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/ThemeModifier.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/ThemeModifier.java
@@ -1,7 +1,9 @@
 package com.vaadin.base.devserver;
 
 import com.helger.css.ECSSVersion;
-import com.helger.css.decl.*;
+import com.helger.css.decl.CSSImportRule;
+import com.helger.css.decl.CSSStyleRule;
+import com.helger.css.decl.CascadingStyleSheet;
 import com.helger.css.reader.CSSReader;
 import com.helger.css.writer.CSSWriter;
 import com.vaadin.flow.server.VaadinContext;
@@ -9,11 +11,15 @@ import com.vaadin.flow.server.frontend.FrontendUtils;
 import com.vaadin.flow.server.startup.ApplicationConfiguration;
 import elemental.json.JsonArray;
 import elemental.json.JsonObject;
+import org.apache.commons.io.IOUtils;
 
 import java.io.File;
+import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
 
 public class ThemeModifier {
 
@@ -23,6 +29,8 @@ public class ThemeModifier {
             + "manual changes may be overwritten while using ThemeEditor";
 
     private VaadinContext context;
+
+    private boolean importPresent;
 
     public ThemeModifier(VaadinContext context) {
         this.context = context;
@@ -63,6 +71,11 @@ public class ThemeModifier {
         CascadingStyleSheet styleSheet = CSSReader.readFromFile(styles,
                 StandardCharsets.UTF_8, ECSSVersion.LATEST);
 
+        if (!importPresent) {
+            insertImportIfNotExists();
+            importPresent = true;
+        }
+
         CSSStyleRule rule = parseStyleRule(selector, property, value);
         removeRuleIfExists(styleSheet, rule);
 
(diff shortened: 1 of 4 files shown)


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/ThemeModifier.java
```diff
@@ -0,0 +1,164 @@
...
+                        "Cannot insert theme-editor.css @import", e);
+            } finally {
+                IOUtils.closeQuietly(writer);
+            }
+        }
+    }
+
+    protected void removeRuleIfExists(CascadingStyleSheet styleSheet,
+            CSSStyleRule rule) {
+        String propertyName = rule.getDeclarationAtIndex(0).getProperty();
```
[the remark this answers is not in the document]
MarcinVaadin: @sonatype-lift ignoreall

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/ThemeModifier.java
```diff
@@ -0,0 +1,317 @@
...
+        } catch (IOException e) {
+            throw new IllegalStateException("Cannot write " + styles.getPath(),
+                    e);
+        }
+    }
+
+    protected void sortStylesheet(CascadingStyleSheet styleSheet) {
+        List<CSSStyleRule> sortedRules = styleSheet.getAllStyleRules().stream()
+                .sorted(Comparator.comparing(
+                        r -> r.getSelectorAtIndex(0).getAsCSSString()))
```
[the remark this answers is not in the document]
MarcinVaadin: @sonatype-lift ignoreall

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/ThemeModifier.java
```diff
@@ -0,0 +1,304 @@
...
+                    .collect(Collectors.toList());
+            rule.removeAllDeclarations();
+            sortedDeclarations.forEach(rule::addDeclaration);
+        }
+        sortedRules.forEach(styleSheet::removeRule);
+        sortedRules.forEach(styleSheet::addRule);
+    }
+
+    protected CSSStyleRule parseStyleRule(String selector, String property,
+            String value) {
```
sissbruecker, line 204:
```suggestion
    protected CSSStyleRule createStyleRule(String selector, String property,
            String value) {
```

That parsing is used to create the rule is an implementation detail.

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/ThemeModifierTest.java
```diff
@@ -0,0 +1,238 @@
...
+        assertThemeEditorCssContains(SELECTOR_WITH_PART + " { color:red; }");
+    }
+
+    @Test
+    public void ruleExists_ruleIsUpdated() {
+        ThemeModifier modifier = new TestThemeModifier();
+        modifier.setCssRule(SELECTOR_WITH_PART, "color", "red");
+        modifier.setCssRule(SELECTOR_WITH_PART, "color", "blue");
+        assertThemeEditorCssNotContains(SELECTOR_WITH_PART + " { color:red; }");
+        assertThemeEditorCssContains(SELECTOR_WITH_PART + " { color:blue; }");
```
sissbruecker, line 156: This and several of the following tests don't guarantee that only a single rule is created. 

In general I feel that working with string patterns is not a good fit for these tests. How about reusing the CSS parser here? It would make the code more verbose, but also more straight-forward, as you don't have to think all the time about the correct string pattern, or care about formatting. A utility function could return all rules in a file. A test could then check: number of rules, order of rules, number of property declarations, and contents of those.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/ThemeModifier.java
```diff
@@ -0,0 +1,304 @@
...
+        if (!themeEditorStyles.canWrite()) {
+            throw new IllegalStateException(
+                    themeEditorStyles.getPath() + " is not writable.");
+        }
+
+        return themeEditorStyles;
+    }
+
+    public boolean createDefaultTheme() {
+        File theme = new File(getFrontendFolder(), "themes/" + DEFAULT_THEME);
```
sissbruecker, line 108: From a glance, I'm not sure if this would work on Windows.
Consider joining paths in an OS independent way, e.g. using `Paths`, as is done here:
https://github.com/vaadin/flow/blob/4a5399073674a581907a96f357f43a63470e44a6/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java#L255-L256

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/ThemeModifierTest.java
MarcinVaadin: I was thinking about it but as the idea of tests was to check the changes directly within files I've used pure file search. But indeed, reusing parser would allow to enhance tests in future.

sissbruecker (approved):

PR #15982, opened 2023-02-22
chore: Update file upload to latest

mshabarov:


flow-server/pom.xml
```diff
@@ -79,15 +79,7 @@
...
             <artifactId>commons-fileupload</artifactId>
-            <version>1.4</version>
-            <!-- commons-fileupload:1.3.3 depends on commons-io:2.2, override
-                it with Flow declaration -->
-            <exclusions>
-                <exclusion>
-                    <groupId>commons-io</groupId>
-                    <artifactId>commons-io</artifactId>
-                </exclusion>
-            </exclusions>
```
knoobie, line 90: Note from discord 🙂 
the fix is more than just updating the version - you probably have to set sane default values (this was clarified on the mailing list) https://lists.apache.org/thread/4xl4l09mhwg4vgsk7dxqogcjrobrrdoy

Artur-: The limits should be set in a separate PR for `main`

Artur- (approved):

flow-server/pom.xml
knoobie: If you think about the correct limit, comment by Mark Thomas: 

> Tomcat, for example, integrated this with the maxParameterCount attribute (default 10,000) so the total number of query string parameters + field parts + file parts is limited to 10,000. That should should be more than sufficient for nearly all applications while still being several orders of magnitude below the point where DoS effects will be observed.

PR #15986, opened 2023-02-22
feat: Enable running npm ci mode (#15579)

### LINKED ISSUE #15579: Enable running npm ci instead of install
### Describe your motivation

CI builds would benefit of the possibility to install the npm packages using `npm ci` instead of `npm install` (or for pnpm use `pnpm install --frozen-lockfile`) to install the exact versions in the lock file.

Add a system variable (`vaadin.ci.build`?) that would order TaskRunNpmInstall to use `ci` instead of `install`.

### Additional context

the vaadin maven plugin executes "npm install" instead of "npm ci" in build flow, and that can potentially overwrite package/package-lock.json.

tepi:
Fixes #15579.

Adds possibility to install the npm packages using npm ci instead of npm install (or for pnpm: pnpm install --frozen-lockfile) to install the exact versions in the lock file. Can be enabled with ci.build parameter when executing a production mode frontend build.


flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
```diff
@@ -216,6 +216,15 @@ public open class VaadinFlowPluginExtension(project: Project) {
      */
     public var frontendHotdeploy: Boolean = false
 
+    /**
+     * Setting this to true will run {@code npm ci} instead of {@code npm install} when using npm.
+     *
+     * If using pnpm, the install will be run with {@code --frozen-lockfile} parameter.
+     *
+     * This makes sure that the package lock file will not be overwritten.
```
mshabarov, line 224:
```suggestion
     * This makes sure that the versions in package lock file will not be overwritten and production builds are reproducible.
```

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
```diff
@@ -92,6 +93,18 @@ public class BuildFrontendMojo extends FlowModeAbstractMojo
...
     private boolean optimizeBundle;
 
+    /**
+     * Setting this to true will run {@code npm ci} instead of
+     * {@code npm install} when using npm.
+     *
+     * If using pnpm, the install will be run with {@code --frozen-lockfile}
+     * parameter.
+     *
+     * This makes sure that the package lock file will not be overwritten.
```
mshabarov, line 103:
```suggestion
     * This makes sure that the versions in package lock file will not be overwritten and production builds are reproducible.
```

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/PluginAdapterBuild.java
```diff
@@ -65,4 +65,16 @@ public interface PluginAdapterBuild extends PluginAdapterBase {
...
     boolean runNpmInstall();
 
+    /**
+     * Setting this to true will run {@code npm ci} instead of
+     * {@code npm install} when using npm.
+     *
+     * If using pnpm, the install will be run with {@code --frozen-lockfile}
+     * parameter.
+     *
+     * This makes sure that the package lock file will not be overwritten.
```
mshabarov, line 75:
```suggestion
     * This makes sure that the versions in package lock file will not be overwritten and production builds are reproducible.
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -138,18 +138,29 @@ public String getPackageManager() {
...
+                    + "` to "
                     + "resolve and optionally download frontend dependencies. "
                     + "This may take a moment, please stand by...");
             runNpmInstall();
 
             updateLocalHash();
         } else {
             packageUpdater.log().info(
-                    "Skipping `{} install` because the frontend packages are already "
+                    "Skipping `{} {}` because the frontend packages are already "
```
mshabarov, line 160: Npm-ci docs https://docs.npmjs.com/cli/v9/commands/npm-ci say that "If a node_modules is already present, it will be automatically removed before npm ci begins its install.", so I think we should not skip the `npm ci` if the node_modules are already installed. Instead, I propose to delete `node_modules` in `ciBuild` mode and always run the command, otherwise the build wouldn't be reproducible always. What do you think?

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
tepi: Seems reasonable to me. Do you think it should be done for the pnpm case as well?

mshabarov: I think yes

tepi: I changed the cleanup method to always delete node modules if ci build is run.

mshabarov (approved):

PR #16002, opened 2023-02-23
fix: Not calculate frontend resource hash for production mode

mshabarov:
Production mode doesn't need hashes for frontend resources and, moreover, the calculation takes time. This patch makes hash calculation happen only for express build dev mode, where these hashes are used actually.

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -195,74 +195,84 @@ function statsExtracterPlugin(): PluginOption {
         })
         .sort()
         .filter((value, index, self) => self.indexOf(value) === index);
-      const npmModuleAndVersion = Object.fromEntries(npmModules.map((module) => [module, getVersion(module)]));
+
+      let stats;
+
+      if (!devBundle) {
+        // production mode: no hashes are needed
+        stats = { npmModules };
+      } else {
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -195,74 +195,84 @@ function statsExtracterPlugin(): PluginOption {
...
-            const themeJsonContent = readFileSync(themeJson, {encoding: 'utf-8'}).replace(/\r\n/g, '\n');
-            const themeJsonContentAsJson = JSON.parse(themeJsonContent);
-            const assets = themeJsonContentAsJson.assets;
-            if (assets) {
-              const hash = createHash('sha256').update(themeJsonContent, 'utf8').digest("hex");
-              themeJsonHashes[`${path.basename(themeFolder)}`] = hash;
-            }
+      if (!devBundle) {
+        // production mode: no hashes are needed
+        stats = { npmModules };
```
Artur-, line 203: This is a different value ?

mshabarov: This is how stats was populated before Vaadin 24.0. Apparently Flow now misses some data in stats for production after these changes.

mshabarov:
This needs re-working because we are going to implement default production bundle feature. So the hashes would be needed.

PR #16003, opened 2023-02-23
refactor: add basic theme editor panel

sissbruecker:
## Description

Adds a basic theme editor tab that allows to pick an element, and then displays stylable parts and properties based on the picked component's metadata. The tab will only be added when the theme editor has been enabled in the backend.

This does not handle the missing theme state currently, planning to add this in a separate PR. Also no special handling for when the user picks an element for which we don't have metadata, as it's not clear what should happen at this point.


vaadin-dev-server/package.json
```diff
@@ -2,13 +2,14 @@
...
     "test": "web-test-runner --node-resolve",
+    "test-watch": "web-test-runner --node-resolve --watch",
     "patch-app": "node patch-application.js",
     "prettier": "prettier --write src/main/resources/META-INF/frontend/**/*.ts"
   },
   "devDependencies": {
-    "@esm-bundle/chai": "^4.3.4-fix.0",
     "@koa/cors": "^4.0.0",
     "@open-wc/dev-server-hmr": "^0.1.2-next.0",
+    "@open-wc/testing": "^3.1.7",
```
sissbruecker, line 12: Added `@open-wc/testing` which includes utils for rendering fixtures in tests. Removed chai, is it's included in `@open-wc/testing`.

vaadin-dev-server/src/main/resources/META-INF/frontend/vaadin-dev-tools/vaadin-dev-tools.ts
```diff
@@ -1050,20 +1055,7 @@ export class VaadinDevTools extends LitElement {
...
-      } else if (message?.command === 'vaadin-dev-tools-code-ok') {
-        if ((window as any).Vaadin.Flow) {
-          this.tabs.push({ id: 'code', title: 'Code', render: () => this.renderCode() });
-        }
-      } else {
-        // eslint-disable-next-line no-console
-        console.error('Unknown message from front-end connection:', JSON.stringify(message));
-      }
-    };
+    frontendConnection.onMessage = (message: any) => this.handleFrontendMessage(message);
```
sissbruecker, line 1058: Extracted this into a public method, to be able to easily send messages from tests. Ideally the whole `Connection` construct would be mockable / provided as dependency, but that requires more work.

MarcinVaadin (approved):

PR #16006, opened 2023-02-23
fix: Rebuild the dev bundle upon changes in theme.json

mshabarov:
Makes a new dev bundle once sees changes in project's theme.json file or in packaged themes. Previously it took only in account the assets list changes from packaged themes.


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -211,34 +214,67 @@ private static boolean exportedWebComponents(JsonObject statsJson,
...
     }
 
-    private static boolean packagedThemeAddedOrUpdated(Options options,
-            JsonObject statsJson) {
-        Map<String, String> packagedThemeHashes = new HashMap<>(1);
+    private static boolean themeConfigurationChanged(Options options,
+            JsonObject statsJson,
+            FrontendDependenciesScanner frontendDependencies)
+            throws IOException {
+        Map<String, String> themeJsonHashes = new HashMap<>(1);
```
Artur-, line 221: nit: I don't think `1` is necessary here

Artur- (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov:
```suggestion
        Map<String, String> themeJsonHashes = new HashMap<>();
```

PR #16015, opened 2023-02-23
chore: Upgrade to Spring Boot 3.0.3

Artur-:
Upgrades Spring to 6.0.5 and Spring Security to 6.0.2


Artur-:
This does not work because of https://github.com/spring-projects/spring-framework/issues/30041

As Spring creates extra instances, the component mapping that uses a thread local gets the wrong component. Iit assumes that one component instance is created when one component instance is requested, but now Spring creates multiple instances and the first one is used. This is not the correct one

PR #16024, opened 2023-02-24
fix!: Use feature detection to show the unsupported browser page (#15813) (CP: 23.3)

### LINKED ISSUE #14366: Use feature detection instead of browser version detection for the "too old" page

Artur-:
Changed to detect ResizeObserver

This removes the `WebBrowser. isTooOldToFunctionProperly()` method as it is no longer possible to determine this information based on the user agent.

Fixes #14366


mshabarov:
Just to double check, is this feature so critical that we want to pick in to the maintenance release with potential breaking changes?

Artur-:
Could be better if it went into 23.4, the day that one is made. It is not supposed to break things, just make detection more accurate

tepi (dismissed):

Artur-:
Let's not do this for 23.3. If we do a 23.4, we can revisit it

mshabarov (approved):

PR #16029, opened 2023-02-24
refactor: skip the node_modules folder completely

vursen:
## Description

Use `Files.walkFileTree` instead of `Files.walk` in the Polymer to Lit converter to prevent it from visiting the `node_modules` folder.

## Type of change

- [x] Refactor

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
@@ -18,10 +18,14 @@ package com.vaadin.flow.plugin.base;
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.nio.file.FileSystems;
+import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
 import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -152,11 +156,31 @@ public class ConvertPolymerCommand implements AutoCloseable {
         PathMatcher matcher = FileSystems.getDefault()
                 .getPathMatcher("glob:" + glob);
 
-        try (Stream<Path> walk = Files.walk(baseDir)) {
-            return walk.filter(path -> matcher.matches(path))
-                    .filter(path -> !path.toString().contains("node_modules"))
-                    .collect(Collectors.toList());
-        }
+        List<Path> matchingPaths = new ArrayList<>();
+
+        Files.walkFileTree(baseDir, new SimpleFileVisitor<Path>() {
+            @Override
+            public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs)
+                    throws IOException {
+                if (dir.toString().contains("node_modules")) {
+                    return FileVisitResult.SKIP_SUBTREE;
+                }
+
+                return FileVisitResult.CONTINUE;
+            }
+
+            @Override
+            public FileVisitResult visitFile(Path file, BasicFileAttributes attrs)
+                    throws IOException {
+                if (matcher.matches(file)) {
+                    matchingPaths.add(file);
+                }
+
+                return FileVisitResult.CONTINUE;
+            }
+        });
+
+        return matchingPaths;
     }
 
     private Path getLookupPath() {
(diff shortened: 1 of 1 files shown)


flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
```diff
@@ -152,11 +156,31 @@ private List<Path> getFilePathsByGlob(Path baseDir, String glob)
...
-                    .filter(path -> !path.toString().contains("node_modules"))
-                    .collect(Collectors.toList());
-        }
+        List<Path> matchingPaths = new ArrayList<>();
+
+        Files.walkFileTree(baseDir, new SimpleFileVisitor<Path>() {
+            @Override
+            public FileVisitResult preVisitDirectory(Path dir,
+                    BasicFileAttributes attrs) throws IOException {
+                if (dir.toString().contains("node_modules")) {
```
Artur-, line 165: If this walks the project directory then it would make sense to skip `target` also... or just walk through `frontend` and `src`

vursen: Right, but given that the developer can provide a custom path to be converted, I think it will be easier to exclude the `target` folder.

czp13:
Is this PR is on progress? Can the flow team help here in any way?

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
```diff
@@ -152,11 +156,31 @@ private List<Path> getFilePathsByGlob(Path baseDir, String glob)
...
+                    return FileVisitResult.SKIP_SUBTREE;
+                }
+
+                return FileVisitResult.CONTINUE;
+            }
+
+            @Override
+            public FileVisitResult visitFile(Path file,
+                    BasicFileAttributes attrs) throws IOException {
```
czp13, line 174: I believe IOException will never thrown:
```suggestion
                    BasicFileAttributes attrs)  {
```

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
```diff
@@ -152,11 +156,31 @@ private List<Path> getFilePathsByGlob(Path baseDir, String glob)
...
-            return walk.filter(path -> matcher.matches(path))
-                    .filter(path -> !path.toString().contains("node_modules"))
-                    .collect(Collectors.toList());
-        }
+        List<Path> matchingPaths = new ArrayList<>();
+
+        Files.walkFileTree(baseDir, new SimpleFileVisitor<Path>() {
+            @Override
+            public FileVisitResult preVisitDirectory(Path dir,
+                    BasicFileAttributes attrs) throws IOException {
```
czp13, line 164: Here as well, maybe we could remove it:
```suggestion
                                                     BasicFileAttributes attrs) {
```

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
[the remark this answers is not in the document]
vursen: I'm afraid this will cause the converter to skip any folder that contains `target` in its name which isn't 100% correct. Only the root `target` folder needs to be skipped, I think.

vursen:
> Is this PR is on progress? Can the flow team help here in any way?

Help would be appreciated. After https://github.com/vaadin/flow/pull/16029#discussion_r1116881380 is resolved, I guess this PR can be merged.

Artur- (approved):
Let's not block this improvement because another improvement could be made.

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
Artur-: `IOException` is defined by the interface

czp13: True, my Intellij let me deleted it, so I think it is semantically correct, also could make sense that the implementation is less restrictive than the interface itself. But I am fine both ways!

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/ConvertPolymerCommand.java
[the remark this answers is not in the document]
czp13: Thanks, true! Then maybe it can be an improvement to this later on as it is merged already :)

PR #16039, opened 2023-02-24
fix: Ignore ?inline suffix for imported stylesheets

mshabarov:
Frontend resources checking ignores ?inline suffixes in the stylesheet path when comparing file names.


Artur- (approved):
LGTM. In some cases Vite uses some other "query parameters" but maybe they are not relevant for this

mshabarov:
> LGTM. In some cases Vite uses some other "query parameters" but maybe they are not relevant for this

Yes, that's why I search for '?', not '?inline'.

Artur-:
Right... I saw was I thought I was going to see :)

PR #16058, opened 2023-02-27
fix: Do not throw exception if no IDE is found

### LINKED ISSUE #16043: java.util.NoSuchElementException: No value present if using component finder when running through mvn
### Description of the bug

If you run a new app from the command line as `mvn spring-boot:run` and use the dev tools component picker, if fails with
```
Caused by: java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.get(Optional.java:143) ~[na:na]
	at com.vaadin.base.devserver.OpenInCurrentIde.getCommandAndArguments(OpenInCurrentIde.java:167) ~[vaadin-dev-server-24.0.0.beta2.jar:na]
	at com.vaadin.base.devserver.OpenInCurrentIde.getIdeaBinary(OpenInCurrentIde.java:195) ~[vaadin-dev-server-24.0.0.beta2.jar:na]
	at com.vaadin.base.devserver.OpenInCurrentIde.isIdea(OpenInCurrentIde.java:191) ~[vaadin-dev-server-24.0.0.beta2.jar:na]
	at com.vaadin.base.devserver.OpenInCurrentIde.findIdeCommand(OpenInCurrentIde.java:159) ~[vaadin-dev-server-24.0.0.beta2.jar:na]
	at com.vaadin.base.devserver.OpenInCurrentIde.findIdeCommandInfo(OpenInCurrentIde.java:154) ~[vaadin-dev-server-24.0.0.beta2.jar:na]
	at com.vaadin.base.devserver.OpenInCurrentIde.openFile(OpenInCurrentIde.java:62) ~[vaadin-dev-server-24.0.0.beta2.jar:na]
	at com.vaadin.base.devserver.IdeIntegration.internalShowInIde(IdeIntegration.java:91) ~[vaadin-dev-server-24.0.0.beta2.jar:na]
	at com.vaadin.base.devserver.IdeIntegration.showComponentCreateInIde(IdeIntegration.java:57) ~[vaadin-dev-server-24.0.0.beta2.jar:na]
	at com.vaadin.base.devserver.DebugWindowConnection.lambda$onMessage$599dae22$1(DebugWindowConnection.java:233) ~[vaadin-dev-server-24.0.0.beta2.jar:na]
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:577) ~[na:na]
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317) ~[na:na]
	at com.vaadin.flow.server.VaadinService.runPendingAccessTasks(VaadinService.java:2028) ~[flow-server-24.0.0.beta2.jar:24.0.0.beta2]
```

### Expected behavior

The component picker is supposed to output the file + line number to the console

vaadin init --next hello
cd hello
mvn
# pick component

### Versions

- Vaadin version: 24 beta5
- Java version: 17
- OS version: macos

Artur-:
Fixes #16043

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/OpenInCurrentIde.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/OpenInCurrentIde.java
@@ -164,7 +164,10 @@ public final class OpenInCurrentIde {
     }
 
     private static String getCommandAndArguments(Info info) {
-        return info.commandLine().get();
+        if (info.commandLine().isPresent()) {
+            return info.commandLine().get();
+        }
+        return null;
     }
 
     private static List<ProcessHandle> getParentProcesses() {
@@ -178,13 +181,17 @@ public final class OpenInCurrentIde {
     }
 
     static boolean isEclipse(Info info) {
-        String lowerCase = info.command().get().toLowerCase(Locale.ENGLISH);
-        // Eclipse has a lot of other products like Temurin and Adoptium so we
-        // cannot
-        // check with "contains"
-        return lowerCase.endsWith("eclipse")
-                || lowerCase.endsWith("eclipse.exe");
+        Optional<String> cmd = info.command();
+        if (cmd.isPresent()) {
+            String lowerCmd = cmd.get().toLowerCase(Locale.ENGLISH);
+            // Eclipse has a lot of other products like Temurin and Adoptium so
+            // we
+            // cannot check with "contains"
+            return lowerCmd.endsWith("eclipse")
+                    || lowerCmd.endsWith("eclipse.exe");
+        }
 
+        return false;
     }
 
     static boolean isIdea(Info info) {
@@ -193,7 +200,8 @@ public final class OpenInCurrentIde {
 
     private static String getIdeaBinary(Info info) {
         String commandAndArguments = getCommandAndArguments(info);
-        if (commandAndArguments.contains("idea_rt.jar")) {
+        if (commandAndArguments != null
+                && commandAndArguments.contains("idea_rt.jar")) {
             String replaced = commandAndArguments
                     .replaceFirst(".*[:;]([^:;]*)(idea_rt.jar).*", "$1$2");
             if (!replaced.equals(commandAndArguments)) {
@@ -208,7 +216,8 @@ public final class OpenInCurrentIde {
                 }
             }
         }
-        if (info.command().get().contains("idea")) {
+        if (info.command().isPresent()
+                && info.command().get().contains("idea")) {
             return info.command().get();
         }
         return null;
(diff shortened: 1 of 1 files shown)


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/OpenInCurrentIde.java
```diff
@@ -164,7 +164,10 @@ static Optional<Info> findIdeCommand(List<Info> processes) {
     }
 
     private static String getCommandAndArguments(Info info) {
-        return info.commandLine().get();
+        if (info.commandLine().isPresent()) {
+            return info.commandLine().get();
+        }
+        return null;
```
mcollovati, line 170: It can perhaps be simplified as
```suggestion
        return info.commandLine().orElse(null);
```

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/OpenInCurrentIde.java
```diff
@@ -208,7 +215,8 @@ private static String getIdeaBinary(Info info) {
                 }
             }
         }
-        if (info.command().get().contains("idea")) {
+        if (info.command().isPresent()
+                && info.command().get().contains("idea")) {
             return info.command().get();
         }
         return null;
```
mcollovati, line 222: nit: I would use a filter here, but current solution is also fine
```suggestion
        return info.command()
                .filter(cmd -> cmd.contains("idea"))
                .orElse(null);
```

mcollovati (commented):
Does it make sense to add a test for this use case?

mcollovati:
As a side note, does it make sense to disable the "Find component in code" button when not running from IDE?
I also wonder if we can cache the IDE detection result in `OpenInCurrentIde` so that it will be performed only once

Artur-:
I thought we had no good way of testing this but we actually do.... added a test

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/OpenInCurrentIde.java
Artur-: Thanks, done

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/OpenInCurrentIde.java
Artur-: Thanks, done

Artur-:
> As a side note, does it make sense to disable the "Find component in code" button when not running from IDE?

One point to having it is that even if we do not detect the IDE, the IDE might make the file+linenumber clickable in the console when we output it

Artur-:
>I also wonder if we can cache the IDE detection result in OpenInCurrentIde so that it will be performed only once

Yes, but let's not do it in this PR :)

mcollovati (approved):

PR #16060, opened 2023-02-27
refactor: disable live reload on CSS changes in theme editor

sissbruecker:
## Description

Temporarily disable live reload for five seconds when saving pending changes from the theme editor, in order to prevent a page reload.

The theme editor doesn't yet update its internal state after saving the changes, will fix that in a follow-up PR.


MarcinVaadin (approved):
LGTM, live theme reloading is also under investigation in https://github.com/vaadin/flow/pull/15631

PR #16063, opened 2023-02-27
fix: Use context class loader for class finder

Artur-:
Fixes https://github.com/vaadin/hilla/issues/791


platosha:
Would it be a good idea to add a unit test using a mock here?

Artur-:
Probably, but what should we test? Explicilty that the code uses the context classloader?

platosha:
Yes, and not the class loader from classes or elsewhere.

Artur-:
Added a basic test

mshabarov (approved):
I understand that the previous code had a moving parts with picking first class loader from the set. Now the context class loader is used that makes the class loading more deterministic. But I still not sure that this approach wouldn't raise other issue with classloading in application servers. Would ask for one more reviewer to check this PR.

Artur-:
It is not only to be deterministic, but if you do not use the restart class loader in a spring boot application then you will not get the latest classes after a dev tools restart. This is the Hilla issue that changes in classes are not picked up after a dev tools restart

PR #16064, opened 2023-02-27
refactor: ThemeEditor - support for setting / removing class names

MarcinVaadin:
## Description
Added support for setting / removing class name of variable inside single code block.
```
public class TestView extends VerticalLayout {

    private TextField pinField = new TextField("Cant touch this"); <--- doesn't work for this case

    public TestView() {
        TextField textField = new TextField("Your name"); <--- works for that case
        [...]
        add(textField, button, new TextField("Cant touch this either")); <--- doesn't work for this case
    }
}
```
New debug connection message `themeEditorClassName` support has been added with message:
```
var data = {
 uiId: 0,
 nodeId: 5,
 add: ['bold', 'beautiful'],
 remove: ['ugly']
};
```
where `uiId` and `nodeId` are returned by component picker, `add` and `remove` are arrays of classnames.

Source files are being modified using https://javaparser.org/.


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/ThemeEditorMessageHandler.java
```diff
@@ -0,0 +1,107 @@
...
+                int uiId = (int) data.getNumber("uiId");
+                int nodeId = (int) data.getNumber("nodeId");
+                List<String> classNamesToBeAdded = toClassNameList(
+                        data.get("add"));
+                List<String> classNamesToBeRemoved = toClassNameList(
+                        data.get("remove"));
+                sourceModifier.setClassNames(uiId, nodeId, classNamesToBeAdded);
+                sourceModifier.removeClassNames(uiId, nodeId,
+                        classNamesToBeRemoved);
+                return true;
```
sissbruecker, line 68: Since there are a lot of things that can go wrong here we should send a response to inform the client whether classes have been successfully applied. For one this allows us to notify the user if they can proceed with styling this specific instance, and it would also allow us to suggest workarounds (refactor code to extract component into variable, or manually add class name `xyz` to the component).

Another thing I'm not sure about is sending a generic list of class names from the client. For now we have explicitly one use-case: generate a unique class name for that specific component instance and apply it to the component. Since we already use CSS parsing in the dev server, and we know the single relevant CSS file, generating a unique class name might actually be easier to do here by checking the existing CSS file for conflicts. So maybe we should rather have a command that either generates a unique class name for a component, or otherwise returns the existing class name. There are some open questions, such as how to identify whether the component already has its own component-instance specific class name, but those would apply regardless whether we solve this on the client or the server. Anyway, not a blocker, just something to keep in mind. For now we can generate some random class name on the client, and use a pattern that identifies the class name as the component instance specific one.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifier.java
```diff
@@ -0,0 +1,264 @@
...
+                    location);
+
+            BlockStmt parentBlock = getParentBlockStmt(node).get();
+            String variableName = getVariableName(node);
+            classNames.forEach(className -> {
+                ExpressionStmt methodCall = createMethodCallExprStmt(
+                        variableName, "addClassName", className);
+                parentBlock.getStatements().stream()
+                        .filter(s -> simpleMethodCallExprStmtFilter(s,
+                                methodCall))
```
sissbruecker, line 124: Comparing this to the add logic, I wonder if we need a custom filter here, when for adding we can just check using `NodeList.contains`. It seems nodes implement `equals`/`hashCode` so that method name, arguments, argument values are accounted for. So could we then try to find the statement that equals `methodCall`, and remove that?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifier.java
```diff
@@ -0,0 +1,264 @@
...
+            IOUtils.write(LexicalPreservingPrinter.print(cu), writer);
+        } catch (IOException e) {
+            throw new ModifierException(
+                    "Cannot update file: " + sourceFile.getPath(), e);
+        } finally {
+            IOUtils.closeQuietly(writer);
+        }
+    }
+
+    public void addThemeAnnotation() {
```
sissbruecker, line 183: Nit: Would suggest to add this in a separate PR, otherwise it's confusing that a comment in `ThemeModifier` refers to a method that doesn't actually do anything.

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifierTest.java
```diff
@@ -0,0 +1,230 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+
+public class JavaSourceModifierTest {
```
sissbruecker, line 39: Tests look good, but for Java sources we might want to add a smoke test that formatting / white-space is preserved. Preferrably with some weird stuff like putting opening parenthesis on a new line, and wrong indents. I imagine our target group is much more particular about their Java sources than about the generated theme editor CSS  🙂.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifier.java
MarcinVaadin: Yes, it's a placeholder for future work.

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifierTest.java
MarcinVaadin: I'm using `LexicalPreservingPrinter` which preserves all whitespaces, but indeed, test would be good.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifier.java
MarcinVaadin: Node consists of many other properties that are not present in new node but are used for comparison, like Range - not available yet in new nodes.

I'm comparing only 3 required parts of MethodCall expression: `this.ofThatMethod(andThisArgument)`.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/ThemeEditorMessageHandler.java
MarcinVaadin: The things I can do currently:
1. Introduce getter operation for already defined classNames and variants for picked component
2. Return response with status of request processing

MarcinVaadin: Left for next PR.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifier.java
MarcinVaadin: Removed.

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifierTest.java
MarcinVaadin: Added a test for file with messed formatting.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifier.java
MarcinVaadin: Updated comparison.

sissbruecker (approved):

PR #16071, opened 2023-02-28
test: Use Jetty 12 beta

Artur-:


Artur-:
Embedding tests get two session ids because of https://github.com/eclipse/jetty.project/issues/9459

Artur-:
Let's revisit this when there is a new beta

Issue #16004, opened 2023-02-23
Advanced checking of the theme.json content for development bundle


mshabarov:
### Describe your motivation

Vaadin 24.0 uses the frontend development bundle by default when app is running in development mode (aka Express Build mode). Among other checks, it compares the hash of the `theme.json` content of the project (part of the custom theme feature) with the hash stored in a bundle, i.e. the `theme.json` which was used to build the bundle.

The hash is calculated simply based in a string content of the file, not taking into account what kind of changes are in the file.
This would lead to generating a new application bundle (instead of using the default bundle from Vaadin platform):
- when the content of theme.json is formatted
- when platform adds a new assets/lumo imports/css imports

Basically if the default bundle is enough to run the project, Flow shouldn't make a new bundle. For example, if the platform's default bundle have extra assets defined in it, this doesn't affect the application unless it starts using these assets.

### Describe the solution you'd like

Flow should parse and analyse the content of the `theme.json` and decide if the default theme.json has all the entries from the application's theme.json. Instead of using hash, Flow can store the content in the `stats.json` of the bundle and using it for resolution.

PR #16076, opened 2023-02-28
feat: Search for missed entries in theme.json

mshabarov:
Flow now analyses the json content and entries in theme.json, not only the string content. It searches for entries in project's theme.jsom whose aren't in the default/app dev bundle. If any entries are found, Flow does re-bundling.

Fixes https://github.com/vaadin/flow/issues/16004

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -1,7 +1,7 @@
 name: Flow Validation
 on:
   push:
-    branches: [main, '23.3', '23.2',  '23.1', '9.0']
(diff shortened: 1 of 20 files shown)


manolo:
Still does a compilation with this project, it is just a default one downloaded from start.vaadin.com

[default-next.zip](https://github.com/vaadin/flow/files/10851407/default-next.zip)

manolo:
After recompiling flow, and platform, I tested and it works fine

manolo (approved):

PR #16077, opened 2023-02-28
refactor: editable host properties in theme editor

sissbruecker:
## Description

- Extends component metadata to include host properties
- Show editors for host properties in theme editor
- Update button metadata to use host properties, as proposed by Jouni
- Update tests to use a test element / metadata in order to reduce changes when iterating on Vaadin component metadata in the future

### DIFF
--- a/vaadin-dev-server/src/main/resources/META-INF/frontend/vaadin-dev-tools/theme-editor/editor.ts
+++ b/vaadin-dev-server/src/main/resources/META-INF/frontend/vaadin-dev-tools/theme-editor/editor.ts
@@ -218,8 +218,9 @@ export class ThemeEditor extends LitElement {
       return;
     }
     const { part, property, value } = e.detail;
+    const partName = part?.partName || null;
     this.hasModifications = true;
-    this.editedComponentTheme.updatePropertyValue(part.partName, property.propertyName, value);
+    this.editedComponentTheme.updatePropertyValue(partName, property.propertyName, value);
     this.effectiveComponentTheme = ComponentTheme.combine(
       this.baseComponentTheme,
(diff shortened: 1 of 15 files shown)


vaadin-dev-server/src/main/resources/META-INF/frontend/vaadin-dev-tools/theme-editor/tests/utils.ts
```diff
@@ -0,0 +1,48 @@
+import { ComponentMetadata } from '../metadata/model';
+
+export class TestElement extends HTMLElement {
```
sissbruecker, line 3: Using a plain JS web component here, as the theme detector currently doesn't work with Lit elements. These render asynchronously, while the detector reads styles synchronously directly after creating the element. 

Should not be a problem with Vaadin components for now as they are still Polymer based. Will try to open a separate PR with a fix for this at some point.

vaadin-dev-server/src/main/resources/META-INF/frontend/vaadin-dev-tools/theme-editor/editor.ts
```diff
@@ -218,8 +218,9 @@ export class ThemeEditor extends LitElement {
       return;
     }
     const { part, property, value } = e.detail;
+    const partName = part?.partName || null;
```
sissbruecker, line 221: The `ComponentTheme` class was already implemented so that a `null` part name is interpreted as targeting the host. Adapted the remaining APIs to do the same. Not very explicit, but should be good enough for now. Maybe we can make this more explicit if we add support for other types of selectors at some point.

MarcinVaadin (approved):

PR #16084, opened 2023-03-01
chore: how to debug flow client in your project guide (and some mino other CONTRIBUTING, README md file changes)

czp13:
## Description
The Vaadin Flow team had an internal conversation that it would be good to add a guide to the README.md about how to debug the flow client in your project.

So relevant changes:
- Added guide/steps on how to debug the flow client JS file in your project.
- Also did some minor changes in other README.md files)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-client/README.md
+++ b/flow-client/README.md
@@ -54,7 +54,7 @@ In short debugging is building flow client JS in pretty/detailed mode and then a
 5. Then you can debug the JS code in your browser's dev tools (e.g.sources tab: `VAADIN/generated/jar-resources/FlowClient.js`)
 
 #### Note for debugging:
-   - you shall add the dependency if it is not present.
+   - you shall add the `flow-client` dependency if it is not present.
      - For example, your project may use Vaadin `24.0.0`, but you rebuild the flow client at version `24.1-SNAPSHOT`. 
 In this case, you should probably add the `flow-client` dependency to overwrite the version the project is bringing in. 
    - Delete the dev-bundle (`src/main/dev-bundle`) directory to be sure the project will not use a previously bundled version of the client


flow-client/README.md
```diff
@@ -35,3 +35,19 @@ Build mode, where the frontend files are pre-compiled and bundled into {project.
    - remove dev-bundle in the current folder: `rm -rf src/main/dev-bundle`
    - remove dev-bundle in the sub-folders: find . -type d -name dev-bundle -delete
 
+## Debugging
+
+In short debugging is building flow client JS in pretty/detailed mode and then adding it to your project and then using it for debugging in the browser.
+1. `git clone git@github.com:vaadin/flow.git`
+2. `git checkout <appropriate branch/tag>`
+3. Edit `flow-client/pom.xml` and change `<gwt.module.style>OBF</gwt.module.style> to DETAILED or PRETTY`
+4. Then build the project: `mvn clean install` in the `flow-client` folder.
```
mcollovati, line 44: I would avoid suggest changing the pom file, that may be then committed accidentally.
Better suggest changing the value at build time with a system property, e.g.
`mvn clean install -Dgwt.module.style=DETAILED`

flow-client/README.md
```diff
@@ -35,3 +35,19 @@ Build mode, where the frontend files are pre-compiled and bundled into {project.
...
+   - Do not mind the many ERRORS happening while building the JAR, eventually build can be SUCCESS still,
+   - the built JAR should be at the /target directory e.g. `target/flow-client-23.3-SNAPSHOT.jar`
+5. Make sure your project is using the non-obfuscated `flow-client.jar`:
+   - Editing `pom.xml` or `gradle.build` or any other way,
```
mcollovati, line 48: It is not clear to me if this `Editing ...` statement refers to changing the property described in point 3

flow-client/README.md
```diff
@@ -35,3 +35,19 @@ Build mode, where the frontend files are pre-compiled and bundled into {project.
...
 
+## Debugging
+
+In short debugging is building flow client JS in pretty/detailed mode and then adding it to your project and then using it for debugging in the browser.
+1. `git clone git@github.com:vaadin/flow.git`
+2. `git checkout <appropriate branch/tag>`
+3. Edit `flow-client/pom.xml` and change `<gwt.module.style>OBF</gwt.module.style> to DETAILED or PRETTY`
+4. Then build the project: `mvn clean install` in the `flow-client` folder.
+   - Do not mind the many ERRORS happening while building the JAR, eventually build can be SUCCESS still,
+   - the built JAR should be at the /target directory e.g. `target/flow-client-23.3-SNAPSHOT.jar`
```
mcollovati, line 46: I would not pin the version here, maybe just use X.Y
```suggestion
   - the built JAR should be at the /target directory e.g. `target/flow-client-X.Y-SNAPSHOT.jar`
```

flow-client/README.md
mcollovati: We can perhaps also suggest adding the `-DskipTests` property to the command line, to skip tests

flow-client/README.md
```diff
@@ -35,3 +35,19 @@ Build mode, where the frontend files are pre-compiled and bundled into {project.
...
+5. Make sure your project is using the non-obfuscated `flow-client.jar`:
+   - Editing `pom.xml` or `gradle.build` or any other way,
+   - OR just uncompress the built JAR file and then copy the `FlowClient.js` from it to your proper directory 
+   (where it is served to frontend) in e.g.: `frontend/generated/jar-resources`
+6. Then you can debug the JS code in your browser's dev tools
+   (e.g.sources tab: `VAADIN/generated/jar-resources/FlowClient.js`)
+
```
mcollovati, line 53: I would add a point stating that you should ensure that `flow-client` dependency in an application project points to the built version (eg: be sure it is `X.Y-SNAPSHOT`) or add the dependency if not present.
For example, your project may use Vaadin 24.0.0, but you rebuild the flow client at version `24.1-SNAPSHOT`. In this case, you should probably add the flow-client dependency to overwrite the version the project is bringing in.

In addition, I think it is worth adding a note to recommend deleting the dev-bundle (`src/main/dev-bundle`) directory, as stated in the previous paragraph, but also cleaning the application project (`mvn vaadin:clean-frontend` or `mvn:flow-clean-frontend`) to be sure to use the rebuilt version of the Flow client.

mcollovati (changes requested):

mcollovati:
After following the instruction, I can see readable Flow client sources in the browser, but is seems to me that some kind of optimization is still done.

For example, in flow-client jar I can see this snippet

```javascript
  function computeScriptBase(){
    function getDirectoryOfFile(path){
      var hashIndex = path.lastIndexOf('#');
      if (hashIndex == -1) {
        hashIndex = path.length;
      }
      var queryIndex = path.indexOf('?');
      if (queryIndex == -1) {
        queryIndex = path.length;
      }
      var slashIndex = path.lastIndexOf('/', Math.min(queryIndex, hashIndex));
      return slashIndex >= 0?path.substring(0, slashIndex + 1):'';
    }
```

but in the browser I get

```javascript
        function computeScriptBase() {
            function f(O) {
                var oe = O.lastIndexOf("#");
                oe == -1 && (oe = O.length);
                var ue = O.indexOf("?");
                ue == -1 && (ue = O.length);
                var de = O.lastIndexOf("/", Math.min(ue, oe));
                return de >= 0 ? O.substring(0, de + 1) : ""
            }
```

The code can be debugged, but it is more difficult to understand what is happening without the variable names.
But maybe there's something I missed in my environment.
@czp13 can you confirm that you see the original source code in the browser?

czp13:
> After following the instruction, I can see readable Flow client sources in the browser, but is seems to me that some kind of optimization is still done.

Thank you for the review! In my case it is/was looking like this:
```
  function computeScriptBase(){
    function getDirectoryOfFile(path){
      var hashIndex = path.lastIndexOf('#');
      if (hashIndex == -1) {
        hashIndex = path.length;
      }
      var queryIndex = path.indexOf('?');
      if (queryIndex == -1) {
        queryIndex = path.length;
      }
      var slashIndex = path.lastIndexOf('/', Math.min(queryIndex, hashIndex));
      return slashIndex >= 0?path.substring(0, slashIndex + 1):'';
    }
```

Worth to mention:
- I used the `DETAILED` option (did you use `PRETTY`?)
- I copied the JS file while the application was running. If you did it through the build/maven, then it is possible that some kind of obfuscation is still happening. Maybe there is a switch to turn off that one as well :)

flow-client/README.md
czp13: Thank you. Sure, we can do it, good idea!

flow-client/README.md
czp13: Yes, good points, -Dskiptests will save some time too. I am adding these changes to the PR.

flow-client/README.md
czp13: No, I intended this to relate to your custom project pom.xml, to make sure you use proper versions, but I can delete this as making confusion.

mcollovati: Got it. I would perhaps make it explicit that editing the files is to pin the flow-client version.
Perhaps we can move here the first point of "Note for debugging"

flow-client/README.md
```diff
@@ -35,3 +35,28 @@ Build mode, where the frontend files are pre-compiled and bundled into {project.
...
+     - this option will make sure you use the non-obfuscated JS file (the other option can cause still some optimization)
+5. Then you can debug the JS code in your browser's dev tools (e.g.sources tab: `VAADIN/generated/jar-resources/FlowClient.js`)
+
+#### Note for debugging:
+   - please make sure that `flow-client` dependency in an application project points to the built version (e.g.: be sure it is `X.Y-SNAPSHOT`), 
+   - OR add the dependency if it is not present.
+     - For example, your project may use Vaadin `24.0.0`, but you rebuild the flow client at version `24.1-SNAPSHOT`. 
+In this case, you should probably add the `flow-client` dependency to overwrite the version the project is bringing in. 
+   - also suggested to delete the dev-bundle (`src/main/dev-bundle`) directory
+   - also recommended to clean the application project (`mvn vaadin:clean-frontend` or `mvn:flow-clean-frontend`)
```
mcollovati, line 61:
```suggestion
   - Delete the dev-bundle (`src/main/dev-bundle`) directory to be sure the project will not use a previously bundled version of the client
   - clean the application project (`mvn vaadin:clean-frontend` or `mvn:flow-clean-frontend`) is also recommanded
```

mcollovati (changes requested):

flow-client/README.md
czp13: Oki, I will add/refactor this, and as I see, this is the last bit I need to tackle :).

czp13: Done! Thank you again!

flow-client/README.md
```diff
@@ -35,3 +35,28 @@ Build mode, where the frontend files are pre-compiled and bundled into {project.
...
+4. Make sure your project is using the non-obfuscated `flow-client.jar`:
+   - Please make sure that your build tool config file (`pom.xml` or `gradle.build`) contains the proper information:
+     - `flow-client` dependency in an application project points to the built version (e.g.: be sure it is `X.Y-SNAPSHOT`),
+   - OR just uncompress the built JAR file and then copy the `FlowClient.js` from it to your proper directory 
+(where it is served to frontend) in e.g.: `frontend/generated/jar-resources`
+     - this option will make sure you use the non-obfuscated JS file (the other option can cause still some optimization)
+5. Then you can debug the JS code in your browser's dev tools (e.g.sources tab: `VAADIN/generated/jar-resources/FlowClient.js`)
+
+#### Note for debugging:
+   - you shall add the dependency if it is not present.
```
mcollovati, line 57:
```suggestion
   - you shall add the `flow-client` dependency if it is not present.
```

mcollovati (changes requested):

PR #16091, opened 2023-03-02
chore: Update file upload to latest (9.0), add default limits

tepi:


flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
```diff
@@ -613,9 +618,42 @@ protected Collection<Part> getParts(VaadinRequest request)
     protected FileItemIterator getItemIterator(VaadinRequest request)
             throws FileUploadException, IOException {
         ServletFileUpload upload = new ServletFileUpload();
+        upload.setSizeMax(getSizeMax());
+        upload.setFileSizeMax(getFileSizeMax());
```
knoobie, line 622: I would reject this change. The file limits are way to low for Vaadin applications and there is literally no way for people to change this because the StreamReceiverHandler is used as protected constructor of  StreamRequestHandler which is literally 10 level to deep for a common Vaadin Flow User's understanding of how flow internals work. How should somebody customize this? Especially the file size limits. 

The limit (CVE related) for the amount of files is totally fine (like tomcat did) but adding the other two limits is literally going to create a lot of problems for you once this is released, because people are wondering in production why large files are rejected, even tho they configured e.g. spring boot to allow such large files correctly.

tepi: I agree, we should let the container handle those limits.

mshabarov:
I would propose to a bit simplify the way to override the limits.
So if we do:
1. Move protected getters from `StreamReceiverHandler` to `StreamRequestHandler`
2. Add class fields for limits and setter to `StreamReceiverHandler`
3. In `StreamReceiverHandler`:
```
        JakSrvltFileUpload upload = new JakSrvltFileUpload();
        upload.setSizeMax(requestSizeMax);
        upload.setFileSizeMax(fileSizeMax);
        upload.setFileCountMax(fileCountMax);
```
4. In `StreamRequestHandler`:
```
protected StreamRequestHandler(StreamReceiverHandler receiverHandler) {
        receiverHandler.setRequestSizeMax(getSizeMax());
        receiverHandler.setFileSizeMax(getFileSizeMax());
        receiverHandler.setFileCountMax(getFileCountMax());
        this.receiverHandler = receiverHandler;
    }
```

Then this API will allow to do:

```
public class CustomServiceInitListener implements VaadinServiceInitListener {
    @Override
    public void serviceInit(ServiceInitEvent event) {
        event.addRequestHandler(new CustomStreamRequestHandler());
    }

    private static final class CustomStreamRequestHandler
            extends StreamRequestHandler {
        @Override
        protected long getFileCountMax() {
            return 10000 * 10;
        }
    }
}
```
which is the standard way of adding request handler in Vaadin.

Then also I propose:
1. Catch exceptions when the limit is exceeded and log a warning with short description how to change the limit
2. Document the limit override here https://vaadin.com/docs/latest/components/upload#upload-restrictions

flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
```diff
@@ -259,6 +272,17 @@ private boolean handleMultipartFileUploadFromInputStream(
                 success = success && itemSuccess;
             }
         } catch (FileUploadException e) {
+            String limitInfoStr = "{} limit exceeded. To increase the limit "
```
mshabarov, line 275: I suggest to extend this message a bit and provide the name of method to override as a second parameter in the message:

```
'{}' limit exceeded. To increase the limit extend StreamRequestHandler, override '{}' method and provide a higher limit. 
The extended class needs to be added to request handlers with ServiceInitEvent::addRequestHandler in an extension of VaadinServiceInitListener.
```

flow-server/src/main/java/com/vaadin/flow/server/communication/StreamRequestHandler.java
```diff
@@ -172,6 +178,36 @@ private static Optional<URI> getPathUri(String path) {
         }
     }
 
+    /**
+     * Returns maximum request size for upload. Override this to increase the
+     * default. Defaults to 100 MB.
```
mshabarov, line 183: Correct the defaults in javadocs: should be 10000 file count and no limit for request size and file size, as given for default values.

mshabarov (changes requested):

mshabarov (approved):

knoobie:
@mshabarov FYI: Tomcat reduced the default from 10.000 to 1.000 (https://github.com/apache/tomcat/commit/f6ffe8dba02082b7f1d3fdbdabfc20a54feed62b) this was just mentioned in the new patch notes for Tomcat 11.0.0-M5. 

This was also backported to 9 / 10 a little bit different (https://github.com/apache/tomcat/commit/4867e4e028f3a4ea801bcf8288ddca1a60d75e8b)

PR #16099, opened 2023-03-02
refactor: added support for themeEditorComponentMetadata command

MarcinVaadin:
## Description
- added support for _themeEditorComponentMetadata_ command
- extracted json messages to records
- added support for requestId (required in each message)
- added response to each request

Added hardcoded value of requestId to vaadin-dev-tools.ts to not to block further work.

## Responses
From this PR each request is followed by response with same requestId, there are currently:
- themeEditorResponse - generic response, with `{ requestId: 'abc' [...] }` and data dependent from operation
- themeEditorError - `{ requestId: 'abc', error: 'Something went wrong' }`

## Component metadata
To get component metadata, send:
```
{ requestId: 'abc', nodeId: 0, uiId: 0}
```

and response:
_themeEditorResponse_ `{ requestId: 'abc', accessible: true/false }`


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -243,9 +247,18 @@ public void onMessage(AtmosphereResource resource, String message) {
...
+            try {
+                JsonObject resultData = themeEditorMessageHandler
+                        .handleDebugMessageData(command, data);
+                send(resource, GenericResponse.COMMAND_NAME, resultData);
+            } catch (ThemeEditorException ex) {
+                getLogger().error(ex.getMessage(), ex);
+                ErrorResponse response = new ErrorResponse(
+                        data.getString("requestId"), ex.getMessage());
+                send(resource, ErrorResponse.COMMAND_NAME,
+                        JsonUtils.beanToJson(response));
```
sissbruecker, line 260: This part is not tested, but might be quite relevant to ensure that errors also have a request id.

Maybe we could move the error handling into `ThemeEditorMessageHandler.handleDebugMessageData` somehow if that would make it easier to test.

Related to that, I'm also not sure about splitting responses and errors into separate commands. The alternative would be a response `code` on all response messages, that could be `ok` or `error`. That would incidently solve the problem above, as we could build all types of responses in `ThemeEditorMessageHandler` since we don't need to distinguish the command type anymore.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifier.java
```diff
@@ -129,6 +143,46 @@ public void removeClassNames(Integer uiId, Integer nodeId,
         });
     }
 
+    /**
+     * Returns metadata for picked component. Supports class names and themes.
```
sissbruecker, line 147: > Supports class names and themes.

It doesn't do that yet 🙂. The exception message below is also confusing and could just be something generic about not being able to generate metadata for now.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/ThemeEditorMessageHandler.java
```diff
@@ -32,76 +38,80 @@ public String getState() {
...
+        ComponentMetadataResponse response = new ComponentMetadataResponse(
+                request.requestId(), metadata.isAccessible());
+        return JsonUtils.beanToJson(response);
     }
 
-    private static Logger getLogger() {
-        return LoggerFactory
-                .getLogger(ThemeEditorMessageHandler.class.getName());
+    private JsonObject ok(String requestId) {
+        return JsonUtils.beanToJson(new GenericResponse(requestId));
```
sissbruecker, line 114: It looks like this generates some wrapper around the response object:
```json
{"command":"themeEditorResponse","data":{"type":"OBJECT","object":{"requestId":"abc-123-def-456"}}}
```
Same goes for the error response I guess.

MarcinVaadin: Fixed

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/JavaSourceModifier.java
MarcinVaadin: Updated

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
MarcinVaadin: Tests added, added 'code' to response

flow-server/src/main/java/com/vaadin/flow/internal/JsonUtils.java
```diff
@@ -53,6 +53,8 @@ public final class JsonUtils {
 
     private static final String CANNOT_CONVERT_NULL_TO_A_JSON_OBJECT = "Cannot convert null to JSON";
 
+    private static final String CANNOT_CONVERT_NULL_JSON_TO_OBJECT = "Cannot convert null JSON object to Java object";
```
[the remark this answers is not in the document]
MarcinVaadin: Updated

sissbruecker (approved):

PR #16103, opened 2023-03-02
feat: Add app vs default dev-bundle statistics

mshabarov:


manolo (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -111,9 +115,20 @@ public NodeTasks(Options options) {
                         frontendDependencies, classFinder)) {
                     options.withRunNpmInstall(true);
                     options.withCopyTemplates(true);
+                    UsageStatistics.markAsUsed("flow/expressBuild/app-bundle",
```
Artur-, line 118: Do we need the `expressBuild` prefix here or can we just use `dev-bundle` and e.g. `app-dev-bundle`?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -182,21 +183,25 @@ protected static boolean needsBuildInternal(Options options,
                 .getPackages();
 
         if (!hashAndBundleModulesEqual(statsJson, packageJson, npmPackages)) {
+            UsageStatistics.markAsUsed("flow/expressBuild/npm-package", null);
```
Artur-, line 186: Are these the reasons for using an app-bundle? If so, the identifier could tell that

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
mshabarov: We can use `default-dev-bundle` and `app-dev-bundle`.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov: the reason is "missing-package". I'll rename it.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
mshabarov: Or maybe just `dev-bundle` and `app-dev-bundle` are enough

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
Artur-: Not just this but all in this file. Like ”rebundle-reason-x”

manolo (approved):

PR #16113, opened 2023-03-03
fix: Remove words about adding bundle dependency

mshabarov:
Because default dev bundle should be added automatically always.


czp13 (approved):
👍

PR #16093, opened 2023-03-02
test: Run with web-test-runner instead of intern

Artur-:

PR #16117, opened 2023-03-03
test: Run with web-test-runner instead of intern (#16093) (CP: 23.3)

Artur-:
Use web-test-runner rather then intern, use a xhr mock that is esm compatible.
This should fix the 23.3 branch from failing build tests with this error:

> Tunnel started
> BUG: suiteEnd was received for invalid session 
> (ノಠ益ಠ)ノ彡┻━┻
> Error: connect ECONNREFUSED ::1:4444


czp13 (approved):
I saw only one line difference between this change and the change that was merged yesterday to `main`:
- https://github.com/vaadin/flow/pull/16093/files

This had in `flow-client/webpack.tests.config.js` the `devtools` added before "./src/test/frontend/vaadin-dev-tools-tests.js" as well in the config and was removed, but I would think this shall not cause any problem as this is a deletion not addition (`main` probably is not using this dependency some time ago)

```Typescript
module.exports = {
  mode: "development",
  entry: {
    flow: "./src/test/frontend/FlowTests.ts",
    // this is used in v23.3 and was removed, while main was not using this, so it was not removed:
    ++ devtools: "./src/test/frontend/vaadin-dev-tools-tests.js", 
  },
```

This was the only one-line difference:
+10,696 −342  vs +10,696 −341

mshabarov (approved):

Artur-:
indeed, vaadin-dev-tools.ts was moved in `main` some time ago to `vaadin-dev-server`

PR #16119, opened 2023-03-03
test: Run with web-test-runner instead of intern (#16093) (CP: 23.2)

Artur-:
Use a xhr mock that is esm compatible


czp13 (approved):
LGTM, same comment as for:
- https://github.com/vaadin/flow/pull/16117

PR #16120, opened 2023-03-03
test: Run with web-test-runner instead of intern (#16093) (CP: 9.0)

Artur-:
Use a xhr mock that is esm compatible


Artur-:
This needs a stamp for the 9.0 branch build to work again

mcollovati (approved):

PR #16123, opened 2023-03-03
refactor: implement theme editor API client

sissbruecker:
## Description

Add an API client to the theme editor that wraps the websocket connection, can send messages and correlate responses to the original request. Each request returns a `Promise` that resolves when a successful response is received, and rejects when an error response is received.

Only covers sending rule updates, we can cover component metadata when adding a related feature.


MarcinVaadin (approved):
LGTM, it looks like requestId could be int.

PR #16124, opened 2023-03-03
feat: Adds helper methods for commonly used CSS properties

mstahv:
The current Style API is basically a dummy string-string map. These helper methods makes finding the right property value easier for Java users and prevents making typos in the property names.

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,424 @@ public interface Style extends Serializable {
...
+    /**
+     * Sets the <code>position</code> property.
+     *
+     * @param value
+     *            the style property value (if <code>null</code>, the property
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setPosition(String value) {
+        set("positio ", value);
```
knoobie, line 330: You mean this kind of typo or the below with scael? 😅

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,424 @@ public interface Style extends Serializable {
...
+     * Sets the <code>background</code> property.
+     *
+     * @see https://drafts.csswg.org/css-backgrounds-3/#propdef-background
+     *
+     * @param value
+     *            the style property value (if <code>null</code>, the property
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setBackground(String value) {
```
knoobie, line 111: Instead of providing so many methods and cluttering the interface, what do you think about overloading the current methods with an enum, like set(Styles.BACKGROUND, "blue")? This could also be used for the has or get methods to make it all type safe. This should also be backwards compatible with the default string methods.

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
mstahv: I'm quite proud of myself it that was the only one 😎

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
mstahv: I considered that, but IMO

 * DX (autocomplete in IDE) is as good
 * Not compatible with further improvements (like enums for certain APIs)
 
I don't see a problem with "cluttering", as long as we don't add a dozen of different variation for e.g. border.

mstahv: Added example of "further api improvements" in https://github.com/vaadin/flow/pull/16124/commits/4fcfce677365fa2406915fa07d9b20bd0704fae0

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
knoobie: Does the set method automatically trim spaces? Otherwise there is still one 🤫

mstahv: I don't think so, but I believe browser would. Fixed. Let me know if you still find some, there sure could be some.

First thought I'd autogenerate based on some specification document, but then thought I have better things to do 😎

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
knoobie: Autocomplete in IDEs is a good call! I was always wondering: is this Style API on the server even recommended? I always thought it was the last resort and should only be used sparingly. Using the theme should have always been superior to this. (correct me if I have wrong assumptions) 

Meaning, if we make it even more accessible / easy to use for newcomers with IDE autocomplete.. I've the feeling that the usage could spike and they are going to use this everywhere..resulting in Bad Performance or Other inconvenience.

mstahv: It should for sure be the default to use css files (~ theme) for most cases. But 

 * there are places where changing some css properties is more like functional feature of a component, instead of a theme feature. There I think using inline styling is just fine approach (compared to using e.g. different classes and then toggling those in Java code)
 * We already have the Style API and people are using it anyways, so don't know why it should be more error prone
 
But in general, we probably should document best practices in the Style class and/or in reference documentation. Regardless of this PR. Didn't check what is the current situation...

knoobie: Good call! Thanks for that clarification 🙇‍♂️

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,501 @@ public interface Style extends Serializable {
...
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setClear(Clear value) {
+        if (value == null) {
+            remove("clear");
+        } else {
+            set("clear", javaEnumToCssValue(value));
+        }
+        return this;
```
knoobie, line 182: Totally missed that Java 17 finally allows private Interface methods! Thanks for enlightening! 

What do you think about proper naming for the methods to show their chaining capabilities? Like `withBackground().withFont()` and so on. 

Another idea for the private method (coding on mobile su***): 

```java

private <T extends Enum<T>> void applyOrErase(String name, @Nullable T value)
{
    if (value == null) {
             remove(name);
         } else {
             set(name, value.name().replace("_", "-").toLowerCase());
         }
}

```

mstahv: Very good my friend! Cleaned up the repeated null checks into single method.

If we'd start from scratch, I'd go with `with` for fluent API, but as the set method already does it this way (and couple of other places in the framework), I think we should stick with the same approach here.

mstahv:
I assume we are pretty much ready with this PR because @knoobie has not commented anything for the last couple of days 😎

knoobie:
Yes that's always a good sign 😏

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,454 @@ public interface Style extends Serializable {
...
+     *
+     * @see https://drafts.csswg.org/css-backgrounds-3/#propdef-background
+     *
+     * @param value
+     *            the style property value (if <code>null</code>, the property
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setBackground(String value) {
+        return set("background", value);
```
knoobie, line 112: nit: not sure how much you guys are using https://github.com/vaadin/flow/blob/554dd02710b5280e582264ebedb92425af854121/flow-server/src/main/java/com/vaadin/flow/dom/ElementConstants.java - the added style "magic" strings could come from there, not sure how much value this adds, but it would help with consistency.

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,454 @@ public interface Style extends Serializable {
...
+     *
+     * @see https://drafts.csswg.org/css-backgrounds-3/#propdef-display
+     *
+     * @param value
+     *            the style property value (if <code>null</code>, the property
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setDisplay(Display value) {
+        return applyOrErase("display", value);
```
knoobie, line 217: The display enum is a good example for: damn those are a lot of possible values that probably 0.05% of the user need or even know. I'm thinking about three alternatives: 

- don't change it, keep it as it is with the burdon to maintain this values 
- only add common values to reduce auto complete overkill for DX and add a second overload method with a string value 
- (Not so good DX) creating of a functional CssInterface / marker for the enums and use that as param allowing people use DisplayEnum as such (auto complete is degraded compared to Display as param) but also allowing them to add () -> return "totally other value"

I'm personally leaning towards 1 or 2, but 3 sounds interesting on paper..

knoobie (dismissed):

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
mstahv: I spotted those constants at least in two places (already before this change) and decided to go with this approach. I don't personally find those helpful.

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
mstahv: As discussed in the description, I find this the best compromise between effort & DX improvement. The enum values for enums in this PR are taken form the excellent w3schools.com site, I'd believe those to be rather complete. If something is really missing, it is likely very rare and there is still the current set(String,String) API available.

mstahv: Ah. So the list would be too complete. I wouldn't be worried about that. I would be worried if some of our user who is using this class don't have any idea what they are looking for. This API should be only used by users who know at least something about html, dom, css when building some lower level integrations.

mstahv: Yeah, I hope they read it, because I really like your valuable insights. Sincerely, I think you should have actual rubber stamping rights 🙇‍♂️

flow-server/src/main/java/com/vaadin/flow/router/InternalServerError.java
```diff
@@ -137,10 +137,9 @@ private void checkLogBinding() {
                             + "As a result the logger doesn't do any real logging. "
                             + "Add some binding as a dependency to your project. "
                             + "See details ");
-            logInfo.getStyle().set("marginTop", "10px");
-            logInfo.getStyle().set("marginBottom", "10px");
+            logInfo.getStyle().setMargin("10px 0");
```
czp13, line 140: this could be, right?:
```suggestion
            logInfo.getStyle().setMargin("10px 10px");
```

mstahv: Why would we want to change the implementation? Then there would be margin also on left and right side of the logInfo element?

czp13: I missed/missee this implementation, this is good like this.
First (before coffee) I thought we changed it to `marginTop: 10px, marginBottom: 0` with your code change, but then looked better, and indeed, this is the correct form this will be the same as before.

mstahv: Where can I send more coffee to you 😁

czp13: Maybe it is a good value proposition, just give miss review comments and then a buy me a coffee link after they are annoyed and want you to stop :))). Smart! ;)

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,429 @@ public interface Style extends Serializable {
...
+
+    /**
+     * Sets the <code>box-sizing</code> property.
+     *
+     * @param value
+     *            the style property value (if <code>null</code>, the property
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setBoxSizing(String value) {
```
tepi, line 133: box-sizing only has a few possible values, how about an enum for it?

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,429 @@ public interface Style extends Serializable {
...
+
+    /**
+     * Sets the <code>overflow</code> property.
+     *
+     * @param value
+     *            the style property value (if <code>null</code>, the property
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setOverflow(String value) {
```
tepi, line 296: Same as for box-sizing: maybe an enmu for overflow values could be added?

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
mstahv: Well spotted, fixed!

tepi (dismissed):

tepi (approved):

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,437 @@ public interface Style extends Serializable {
...
+     *            the style property value (if <code>null</code>, the property
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setOverflow(Overflow value) {
+        return applyOrErase("overflow", value);
+    }
+
+    /**
+     * Sets the <code>padding</code> property. * @param value the style property
```
Artur-, line 309:
```suggestion
     * Sets the <code>padding</code> property. 
```

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,437 @@ public interface Style extends Serializable {
...
+     * @param value
+     *            the style property value (if <code>null</code>, the property
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setDisplay(Display value) {
+        return applyOrErase("display", value);
+    }
+
+    // PostFixed with "Css" to avoid collician with java.lang.Float
```
Artur-, line 212:
```suggestion
    // PostFixed with "Css" to avoid a collision with java.lang.Float
```

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,437 @@ public interface Style extends Serializable {
...
+     * @param value
+     *            the style property value (if <code>null</code>, the property
+     *            will be removed)
+     * @return this style instance
+     */
+    default Style setBorder(String value) {
+        return set("border", value);
+    }
+
+    public enum BoxSizing {
```
Artur-, line 125: There are already 2 BoxSizing classes.. Any other options?

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
```diff
@@ -97,4 +98,437 @@ public interface Style extends Serializable {
...
+     */
+    default Style setWidth(String value) {
+        return set("width", value);
+    }
+
+    /**
+     * Css values for the white-space property.
+     */
+    public enum WhiteSpace {
+        NORMAL, NOWRAP, PRE, PRE_LINE, PRE_WRAP, INITIAL, INHERIT
```
Artur-, line 442: This is the same as com.vaadin.flow.component.HasText.WhiteSpace then?

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
mstahv: I don't think the two earlier ones are not really good this. They could probably be removed in the future and start to use these constants.

Artur-: In fact many of the classes here collide with the names used for Lumo utility classes. Can we avoid making using the utility classes more difficult? They are used way more than inline styles

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
mstahv: Same here as in the other duplicate css enum comment. The fact that those are currently found "all over the place" kind of highlights that this should have been done earlier.

flow-server/src/main/java/com/vaadin/flow/dom/Style.java
mstahv: I don't see how this makes using utility classes more difficult.

I also wouldn't be so sure that utility classes are used more than inline classes, except in our own templates created by Jonte ;-)

tepi (approved):

mstahv:
Yeah, I agree those could be added as well. But I don't think those are that often needed. Could do some research within my projects at some point. Check in Flow project itself could give some insights as well.

mshabarov:
Yes, I agree that the usage of hasA() and getA() are way less than setA(). For consistency of API we might consider adding them in V24.1.

mstahv:
If only we could ask ChatGPT to code those for us, we could have all the features🤔

knoobie:
Don't trick me in reviewing bot generated code 😝

Dudeplayz:
Hi, the Jdocs for the `Display` Enum are missing.

Issue #16109, opened 2023-03-03
V24: error during dev-bundle build in add-on project with css file in src/main/resources


taefi:
### Description of the bug

For Vaadin add-on components, it is normal to have the styles and other resources in `src/main/resources/META-INF/resources/frontend` instead of the default `[project root]/frontend` folder, and referencing them by `@CssImport` on the component class.

Migrating an add-on project to v24 and running it (e.g. by `mvn jetty:run`), it tries to build the `dev-bundle` and it complains about not having the referenced resource in the `/frontend` folder:
> [main] INFO com.vaadin.flow.server.frontend.scanner.FullDependenciesScanner - Visited 92 classes. Took 38 ms.
[main] INFO com.vaadin.flow.server.frontend.TaskRunDevBundleBuild - Checking if an express mode bundle build is needed
[main] INFO com.vaadin.flow.server.frontend.TaskRunDevBundleBuild$1 - Added 5 default dependencies to main package.json
[main] ERROR com.vaadin.flow.server.frontend.TaskRunDevBundleBuild - Error when checking if an express mode bundle build is needed
java.lang.IllegalStateException: 
  Failed to find the following css files in the `node_modules` or `/Users/soroosh/tmp/toggle-button-group/frontend` directory tree:
      - ./addons-styles/toggle-button-group.css
  Check that they exist or are installed. If you use a custom directory for your resource files instead of the default `frontend` folder then make sure it's correctly configured (e.g. set 'vaadin.frontend.frontend.folder' property)
	at com.vaadin.flow.server.frontend.AbstractUpdateImports.getCssLines(AbstractUpdateImports.java:230)
	at com.vaadin.flow.server.frontend.AbstractUpdateImports.run(AbstractUpdateImports.java:101)
	at com.vaadin.flow.server.frontend.GenerateMainImports.run(GenerateMainImports.java:53)
	at com.vaadin.flow.server.frontend.TaskRunDevBundleBuild.frontendImportsFound(TaskRunDevBundleBuild.java:343)
	at com.vaadin.flow.server.frontend.TaskRunDevBundleBuild.needsBuildInternal(TaskRunDevBundleBuild.java:152)
	at com.vaadin.flow.server.frontend.TaskRunDevBundleBuild.needsBuild(TaskRunDevBundleBuild.java:105)
	at com.vaadin.flow.server.frontend.NodeTasks.<init>(NodeTasks.java:110)
[...]

PR #16125, opened 2023-03-03
fix: Exclude Vite specific import parameters for project's file imports

mshabarov:
Excludes parameters added by Vite (e.g. ?inline) for frontendFiles entries in the stats.json of dev-bundle.
These parameters cause a mistake in import string comparison and false re-bundling.

Related-to https://github.com/vaadin/flow/issues/16109
Follow-up for https://github.com/vaadin/flow/pull/16039


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -225,7 +225,8 @@ function statsExtracterPlugin(): PluginOption {
       const generatedImports = readFileSync(path.resolve(generatedFlowImportsFolder, "generated-flow-imports.js"), {encoding: 'utf-8'})
           .split("\n")
           .filter((line: string) => line.startsWith("import"))
-          .map((line: string) => line.substring(line.indexOf("'")+1, line.lastIndexOf("'")));
+          .map((line: string) => line.substring(line.indexOf("'")+1, line.lastIndexOf("'")))
```
czp13, line 228: smallie :), but this line could be reformated like:
```suggestion
          .map((line: string) => line.substring(line.indexOf("'") + 1, line.lastIndexOf("'")))
```

czp13 (commented):
As I see we have new tests for the CSS bundle part, but for the other changes about the modules how do we handle the `?` in the stream (was refactored and slightly changed to happen in one step before then it happened). I do not see tests (maybe I missed them):
`flow-server/src/main/resources/vite.generated.ts`. 

If it is possible to add, then maybe would be good to cover the changes that happened in the `vite.generated.ts`.

flow-server/src/main/resources/vite.generated.ts
mshabarov, line 228:
```suggestion
          .map((line: string) => line.substring(line.indexOf("'") + 1, line.lastIndexOf("'")))
```

czp13 (approved):
Thank you very much for covering my question. It is making sense, we do not win big with this refactor, mock, but can lose a lot of time, I agree! I am approving from my side.

PR #16128, opened 2023-03-04
test: Use built-in Selenium Manager instead of WebDriverManager

Artur-:


flow-tests/test-pwa/pom-production.xml
```diff
@@ -20,6 +20,12 @@
             <artifactId>flow-test-resources</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.brotli</groupId>
+            <artifactId>dec</artifactId>
```
mshabarov, line 25: What is this for?

Artur-: It is used in the ITs

mshabarov (approved):

PR #16147, opened 2023-03-06
test: Increase build timeout to 30min

Artur-:
Should make it easier to see what goes wrong in the builds that cause them to take longer


mshabarov (approved):

czp13 (approved):
👍

PR #16153, opened 2023-03-06
chore: Align Flow test dev bundle with the default dev bundle in platform

mshabarov:
Makes dev-bundle in `test-dev-bundle` and in `vaadin-dev-bundle` be in sync regarding dependencies they are inlcuded.
Test module is removed, because it checked only that line-awesome resource is in default bundle, which is not the case now.
Also some polymer dependencies are removed from `test-dev-bundle`.


flow-tests/test-express-build/pom.xml
```diff
@@ -35,7 +35,6 @@
             <modules>
                 <module>test-dev-bundle-no-plugin</module>
                 <module>test-theme-dev-bundle</module>
-                <module>test-theme-default-dev-bundle</module>
```
Artur-, line 38: What did this test and do we still want to test something regarding a customized theme.json?

Artur- (approved):

PR #16157, opened 2023-03-06
chore: Add package and package lock files in tests to VCS

mshabarov:
Updates web-components to `24.1.0-alpha1` and adds package.json and package-lock.json to Git to make Flow tests run a bit faster (when installing npm packages).


mshabarov:
Doesn't really improve the situation.

PR #16158, opened 2023-03-06
refactor: add range property editor for theme editor

sissbruecker:
## Description

Adds a custom property editor for selecting a preset from a range of values, or entering a custom value.

Bildschirmfoto 2023-03-06 um 17 38 00


jouni:
I’m not sure why I’m not seeing the slider control when I try this locally. I cloned this branch, and ran `mvn clean install -DskipTests`, and then used `24.1-SNAPSHOT` in a local starter project.

PR #16160, opened 2023-03-06
refactor: Reuse already fetched I18NProvider in Component.getTranslation

Dudeplayz:
## Description

The I18NProvider should be stored in a local reference, because it will always be present for projects were it is set. This results in doubled access to the backing TheadLocal, which can reduce performance when translation is heavily used.

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


Dudeplayz:
I also added a commit to refactor `getLocale` and `getI18NProvider` to use Optional. This should increase readability. Perhaps the commit could be split. Also it is possible to decide that the runtime speed is prioritized over the readability in this case.

Dudeplayz:
I'm also thinking about to create a private getLocale function, which gets a supplier for I18NProvider. This is then used by the  normal function and the getTranslation functions without a locale param to reduce the I18NProvider resolution for this case.

Dudeplayz:
The usage of the wrapping `Supplier` around the Optional could be removed. The reason for adding this is, that the protected getLocale function would always fetch the I18NProvider, even if it is not needed. But if this performance improvement is worth the decreased readability is up to you.

czp13:
Hey @Dudeplayz,

Thank you very much for your contribution! 🙇  
And nice catch, indeed with your code change in the Component class.

The `getI18NProvider()` was executed twice many times before, but with your optimization, it is only once, and it can make a difference when this method is called many times. I agree with you.

The optional, and code optimization-related changes seem safe enough in the `getTranslation(...)`, and `getI18NProvider()` methods

The `protected Locale getLocale() {`  related changes with my eyes and testing some stream API/Optional API way of working also seems safe.

I pushed the builds as well in our server and except for one job all seem good ✅ 

Two topics remained:
- formatting the code,
- testing,

**Formatting:**
**Could you please execute this command in your root?**
`mvn formatter:format`
This maven command shall format your changes to be consistent with our format standards, and then commit/push it as well please.

**Testing:**
Were you able to test the changes in your custom app, to make sure all is good? (sometimes things can happen we do not expect or think). Plus for the same reason, I feel some test coverage would be helpful, also I did not find any tests covering getTranslation and other affected methods here (from before)

I understand hard to write integration tests and mocking as well in that context :(.

But I would be much more comfortable merge this if we could write some covering tests for the 4 cases basically here:
`getLocale()`

You can use maybe reflection to call this, and potentially Mockito as well to mock most of the objects, because we would just test that the logic is good, do not need to care about injections, etc.

If too much hassle then let me know and maybe we can revert those changes back, as it is more about readability there than performance, but for the getTranslation(), we can merge it for sure after formatting the code.

czp13 (changes requested):
Seems good, but we need to clear up two things before merging:
- formatting (to pass CI builds)
- and potential covering test cases for the `getLocale` changes (although it seems safe from eyeball check)

Thank you @Dudeplayz for the work so far! 🙇

Dudeplayz:
Hi, I was in vacation.
I've run the formatter.
I'm not sure how to test the getLocale method, which is accessing the VaadinService, I am not that experienced with writing tests. It would be nice if you could help or give some additional hints how to do it.
Thanks for the review until now 😊.

czp13:
Hi @Dudeplayz ,

Thanks for getting back to me and for the follow-up commit ˆˆ. 
With your formatting changes/commits the tests and so build should pass ;).
(I re-executed them now).

About writing tests, yap we need to mock the objects we are using in the method, basically two things:
- Vaadinservice for giving back Vaadinservice.getCurrent().getInstantiator().getI18NProvider());
- and UI object for giving back UI.getCurrent().getLocale();

We are lucky because there is already a working example in ComponentTest class:
```
    @Test
    public void getComponentLocale_noCurrentUI_returnsDefaultLocale() {
        UI.setCurrent(null);
        Instantiator instantiator = mocks.getService().getInstantiator();
        Mockito.when(instantiator.getI18NProvider()).thenReturn(null);
        Component test = new TestButton();
        final Locale locale = test.getLocale();
        Assert.assertEquals("System default locale should be returned",
                Locale.getDefault(), locale);
    }
```

Which is covering one branch of the 4 branches. This execution branch is covered now:
```         ...
            final I18NProvider i18NProvider = getI18NProvider();
            // If a i18nProvider is not defined we should just return the
            // default locale.
            if (i18NProvider == null) {
                return Locale.getDefault();
            }
            ...
```

If you feel the power 💪 , based on this example you can try to add the remaining cases, or potentially we merge and then I add them in a separate PR, or even better if I can commit to your project, because it is a forked project, I need to check out your repo, branch and commit there I believe :). Let me know how to proceed.

czp13:
I helped/got more proactive a bit more. I hope you do not mind and added the mentioned tests here:
- https://github.com/vaadin/flow/pull/16367
As soon as I can merge this to the main, then we can merge it here and finally squash it :).
I keep you in the loop.

czp13:
I merged the tests to master and now merging them here. If this is good, then we are good to go with this! 👍

czp13 (approved):

Dudeplayz:
Hi @czp13, sorry for the very late reply, I didn't saw the notification and needed some time to get back after vacation 😄.

Thanks for the clarification, it helped me to understand how to do such things. And thanks for writing and adding the tests, I really appreciate it!

I'm glad my first PR was merged 😁.

p.s. I had a very nice vacation, thanks. Hope yours was fine, too. 😊

Dudeplayz:
It was a pleasure 😊.
I'm more happy to hear that my changes get back-ported!
I hope too, I already have some ideas ;D

czp13:
Thank you 😊. I am glad to help, and it was I think a nice/providing value change :).
Hehe, look forward to seeing your ideas ;)!

PR #16161, opened 2023-03-06
chore: Add package and package lock files in tests to VCS

mshabarov:
To make Flow tests run a bit faster (when installing npm packages).


mshabarov:
Doesn't really improve the situation.

PR #16173, opened 2023-03-07
test: Run with web-test-runner instead of intern (#16093) (CP: 9.0)

tepi:


mshabarov (approved):

tepi:
> Ah, I overlooked this one #16120, which does the same. Number of changed lines are different

Right, let's go with that one then. I'll drop this one.

PR #16189, opened 2023-03-07
test: Use built-in Selenium Manager instead of WebDriverManager (#16128) (CP: 23.2)

Artur-:


flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeDriverLocator.java
```diff
@@ -18,8 +18,6 @@
 
 import java.io.File;
 
-import io.github.bonigarcia.wdm.WebDriverManager;
```
czp13, line 21: This could be removed as well, and then the build errors will be ceased to exist 🤞 .
```
    /**
     * Fills {@link ChromeDriverLocator#WEBDRIVER_CHROME_DRIVER} system property
     * with chromedriver path, does not override already existing value.
     */
    static void fillEnvironmentProperty() {
        if (AbstractTestBenchTest.USE_HUB) {
            return;
        }
        String chromedriverProperty = System
                .getProperty(WEBDRIVER_CHROME_DRIVER);
        if (chromedriverProperty == null
                || !new File(chromedriverProperty).exists()) {
            // This sets the same property
            WebDriverManager.chromedriver().setup();
        } else {
            System.out
                    .println("Using chromedriver from " + chromedriverProperty);
        }
    }
```

czp13: And in the other PR-s too I think these are blocking the build/compilation success.

czp13 (approved):

PR #16190, opened 2023-03-07
test: Use built-in Selenium Manager instead of WebDriverManager (#16128) (CP: 23.3)

Artur-:


Artur-:
Let's see that this actually works when not manually installing the chromedriver before merging it: https://github.com/vaadin/flow/pull/16201

Artur-:
Works fine

PR #16191, opened 2023-03-07
test: Use built-in Selenium Manager instead of WebDriverManager (#16128) (CP: 9.0)

Artur-:


Artur-:
This probably won't work really before Selenium is updated (through a new TestBench)

PR #16207, opened 2023-03-08
feat: support ES-module applications

cromoteca:
As part of fixing https://github.com/vaadin/react-components/issues/64, we want to offer Hilla developers the possibility to switch to ES modules by adding `"type": "module"` to `package.json`.

This PR makes the necessary changes to make sure that existing code will work in such applications.

Related to https://github.com/vaadin/hilla/pull/817.


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -380,7 +380,7 @@ function vaadinBundlesPlugin(): PluginOption {
       if (command !== 'serve') return false;
 
       try {
-        const vaadinBundleJsonPath = require.resolve('@vaadin/bundles/vaadin-bundle.json');
+        const vaadinBundleJsonPath = path.resolve(modulesDirectory, '@vaadin/bundles/vaadin-bundle.json');
```
Artur-, line 383: `path.resolve` and ´require.resolve` are not the same, but will it work anyway?

cromoteca: In this case, the path corresponds to the module name, so I can't see any issue.

cromoteca: Finally I had to amend the syntax: it worked in final apps, but some tests were resolving elsewhere, so I guess that syntax is needed for express build or some other feature.

Artur- (approved):

platosha (approved):

PR #16212, opened 2023-03-09
refactor: add build for dev tools frontend

sissbruecker:
## Description

Adds a build step for the dev tools frontend. This would allow bundling 3rd-party dependencies from NPM into the dev tools without having to add those to an application's `package.json`, or having to add the dependencies sources to the dev tools frontend.

Changes:
- Moves frontend files to `/frontend`
- Adds a Vite build that bundles the dev tools sources and writes the output to `src/main/resources/META-INF/frontend/vaadin-dev-tools` where it is picked up by Maven for packaging
- Adds a Maven plugin for running the frontend build as part of the Maven build


vaadin-dev-server/pom.xml
```diff
@@ -89,16 +89,42 @@
...
-                    <!-- Exclude all frontend tests and utils -->
-                    <exclude>META-INF/frontend/vaadin-dev-tools/**/*.test.ts</exclude>
-                    <exclude>META-INF/frontend/vaadin-dev-tools/theme-editor/tests/*.*</exclude>
-                </excludes>
-            </resource>
-        </resources>
+        <plugins>
+            <plugin>
+                <groupId>com.github.eirslett</groupId>
+                <artifactId>frontend-maven-plugin</artifactId>
```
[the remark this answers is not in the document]
Artur-: I think that `flow-client/pom.xml` runs npm install also. Should probably do it the same way here

vaadin-dev-server/vite.config.js
```diff
@@ -0,0 +1,26 @@
+import { fileURLToPath } from 'url';
+import { defineConfig } from 'vite';
+
+export default defineConfig({
+  build: {
+    // Write output to resources to include it in Maven package
+    outDir: 'src/main/resources/META-INF/frontend/vaadin-dev-tools',
```
Artur-, line 7: Do I understand it correctly that this will bundle everything into `vaadin-dev-tools.js`? 
What happens if dev tools uses a web component, like vanilla colorful, that you also use in the application?

sissbruecker: > Do I understand it correctly that this will bundle everything into vaadin-dev-tools.js?

Yes, everything that is imported and does not only reference a type. Also `lit` and `@vaadin` related imports are kept external as they are expected to be available through the app, and both don't like being loaded in different versions.

> What happens if dev tools uses a web component, like vanilla colorful, that you also use in the application?

It's definitely possible to end up with duplicate modules being loaded and whatever the side-effects of that are. For vanilla-colorful specifically there is [an option](https://github.com/web-padawan/vanilla-colorful#base-classes) to import the component class without automatically defining it as custom element, so that would allow to define a dev-tools specific element name.

Artur-: Sounds like it would work reasonably well to do this then. By shipping a bundle instead of the TS sources, we at the same time resolve issues where vaadin-dev-tools doesn't follow the strictest possible tsconfig settings (#16237).

What we should avoid though is using anything in vaadin-dev-tools that has global side effects. So using vanilla-colorful is fine but we should rename the tag to a dev tools specific one. Then we minimize the risk of causing issues in an application because of what we are doing in development tools.

Artur- (dismissed):

vaadin-dev-server/vite.config.js
sissbruecker: > What we should avoid though is using anything in vaadin-dev-tools that has global side effects.

Yes, this needs to be considered on a case by case basis when adding a new dependency. 

I'll remove the vanilla colorful copy, and add the distinct element name in a separate PR.

MarcinVaadin (approved):

PR #16216, opened 2023-03-09
fix: Add 'remote allow origins' option to chrome driver

mshabarov:
Fixes status code=403 errors when running tests with chrome driver described here https://stackoverflow.com/questions/75678572/java-io-ioexception-invalid-status-code-403-text-forbidden


Artur-:
Let's not, let's upgrade TestBench instead

PR #16218, opened 2023-03-09
fix: Apply parent theme located in project frontend folder

mshabarov:
Makes styles from parent theme apply properly if the parent theme is located in the project's folder, not in the JAR.
Re-bundles if the parent theme in frontend folder has new entries in theme.json.
Adds theme.json content to stats.json if parent theme is in the frontend folder.
Adds link tag to the styles.css of the parent theme onto the page.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -1684,7 +1684,17 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
             }
         }
 
-        // Secondly, add a link for the project's custom theme, if it exists
+        // Secondly, check if a parent theme is in the project's frontend/themes
+        // folder
+        Optional<String> parentThemeName = FrontendUtils
+                .getParentThemeNameInFrontend(frontendFolder, themeName.get());
+        if (parentThemeName.isPresent()) {
+            T reference = referenceProvider.apply(parentThemeName.get(),
+                    fileName);
+            references.add(reference);
+        }
+
+        // Finally, add a link for the project's custom theme, if it exists
         // and not yet added by the packaged themes loop
         if (!projectCustomThemeAdded) {
             T reference = referenceProvider.apply(themeName.get(), fileName);
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -1685,7 +1685,7 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
         }
 
         // Secondly, check if a parent theme is in the project's frontend/themes
-        // folder
+        // folder and add a link if any
         Optional<String> parentThemeName = FrontendUtils
                 .getParentThemeNameInFrontend(frontendFolder, themeName.get());
         if (parentThemeName.isPresent()) {
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
@@ -659,7 +659,13 @@ public class TaskRunDevBundleBuild implements FallibleCommand {
 
         JsonObject dependencies = packageJson.getObject("dependencies");
 
-        List<String> missingFromBundle = Arrays.stream(dependencies.keys())
+        List<String> dependenciesList = Arrays.stream(dependencies.keys())
+                // skip checking flow-frontend as it was used in previous
+                // versions as an alias for ./target/flow-frontend
+                .filter(pkg -> !"@vaadin/flow-frontend".equals(pkg))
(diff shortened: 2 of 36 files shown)


flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1684,7 +1684,17 @@ private static <T> Collection<T> getStylesheetReferences(
             }
         }
 
-        // Secondly, add a link for the project's custom theme, if it exists
+        // Secondly, check if a parent theme is in the project's frontend/themes
```
Artur-, line 1687: Why do we need to check if it exists? Why don't we always load it?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -333,8 +344,9 @@ private static boolean themeConfigurationChanged(Options options,
                 if (!objectIncludesEntry(content, themeContent.getValue(),
                         missedKeys)) {
                     getLogger().info(
-                            "Packaged custom theme '{}' has imports/assets in 'theme.json' not present in the bundle",
+                            "Parent/Re-used custom theme '{}' has imports/assets in 'theme.json' not present in the bundle",
```
Artur-, line 347:
```suggestion
                            "Parent custom theme '{}' has imports/assets in 'theme.json' not present in the bundle",
```

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
mshabarov: Parent theme can we in the frontend/themes or in the JAR, Flow doesn't know in advance.
This block of codes checks if the `"parent": "my-parent-theme"` is located in the frontend folder, and, if so, adds a link to it.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
Artur-: Is the theme loaded from different URLs depending on where it can be found? Sounds odd but the only reason I can come up with here

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuildTest.java
```diff
@@ -1336,6 +1337,15 @@ public void themeJsonUpdates_statsHasNoThemeJson_projectHasThemeJson_bundleRebui
             utils.when(() -> FrontendUtils
                     .findBundleStatsJson(temporaryFolder.getRoot()))
                     .thenReturn(stats.toJson());
+            utils.when(() -> FrontendUtils.getThemeJsonInFrontend(
```
Artur-, line 1340: Did you count how many times you copy pasted this?

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
mshabarov: > What happens if it is in a jar? Isn't a link added to it then also?

Yes, the link is added also in this case. The code block for this is a few lines above.

Artur- (approved):

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
mshabarov: > Is the theme loaded from different URLs depending on where it can be found? Sounds odd but the only reason I can come up with here

Doesn't depend, this can be refactored to visit all the possible locations with a common code/method.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuildTest.java
mshabarov: Many times :wink: I wasn't sure what is better: to leave it explicitly where need so make the test more readable, or extract a method. I'll make some refactoring.

Artur- (approved):
much better

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
mshabarov: There might be a case when the parent theme in the `frontend/themes` extends another parent theme in `frontend/themes`. For this case the tag won't be added. I have to fix it.

Artur-: In this or another PR?

mshabarov: Better to make another PR for it IMO, because this one is already too big.

PR #16224, opened 2023-03-10
refactor: reduce boilerplate for property editors

sissbruecker:
## Description

- Extract a base class for property editors that:
  - contains common properties
  - has rendering logic for the basic layout
  - shows whether a property is modified
- Removed specific property editor tests for modified state that are now covered by base class, and added some basic tests for text property editor
- Update editor factory logic in property-list component to render a single template with a dynamic tag name


MarcinVaadin (approved):
LGTM

PR #16225, opened 2023-03-10
refactor: Add mode enum to make code easier to read

Artur-:


flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
```diff
@@ -982,8 +983,7 @@ private static void warnAboutPreserveOnRefreshAndLiveReloadCombo(UI ui) {
         // Show a warning that live-reload may work counter-intuitively
         DeploymentConfiguration configuration = ui.getSession()
                 .getConfiguration();
-        if (!configuration.isProductionMode()
-                && configuration.isDevModeLiveReloadEnabled()) {
```
mshabarov, line 986: `isDevModeLiveReloadEnabled` refers to parameter `devmode.liveReload.enabled` which enables/disables browser live reload mechanism in Flow.

I.e. while the app runs with Vite dev server, the live reload can be turned off through the config property, so I think the condition here must be:

```java
if (configuration.getMode() == Mode.DEVELOPMENT_FRONTEND_LIVERELOAD && configuration.isDevModeLiveReloadEnabled())
```

Artur-: True that, fixed

mshabarov (approved):

PR #16230, opened 2023-03-10
refactor: add theme editor license check

sissbruecker:
## Description

WIP


vaadin-dev-server/src/main/resources/META-INF/frontend/vaadin-dev-tools/theme-editor/editor.ts
```diff
@@ -162,9 +168,20 @@ export class ThemeEditor extends LitElement {
     `;
   }
 
+  renderMissingLicenseNotice() {
```
MarcinVaadin, line 171: Can we reuse messageHtml from License.ts? it's no point of duplicating same. It could also simplify handling of no license response.

PR #16233, opened 2023-03-10
Upgrade commons-io to 2.11.0

tepi:


czp13 (approved):

flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/production/ProductionModeCopyStep.java
```diff
@@ -130,12 +130,8 @@ private String getPackageDirectory(String bowerJsonPath) {
     }
 
     private String getPackageName(ArtifactData webJar, String nameSourceJarPath) {
-        String fileContents;
-        try {
-            fileContents = IOUtils.toString(jarContentsManager.getFileContents(webJar.getFileOrDirectory(), nameSourceJarPath), StandardCharsets.UTF_8.displayName());
-        } catch (IOException e) {
-            throw new UncheckedIOException(String.format("Unable to read file '%s' from webJar '%s'", nameSourceJarPath, webJar.getFileOrDirectory()), e);
-        }
+        String fileContents = IOUtils.toString(jarContentsManager.getFileContents(webJar.getFileOrDirectory(), nameSourceJarPath), StandardCharsets.UTF_8.displayName());
```
czp13, line 133: Checked out the branch. Indeed it is not throwing any more `IOException`, but it can throw `NullPointerException`.

If this part never produces null:
``` 
jarContentsManager.getFileContents(webJar.getFileOrDirectory(), nameSourceJarPath)
```
And for the first look, it seems it will not, then we are safe!

So I will approve this, just needed to think it through :), and share my thought process.

czp13 (approved):

PR #16240, opened 2023-03-13
refactor: add vanilla-colorful dependency to dev tools

sissbruecker:
## Description

Adds the vanilla-colorful NPM package to the dev tools in order to be used by the theme editor color picker. Adding the package in source-form, as it is currently not possible to add NPM dependencies to the dev tools otherwise.

This can be improved in the future by [adding a build step to the dev tools](https://github.com/vaadin/flow/pull/16212), which would allow just installing dependencies from NPM.


MarcinVaadin (approved):
Approved for the moment but in final solution it should be used as external dependency.

PR #16242, opened 2023-03-13
refactor: implement color property editor in theme editor

sissbruecker:
## Description

Adds a custom editor for CSS properties that represent colors. The component is structured into two parts:
- A more general color picker component, which is mostly a wrapper for putting a vanilla-colorful picker into an overlay. 
- The actual color property editor that uses the same layout as the other property editors and contains a color-picker as well as an input for displaying and entering color values.

### DIFF
--- a/vaadin-dev-server/package.json
+++ b/vaadin-dev-server/package.json
@@ -19,7 +19,8 @@
     "sinon": "^15.0.1"
   },
   "dependencies": {
-    "@vaadin/button": "24.0.0-rc1",
+    "@vaadin/button": "24.0.0",
+    "@vaadin/overlay": "24.0.0",
     "lit": "^2.6.1"
   }
 }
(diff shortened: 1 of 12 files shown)


sissbruecker:
Building the dev bundle currently fails with
```
[vite]: Rollup failed to resolve import "@vaadin/overlay/src/vaadin-overlay-position-mixin" from "/home/runner/work/flow/flow/flow-tests/test-express-build/test-dev-bundle/frontend/generated/jar-resources/vaadin-dev-tools/theme-editor/editors/color-picker.ts".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
```
I would have assumed that `@vaadin/overlay` would be available when building the dev bundle. We need to look into how we can reference Vaadin component modules from the dev tools.

sissbruecker:
> I would have assumed that @vaadin/overlay would be available when building the dev bundle. We need to look into how we can reference Vaadin component modules from the dev tools.

So it seems we can not import anything from `@vaadin` packages, as that would cause the Vite build in the dev bundle tests to fail where these packages can not be installed. As a long term fix the dev tools UI needs to be factored out into a separate project somehow.

For now I've:
- Removed the import for the Overlay class. Replaced that by getting the class from the browsers `customElements` API, and adding a basic TS interface for it.
- Added a copy of the position mixin to the code-base that we need for positioning the overlay around the color picker's toggle button.

MarcinVaadin (approved):
LGTM

Issue #16231, opened 2023-03-10
CSSImport does not override Lumo


Artur-:
### Description of the bug

The loading order of CSS described in https://vaadin.com/docs/latest/advanced/css-loading-order is
1. The Lumo theme.
2. Application-specific bundled CSS added with a `@CssImport'.
3. Styles added on the Java side via Page::addStylesheet().
4. Application-specific unbundled CSS added with @StyleSheet.
5. Parent theme.
6. Current theme.

This works as expected when running with Vite except that `@CssImport` does not override the Lumo theme.

![image](https://user-images.githubusercontent.com/260340/224025108-d9c350f8-520e-4d9c-89ec-e67d51cbe7e0.png)

### Expected behavior

Lumo styles should always be only default styles and everything else should override

### Minimal reproducible example

-

### Versions

- Vaadin / Flow version: 24.0.0

PR #16243, opened 2023-03-13
fix: Do not import Lumo twice for the document

Artur-:
All Lumo modules except utility are loaded and injected into <head> by https://github.com/vaadin/flow-components/blob/main/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java We should not do that another time

This does not guarantee that the utility classes are injected before other styles but they are not meant to be overridden either


Artur-:
This depends on  #16230 for the test

Artur-:
Scrap that, it does not need that test nor does it yet resolve #16231. That needs another PR

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
```diff
@@ -186,15 +172,13 @@ function generateThemeFile(themeFolder, themeName, themeProperties, options) {
       imports.push(`import { ${lumoImport} } from '@vaadin/vaadin-lumo-styles/${lumoImport}.js';\n`);
     });
 
-    if (useDevServer) {
-      lumoImports.forEach((lumoImport) => {
+    lumoImports.forEach((lumoImport) => {
+      if (lumoImport === 'utility') {
```
mshabarov, line 176:
```suggestion
      // Lumo is injected to the document by Lumo itself, except for the utility module
      if (lumoImport === 'utility') {
```

mshabarov (changes requested):
Any test covering this patch?

Artur-:
All existing tests that use Lumo pass so they should test that Lumo still works. I would not add a test that Lumo is not loaded twice because that tests the code and not any feature. 

The real problem that should be resolved is #16231, which has a disabled test in #16241. It can be fixed more easily once the test and this PR have been merged. It is an attempt to keep PRs reasonably sized

mshabarov (approved):

Artur-:
Do we need this for 24.0? I think the theme loading changes that are needed are not in general suitable for a bugfix release

mshabarov:
> Do we need this for 24.0? I think the theme loading changes that are needed are not in general suitable for a bugfix release

Since this patch fixes what was recently added to 24.0, I assumed 24.0 deserves this fix, especially that the codebase is almost equivalent and this won't require any additional maintenance effort from us.

PR #16247, opened 2023-03-13
fix: Import all document styles using style tags

### LINKED ISSUE #16231: CSSImport does not override Lumo

Artur-:
This adds to comments into `<head>`: `CSSImport end` and `Stylesheet end`. The corresponding styles are always inserted right before these comments so the order of CSS imports is maintained, even when you do not add styles in the order that they should override each other.

After this, it should be easy to support hot reload for the theme CSS

Fixes #16231


flow-client/src/main/java/com/vaadin/client/ResourceLoader.java
```diff
@@ -639,8 +638,27 @@ public void inlineStyleSheet(String styleSheetContents,
 
             addCssLoadHandler(styleSheetContents, event, styleSheetElement);
 
-            getHead().appendChild(styleSheetElement);
+            addBeforeComment(styleSheetElement, "Stylesheet end");
+        }
+    }
+
+    private void addBeforeComment(Element element, String comment) {
+        elemental.dom.Node commentNode = findComment(getHead(), comment);
```
mshabarov, line 646: `commentNode` might be `null` (which signals a bug in the code most likely, because the comment node is expected), so would it be better to fail fast if nothing was found?

flow-server/src/main/resources/META-INF/frontend/theme-util.js
```diff
@@ -42,21 +42,48 @@ const createLinkReferences = (css, target) => {
...
+
+  let beforeThis = undefined;
+  if (referenceComment) {
+    const comments = Array.from(document.head.childNodes).filter(elem => elem.nodeType === Node.COMMENT_NODE);
+    const container = comments.find(comment => comment.data.trim() === referenceComment);
+    if (container) {
+      beforeThis = container;
+    }
+  }
+  document.head.insertBefore(styleTag, beforeThis);
```
mshabarov, line 68: What does insertBefore do if the beforeThis is undefined? :thinking_face: 
Apparently it doesn't throw error, but will it add the tag to the end?

mshabarov (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuildTest.java
```diff
@@ -78,6 +96,11 @@ private JsonObject getBasicStats() {
                     dependency.getValue());
         }
 
+        bundleImports.set(bundleImports.length(),
+                "./generated/jar-resources/theme-util.js");
+        frontendHashes.put("theme-util.js",
```
mshabarov, line 101: Off-topic: just realised that the refactoring with adding new `theme-util.js` file will trigger app bundle rebuild in the projects after upgrading to 24.0.x. Not a big problem, but good to avoid such a things for 24.0.

Artur-: None of this should go into 24.0. Why would it trigger a rebuild when upgrading to 24.1 except for when you use an app specific bundle, and then you rebuilt anyway?

flow-server/src/main/resources/META-INF/frontend/theme-util.js
Artur-: It will append

Artur-: "The node before which newNode is inserted. If this is null, then newNode is inserted at the end of node's child nodes."

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuildTest.java
mshabarov: Yes, right. I thought this fix will target 24.0 as well, but not.

flow-client/src/main/java/com/vaadin/client/ResourceLoader.java
```diff
@@ -639,8 +638,27 @@ public void inlineStyleSheet(String styleSheetContents,
...
+        getHead().insertBefore(element, commentNode);
+    }
+
+    private elemental.dom.Node findComment(HeadElement head, String comment) {
```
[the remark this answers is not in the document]
Artur-: Fixed

flow-client/src/main/java/com/vaadin/client/ResourceLoader.java
Artur-: Added error logging for that case. Sounds better than showing a blank page

mshabarov (approved):

PR #16264, opened 2023-03-15
chore: Delete code that has no effect

Artur-:


Artur-:
Alright, so this is still needed for exported web components..

PR #16267, opened 2023-03-15
fix: Handle hierarchy of parent themes in frontend folder

mshabarov:
Handles hierarchy of themes, like shown below:
```
frontend/themes
   |_ parent-theme-2
   |_ parent-theme-1 (refers parent-theme-2 in theme.json)
   |_ custom-theme (refers parent-theme-1 in theme.json)
```


flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1678,30 +1685,32 @@ private static <T> Collection<T> getStylesheetReferences(
                 String packagedThemeName = themeFolder.getName();
```
Artur-, line 1685: Is it so that you first loop through jar resources here and try to look for the parent theme, and if you find it you call `addThemeReferenceCallback.accept(themeName);`. Then you look through the frontend folder and if you find the theme, you call `addThemeReferenceCallback.accept(themeName);` (if it was found in both you ignore it).

What is the difference to just calling `addThemeReferenceCallback.accept(themeName);` for the theme and its parents?

mshabarov: Good comment, I agree that it's better and more clear to collect all the themes into `LinkedHashSet` first (ordering does matter), then iterate through it and add a link for each of them. (Hope this is that you meant :) )

Artur-: Yes but ordering matters. Parent before child

mshabarov: Yes, it matters for sure.

mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1662,48 +1664,61 @@ private static <T> Collection<T> getStylesheetReferences(
...
-        // Secondly, check if a parent theme is in the project's frontend/themes
-        // folder and add a link if any
-        Optional<String> parentThemeName = FrontendUtils
-                .getParentThemeNameInFrontend(frontendFolder, themeName.get());
-        if (parentThemeName.isPresent()) {
-            T reference = referenceProvider.apply(parentThemeName.get(),
-                    fileName);
-            references.add(reference);
-        }
+        // Secondly, collect the themes in {project.root}/frontend/themes,
```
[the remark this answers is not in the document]
mshabarov: There might be a hierarchy of themes: parent theme may have other parent theme in the frontend (this is probably a rare case, but anyway). Without scanning the theme.json files in frontend/themes we cannot figure out how many parent themes we have in the hierarchy.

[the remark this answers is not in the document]
Artur-: Ok but how do you get the order correct this way when you separately scan the packaged themes and the frontend folder? Shouldn't you find the active theme, no matter where it is, then its parent wherever it is etc?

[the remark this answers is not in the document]
mshabarov: Hmm, yes, you are right. The themes from `generated/jar-resources/themes` are not checked vs their activeness, and order is not taken into account. Better to pick recursively only these themes from JAR that are referenced from theme.json of the themes in frontend/themes (active themes). This approach should also fix the ordering uncertainty.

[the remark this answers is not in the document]
mshabarov: Reworked - active themes are collected recursively in one pass, preserving an order.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1659,48 +1660,64 @@ private static <T> Collection<T> getStylesheetReferences(
...
+        });
 
         return references;
     }
 
+    private static <T> void collectActiveThemes(String themeName,
+            File frontendFolder, Set<String> packagedThemes,
+            Set<String> activeThemes) throws IOException {
+        // Look for parent theme of themeName in {project}/frontend/themes/
+        Optional<String> parentThemeName = FrontendUtils
```
Artur-, line 1701: This version is much better but it is still odd. Why do we gather packaged themes separately? We know the theme is in `frontend/themes` or `frontend/generated/jar-resources/themes` and can just check one first and then the other. So e.g. have a helper that is `FrontendUtils.getThemeFolder(themeName)` instead of a weird `getParentThemeName` that apparently only gets the parent theme name if the theme is in one of the two possible locations?

Artur-: Something like this is needed for #16455 also, can be done as part of that

Artur- (approved):

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
mshabarov: Let's do more refactoring separately, yes.

mshabarov: > With this, wouldn't getStylesheetReferences basically become a loop through the active themes and mapping them to another value through the callback?

Yes, the idea is at least in what you described, get the active themes hierarchy and create a reference for each.

mshabarov: > We know the theme is in frontend/themes or frontend/generated/jar-resources/themes and can just check one first and then the other. So e.g. have a helper that is FrontendUtils.getThemeFolder(themeName) instead of a weird getParentThemeName that apparently only gets the parent theme name if the theme is in one of the two possible locations?

Or we can collect all the themes from these two locations into a single collection and recursively build the active themes list, crawling through this collection.

PR #16268, opened 2023-03-15
refactor: theme scope selection

sissbruecker:
## Description

- Adds a scope selector for switching theming between local (instance) and global (all components of type) scope
- Updates API calls to also send instance's node ID when using local scope
- Align data structures with backend changes
- Shows notice if selected instance is not stylable
- Move all sub-components of the theme editor into a `components` folder


MarcinVaadin (approved):
LGTM

PR #16270, opened 2023-03-15
refactor: highlight selected component in theme editor

sissbruecker:
## Description

Adds an outline to the selected component in the theme editor. The outline gets removed when the component picker is opened again, or when the editor is removed from the DOM.

Bildschirmfoto 2023-03-15 um 21 27 39
Bildschirmfoto 2023-03-15 um 21 27 29

### DIFF
--- a/vaadin-dev-server/frontend/theme-editor/editor.ts
+++ b/vaadin-dev-server/frontend/theme-editor/editor.ts
@@ -15,6 +15,14 @@ import './components/scope-selector';
 import './components/property-list';
 import '../component-picker.js';
 import { ComponentReference } from '../component-util';
+import { injectGlobalCss } from './styles';
+
+injectGlobalCss(css`
+  .vaadin-theme-editor-highlight {
+    outline: solid 3px #9e2cc6;
+    outline-offset: 3px;
+  }
+`);
 
 @customElement('vaadin-dev-tools-theme-editor')
 export class ThemeEditor extends LitElement {
@@ -141,6 +149,12 @@ export class ThemeEditor extends LitElement {
     this.historyActions = this.history.allowedActions;
   }
 
+  disconnectedCallback() {
+    super.disconnectedCallback();
+
+    this.removeElementHighlight(this.context?.component.element);
+  }
+
   render() {
     if (this.themeEditorState === ThemeEditorState.missing_theme) {
       return this.renderMissingThemeNotice();
@@ -257,6 +271,8 @@ export class ThemeEditor extends LitElement {
   }
 
   private async pickComponent() {
+    this.removeElementHighlight(this.context?.component.element);
+
     this.pickerProvider().open({
       infoTemplate: html`
         <div>
@@ -275,6 +291,8 @@ export class ThemeEditor extends LitElement {
           this.effectiveTheme = null;
           return;
         }
+
+        this.highlightElement(component.element);
 
         this.refreshTheme({
           scope: this.context?.scope || ThemeScope.local,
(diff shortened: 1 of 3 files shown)


sissbruecker:
I think we should align the style used in the component picker with the one used in the theme editor. I based the highlight style on the Figma design, but I'm not sure about the outline color, it seems it might have a low contrast on darker backgrounds. 

@jouni Any opinion on this?

vaadin-dev-server/frontend/theme-editor/editor.ts
```diff
@@ -15,6 +15,14 @@ import './components/scope-selector';
...
 import '../component-picker.js';
 import { ComponentReference } from '../component-util';
+import { injectGlobalCss } from './styles';
+
+injectGlobalCss(css`
+  .vaadin-theme-editor-highlight {
+    outline: solid 3px #9e2cc6;
+    outline-offset: 3px;
+  }
+`);
```
sissbruecker, line 25: I tried reusing the [global CSS class defined in the component picker](https://github.com/vaadin/flow/blob/bd2ef108fb1d7e0b5bb26e2718d9d2146df75d2f/vaadin-dev-server/frontend/component-picker.ts#L57-L60), but there are currently timing issues with the theme editor adding the class name when the element is picked, and the component picker removing the same class when it closes itself. Seems better to use different classes and avoid any side-effects between the two.

vaadin-dev-server/frontend/theme-editor/editor.ts
sissbruecker, line 1: I think we should align the style used in the component picker with the one used in the theme editor. I based the highlight style on the Figma design, but I'm not sure about the outline color, it seems it might have a low contrast on darker backgrounds. 

@jouni Any opinion on this?

sissbruecker: As demonstrated by Tarek, it's not immediately obvious that something is happening if you hover over something that spans the whole client area, because the outline is cut off. A more sophisticated solution could be to add an semi-transparent overlay element, similar to the browsers dev tools. Still, maybe we should go with a low-effort solution for now, and improve this later.

jouni:
> I'm not sure about the outline color, it seems it might have a low contrast on darker backgrounds.

It should have a 2px solid white border/outline inside the colored border/outline as well.

vaadin-dev-server/frontend/theme-editor/editor.ts
jouni: > A more sophisticated solution could be to add an semi-transparent overlay element

Yes, that’s what I was using in the Figma prototypes.

jouni: > I'm not sure about the outline color, it seems it might have a low contrast on darker backgrounds.

It should have a 2px solid white border/outline inside the colored border/outline as well.

vaadin-dev-server/frontend/theme-editor/editor.ts
jouni: Oh, so the implementation is actually adding a class name on the target element itself, is it? 

I thought we would be using a “scrim” instead, where the outline is overlaid on top of the actual UI. We can't rely on the target element, as it could override the outline styles we rely on, and some parent element can end up clipping the outline as well, making it look broken. It really should eventually be a separate element with full control of its styles.

vaadin-dev-server/frontend/theme-editor/editor.ts
jouni: > I think we should align the style used in the component picker with the one used in the theme editor

100% agreed.

sissbruecker: Decided to go with a simple solution for now: Just use the outline property with the styles proposed in this PR, and also align the component picker to use the same styles.

vaadin-dev-server/frontend/theme-editor/editor.ts
sissbruecker: After discussion, we go with the simple solution for now, and add a overlay solution later on.

MarcinVaadin (approved):
Tested and accepted

jouni (dismissed):

MarcinVaadin (approved):

PR #16273, opened 2023-03-16
refactor: fix base theme detection in theme editor

sissbruecker:
## Description

After merging the scope changing, the base theme detection used for showing default values in the property editor does not work correctly anymore. For example in the following case:
- Change background color globally
- Switch to local scope
- Property editor shows initial background color, because the base theme has not been updated with the modified global background color

A similar problem exists with undo/redo, which can also affect properties outside of the current scope.

This fix updates the base theme whenever the scope is changed, or undo / redo is performed.


MarcinVaadin (approved):
LGTM

PR #16296, opened 2023-03-17
fix: Make applications esm modules instead of cjs

Artur-:
This is not supposed to change any behavior as Vite already requires all project frontend files to be ESM.


Artur-:
@platosha seems to work

mshabarov (approved):

PR #16300, opened 2023-03-17
test: Delete nonsense test

Artur-:


mshabarov (approved):
This test looks cryptic indeed. And also do we really need to check polyfills in V24?

Artur-:
I think the polyfill it tried to test was removed years ago

PR #16330, opened 2023-03-21
fix: Always set development mode when in development mode

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -136,6 +136,12 @@ public boolean synchronizedHandleRequest(VaadinSession session,
         // modify the page based on registered IndexHtmlRequestListener:s
         service.modifyIndexHtmlResponse(indexHtmlResponse);
 
+        if (!config.isProductionMode()) {
+            // Ensure no older tools incorrectly detect a bundle as production
+            // mode
+            addScript(indexDocument,
+                    "window.Vaadin = window.Vaadin || {}; window.Vaadin.developmentMode = true;");
```
mshabarov, line 143: Exported web components probably also need this, so let's include it into `WebComponentBootstrapHandler` (or even only into `BootstrapHandler`) ?

Artur-: Should exported web components affect global variables on the page? They probably do it already to some degree but say that you would run a Flow app in production mode and import an exported wc that is accidentally running in development mode. Should this say that the Flow app is running in development mode?

mshabarov (approved):

PR #16332, opened 2023-03-21
chore: also cancel builds for main branches when newer pushes (#16327) (CP: 24.0)

manolo:


czp13:
I think this should be good, just some flaky tests... Merging, rerunning, and approving this as well from the flow side. 🙇
Thanks.

PR #16337, opened 2023-03-21
chore: Throw an exception when a download fails

Artur-:
Makes it easier to understand that download of Node failed instead of that the archive was corrupt


czp13 (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/DefaultFileDownloader.java
```diff
@@ -21,10 +21,12 @@
 import java.net.URI;
 import java.net.URL;
 import java.nio.channels.Channels;
+import java.nio.channels.FileChannel;
```
czp13, line 24: Is this import needed?
```suggestion
```

Artur-: No :)

czp13 (approved):

PR #16351, opened 2023-03-22
Revert lumo changes from 24.0

Artur-:
These changes break typography import


Artur-:
The problem seems to be noticeable only when you have e.g. texts that are not inside any Vaadin web component. Then the correct font is not applied

mshabarov (approved):

PR #16364, opened 2023-03-22
refactor: hide selection highlight when closing theme editor

sissbruecker:
## Description

Removes / restores the selected element highlight when closing / opening the dev tools.

Addresses feedback from DX tests that highlight is still visible after closing dev tools.

https://user-images.githubusercontent.com/357820/226929991-68a54d70-f252-4872-8bea-c2782448963f.mp4

### DIFF
--- a/vaadin-dev-server/frontend/theme-editor/editor.test.ts
+++ b/vaadin-dev-server/frontend/theme-editor/editor.test.ts
@@ -47,6 +47,7 @@ describe('theme-editor', () => {
     };
     const pickerProvider: PickerProvider = () => pickerMock as any;
     const editor = (await fixture(html` <vaadin-dev-tools-theme-editor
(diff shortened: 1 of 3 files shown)


sissbruecker:
> How it would look after we implement dragging of theme editor / dev tools?

I don't know, it's not really clear yet how we want this to work. But I assume you'll still be able to close the dev tools somehow, at which point we want to remove the highlight.

MarcinVaadin (approved):

jouni:
> How it would look after we implement dragging of theme editor / dev tools?

The plan/idea I have at the moment is to redesign/refactor the dev tools into separate pieces you can open independently of each other, instead of a single dialog. The current collapsed button in the bottom right corner would then be more like a popup button/menu, from which you can open the individual dev tools.

PR #16365, opened 2023-03-22
refactor: theme editor api updates

MarcinVaadin:
## Description

In this PR:
- tag and part has been replaced by selectors
- javasource modifications has been unified
- local classname operations has been updated and new classname suggestion has been added
- updated javaparser-core version to 3.24.10 same as in https://github.com/vaadin/flow-components/pull/4848
- removed javaparser-symbol-solver-core dependency (not required currently)

### themeEditorComponentMetadata
Returns component **className** if present, otherwise **suggestedClassName**. Includes **accessible** if component can be modified in Local mode.

BaseRequest
`{ requestId: string, uiId: int, nodeId: int }`
ComponentMetadataResponse
`{code: 'ok', accessible: boolean, className: string | null, suggestedClassName: string | null }`

### themeEditorHistory
No changes

### themeEditorLoadPreview
No API changes, response CSS with removed comments and unnecessary whitespaces.

### themeEditorLoadRules
Selector filter replaced by list of **selectors**.

LoadRulesRequest
`{ requestId: string, uiId: int, selectors: [string] }`
LoadRulesResponse
`{ requestId: string, rules: [ { selector: string, properties: { name: value } } ] }`

### themeEditorLocalClassName
Different command name to be inline with rest of code. Supports single **className**.

SetClassNameRequest
`{ requestId: string, uiId: int, nodeId: int, className: string }`
BaseResponse
`{ requestId: string, code: 'ok' }`

### themeEditorRules
**selector** instead of tag and part. No further changes.


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/utils/LocalClassNameVisitor.java
```diff
@@ -0,0 +1,30 @@
...
+import com.github.javaparser.ast.stmt.ExpressionStmt;
+import com.github.javaparser.ast.visitor.GenericVisitorAdapter;
+import com.vaadin.base.devserver.themeeditor.JavaSourceModifier;
+
+/**
+ * Implementation of {@link com.github.javaparser.ast.visitor.GenericVisitor}
+ * that searches for local classname by comparing expression type, method call
+ * scope and expression comment.
+ */
+public class LocalClassNameVisitor
```
sissbruecker, line 13: It looks like this could use a sanity check that the method call is for `addClassName`.

For example if the code would be manually modified to insert some other statement / method call for the same field/variable in between the comment and the `addClassName` call, we would modify that statement instead.

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/themeeditor/ThemeEditorMessageHandlerTest.java
```diff
@@ -81,218 +83,165 @@ public void testCanHandle() {
...
+        ComponentMetadataResponse metadataResponse = (ComponentMetadataResponse) response;
+        Assert.assertTrue(metadataResponse.isAccessible());
+        Assert.assertNull(metadataResponse.getClassName());
+        Assert.assertNotNull(metadataResponse.getSuggestedClassName());
+        Assert.assertEquals("span-1", metadataResponse.getSuggestedClassName());
     }
 
     @Test
-    public void testHandleLoadPreview() {
+    public void testHandle_LocalClassName() {
```
sissbruecker, line 119: This test is not complete, it does not execute the request, and does not assert anything.

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/themeeditor/ThemeEditorMessageHandlerTest.java
```diff
@@ -327,53 +305,36 @@ private void assertResponseError(BaseResponse response, String requestId) {
...
     }
 
+    private BaseResponse setRule(Integer uiId, String requestId,
+            ThemeEditorMessageHandler handler, String selector, String property,
+            String value) {
+        return setRule(uiId, null, requestId, handler, selector, property,
+                value);
+    }
+
     private BaseResponse setRule(Integer uiId, Integer nodeId, String requestId,
```
sissbruecker, line 329: It looks like we don't need this overload? We don't need to pass a `nodeId` when setting rules, and the argument is unused.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/themeeditor/handlers/ComponentMetadataHandler.java
```diff
@@ -0,0 +1,65 @@
...
+import com.vaadin.flow.internal.JsonUtils;
+import elemental.json.JsonObject;
+
+import java.util.Optional;
+
+public class ComponentMetadataHandler implements MessageHandler {
+
+    private final HasSourceModifier hasSourceModifier;
+
+    private class FinalsHolder {
```
sissbruecker, line 18: Holder classes can be static.

sissbruecker (approved):

PR #16372, opened 2023-03-23
refactor: use selectors in theme editor component metadata

sissbruecker:
## Description

- Updates component metadata model to contain a list of elements, each element having a selector and a list of properties. The host itself is now also just an element.
- Update API calls to pass/receive selectors when loading / setting rules
- Update theme model to work with selectors, add logic for generating local and global scoped selectors
- Update theme detector to work with selectors
- Set local class name before sending rules

### DIFF
--- a/vaadin-dev-server/frontend/theme-editor/components/property-list.ts
+++ b/vaadin-dev-server/frontend/theme-editor/components/property-list.ts
@@ -1,7 +1,7 @@
 import { css, html, LitElement } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 import { html as staticHtml, literal, StaticValue } from 'lit/static-html.js';
-import { ComponentMetadata, ComponentPartMetadata, CssPropertyMetadata, EditorType } from '../metadata/model';
+import { ComponentMetadata, ComponentElementMetadata, CssPropertyMetadata, EditorType } from '../metadata/model';
 import { ComponentTheme } from '../model';
 import './editors/text-property-editor';
 import './editors/range-property-editor';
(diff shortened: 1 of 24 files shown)


vaadin-dev-server/frontend/theme-editor/components/property-list.ts
```diff
@@ -29,26 +29,23 @@ export class PropertyList extends LitElement {
   public theme!: ComponentTheme;
 
   render() {
-    const sections = [
-      this.renderSection(null, this.metadata.properties),
-      ...this.metadata.parts.map((part) => this.renderSection(part, part.properties))
-    ];
+    const sections = this.metadata.elements.map((element) => this.renderSection(element));
```
sissbruecker, line 32: For now the section containing the host properties will also have a header / title, which is just `Host`.

We can discuss if we want to hide the header again, or use a different name.

vaadin-dev-server/frontend/theme-editor/detector.ts
```diff
@@ -1,33 +1,46 @@
...
-      const propertyValue = hostStyles.getPropertyValue(property.propertyName);
-      componentTheme.updatePropertyValue(null, property.propertyName, propertyValue);
-    });
+      let element = document.querySelector(lightDomSelector);
+      // If we target a part in shadow DOM, query for that within shadow DOM
+      if (element && partNameMatch) {
+        const partName = partNameMatch[1];
+        const shadowDomSelector = `[part~="${partName}"]`;
+        element = element!.shadowRoot!.querySelector(shadowDomSelector);
+      }
```
sissbruecker, line 34: This is a bit more complex because we can not just pass a selector like `vaadin-button::part(label)` to `document.querySelector`. Querying shadow parts is not possible with that API. So now we need to split the selector, first query the light DOM element, and then the part from it's shadow DOM.

MarcinVaadin (approved):
Tested and working as expected

Issue #16319, opened 2023-03-20
Add license checking to Vaadin Servlet for Flow 9.1 (Vaadin 22.1.0)


mshabarov:
Upcoming Vaadin 22.1.0 (Flow 9.1.0) will be under commercial license. Thus Flow needs a license checking in a runtime during initialising `VaadinServlet`.

jhult:
@mshabarov, does this mean Flow is no longer open source? Has there been an announcement about this?

jhult:
Related: #16301

knoobie:
Flow 9.1 is  the release targeted for Vaadin 22.0 which is EoL since March. Like with V8, they  change the license for the old version for commercial customer support if those need this EoL Version longer. Flow 9.1 == Vaadin 22.1 

This does not change anything for people using already not EoL versions.

mshabarov:
@jhult thanks for asking!
This ticket only affects Flow 9.1 (targets Vaadin 22.1). Vaadin 22.0 is end of life in March 2023 and Vaadin 22.1 has an extended maintenance for Prime+ customers, thus we are adding a license into it.
See https://vaadin.com/releases
```
Vaadin 22 is the latest stable version to support Java 8, and will be exceptionally maintained for an extra year, until March 2023, to help transition to Java 11. Prime customers will get support until Match 2024.
```
Also thanks @knoobie for fast reply and explanations!

PR #16379, opened 2023-03-23
feat!: Add runtime license checking for Flow 9.1 (Vaadin 22.1)

mshabarov:
Fixes https://github.com/vaadin/flow/issues/16319


flow-server/bnd.bnd
```diff
@@ -2,7 +2,8 @@ Bundle-SymbolicName: ${project.groupId}.flow.server
 Bundle-Name: Vaadin Flow Server
 Bundle-Version: ${osgi.bundle.version}
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
-Bundle-License: http://www.apache.org/licenses/LICENSE-2.0
+Bundle-License: https://vaadin.com/commercial-license-and-service-terms
 Import-Package: org.atmosphere*;resolution:=optional;version='${atmosphere.runtime.version}',\
     org.apache.http*;resolution:=optional;,*
 Export-Package: !com.vaadin.flow.push*, com.vaadin.flow*;-noimport:=true
+CvdlName: ${cvdlName}
```
mshabarov, line 9: As the license checking is only in flow-server, I'm not sure whether we need cvdl name for other modules or not.

flow-server/pom.xml
```diff
@@ -227,6 +237,9 @@
                         <manifest>
                             <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                         </manifest>
+                        <manifestEntries>
+                            <CvdlName>${cvdlName}</CvdlName>
```
mshabarov, line 241: Same for CVDL entries in manifest for other Flow modules. Not sure if we need this entry for all the modules or only for flow-server, as other modules might hardly be used without flow-server.

czp13:
Executed the tests some times, ended up with these errors:
- https://github.com/vaadin/flow/pull/16379/checks?check_run_id=12429972674

All of the 40 tests are failing having the same characteristics:
- error message:
```
com.vaadin.pro.licensechecker.LicenseException: 
Unable to validate the license, please check your internet connection. If you need to work offline then please go to https://vaadin.com/pro/validate-license?getOfflineKey=mid-d9819704-d22fea3c to retrieve an offline key.
For CI/CD build servers, you need to download a server license key, which can work offline to create production builds. You can download a server license key from https://vaadin.com/myaccount/licenses.
For troubleshooting steps, see https://vaadin.com/licensing-faq-and-troubleshooting.
	at com.vaadin.pro.licensechecker.LicenseChecker.checkLicense(LicenseChecker.java:292)
	at com.vaadin.pro.licensechecker.LicenseChecker.checkLicense(LicenseChecker.java:226)
	at com.vaadin.pro.licensechecker.LicenseChecker.checkLicense(LicenseChecker.java:129)
	at com.vaadin.pro.licensechecker.LicenseChecker.checkLicense(LicenseChecker.java:104)
```
- coming from `flow-polymer-template`sub project

The related code is:
```        if (proKey == null) {
            // No proKey found - probably first launch

            proKey = vaadinComIntegration.openBrowserAndWaitForKey(product, noKeyUrlHandler);
            if (proKey != null) {
                LocalProKey.write(proKey);
                if (proKeyValidator.validate(product, proKey, machineId, buildType) == Result.OK) {
                    // Online validation OK
                    return;
                }
            }
        }
````

My best guess really is that we do not have a server key or pro key (not sure which one we use for building servers), added here).  I will catch up with Zhe this Monday.

**Review by czp13 — CHANGES_REQUESTED** [review]
ProKey/Server key thin blocking this and manual testing.

**Review by czp13 — APPROVED** [review]

File: flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java [hunk 3]
```diff
@@ -489,4 +494,9 @@ public static void updateBuildFile(PluginAdapterBuild adapter) {
             adapter.logWarn("Unable to read token file", e);
         }
     }
+
+    private static void validateLicense() {
+        LicenseChecker.checkLicense("flow", Version.getFullVersion(),
```
mshabarov, line 499: The product name for Flow is "flow", hope it is fine and not overlap any existing product on the license server. After this getting merged, we need to add it to the product list.

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
```diff
@@ -400,6 +403,8 @@ private static void runFrontendBuildTool(PluginAdapterBase adapter,
             Map<String, String> environment, String... params)
             throws TimeoutException, URISyntaxException {
 
+        validateLicense();
```
mshabarov, line 406: Commercial components check was not backported into Flow 9.0, thus, there is no check like in the latest versions, see https://github.com/vaadin/flow/blob/ca8bdee13523451da585b5afe865ffa67f4a6a9a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java#L515.

Since the Flow 9.1 is now a commercial itself, maybe components check is not needed, but I'm not sure 100%.

Artur-: Sounds like out of scope for this change. The other parts can be backported but shouldn't be unless there is a good reason.

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
Artur-: Should be fine

flow-server/bnd.bnd
Artur-: I don't think we need CvdlName at all

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
```diff
@@ -541,4 +544,14 @@ private VaadinServletContext initializeContext() {
         return vaadinServletContext;
     }
 
+    private void validateLicense(
+            DeploymentConfiguration deploymentConfiguration) {
+        // Check the license at runtime if in development mode
+        if (!deploymentConfiguration.isProductionMode()) {
+            // Using a null BuildType to allow trial licensing builds
```
Artur-, line 551: Where does this come from? Why not use `BuildType.DEVELOPMENT`?

pom.xml
```diff
@@ -265,6 +265,11 @@
                 <artifactId>hamcrest-all</artifactId>
                 <version>1.3</version>
             </dependency>
+            <dependency>
+                <groupId>com.vaadin</groupId>
+                <artifactId>license-checker</artifactId>
+                <version>1.12.3</version>
```
Artur-, line 271: @ZheSun88 should this be a property instead for some auto-updating?

Artur- (changes requested):

flow-server/bnd.bnd
mshabarov: Hmm, since we call the LC directly for Flow, probably this parameter is not needed, yes. It is needed to let LC know what this is a commercial product and what is the name of it (e.g. commercial components)

flow-server/src/main/java/com/vaadin/flow/server/VaadinServlet.java
mshabarov: I spied it here https://github.com/vaadin/sso-kit/pull/58/files#diff-ebb4a6381d3d39073dbcdd3726cf525e28136f738aad157c7d7fa6c8c098c762R45

Will change to BuildType.DEVELOPMENT. Agree

mshabarov: Done

flow-server/bnd.bnd
mshabarov: Removed

Artur- (approved):

PR #16384, opened 2023-03-23
refactor: add menubar metadata to theme editor

sissbruecker:
## Description

Adds metadata for the MenuBar component. 

Also:
- Improves theme detection to work with more complex components. To detect styles for components that have children or overlays we need an instance that has all of these set up. Relying on the selected element seems brittle, as that might not have all of the elements we want to measure, and its custom styles would lead to incorrect results for detecting styles for the global scope. It gets even more difficult for data-driven components like ComboBox. Having a generic setup for all types of components seems impossible, so for now, for each complex component the metadata defines functions on how to properly set up or clean up an instance from which styles can be detected.
- Adds a property editor with a checkbox to allow toggling bold / italic text styles. Could be improved in the future to use a single editor for multiple properties. Checkbox is not styled at the moment.
- Adds a customization to the server rule generation, so that when setting a `border-width` larger than zero, it also sets the `border-style` property to solid. This avoids the user having to set the border style separately for a border to show up.

The generated style rules for the menu bar overlay depend on the overlay class name being set on the menu bar, which means the backend also needs to be changed so that it applies the local class name as overlay class name. In the meanwhile for testing this, the overlay class name can be applied manually:
```java
menuBar.setOverlayClassName("local-class-name");
```

~~PR depends on the changes from https://github.com/vaadin/flow/pull/16372, and is currently set to merge against it.~~

### DIFF
--- a/vaadin-dev-server/frontend/theme-editor/components/editors/checkbox-property-editor.test.ts
+++ b/vaadin-dev-server/frontend/theme-editor/components/editors/checkbox-property-editor.test.ts
@@ -1,0 +1,75 @@
+import { elementUpdated, expect, fixture, html } from '@open-wc/testing';
+import { ComponentTheme } from '../../model';
+import { ComponentElementMetadata, CssPropertyMetadata } from '../../metadata/model';
+import { testElementMetadata } from '../../tests/utils';
+import { TextPropertyEditor } from './text-property-editor';
+import './checkbox-property-editor';
+import sinon from 'sinon';
+
+const fontWeightMetadata: CssPropertyMetadata = {
+  propertyName: 'font-weight',
+  displayName: 'Bold',
+  checkedValue: 'bold'
+};
+const labelMetadata: ComponentElementMetadata = {
+  selector: 'test-element::part(label)',
+  displayName: 'Label',
+  properties: [fontWeightMetadata]
+};
+
+describe('checkbox property editor', () => {
+  let theme: ComponentTheme;
+  let editor: TextPropertyEditor;
+  let valueChangeSpy: sinon.SinonSpy;
+
+  function getCheckbox() {
+    return editor.shadowRoot!.querySelector('input') as HTMLInputElement;
+  }
+
+  function cloneTheme() {
+    const result = new ComponentTheme(testElementMetadata);
+    result.addPropertyValues(theme.properties);
+    return result;
+  }
+
+  beforeEach(async () => {
+    theme = new ComponentTheme(testElementMetadata);
+    theme.updatePropertyValue(labelMetadata.selector, 'font-weight', 'bold');
+    valueChangeSpy = sinon.spy();
+
+    editor = await fixture(html` <vaadin-dev-tools-theme-checkbox-property-editor
+      .theme=${theme}
+      .elementMetadata=${labelMetadata}
+      .propertyMetadata=${fontWeightMetadata}
+      @theme-property-value-change=${valueChangeSpy}
+    >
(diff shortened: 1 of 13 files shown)


MarcinVaadin (approved):
It is bit confusing that picking `menu-bar-item` does nothing. Picking `menu-bar` works fine. IMO if child components are included within parent component they shouldn't be picked individually or "redirected" to styling parent component.

IMO border width could also be slider by default.

Bold and italic are fine but eventually should look like in text editors - as button group (maybe with underline also).

PR does LGTM although needs to be addressed in next iteration. Please merge if no further updates in given PR.

vaadin-dev-server/frontend/theme-editor/components/editors/checkbox-property-editor.test.ts
```diff
@@ -0,0 +1,75 @@
...
+import { testElementMetadata } from '../../tests/utils';
+import { TextPropertyEditor } from './text-property-editor';
+import './checkbox-property-editor';
+import sinon from 'sinon';
+
+const fontWeightMetadata: CssPropertyMetadata = {
+  propertyName: 'font-weight',
+  displayName: 'Bold',
+  checkedValue: 'bold'
+};
```
jouni, line 13: Font weight is actually nine different values (100,200,...,800,900), not a boolean. That said, this isn't a big deal at this point, and can be improved later. There might eventually be Lumo design tokens for font weight as well.

vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-menu-bar.ts
```diff
@@ -0,0 +1,166 @@
...
+          displayName: 'Bold',
+          editorType: EditorType.checkbox,
+          checkedValue: 'bold'
+        },
+        {
+          propertyName: 'font-style',
+          displayName: 'Italic',
+          editorType: EditorType.checkbox,
+          checkedValue: 'italic'
+        }
```
jouni, line 77: While this is easy to add, I think it adds noise to the editor 99% of the time, as italics are rarely used in UI design.

vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-menu-bar.ts
```diff
@@ -0,0 +1,166 @@
...
+        },
+        {
+          propertyName: 'border-color',
+          displayName: 'Border color',
+          editorType: EditorType.color
+        },
+        {
+          propertyName: 'border-width',
+          displayName: 'Border width'
+        },
```
jouni, line 97: Missing border-radius (as discussed, all three would always be added together).

vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-menu-bar.ts
sissbruecker: OK, let's remove it then. It seemed like something obvious for text, but then again you usually use this in some form of rich text rather than UIs.

vaadin-dev-server/frontend/theme-editor/components/editors/checkbox-property-editor.test.ts
sissbruecker: Discussed this, for now a simple toggle is enough.

sissbruecker:
After discussing with Jouni, added a few tweaks:
- Use range slider for border-width
- Added border-radius property
- Removed font-style / italics property

MarcinVaadin (approved):
Looks better after updates.

PR #16399, opened 2023-03-24
fix: Decode project folder when using classpath

### LINKED ISSUE #16396: Running an application with spaces in path fails when using IDEA
### Description of the bug

When I use IDEA to run a newly-created Flow project in a directory which has a space in its name, another directory is created aside of it, with `%20` instead of space, and some Vaadin-specific files are created there instead of the real project dir. Then, execution fails.

If I execute the project using Maven, everything works. And once that is done, I'm also able to run the project using IDEA.

`com.vaadin.flow.server.ExecutionFailedException: Discovered @Theme annotation with theme name 'myapp', but could not find the theme directory in the project or available as a jar dependency. Check if you forgot to create the folder under 'C:\Users\Luciano\Projects\my%20app\.\frontend\themes' or have mistyped the theme or folder name for 'myapp'.`

I can reproduce the same behavior in Hlla 2.1 projects.

### Expected behavior

The application should create all files inside the project dir.

- Go to start.vaadin.com and create a project with pre-release version.
- Unpack the `my-app` dir and rename it to `my app`
- Open the project in IDEA
- Run the `Application` class using the "play" button

### Versions

- Vaadin / Flow version: 24.1.0.alpha1or 24.1-SNAPSHOT
- Java version: 17
- OS version: Windows 10 or Linux
- IDE (if applicable): IntelliJ IDEA

Artur-:
Fixes #16396

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
@@ -18,8 +18,12 @@ package com.vaadin.flow.server;
 import java.io.File;
 import java.io.Serializable;
 import java.net.URL;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+
+import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.internal.hilla.EndpointRequestUtil;
 import com.vaadin.flow.server.frontend.FrontendUtils;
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
```diff
@@ -18,9 +18,13 @@
 import java.io.File;
 import java.io.Serializable;
 import java.net.URL;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
```
mcollovati, line 22: Can these imports be removed?

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
Artur-: ✅

czp13 (approved):
Nice, LGTM, wanted to say the imports thingie, but already was removed :)

PR #16400, opened 2023-03-24
refactor: add select metadata to theme editor

sissbruecker:
## Description

Adds metadata for the Select component. Also updates the theme detector to be able to detect pseudo element styles.

We could think about reusing some of the property definitions somehow, but let's maybe add some more components first to see what patterns come up.


MarcinVaadin (approved):
Tested and works fine

PR #16401, opened 2023-03-24
fix: add .js to imports used by Hilla 2.0 (target: 24.0)

cromoteca:
Those imports are marked as errors since Hilla projects are supposed to use ES modules.


Artur- (approved):

Artur-:
Is this not a problem in `main`?

cromoteca:
A more general fix is already in place in 24.1, but not backported.

czp13 (approved):

PR #16413, opened 2023-03-25
refactor: add basic text fields metadata to theme editor

sissbruecker:
## Description

Add / improve metadata for text-field based components:
- BigDecimalField
- EmailField
- IntegerField
- NumberField
- TextArea
- TextField

PR #16414, opened 2023-03-25
refactor: allow theming non-Vaadin components

sissbruecker:
## Description

Allows styling non-Vaadin components, such as `H1`, `Div`, `Span`, etc., using a generic set of metadata. The generic metadata is used whenever a component's tag name does not start with `vaadin-`. For Vaadin components we should attempt to cover them with specific metadata, so the generic metadata is not used as fallback for Vaadin components that don't have their own metadata yet.

~~Based on the changes from https://github.com/vaadin/flow/pull/16413 and is currently set to merge against it.~~


MarcinVaadin (approved):
Tested and LGTM

PR #16421, opened 2023-03-27
refactor: add metadata for layout components

sissbruecker:
## Description

Adds initial metadata for all layout based components.
- Most layouts have properties for background, border, padding.
- `HorizontalLayout` and `VerticalLayout` also have `gap`. This could also be configured using the spacing theme variants in Java code, but since we can't add those yet, configuring gap is helpful.
- `VirtualList` does not have padding, might cause issues with the virtual scrolling implementation.
- `AppLayout` does not expose border, which doesn't make sense for the whole component, and does not work correctly for individual parts. Also no padding on the host element, because it doesn't make sense.
- `SplitLayout` also doesn't have padding because the splitter needs to cover the whole element. Padding for the primary/secondary content should be set in a nested layout.


MarcinVaadin (approved):
LGTM

MarcinVaadin (approved):

PR #16423, opened 2023-03-27
feat: Enable Vite hot replace for theme style changes

Artur-:
At the moment only works when you do not inject the theme into shadow roots and only in frontend hotdeploy mode


Artur-:
Depends on #15631

flow-server/src/main/resources/plugins/application-theme-plugin/theme-generator.js
sissbruecker, line 1: For me the HMR stops working after around 40 updates, logging an error like this to the browser console:
```
Uncaught (in promise) SyntaxError: JSON.parse: unexpected end of data at line 1 column 167 of the JSON data
    setupWebSocket client.ts:91
    setupWebSocket client.ts:90
    <anonymous> client.ts:68
```

Can reproduce this in Chrome and Firefox. It seems the server is still pushing updates when doing more changes (new logs are being appended to the Java logs), but the client does not process them afterwards.

I basically just added a single rule like this to the CSS, and then changed the value about 40 times:
```css
.foo {
  --foo: bar1111...;
}
```

Artur-: Any idea what the invalid JSON is or who tries to parse it? Sounds like vite-client

Artur-: Sounds like a Vite issue so it should be checked with a ”npm init vite” project

Artur-: It can also be a problem in https://github.com/vaadin/flow/tree/main/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy

Artur-: This is fixed by #16454

Artur-:
Now this should work..

tepi (approved):

PR #16443, opened 2023-03-28
chore: Remove unused imports (#13213)

Artur-:
Using find . -type f -name "*.java"|xargs java -jar google-java-format-1.14.0-all-deps.jar --fix-imports-only --skip-sorting-imports  -r


czp13 (dismissed):
Very good idea. Seems safe, I checked some static import removals, all were good.
🙇

czp13 (approved):

PR #16445, opened 2023-03-28
refactor: import constructable stylesheets polyfill from dev tools

sissbruecker:
## Description

The theme editor injects global styles using constructable stylesheets right when its JS modules are loaded. While we use a polyfill for constructable stylesheets for Safari, the current loading order does not guarantee that the polyfill is loaded before the theme editor modules. Adding an explicit import to the polyfill in the dev tools should fix that.


MarcinVaadin (approved):
Fixes Safari issue.

PR #16446, opened 2023-03-28
refactor: show theme editor only in Flow apps

sissbruecker:
## Description

The theme editor only works with Flow apps and should not be visible in a Hilla app.


MarcinVaadin (approved):
LGTM

PR #16447, opened 2023-03-28
refactor: indicate that theme editor is going to be a commercial feature on stable release

sissbruecker:
## Description

Theme editor is going to be a commercial feature. While it is going to stay behind a feature flag for now and will be freely accessible, it has been decided that it should already indicate that it is not going to be accessible without a subscription in a future stable release. Added a "Free Preview" to the feature flag name, and the dev tools tab title.

Bildschirmfoto 2023-03-28 um 12 08 26

### DIFF
--- a/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
+++ b/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
@@ -58,8 +58,8 @@ public class FeatureFlags implements Serializable {
             "Collaboration Engine backend for clustering support",
             "collaborationEngineBackend",
             "https://github.com/vaadin/platform/issues/1988", true, null);
-    public static final Feature THEME_EDITOR = new Feature("Theme Editor",
-            "themeEditor", null, true, null);
+    public static final Feature THEME_EDITOR = new Feature(
+            "Theme Editor (Free Preview)", "themeEditor", null, true, null);
     private List<Feature> features = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


MarcinVaadin (dismissed):

MarcinVaadin (approved):

knoobie:
@tarekoraby That this a paid feature ;) I was hoping for it to be free for all (not especially for me, because we have a license, but more for the whole Vaadin community where this new feature would really shine if it would be available for everyone and help to get more people on board with how easy development with Vaadin is)

tarekoraby:
👍 thanks for the feedback. You are right that this feature would be valuable for all Flow users: so far, you can use Flow to create web apps without touching HTML and JS. With the Theme Editor, you can also create Flow apps without touching CSS. 

We take your feedback on board 🙌

knoobie:
Thanks for considering! You can go ahead with the current direction - like with e.g. ConfirmDialog or the VCF, things can change over time ;) 

I just had the feeling that feature is too good to be behind some "paywall" and could drastically increase the hardest part (looking at the community) of Vaadin Flow: Styling Applications. 

Not sure if it's worth or even feasible, but you could also consider some freemium model, where e.g. "easy" things are included by default (Margins, Paddings and other simple stuff) and advanced things like state styling (`:hover` or `active`) are paid.

kimweins:
@knoobie We also have additional "low-code" type tools coming soon that will be freely available to all. Stay tuned!

PR #16453, opened 2023-03-28
feat: HMR for changing styles injected into shadow roots

Artur-:


flow-tests/test-live-reload/frontend/component-with-theme.ts
```diff
@@ -16,6 +18,9 @@ export class ComponentWithTheme extends LitElement {
   }
   connectedCallback() {
     super.connectedCallback();
-    applyTheme(this.renderRoot);
+    if (!this.applied) {
+      applyTheme(this.renderRoot);
+      this.applied = true;
+    }
```
caalador, line 24: Does this have an implication that all components should check applyTheme usage so it is only called once?
If so, could applyTheme have book keeping on the applied elements.

Artur-: Maybe this should be in `applyTheme` instead. But has anything changed here? Didn't the old code apply the theme multiple times?

flow-server/src/main/resources/META-INF/frontend/theme-util.js
```diff
@@ -81,9 +88,9 @@ export const injectGlobalCss = (css, referenceComment, target, first) => {
 
   // We avoid mixing style tags and adoptedStyleSheets to make override order clear
   if (target === document) {
-    addStyleTag(cssText, referenceComment);
+    return addStyleTag(cssText, referenceComment);
   } else {
-    addAdoptedStyle(cssText, target, first);
+    return addAdoptedStyle(cssText, target, first);
   }
```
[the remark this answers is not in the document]
Artur-: Fixed

flow-tests/test-live-reload/frontend/component-with-theme.ts
Artur-: I reverted this test change as it does not seem to be needed. If `applyTheme` applies twice when run twice, then it all the rules should also be removed with this approach, as all removers are collected into an array

So maybe it should be fixed but not in this PR, unless something breaks now

caalador (approved):

PR #16455, opened 2023-03-29
feat: HMR for theme styles.css in bundle mode

### LINKED ISSUE #16366: Images from theme folder are not loaded in UI after upgrade to Vaadin 24
### Description of the bug

After upgrading from 23.3.x to Vaadin 24.0.1 my images located in the theme folder are not resolved (loaded) anymore.
This problem is not appearing when  Vite is activated via `frontend.hotdeploy=true`.

Theme folder setup
```
themes
  common
    components
    styles.css
  theme1
    styles.css
    theme.json
    img
      logo.png
```

CSS
```
.mydiv {
  background-image: url("./img/logo.png")
}
```

This is what is looks like:
image

### Expected behavior

Image is being loaded.

This is what it looks like with Vite enabled:
image

[my-app2 2.zip](https://github.com/vaadin/flow/files/11041617/my-app2.2.zip)

### Versions

- Vaadin / Flow version: 24.0.1
- Java version: 17.0.5

Artur-:
Provides hot replacement of CSS when running in development bundle mode.
Inlines imported local CSS so these are also hot replaced.
Rewrites image URLs so they work correctly now that link tags are changed into style tags.

Fixes #16366

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -38,7 +38,6 @@ import java.util.Collections;
 import java.util.EnumMap;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -73,7 +72,6 @@ import com.vaadin.flow.component.page.Viewport;
 import com.vaadin.flow.di.Lookup;
 import com.vaadin.flow.di.ResourceProvider;
 import com.vaadin.flow.function.DeploymentConfiguration;
-import com.vaadin.flow.function.SerializableBiFunction;
 import com.vaadin.flow.internal.AnnotationReader;
 import com.vaadin.flow.internal.BootstrapHandlerHelper;
 import com.vaadin.flow.internal.ReflectTools;
@@ -86,6 +84,7 @@ import com.vaadin.flow.server.communication.AtmospherePushConnection;
 import com.vaadin.flow.server.communication.IndexHtmlRequestHandler;
 import com.vaadin.flow.server.communication.PushConnectionFactory;
 import com.vaadin.flow.server.communication.UidlWriter;
+import com.vaadin.flow.server.frontend.CssBundler;
 import com.vaadin.flow.server.frontend.DevBundleUtils;
 import com.vaadin.flow.server.frontend.FrontendUtils;
 import com.vaadin.flow.server.frontend.ThemeUtils;
@@ -1625,7 +1624,7 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
     protected static Collection<Element> getStylesheetTags(
             AbstractConfiguration config, String fileName) throws IOException {
         return ThemeUtils.getActiveThemes(config).stream()
-                .map(theme -> getLinkTagWithStyleRef(theme, fileName)).toList();
+                .map(theme -> getDevModeStyleTag(theme, fileName)).toList();
     }
 
     /**
@@ -1644,20 +1643,25 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
     protected static Collection<String> getStylesheetLinks(
             AbstractConfiguration config, String fileName) throws IOException {
         return ThemeUtils.getActiveThemes(config).stream()
-                .map(theme -> getThemeFilePath(theme, fileName)).toList();
+                .map(theme -> ThemeUtils.getThemeFilePath(theme, fileName))
+                .toList();
     }
 
-    private static String getThemeFilePath(String themeName, String fileName) {
-        return Constants.VAADIN_MAPPING + Constants.APPLICATION_THEME_ROOT + "/"
-                + themeName + "/" + fileName;
-    }
-
-    private static Element getLinkTagWithStyleRef(String themeName,
+    private static Element getDevModeStyleTag(String themeName,
             String fileName) {
-        Element element = new Element("link");
-        element.attr("rel", "stylesheet");
-        element.attr("type", "text/css");
-        element.attr("href", getThemeFilePath(themeName, fileName));
+        Element element = new Element("style");
+        element.attr("data-dev-id",
+                ThemeUtils.getThemeFilePath(themeName, fileName));
+        File frontendDirectory = new File("frontend");
+        File stylesCss = new File(
+                ThemeUtils.getThemeFolder(frontendDirectory, themeName),
+                fileName);
+        try {
+            element.text(CssBundler.inlineImports(stylesCss));
+        } catch (IOException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
+        }
         return element;
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -1650,7 +1650,7 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
     private static Element getDevModeStyleTag(String themeName,
             String fileName) {
         Element element = new Element("style");
-        element.attr("data-dev-id",
+        element.attr("data-file-path",
                 ThemeUtils.getThemeFilePath(themeName, fileName));
         File frontendDirectory = new File("frontend");
         File stylesCss = new File(
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -1658,7 +1658,8 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
                 ThemeUtils.getThemeFolder(frontendDirectory, themeName),
                 fileName);
         try {
-            element.text(CssBundler.inlineImports(stylesCss));
+            element.text(CssBundler.inlineImports(stylesCss.getParentFile(),
+                    stylesCss));
         } catch (IOException e) {
             throw new RuntimeException(
                     "Unable to read theme file from " + stylesCss, e);
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -1624,7 +1624,8 @@ public class BootstrapHandler extends SynchronizedRequestHandler {
     protected static Collection<Element> getStylesheetTags(
             AbstractConfiguration config, String fileName) throws IOException {
         return ThemeUtils.getActiveThemes(config).stream()
-                .map(theme -> getDevModeStyleTag(theme, fileName)).toList();
+                .map(theme -> getDevModeStyleTag(theme, fileName, config))
+                .toList();
     }
 
     /**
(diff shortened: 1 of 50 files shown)


vaadin-dev-server/frontend/vaadin-dev-tools.ts
```diff
@@ -881,6 +881,15 @@ export class VaadinDevTools extends LitElement {
       window.sessionStorage.setItem(VaadinDevTools.TRIGGERED_KEY_IN_SESSION_STORAGE, 'true');
       window.location.reload();
     };
+    const onUpdate = (path: string) => {
+      let linkTag = document.head.querySelector(`link[href^='${path}?']`);
+      if (linkTag) {
+        this.log(MessageType.INFORMATION, 'Hot update of ' + path);
+        (linkTag as any).href = `${path}?${new Date().getTime()}`;
```
sissbruecker, line 888: In Firefox, stylesheets that are `@import`ed from the main stylesheet are not updated when changing the URL.


https://user-images.githubusercontent.com/357820/229753888-d8845cfc-2f93-4fba-accb-b844ee7a024c.mp4

Chrome and Safari do not have this problem.

sissbruecker: Just noting that the refresh also quite sluggish on my system (around 1 second), and causes some flashing of the styles, at least in Chrome and Firefox.

Artur-: The flashing can probably be fixed by first loading the new styles and then removing the old. Any idea where the sluggishness come from?

Artur-: Ah I see, you should test with #16458. The current file watcher is slow

Artur-: Any other ideas to fix the Firefox issue except rewiting all imports so they also have a query parameter?

sissbruecker: > Ah I see, you should test with https://github.com/vaadin/flow/pull/16458. The current file watcher is slow

Nice, that's a huge improvement 👍

> The flashing can probably be fixed by first loading the new styles and then removing the old.

Sounds good. One thing to keep in mind maybe is that updates can happen very frequently with the theme editor (rapidly clicking undo/redo), so this could result in loading multiple stylesheets in parallel.

> Any other ideas to fix the Firefox issue except rewiting all imports so they also have a query parameter?

The only other thing I can think of is recursively inlining all imported files into `styles.css` and then push that to the dev tool. Dev tool replaces `<link>` with style tag or constructed stylesheet. I'm not sure how difficult inlining the styles would be, there are probably several edge-cases like importing from an URL rather than a file.

As for your suggestion, I'm not sure where we would rewrite the imports - we don't want to change the actual project files?

Artur-: Right now, it feels like we should read and combine styles.css on the server in `DevModeHandlerManagerImpl` when the file watcher is triggered. Instead of sending the path to the dev tools, we should send the combined content directly. Then we can use a `<style>` instead of `<link>` and make everything work + get rid of the flashing

sissbruecker: Yes, that's what I meant. The only tricky part could be resolving imports, or leaving them as is if they reference a URL.

Artur-: Updated to use `<style>` tag and combining the CSS on the server. It still has a few shortcuts that need to be fixed but should now work also in Firefox

sissbruecker: Can confirm it works 👍. Flashing seems to be gone as well.

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.regex.Pattern;
+
+import org.apache.commons.io.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CssBundler {
+
+    private static Pattern urlMatcher = Pattern
+            .compile("@import\\s*(url\\(\\s*)(\\'|\\\")?(\\S*)(\\2\\s*\\));?");
```
sissbruecker, line 16: Wrapping with `url("...")` seems to be optional: https://developer.mozilla.org/en-US/docs/Web/CSS/@import#importing_css_rules

Artur-: Yes, this file definitely needs some tests and then some fixes The syntax is
```
@import =    @import [ <url> | <string> ] [ layer | layer( <layer-name> ) ]? <import-conditions> ;
```

so I guess we should as a minimum handle
```
@import "foo.css"
@import 'foo.css'
import url(foo.css)
import url('foo.css');
@import url("foo.css");
```

Artur-: Now it should be a little bit more generic and even has a test

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1643,94 +1642,26 @@ protected static Collection<Element> getStylesheetTags(
...
+    private static Element getDevModeStyleTag(String themeName,
             String fileName) {
-        Element element = new Element("link");
-        element.attr("rel", "stylesheet");
-        element.attr("type", "text/css");
-        element.attr("href", getThemeFilePath(themeName, fileName));
+        Element element = new Element("style");
+        element.attr("data-dev-id",
+                ThemeUtils.getThemeFilePath(themeName, fileName));
+        File frontendDirectory = new File("frontend");
```
mshabarov, line 1655: This should get the frontend folder from configuration.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1643,94 +1642,26 @@ protected static Collection<Element> getStylesheetTags(
...
-
-    private static Element getLinkTagWithStyleRef(String themeName,
+    private static Element getDevModeStyleTag(String themeName,
             String fileName) {
-        Element element = new Element("link");
-        element.attr("rel", "stylesheet");
-        element.attr("type", "text/css");
-        element.attr("href", getThemeFilePath(themeName, fileName));
+        Element element = new Element("style");
+        element.attr("data-dev-id",
```
mshabarov, line 1653: This can be any key name basically. However `data-dev-id` doesn't say much what is it. Maybe `styles-id` or `styles-path` would be better?

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1643,94 +1642,26 @@ protected static Collection<Element> getStylesheetTags(
...
+                ThemeUtils.getThemeFilePath(themeName, fileName));
+        File frontendDirectory = new File("frontend");
+        File stylesCss = new File(
+                ThemeUtils.getThemeFolder(frontendDirectory, themeName),
+                fileName);
+        try {
+            element.text(CssBundler.inlineImports(stylesCss));
+        } catch (IOException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
```
mshabarov, line 1663: Don't forget to handle it properly: log that the styling fails or throw.

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
```diff
@@ -0,0 +1,88 @@
...
+            File potentialFile = new File(cssFile.getParentFile(), url.trim());
+            if (potentialFile.exists()) {
+                try {
+                    return inlineImports(potentialFile);
+                } catch (IOException e) {
+                    getLogger()
+                            .warn("Unable to inline import: " + result.group());
+                }
+            }
+            return result.group();
```
mshabarov, line 69: The logic behind this method is a bit encrypted for me, need to make another attempt to understand it later.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -278,25 +278,27 @@ private static boolean themeConfigurationChanged(Options options,
...
-
-        JsonObject contentsInStats = statsJson.getObject("themeJsonContents");
-        if (contentsInStats == null && (!themeJsonContents.isEmpty()
+        Optional<String> maybeThemeName = Optional
+                .ofNullable(frontendDependencies.getThemeDefinition())
+                .map(def -> def.getName()).filter(name -> !name.isBlank());
+        Optional<JsonObject> projectThemeJson = maybeThemeName
+                .flatMap(themeName -> ThemeUtils.getThemeJson(
+                        options.getFrontendDirectory(), themeName));
+        String projectThemeName = maybeThemeName.orElse(null);
```
mshabarov, line 287: What happens if `projectThemeName` is null ? :thinking_face:

mshabarov (changes requested):
Tested with Vaadin 24.1 pre starter and HMR works like a charm for me in Chrome.
Some minor comments.

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
Artur-: I would keep `styles` out of it as the `styles` tag already tells that. If we want to avoid conflicts then custom attributes should start with `data-`. Vite uses `data-vite-dev-id`, that's where the current `data-id` comes from. Would `data-file-path` be better?

mshabarov: > Would data-file-path be better?

Yes, would be better +

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
Artur-: That's not even from this commit, it's from the parent :)

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
Artur-: Renamed

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
Artur-: Foxed

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
Artur-: Well that's the only difficult part of this PR :)

I added a comment about the source of the regexp. Otherwise it just tries to find imports and inline them in the CSS if they are in the project

flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
Artur-: Fixed to use correct method

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
Artur-: The tests are the one that are supposed to prove then that it actually does what it promises

Artur-:
Now this also fixes #16366 ...

mshabarov (approved):

PR #16457, opened 2023-03-29
refactor: add open css button to theme editor

sissbruecker:
## Description

Adds a button for opening / editing each element's respective CSS rule in the IDE. Kept the "open CSS" terminology in the code, but thought that "Edit CSS" would be a better label for the button itself. When using local theming, the action also adds a local class name to the component if it doesn't have one yet, so that a proper selector / CSS rule can be generated.

Bildschirmfoto 2023-03-29 um 11 21 42

### DIFF
--- a/vaadin-dev-server/frontend/vaadin-dev-tools.test.ts
+++ b/vaadin-dev-server/frontend/vaadin-dev-tools.test.ts
@@ -12,6 +12,10 @@ describe('vaadin-dev-tools', function () {
     // @ts-ignore
     window.Vaadin = {};
     // @ts-ignore
+    window.Vaadin.Flow = {};
+    // @ts-ignore
+    window.Vaadin.Flow.clients = {};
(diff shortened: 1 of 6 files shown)


vaadin-dev-server/frontend/vaadin-dev-tools.test.ts
```diff
@@ -12,6 +12,10 @@ describe('vaadin-dev-tools', function () {
     // @ts-ignore
     window.Vaadin = {};
     // @ts-ignore
+    window.Vaadin.Flow = {};
+    // @ts-ignore
+    window.Vaadin.Flow.clients = {};
+    // @ts-ignore
```
sissbruecker, line 18: Unrelated test fix that I missed in previous PRs, same with `package.json`

MarcinVaadin (approved):
Tested and working

PR #16458, opened 2023-03-29
chore: Change to io.methvin.directory-watcher which has native Mac support

Artur-:
.


Artur-:
Would need a new version with https://github.com/gmethvin/directory-watcher/pull/93 but we can exclude that jna for now to avoid conflicts with oshi

mshabarov (approved):

oryan-block:
Hello guys, this move to directory-watcher fails for me on an M2 Macbook. Error is:
>fat file, but missing compatible architecture (have 'i386,x86_64', need 'arm64')

This seems to be caused because of the JNA version this library is using:
https://github.com/gmethvin/directory-watcher/blob/main/core/src/main/java/io/methvin/watchservice/jna/CarbonAPI.java#L19-L20
which is 4.5.0 but should be 5.7.0 according to this:
https://groups.google.com/g/jna-users/c/UwUa5qWbIco

Is there a solution for this?

oryan-block:
@Artur- thanks. It seems to be conflicting with elasticsearch.

PR #16463, opened 2023-03-29
test: added suffix for test reports

mcollovati:
Configured maven to add a suffix for test reports in modules that have multiple pom files, to prevent results to be overwritten during test-results validation job.


czp13 (approved):

mcollovati:
Compared with another PR build, as expected, the number of tests in Test Results is increased with this change

image

mcollovati:
> We have some tests that are included as dependencies and thus also run in multiple modules with the same class name. Do they show up separately already as the folder is different?

The test summary in the compressed files preserves their relative path to the project root. So, if the same test class runs in different modules should result in different report files in each module

PR #16465, opened 2023-03-30
refactor: Use config.createResolver to be compatible with Vite 4.3+

Artur-:


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -126,16 +126,23 @@ function buildSWPlugin(opts): PluginOption {
...
-      'vite:terser',
-    ]
+      'vite:terser'
+    ];
     const plugins: rollup.Plugin[] = config.plugins.filter((p) => {
-      return includedPluginNames.includes(p.name)
+      return includedPluginNames.includes(p.name);
     });
+    const resolver = config.createResolver();
+    const resolvePlugin:rollup.Plugin = {
```
vursen, line 138:
```suggestion
    const resolvePlugin: rollup.Plugin = {
```

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -126,16 +126,23 @@ function buildSWPlugin(opts): PluginOption {
...
+      name: 'resolver',
+      resolveId(source, importer, _options) {
+        const result = resolver(source, importer);
+        return result;
```
vursen, line 142:
```suggestion
        return resolver(source, importer);
```

vursen:
**As a side note:** We might want to try out building the service worker with ESBuild instead of Rollup: https://vitejs.dev/guide/api-javascript.html#transformwithesbuild

Artur-:
> **As a side note:** We might want to try building the service worker with ESBuild instead of Rollup: https://vitejs.dev/guide/api-javascript.html#transformwithesbuild

Feel free, but let's fix this issue first :)

vursen (approved):

vursen:
It seems that I tested the wrong version. Ignore my approval, please.

vursen (approved):
I tested it locally and it works. Approved.

PR #16468, opened 2023-03-30
refactor: add RadioButtonGroup metadata

sissbruecker:
## Description

Adds metadata for RadioButtonGroup. Mostly just adapted from the metadata for Checkbox and CheckboxGroup.

This PR is based on the changes from https://github.com/vaadin/flow/pull/16462 and is currently set to merge against it.

### DIFF
--- a/vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-radio-group.ts
+++ b/vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-radio-group.ts
@@ -1,0 +1,97 @@
+import { ComponentMetadata, EditorType } from '../model';
+import { iconProperties, shapeProperties, textProperties } from './defaults';
+import { errorMessageProperties, helperTextProperties, labelProperties } from './vaadin-text-field';
+import { presets } from './presets';
+
+export default {
+  tagName: 'vaadin-radio-group',
+  displayName: 'RadioButtonGroup',
+  elements: [
+    {
+      selector: 'vaadin-radio-group',
+      displayName: 'Group',
+      properties: [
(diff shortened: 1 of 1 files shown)


vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-radio-group.ts
```diff
@@ -0,0 +1,97 @@
...
+      displayName: 'Radio buttons (when checked)',
+      properties: [
+        shapeProperties.backgroundColor,
+        shapeProperties.borderColor,
+        shapeProperties.borderWidth,
+      ]
+    },
+    {
+      selector: 'vaadin-radio-group vaadin-radio-button::part(radio)::after',
+      displayName: 'Radio dots',
```
jouni, line 69: I feel this shouldn't be plural, as there’s always just one radio button selected at any given time. My first thought is something like “Selection indicator”, but that doesn't sound very descriptive…

jouni (commented):
As this is very similar to #16462, I think this is good to go once we agree on that.

vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-radio-group.ts
sissbruecker: Yes, makes sense. "Checked indicator" maybe? Let's sleep on it.

MarcinVaadin (approved):
Tested and working, "Indicator" name seems short and simple.

sissbruecker:
Updated metadata to use the `--vaadin-radio-button-size` property.

MarcinVaadin (approved):

PR #16471, opened 2023-03-30
refactor: add clear property button

sissbruecker:
## Description

Adds a button to property editors that allows clearing customized property values. The button will only be shown if the property has been previously customized by the theme editor.

Bildschirmfoto 2023-03-30 um 16 20 55

Note that until HMR for CSS is implemented, the component will still be styled with the modified value if it was present in the CSS file when the page was loaded.

### DIFF
--- a/vaadin-dev-server/frontend/theme-editor/model.ts
+++ b/vaadin-dev-server/frontend/theme-editor/model.ts
@@ -61,6 +61,12 @@ export class ComponentTheme {
   }
 
   public updatePropertyValue(elementSelector: string, propertyName: string, value: string, modified?: boolean) {
+    // Remove property when value is an empty string
+    if (!value) {
+      delete this._properties[propertyKey(elementSelector, propertyName)];
+      return;
+    }
+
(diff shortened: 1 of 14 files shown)


vaadin-dev-server/frontend/theme-editor/model.ts
```diff
@@ -154,8 +160,12 @@ export function generateThemeRule(
...
 
   // Enable border style when setting a border width
-  if (propertyName === 'border-width' && parseInt(value) > 0) {
-    properties['border-style'] = 'solid';
+  if (propertyName === 'border-width') {
+    if (parseInt(value) > 0) {
+      properties['border-style'] = 'solid';
+    } else {
+      properties['border-style'] = '';
+    }
```
sissbruecker, line 168: Added clearing of the border-style as well.

MarcinVaadin (approved):
Accepted, let's get back to it after CSS hot reload will be merged.

jouni (approved):

PR #16474, opened 2023-03-30
fix: Do not downgrade es-version

### LINKED ISSUE #16305: Do not enforce target change in tsconfig.json
### Description of the bug

I want to customize `tsconfig.json` in the application to have a newer target setting  `es2021`. But after I edit the file, when starting the dev server, the target is changed back to `es2020`.

### Expected behavior

`tsconfig.json` is not edited if the `flow_version` key matches the recent one in Flow.

See https://github.com/vaadin/hilla-crm-tutorial

- Put  `"target": "es2021"` in `tsconfig.json`
- run `mvn`

### Versions

- Vaadin / Flow version: 24.0.0 (Hilla 2.0.0)
- Java version:
- OS version:
- IDE (if applicable):

Artur-:
Fixes #16305

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
@@ -111,7 +111,7 @@ public class TaskGenerateTsConfig extends AbstractTaskClientGenerator {
             String current = FileUtils.readFileToString(projectTsconfig,
                     StandardCharsets.UTF_8);
             String currentEsVersion = getEsTargetVersion(current);
-            if (!currentEsVersion.equals(esVersion)) {
+            if (isOlder(currentEsVersion, esVersion)) {
                 current = current.replace(currentEsVersion, esVersion);
                 writeIfChanged(projectTsconfig, current);
             }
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
```diff
@@ -111,7 +111,7 @@ private void ensureTarget(String esVersion) {
             String current = FileUtils.readFileToString(projectTsconfig,
                     StandardCharsets.UTF_8);
             String currentEsVersion = getEsTargetVersion(current);
-            if (!currentEsVersion.equals(esVersion)) {
```
czp13, line 114: I would rewrite:
-  the condition to make it more clear to read and know when we step in,
- plus would make private the methods,
- plus would use currentVersion, oldVersion or just current, old naming,

Tiny things as this will only be used here, 

```Java
if(  (isEsPrefix(current, old) && isOlder(current, old)) ||
      !isEsPrefix(current,old)  && !current.equals(old)) {
      ...


private static boolean isEsPrefix{ return (esVersion1.startsWith("es") && esVersion2.startsWith("es")); } 

private static boolean isOlder{ return esVersion1.compareTo(esVersion2) < 0; }
```

platosha (commented):
Seems alright, but are there any reasons to keep this feature now that the entire `tsconfig.json` is auto-updated? I was hoping we could remove it and simplify the editing behavior...

Artur-:
From a users point of view, this is what we should have and the "update the whole file" mechanism is flawed... I think changing the mechanism is a separate discussion

Artur-:
It would be much easier to understand if we instead had some specific requirements for some specific values, like 
* "es version must be at least 2020"
* "the folder foo should be excluded"

Then we could also auto-update the file in a non-disruptive way like we do today. It is not really anybody who wants us to overwrite the whole file and then say "oops, you had some customizations. Well, you can re-do them"

platosha (approved):

PR #16477, opened 2023-03-31
refactor: Use dnd from npm instead of as a webjar

### LINKED ISSUE #12315: [Grid]-Drag and Drop - Not working when application is deployed on a different context than Root context
### Description

In a Drag and Drop enabled Grid, using Grid#setRowsDraggable(true):

- if application is deployed under Root context ( http://127.0.0.1:8080 ), then Drag and Drop works both on a Desktop browser and also on a mobile device
- if application is deployed under a different context than Root ( http://127.0.0.1:8080/dnd ), then Drag and Drop works on a Desktop browser, but it is not working on a mobile device, and it fails with the following error:

Screenshot 2021-11-09 at 08 01 38

### Expected outcome

Drag and Drop among the Grid rows should work both on Desktop and Mobile, not matter if the application is deployed in the Root context or a different one.

### Actual outcome

Drag and Drop is not working on a mobile device, when the application is deployed on a different context than Root and if tails with the following error:
Screenshot 2021-11-09 at 08 03 28

The issue can be reproduced from a Vaadin 14.7.4 starter.
Please see archive attached: 
[vaadin14-dnd-mobile.zip](https://github.com/vaadin/web-components/files/7502807/vaadin14-dnd-mobile.zip)

### Steps to reproduce

1. Start the application: mvn jetty:run .
2. Open the Google Chrome Desktop browser along with the Developer Tools and set the Device to, for example "iPad Pro" (like in the screenshot). It is important to do this step first, in order to simulate a mobile device.
3. Access the application at: http://127.0.0.1:8080/dnd
4. You should obtain the error described above.

### Environment

- Vaadin 14.7.4
- Jetty: 9.4.42.v20210604 with web context set to "/dnd" => http://127.0.0.1:8080/dnd
- Google Chrome set to simulate an "iPad Pro"

### Notes
Please let me know if you need additional info
Thank you.

Artur-:
There is a version conflict between the webjars used and the vaadin-mobile-drag-drop package is archived. It will be easier to use the same mechanism for the mobile-polyfill as for all other npm dependencies

Fixes #12315


Artur-:
It fixes #12315, I used that as a test case

caalador (approved):

PR #16480, opened 2023-03-31
fix: Make @Push inherited

### LINKED ISSUE #16433: Honor @Push annotations on superclasses implementing AppShellConfigurator
### Describe your motivation

We have several applications that use the same look & feel and other resources. So it makes sense to use a common abstract superclass for the AppShellConfigurator. This works great for overridden methods, e.g. sharing the same favicon in each app via configurePage(), and the @Theme annotation. But this does not work for the @Push annotation, as it does not have the @Inherited meta annotation.

The @Push annotation should receive the @Inherited annotation that @Theme already uses, if possible.

Currently we apply the @Push annotation to the concrete class in each application.

### Additional context

Nothing I can think of right now.

Artur-:
Allows you to use a super class for configuration

Fixes #16433

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/Push.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/Push.java
@@ -17,6 +17,7 @@
 package com.vaadin.flow.component.page;
 
 import java.lang.annotation.ElementType;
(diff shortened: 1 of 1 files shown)


czp13 (approved):
I wanted to do this as well, seems safe.

PR #16482, opened 2023-03-31
fix: Remove /VAADIN/dev-bundle/ from bundle url

### LINKED ISSUE #16210: Extra /VAADIN/dev-bundle in URL when using a dev bundle
### Description of the bug

When running a bundle build, Vite is configured with
```
base: ""
assetsDir: 'VAADIN/build',
outDir: '/project/src/main/dev-bundle/webapp'
```

If I understand correctly, this means that you can assume that an asset that is put into `src/main/dev-bundle/webapp/VAADIN/build/foo.png` is made available at `base+assetsDir` so `VAADIN/build/foo.png` in the browser.

This is not true as Flow adds an additional `/VAADIN/dev-bundle` in front of the URL and only serves the dev bundle assets when that is present. 

### Expected behavior

If you add `foo.png` to the `assetsDir` (`src/main/dev-bundle/webapp/foo.png` when in bundle mode) you can load it using `VAADIN/build/foo.png` both when in development mode (Vite or dev bundle) and in production mode.

In a project with a custom bundle, do
```
echo hello > src/main/dev-bundle/webapp/VAADIN/build/foo.txt
```
The file is not available at http://localhost:8839/VAADIN/build/foo.txt but at http://localhost:8839/VAADIN/dev-bundle/VAADIN/build/foo.txt

In the same project, run it with `-Pproduction` and do
```
echo hello > target/classes/META-INF/VAADIN/webapp/VAADIN/build/foo.txt   
```
The file is now available at http://localhost:8839/VAADIN/build/foo.txt and not at http://localhost:8839/VAADIN/dev-bundle/VAADIN/build/foo.txt

### Versions

- Vaadin / Flow version: 24.0.0
- Java version:17

Artur-:
Fixes #16210


flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -260,15 +259,14 @@ public boolean serveStaticResource(HttpServletRequest request,
...
+            if (!"/index.html".equals(filenameWithPath)) {
                 resourceUrl = FrontendUtils.findBundleFile(
                         deploymentConfiguration.getProjectFolder(),
-                        "webapp" + filenameInsideBundle);
-            } else if (APP_THEME_PATTERN.matcher(filenameWithPath).find()
+                        "webapp" + filenameWithPath);
+            }
+
+            if (resourceUrl == null
+                    && APP_THEME_PATTERN.matcher(filenameWithPath).find()
```
mshabarov, line 269: Conditional expression should be
```
if (resourceUrl == null
                    && (APP_THEME_PATTERN.matcher(filenameWithPath).find()
                    || APP_THEME_ASSETS_PATTERN.matcher(filenameWithPath)
                            .find()))
```
right ? (fall into this block only if resource URL is null)

Artur-: right, the joy of || and &&. Which goes first again? :)

Artur-: Added `()`

mshabarov: Logical AND has a higher precedence over Logical OR, if I haven't lost my Java expertise yet :)

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
Artur-: So you mean (a && b || c) === (a && (b || c)) and the code was correct?

Artur-: Or the other way around. Well it does not matter because nobody who reads the code will know without `()`

mshabarov: > So you mean (a && b || c) === (a && (b || c)) and the code was correct?
No, I meant the opposite. These two expressions aren't equal. In our case B and C should be checked if A is false. This is achieved when you have brackets around B || C.

Artur-: I so want to make a JDK PR to make it a compile error instead..

PR #16497, opened 2023-03-31
chore: Remove SERVER_ROUTING

Artur-:
I don't think we have something like SERVER_ROUTING anymore, whatever it is


flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
```diff
@@ -69,15 +69,6 @@ public ApplicationConnection(
...
         Binder.bind(rootNode, body);
 
-        // When app is run as a WC do not add listener for routing events.
-        // Routing is responsability of the hosting application (#6108)
-        if (!applicationConfiguration.isWebComponentMode()
-                // client router is responsible of routing events
-                && !applicationConfiguration.isClientRouting()) {
-            new PopStateHandler(registry).bind();
-            RouterLinkHandler.bind(registry, body);
-        }
```
caalador, line 79: Shouldn't this only remove the !isClientRouting part and leave the rest as no router should be set for a web-component

caalador (changes requested):

flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
Artur-: !isClient == isServer, which is what is removed here?

Artur-: and && then means `&& false` ?

Artur-: I am not sure "this" is here. Should a web component have any route link handler or pop state handler? I though they are supposed to ignore everything related to routing and the URL

caalador: Wow that was the most confusing if.
So PopStateHandler and RouterLinkHandler could be removed completely as they are not used anymore as we do not have server-side-bootstrap.

Artur-: that's my understanding

Artur-: if they are not used elsewhere

Artur-: Removed

flow-client/src/main/java/com/vaadin/client/DefaultRegistry.java
```diff
@@ -117,15 +117,6 @@ public DefaultRegistry(ApplicationConnection connection,
         set(XhrConnection.class, new XhrConnection(this));
         set(PushConfiguration.class, new PushConfiguration(this));
         set(ReconnectConfiguration.class, new ReconnectConfiguration(this));
-        if (!applicationConfiguration.isClientRouting()) {
-            if (applicationConfiguration.isWebComponentMode()) {
-                set(ScrollPositionHandler.class,
```
caalador, line 122: The ScrollPositionHandler seems to not be in use either with client routing.
This could also be cleaned away. here or in its own PR if you wish, but I feel we shouldn't leave it to clutter the code base.

caalador (approved):

Artur-:
Seems like `GwtAtmoshperePushConnectionTest` has some global side effects as moving it last in the suite resolves the problem in `GwtDefaultConnectionStateHandlerTest` ...

caalador (approved):

PR #16498, opened 2023-04-01
fix: Initialize the route only once when using eager server load

### LINKED ISSUE #16046: HasUrlParameter.setParameter is executed twice with vaadin.eagerServerLoad=true
### Description of the bug

I have a View which  `implements HasUrlParameter<String>`. 

Also I configured:

`vaadin.eagerServerLoad=true` (for IndexHtmlRequestListener)

with such setup, `HasUrlParameter.setParameter()` method is executed twice during the first page load

### Expected behavior

`HasUrlParameter.setParameter()` method is executed only one time during the first page load with `vaadin.eagerServerLoad=true`

### Versions

- Vaadin / Flow version: 23.3.6

### LINKED ISSUE #16496: Eager server load does not show the eager version
### Description of the bug

When you enable eager server load using `vaadin.eagerServerLoad=true` then the UI is created during the index page request and the initial UIDL is sent to the browser. In the browser, the UIDL is processed, the state tree is updated and the DOM elements are created. Unfortunately, the DOM elements are not created inside the designated "outlet" div but appended to the body tag. This means they are rendered right below the viewport so nobody can see them. It works however, because after this an xhr is sent to the server (the same as when not using eager mode) and this xhr triggers navigation again, causing #16046 and probably other issues. The result of the second navigation is sent back to the client, and now the elements rendered outside the viewport are removed and the second view instance is rendered correctly.

### Expected behavior

When you enable eager server load, there is only one server request and the view is rendered so you can see it.

See `flow-tests/test-eager-bootstrap`

### Versions

- Vaadin / Flow version: main

Artur-:
Changes bootstrap for eager mode so that the index page contains the <flow-container-[appId]> div and the virtual child for this element is created eagerly on the server. This allows the initial render on the client side to put the content inside the flow-container instead of document body as happened before.
    
By using a convention on how the container div is created and id is set, we no longer need to communicate the element tag and id from the client.
    
Fixes #16046
Fixes #16496

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -117,9 +117,6 @@ public class UI extends Component
 
     private static final String NULL_LISTENER = "Listener can not be 'null'";
 
-    private static final Pattern APP_ID_REPLACE_PATTERN = Pattern
-            .compile("-\\d+$");
-
     /**
      * The id of this UI, used to find the server side instance of the UI form
      * which a request originates. A negative value indicates that the UI id has
@@ -262,8 +259,15 @@ public class UI extends Component
         }
         this.uiId = uiId;
 
-        appId = APP_ID_REPLACE_PATTERN.matcher(appId).replaceAll("");
-        getInternals().setAppId(appId);
+        getInternals().setFullAppId(appId);
+
+        // Create flow reference for the client outlet element
+        wrapperElement = new Element(getInternals().getContainerTag());
+
+        // Connect server with client
+        getElement().getStateProvider().appendVirtualChild(
+                getElement().getNode(), wrapperElement,
+                NodeProperties.INJECT_BY_ID, appId);
 
         // Add any dependencies from the UI class
         getInternals().addComponentDependencies(getClass());
@@ -1637,6 +1641,8 @@ public class UI extends Component
 
     private String forwardToClientUrl = null;
 
+    private boolean firstNavigation = true;
+
     /**
      * Gets the new forward url.
      *
@@ -1650,10 +1656,6 @@ public class UI extends Component
      * Connect a client with the server side UI. This method is invoked each
      * time client router navigates to a server route.
      *
-     * @param clientElementTag
-     *            client side element tag
-     * @param clientElementId
-     *            client side element id
      * @param flowRoutePath
      *            flow route path that should be attached to the client element
      * @param flowRouteQuery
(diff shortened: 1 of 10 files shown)


Artur-:
This depends on #16497

flow-tests/test-eager-bootstrap/src/test/java/com/vaadin/flow/eagerbootstrap/ParameterIT.java
```diff
@@ -0,0 +1,49 @@
...
+        return "/parameter";
+    }
+
+    @Test
+    public void viewInitializedOnce() {
+        openWithParameter("foo");
+        int firstInstance = getInstance();
+        openWithParameter("foo");
+        Assert.assertEquals(firstInstance + 1, getInstance());
+    }
```
mcollovati, line 21: This test passes also without the fix if it runs isolated.
Is there another way to ensure that the view is not initialized twice?

mcollovati (changes requested):

flow-tests/test-eager-bootstrap/src/test/java/com/vaadin/flow/eagerbootstrap/ParameterIT.java
Artur-: Right, the real test is `setParameterCalledAsExpected` and apparently the same view instance is reused even though navigation happens twice. Maybe we can just delete this test as `setParameterCalledAsExpected` ensures that `setParameter` is only called once

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1692,21 +1693,17 @@ public void connectClient(String clientElementTag, String clientElementId,
...
                     event -> renderViewForRoute(event.getLocation(),
                             event.getTrigger()));
 
-            // Render the flow view that the user wants to navigate to.
-            renderViewForRoute(location, navigationTrigger);
+            if (!getInternals().getActiveRouterTargetsChain().isEmpty()) {
+                // The view has already been rendered eagerly
+            } else {
+                renderViewForRoute(location, navigationTrigger);
+            }
```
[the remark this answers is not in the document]
Artur-: Fixed

flow-tests/test-eager-bootstrap/src/test/java/com/vaadin/flow/eagerbootstrap/ParameterIT.java
Artur-: Removed and simplified

mcollovati (approved):

PR #16499, opened 2023-04-02
Revert "chore(deps): bump javaparser-core from 3.24.10 to 3.25.2"

Artur-:
Reverts vaadin/flow#16485

we need to match what poi uses


czp13 (approved):
Ok, sorry 😥. (Apache POI?)

Issue #13437, opened 2022-04-06
lang attribute of <html> tag ignores value of default locale and defaults to "en"


sephiroth-j:
### Description of the bug

The generated app shell contains `<html lang="en">` even if `Locale.getDefault()` returns "de_DE". This is different from Vaadin 14. No `I18NProvider` is used.

But even if one is used (such as [this example](https://github.com/vaadin/flow-spring-examples/blob/v23/src/main/java/org/vaadin/spring/tutorial/SimpleI18NProvider.java)), the value does not change.

### Expected behavior

The lang attribute of the <html> tag should follow the value of the default locale `Locale.getDefault()`.

### Minimal reproducible example

Run the application with `LANG="de_DE.utf8"`.

### Versions

- Vaadin / Flow version: 23.0.3
- Java version: 17.0.1
- OS version: Windows 10, Linux with LANG = de_DE

sephiroth-j:
### Workaround (Spring Based)

Based on https://vaadin.com/docs/latest/flow/advanced/modifying-the-bootstrap-page/#IndexHtmlRequestListener-interface and https://vaadin.com/docs/latest/flow/advanced/service-init-listener  a possible workaround could look like this.

```java
@Bean
VaadinServiceInitListener vaadinServiceInitListener() {
	return event -> {
		event.addIndexHtmlRequestListener(indexHtmlResponse -> {
			// workaround for https://github.com/vaadin/flow/issues/13437
			indexHtmlResponse.getDocument().head().parent().attr("lang", Locale.getDefault().getLanguage());
		});
	};
}
```

HermanBovens:
Note that the current "workaround" uses the system locale, which is usually not what you want in a multilingual app.  The actual language will need to come from the stored user settings, which you hopefully have access to in the service init listener.

mshabarov:
The issue was triaged and currently added to the backlog priority queue for further investigation.

czp13:
Thank you for reporting it, and for your patience. 

The relevant [fix](https://github.com/vaadin/flow/pull/16410) was developed, tested, added covering tests, and merged to the master, so the next Vaadin 24 release will contain it.

It is using the same logic as in the Component class for consistency within the framework:

0. It is checking if the language attribute is already added to the HTML tag, this was needed to not have any regression for customers having some solution in place.  If not, then it will try to add it based on the following rules:
1. trying to get the locale from the UI
2. if there is an I18N provider then the first locale will be used,
[...]

PR #16502, opened 2023-04-02
fix/feature/optimization: use lang attribute in the html tag (#16410) (CP:24.0) (may not need)

czp13:
## Description

**Cherry-picked from https://github.com/vaadin/flow/pull/16410 to 24.0**

(introducing language to the HTML)
```html
<html theme=""> ...</html>
```
to:
```html
<html lang="en" theme="">
 ...</html>
```

The actual rule/way of working is quite similar to how the Component getLocale() is working:
1. trying to get the locale from the UI
2. if there is an I18N provider then the first locale will be used
3. last option is to fallback to the Locale.getDefault()

Now it is added to the `IndexHtmlRequestHandler` class

Pros:
+ it will be calculated for all requests
+ so a dynamic new I18N provider will be 
Cos:
- less performant

It could be added to other places to only do this once. The benefit would be to be more performant, but also it would not be dynamic.

I have tested it locally and it is adding the attribute for all requests:
image

Fixes # (issue)
https://github.com/vaadin/flow/issues/13437

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have performed a self-review and corrected misspellings.

- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria was created.


czp13:
Tested locally manually and works.

PR #16504, opened 2023-04-02
chore: Remove mentions of express mode

Artur-:


czp13 (dismissed):
Nice, safe to merge. Just noting here that:
- LitTemplateParserImpl
and:
- NpmTemplateParser

Are as well containing the following lines:
```
            // In express mode, template sources are stored in
```

Artur-:
Thanks, seems like I searched only inside `flow-server`. Fixed those also

czp13 (approved):

PR #16521, opened 2023-04-03
fix: Use default XOR CSRF token mechanism in Spring

### LINKED ISSUE #14923: Support XorCsrfTokenRequestAttributeHandler
### Describe your motivation

Spring Security 6 switched to use `XorCsrfTokenRequestAttributeHandler` for the CSRF token by default in https://github.com/spring-projects/spring-security/issues/11960. This is not supported by Hilla endpoints right now. The login handler returns the `CsrfTokenRequestAttributeHandler` token instead and the next endpoint request then fails.

Endpoints and Hilla should work with both `CsrfTokenRequestAttributeHandler` and `XorCsrfTokenRequestAttributeHandler`

Artur-:
The stateless configuration is copied from https://docs.spring.io/spring-security/reference/5.8/migration/servlet/exploits.html#_i_am_using_angularjs_or_another_javascript_framework

Fixes #14923

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/VaadinStatelessSecurityConfigurer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/VaadinStatelessSecurityConfigurer.java
@@ -34,7 +34,9 @@ import org.springframework.security.oauth2.jose.jws.MacAlgorithm;
 import org.springframework.security.web.context.SecurityContextRepository;
 import org.springframework.security.web.csrf.CookieCsrfTokenRepository;
 import org.springframework.security.web.csrf.CsrfTokenRepository;
+import org.springframework.security.web.csrf.CsrfTokenRequestHandler;
 import org.springframework.security.web.csrf.LazyCsrfTokenRepository;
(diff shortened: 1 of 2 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/VaadinStatelessSecurityConfigurer.java
```diff
@@ -98,7 +100,13 @@ public void init(H http) {
             // session (double-submit cookie pattern)
             CsrfTokenRepository csrfTokenRepository = CookieCsrfTokenRepository
                     .withHttpOnlyFalse();
+            XorCsrfTokenRequestAttributeHandler delegate = new XorCsrfTokenRequestAttributeHandler();
```
platosha, line 103: Would be nice to have a comment explaining why the handler is customized like this. Maybe linking this: https://docs.spring.io/spring-security/reference/5.8/migration/servlet/exploits.html#_i_am_using_angularjs_or_another_javascript_framework

Artur-: Added

platosha (approved):

PR #16533, opened 2023-04-05
refactor: Split dev bundle related utils to a separate class

Artur-:


czp13 (approved):
This seems safe to me, checked out, imports good, etc. 
Just refactoring, with a small optimization /fix around logging.

czp13:
I can fix the tests later, as they are because of refactoring (only unit tests, because how we mock could be changed, logging is changed, etc)

Artur-:
I think #16537 will solve most of the issues

czp13 (approved):
This seems safe to me, checked out, imports good, etc. 
Just refactoring, with a small optimization /fix around logging.

PR #16539, opened 2023-04-05
refactor: show notice if styling component is not supported

sissbruecker:
## Description

Shows a notice if a component is not supported by the theme editor. This should currently only affect Vaadin components that do not have metadata.

Bildschirmfoto 2023-04-05 um 12 41 16


MarcinVaadin (approved):
Tested and working

PR #16541, opened 2023-04-05
refactor: fix for picking elements that have inaccessible source file

MarcinVaadin:
## Description

While picking component that has inaccessible source files (like `<body>`) nasty exception has been thrown. Generic component inaccessible message is now presented to the user.

image


sissbruecker:
I'm having a similar problem with Joel's sandbox app, where the Java parser chokes on Java files that contain something this:
```java
if (component instanceof HasLabel hasLabel) {
}
```
That results in:
```
Caused by: com.github.javaparser.ParseProblemException: (line 73,col 13) Use of patterns with instanceof is not supported.
```
which is not covered by the exception handling. 

We could extend the catch clause to catch everything and then maybe log the underlying reason with a low log level, and/or adapt the notice in the frontend to state that the Java file could not be parsed.

sissbruecker (approved):

PR #16555, opened 2023-04-06
refactor: added UsageStatistics for ThemeEditor

MarcinVaadin:
## Description

Added UsageStatistics `flow/ThemeEditor` to Theme Editor.

### DIFF
--- a/vaadin-dev-server/frontend/theme-editor/editor.ts
+++ b/vaadin-dev-server/frontend/theme-editor/editor.ts
@@ -28,7 +28,6 @@ import { injectGlobalCss } from './styles';
 import { ComponentMetadata } from './metadata/model';
 import { ClassNameChangeEvent } from './components/class-name-editor';
 import { OpenCssEvent } from './components/property-list';
-import { ServerInfo } from '../vaadin-dev-tools';
 
 injectGlobalCss(css`
   .vaadin-theme-editor-highlight {
@@ -49,8 +48,6 @@ export class ThemeEditor extends LitElement {
   public connection!: Connection;
   private api!: ThemeEditorApi;
   private history!: ThemeEditorHistory;
-  @property({})
-  public serverInfo!: ServerInfo;
   @state()
   private historyActions?: ThemeEditorHistoryActions;
 
@@ -196,6 +193,7 @@ export class ThemeEditor extends LitElement {
     this.api = new ThemeEditorApi(this.connection);
     this.history = new ThemeEditorHistory(this.api);
     this.historyActions = this.history.allowedActions;
+    this.api.markAsUsed();
   }
 
   protected update(changedProperties: PropertyValues) {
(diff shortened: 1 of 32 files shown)


vaadin-dev-server/frontend/theme-editor/editor.ts
```diff
@@ -28,6 +28,7 @@ import { injectGlobalCss } from './styles';
 import { ComponentMetadata } from './metadata/model';
 import { ClassNameChangeEvent } from './components/class-name-editor';
 import { OpenCssEvent } from './components/property-list';
+import { ServerInfo } from '../vaadin-dev-tools';
```
sissbruecker, line 31: This creates a circular dependency between `editor.ts` and `vaadin-dev-tools.ts`, which itself imports `editor.ts`. Might not be an issue here, since it only imports a type for now, but should be avoided in general. Would be better to extract the interface into a separate module and import that in both dev tools and editor. Or check if the Flow version is available through some global via `window.Vaadin.Flow`, then passing the server info as property is not necessary.

vaadin-dev-server/frontend/theme-editor/editor.ts
```diff
@@ -599,4 +607,14 @@ export class ThemeEditor extends LitElement {
       element.classList.remove('vaadin-theme-editor-highlight');
     }
   }
+
+  private pushStatistics() {
+    const registrations = window.Vaadin.registrations;
+    if (registrations !== undefined && this.serverInfo !== undefined) {
+      const registrationId = 'flow/ThemeEditor';
+      if (registrations.find((p) => p.is == registrationId) == undefined) {
+        registrations.push({ is: registrationId, version: this.serverInfo.flowVersion });
```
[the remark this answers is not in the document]
Artur-: Why do you build your own tracker in TS when you could send a message to Java (or use an existing one) and call `markAsUsed`?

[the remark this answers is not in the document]
MarcinVaadin: TBH using `UsageStatistics.markAsUsed` was my initial idea, approach updated

Artur- (changes requested):
What's with all the unrelated changes?

MarcinVaadin:
> What's with all the unrelated changes?

I've run 'npm prettier' same as I run maven formatter. It was mentioned in readme.

vaadin-dev-server/frontend/vaadin-dev-tools.ts
```diff
@@ -14,7 +14,7 @@ import { licenseCheckFailed, licenseInit, Product } from './License';
 import { Connection, ConnectionStatus } from './connection';
 import { popupStyles } from './styles';
 
-interface ServerInfo {
+export interface ServerInfo {
```
Artur-, line 17: Should this be here?

Artur- (dismissed):

vaadin-dev-server/frontend/vaadin-dev-tools.ts
MarcinVaadin, line 17:
```suggestion
interface ServerInfo {
```

MarcinVaadin: No, leftover. Removed.

Artur- (approved):

PR #16476, opened 2023-03-31
chore: enable dependencyConvergence rule for flow project

ZheSun88:

PR #16564, opened 2023-04-10
chore: apply dependencyConvergence rule

taefi:
Enabling the enforcer plugin's dependencyConvergence 
rule will prevent having multiple versions of transitive 
dependencies during the builds and tests, which can 
improve the quality of tests and reveals hidden bugs 
related to dependency misconfiguring.

This is an alternative PR for #16476


flow-test-util/pom.xml
```diff
@@ -36,7 +39,106 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-codec-http</artifactId>
+            <version>${netti.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${guava.version}</version>
```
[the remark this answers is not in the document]
taefi: No patch is available yet, so 
@sonatype-lift ignore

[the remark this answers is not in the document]
taefi: @sonatype-lift ignore

flow-test-util/pom.xml
```diff
@@ -36,7 +39,106 @@
...
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-handler</artifactId>
+            <version>${netti.version}</version>
```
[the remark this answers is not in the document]
taefi: Fixed!

flow-test-util/pom.xml
```diff
@@ -36,7 +39,133 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-common</artifactId>
+            <version>${netti.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-codec-http</artifactId>
+            <version>${netti.version}</version>
```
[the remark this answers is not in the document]
taefi: No newer version is available yet.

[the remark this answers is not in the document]
taefi: @sonatype-lift ignore

flow-test-util/pom.xml
mcollovati, line 1: I wonder if this pom can be simplified by adding an explicit dependency to `selenium-remote-driver`, excluding guava and all netty transitive deps, instead of adding all the netty artifacts one by one.

```xml
<dependency>
    <groupId>org.seleniumhq.selenium</groupId>
    <artifactId>selenium-remote-driver</artifactId>
    <version>4.8.3</version>
    <exclusions>
        <exclusion>
            <groupId>io.netty</groupId>
            <artifactId>*</artifactId>
        </exclusion>
        <exclusion>
            <groupId>com.google.guava</groupId>
            <artifactId>guava</artifactId>
        </exclusion>
    </exclusions>
</dependency>
```

flow-tests/vaadin-spring-tests/pom.xml
```diff
@@ -134,7 +136,13 @@
...
             <artifactId>httpclient5</artifactId>
-            <version>5.2.1</version>
+            <version>${httpclient5.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.nimbusds</groupId>
+            <artifactId>nimbus-jose-jwt</artifactId>
+            <version>${nimbus-jose-jwt.version}</version>
         </dependency>
```
mcollovati, line 146: I wonder if adding the declaration of these dependencies on top of the `dependency-management` section, will prevent the need of exclusion in all other poms :thinking:

mcollovati:
After merging the main branch, enforcer fails for the `vaadin-test-spring-war` module because of jetty: spring boot imports jetty-bom `11.0.14` but Flow defines `11.0.15` for tests.

taefi:
> After merging the main branch, enforcer fails for the `vaadin-test-spring-war` module because of jetty: spring boot imports jetty-bom `11.0.14` but Flow defines `11.0.15` for tests.

That was kinda predictable since IMO this is an evolutionary process that can finally reach a more mature state. Currently, I would expect that every now and then after adding a new dependency or similar changes, there would be a need to refine the exclusions or to get rid of redundant or duplicate dependency declarations in the project. I would say, we should exclude the conflicting dependency (or find another way) to make this pass and merge it as soon as possible to prevent further misconfigurations.

After having this in place, I suggest we add the other already existing enforcer plugin's rule `banDuplicatePomDependencyVersions` (read more [here](https://maven.apache.org/enforcer/enforcer-rules/banDuplicatePomDependencyVersions.html)) so that we can simplify our hierarchy of dependency in general, which probably helps the convergence problem as well.

flow-test-util/pom.xml
taefi: Maybe it is a good idea for now, although, in a near future we might end up having a situation that needs to use a security hotfix for one of them while the rest remain on the same existing version (until a full release happens). So there is a trade-off in choosing between these approaches.

flow-tests/vaadin-spring-tests/pom.xml
taefi: I know that the order of dependencies overrides dependencies within the scope of `<dependencies> ... </dependencies>` but have no idea whether there is a possibility of having something on top of `<dependencyManagement>` to override something in that way.

flow-test-util/pom.xml
mcollovati: Good point. But wouldn't it be cleaner and easier to maintain to avoid adding all the dependencies right now, add just add an entry for the vulnerable one, and then remove it once the version is fixed in the main dep?

flow-tests/vaadin-spring-tests/pom.xml
mcollovati: If we pin the version of these dependencies in `<dependencyManagement>` all submodules should inherit that one.
I'll give it a try in another branch. If it works, I can commit changes here; otherwise we can proceed with the current solution

flow-test-util/pom.xml
taefi: Done.

flow-tests/vaadin-spring-tests/pom.xml
mcollovati: @taefi I tried to reduce a bit the number of pom files to change with the dependency management solution in #16609
Do you mind taking a look when you have time?

mcollovati:
@taefi last build passed. I've merged master again.
I'll approve the PR once the build completes.
Do you want to take a last look before merging?

mcollovati (approved):

PR #16565, opened 2023-04-11
refactor!: Move generated Flow files into frontend/generated/flow

Artur-:
Removes the need for importing from the target folder and simplifies when all generated frontend files are in the same folder


Artur-:
Hm, now `versions.json` go into `frontend/generated/flow` which is probably wrong as that is not and should not be part of the frontend build

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
```diff
@@ -295,15 +306,16 @@ private void collectModules(List<String> lines) {
...
 
         modules.removeIf(UrlUtil::isExternal);
 
         ArrayList<String> externals = new ArrayList<>();
         ArrayList<String> internals = new ArrayList<>();
 
         for (String module : getModuleLines(modules)) {
-            if (FRONTEND_IMPORT_LINE.matcher(module).matches()) {
+            if (FRONTEND_IMPORT_LINE.matcher(module).matches()
+                    && !module.contains("Frontend/generated/flow/")) {
```
caalador, line 318: Would be good to have `"Frontend/generated/flow/"` as a constant so breaking it with a typo later is less likely.

Already in this file I see it used 3 times.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -1174,4 +1159,19 @@ public static String getFrontendServletPath(ServletContext servletContext) {
...
+
+    public static File getFlowGeneratedImports(File frontendFolder) {
+        return new File(getFlowGeneratedFolder(frontendFolder), IMPORTS_NAME);
+    }
+
+    public static File getFlowGeneratedWebComponentsFolder(
+            File frontendDirectory) {
+        return new File(getFlowGeneratedFolder(frontendDirectory),
+                "web-components");
+    }
```
caalador, line 1175: JDocs missing for public methods

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
```diff
@@ -222,7 +222,11 @@ private class UpdateFallBackImportsFile extends AbstractUpdateImports {
 
         UpdateFallBackImportsFile(ClassFinder classFinder, Options options) {
             super(options);
-            generatedFallBack = new File(options.getGeneratedFolder(),
+            generatedFallBack = new File(
+                    FrontendUtils
+                            .getFlowGeneratedImports(
+                                    options.getFrontendDirectory())
+                            .getParentFile(),
                     FrontendUtils.FALLBACK_IMPORTS_NAME);
```
caalador, line 230:
```suggestion
                    FrontendUtils
                            .getFlowGeneratedFolder(
                                    options.getFrontendDirectory()),
                    FrontendUtils.FALLBACK_IMPORTS_NAME);
```
Lets use the simple folder instead of the one with the extra file name

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateImports.java
Artur-: Fixed

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
Artur-: Added

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
Artur-: Done

PR #16568, opened 2023-04-11
chore!: Remove the themeModule concept

Artur-:
When splitting code into multiple chunks, there is no way to guarantee that some modules are globally loaded before others

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/EntryPointData.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/EntryPointData.java
@@ -41,7 +41,6 @@ public final class EntryPointData implements Serializable {
     String layout;
     ThemeData theme = new ThemeData();
     final LinkedHashSet<String> modules = new LinkedHashSet<>();
-    final LinkedHashSet<String> themeModules = new LinkedHashSet<>();
     final LinkedHashSet<String> scripts = new LinkedHashSet<>();
     final transient List<CssData> css = new ArrayList<>();
 
@@ -62,10 +61,6 @@ public final class EntryPointData implements Serializable {
 
     LinkedHashSet<String> getModules() {
         return modules;
-    }
-
-    LinkedHashSet<String> getThemeModules() {
-        return themeModules;
     }
 
     LinkedHashSet<String> getScripts() {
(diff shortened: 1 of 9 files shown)


Artur-:
This probably requires https://github.com/vaadin/web-components/pull/5752 first so that Material styles are also put first in `<head>`. Then we can simplify but not considering some modules "special" in some way, which is problematic in the long run, e.g. when code splitting

caalador:
Will this/chunks have an impact to theme vs `@CssImport`/`@StyleSheet` load order in views?

Artur-:
All tests pass so surely not :)

Artur-:
As I understand it, the feature is used to ensure that annotations on `Lumo` or `Material` end up as imports before other imports in `generated-flow-imports.js`. With the changes for Lumo already done and the changes for Material done in https://github.com/vaadin/web-components/pull/5752, it no longer matters at which point they are imported, because the global styles are always added in the beginning of `<head>`

caalador:
Not all tests seem to pass =P
```

Error:  Failures: 
Error:    FrontendDependenciesTest.annotationsInRouterLayoutWontBeFlaggedAsBelongingToTheme:263 Theme's annotations should come first expected:<[theme-foo].js> but was:<[@vaadin/common-frontend/ConnectionIndicator].js>
Error:    ScannerThemeTest.should_takeThemeFromLayout_ifLayoutAlreadyVisited:190 expected:<com.vaadin.flow.server.frontend.scanner.ScannerTestComponents$Theme1> but was:<null>
Error:    ScannerThemeTest.should_visitDefaultTheme_when_noThemeAnnotationIsGiven:137 Theme4 should have been visited and JsModule collected
```

Artur-:
Yes, I was talking about the integration tests that validate that apps still work. The unit tests test that the code works the old way except for maybe the last that is still a problem

Artur-:
Fixed the tests. As the theme of an entrypoint is only used to get the application theme, it does not seem fruitful to test it in `should_visitDefaultTheme_when_noThemeAnnotationIsGiven`

Artur-:
Rebased... hope it is still the same code

caalador (approved):

caalador (approved):

Issue #16032, opened 2023-02-24
Version information warning when building for production


Artur-:
### Description of the bug

```
mvn install -Pproduction
```

logs `Unable to determine version information. No vaadin-core-versions.json found`

```
[INFO] --- vaadin:24.0.0.beta4:build-frontend (default) @ spring-skeleton ---
[INFO] Reflections took 198 ms to scan 132 urls, producing 5454 keys and 28593 values
[INFO] Scanning classes to find frontend configurations and dependencies...
[INFO] Visited 2345 classes. Took 861 ms.
[INFO] Visited 102 classes. Took 38 ms.
[INFO] Unable to determine version information. No vaadin-core-versions.json found
[INFO] Running `npm install` to resolve and optionally download frontend dependencies. This may take a moment, please stand by...
```

### Expected behavior

It should find what it needs or expects

### Minimal reproducible example

https://github.com/vaadin/skeleton-starter-flow-spring

### Versions

- Vaadin / Flow version: 24 beta 4

czp13:
Not happening to me anymore. Executed today.
`mvn install -Pproduction` on `V24` branch.

With these properties from the pom.xml:
```
    <properties>
        <java.version>17</java.version>
        <vaadin.version>24.0.2</vaadin.version>
        <selenium.version>4.8.1</selenium.version>
    </properties>
```

I guess 24.0.2 fixed this. Maybe we can close this?

Artur-:
```
% npx @vaadin/cli init test-prod
% cd test-prod
% mvn install -Pproduction
[INFO] Scanning for projects...
[INFO] 
[INFO] -----------------< com.example.application:test-prod >------------------
[INFO] Building test-prod 1.0-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
Downloading from central: https://repo.maven.apache.org/maven2/org/webjars/npm/mobile-drag-drop/maven-metadata.xml
Downloading from Vaadin Directory: https://maven.vaadin.com/vaadin-addons/org/webjars/npm/mobile-drag-drop/maven-metadata.xml
Downloaded from central: https://repo.maven.apache.org/maven2/org/webjars/npm/mobile-drag-drop/maven-metadata.xml (456 B at 2.0 kB/s)
[INFO] 
[INFO] --- resources:3.3.0:resources (default-resources) @ test-prod ---
[INFO] Copying 1 resource
[INFO] Copying 3 resources
[INFO] 
[...]

PR #16575, opened 2023-04-13
refactor: Do not write a versions.json

Artur-:


czp13:
I have some changes stashed related to this part, but I can just discarding if it solved this issue:
- https://github.com/vaadin/flow/issues/16032
Does this change fix the version find the issue?

Artur-:
> I have some changes stashed related to this part, but I can just discarding if it solved this issue:
> 
> * [Version information warning when building for production #16032](https://github.com/vaadin/flow/issues/16032)
> 
> Does this change fix the version find the issue?

It is not related to that

caalador:
As we now see from the slow tests `versions.json` was written to file because pnpm uses it to lock defined versions in the json file.

caalador:
It would seem that pnpm supports oveerride in the format:
```
{
  "pnpm": {
    "overrides": {
      "foo": "^1.0.0",
      "quux": "npm:@myorg/quux@^1.0.0",
      "bar@^2.1.0": "3.0.0",
      "qar@1>zoo": "2"
    }
  }
}
```
Or does pnpm now support the npm overrides object? 
The documentation left me wondering a bit: https://pnpm.io/package_json#pnpmoverrides

Artur-:
> As we now see from the slow tests `versions.json` was written to file because pnpm uses it to lock defined versions in the json file.

Yes, now it writes it into pnpmfile instead

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -110,18 +109,13 @@ public void execute() {
         }
     }
 
-    boolean lockVersionForNpm(JsonObject packageJson, String versionsPath)
+    boolean lockVersionForNpm(JsonObject packageJson, JsonObject versionsJson)
```
caalador, line 112: I guess we could just use the package protected versionsJson directly instead of giving it as a parameter

Artur-: We could. Done

caalador (dismissed):

caalador (approved):

PR #16584, opened 2023-04-14
refactor: Collect reachable class info per entry point

Artur-:
This changes the class visitor so that we
1. Find all entry points
2. Scan through all classes reachable from any of the entry points and store metadata for that class
3. Collect the metadata for all classes reachable for each individual entry point

It should allow us to generate an import file separately for each entry point and then lazy load scripts and css for some of the entry points

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
@@ -70,7 +70,8 @@ public class FrontendDependencies extends AbstractDependenciesScanner {
     private ThemeDefinition themeDefinition;
     private AbstractTheme themeInstance;
     private final HashMap<String, String> packages = new HashMap<>();
-    private final Set<String> visited = new HashSet<>();
+    private final Map<String, ClassInfo> visitedClasses = new HashMap<>();
+
     private PwaConfiguration pwaConfiguration;
 
     /**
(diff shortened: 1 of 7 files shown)


Artur-:
Do you think this makes sense @caalador ? There are some unit tests to fix yet but the ITs seem to have no problem with this change

caalador:
So the main change is how we store the entrypoint contents?
I don't see any specific problem with the approach.

Artur-:
The only change here is that the we store info per class and then collect the info together into info per entry point. Earlier everything was stored in a bit of a mixed way and classes were some times visited many times

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -128,30 +129,85 @@ public FrontendDependencies(ClassFinder finder,
...
                     "Unable to compute frontend dependencies", e);
         }
     }
 
+    private void aggregateEntryPointInformation() {
+        for (Entry<String, EntryPointData> entry : entryPoints.entrySet()) {
+            EntryPointData entryPoint = entry.getValue();
+            for (String className : entryPoint.reachableClasses) {
+                entryPoint.getModules()
+                        .addAll(visitedClasses.get(className).modules);
```
caalador, line 155: nit: `visitedClasses.get(className)` could be a variable as it is used 3 times and I got a bit confused what are we doing with the entrypoint we are looking at at the moment. Having `ClassInfo` written out for the variable would help readability.

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
Artur-: refactored

PR #16599, opened 2023-04-17
refactor: Extract writeIfChanged to a util method

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/frontend/FileIOUtils.java
```diff
@@ -0,0 +1,94 @@
...
+        if (content.equals(existingFileContent)) {
+            // Do not write the same contents to avoid frontend recompiles
+            log().debug("skipping writing to file '{}' because content matches",
+                    file);
+            return false;
+        }
+
+        log().debug(
+                "skipping writing to file '{}' because content does not match",
+                file);
```
czp13, line 76: Should not it be?:
```suggestion
        log().debug(
                "writing to file '{}' because the content does not match",
                file);
```
Because we actually are writing to the file at 79. line.

czp13 (approved):
Ok, now it seems to same to me.
updateImportsFile was merged to writeIfChanged (and overloaded with different parameters to support places where it is called) and extracted all related methods to FileIOUTils.
🙇

PR #16601, opened 2023-04-17
refactor: Simplify imports updater

Artur-:


czp13 (approved):

czp13 (approved):
Checking this thoroughly I believe the changes are safe (basically doing the same without overriding methods from the parent abstract class, but rather using the parent finder and scanner and related methods from the AbstractUpdateImports) + deleting those not used anymore methods from the children classes.


`getGeneratedModules` in `AbstractUpdateImports.java` and `GenerateMainImports.java` is the same just the Set.of() vs Stream.of() and collecting is different
```
    abstract Collection<String> getGeneratedModules();
..
                Set.of(FrontendUtils.IMPORTS_NAME)); //vs Stream.of(...) and collecting to Set
    }
 ```

`GenerateMainImports.java` seems different, but eventually, it is the same as well, fetching the generated-flow-imports.js, in a slightly different way, using different naming, and code for excluding, but with the same end result.

**I think it is safe to merge, especially with the covering test.** 🙇

PR #16602, opened 2023-04-17
test: Maybe make test more stable

Artur-:


czp13 (approved):
- SessionValueView refactored to wrap div into a span to have an ID.
- SessionValueIT using that id, to find the element and wait for it, otherwise give an exception.

ThemeSwitchLiveReloadIT changes seems a bit unrelated to change remove the view from the path, but makes sense as both ThemeSwitch and SessionValue views are at the same place/path.

Artur-:
ThemeSwitchLiveReloadIT did not actually use the view class but it used the RouteNotFound view...Changing the global theme still had an effect though

PR #16609, opened 2023-04-18
WIP: apply dependency convergence (alternative to 16564)

mcollovati:


taefi (approved):
This is definitely a more organized way of pinning and managing the exclusions. Even though the build is failing because of different versions of spring-web being pulled into the classpath in `flow-test-redeployment` module (which can be solved by adding the proper exclusion), the changes look fine to me and can be merged right after the build is green.

PR #16632, opened 2023-04-19
test: Add production-bundle test build

### LINKED ISSUE #16628: Add a test module in Flow that makes the default prod bundle in the same way as platform does, to be used for testing default prod bundle in Flow

caalador:
Add a test build for production bundle.

Closes #16628

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -670,7 +670,7 @@ export const vaadinConfig: UserConfigFn = (env) => {
       devMode && vaadinBundlesPlugin(),
       devMode && showRecompileReason(),
       settings.offlineEnabled && buildSWPlugin({ devMode }),
-      !devMode && statsExtracterPlugin(),
+      statsExtracterPlugin(),
       devBundle && preserveUsageStats(),
       themePlugin({devMode}),
       postcssLit({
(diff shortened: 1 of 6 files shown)


flow-tests/test-express-build/test-prod-bundle/.gitignore
```diff
@@ -0,0 +1,2 @@
+dev-bundle
```
czp13, line 1: Probably no separate .gitignore is needed. I did not see it in other `test-...` projects.
The root dir `.gitignore` already contains some rules for the other flow tests:
- https://github.com/vaadin/flow/blob/main/.gitignore#L49

Maybe we could just add there these lines?
```
flow-tests/test-express-build/test-prod-bundle/.src/main/dev-bundle
flow-tests/**/vite.*.ts
```

czp13 (approved):

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -670,7 +670,7 @@ export const vaadinConfig: UserConfigFn = (env) => {
       devMode && vaadinBundlesPlugin(),
       devMode && showRecompileReason(),
       settings.offlineEnabled && buildSWPlugin({ devMode }),
-      !devMode && statsExtracterPlugin(),
+      statsExtracterPlugin(),
```
mshabarov, line 673: I'm not against removing devMode check, but doesn't it increase hot reload time of Vite in hot deploy mode?

mshabarov (approved):

PR #16638, opened 2023-04-19
fix: Try polling again on connection refused

Artur-:
This should fix problems with the dev mode not ready page showing forever on startup

### DIFF
--- a/vaadin-dev-server/src/main/resources/com/vaadin/base/devserver/dev-mode-not-ready.html
+++ b/vaadin-dev-server/src/main/resources/com/vaadin/base/devserver/dev-mode-not-ready.html
@@ -1,78 +1,76 @@
 <html>
-<head>
-<meta name="viewport" content="width=device-width, initial-scale=1">
-<meta name="color-scheme" content="dark light">
-<style>
-  body {
-    display: flex;
-    align-items: center;
-    justify-content: center;
-    margin: 0;
-    padding: 1rem;
-    min-height: 100vh;
-    box-sizing: border-box;
-  }
-
-  .message {
-    background-color: rgba(50,50,50,1);
-    color: white;
-    border-radius: 2rem;
-    box-sizing: border-box;
-    padding: 0.75em 1.5em 0.75em 0.75em;
-    font: 0.875rem/1.2 system-ui, ui-sans-serif, sans-serif;
-    display: flex;
-    align-items: center;
-    gap: 0.75em;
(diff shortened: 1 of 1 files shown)


czp13 (approved):

vaadin-dev-server/src/main/resources/com/vaadin/base/devserver/dev-mode-not-ready.html
```diff
@@ -54,7 +54,7 @@
 const poll = () => {
   try {
     fetch(window.location.href, { headers: { 'X-DevModePoll': 'true' } }).then(response => {
```
platosha, line 56: Needs an `await` to catch async errors in the outer `try ... catch`:
```suggestion
    await fetch(window.location.href, { headers: { 'X-DevModePoll': 'true' } }).then(response => {
```

vaadin-dev-server/src/main/resources/com/vaadin/base/devserver/dev-mode-not-ready.html
```diff
@@ -54,7 +54,7 @@
...
-      if (response.headers.has("X-DevModePending")) {
+      if (!response || !response.headers || response.headers.has("X-DevModePending")) {
```
platosha, line 57: Seems superficial, `fetch` must resolve to `Response` as per spec.

```suggestion
      if (response.headers.has("X-DevModePending")) {
```

Artur-: Not sure what the case was there `then` was called with an undefined `response` but maybe it is not important

vaadin-dev-server/src/main/resources/com/vaadin/base/devserver/dev-mode-not-ready.html
Artur-: There is not top level await support in all Vaadin 23 browsers, so maybe `.catch()` is better then

Artur-: ah but this isn't even top level...

Artur-: Maybe the new version makes more senes

platosha (approved):

czp13 (approved):

PR #16643, opened 2023-04-20
chore: Add missing semicolon

Artur-:


czp13 (approved):
I cannot test this now, but I have like 3 things I wanna test soonish on Windows... I just need to setup a Win machine first :D.

But the path seems safe, and could work.

PR #16644, opened 2023-04-20
feat: Added single and double click event listeners registration methods

TatuLund:
Fixes https://github.com/vaadin/flow-components/issues/2173


flow-server/src/main/java/com/vaadin/flow/component/ClickNotifier.java
```diff
@@ -52,6 +52,73 @@ default Registration addClickListener(
...
+     * Adds a double click listener to this component.
+     *
+     * @param listener
+     *            the listener to add, not <code>null</code>
+     * @return a handle that can be used for removing the listener
+     */
+    default Registration addDoubleClickListener(
+            ComponentEventListener<DoubleClickEvent<T>> listener) {
+        if (this instanceof Component) {
+            Registration internalEventReg = this.addClickListener(event -> {
```
Legioth, line 65: This seems overly complicated. What about instead doing something like this (along with adding `@DomEvent("click")` to the event type):
```
return ComponentUtil.addListener(this, DoubleClickEvent.class, listener,
  registration -> registration.setFilter("event.detail == 2"));
```

Aside from less code, this would also have the benefit that filtering happens already on the client so that redundant requests aren't even sent to the server.

flow-server/src/main/java/com/vaadin/flow/component/DoubleClickEvent.java
```diff
@@ -0,0 +1,85 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component;
+
+/**
+ * Event fired when a component is single clicked.
```
Legioth, line 19: double clicked

flow-server/src/main/java/com/vaadin/flow/component/DoubleClickEvent.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Creates a new server-side click event with no additional information.
+     *
+     * @param source
+     *            the component that fired the event
+     */
+    public DoubleClickEvent(Component source) {
+        // source, notClient, 4 coordinates, clickCount, button, 4 modifier
```
Legioth, line 81: There's 12 items in this list but only 11 passed parameters

flow-server/src/main/java/com/vaadin/flow/component/AbstractClickEvent.java
```diff
@@ -0,0 +1,190 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component;
+
+/**
+ * Abstract super class event for e.g. DoubleClickEvent and SingleClickEvent.
```
Legioth, line 19: Any reason to not consistently use this for all click events, including the existing `ClickEvent` class?

flow-server/src/main/java/com/vaadin/flow/component/DoubleClickEvent.java
TatuLund: Forgot to delete clickCount from the comment

flow-server/src/main/java/com/vaadin/flow/component/AbstractClickEvent.java
Legioth: Now when I think about it further, I wonder whether separate event types are really needed at all when the only practical difference is the `getClickCount()` method?

TatuLund: Especially filtering the single click event, which currently requires hacks. So this is convenience, and DX improvement.

TatuLund: > Any reason to not consistently use this for all click events, including the existing ClickEvent class?**

I could modify ClickEvent to extend this if that is preferred. I was just conservative with my approach.

TatuLund: Fixed now

flow-server/src/main/java/com/vaadin/flow/component/ClickNotifier.java
TatuLund: Fixed

flow-server/src/main/java/com/vaadin/flow/component/AbstractClickEvent.java
Legioth: From the DX perspective, it's just `button.addSingleClickListener(event -> doSomething())` regardless of whether there's a dedicated event type.

As for the hacks, I would still suggest to do filtering in the browser instead by using the `ComponentUtil.addListener` overload with a callback for modifying the low-level `DomListenerRegistration` to add a filter.

Legioth:
No further comment from my side, but I leave final review in the hands of the Flow team who maintains the code.

I would recommend them to discuss what they think on the question about using multiple different event classes vs using the familiar `ClickEvent` class throughout (and then changing the double click to filter on `event.detail == 2` instead of listening for the `dblclick` event).

mcollovati:
Personally, I would prefer the solution with only one `ClickEvent` and the filter: 
* the effect is the same as having multiple events and it will provide the same DX as there will be a `addDoubleClickListener` method
* less code to maintain
* a simpler API, as there will be only `addClickListener` and `addDoubleClickListener`
* we can consider to add also the proposed `addSingleClickListener`, using a similar filter or an overload of `addClickListener` that takes the number of expected clicks

TatuLund:
> we can consider to add also the proposed addSingleClickListener,

@mcollovati This was the actual need of my customer.

mcollovati:
@TatuLund Sorry, I missed that part. So, it makes sense to add that method

TatuLund:
> Personally, I would prefer the solution with only one ClickEvent and the filter:

@mcollovati simplification is done now, tests pass. So I hope it is finally ok.

flow-server/src/main/java/com/vaadin/flow/component/ClickNotifier.java
```diff
@@ -52,6 +52,51 @@ default Registration addClickListener(
         }
     }
 
+    /**
+     * Adds a single click listener to this component.
```
mcollovati, line 56:
```suggestion
     * Adds a double click listener to this component.
```

mcollovati (changes requested):
LGTM.
There's just a copy/paste typo in javadocs.
Could you also please change the PR title, as now the mentioned event does not exist anymore? 
Maybe something like  "Added single and double click event listeners registration methods"?

PR #16647, opened 2023-04-20
fix: clean Frontend/generated folder

### LINKED ISSUE #16640: Generated files under Frontend not cleaned correctly
### Description of the bug

Generated files created or copied to `Frontend/generated` are not cleaned correctly using the build system `clean` commands as they are outside the build directory (`build`/`target`) which can lead to there being old files that then used for execution and bundle generation.

### Expected behavior

Files in `Frontend/generated` should be only currently active and up to date files and not contain old data.

Run a production build with 24.1.0.alpha3 then with 24.1-SNAPSHOT and see that the resulting generated `generated-flow-imports.js` contains all components as the fallback bundle gets added to the `generated-flow-imports.js` file as a direct import in `Frontend/generated/flow/`

### Versions

- Vaadin / Flow version: 24.1

caalador:
Clean the frontend generated folder
on prepare-frontend task (maven/gradle)
and spring contextInitialization.

closes #16640

### DIFF
--- a/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPrepareFrontendTask.kt
+++ b/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPrepareFrontendTask.kt
@@ -48,6 +48,8 @@ public open class VaadinPrepareFrontendTask : DefaultTask() {
     @TaskAction
     public fun vaadinPrepareFrontend() {
         val extension: VaadinFlowPluginExtension = VaadinFlowPluginExtension.get(project)
+        // Remove Frontend/generated folder to get clean files copied/generated
+        project.delete(extension.generatedTsFolder.absolutePath)
         logger.info("Running the vaadinPrepareFrontend task with effective configuration $extension")
         val adapter = GradlePluginAdapter(project, true)
         val tokenFile = BuildFrontendUtil.propagateBuildInfo(adapter)
--- a/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojoTest.java
+++ b/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojoTest.java
@@ -224,8 +224,7 @@ public class PrepareFrontendMojoTest {
         }
         final File oldFile = new File(flowFolder, "old.js");
         if (!oldFile.createNewFile()) {
-            Assert.fail(
-                    "Failed to generate old.js in Frontend/generated/flow");
+            Assert.fail("Failed to generate old.js in Frontend/generated/flow");
         }
         ;
 
--- a/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojoTest.java
+++ b/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojoTest.java
@@ -225,7 +225,7 @@ public class PrepareFrontendMojoTest {
         final File oldFile = new File(flowFolder, "old.js");
         if (!oldFile.createNewFile()) {
             Assert.fail(
-                    "Failed to generate old file to Frontend/generated/flow");
+                    "Failed to generate old.js in Frontend/generated/flow");
         }
         ;
 
--- a/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojoTest.java
+++ b/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojoTest.java
@@ -213,6 +213,32 @@ public class PrepareFrontendMojoTest {
     }
 
     @Test
+    public void mavenGoal_when_frontendGeneratedExists_shouldClearFolder()
+            throws Exception {
+        if (!generatedTsFolder.mkdirs()) {
+            Assert.fail("Failed to generate Frontend/generated folders.");
+        }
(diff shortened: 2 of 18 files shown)


flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPrepareFrontendTask.kt
```diff
@@ -48,6 +48,8 @@ public open class VaadinPrepareFrontendTask : DefaultTask() {
     @TaskAction
     public fun vaadinPrepareFrontend() {
         val extension: VaadinFlowPluginExtension = VaadinFlowPluginExtension.get(project)
+        // Remove Frontend/generated folder to get clean files copied/generated
+        project.delete(extension.generatedTsFolder.absolutePath)
```
Artur-, line 52: Can Gradle not fail at this?

flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojoTest.java
```diff
@@ -212,6 +212,32 @@ public void mavenGoal_when_packageJsonMissing_shouldNotGenerateDefault()
...
+            Assert.fail("Failed to generate Frontend/generated folders.");
+        }
+        final File flowFolder = new File(generatedTsFolder, "flow");
+        if (!flowFolder.mkdir()) {
+            Assert.fail("Failed to generate flow folder");
+        }
+        final File oldFile = new File(flowFolder, "old.js");
+        if (!oldFile.createNewFile()) {
+            Assert.fail(
+                    "Failed to generate old file to Frontend/generated/flow");
```
Artur-, line 228:
```suggestion
                    "Failed to generate old.js in Frontend/generated/flow");
```

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -439,6 +445,23 @@ public void failFastContextInitialized(ServletContextEvent event)
                 return;
             }
 
+            if (config.getProjectFolder() != null) {
+                final String frontendFolder = config.getStringProperty(
```
Artur-, line 449: This is a weird place to do it. Where does the cleaning happen if you run an embedded Jetty?

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPrepareFrontendTask.kt
caalador: I went according to VaadinCleanTask and also what I get from the API documentation it only returns  true/false for deleted something or not: https://gradle.github.io/kotlin-dsl-docs/api/org.gradle.api/-project/delete.html

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
caalador: Yes. Did this part a bit in a hurry yesterday. Moved to DevModeInitializer

caalador: That was a worse place to have it.
Can't come up with where this cleaning should happen for a start from app.

Artur-: If we would tie it to startup of Vite, then https://github.com/vaadin/flow/blob/38ffaa84158981de846b0f991139f092c6ce72d7/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/AbstractDevServerRunner.java#L202 should only be reached on the first startup and not on server restarts?

caalador: But that is too late in the initialization as we have already generated all the new files.

Artur-: If there is no difference between a server startup and restart, maybe we need to do it in a Spring specific way then? Spring can tell the difference I think as spring dev tools related classes are retained over a restart

caalador: What I have gotten to is that we can have an `ApplicationListener` for `ApplicationStartingEvent`  or a bit later `ApplicationContextInitializedEvent`, but I don't know how I would see if the event is from a dev tools restart so it cleans the folder for each run. 
Also a problem is that it does the clean even for running from maven which is not a big deal as it should be cleared anyway and we don't have the configuration so we could distinguish production execution.

caalador: As I've not found a good solution for this I made issue #16670 for solving the spring app start case.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -86,6 +89,9 @@
 import com.vaadin.flow.spring.VaadinScanPackagesRegistrar.VaadinScanPackages;
 import com.vaadin.flow.theme.Theme;
 
+import static com.vaadin.flow.server.frontend.FrontendUtils.DEFAULT_FRONTEND_DIR;
+import static com.vaadin.flow.server.frontend.FrontendUtils.PARAM_FRONTEND_DIR;
```
mshabarov, line 93: Changes in this class have been reverted, so these imports are not needed too.

mshabarov (changes requested):

Issue #16627, opened 2023-04-19
Add a logic to decide to re-use default bundle or make a custom one


mshabarov:
Vaadin applications usually set up `vaadin-maven-plugin` to run `build-frontend` goal to generate a production bundle.
This goal must re-use the default production bundle shipped by the platform if possible.

Acceptance criteria:
- [ ] `BuildFrontendMojo` uses `BundleValidationUtils` to check if the default prod bundle is sufficient or not. Default prod bundle is available in classpath under `vaadin-prod-bundle` folder, similarly like default dev bundle is.
- [ ] If a new bundle is needed, `BuildFrontendMojo` builds a new bundle as usual
- [ ] If the default one is enough, `BuildFrontendMojo` copies the bundle files from default bundle JAR to the application's JAR/WAR, so to not making any difference be it a default bundle or custom one in runtime.

PR #16653, opened 2023-04-20
feat: Use default prod bundle if possible

mshabarov:
Re-uses re-bundle checker in production mode. Copies default bundle files from JAR to application's resource folder in `target` and skips re-bundling or executes usual frontend build codes if it sees any customisations. Uses class finder to look for frontend resources from flow plugin.

Fixes https://github.com/vaadin/flow/issues/16627
Fixes https://github.com/vaadin/flow/issues/16671


flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
```diff
@@ -117,7 +117,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                     "Could not execute build-frontend goal", exception);
         }
 
-        if (generateBundle()) {
+        if (generateBundle() && !BuildFrontendUtil.bundleExists(this)) {
```
mshabarov, line 120: This means that to re-build the prod bundle, one should `clean` the project, i.e. completely remove Vaadin generated files from build/output folder. If it's may break projects in some way, we have to pass/throw the `needBuild` result this place to not run the checker again.

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
```diff
@@ -611,6 +708,20 @@ public static void logChangedFiles(List<String> frontendFiles,
         getLogger().info(message, handledFiles);
     }
 
+    public static String findProdBundleStatsJson() throws IOException {
+        URL statsJson = getProdBundleResource("config/stats.json");
+        if (statsJson == null) {
+            getLogger().warn("There is no production bundle in the classpath.");
```
knoobie, line 714: Is this a warning or just an information? Looking at the following scenario:
- use Vite instead of Dev Bundle
- don't have a Prod Bundle available in CI because Vite is also running in CI to create a bundle every time the application is build (no shared storage in between CI runs)

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
```diff
@@ -41,6 +45,99 @@
...
+        File npmFolder = options.getNpmFolder();
+
+        if (!options.isProductionMode()
+                && !DevBundleUtils.getDevBundleFolder(npmFolder).exists()
+                && !BundleValidationUtil.hasJarBundle(DEV_BUNDLE_JAR_PATH)) {
+            getLogger().info("No dev-bundle found.");
+            return true;
+        }
+
+        if (options.isSkipDevBundle()) {
```
mshabarov, line 84: We can perhaps extend this parameter to production mode as well.

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
knoobie: Are you sure that's always in the class path? Looking at the recommendation to exclude `vaadin-dev`from production builds to reduce the final artefact size, I'm not sure if it's also there.. or are you goint to add it to the `maven/gradle` plugin as dependency?

knoobie: You can close/resolve this :) Just saw the PR at the platform repo where it's done (added to the plugins, instead of flow-server)

mshabarov: > maven/gradle plugin as dependency

Yes, this is the plan so far.

flow-tests/test-express-build/test-dev-bundle-no-plugin/src/test/java/com/vaadin/flow/testnpmonlyfeatures/nobuildmojo/NoAppBundleIT.java
```diff
@@ -17,22 +23,31 @@ public void init() {
...
                 new File(baseDir, "node_modules").exists());
 
-        // These should not be generated either, but at the moment they are
-        Assert.assertFalse("No package.json should be created",
-                new File(baseDir, "package.json").exists());
-        Assert.assertFalse("No vite generated should be created",
-                new File(baseDir, "vite.generated.ts").exists());
-        Assert.assertFalse("No vite config should be created",
-                new File(baseDir, "vite.config.ts").exists());
+        // TODO: for default prod bundle these are still generated
```
mshabarov, line 41: https://github.com/vaadin/flow/issues/16696

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -107,18 +108,28 @@ public NodeTasks(Options options) {
...
+                if (!needBuild) {
+                    commands.add(new TaskCopyBundleFiles(options));
+                }
+            }
+
             // The dev bundle check needs the frontendDependencies to be able to
             // determine if we need a rebuild as the check happens immediately
             // and no update tasks are executed before it.
-            if (!options.isProductionMode() && options.isDevBundleBuild()) {
+            if (!options.isProductionMode() && options.isBundleBuild()) {
```
caalador, line 124: The comment can be moved inside if and the productionMode flag removed when this is added as an `else if {` after the `if(productionMode)`

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
```diff
@@ -41,6 +45,99 @@
...
+            return needsBuild;
+        } catch (Exception e) {
+            getLogger().error(
+                    "Error when checking if a production bundle build is needed",
+                    e);
+            return true;
+        }
+    }
+
+    public static boolean needsBundleBuild(Options options,
```
caalador, line 72: I would perhaps move these 2 methods out from BundleValidationUtil as they are production bundle specific.
Could perhaps have a class `CheckBundleRequirement` that would have the needsBuild checks for both production bundle and the dev bundle that is now in `TaskRunDevBundleBuild`

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
```diff
@@ -41,6 +45,99 @@
...
+            FrontendDependenciesScanner frontendDependencies,
+            ClassFinder finder) throws IOException {
+        File npmFolder = options.getNpmFolder();
+
+        if (!options.isProductionMode()
```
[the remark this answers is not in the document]
caalador: Seems this is moved from TaskRunDevBundleBuild here, but it is development mode specific even though it is used from the `needsBuildProdBundle` method.

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
```diff
@@ -41,6 +45,99 @@
...
+
+        String statsJsonContent = options.isProductionMode()
+                ? BundleValidationUtil.findProdBundleStatsJson()
+                : DevBundleUtils.findBundleStatsJson(npmFolder);
+
+        if (statsJsonContent == null) {
+            // without stats.json in bundle we can not say if it is up-to-date
+            getLogger().info("No bundle's stats.json found for validation.");
+            return true;
+        }
```
caalador, line 100: I would perhaps have the start separate for development and production bundle and have the rest of the checks mutual in the Util class.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
caalador: Do we need to have this check? shouldn't it be fine to run the checker to see if a build is needed or not?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyBundleFiles.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+import com.vaadin.flow.server.Constants;
+import com.vaadin.flow.server.ExecutionFailedException;
+
+public class TaskCopyBundleFiles implements FallibleCommand {
```
caalador, line 11: JavaDoc missing

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyBundleFiles.java
```diff
@@ -0,0 +1,43 @@
+package com.vaadin.flow.server.frontend;
```
caalador, line 1: Header missing.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyTemplateFiles.java
```diff
@@ -73,7 +73,7 @@ public void execute() throws ExecutionFailedException {
                 }
 
                 File templateDirectory;
-                if (options.isDevBundleBuild()) {
+                if (!options.isProductionMode() && options.isBundleBuild()) {
```
caalador, line 76: Could add the `isDevBundleBuild()` to Options as the written out check. so it would just combine the not production want bundle build.

caalador (changes requested):

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
mshabarov: The checker is used on the previous step `BuildFrontendUtil.runNodeUpdater` when Flow makes a decision to generate `package.json`, install packages etc. We could check here again, but this is double work. I'd prefer to re-use the result of checker also here, but I don't know a good way to pass this flag here without changing the API.

caalador: We could perhaps utilize the token file so in the check when we know the result we would do:
```
        final JsonObject token = Json.parse(
                FileUtils.readFileToString(options.getTokenFile(),
                        StandardCharsets.UTF_8));
        token.put("needsBundle", false);
        FileUtils.write(options.getTokenFile(), token.toJson(), StandardCharsets.UTF_8);
```
And then change `bundleExists` to 
```
        final File tokenFile = getTokenFile(adapter);
        if (!tokenFile.exists()) {
            adapter.logWarn(
                    "Require bundle build due to missing token file.");
            return false;
        }
        try {
            String json = FileUtils.readFileToString(tokenFile,
                    StandardCharsets.UTF_8.name());
            JsonObject buildInfo = JsonUtil.parse(json);
            if(!buildInfo.hasKey("needsBundle")) {
                return false;
            }
            return buildInfo.getBoolean("needsBundle");
        }catch(IOException e){}
```

caalador: And clean the field in `updateBuildFile()` so we don't have leftovers in the end result.

mshabarov: Thanks for the suggestion, I did it in https://github.com/vaadin/flow/pull/16653/commits/90cde687aacfafff0c49aae878fdd8fd6f7682a2 , but seems something overrides the token file between `BuildFrontendUtil.runNodeUpdater(this);` and `BuildFrontendUtil.needsBundleBuild(this)` - i.e. after I write the flag into it and before I try to restore the flag. 

It looks as though it is done by `BuildFrontendUtil.updateBuildFile(this);`, however this line goes even after I try to restore the flag, so I'm confused. Need to check later on.

caalador: I don't see anything that would touch the file in between.
It will be cleaned after, but for that part it should just stay as it is.

mshabarov: Two thing that I get and that I don't completely understand: this file is being overridden and it always has result `true`, which I cannot explain at all. To not depend on the logic of handling token file, I made a separate temp file which has this flag.

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
mshabarov: Refactored to have one public method taking a `Mode` as a parameter, then two private methods that has mode specific things and one internal method that has overall checking logic.

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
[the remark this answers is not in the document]
mshabarov: Refactored

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
mshabarov: Refactored to have one public method taking a `Mode` as a parameter, then two private methods that has mode specific things and one internal method that has overall checking logic.

caalador:
Seems ok, but build has a failing test
`BundleValidationTest.bundleMissesSomeEntries_skipBundleBuildSet_noBundleRebuild:1602 Rebuild should be skipped`

mshabarov:
> Seems ok, but build has a failing test `BundleValidationTest.bundleMissesSomeEntries_skipBundleBuildSet_noBundleRebuild:1602 Rebuild should be skipped`

Fixed, this test is for dev mode currently, since production mode has no yet such a feature to skip rebundling.

caalador (approved):

PR #16657, opened 2023-04-20
feat: More eager cleanup for UIs using Beacon API

### LINKED ISSUE #6293: Immediately detach the old UI when the browser tab is closed or refreshed
The [Beacon API](https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API) is now relatively widely available, making it possible to propagate an unload event to the server when a tab is closed/refresh without affecting user experience.

This could be used to close UIs on close/refresh immediately, without waiting for expiration due to missed heartbeats. This has the potential to improve performance, primarily by decreasing memory consumption.

The heartbeat expiration or similar is still needed for when the unload event can not be propagated to the server, e.g. when the user loses network connectivity.

mstahv:
Notifies server about closed UIs using a beacon request on pagehide event.

At the same time unifies behaviour in certain edge cases, where Safari maintained the state when brought back from page cache. Previously Safari in some situations kept the state.

Tested manually with Safari, Chrome, Firefox and validated results using VisualVM.

Closes #6293 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
+++ b/flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
@@ -141,7 +141,7 @@ public class ApplicationConnection {
             // with browser buttons
             // Chrome discards our state as beforeunload is used
             // As state is most likely cleared on the server already (especially
-            // now with beacon API request, it is probably
+            // now with Beacon API request, it is probably
             // better resyncronize the state (would happen on first server
             // visit)
             Browser.getWindow().getLocation().reload();
--- a/flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
+++ b/flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
@@ -29,6 +29,7 @@ import com.vaadin.client.flow.dom.DomApi;
 import com.vaadin.client.flow.util.NativeFunction;
 
 import elemental.client.Browser;
+import elemental.dom.Document;
 import elemental.dom.Element;
 import elemental.dom.Node;
 
@@ -131,6 +132,21 @@ public class ApplicationConnection {
             registry.getRequestResponseTracker().startRequest();
             registry.getMessageHandler().handleMessage(initialUidl);
         }
+
+        Browser.getWindow().addEventListener("pagehide", e -> {
+            registry.getMessageSender().sendUnloadBeacon();
+        });
+
+        Browser.getWindow().addEventListener("pageshow", e -> {
+            // Currently only Safari gets here, sometimes when going back/foward
+            // with browser buttons
+            // Chrome discards our state as beforeunload is used
+            // As state is most likely cleared on the server already (especially
+            // now with beacon API request, it is probably
+            // better resyncronize the state (would happen on first server
+            // visit)
+            Browser.getWindow().getLocation().reload();
+        });
     }
 
     /**
--- a/flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
+++ b/flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
@@ -29,7 +29,6 @@ import com.vaadin.client.flow.dom.DomApi;
 import com.vaadin.client.flow.util.NativeFunction;
 
 import elemental.client.Browser;
-import elemental.dom.Document;
 import elemental.dom.Element;
 import elemental.dom.Node;
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
@@ -349,9 +349,14 @@ public class AtmospherePushConnection implements PushConnection {
                 try {
                     outgoingMessage.get(1000, TimeUnit.MILLISECONDS);
                 } catch (TimeoutException e) {
-                    getLogger().info(
-                            "Timeout waiting for messages to be sent to client before disconnect",
-                            e);
+                    if (ui.isClosing()) {
+                        getLogger().debug(
+                                "Something was not sent to client on an UI that was already closed by beacon reqeuest or similar. This seems to happen with Safari occassionally when navigating away from a UI.");
+                    } else {
+                        getLogger().info(
+                                "Timeout waiting for messages to be sent to client before disconnect",
+                                e);
+                    }
                 } catch (Exception e) {
                     getLogger().info(
                             "Error waiting for messages to be sent to client before disconnect",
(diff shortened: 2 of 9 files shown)


mstahv:
Included IT only runs in Chrome 🤔 Might be good to get it run also on Chrome and Safari + permutation with Push on, but some team member can probably configure that about 10k time faster than me 👼

mstahv:
Looks like Push is on by default for these tests, I think that is enough as the xhr mode has much less moving parts. Also heard that in "platform tests" these tests are run on other browsers, so all should be covered.

flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
```diff
@@ -131,6 +132,21 @@ public void start(ValueMap initialUidl) {
             registry.getRequestResponseTracker().startRequest();
             registry.getMessageHandler().handleMessage(initialUidl);
         }
+
+        Browser.getWindow().addEventListener("pagehide", e -> {
```
knoobie, line 136: General Note / Question: Should this be enabled / toggled by the developer on application level? 

I was wondering if this could interfere with people's expectation that the UI is still present after the UI was redirected to e.g. PayPal or other payment providers and afterwards gets redirected again to the Vaadin UI.. where the user's session could be instantly lost because all UIs were immediately closed?

mstahv: Finally, I already told the Flow team that this is all done, except it is still waiting for a review by @knoobie  :-)

Now that I changed the behaviour on Monday that the beacon request is ignored on the server side for UIs with "preserveonrefresh" on, no need for that. For other UIs the state is lost anyways and we are still never dropping the session, that can(and should be) used for such state.

knoobie: You can always expect me lurking somewhere 😉 

Thanks for thinking this though before I was here 😅 My job here is done 😎

mstahv: I seriously value your insights a lot. You have good eye for code and so much experience of Vaadin that I don't see why your reviews wouldn't be enough 🙇

flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
```diff
@@ -349,9 +349,14 @@ public void disconnect() {
                 try {
                     outgoingMessage.get(1000, TimeUnit.MILLISECONDS);
                 } catch (TimeoutException e) {
-                    getLogger().info(
-                            "Timeout waiting for messages to be sent to client before disconnect",
-                            e);
+                    if (ui.isClosing()) {
+                        getLogger().debug(
+                                "Something was not sent to client on an UI that was already closed by beacon reqeuest or similar. This seems to happen with Safari occassionally when navigating away from a UI.");
```
mcollovati, line 354:
```suggestion
                                "Something was not sent to client on an UI that was already closed by beacon request or similar. This seems to happen with Safari occassionally when navigating away from a UI.");
```

flow-server/src/main/java/com/vaadin/flow/server/communication/ServerRpcHandler.java
```diff
@@ -340,6 +343,24 @@ public void handleRpc(UI ui, Reader reader, VaadinRequest request)
             // signature for source and binary compatibility
             throw new ResynchronizationRequiredException();
         }
+        if (rpcRequest.isUnloadBeaconRequest()) {
+            if (isPreserveOnRefreshTarget(ui)) {
+                getLogger().debug(
+                        "Eager UI close ignorer for PreserveOnRefresh view");
```
mcollovati, line 349:
```suggestion
                        "Eager UI close ignored for @PreserveOnRefresh view");
```

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/UIsCollectedWithBeaconAPIIT.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.Test;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+
+import com.vaadin.flow.testutil.ChromeBrowserTest;
+
+public class UIsCollectedWithBeaconAPIIT extends ChromeBrowserTest {
+
+    @Test
+    public void firstRun() {
```
mcollovati, line 28: nit: I would rename so that it should be easier to immediately understand what we are testing.
Something like `beaconHandling_navigateAwayFromApplication_uiClosedEarly`

mcollovati (changes requested):
Some typo and nits.
I would also like @heruan to have a brief look, to be sure this change may not break beacon handling in Collaboration Kit

mstahv:
I briefly looked at what @heruan  et co have done in collaboration engine with beacon api, and my hunch is that no issues with that. Currently you seem to be hooking on unload event, which is kind of deprecated but works most often. Depending on how you use it, you might in some cases now (after this change) use normal detach listeners, but I'd expect that you should at least in upcoming versions also detect visibility changes as well (e.g. to gray out a user who has switched to another tab) 🤷‍♂️

mcollovati (approved):
LGTM (unless there are troubles with Collaboration KIT)

flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
```diff
@@ -29,6 +29,7 @@
 import com.vaadin.client.flow.util.NativeFunction;
 
 import elemental.client.Browser;
+import elemental.dom.Document;
```
czp13, line 32: Not used, could be removed:
```suggestion
```

flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
```diff
@@ -131,6 +132,21 @@ public void start(ValueMap initialUidl) {
...
+        Browser.getWindow().addEventListener("pagehide", e -> {
+            registry.getMessageSender().sendUnloadBeacon();
+        });
+
+        Browser.getWindow().addEventListener("pageshow", e -> {
+            // Currently only Safari gets here, sometimes when going back/foward
+            // with browser buttons
+            // Chrome discards our state as beforeunload is used
+            // As state is most likely cleared on the server already (especially
+            // now with beacon API request, it is probably
```
czp13, line 145:
```suggestion
            // now with Beacon API request, it is probably
```

mcollovati (approved):

czp13 (approved):
👍

PR #16660, opened 2023-04-21
fix: postpone serverSyncId computation after changes are applied

### LINKED ISSUE #16635: UI::push inside attachment listener of a dialog child makes it fail
### Description

When using ui push inside a grid attach listener inside a dialog (while manual push mode is enabled) makes the whole ui fail to load.

### Outcome
Ui fails to load, then resyncronizes UI.
The only error message is in the javascript console `(TypeError) : Cannot read properties of undefined (reading 'a')` and `Uncaught (in promise) Error: Client is resynchronizing`

### Expected outcome

The ui loads as expected, or at least an error is thrown.

```java
package com.example.application.views.main;

import com.vaadin.flow.component.dialog.Dialog;
import com.vaadin.flow.component.grid.Grid;
import com.vaadin.flow.component.orderedlayout.HorizontalLayout;
import com.vaadin.flow.router.PageTitle;

import java.util.ArrayList;

@PageTitle("Main")
@Route(value = "")
public class MainView extends HorizontalLayout {
    private static class Item {
        public String name;

        public Item(String s) {
            name = s;
        }
    }

    public MainView() {
        setSizeFull();

        var items = new ArrayList<Item>();

        for (int i = 0; i < 100; i++) {
            items.add(new Item("Item " + i));
        }

        var grid = new Grid<Item>();
        grid.setSizeFull();
        grid.setItems(query -> items.stream().skip(query.getOffset()).limit(query.getLimit()));
        grid.addColumn(item -> item.name);

        grid.addAttachListener(event -> {
            // this makes everything fail to load
            event.getUI().push();
        });

        var dialog = new Dialog();
        dialog.add(grid);
        dialog.setSizeFull();
        dialog.open();
    }
}

```

### Steps to reproduce

1. Download the Vaadin 24.0.4 starter project
2. Add `@Push(PushMode.MANUAL)` to the Application.java
3. Replace the MainView.java with the provided example
4. Notice the Dialog is not loading properly
5. Comment out `event.getUI().push();` and see how everything load correctly.
6. Additionally, notice that `event.getUI().accessSynchronously(() -> event.getUI().push());` fails aswell, while `event.getUI().access(() -> event.getUI().push());` seems to delay the push but work

### Environment

Vaadin version(s): 24.0.4
OS: Windows 11 x64

### Browsers

_No response_

mcollovati:
## Description

While UIDLWriter is encoding changes, it may happen that a push operation is invoked, thus incrementing the serverSyncId counter. If this happens, the client may receive the messages in the wrong order and and subsequently there may be a gap in the counter, so that a message will never be created for an expected id, and the client will force a resynchronization.
This change reads the serverSyncId for the current request after all changes are applied, so the client will receive messages in the expected order.

Fixes #16635

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/server/communication/UidlWriter.java
```diff
@@ -194,6 +189,14 @@ public JsonObject createUidl(UI ui, boolean async, boolean resync) {
         if (service.getDeploymentConfiguration().isRequestTiming()) {
             response.put("timings", createPerformanceData(ui));
```
czp13, line 190: Maybe not related closely to the PR, but I see the inconsistency between 
`response.put()`code lines:
- sometimes we use constants, 
- sometimes we do not:

`response.put("timings", createPerformanceData(ui));`  
VS 
`response.put(ApplicationConstants.CLIENT_TO_SERVER_ID,
                nextClientToServerMessageId);`
                
Probably would make sense to add "changes", "constants", "meta", "timings" to the ApplicationConstants.

czp13 (commented):
I think it is a nice, simple, straightforward change in the `createUidl()` logic (ordering). 👍 
Also makes sense to know the previous and actual syncId for trace logging.
This is a new topic to me, but it could work.

flow-server/src/main/java/com/vaadin/flow/server/communication/UidlWriter.java
mcollovati: I suppose it is because the keys you listed are used only in this code block, so there's no need for a constant.

czp13 (approved):

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/server/communication/UidlWriter.java
```diff
@@ -194,6 +189,14 @@ public JsonObject createUidl(UI ui, boolean async, boolean resync) {
...
         }
+
+        // Get serverSyncId after all changes has been computed, as push may
+        // have been invoked, thus incrementing the counter.
+        // This way the client will receive messages in the correct order
+        int syncId = service.getDeploymentConfiguration().isSyncIdCheckEnabled()
+                ? uiInternals.getServerSyncId()
+                : -1;
+        response.put(ApplicationConstants.SERVER_SYNC_ID, syncId);
```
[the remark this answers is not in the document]
mcollovati: If (A + 2) is received before (A + 1), the client waits for a while to receive (A +1).
If (A + 1) is received, then also (A + 2) is applied, otherwise the client forces a resynchronization

PR #16682, opened 2023-04-24
refactor: Unify getters so List is used for modules, scripts and css

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -249,12 +249,12 @@ public List<String> getModules() {
      * @return the set of JS files
      */
     @Override
-    public Set<String> getScripts() {
-        Set<String> all = new LinkedHashSet<>();
+    public List<String> getScripts() {
+        LinkedHashSet<String> all = new LinkedHashSet<>();
```
caalador, line 253: Nit: could have kept this as Set<String> like for getCss

caalador (approved):

PR #16691, opened 2023-04-25
fix: Only include generated web component files into generated-imports

Artur-:


flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
```diff
@@ -84,22 +84,28 @@ public void execute() {
     }
 
     @Test
-    public void getGeneratedModules_should_excludeByFileName()
+    public void getGeneratedModules_should_includeWebComponents()
```
caalador, line 87:
```suggestion
    public void getGeneratedModules_should_includeOnlyWebComponents()
```
Only webcomponents should be added an no others.

caalador (commented):
Does this change anything apart from mistaken imports due to old files in generated folder?

Artur-:
It only prevents extra files from being included, if they happen to be in the flow folder

caalador (approved):

PR #16693, opened 2023-04-25
feat: Support lazy loading of route dependencies

Artur-:
Adds a `@LoadDependenciesOnStartup` annotation you can set on your `AppShellConfigurator` class to define for which routes the dependencies should be loaded eagerly when the application is first visited, e.g. `@LoadDependenciesOnStartup(MyMainView.class)`

Defining only `@LoadDependenciesOnStartup` causes all dependencies to be loaded eagerly 

Not defining any `@LoadDependenciesOnStartup` annotation uses the Flow default, which currently is to load everything eagerly but is bound to be changed to load `""` and `"login"` routes, if defined, eagerly and dependencies for the remaining views on demand.

Also adds a `@DependencyTrigger` that overrides which class triggers loading of the dependencies for a route not loaded on startup. It is meant to be used internally to create the default production bundle. It is not meant for any other usage.


caalador (dismissed):

Artur-:
Based on some testing, I think it would make sense NOT to have it as a route attribute, especially not with eager as the default.

It is natural to load resources when you need them, that is how the web works. To accomplish this you would need to go through almost all routes in the app and change them.

We could make on-demand the default but then you have the opposite problem if you for some reason want to load everything upfront.

I will see if we can make it just a property instead, with some kind of wildcard matching

Artur-:
Based on some testing, I think it would make sense NOT to have it as a route attribute, especially not with eager as the default.

It is natural to load resources when you need them, that is how the web works. To accomplish this you would need to go through almost all routes in the app and change them.

We could make on-demand the default but then you have the opposite problem if you for some reason want to load everything upfront.

I will see if we can make it just a property instead, with some kind of wildcard matching

caalador:
Or perhaps an `AppShellConfigurator` annotation/method?
Properties are always complicated when not using spring, so programmatic might be better.

Artur-:
Defining the empty string as one value is always problematic with properties also.. `<eagerRoutes>"", "login"</eagerRoutes>` look odd as a Maven option too, nested tags become complex..

If it is an app shell annotation, it could even refer to route classes like
```
@LoadDependenciesOnStartup({ MainView.class, LoginView.class });
```
but you would want a way to get the old behavior of everything eager without listing all classes... Would it work to have that behavior with only the annotation? 
```
@LoadDependenciesOnStartup
```

The default of having no annotation would then be defined by Flow so something that is applicable to all applications.

caalador:
Written out like this I feel it would work quite well.
Also not giving any routes matching to take all would probably be quite fine also.

Artur-:
Updated

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -491,6 +558,31 @@ private void computePackages() throws ClassNotFoundException, IOException {
...
+            throws ClassNotFoundException {
+        FrontendAnnotatedClassVisitor loadDependenciesOnStartupVisitor = new FrontendAnnotatedClassVisitor(
+                getFinder(), LoadDependenciesOnStartup.class.getName());
+        Class<?> appShellConfiguratorClass = getFinder()
+                .loadClass(AppShellConfigurator.class.getName());
+        for (Class<?> hopefullyAppShellClass : getFinder().getAnnotatedClasses(
+                LoadDependenciesOnStartup.class.getName())) {
+            if (!Arrays.asList(hopefullyAppShellClass.getInterfaces())
+                    .contains(appShellConfiguratorClass)) {
+                throw new IllegalStateException(ERROR_INVALID_PWA_ANNOTATION);
```
caalador, line 571: This should perhaps be changed as I don't think it will tell much to get a
`"There can only be one @PWA annotation and it must be set on the "    + AppShellConfigurator.class.getSimpleName() + " implementor."` for LoadDependenciesOnStartup class

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
Artur-: The whole way of finding app shell in multiple ways in multiple places is horrible.. 

Fixed the text

caalador (approved):

PR #16694, opened 2023-04-26
fix: TaskUpdatePackages should use finder (#16692)

### LINKED ISSUE #16032: Version information warning when building for production

caalador:
The tasks should use finder for getting
resources instead of using a XYZ.class.getClassLoader() as that doesn't contain required content.

Fixes #16032


czp13 (approved):
I think we can ignore sonatype attacker exposure message, this one:
- `try (InputStream vaadinVersionsStream = coreVersionsResource
                .openStream()) {`
                
as coreVersionResource is not coming from a potential attacker, browser request, but from the finder/classFiender:
- `coreVersionsResource = finder
                .getResource(Constants.VAADIN_CORE_VERSIONS_JSON`
- and only the `vaadin-core-version.json` file is looked for.         
     
The change itself is nice, the related issue was in my name as I spent some hours investigating it, but my stashed changes were much more complex and not so straightforward. I was using:
- `ClassFinder.CachedClassFinder(options.getClassFinder());` 

and for it to make it work options would have been needed, and so more refactor, test fixes, and manual tests, that is why I postponed the PR creation.

This is a much cleaner solution, thanks for it.
Approved it.

Issue #16671, opened 2023-04-24
Access application resources from Vaadin plugin


mshabarov:
### Describe your motivation

Suppose I want to get frontend resources (for instance, Vaadin web component) from inside the Vaadin Maven or Gradle plugin with `getClassLoader().getResources("META-INF/frontend/some-resource.js")` when I run `build-frontend` goal, i.e. these resources included as application's dependencies, not a dependencies of Vaadin plugin.
I cannot do it for the time being, because they are invisible for plugin's class loader.

### Describe the solution you'd like

Add resources defined in application's `pom`/`build` dependencies to the class-path before running Vaadin plugin, so they would be available.

### Additional context

This is needed in particular for re-bundle checker for production bundle: the checker's code runs within the `build-frontend` goal, i.e. plugin's code, and needs to load frontend resources, defined by application, to be able to compare with what is already included in the bundle.

PR #16701, opened 2023-04-26
feat: Use class finder to access project resources and classes

mshabarov:
Depends on https://github.com/vaadin/flow/pull/16653

ClassFinder is used instead of regular class loader in the Flow plugins to access the resources/classes defined and imported to the project's build file, for instance, Vaadin components.

Thanks @caalador for essential code changes and testing.

Fixes https://github.com/vaadin/flow/issues/16671


mshabarov:
> Seems good to me, but maybe safe just to commit the last one:
> 
> * [feat: Use class finder to access project resources and classes](https://github.com/vaadin/flow/pull/16701/commits/48ba6c91cfd8f35270f9d1003be451a09ea86223)
> 
> Maybe this branch was created from the `use-default-prod-bundle` caused the other commits here related to that PR? 🤔

Yes, other commits are unrelated and should be handled in another PR. I've just made this PR to check that the last commit with the class finder lets all tests pass.

mshabarov:
Included into https://github.com/vaadin/flow/pull/16653

PR #16708, opened 2023-04-27
feat: deprecate Label and add NativeLabel as replacement

### LINKED ISSUE #3532: Warn users about incorrect usage of Label
When I incorrectly use `Label` in V10, I want to get a warning message. 

In Vaadin 8 and earlier `Label` was used for text content. In V10, it maps to `<label>`. This will likely lead to a lot of incorrect markup that's possibly confusing for screen readers. 

Suggestion: Log a warning in the console if a `Label` is used without `for` or an `input` child.

### LINKED ISSUE #4384: Rename the current Label component to NativeLabel and create a new Label that is compatible with FW8 Label
Currently the Label component creates a `<label>` element in the client, which makes it not suitable to rendering loose text in the page, since the `<label>` in meant to be used in conjunction with other elements (usually `<input>`). This makes it semantically different from the old Label in FW8, which creates a `<div>` instead.

On top of that, the `<label>` element is an [interactive content](https://www.w3.org/TR/2013/CR-html5-20130806/dom.html#interactive-content) element, which makes it behave differently than a regular text in the page. For example, when used inside a Grid cell, clicking on a label doesn't trigger the selection of the row (see https://github.com/vaadin/flow/issues/4313 for details).

My proposal to this issue is the following:

1. Rename the current class from `Label` to `NativeLabel` (using the same naming structure as the `NativeButton`);
2. Create a new `Label` class that behaves exactly like the old Label from FW8;
3. Update demos and starters to use the new Label class;
4. Update the documentation explaining the difference between those two classes.

knoobie:
## Description

Adds `NativeLabel` as direct replacement for the HTML-based `Label` component / element. 

Fixes #4384
Fixes #3532

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-html-components/src/main/java/com/vaadin/flow/component/html/Label.java
```diff
@@ -48,6 +48,7 @@
  * @since 1.0
  */
 @Tag(Tag.LABEL)
+@Deprecated(since = "24.1", forRemoval = true)
```
tltv, line 51: This deprecation should be documented well in Javadoc. Deprecating `Label` alone may not help those who are doing migration from old versions. 
How about adding Javadoc with `@deprecated` telling shortly why `Label` is deprecated and hint to use `Span` or `Div` instead to show text label and `NativeLabel` only if also `for` attribute is used.

knoobie: So, I take this comment as: 'Go for it.'? See the initial PR comment. For further information regarding this change, see the discussion and last comments in https://github.com/vaadin/flow/issues/3532

mshabarov: @knoobie @tltv thanks for the contribution and review! And sorry for a long reply, I took a break to dig into context. I like the idea, and see no obstacles to do it. I've heard a talks about usage of `Label` quite many times, so the misunderstanding of it turns into an issue in our framework. 

Not sure we can target this to Vaadin 24.1, because only few days remain to Beta release, so I propose to thoughtfully complete this feature and target it to 24.2. 

So in short, I vote to continue with this pull request, add a warning and explanatory javadoc. We can auto-replace all the usages of `Label` to `NativeLabel` in Flow tests at this point I think. Replacing it with `Span` would take an extra efforts.

Ping @Legioth just in case, but this feature is in accordance with his suggestion, so I suppose move forward.

mshabarov: I'm going to replace `!` from the title, because it is not a breaking change so far.

knoobie: I'm going to finish the PR on the weekend to get it rdy for you guys.

knoobie: Added proper deprecation notice and refactored tests to use `NativeElement` and `NativeElementLabel`. Should be ready for a proper review now.

knoobie: > add a warning

Added a warning about wrong usage of the Label in https://github.com/vaadin/flow/pull/16708/commits/4644275c69dbed49242afc951c41e1096e5227fd. This should only be logged in development mode, so that this does not create problems for Ops Teams and the Developer isn't bad-mouthed by the framework in front of their clients in production logs.. I was considering sanitizing `getText()` before logging it, but it wasn't really worth it, because it's only logged in development mode, where it shouldn't be problem.

flow-html-components/src/main/java/com/vaadin/flow/component/html/Label.java
```diff
@@ -118,4 +128,51 @@ public void setFor(String forId) {
     public Optional<String> getFor() {
         return get(forDescriptor);
     }
+
+    private static Boolean productionMode = null;
```
tltv, line 132: Move to top of the class next to the other static field.

flow-html-components/src/main/java/com/vaadin/flow/component/html/Label.java
```diff
@@ -118,4 +128,51 @@ public void setFor(String forId) {
...
+            return; // skip check in production so that customer / clients /
+                    // ops-teams are not complaining about this warning to the
+                    // devs. This should be dealt with by devs in development
+                    // mode.
+        }
+        // Label was not associated with a for-attribute
+        // AND
+        // Label was not associated by adding a nested component
+        if (getFor().isEmpty() && getChildren().findAny().isEmpty()) {
+            LoggerFactory.getLogger(Label.class.getName())
```
tltv, line 147: Showing a warning only when label is detached is a bit late. It would be great to get this warning earlier.
Instead of doing this on detach, how about doing it earlier in following way on attach:
```
@Override
   super.onAttach(attachEvent);
   if (!isProductionMode(attachEvent)) {
      onetTimeExecutionRegistration = attachEvent.getUI().beforeClientResponse(this, ctx -> {
         onetTimeExecutionRegistration.remove();
         
         // show warning here when needed
      });
   }
}
```

tltv (changes requested):

flow-html-components/src/main/java/com/vaadin/flow/component/html/Label.java
mshabarov, line 147: I suppose the same warning should be added to `NativeLabel`, as it is valid/expected that even `NativeLabel` might be used as a plain text erroneously.

tltv: Maybe also limit this to only to initial attach with `attachEvent.isInitialAttach()`.

knoobie: done

knoobie: Once you guys are happy with the impl in Labe, I'm going to copy paste it to NativeLabel. (Even tho, I personally would not add it there; people using the new `NativeLabel` should know what they do (hopefully))

flow-html-components/src/main/java/com/vaadin/flow/component/html/Label.java
```diff
@@ -118,4 +133,55 @@ public void setFor(String forId) {
...
+    protected void onAttach(AttachEvent attachEvent) {
+        super.onAttach(attachEvent);
+        if (skipForAttributeCheck(attachEvent)
+                || !attachEvent.isInitialAttach()) {
+            return; // skip check in production so that customer / clients /
+            // ops-teams are not complaining about this warning to the
+            // devs. This should be dealt with by devs in development
+            // mode.
+        }
+        checkForAttributeOnAttach = attachEvent.getUI()
```
mshabarov, line 147: I suppose that before adding a new listener by `beforeClientResponse`, it should check
```java
if (checkForAttributeOnAttach == null) { 
```
otherwise I don't understand why we use `beforeClientResponse` at all.

flow-html-components/src/main/java/com/vaadin/flow/component/html/Label.java
mshabarov: Yes, good comment. For `NativeLabel` we can skip adding a warning.

mshabarov:
The build validation stops because of compilation errors:
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project flow-html-components: Compilation failure
Error:  flow/flow-html-components/src/main/java/com/vaadin/flow/component/html/Label.java:[184,33] cannot access jakarta.servlet.http.HttpSessionBindingListener
Error:    class file for jakarta.servlet.http.HttpSessionBindingListener not found
```
This pull request doesn't add.import this class, this is weird.

flow-html-components/src/main/java/com/vaadin/flow/component/html/Label.java
knoobie: done

flow-html-components/pom.xml
```diff
@@ -26,6 +26,11 @@
             <artifactId>flow-data</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>jakarta.servlet</groupId>
+            <artifactId>jakarta.servlet-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
knoobie, line 33: @mshabarov 

> he build validation stops because of compilation errors

Looks like the culprit is the usage of `VaadinSession` because it has `jakarta` imports.. My first try would be to add the servlet-api as provided, if you don't like. We could think about changing the logik to get the production mode information from `VaadinService.getCurrent().getDeploymentConfiguration().isProductionMode()` instead of using the AttachEvent.

tltv: I like usage of `VaadinService` as you suggested as long as it helps to avoid adding new dependency. Otherwise import should be fine too.

knoobie: Done

tltv (approved):

mshabarov (approved):

knoobie:
@mshabarov I just realized that there are a lot of usages within the flow-components that are still using the old Label and polluting the logs because they are not associated on the server side, but on the client side thanks to the slot=label Attribute. I'm wondering if we should add this condition also to the list of valid usages to reduce the amount of work the component team has to do urgendly to remove the log spam for users. 

Example FormLayout or RadioGroup

PR #16709, opened 2023-04-27
fix: Lazy load CSS also

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
```diff
@@ -271,20 +286,17 @@ String resolveGeneratedModule(String module) {
         return FrontendUtils.FRONTEND_GENERATED_FLOW_IMPORT_PATH + module;
     }
 
-    protected List<String> getCssLines(Map<ChunkInfo, List<CssData>> css) {
+    protected List<String> getCssLines(List<CssData> css) {
```
caalador, line 289: The index for cssLine should be a rising global as now we will get 0 for each css key

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
caalador: This should now be fine in master, but needs conflict resolution.

Artur-: Is it fine to have
```
import $cssFromFile_0 from 'Frontend/foo.css?inline';
injectGlobalCss($cssFromFile_0.toString(), 'CSSImport end', document);
```

in multiple JS files?

caalador: I think it should be fine to have the same import identifier in different files, but ofc not having the same numbering might be better.

Artur-: The question is not really the numbering but does `injectGlobalCss` handle deduplication and does having the import in multiple files cause overhead or not? I think the answer is "yes" and "does not cause overhead" but maybe it needs to be checked

caalador: Aa. The css is checked for if it has been injected with the exact same hash when target is document so it handles duplication. So overhead is duplicate data transmitted/stored in bundle and hash generation.

Artur-: Ok, then this version could work

caalador (approved):

PR #16721, opened 2023-05-02
chore: pinned maven enforcer plugin version

mcollovati:
maven-enforecer-plugin version is set in the pluginManagement section of the POM file, but for some reason the version is ignored when running the build with release profiles active.
This change pins the plugin version also in the plugin section.


czp13 (approved):
Fixed the enforcer errors for me. Thanks and nice Marco!

mcollovati:
The reason for the "wrong" enforcer plugin version is that the version is explicitly set to `3.0.0` by the `release` profile in  `vaadin-parent` POM

PR #16723, opened 2023-05-02
feat: Make "" and "login" eager and the rest lazy by default

Artur-:


flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractNodeUpdateImportsTest.java
```diff
@@ -41,7 +41,9 @@
 import org.mockito.Mockito;
 import org.slf4j.Logger;
 
+import com.vaadin.flow.component.page.AppShellConfigurator;
 import com.vaadin.flow.di.Lookup;
+import com.vaadin.flow.server.LoadDependenciesOnStartup;
```
czp13, line 46: Not used:
```suggestion
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -398,6 +399,30 @@ private boolean isEagerRoute(Class<?> route) {
         return this.eagerRoutes.contains(route.getName());
     }
 
+    private boolean defaultIsRouteEager(Class<?> route) {
```
knoobie, line 402: Should you also try to get the "real" login component configured within https://github.com/vaadin/flow/blob/d54e6af4b096b8d0ccf8dd4a5d70a5917fa33867/flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java

or would that be to much magic?

Artur-: That is not available during a production build

mshabarov (approved):

Issue #16668, opened 2023-04-22
Remove deadcodes .


czp13:
### Description of the bug

I believe there are some opportunities to remove unused codes.

Similar to this PR:
- https://github.com/vaadin/flow/pull/16665

This dead code was blocking to update to the newest 4.9.0 selenium (and to any new selenium versions after this).
Tests can be a low-hanging fruit, but maybe there are some other utils or other classes which are easier to remove dead code.

Short analysis exacuted shown 562 opportunities:
<img width="684" alt="image" src="https://user-images.githubusercontent.com/61667986/233798559-d0c2047e-54ca-4994-b017-12ef279518ac.png">

Probably not all of them can be removed as some of them part of abstract classes which used elsewhere, but this one for example can be removed:

```

GwtEventHandlerTest.assertPolymerMethods

JsonUtilsTest
.createNumberStreamForNull()
.createObjectStreamForNull()
.createStreamForNull()
```

And so on...

### Expected behavior

The repository could contain fewer unused code lines, which makes:
- dependency addition, removal, bumping,
- refactoring, changes also easier,
- and just to overview understand the logic, code base better as code is simplified a bit.
- potentially builds/compilation time can be faster if we remove a lot of dead codes, although by not much probably :D

### Minimal reproducible example

In IntelliJ:
 Analyze | Run Inspection by Name... | Unused declaration.
 (Potentially worth to add Test regexp or Util regexp, as there will be more false positives otherwise)

### Versions

- Vaadin / Flow version: latest

PR #16745, opened 2023-05-05
test: Delete ignored and non-functional test

### LINKED ISSUE #2119: Fix memory leak tests
`test-memory-leaks` module started to fail randomly hence was marked with @Ignore junit annotation.
Need to decide what to do with it and either remove or fix it.

Artur-:
Fixes #2119


czp13 (approved):
Very nice catch, also progress with this one:
- https://github.com/vaadin/flow/issues/16668

Issue #16718, opened 2023-05-02
Support for custom theme when pre-compiled prod bundle is used


mshabarov:
Since application specific styles/custom theme cannot be included into pre-compiled prod bundle, Flow should manage the custom theme resources and apply theme to the application separately from the pre-compiled production bundle in the same way as it is done for pre-compiled development mode (by adding style and link tags to the head of document).

PR #16751, opened 2023-05-05
feat: Enable custom theme for pre-compiled prod bundle

mshabarov:
Uses the same mechanism with adding style tags to `head` to style application when the pre-compiled prod bundle is used, as Flow does for dev bundle.
Copies project's custom theme files to the output folder, so they are available in the runtime in class-path.
Also loads application's theme from AppShellRegistry.

Fixes https://github.com/vaadin/flow/issues/16718


flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
```diff
@@ -62,30 +65,38 @@ private static Logger getLogger() {
...
      * @return custom theme name or empty optional if no theme is used
      * @throws IOException
      *             if I/O exceptions occur while trying to extract the theme
      *             name.
      */
-    public static Optional<String> getThemeName(File projectFolder)
-            throws IOException {
-        File themeJs = new File(projectFolder, FrontendUtils.FRONTEND
-                + FrontendUtils.GENERATED + FrontendUtils.THEME_IMPORTS_NAME);
+    public static Optional<String> getThemeName(VaadinContext context,
```
Artur-, line 77: Does it make sense to pass the config separately as you can get it using `ApplicationConfiguration.get(context)`

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -182,9 +182,11 @@ public boolean synchronizedHandleRequest(VaadinSession session,
     private static void addDevBundleTheme(Document document,
             VaadinContext context) {
         ApplicationConfiguration config = ApplicationConfiguration.get(context);
-        if (config.getMode() == Mode.DEVELOPMENT_BUNDLE) {
+        if (config.getMode() == Mode.DEVELOPMENT_BUNDLE
+                || config.getMode() == Mode.PRODUCTION) {
             try {
-                BootstrapHandler.getStylesheetTags(config, "styles.css")
+                BootstrapHandler
```
Artur-, line 188: Wouldn't this always add the stylesheet in production mode, also when a custom production bundle was built?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskPrepareProdBundle.java
```diff
@@ -16,31 +16,76 @@
...
 
     private final Options options;
 
-    public TaskCopyBundleFiles(Options options) {
+    private final FrontendDependenciesScanner frontendDependenciesScanner;
+
+    public TaskPrepareProdBundle(Options options,
+            FrontendDependenciesScanner frontendDependencies) {
         this.options = options;
+        this.frontendDependenciesScanner = frontendDependencies;
```
Artur-, line 57: Where is the scanner needed?

flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
```diff
@@ -62,30 +65,38 @@ private static Logger getLogger() {
...
+            AbstractConfiguration config) throws IOException {
+        if (config.isProductionMode()) {
```
Artur-, line 79: Why should the theme lookup work differently in dev and prod mode?

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
mshabarov: Because theme name is needed in runtime, and in production mode you don't know anything about project files, also don't have a generated theme file where the theme name is placed.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
mshabarov: Currently yes, and this is not correct. Should use stylesheet only for pre-compiled bundle. I'm looking for a good way to pass this info (pre-compiled bundle or custom) into runtime. Probably an extra entry in `stats.json` would work.

flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
mshabarov: Yes, I think it would be better. Will fix it.

mshabarov: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskPrepareProdBundle.java
mshabarov: Not needed. This was a leftover from previous versiins.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
mshabarov: Added a new entry `pre-compiled` to the stats. Not sure if we need a new enum name for pre-compiled bundle, like `Mode.PRODUCTION_PRE_COMPILED`, having also `Mode.PRODUCTION` that means custom bundle.

flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
```diff
@@ -137,23 +184,22 @@ public static Optional<String> getParentThemeName(JsonObject themeJson) {
...
      * @param config
      *            the application configuration
      * @return a list of active themes, in parent to child order
      */
-    public static List<String> getActiveThemes(AbstractConfiguration config)
-            throws IOException {
-        File projectFolder = config.getProjectFolder();
-        Optional<String> applicationTheme = getThemeName(projectFolder);
+    public static List<String> getActiveThemes(VaadinContext context,
+            AbstractConfiguration config) throws IOException {
```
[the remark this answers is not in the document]
mshabarov: No. Removed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
mshabarov: Extra config parameter exluded.

flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
mshabarov: Reverted back the code for extracting from generated file, because frontend live reload initialiser doesn't get the theme name when trying to get it from annotation (in DevModeHandlerManagerImpl).

flow-tests/test-express-build/pom.xml
```diff
@@ -42,7 +43,9 @@
                 <module>test-dev-bundle-java-add-on</module>
                 <module>test-reusing-theme-express-build</module>
                 <module>test-parent-theme-express-build</module>
+                <module>test-parent-theme-express-build/pom-prod.xml</module>
```
caalador, line 46: Is there any other wat o actually have the same sources for multiple different modules?
Here also both poms should start with cleaning any generated files so that for instance the production doesn't by mistake use dev files and vice versa.
Also this makes at least idea have bad module handling (could be other ides also, but I don't use any other at the moment)

flow-tests/test-express-build/test-parent-theme-express-build/src/test/java/com/vaadin/flow/devbuild/ParentThemeIT.java
```diff
@@ -52,18 +53,24 @@ public void init() {
...
-        waitUntilExternalAsset();
+        try {
+            findElement(By.tagName("vaadin-dev-tools"));
+            // app dev bundle has reusable theme in stats.json
+            verifyThemeJsonHashReusableThemeRecord();
+            // check that the bundle has the expected asset for dev mode
+            verifyExternalAssetInBundle();
+            waitUntilExternalAsset();
+        } catch (NoSuchElementException e) {
+            // ignore in prod mode
```
caalador, line 72: The need for these not in X mode would warrant having own modules for prod and dev modes.

mshabarov: It's better to make its own modules and codes for prod mode. Taking into account possible additions in the future, I think this would be a better way. Will fix.

mshabarov: done

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleUtils.java
```diff
@@ -113,6 +113,18 @@ public static String getChunkId(String className) {
...
     }
 
+    /**
+     * Returns whether the application uses pre-compiled production bundle or a
+     * custom bundle.
+     *
+     * @return <code>true</code> in case of pre-compiled bundle,
+     *         <code>false</code> otherwise
+     */
+    public static boolean isPreCompiledBundle() {
```
Artur-, line 123:
```suggestion
    public static boolean isPreCompiledProductionBundle() {
```

Artur-, line 123:
```suggestion
    public static boolean isPreCompiledProductionBundle() {
```

or is this also for dev bundle?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskPrepareProdBundle.java
```diff
@@ -0,0 +1,122 @@
...
+    private void copyBundleFilesFromJar() {
+        URL statsJson = BundleValidationUtil.getProdBundleResource(
+                "config/stats.json", options.getClassFinder());
+        if (statsJson == null) {
+            throw new IllegalStateException(
+                    "Could not copy production bundle files, because couldn't find production bundle in the class-path");
+        }
+        String pathToJar = statsJson.getPath();
+        int index = pathToJar.lastIndexOf(".jar!/");
+        if (index >= 0) {
```
Artur-, line 92: So if the path to statsJson does not contain `.jar!/` then stats.json is the jar?

Seems like this code has not changed though..

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskPrepareProdBundle.java
```diff
@@ -0,0 +1,122 @@
...
+            pathToJar = pathToJar.substring(0, index + 4);
+        }
+        try {
+            URI jarUri = new URI(pathToJar);
+            JarContentsManager jarContentsManager = new JarContentsManager();
+            jarContentsManager.copyIncludedFilesFromJarTrimmingBasePath(
+                    new File(jarUri), Constants.PROD_BUNDLE_NAME,
+                    options.getResourceOutputDirectory(), "**/*.*");
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
```
Artur-, line 103: Why `RuntimeException` when the the method below throws `ExecutionFailedException`?

flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
Artur-: Why doesn't it get it?

flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
```diff
@@ -105,10 +119,58 @@ public static Optional<Theme> getThemeAnnotation(VaadinContext context) {
...
+        return ThemeUtils.class.getClassLoader()
+                .getResource(Constants.RESOURCES_JAR_DEFAULT + themeAssetPath);
+    }
+
+    public static URL getThemeResourceFromBundle(String themeAssetPath) {
+        return ThemeUtils.class.getClassLoader().getResource(
+                VAADIN_WEBAPP_RESOURCES + "VAADIN/static/" + themeAssetPath);
+    }
+
+    public static Optional<JsonObject> getThemeJson(String themeName,
```
Artur-, line 167: Looks like you are leaking `Options` outside node tasks. Pass the frontend directory instead

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -327,6 +334,16 @@ && isIncorrectWebjarPath(filenameWithPath)) {
         return true;
     }
 
+    private static URL findAssetInProdBundleOrClassPath(String assetPath) {
+        // lookup in the prod bundle, where themes are copied from project's
+        URL resourceUrl = ThemeUtils.getThemeResourceFromBundle(assetPath);
```
Artur-, line 339: Maybe `getThemeResourceFromBundle` and `getThemeResourceFromJar` should be combined into `getThemeResourceFromPrecompiledProductionBundle`?

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -388,14 +390,17 @@ protected void writeBootstrapPage(String contentType,
 
         DeploymentConfiguration config = response.getService()
                 .getDeploymentConfiguration();
-        if (config.getMode() == Mode.DEVELOPMENT_BUNDLE) {
+        VaadinContext context = response.getService().getContext();
+        if (config.getMode() == Mode.DEVELOPMENT_BUNDLE
+                || (config.getMode() == Mode.PRODUCTION
```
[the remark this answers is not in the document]
mshabarov: Good idea! Done

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -276,6 +278,11 @@ public boolean serveStaticResource(HttpServletRequest request,
                         deploymentConfiguration.getProjectFolder(),
                         filenameWithPath.replace(VAADIN_MAPPING, ""));
             }
+        } else if (deploymentConfiguration.getMode() == Mode.PRODUCTION
+                && BundleUtils.isPreCompiledBundle()
+                && APP_THEME_PATTERN.matcher(filenameWithPath).find()) {
+            resourceUrl = findAssetInProdBundleOrClassPath(filenameWithPath
```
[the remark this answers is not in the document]
mshabarov: The themes located in the project frontend are copied to the same `VAADIN/static/themes` directory as it's done with the Vite application theme plugin. Should work, but not tested. I also aligned the copy logic with the plugin to take `webappResourcesDirectory` parameter into account.

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleUtils.java
mshabarov: Done

mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskPrepareProdBundle.java
mshabarov: Replaced by EFE

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
mshabarov: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskPrepareProdBundle.java
mshabarov: The expectation is always have stats.json in the JAR always at this point of app execution.
So maybe we should `assert` instead of `if()`

flow-server/src/main/java/com/vaadin/flow/server/frontend/ThemeUtils.java
mshabarov: I'm not sure why. If I run in debug mode and stop at the point when DevModeHandlerManager tries to fetch the themes, I see that the annotation way works and then after app start-up the live reload also works. If I run without dubugging, I always get it not working.
Thus, I suspect this is a timing issue when it's fetched from annotation, because dev server starts in parallel with the servlet initialisation and we may not have the theme information available sometimes.

Artur-: ok so it could be improved in a separate PR at some point

mshabarov:
I made an issue about static resource here https://github.com/vaadin/flow/issues/16780 . So propose to fix it in a separate patch.

Artur- (approved):

PR #16757, opened 2023-05-09
feat: add parameter to force prod build

### LINKED ISSUE #16752: Add a new config parameter for always compiling a new prod bundle
Add a new config parameter to Flow that allows bypass the bundle checker and always build a new production bundle, to let developers have a fallback in case if they experience issues with the pre-compiled bundle, or if they know in advance that a new bundle is always an option for them.
Acceptance criteria:
- [x] a new config parameter `vaadin.force-new-build` (or similar name) is added
- [x] when it's set to `true`, Flow always makes a new production bundle and skips bundle checker. When `false`, it rebuild a new bundle or reuses the pre-compiled one based on checker resolution.
- [x] for Hilla projects this parameter is `true`, so a new bundle is always needed.

TBD: This parameter might be combined with the existing `skip.dev.bundle` and `frontend.hotdeploy` parameters, for instance, to always start Vite dev server and rebuild a new prod bundle with a same single parameter.

caalador:
Add a parameter that forces a
production bundle build even
when an applicable bundle exists.

Closes #16752

### DIFF
--- a/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
+++ b/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
@@ -210,6 +210,11 @@ public class BuildDevBundleMojo extends AbstractMojo
         return false; // ci build not applicable for dev mode
     }
 
+    @Override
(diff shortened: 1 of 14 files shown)


mshabarov (dismissed):

mshabarov (approved):

mshabarov:
@caalador please don't forget to add this parameter to the documentation.

PR #16758, opened 2023-05-09
fix: Restore JRebel support

Artur-:
JRebel detects `BrowserLiveReloadAccess` and uses it to trigger the browser reload


jcgueriaud1 (approved):
Is it still a deprecated class? since it's useful at least for JRebel.

Artur-:
There is no other use for it. Also JRebel should move to use `BrowserLiveReloadAccessor`

PR #16765, opened 2023-05-10
VaadinService: Avoid silently clearing the thread interrupted state

### LINKED ISSUE #16764: VaadinService#ensureAccessQueuePurged silently clears thread interrupted state
### Description of the bug

VaadinService#ensureAccessQueuePurged contains try/catch block that catches InterruptedException, but doesn't restore the interrupted state. This has the effect of silently swallowing Thread interruption which makes it impossible for threads that use, for instance, UI#access to correctly be interrupted.

### Expected behavior

UI#access should not silently clear the interrupted state.

This snippet sketches a bit of UI that starts some work on a separate thread. The worker thread periodically updates the UI using UI#access. A cancel button aborts the background task by calling Future#cancel and allow thread interruption.

This currently does not work reliably because UI#access may silently discard the interrupted state.

```
UI ui = UI.getCurrent();
Label label = new Label();

Future<Object> future = Executors.newSingleThreadExecutor().submit(() -> {
    while (!Thread.interrupted()) {
        // do some work
...

        // Update progress indicator
        ui.access(() -> label.setText(Instant.now().toString()));
    }
});

Button cancel = new Button();
cancel.addClickListener(event -> {
    future.cancel(true);
});
```

### Versions

- Vaadin / Flow version: 24.0.5
- Java version: 17.0.7 (not relevant to issue)
- OS version: macOS 13.3.1 (not relevant to issue)
- Browser version (if applicable): not applicable
- Application Server (if applicable): not applicable
- IDE (if applicable): not applicable

pepijnve:
## Description

Instead of silently catching the thrown InterruptedException, restore the interrupted state of the current thread. This allows calling code to correctly inspect the interrupted state and handle it.

Fixes #16764

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


pepijnve:
I only see 'This job failed' for the integration tests. Is running these locally a matter of running the corresponding maven task(s)?

czp13:
This is a very nice change @pepijnve. It is much appreciated that:
-  you created an issue, with -I think- a quite frequent use case,
- and also opened a PR,
Thank you! 🙇 

I think the only thing that is missing is to somehow write a unit test/integration test to show that this change is working indeed (it should be working in your example). Could you try out maybe in your custom app change with your snippet/code?

pepijnve:
I'm afraid I don't have time right now to set up a full reproduction app. In our production application we do have code similar to the snippet I provided. Calls to UI#access are all made through a single static helper method in that codebase. I've inlined UI#access and the methods it calls in that method and applied the suggested patch as a quick fix and can confirm that this resolves the issue.

czp13:
Thanks @pepijnve, sure I understand, basically this manual testing is:
1. build manually the flow jar to your maven local repository
2.  include it in your project (proper artifact version in pom.xml dependencies),
3. and add the code snippet/view and test.
(Just for future reference, if needed you can try out this way if your fix/changes are working or not).

 I can test this change in my custom project to help out as you help out as well :). And if we do not have any idea where to put the tests I can as well investigate a bit more and see where we could put such a test. Probably later this week I can pick this up (manual test, potentially adding tests).

czp13:
We have VaadinServiceTest so probably it would be a good place to put some covering tests there. 
I am on it to help close this ASAP: 
- manual test, 
- and creating covering test if possible easy

czp13:
### Tested in custom project.

Based on your idea with a slightly changed code snippet, this:
```
        Label label = new Label();

        Future<?> future = Executors.newSingleThreadExecutor().submit(() -> {
            while (!Thread.interrupted()) {
                // do some work
                for (int i = 0; i < 10000; i++) {
                    // do nothing
                }
                System.out.println("For loop done");
                // Update progress indicator
                ui.access(() -> label.setText(Instant.now().toString()));
                System.out.println("UI updated");
            }
        });

        Button cancel = new Button();
        cancel.addClickListener(event -> {
            future.cancel(true);
        });

        add(label, cancel);

```

### **With the fix**
thread_interrupted_after

### **Without the fix**
thread_interrupted_before

**The key differences between the two versions are:**
- the long-running job cannot be interrupted with thread interrupt (flag is not set)
- showing that you can push multiple times click, but still, the date will be updated as the job is running constantly
- note: probably not the best use case, as this is constantly running on the server making it slow, but I can imagine other use cases where interrupting a thread and handling the interrupted state properly can be useful

### About test coverage:
I spent some hours, and I had some ideas to do it (mock object and throw an exception and see in both cases it works properly), but unfortunately, we would need to mock `VaadinService`, and then mock ensureAccessQueuePurged method, but because VaadinService contains static fields, methods `Mockito.mock` cannot mock it properly,
We have two solutions:
1. Refactor VaadinService to put static parts into a util method, which could be a nice addition at some point potentially, but it can be a breaking change to many, so it needed more careful consideration and decision, not a rushed one.
(in mind that some tests try to mock VaadinService, but cannot, so it is a mistake in the code, confusion)
2. Using `PowerMock`, which is capable of mocking in these cases potentially, adding a new dependency is also troublesome.

So because of manual testing and code review,  I would accept and merge this change!

PR #16766, opened 2023-05-10
fix: not require stats.json output when generateBundle is disabled

### LINKED ISSUE #16763: Unable to run mvn vaadin:build-frontend when generateBundle is disabled
### Description of the bug

When building my application for production, I want to have some custom executions of Maven plugins between `npm install` and frontend bundling steps. For this purpose, I set up two executions for `build-frontend` goal:
```xml
<plugin>
  <groupId>com.vaadin</groupId>
  <artifactId>flow-maven-plugin</artifactId>
  <executions>
    <execution>
      <!-- Only installs / updates npm dependencies -->
      <id>default</id>
      <goals>
        <goal>prepare-frontend</goal>
        <goal>build-frontend</goal>
      </goals>
      <phase>compile</phase>
      <configuration>
        <generateBundle>false</generateBundle>
      </configuration>
    </execution>
    <execution>
      <!-- Only does the production bundle -->
      <id>main-frontend-build</id>
      <goals>
        <goal>build-frontend</goal>
      </goals>
      <phase>process-classes</phase>
      <configuration>
        <nodeAutoUpdate>false</nodeAutoUpdate>
        <runNpmInstall>false</runNpmInstall>
        <productionMode>true</productionMode>
      </configuration>
    </execution>
  </executions>
</plugin>
```

This throws an exception in the first execution with the message: “No production bundle stats.json available.”

### Expected behavior

The build works.

See https://github.com/vaadin/hilla/tree/main/packages/java/tests/spring/endpoints-maven-engine

Run `mvn spring-boot:run` to reproduce

### Versions

- Vaadin / Flow version: Flow v24.1.0.alpha5 and v24.1.0.alpha6
- Java version:
- OS version:
- IDE (if applicable):

platosha:
Fixes #16763

### DIFF
--- a/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/BuildFrontendMojoTest.java
+++ b/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/BuildFrontendMojoTest.java
@@ -142,12 +142,6 @@ public class BuildFrontendMojoTest {
                 "jar-resources-source/META-INF/frontend");
         jarResourcesSource.mkdirs();
 
-        File statsfile = new File(resourceOutputDirectory,
-                Constants.VAADIN_CONFIGURATION + "/stats.json");
-
-        statsfile.getParentFile().mkdirs();
-        FileUtils.fileWrite(statsfile, "UTF-8", "{}");
-
         projectFrontendResourcesDirectory = new File(npmFolder,
                 "flow_resources");
 
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -513,7 +513,7 @@ public class BuildFrontendUtil {
 
         if (statsJsonContent == null) {
             // without stats.json in bundle we can not say if it is up-to-date
-            getLogger().debug("No production bundle stats.json available.");
+            getLogger().debug("No production bundle stats.json available for licenses validation.");
             statsJsonContent = "{}";
         }
 
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -513,8 +513,8 @@ public class BuildFrontendUtil {
 
         if (statsJsonContent == null) {
             // without stats.json in bundle we can not say if it is up-to-date
-            throw new RuntimeException(
-                    "No production bundle stats.json available.");
+            getLogger().debug("No production bundle stats.json available.");
+            statsJsonContent = "{}";
         }
 
         FrontendDependenciesScanner scanner = new FrontendDependenciesScanner.FrontendDependenciesScannerFactory()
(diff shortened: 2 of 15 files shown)


czp13:
I might be wrong, but `BuildFrontendMojo/VaadinBuildFrontendTask.kt` is for the build-frontend goal (maven/gradle), and with this change, the license check will only be executed if the bundle creation is on/happening:
``` 
    /**
     * Whether to generate a bundle from the project frontend sources or not.
     */
    @Parameter(defaultValue = "true")
    private boolean generateBundle;
```

The license check was put under this condition:
```
        if (generateBundle()) {
....
}
 ```
 
I would think a license check shall happen every time, not just related to bundle creation, so every time a frontend build is happening ` BuildFrontendUtil.runNodeUpdater(this);` the license check shall run, so the original condition, execution chain was good imho.

I think another change or refactor is needed to achieve/surpass this error: `No production bundle stats.json available.`. 
But I would need to test and debug it to see exactly what changes. 

I can test this theory (license check is not happening if the bundle is not created) if no one reviews, and confirms it soonish, this week.

**Review by czp13 — CHANGES_REQUESTED** [review]
It would be good to have some covering tests for the changed logic (now the relevant part was just deleted). 

And manually test through some cases (bundle creation is needed, or not, and if there was already a created bundle or not)

**Review by caalador — CHANGES_REQUESTED** [review]
This change is not ok as the license should be checked in all cases. There is a default bundle that is pacakged with the `vaadin-maven-plugin` that can always be checked against even when generate bundle is false.

This will need another type of fix which we need to ponder about for a bit @mshabarov

**Comment by caalador** [comment]
Actually it should be changed as artur noted to what it was here in the before https://github.com/vaadin/flow/pull/16728/commits/a7245c2818d5693a162f5c6e68b1f21b122753c8

File: flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java [hunk 1]
```diff
@@ -513,8 +513,8 @@ public static void validateLicenses(PluginAdapterBase adapter) {
 
         if (statsJsonContent == null) {
             // without stats.json in bundle we can not say if it is up-to-date
-            throw new RuntimeException(
-                    "No production bundle stats.json available.");
+            getLogger().debug("No production bundle stats.json available.");
```
mshabarov, line 516:
```suggestion
            getLogger().debug("No production bundle stats.json available for licenses validation.");
```

mshabarov (approved):
Looks good to me, so if no bundle generated/copied then nothing to check.

czp13 (approved):
I think with @caalador changes this seems good!
Only thing missing is test case when `statJson={}`, but I can approve to not block it if we feel like.

flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/BuildFrontendMojoTest.java
```diff
@@ -142,12 +142,6 @@ public void setup() throws Exception {
                 "jar-resources-source/META-INF/frontend");
         jarResourcesSource.mkdirs();
```
czp13, line 144: Should we put this back as the original change was reverted as I see?:
```suggestion

        File statsfile = new File(resourceOutputDirectory,
                Constants.VAADIN_CONFIGURATION + "/stats.json");

        statsfile.getParentFile().mkdirs();
        FileUtils.fileWrite(statsfile, "UTF-8", "{}");
```

Artur- (approved):

flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/BuildFrontendMojoTest.java
Artur-: It should not be needed now that `{}` is used when stats.json is not found

czp13: yes, correct, I was on autopilot, morning coffee needs before reviews!

PR #16769, opened 2023-05-11
fix: set Lookup attribute to servlet context from OSGi servlet context

tltv:
This change ensures that Lookup is set in servlet context attributes when initializing VaadinServlet in OSGi environment.

Fixes: vaadin/base-starter-flow-osgi/issues/296

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/OSGiVaadinServletTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/OSGiVaadinServletTest.java
@@ -20,6 +20,7 @@ import javax.servlet.ServletContext;
 import javax.servlet.ServletException;
 
 import java.util.Collections;
+import java.util.List;
 
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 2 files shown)


flow-server/src/test/java/com/vaadin/flow/server/OSGiVaadinServletTest.java
```diff
@@ -48,12 +49,50 @@ public void init_onlyLookupAttributeFromFakeOSGiContextAreSetFromServletContext(
...
+                .thenReturn(Collections.enumeration(servletContextAttributes));
+
+        VaadinServlet servlet = createVaadinServletStub();
+
+        servlet.init(config);
+
+        Mockito.verify(servletContext).setAttribute(Lookup.class.getName(),
+                context.getAttribute(Lookup.class.getName()));
+        Mockito.verify(servletContext, Mockito.times(0)).setAttribute("foo",
+                "bar");
```
czp13, line 87: nitty-picky, so very small thing:
I think these lines could be moved to the createVaadinServletStub method and potentially some other lines as well if I am correct.  (SerlvetConfig config maybe too).

I can try to refactor it a bit later to see we can deduct some lines, if you feel, or we can just merge it as well :) Fine with both ways.

mcollovati (approved):

czp13 (approved):

czp13 (approved):

mcollovati (approved):

Issue #8740, opened 2020-07-20
Exception in StreamResourceWriter or InputStreamFactory will set Internal Server Error after Response has been written


roeltje25:
### Description of the bug
In our application we generate reports which can be downloaded. We have use a StreamResource with a StreamResourceWriter (we also tried an InputStreamFactory with the same results) which will generate the output for the client. When the generation of the data generates an exception Vaadin sets the response status code to 500 Internal Server Error in StreamResourceHandler.java line 80 to 85 (vaadin 14.3.1). However in the try-with-resource block the OutputStream is closed BEFORE the catch block is entered. Closing the OutputStream causes the response to write to the client (at least in our WildFly server using undertow). The status code is then not picked up anymore by the browser resulting in a download popup which has an empty file.
### Minimal reproducible example
A minimal example can be found here: https://github.com/roeltje25/streamresourcehandlerissue
### Expected behavior
When the link in the minimal example is clicked the download should not start, and in the browser network log the request should return with status code 500.
### Actual behavior
When the link in the minimal example is clicked the download starts (with a popup in chrome e.g.), the status code in the network log is 200 and the downloaded file is empty.
### Versions:
    - Vaadin / Flow version: 14.3.1
    - Java version: 
      openjdk version "13" 2019-09-17
      OpenJDK Runtime Environment AdoptOpenJDK (build 13+33)
      OpenJDK 64-Bit Server VM AdoptOpenJDK (build 13+33, mixed mode, sharing)
    - OS version:
    - Browser version (if applicable): Firefox 78.0.2
    - Application Server (if applicable): Jetty built into maven (jetty-9.4.28.v20200408; built: 2020-04-08T17:49:39.557Z; git: ab228fde9e55e9164c738d7fa121f8ac5acd51c9; jvm 13+33)

roeltje25:
As an additional remark. It is not possible to easily add a global streamresource failure handler. For example to show a notification or things like that

denis-anisimov:
Confirmed.

I disagree with the expected behavior though a bit.
>When the link in the minimal example is clicked the download should not start

The download should start. The mentioned example is very specific: it throws an exception immediately but that doesn't have to happen generally. The exception may be thrown at any point: some part of file may be read and downloaded and an exception may be thrown somewhere in the middle.
We may not prevent downloading : it should always start.
But we should return a server error status in case there is an error in the stream.

roeltje25:
Dear Denis,

[...]

PR #16776, opened 2023-05-11
Fix homebrew try-with-resources to behave like it is supposed to.

archiecobbs:
## Description

The fix for #8740 is broken because it doesn't properly replicate the behavior of the try-with-resources block that it replaced.

In particular, if an exception is thrown, and then another exception is thrown when trying to `close()` the stream, the exception that gets propagated to the outside world is the bogus/meaningless `close()` exception, not the actual root cause exception. This makes debugging much harder because the root cause exception is being completely lost.

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist


archiecobbs:
If accepted, please back-port this to 23.x as well. Thanks.

mcollovati:
@archiecobbs thank you for the contribution.
The change looks good to me, and it is already covered by an existing test.
Can you please format the code, as reported in the above comment?

mcollovati:
Just one more thing. To avoid regressions in the future, may I ask you to update the `StreamResourceHandlerTest` test?

* adding an assertion to ensure the expected exception is propagated in `inputStreamFactoryThrowsException_responseStatusIs500` and `inputStreamResourceWriterThrows_responseStatusIs500`
  ```java
    } catch (RuntimeException exception) {
        Assert.assertEquals("Simulated", exception.getMessage());
    }
  ```
* add a new test that simulates an exception on `ServletOutputStream.close()` and verifies that the expected exception is propagated
  ```java
  @Test
  public void inputStreamResourceWriterAndResponseThrows_streamResourceWriterExceptionIsPropagated()
          throws IOException {
      ServletOutputStream servletOutputStream = Mockito.mock(ServletOutputStream.class);
      Mockito.when(response.getOutputStream()).thenReturn(servletOutputStream);
      Mockito.doThrow(new RuntimeException("Error on close")).when(servletOutputStream).close();
      StreamResource res = new StreamResource("readme.md",
              (StreamResourceWriter) (stream, session) -> {
                  throw new RuntimeException("Simulated");
              });
      try {
          handler.handleRequest(session, request, response, res);
      } catch (RuntimeException exception) {
          Assert.assertEquals("Simulated", exception.getMessage());
      }
      Mockito.verify(response)
            .setStatus(HttpStatusCode.INTERNAL_SERVER_ERROR.getCode());
    }

  ```

mcollovati (dismissed):

mcollovati (approved):

mcollovati:
@archiecobbs thank you for the contribution!

mcollovati:
I'll take care of the manual cherry-pick

PR #16782, opened 2023-05-12
feat: Store and use package-lock for bundle

### LINKED ISSUE #16041: Production build is not reproducible with the new dev bundle mode
### Description of the bug

When you use frontend hotdeploy mode, you have a `package.json` and `package-lock.json` in the project and they are always up to date. Assuming you commit `package-lock.json` to the repository your builds will be reproducible, especially when using `npm ci` instead of `npm install`(#15986). Otherwise, you might get newer frontend dependencies than originally used and they might be incompatible or have new bugs.

When you are using dev bundle mode, you have no package.json and the package-lock.json used to create the bundle (both for the default and an app specific) is not available for the production build. It might always pick up newer versions which can break the build.

### Expected behavior

When using dev bundle mode, the production build should use the same package-lock.json as used when building the dev bundle. This will ensure reproducible builds and that the production build works like development mode.

For this to work, `package-lock.json` (and maybe `package.json`) should be added to the bundle folder (default and app bundle) after the build is complete. A production build should use this `package-lock.json` when performing ´npm install` or `npm ci`.

At the same time solving this would solve the problem that `npm install` takes a long time for production builds as `package-lock.json` typically is not in the repository.

```
npx @vaadin/cli init --next hello
cd hello
mvn clean vaadin:clean-frontend install -Pproduction

# Store the bundle
cp -R target/classes/META-INF/VAADIN/webapp/ webapp-1

# Remove all created files
mvn clean vaadin:clean-frontend

# Wait for a new dependency to be released

mvn clean vaadin:clean-frontend install -Pproduction
# Store the bundle
cp -R target/classes/META-INF/VAADIN/webapp/ webapp-2

# Compare
diff -r webapp-1 webapp-2
```

### Versions

- Vaadin version: 24 beta 5

caalador:
When making a bundle build if
no package-lock file exists in
project, use the dev bundle package-lock
that is stored when a dev bundle is built.

Acceptance order of lock files is:
project lock, src/dev-bundle, jar bundle.

Fixes #16041


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -204,6 +204,24 @@ private void runFrontendBuildTool(String toolName, String executable,
         }
     }
 
+    private void copyPackageLockJson() {
+        File devBundleFolder = new File(options.getNpmFolder(),
+                Constants.DEV_BUNDLE_LOCATION);
+        assert devBundleFolder.exists() : "No dev-bundle folder created";
+
+        File packageLockJson = new File(options.getNpmFolder(),
```
Artur-, line 212: What about pnpm?

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleUtils.java
```diff
@@ -128,4 +132,49 @@ public static boolean isPreCompiledProductionBundle() {
...
     }
+
+    /**
+     * Copy package-lock.json file from existing dev-bundle for building new
+     * bundle.
+     *
+     * @param options
+     *            task options
+     */
+    public static void copyPackageLock(Options options) {
```
Artur-, line 143:
```suggestion
    public static void copyPackageLockFromBundle(Options options) {
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -204,6 +204,24 @@ private void runFrontendBuildTool(String toolName, String executable,
         }
     }
 
+    private void copyPackageLockJson() {
```
Artur-, line 207:
```suggestion
    private void copyPackageLockJsonToBundleFolder() {
```

caalador: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleUtils.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
caalador: As we only package using npm the default won't have .yaml file available, but pnpm would have `pnpm import` that should be able to import a `package-lock.json` file. For the custom bundle we could store the pnpm lock file.

Artur-: For a project that uses pnpm we should indeed store the pnpm-lock so that a production build is reproducible

caalador: added

mshabarov (dismissed):

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleUtils.java
```diff
@@ -128,4 +132,57 @@ public static boolean isPreCompiledProductionBundle() {
...
+        if (packageLock.exists()) {
+            // NO-OP due to existing package-lock
+            return;
+        }
+
+        try {
+            copyAppropriatePackageLock(options, packageLock);
+        } catch (IOException ioe) {
+            getLogger().error(
+                    "Failed to copy existing `package-lock.json` to use", ioe);
```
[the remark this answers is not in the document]
caalador: Changed to log correct lock file.

mshabarov (approved):

PR #16784, opened 2023-05-12
feat: Add usage statistics for pre-compiled production bundle

mshabarov:


flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
```diff
@@ -147,6 +147,8 @@ private static boolean needsBuildProdBundle(Options options,
                 .isEmpty()) {
             getLogger()
                     .info("Custom eager routes defined. Require bundle build.");
+            UsageStatistics.markAsUsed(
```
Artur-, line 150: Not sure what this means. I would have thought "optimization" is the reason when you force a production build

mshabarov: So it really is: if Flow spots `LoadDependenciesOnStartup`, it triggers a new bundle to apply custom loading rules. That's why this is "bundle-optimisation". I'm fine with any better name :)

mshabarov: Or, right, bundle optimisation is when you force the production build and have routes loading on demand. If you use annotation, this would be better named as "flow/rebundle-reason-custom-bundle-loading" or similar.

Artur- (dismissed):

Artur- (approved):

PR #16794, opened 2023-05-15
fix: Store and reuse index.html modifications for dev bundles

### LINKED ISSUE #16792: Tags added to the index page by Vite plugins are ignored when running in dev bundle mode
### Description of the bug

The Vite Cesium plugin used in #16208 adds a `<link>` tag to index.html using the `transformIndexHtml` hook in Vite: 
```
    transformIndexHtml() {
      const tags = [
        {
          tag: "link",
          attrs: {
            rel: "stylesheet",
            href: (0, import_vite.normalizePath)(import_path.default.join(CESIUM_BASE_URL, "Widgets/widgets.css"))
          }
        }
      ];
//...
      return tags
    }
```

This ends up in `src/main/dev-bundle/webapp/index.html` as 
```
<head>
  <link rel="stylesheet" href="/cesium/Widgets/widgets.css">
```

However, when running in dev bundle mode, this file is never used: https://github.com/vaadin/flow/blob/a23d76d6d3eca93d420a362f7591498e3d6b39ee/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java#L417-L422

### Expected behavior

You can add tags to index.html using a Vite plugin and the result is included in the final index.html page

See #16208 

### Versions

- Vaadin / Flow version:24.1.0.alpha5

Artur-:
Fixes #16792

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
@@ -25,7 +25,6 @@ import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
-import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -49,7 +48,6 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.apache.commons.io.IOUtils;
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.DataNode;
 import org.jsoup.nodes.Document;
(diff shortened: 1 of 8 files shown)


flow-server/src/main/java/com/vaadin/flow/server/BootstrapHandler.java
```diff
@@ -1676,31 +1674,30 @@ private static Element getStyleTag(String themeName, String fileName,
...
+        Optional<String> webComponentScript = JsonUtils
+                .stream(statsJson.getArray("entryScripts"))
+                .map(value -> value.asString())
+                .filter(script -> script.contains("webcomponenthtml"))
+                .findFirst();
 
-            if (entryScript.contains("webcomponenthtml")) {
-                addIndexHtml = false;
-            }
+        if (webComponentScript.isPresent()) {
```
caalador, line 1689: Doesn't this now differ in that we with this change only add the first webcomponenthtml entry script where as before we added all except indexhtml?

Artur-: Are there many of them?

Artur-: For instance in `flow-tests/test-express-build/test-embedding-express-build` there are two exported wcs but `stats.json` contains 
```
 "entryScripts": [
  "VAADIN/build/webcomponenthtml-195b78ff.js",
  "VAADIN/build/indexhtml-f9743387.js"
 ],
```

I thought that webcomponenthtml here refers to `/frontend/web-component.html` which there is always only one of

caalador (approved):

PR #16795, opened 2023-05-16
chore: Tweak the feature flag message to give more info

Artur-:


czp13 (approved):
It is good from the flow side, but the components team can check as well 👍

mshabarov (approved):

Issue #15825, opened 2023-02-04
Order of the loaded @JsModule is not as the documentation suggests


johanneshayry:
### Description of the bug

@JsModule advertises that `The files loaded will be in the same order as the annotations were on the class.` I understand this as the order written on the annotated class, not just `some` predictable repeated order. However, the order is not the order written.

### Expected behavior

Expecting this code to execute the side effects of script-2 first and then script-1. However, the order is the opposite.
```java
@JsModule("./script-2.js")
@JsModule("./script-1.js")
public class EmptyView extends VerticalLayout {

    public EmptyView() {
    }
}
```

Now `target/frontend/generated-flow-imports.js` file has the imports as:
```javascript
import 'Frontend/generated/jar-resources/script-1.js';
import 'Frontend/generated/jar-resources/script-2.js';
```

### Minimal reproducible example

https://github.com/johanneshayry/jsmodule-order-bug

### Versions

- Vaadin / Flow version: 23.3.5
- Java version: Amazon.com Inc. 17.0.2
- OS version: amd64 Windows 10 10.0

Artur-:
If you have control over script-2 it might be better to import script-1 into that. Then you will avoid the potential situation of another class having the same annotations but in the oppposite order. I wonder which order ”the same order as the annotations were on the class” is then…

Artur-:
The sorting looks like an attempt to make behavior deterministic but it certainly feels like it is in the wrong place.

Artur-:
To fix this, I would go ahead and remove the sorting and rely on what the scanner produces. Then you would of course need a couple of tests to validate that the class order promise is actually kept.

I don’t think much will break if the order is currently alphabetical

mshabarov:
@johanneshayry this has been fixed in Vaadin 24.1 pre-release. Picking it to V23 and V24.0 need more efforts. Do you still need this for V23.3 ?

PR #16820, opened 2023-05-17
fix: remove natural order sorting of jsmodules and javascript imports

tltv:
Removes sorting of jsmodule and javascript imports in natural order in generated-flow-imports.js. Sorting is removed only for imports added via JsModule and Javascript annotation. Natural order was not in line with the promised order described in these annotations, which is to keep the same order as the annotations were on the class. However, order is only guaranteed on a class level.

Fixes: #15825


czp13 (approved):
Nice one, checked it, no unused imports, covered with tests, minor changes could happen like:
-         `expectedImports.addAll(expectedJsModuleImports);` ->   `List<String> expectedImports = new ArrayList<>(expectedJsModuleImports);`

But super minor, it is good to merge I think.

PR #16834, opened 2023-05-19
chore: add owner to JavaScriptInvocation class.

czp13:
## Description

Adding owner to JavaScriptInvocation class.

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -125,9 +125,9 @@ public class UIInternals implements Serializable {
         }
 
         /**
-         * Gets the state node that this invocation belongs to.
+         * Gets the state node that this invocation belongs to, if any.
          *
-         * @return the owner state node
+         * @return the optional owner state node, or an empty optional if no state node has been provided for this invocation.
          */
         public Optional<StateNode> getOwner() {
             return Optional.ofNullable(owner);
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -46,6 +46,7 @@ import com.vaadin.flow.internal.AnnotationReader;
 import com.vaadin.flow.internal.ConstantPool;
 import com.vaadin.flow.internal.ReflectTools;
 import com.vaadin.flow.internal.StateTree;
+import com.vaadin.flow.internal.StateNode;
 import com.vaadin.flow.internal.nodefeature.LoadingIndicatorConfigurationMap;
 import com.vaadin.flow.internal.nodefeature.NodeFeature;
 import com.vaadin.flow.internal.nodefeature.PollConfigurationMap;
@@ -88,6 +89,7 @@ public class UIInternals implements Serializable {
     public static class JavaScriptInvocation implements Serializable {
         private final String expression;
         private final List<Serializable> parameters = new ArrayList<>();
+        private final StateNode owner;
 
         /**
          * Creates a new invocation.
@@ -98,10 +100,22 @@ public class UIInternals implements Serializable {
          *            a list of parameters to use when invoking the script
          */
         public JavaScriptInvocation(String expression,
+                                    StateNode owner,
                 Serializable... parameters) {
             this.expression = expression;
+            this.owner = owner;
             Collections.addAll(this.parameters, parameters);
         }
+
+        /**
+         * Gets the state node that this invocation belongs to.
+         *
+         * @return the owner state node
+         */
+        public StateNode getOwner() {
+            return owner;
+        }
+
 
         /**
          * Gets the JavaScript expression to invoke.
@@ -557,7 +571,7 @@ public class UIInternals implements Serializable {
     public void setTitle(String title) {
         assert title != null;
         JavaScriptInvocation invocation = new JavaScriptInvocation(
-                "document.title = $0", title);
+                "document.title = $0", getStateTree().getRootNode(), title);
 
         pendingTitleUpdateCanceler = addJavaScriptInvocation(invocation);
 
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -100,8 +100,25 @@ public class UIInternals implements Serializable {
          *            a list of parameters to use when invoking the script
          */
         public JavaScriptInvocation(String expression,
+                                    Serializable... parameters) {
+            this.expression = expression;
+            this.owner = new StateNode();
+            Collections.addAll(this.parameters, parameters);
+        }
+
+        /**
+         * Creates a new invocation.
+         *
+         * @param expression
+         *            the expression to invoke
+         * @param owner
+         *            the state node that this invocation belongs to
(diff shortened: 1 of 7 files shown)


mcollovati (commented):
The change looks good to me.
I just wonder if we should preserve and deprecate the previous `JavaScriptInvocation` constructor, to avoid breaking changes, since this is a public API.

Waiting @mshabarov considerations

mshabarov (changes requested):
I agree that @mcollovati said: let's preserve the existing API as is, but deprecate it and add a new constructor. Also, since it is possible to construct the invocation without the owner, let's highlight it in the getter's javadoc, that owner can be `null`.

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -1463,6 +1464,10 @@ public void callFunction(String functionName, Serializable... arguments) {
 
     private void doCallFunction(UI ui, String functionName,
             Serializable... arguments) {
+        UIInternals.JavaScriptInvocation invocation = new UIInternals.JavaScriptInvocation(
+                functionName, getNode(), arguments);
+        ui.getInternals().addJavaScriptInvocation(invocation);
```
mcollovati, line 1469: Isn't this causing the execution to be sent twice to the client?

flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
```diff
@@ -262,6 +263,26 @@ public ExecutionCanceler executeJavaScript(String expression,
         return ui.getInternals().addJavaScriptInvocation(invocation);
```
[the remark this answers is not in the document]
czp13: Thanks for the further review, it makes perfect sense. I also was thinking about that rapidly yesterday to avoid code duplication, but at that point I though it is more complicated or produce not nice code, but after did it, it seems way more clean.

Done 👍

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -1475,7 +1476,7 @@ private void doCallFunction(UI ui, String functionName,
                             .toArray(Serializable[]::new);
 
                     ui.getPage().executeJavaScript("$0." + functionName + "("
-                            + paramPlaceholderString + ")", jsParameters);
+                            + paramPlaceholderString + ")", getNode(),  jsParameters);
```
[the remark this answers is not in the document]
czp13: I did not catch this one but now fixed this as well. Thanks for noticing it!

flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
```diff
@@ -245,6 +246,42 @@ private void addDependency(Dependency dependency) {
...
+     * @param expression
+     *           the JavaScript expression to invoke
+     * @param owner
+     *          the owner StateNode of the JavaScript expression
+     * @param parameters
+     *      parameters to pass to the expression
+     * @return  a callback for canceling the execution if not yet sent to browser
+     */
+
+    public ExecutionCanceler executeJavaScript(String expression, StateNode owner,
```
mshabarov, line 283: I don't understand why one needs an owner node to run the JS on the Page? The owner here is always a root node, or I missed something?

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -1463,19 +1464,20 @@ public void callFunction(String functionName, Serializable... arguments) {
...
+        ui.getInternals().getStateTree().beforeClientResponse(node,
                 context -> {
                     // $0.method($1,$2,$3)
                     String paramPlaceholderString = IntStream
                             .range(1, arguments.length + 1)
                             .mapToObj(i -> "$" + i)
                             .collect(Collectors.joining(","));
                     Serializable[] jsParameters = Stream
-                            .concat(Stream.of(this), Stream.of(arguments))
+                            .concat(Stream.of(node), Stream.of(arguments))
```
mcollovati, line 1476: this seems to be wrong. `this` is not the `Element` instance, not the `StateNode`.
My bad, I didn't notice that capture of `this`, so using `getNode()` is fine.
I apologize for the mistake

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -1463,19 +1464,20 @@ public void callFunction(String functionName, Serializable... arguments) {
...
                             .toArray(Serializable[]::new);
 
                     ui.getPage().executeJavaScript("$0." + functionName + "("
-                            + paramPlaceholderString + ")", jsParameters);
+                            + paramPlaceholderString + ")", node,  jsParameters);
```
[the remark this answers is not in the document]
mcollovati: And if we want to preserve the parameters type check, we can probably just add a `.peek(JsonCodec::encodeWithTypeInfo)` to the `jsParameters` stream flow.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -98,11 +100,40 @@ public static class JavaScriptInvocation implements Serializable {
          *            a list of parameters to use when invoking the script
          */
         public JavaScriptInvocation(String expression,
-                Serializable... parameters) {
+                                    Serializable... parameters) {
             this.expression = expression;
+            this.owner = new StateNode();
```
mshabarov, line 105: I propose to assign a root node to the owner here.

flow-server/src/main/java/com/vaadin/flow/component/page/Page.java
czp13: This will be deleted as discussed. Thanks!

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
czp13: Okey, yap, indeed, fixing it! Thanks!

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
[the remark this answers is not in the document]
czp13: Refactored, changed. Thanks for review ✅

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
czp13: As discussed internally, it will be null and Optional used for the getter return value!

mshabarov (dismissed):

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -98,11 +100,40 @@ public static class JavaScriptInvocation implements Serializable {
...
+                                    Serializable... parameters) {
+            this.expression = expression;
+            this.owner = owner;
+            Collections.addAll(this.parameters, parameters);
+        }
+
+        /**
+         * Gets the state node that this invocation belongs to.
+         *
+         * @return the owner state node
```
mcollovati, line 130: nit:
```suggestion
         * Gets the state node that this invocation belongs to, if any.
         *
         * @return the optional owner state node, or an empty optional if no state node has been provided for this invocation.
```

mshabarov (approved):

mcollovati (approved):

Issue #16780, opened 2023-05-12
Static resource cannot be found if it referenced from CSS with relative path


mshabarov:
### Description of the bug

If I place my static resource, say, image `vaadin-logo.png` into `src/main/webapp/images/vaadin-logo.png` and reference it from the `styles.css` of my custom theme, like shown below:

```css
.vaadin-logo {
    background-image: url("./images/vaadin-logo.png");
}
```
it gives 404 NOT FOUND response from server, so Flow doesn't resolve the URL properly.
However, if I put this image file into `${project.root}/frontend/themes/my-theme/images/`, it would work as expected.
Also, if I remove the `.` and set the URL as an absolute path, it also works: url("/images/vaadin-logo.png").

I expect that the relative URL should properly be resolved by Flow and properly return the resource.

This issue occurs for both pre-compiled and custom production bundle.

### Expected behavior

Static resource should be served properly with no matter where the resource is placed.

### Minimal reproducible example

See the test module in Flow `flow-tests/test-express-build/test-parent-theme-in-frontend-prod` and CSS file
https://github.com/vaadin/flow/blob/958bd32a92834c5e6bf752c8182a38421a8fc861/flow-tests/test-express-build/test-parent-theme-in-frontend-prod/frontend/themes/specific-theme/styles.css#L11

### Versions

- Vaadin / Flow version: Flow 24.1.0.alpha7

mshabarov:
Vaadin 24.1 pre uses different approaches for making and adding style tags into `head` on the page: with pre-compiled dev bundle resolves url in CSS and inlines the resulting CSS into `style` tag, whereas with pre-compiled prod bundle Flow adds a `link` tag with the reference `VAADIN/themes/my-theme/styles.css`. In latter case the image url is resolved into ``VAADIN/themes/my-theme/images/vaadin-logo.png`, which point to nothing.

For pre-compiled prod bundle we need to inline CSS similarly or rewrite the urls.
This workaround below leads to proper resource handling:
```java
@Component
public class ApplicationServiceInitListener
        implements VaadinServiceInitListener {
    @Override
    public void serviceInit(ServiceInitEvent event) {
        event.addIndexHtmlRequestListener(response -> {
            Document document = response.getDocument();
            Element head = document.head();
            Element styleTag = new Element("style");
            styleTag.attr("workaround", "workaround");
           // inline CSS
            styleTag.appendText("""
                    .vaadin-logo {
                      background-image: url("./images/vaadin-logo.png");
[...]

PR #16869, opened 2023-05-23
fix: Look up for theme resource in static resource folder

mshabarov:
Looks up for the static resource in the app's class-path needed for custom theme ~and inlines the content of it into style tag, like it's done for development mode~. This resolves `url(./images/foo.png)` inside CSS properly if a resource is located in the static resource directory, not in the theme folder.

Fixes https://github.com/vaadin/flow/issues/16780

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -1,7 +1,7 @@
 name: Flow Validation
 on:
   push:
-    branches: [main, '24.0', '23.3', '23.2',  '23.1', '9.0']
+    branches: [main, '24.1', '24.0', '23.3', '9.1']
   workflow_dispatch:
   pull_request_target:
     types: [opened, synchronize, reopened, edited]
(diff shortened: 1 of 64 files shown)


Artur-:
This would make it impossible for the browser to cache the css

mshabarov:
> This would make it impossible for the browser to cache the css

We have a lots of other styles generated by Flow and inlined on the page. With inline style you don't need an extra request to the server, so in some cases it's beneficial. However, suddenly I start to think about looking up for the resource in the static resources folder in 
```java
    public static URL getThemeResourceFromPrecompiledProductionBundle(
            String themeAssetPath) {
        // lookup in the prod bundle, where themes are copied from project's
        URL resourceUrl = ThemeUtils.class.getClassLoader()
                .getResource(Paths.get(VAADIN_WEBAPP_RESOURCES,
                        VAADIN_STATIC_FILES_PATH, themeAssetPath).toString());
        if (resourceUrl == null) {
            // lookup in the JARs for packaged themes
            resourceUrl = ThemeUtils.class.getClassLoader().getResource(
                    Constants.RESOURCES_JAR_DEFAULT + themeAssetPath);
        }

        // Add this:
        if (resourceUrl == null) {
            // trim 'themes/my-theme/' and look up in the static resources:
            resourceUrl = getStaticResource(trimmedFilenameWithPath);
        }
        return resourceUrl;
    }
```

Artur-:
> We have a lots of other styles generated by Flow and inlined on the page
Should they be done that way?

mshabarov:
Not an option, as it might expose static resources wrapped into theme request.

PR #16836, opened 2023-05-19
fix: remove detach listener when javascript execution completes

mcollovati:
## Description

A detach listener is added for pending javascript invocation owner nodes to clean up the queue, if the owner gets detached. This change also removes the detach listener when the javascript execution completes.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #16875, opened 2023-05-24
chore: disable sending updates to client for effectively non-visible nodes, and some additional cleanup, changes to apply CP-s (#15885)  (#16855) and (#16836) (CP: 1.0)

czp13:
## Description
Manually cherry-picked #15885 PR for 1.0:
- Disable sending updates to client for effectively non-visible nodes
- Fix test: Always set tested component visible first; only assume node has changed if testvalue differs from original value
- Cleanup, and refactoring, in Element, StateNode, UIInternals classes + mvn formatter.

Plus manually cherry-picked #16836 and #16855 PR-s commits, and fine-tuned for this codebase/branch.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -22,11 +22,13 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import com.vaadin.flow.server.Command;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -174,6 +176,8 @@ public class UIInternals implements Serializable {
     private long lastHeartbeatTimestamp = System.currentTimeMillis();
 
     private List<JavaScriptInvocation> pendingJsInvocations = new ArrayList<>();
+
+    private final HashMap<StateNode, JavaScriptInvocationDetachListener> pendingJsInvocationDetachListeners = new HashMap<>();
 
     /**
      * The related UI.
@@ -543,9 +547,7 @@ public class UIInternals implements Serializable {
             JavaScriptInvocation invocation) {
         session.checkHasLock();
         pendingJsInvocations.add(invocation);
-      invocation.getOwner().ifPresent(node -> node.addDetachListener(() -> pendingJsInvocations
-                     .removeIf(pendingInvocation -> pendingInvocation
-                             .equals(invocation))));
+
         return () -> pendingJsInvocations.remove(invocation);
     }
 
@@ -566,12 +568,78 @@ public class UIInternals implements Serializable {
                 .filter(invocation -> invocation.getOwner().map(StateNode::isVisible).orElse(true))
                 .collect(Collectors.toList());
 
+        StateNode rootNode = ui.getInternals().getStateTree().getRootNode();
+        readyToSend.forEach(i -> {
+            JavaScriptInvocationDetachListener listener = pendingJsInvocationDetachListeners.get(i.getOwner().orElse(rootNode));
+            if (listener != null) {
+                listener.removePendingStatusAndRegistrationAndDetachListener(i);
+            }
+        });
+
         pendingJsInvocations = getPendingJavaScriptInvocations()
                 .stream()
                 .filter(invocation -> invocation.getOwner().map(node -> !node.isVisible()).orElse(false))
                 .collect(Collectors.toCollection(ArrayList::new));
 
+        pendingJsInvocations
+                .forEach(this::registerDetachListenerForPendingInvocation);
+
         return readyToSend;
+    }
+
+    /**
+     *  Registers a detach listener for the given invocation.
+     *
+     * @param invocation - invocation should always be having an existing owner if we call with this.
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private void registerDetachListenerForPendingInvocation(
+            JavaScriptInvocation invocation) {
+
+        StateNode ownerNode = invocation.getOwner().orElse(ui.getInternals().getStateTree().getRootNode());
+
+        JavaScriptInvocationDetachListener listener = pendingJsInvocationDetachListeners
+                .computeIfAbsent(ownerNode, node -> {
+                    JavaScriptInvocationDetachListener detachListener = new JavaScriptInvocationDetachListener();
+                    Registration registration = node.addDetachListener(detachListener);
+                    detachListener.registration = () -> {
+                        pendingJsInvocationDetachListeners.remove(node);
+                        registration.remove();
+                    };
+                    return detachListener;
+                });
+        listener.invocationList.add(invocation);
+    }
+
+    private class JavaScriptInvocationDetachListener implements Command {
+        private final Set<JavaScriptInvocation> invocationList = new HashSet<>();
+
+        private Registration registration;
+
+        @Override
+        public void execute() {
+            if (!invocationList.isEmpty()) {
+                List<JavaScriptInvocation> copy = new ArrayList<>(
+                        invocationList);
+                invocationList.clear();
+                copy.forEach(this::removePendingInvocation);
+            }
+        }
+
+        private void removePendingInvocation(
+                JavaScriptInvocation invocation) {
+            UIInternals.this.pendingJsInvocations.removeIf(
+                    pendingInvocation -> pendingInvocation.equals(invocation));
+            if (invocationList.isEmpty() && registration != null) {
+                registration.remove();
+                registration = null;
+            }
+        }
+
+        void removePendingStatusAndRegistrationAndDetachListener(JavaScriptInvocation invocation) {
+            invocationList.remove(invocation);
+            removePendingInvocation(invocation);
+        }
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -542,6 +543,9 @@ public class UIInternals implements Serializable {
             JavaScriptInvocation invocation) {
         session.checkHasLock();
         pendingJsInvocations.add(invocation);
+      invocation.getOwner().ifPresent(node -> node.addDetachListener(() -> pendingJsInvocations
+                     .removeIf(pendingInvocation -> pendingInvocation
+                             .equals(invocation))));
         return () -> pendingJsInvocations.remove(invocation);
     }
 
@@ -557,11 +561,17 @@ public class UIInternals implements Serializable {
             return Collections.emptyList();
         }
 
-        List<JavaScriptInvocation> currentList = pendingJsInvocations;
+        List<JavaScriptInvocation> readyToSend = getPendingJavaScriptInvocations()
+                .stream()
+                .filter(invocation -> invocation.getOwner().map(StateNode::isVisible).orElse(true))
+                .collect(Collectors.toList());
 
-        pendingJsInvocations = new ArrayList<>();
+        pendingJsInvocations = getPendingJavaScriptInvocations()
+                .stream()
+                .filter(invocation -> invocation.getOwner().map(node -> !node.isVisible()).orElse(false))
+                .collect(Collectors.toCollection(ArrayList::new));
 
-        return currentList;
+        return readyToSend;
     }
 
     /**
@@ -598,7 +608,7 @@ public class UIInternals implements Serializable {
     /**
      * Gets the page title recorded with {@link Page#setTitle(String)}.
      * <p>
-     * <b>NOTE</b> this might not be up to date with the actual title set since
+     * <b>NOTE</b> this might not be up-to-date with the actual title set since
      * it is not updated from the browser and the update might have been
      * canceled before it has been sent to the browser with
      * {@link #cancelPendingTitleUpdate()}.
--- a/flow-server/src/test/java/com/vaadin/flow/component/internal/UIInternalsTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/internal/UIInternalsTest.java
@@ -1,0 +1,196 @@
+package com.vaadin.flow.component.internal;
+
+import java.util.Optional;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.Tag;
+import com.vaadin.flow.component.UI;
+import com.vaadin.flow.component.page.Push;
+import com.vaadin.flow.di.DefaultInstantiator;
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.internal.StateNode;
+import com.vaadin.flow.internal.nodefeature.ElementChildrenList;
+import com.vaadin.flow.internal.nodefeature.ElementData;
+import com.vaadin.flow.router.ParentLayout;
+import com.vaadin.flow.router.Route;
+import com.vaadin.flow.router.RouterLayout;
+import com.vaadin.flow.server.VaadinService;
+import com.vaadin.tests.util.AlwaysLockedVaadinSession;
+
+public class UIInternalsTest {
+
+    @Mock
+    UI ui;
+    @Mock
+    VaadinService vaadinService;
+
+    UIInternals internals;
+
+    @Route
+    @Push
+    @Tag(Tag.DIV)
+    public static class RouteTarget extends Component implements RouterLayout {
+
+    }
+
+    @Route(value = "foo", layout = RouteTarget.class)
+    @Tag(Tag.DIV)
+    public static class RouteTarget1 extends Component {
+
+    }
+
+    @Tag(Tag.DIV)
+    static class MainLayout extends Component implements RouterLayout {
+        static String ID = "main-layout-id";
+
+        public MainLayout() {
+            setId(ID);
+        }
+    }
+
+    @Tag(Tag.DIV)
+    @ParentLayout(MainLayout.class)
+    static class SubLayout extends Component implements RouterLayout {
+        static String ID = "sub-layout-id";
+
+        public SubLayout() {
+            setId(ID);
+        }
+    }
+
+    @Tag(Tag.DIV)
+    @Route(value = "child", layout = SubLayout.class)
+    static class FirstView extends Component {
+        static String ID = "child-view-id";
+
+        public FirstView() {
+            setId(ID);
+        }
+    }
+
+    @Tag(Tag.DIV)
+    static class AnotherLayout extends Component implements RouterLayout {
+        static String ID = "another-layout-id";
+
+        public AnotherLayout() {
+            setId(ID);
+        }
+    }
+
+    @Tag(Tag.DIV)
+    @Route(value = "another", layout = MainLayout.class)
+    static class AnotherView extends Component {
+        static String ID = "another-view-id";
+
+        public AnotherView() {
+            setId(ID);
+        }
+    }
+
+    @Before
+    public void init() {
+        MockitoAnnotations.initMocks(this);
+
+        Mockito.when(ui.getUI()).thenReturn(Optional.of(ui));
+        Element body = new Element("body");
+        Mockito.when(ui.getElement()).thenReturn(body);
+
+        internals = new UIInternals(ui);
+        AlwaysLockedVaadinSession session = new AlwaysLockedVaadinSession(
+                vaadinService);
+        Mockito.when(vaadinService.getInstantiator())
+                .thenReturn(new DefaultInstantiator(vaadinService));
+        internals.setSession(session);
+        Mockito.when(ui.getSession()).thenReturn(session);
+
+        Mockito.when(ui.getInternals()).thenReturn(internals);
+    }
+
+    @Test
+    public void dumpPendingJavaScriptInvocations_detachListenerRegisteredOnce() {
+        StateNode node = Mockito.spy(new StateNode(ElementData.class));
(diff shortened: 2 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +564,18 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
             return Collections.emptyList();
         }
 
-        List<JavaScriptInvocation> currentList = pendingJsInvocations;
+        List<JavaScriptInvocation> readyToSend = getPendingJavaScriptInvocations()
+                .stream()
+                .filter(invocation -> invocation.getOwner().isPresent() && invocation.getOwner().get().isVisible())
+//                .peek(JavaScriptInvocation::setSentToBrowser)
```
czp13, line 570: The only difference between the original and cherry-picked commit, this part can be deleted as this concept does not exist in `1.0` and is not used anywhere else in `1.0`. 

TLDR: So it is safe to delete the setting part (as getting, or using is not existing)

mcollovati:
I think this manual cherry pick should also consider changes in #16836 and #16855

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +564,17 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
             return Collections.emptyList();
         }
 
-        List<JavaScriptInvocation> currentList = pendingJsInvocations;
+        List<JavaScriptInvocation> readyToSend = getPendingJavaScriptInvocations()
+                .stream()
+                .filter(invocation -> invocation.getOwner().isPresent() && invocation.getOwner().get().isVisible())
```
mshabarov, line 569: Invocations that have no owners are ready to be send as well - we assume they are for root node, i.e. always visible.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +564,17 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
...
+                .collect(Collectors.toList());
 
-        pendingJsInvocations = new ArrayList<>();
+        pendingJsInvocations = getPendingJavaScriptInvocations()
+                .stream()
+                .filter(invocation -> !(invocation.getOwner().isPresent() && invocation.getOwner().get().isVisible()))
```
mshabarov, line 574: Same here: this filter condition needs change, because JS invocations for root node would be pending forever.

mshabarov (changes requested):

mshabarov:
> @mcollovati: I was thinking about merging this one soon.
> 
> And then merge the mentioned two commits together as the third (#16855) fix is a small one, but the second fix/CP (#16836) is quite big and building on top of this. It could be nice for historical reasons, reverting, git tree history to unite the PR-s, and commits like this.
> 
> p.s: but if you feel, I can add those ones to here 🙇

You can add the commits to this branch, but then do "Rebase & Merge", not "Squash" and they will be preserved in history.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -542,6 +543,10 @@ public ExecutionCanceler addJavaScriptInvocation(
             JavaScriptInvocation invocation) {
         session.checkHasLock();
         pendingJsInvocations.add(invocation);
+      invocation.getOwner().ifPresent(node -> invocation.getOwner().get()
+             .addDetachListener(() -> pendingJsInvocations
+                     .removeIf(pendingInvocation -> pendingInvocation
+                             .equals(invocation))));
```
mcollovati, line 549: the `node` param in the lambda is already `invocation.getOwner().get()`
```suggestion
      invocation.getOwner().ifPresent(node -> 
              node.addDetachListener(() -> pendingJsInvocations
                     .removeIf(pendingInvocation -> pendingInvocation
                             .equals(invocation))));
```

czp13: Good one, shorter code is usually better. The only reason I do not commit this is I want to have only one commit for this change, then another commit for the other CP-s, to not push like 5-10 commits to the branch. 

So I squash logically related commits together.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
czp13: Thanks, fixed!

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +561,17 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
             return Collections.emptyList();
         }
 
-        List<JavaScriptInvocation> currentList = pendingJsInvocations;
+        List<JavaScriptInvocation> readyToSend = getPendingJavaScriptInvocations()
+                .stream()
+                .filter(invocation -> invocation.getOwner().isEmpty() || invocation.getOwner().get().isVisible())
```
mcollovati, line 566:
```suggestion
                .filter(invocation -> invocation.getOwner().map(StateNode::isVisible).orElse(true))
```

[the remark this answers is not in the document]
czp13: Thanks, changed to negated isPresent() shall be the same.

czp13: This is good :), I've double-checked.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +561,17 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
...
+                .collect(Collectors.toList());
 
-        pendingJsInvocations = new ArrayList<>();
+        pendingJsInvocations = getPendingJavaScriptInvocations()
+                .stream()
+                .filter(invocation -> invocation.getOwner().isPresent() && !invocation.getOwner().get().isVisible())
```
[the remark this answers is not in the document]
czp13: Ok, this seems the same by eye checks. 

Asked my bot friend, who said it is different, then I executed this code for test:

```

import lombok.Data;

import java.util.List;
import java.util.Optional;
import java.util.stream.Collectors;

public class InvocationLogicTest {

    @Data
    static class Owner {
        private boolean isVisible;

        public Owner(boolean isVisible) {
            this.isVisible = isVisible;
        }

        public boolean isVisible() {
            return isVisible;
        }
    }

    @Data
    static class Invocation {
        private Owner owner;

        public Invocation(Owner owner) {
            this.owner = owner;
        }

        public Invocation() {
        }

        public Optional<Owner> getOwner() {
            return owner == null ? Optional.empty() : Optional.of(owner);
        }
    }

        List<Invocation> invocations = List.of(
                new Invocation(new Owner(true)),
                new Invocation(),
                new Invocation(new Owner(false))
        );


        System.out.println("Logic for: invocations.stream().filter(invocation -> invocation.getOwner().isPresent() && !invocation.getOwner().get().isVisible()).collect(Collectors.toList())");
        List<Invocation> filteredList1 = invocations.stream()
                .filter(invocation -> invocation.getOwner().isPresent() && !invocation.getOwner().get().isVisible())
                .collect(Collectors.toList());

        System.out.println(filteredList1.size());  // Output: 1
        System.out.println(filteredList1.get(0) == invocations.get(2));  // Output: true
        System.out.println(filteredList1.get(0));

        System.out.println("Logic for: invocations.stream().filter(invocation -> invocation.getOwner().map(node -> !node.isVisible()).orElse(false)).collect(Collectors.toList())");
        List<Invocation> filteredList2 = invocations.stream()
                .filter(invocation -> invocation.getOwner().map(node -> !node.isVisible()).orElse(false))
                .collect(Collectors.toList());

        System.out.println(filteredList2.size());  // Output: 2
        System.out.println(filteredList2.get(0) == invocations.get(2));  // Output: true
        System.out.println(filteredList2.get(0));


    }

}

```

It showed this output:
```
Logic for: invocations.stream().filter(invocation -> invocation.getOwner().isPresent() && !invocation.getOwner().get().isVisible()).collect(Collectors.toList())
1
true
InvocationLogicTest.Invocation(owner=Optional[InvocationLogicTest.Owner(isVisible=false)])
Logic for: invocations.stream().filter(invocation -> invocation.getOwner().map(node -> !node.isVisible()).orElse(false)).collect(Collectors.toList())
1
true
InvocationLogicTest.Invocation(owner=Optional[InvocationLogicTest.Owner(isVisible=false)])
```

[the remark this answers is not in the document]
czp13: And then my bot friend asked for an excuse:

```
Apologies for the confusion caused by the previous response. Upon reviewing the code snippets, I realize that I made an error in explaining the difference. I apologize for any inconvenience caused.

In the given code snippets:

.filter(invocation -> invocation.getOwner().isPresent() && !invocation.getOwner().get().isVisible())
.filter(invocation -> invocation.getOwner().map(node -> !node.isVisible()).orElse(false))
Both snippets will produce the same output and behave identically in terms of filtering the invocations.

To clarify, in the second snippet, when invocation.getOwner() returns an empty optional (Optional.empty()), the map function will not be executed, and the orElse(false) part will immediately provide the default value of false. This means that invocations without an owner will be filtered out in both cases.

I apologize for the incorrect explanation in the previous response. Thank you for pointing it out, and I appreciate your understanding.
```

You can never believe anyone, even if it is a bot. Maybe too much effort but was interesting, I will push your suggestion because it was overtested as well.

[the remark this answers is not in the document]
czp13: Oh, yeah, and first it said it is different the 3 cases (he accidentally shown 4 cases, because duplicated one).

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -16,16 +16,14 @@
...
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
czp13, line 19: I will fix this, accidentally united.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +558,93 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
...
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private void registerDetachListenerForPendingInvocation(
+            JavaScriptInvocation invocation) {
+
+        StateNode ownerNode;
+        if (!invocation.getOwner().isPresent()) {
+//            return;
+            // or throw error?
+//            throw new IllegalStateException("Owner is not present for invocation: " + invocation);
+            ownerNode = ui.getInternals().getStateTree().getRootNode();
```
czp13, line 591: Thinking it through:
1. if `invocation.getOwner()` is somehow is null should be equal with `getInternals().getStateTree().getRootNode()` (as we discussed before)
2. also the `registerDetachListenerForPendingInvocation` method should be called with the filled-out owner (and not visible rootnode) based on the 568. line logic

So I can remove those comments.

flow-server/src/test/java/com/vaadin/flow/component/internal/UIInternalsTest.java
```diff
@@ -0,0 +1,259 @@
...
+//
+//        UIInternals.JavaScriptInvocation invocation =
+//                new UIInternals.JavaScriptInvocation("", node);
+//        internals.addJavaScriptInvocation(invocation);
+//        internals.dumpPendingJavaScriptInvocations();
+//
+//        Mockito.verify(node, Mockito.times(1))
+//                .addDetachListener(() -> {});
+//
+//        invocation.cancelExecution();
```
czp13, line 233: no `cancelExecution`, nor `completeExceptionally`, `complete` methods are implemented for JavaScriptInvocation so I think this test can be deleted. (manipulating and setting up successHandler and errorHandler, which is also missing from this codebase)

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +564,80 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
...
+                            node.addDetachListener(detachListener));
+                    return detachListener;
+                });
+        listener.invocationList.add(invocation);
+
+        SerializableConsumer callback = unused -> listener
+                .onInvocationCompleted(invocation);
+        // not then part for JavaScript invocation, no successHandler or errorHandler:
+        // so we can remove this one too, I believe:
+        // invocation.then(callback, callback);
```
czp13, line 609: `invocation.then` is the last method that is missing, but also related to `successHandler()` and `errorHandler()`, did not go that deep to see how much effort would be to add these to the code base in a separate PR.

flow-server/src/main/java/com/vaadin/flow/shared/Registration.java
```diff
@@ -29,4 +31,20 @@ public interface Registration extends Serializable {
...
+     * @return a registration that removes all provided registrations not
+     *         <code>null</code>
+     * @since
+     */
+    static Registration combine(Registration... registrations) {
+        Objects.requireNonNull(registrations);
+        Arrays.asList(registrations).forEach(Objects::requireNonNull);
+
+        return () -> Arrays.asList(registrations).forEach(Registration::remove);
+    }
```
mcollovati, line 49: Probably it does not make sense to add this new API, that will be used only in one place.
It should be easy to apply this logic in the caller.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
mcollovati: Instead of adding the missing APIs, I wonder if we can simply call `onInvocationCompleted` in `dumpPendingJavaScriptInvocations`  for every invocation that will be sent to the browser (`readyToSend` list).
In this case, I would rename `onInvocationCompleted` to something that better explains that we are removing the detach listener because the invocation will be sent to the client and executed.
@mshabarov what do you think?

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/shared/Registration.java
czp13: Thanks Marco! 
My idea was that:
- this is a nice addition/improvement to the codebase, so it was a low hanging fruit, 
- and maybe also next time we will need it, we can find it here, but if it is a private method or part of a method then we will duplicate the code.

If you still feel I can refactor this part and do as you wish! 🙇

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
czp13: Thanks! Interesting idea,  I will wait for @mshabarov's thoughts on this, then I can check and do the necessary changes. 🙇

flow-server/src/main/java/com/vaadin/flow/shared/Registration.java
mcollovati: My opinion is that we should not be ad new features to 1.0 if not strictly necessary.
In addition, if you want to peek at this a public API you should also port/implement related tests

czp13: Adding tests, a very fair point! 👍  I have no problem adding more tests, other than the PR starting to be quite big and making it hard to review it properly (so maybe a separate PR would have been the best solution if we want to add it).

Adding new features to 1.0, but with this you bought me :D, indeed this branch should be just for security fixes, and absolutely necessary commits, and so not much point to add new API-s to it, as probably we will not use it, so yap, let's go with your approach.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -594,7 +596,7 @@ private void registerDetachListenerForPendingInvocation(
         JavaScriptInvocationDetachListener listener = pendingJsInvocationDetachListeners
                 .computeIfAbsent(ownerNode, node -> {
                     JavaScriptInvocationDetachListener detachListener = new JavaScriptInvocationDetachListener();
-                    detachListener.registration = Registration.combine(
+                    detachListener.registration = combine(
                             () -> pendingJsInvocationDetachListeners
                                     .remove(node),
                             node.addDetachListener(detachListener));
```
mcollovati: I would remove the `combine` method, as it introduces a lot of overhead for a single call (e.g, creating `List`s, as it expects potentially more than one registration)
Instead, I would simply create a lambda here

```suggestion
                    Registration registration = node.addDetachListener(detachListener);
                    detachListener.registration = () -> {
                        pendingJsInvocationDetachListeners.remove(node);
                        registration.remove();
                    };
```

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +566,96 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
...
+        StateNode ownerNode = invocation.getOwner().orElse(ui.getInternals().getStateTree().getRootNode());
+
+        JavaScriptInvocationDetachListener listener = pendingJsInvocationDetachListeners
+                .computeIfAbsent(ownerNode, node -> {
+                    JavaScriptInvocationDetachListener detachListener = new JavaScriptInvocationDetachListener();
+                    detachListener.registration = combine(
+                            () -> pendingJsInvocationDetachListeners
+                                    .remove(node),
+                            node.addDetachListener(detachListener));
+                    return detachListener;
```
mcollovati, line 603: I would remove the combine method, as it introduces a lot of overhead for a single call (e.g, creating Lists, as it expects potentially more than one registration)
Instead, I would simply create a lambda here
```suggestion
                    Registration registration = node.addDetachListener(detachListener);
                    detachListener.registration = () -> {
                        pendingJsInvocationDetachListeners.remove(node);
                        registration.remove();
                    };
```

(Sorry for duplicate, I messed up with comments)

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
mshabarov: I agree with this approach, but how can you retrieve `JavaScriptInvocationDetachListener` for `readyToSend` invocations? You need this listener to get the registration to call remove on it.

Ah...you already have a map with node vs listener. Let do it then.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
czp13: Thank you vm, makes sense, cleaner! Fixed it, but I could not accept the suggestion (will squash all last commits also no worries), because  `return detachListener;` was needed still.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +567,84 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
...
+                .filter(invocation -> invocation.getOwner().map(StateNode::isVisible).orElse(true))
+                .collect(Collectors.toList());
+
+        pendingJsInvocationDetachListeners.entrySet().stream()
+                .flatMap(invocationListeners -> readyToSend.stream()
+                        .filter(invocation -> invocation.getOwner()
+                                .orElseGet(() -> ui.getInternals().getStateTree().getRootNode())
+                                .getId() == invocationListeners.getKey().getId()))
+                .forEach(invocation -> pendingJsInvocationDetachListeners.values()
+                        .forEach(listener -> listener.removePendingInvocation(invocation)));
```
mcollovati, line 577: It seems a bit complicated to me. Also, it should call `listener.onInvocationCompleted()` instead of `listener.removePendingInvocation()` otherwise the invocation will not be removed from the `JavaScriptInvocationDetachListener`

And as commented before, method `onInvocationCompleted` should be renamed with something more related to the context where it is applied in this PR

```suggestion
        StateNode rootNode = ui.getInternals().getStateTree().getRootNode();
        readyToSend.forEach(i -> {
            JavaScriptInvocationDetachListener listener = pendingJsInvocationDetachListeners.get(i.getOwner().orElse(rootNode));
            if (listener != null) {
                listener.onInvocationCompleted(i);
            }
        });
```

mcollovati (changes requested):

mcollovati (approved):

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -557,11 +563,83 @@ public List<JavaScriptInvocation> dumpPendingJavaScriptInvocations() {
...
+            UIInternals.this.pendingJsInvocations.removeIf(
+                    pendingInvocation -> pendingInvocation.equals(invocation));
+            if (invocationList.isEmpty() && registration != null) {
+                registration.remove();
+                registration = null;
+            }
+        }
 
-        return currentList;
+        void removePendingStatusAndRegistrationAndDetachListener(JavaScriptInvocation invocation) {
```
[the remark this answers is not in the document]
czp13: Super! Thank you! 🎉 
Yes, not the best, I agree, but if the scope is narrower then the name can be longer, based on Uncle Bob clean coding paradigm :), so that is why I was brave enough to make it longer. (https://bpoplauschi.github.io/2021/01/30/Clean-Code-Naming-by-Uncle-Bob-part-3.html)

I will squash the commits and then we can merge! 🙇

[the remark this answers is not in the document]
mcollovati: Let's also wait for @mshabarov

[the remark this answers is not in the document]
czp13: Oki, sure. I squashed the commits for clearer `rebase and merge`.

mshabarov (approved):

mshabarov:
@mcollovati if you have no comments, we can merge. @czp13 excellent job 👍

mcollovati (approved):
LGTM. @czp13 great job

czp13:
Thank you very much for the constant reviews and ideas for both of you! 🙇

PR #16876, opened 2023-05-24
chore: Upgrade to Vite 4.3.9

Artur-:


czp13 (approved):
Shall be safe as a minor version upgrade if validation passing.

Issue #16880, opened 2023-05-26
The property used for forcing a production build is missing the vaadin. prefix


Artur-:
### Description of the bug

To force a production build, you need to do
```
mvn install -Pproduction -Dforce.production.build=true
```

### Expected behavior

Shouldn't this be
```
mvn install -Pproduction -Dvaadin.force.production.build=true
```
?

### Minimal reproducible example

https://github.com/vaadin/skeleton-starter-flow-spring, v24.1

### Versions

- Vaadin / Flow version: 24.1.beta2

Issue #15969, opened 2023-02-21
Unclear when to use vaadin. -prefix for system properties


tepi:
### Description of the bug

https://vaadin.com/docs/latest/configuration/npm-pnpm states the following:

"Vaadin also supports using [pnpm](https://pnpm.io/) (also known as performant npm). To switch to pnpm, you can set the `vaadin.pnpm.enable` system property to `true`."

However, when running this: `mvn jetty:run -Pproduction -Dvaadin.pnpm.enable=true`, npm will be used.

When running `mvn jetty:run -Pproduction -Dpnpm.enable=true`, pnpm will be used.

### Expected behavior

As per the documentation, I expect the vaadin. -prefixed system property would work in the above case.

### Minimal reproducible example

Just run any starter with 24.1-SNAPSHOT with the commands mentioned above.

### Versions

- Vaadin / Flow version: 24.1-SNAPSHOT

Artur-:
All system properties are supposed to start with `vaadin.` to avoid conflicts with other libraries and frameworks. Thus pnpm works incorrectly if it works as described above

tepi:
The difference might be that when system properties are read through `com.vaadin.flow.server.AbstractPropertyConfiguration#getSystemProperty` it adds the "vaadin." prefix to the property name, but I believe in the above case the property is read directly by the maven plugin's `FlowModeAbstractMojo`, and I don't find similar logic there.

tepi:
`vaadin.` -prefix has been added to properties which are only used for the plugin configuration, so closing this ticket as fixed.

PR #16882, opened 2023-05-26
chore!: Add vaadin. prefix to recently added properties

mshabarov:
Add `vaadin.` prefix to `ci.build` and `force.production.build` properties added in 24.1 to satisfy naming rules.

Also see https://github.com/vaadin/docs/pull/2450

Fixes https://github.com/vaadin/flow/issues/16880

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -1,7 +1,7 @@
 name: Flow Validation
 on:
   push:
-    branches: [main, '24.0', '23.3', '23.2',  '23.1', '9.0']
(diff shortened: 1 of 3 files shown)


czp13 (approved):

mshabarov:
This issue is relevant https://github.com/vaadin/flow/issues/15969. 
IMO we need to review when/how prefix is added currently by Flow and possibly add a test checking that the all the properties can be set with prefix. This is for another PR though.

czp13 (approved):

PR #16884, opened 2023-05-26
fix: Change from ? extends Component to ? for VS Code

Artur-:
VS Code refuses to understand that a View which extends a Component can be used with the annotation when it is `Class<? extends Component>`. VS Code is broken here but this is a workaround to make VS Code work.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/LoadDependenciesOnStartup.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/LoadDependenciesOnStartup.java
@@ -20,7 +20,6 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-import com.vaadin.flow.component.Component;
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/LoadDependenciesOnStartup.java
```diff
@@ -55,6 +54,6 @@
      * @return a collection of views to load eagerly or an empty array to load
      *         all dependencies eagerly
      */
-    Class<? extends Component>[] value() default {};
+    Class<?>[] value() default {};
```
mshabarov, line 57: To avoid confusion, can we state in the javadoc that the classes should be Vaadin `Component` instances. Basically now anything can be passed as a parameter and won't work as expected.

Artur-: Done

mshabarov (approved):

czp13 (approved):

PR #16897, opened 2023-05-26
chore: use SDK http client instead of a separate http-client dependency

### LINKED ISSUE #14538: Explore removal of Apache's httpclient dependency from flow-server
### Describe your motivation

Currently vaadin-flow (flow-server) has a dependency to the httpclient (4.5.13) from Apache. The only class that needs it in flow-server [DefaultFileDownloader](https://github.com/vaadin/flow/blob/bf2b84fed9071d399248b3fedfc0c8922c2282c3/flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/DefaultFileDownloader.java) which is only used on build / development time.

Remove the dependency to Apache's httpclient by e.g. using Java 9's HttpClient or by moving of the whole class and its callers to the dedicated maven / gradle base plugin to reduce the dependency on runtime / production mode.

mstahv:
Less dependencies that might cause conflicts in user projects

Fixes #14538

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/StatisticsSender.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/StatisticsSender.java
@@ -22,14 +22,12 @@ import java.util.concurrent.atomic.AtomicReference;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.logging.Level;
 
-import org.apache.http.HttpResponse;
-import org.apache.http.HttpStatus;
-import org.apache.http.client.HttpClient;
-import org.apache.http.client.methods.HttpPost;
-import org.apache.http.entity.StringEntity;
-import org.apache.http.impl.client.HttpClientBuilder;
-import org.apache.http.util.EntityUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -44,8 +42,7 @@ public class StatisticsSender {
     /**
      * Creates a new instance connected to the given storage.
      *
-     * @param storage
-     *            the storage to use
+     * @param storage the storage to use
      */
     public StatisticsSender(StatisticsStorage storage) {
         this.storage = storage;
@@ -63,10 +60,9 @@ public class StatisticsSender {
     /**
      * Get the last server message.
      *
-     * @param json
-     *            The json returned by {@link StatisticsStorage#read()}
+     * @param json The json returned by {@link StatisticsStorage#read()}
      * @return The message string returned from server in last successful
-     *         requests.
+     * requests.
      */
     String getLastServerMessage(ObjectNode json) {
         return json.has(StatisticsConstants.FIELD_SERVER_MESSAGE)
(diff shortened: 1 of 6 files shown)


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/StatisticsSender.java
```diff
@@ -281,20 +278,20 @@ String sendStatistics(ObjectNode json) {
...
-            HttpResponse response = client.execute(post);
-            String responseStatus = response.getStatusLine().getStatusCode()
-                    + ": " + response.getStatusLine().getReasonPhrase();
+            HttpClient client = HttpClient.newHttpClient();
+            HttpRequest request = HttpRequest.newBuilder()
+                    .uri(URI.create(postUrl))
+                    .POST(HttpRequest.BodyPublishers.ofString(data))
+                    .header("Content-Type", "application/json").build();
+
+            HttpResponse<String> r = client.send(request,
```
mcollovati, line 287: nit: can we rename the variable to 'response' or something similar?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/StatisticsSender.java
```diff
@@ -304,11 +301,12 @@ private static ObjectNode postData(String postUrl, String data) {
...
             // Update the status and return the results
-            result.put(StatisticsConstants.FIELD_LAST_STATUS, responseStatus);
+            result.put(StatisticsConstants.FIELD_LAST_STATUS,
+                    statusCode + ": "); // TODO is reason phrase needed here
             return result;
 
-        } catch (IOException e) {
-            getLogger().debug("Failed to send statistics.", e);
+        } catch (IOException | InterruptedException ex) {
+            getLogger().debug("Failed to send statistics.", ex);
```
mcollovati, line 309: Interrupt flag should probably be reset when catching `InterruptedException`

`Thread.currentThread().interrupt();`

mcollovati (changes requested):
Some minor changes.
The hard-coded timeout settings are a bit opinionated.
It would be good to make them configurable.
Maybe trying to get the value from system properties, otherwise fallback to default?

mstahv:
@mcollovati timeouts were opinionated indeed, copied from some tutorial 😂 I removed them now, going with defaults would be fine I guess? Or were those configurable in the past 🤔

mcollovati:
I was thinking they were configurable through system properties, as the client builder had `useSystemProperties()` setting.
But it seems like there are no properties for the timeouts in httpclient.
So removing them from the PR could be fine.

mstahv:
👌 That was done in https://github.com/vaadin/flow/pull/16897/commits/9becd05f08c1509ea97d11f2823c807e74ef4488

mcollovati:
And this could be interesting to set up tests with proxy: https://github.com/pan-dora/jdk11-httpclient-tests/blob/master/httpclient/src/test/java/net/httpclient/ProxyAuthTest.java

knoobie:
> Another suggested workaround is to set the Proxy-Authorization header in the request.

keep in mind this old security fix :) https://github.com/apache/httpcomponents-client/commit/a572756592c969affd0ce87885724e74839176fb

mstahv:
Would be great if we'd know if and how developers use proxies 😬 Building extensive integration tests for this is probably not worth the investement but some smaller smoke test would be great.

mcollovati:
I propose not adding custom code for handling proxy authentication, but instead updating the documentation at https://vaadin.com/docs/latest/configuration/development-mode/node-js#proxy-settings-for-downloading-the-front-end-toolchain to suggest setting the `-Djdk.http.auth.tunneling.disabledSchemes` if needed

mcollovati (approved):

mcollovati:
Created vaadin/docs#2475 and #16966

mcollovati (approved):

platosha:
Would it be feasible to backport this to 24.1? Hilla native compilation is broken by httpclient dependency that brings `commons-logging`, this would help.

Issue #16881, opened 2023-05-26
Maven plugin validation issue report


Artur-:
### Description of the bug

When you run `mvn install -Pproduction -Dforce.production.build=true -Dmaven.plugin.validation=VERBOSE` in 24.1 beta2

```
[WARNING] Plugin validation issues were detected in 1 plugin(s)
[WARNING] 
[WARNING]  * com.vaadin:vaadin-maven-plugin:24.1.0.beta2
[WARNING]   Declared at location(s):
[WARNING]    * com.example:demo:0.0.1-SNAPSHOT (pom.xml) @ line 75
[WARNING]   Used in module(s):
[WARNING]    * com.example:demo:0.0.1-SNAPSHOT (pom.xml)
[WARNING]   Plugin issue(s):
[WARNING]    * Plugin should declare these Maven artifacts in `provided` scope: [org.apache.maven:maven-resolver-provider:3.8.6, org.apache.maven:maven-settings-builder:3.8.6, org.apache.maven:maven-repository-metadata:3.8.6, org.apache.maven:maven-artifact:3.8.6, org.apache.maven:maven-core:3.8.6, org.apache.maven:maven-plugin-api:3.8.6, org.apache.maven:maven-model-builder:3.8.6, org.apache.maven:maven-model:3.8.6, org.apache.maven:maven-builder-support:3.8.6, org.apache.maven:maven-settings:3.8.6]
```

### Expected behavior

There should be no validation issues

### Minimal reproducible example

```
npx @vaadin/cli init hello
cd hello
mvn install -Pproduction -Dforce.production.build=true -Dmaven.plugin.validation=VERBOSE
```

### Versions

- Vaadin / Flow version: 24 beta2

knoobie:
To get this warnings: Maven 3.9.2+ are needed

PR #16911, opened 2023-05-30
fix: update plugins

caalador:
Update available plugins
to not get validation issues.

This lessens some of them when
building flow.

partly to #16881

### DIFF
--- a/flow-plugins/flow-gradle-plugin/pom.xml
+++ b/flow-plugins/flow-gradle-plugin/pom.xml
@@ -62,7 +62,6 @@
             <!-- copy generated JARs -->
             <plugin>
                 <artifactId>maven-resources-plugin</artifactId>
-                <version>3.3.1</version>
                 <executions>
                     <execution>
                         <id>copy-gradle-jars</id>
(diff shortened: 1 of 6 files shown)


Artur- (approved):

Artur- (approved):
There is something fishy here as the api diff build fails with

[ERROR] We have duplicated artifacts attached.
[11:44:09 ]Failed to execute goal org.apache.maven.plugins:maven-source-plugin:3.3.0:jar-no-fork (attach-sources) on project flow-commons-upload: Presumably you have configured maven-source-plugn to execute twice times in your build. You have to configure a classifier for at least on of them.

Artur- (changes requested):
There is something fishy here as the api diff build fails with

[ERROR] We have duplicated artifacts attached.
[11:44:09 ]Failed to execute goal org.apache.maven.plugins:maven-source-plugin:3.3.0:jar-no-fork (attach-sources) on project flow-commons-upload: Presumably you have configured maven-source-plugn to execute twice times in your build. You have to configure a classifier for at least on of them.

Artur-:
.... it was using `mvn verify clean install`... maybe it works now that it uses `mvn clean install` only

Artur- (approved):

PR #16935, opened 2023-06-01
fix: log error messages

caalador:
Log error messages and throw more
generic exceptions.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/PublishedServerEventHandlerRpcHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/PublishedServerEventHandlerRpcHandler.java
@@ -26,6 +26,7 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/PublishedServerEventHandlerRpcHandler.java
```diff
@@ -169,7 +171,9 @@ private static Optional<Method> findMethod(Component instance,
             String msg = String.format("Class '%s' contains "
                     + "several event handler method with the same name '%s'",
                     instance.getClass().getName(), methodName);
-            throw new IllegalStateException(msg);
+            getLogger().error(msg);
+            throw new IllegalStateException(
+                    "Method conflict in event handler with multiple methods with same name.");
```
Legioth, line 176: Could maybe also explicitly say that more information has been logged. Would also be good to use the same terminology (e.g. "Faulty method invocation") both in the exception message and in the log message so that the corresponding log message is easier to find.

caalador: Added see logs, updated both to begin with same message

mcollovati (approved):

PR #16942, opened 2023-06-01
fix: Find production bundle theme on Windows

Artur-:
Classloader always uses a / separated path


manolo:
I tested it in windows, and fixes the problem

czp13 (approved):

PR #16976, opened 2023-06-07
feat: Overloaded RouteParam constructor with Integer and Long

### LINKED ISSUE #16972: RouteParameter should support simple data types other than String
RouteParameters has methods like getLong and getInteger.
But when I create a RouteParamter it just takes a String as the value.

Therefore I must convert the value into a String. It would be more convenient to have RouteParameter constructors that take Integer and Long as the value in addition to the constructor that takes a String as the value.

mstahv:
Closes #16972

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/RouteParam.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/RouteParam.java
@@ -43,8 +43,8 @@ public class RouteParam extends Pair<String, String> {
      * @param value
      *            the value of the parameter.
      */
-    public RouteParam(String name, Integer value) {
-        super(name, value.toString());
+    public RouteParam(String name, int value) {
+        super(name, Integer.toString(value));
     }
 
     /**
@@ -55,8 +55,8 @@ public class RouteParam extends Pair<String, String> {
      * @param value
      *            the value of the parameter.
      */
-    public RouteParam(String name, Long value) {
-        super(name, value.toString());
+    public RouteParam(String name, long value) {
+        super(name, Long.toString(value));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/router/RouteParam.java
```diff
@@ -35,6 +35,30 @@ public RouteParam(String name, String value) {
...
 
+    /**
+     * Creates a new route parameter.
+     *
+     * @param name
+     *            the name of the parameter.
+     * @param value
+     *            the value of the parameter.
+     */
+    public RouteParam(String name, Integer value) {
```
mcollovati, line 46: I wonder if we can use a primitive int instead of a wrapper object.

flow-server/src/main/java/com/vaadin/flow/router/RouteParam.java
```diff
@@ -35,6 +35,30 @@ public RouteParam(String name, String value) {
...
+
+    /**
+     * Creates a new route parameter.
+     *
+     * @param name
+     *            the name of the parameter.
+     * @param value
+     *            the value of the parameter.
+     */
+    public RouteParam(String name, Long value) {
```
mcollovati, line 58: I wonder if we can use a primitive long instead of a wrapper object.
Otherwise, if null is an acceptable value, we should handle it

flow-server/src/main/java/com/vaadin/flow/router/RouteParam.java
mcollovati: Otherwise, if null is an acceptable value, we should handle it

mcollovati (approved):

PR #16977, opened 2023-06-07
feat: Added shorthands for QueryParameters class

### LINKED ISSUE #16975: QueryParameter should have a method to get a single parameter with a single value
QueryParameters has a method simple() to create quey parameters with a single value.

But there is no method to get a parameter with a single value.

mstahv:
Closes #16975

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -218,7 +218,8 @@ public class QueryParameters implements Serializable {
      * @return query parameter value
      */
     public Optional<String> getSingleParameter(String key) {
-        return parameters.getOrDefault(key, Collections.emptyList()).stream().findFirst();
+        return parameters.getOrDefault(key, Collections.emptyList()).stream()
+                .findFirst();
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -204,7 +204,7 @@ public class QueryParameters implements Serializable {
      *         given key
      */
     public List<String> getParameter(String key) {
-        return parameters.get(key);
+        return parameters.getOrDefault(key, Collections.emptyList());
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -203,7 +203,7 @@ public class QueryParameters implements Serializable {
      * @return query parameters or null if there are no parameters with the
      *         given key
      */
-    public List<String> getParameter(String key) {
+    public List<String> getParameters(String key) {
         return parameters.getOrDefault(key, Collections.emptyList());
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -200,7 +200,7 @@ public class QueryParameters implements Serializable {
      *
      * @param key
      *            the key of query parameters to fetch
-     * @return query parameters or null if there are no parameters with the
+     * @return query parameters or an empty list if there are no parameters with the
      *         given key
      */
     public List<String> getParameters(String key) {
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -30,6 +30,7 @@ import java.util.stream.Stream;
 
 import com.vaadin.flow.internal.UrlUtil;
 import java.util.NoSuchElementException;
+import java.util.Optional;
 
 /**
  * Holds query parameters information.
@@ -217,11 +218,11 @@ public class QueryParameters implements Serializable {
      *            the key of query parameters to fetch
      * @return query parameter value
      */
-    public String getSingleParameter(String key) {
+    public Optional<String> getSingleParameter(String key) {
         try {
-            return parameters.get(key).get(0);
+            return Optional.ofNullable(parameters.get(key).get(0));
         } catch (Exception e) {
-            throw new NoSuchElementException(e);
+            return Optional.empty();
         }
     }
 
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -189,6 +191,40 @@ public Map<String, List<String>> getParameters() {
...
+     * Returns the first query parameter values mapped with the given key.
+     * <p>
+     * Example: Calling with key value "one" with
+     * {@code https://example.com/?one=1&two=2&one=3} will return {@code 1}
+     *
+     * @param key
+     *            the key of query parameters to fetch
+     * @return query parameter value
+     */
+    public String getSingleParameter(String key) {
```
knoobie, line 220: Another overload with (String key, String default) or (String key, T default)? (I personally hate exceptions and prefer default values)

mstahv: Or should this actually return Optional, although you know my generic attitude against them ;-)

mstahv: Now returns Optional so you ought to be able to provide the default value to the chain. Open for other options still as well.

knoobie: I prefer overloaded methods more but that's also a personal taste :) Thanks for the change!

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -189,6 +192,40 @@ public Map<String, List<String>> getParameters() {
...
+     * @param key
+     *            the key of query parameters to fetch
+     * @return query parameter value
+     */
+    public Optional<String> getSingleParameter(String key) {
+        try {
+            return Optional.ofNullable(parameters.get(key).get(0));
+        } catch (Exception e) {
+            return Optional.empty();
+        }
```
mcollovati, line 226: Can this method be rewritten to avoid using exceptions for its logic?
Something like
```suggestion
        parameters.getOrDefault(key, Collections.emptyList()).stream().findFirst()
```
Or just get the parameter and check for null and empty

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -189,6 +192,40 @@ public Map<String, List<String>> getParameters() {
...
+     * Example: Calling the method with key "one" for a parameters like
+     * {@code https://example.com/?one=1&two=2&one=3} will result in the
+     * corresponding list: {@code [1, 3]}
+     *
+     * @param key
+     *            the key of query parameters to fetch
+     * @return query parameters or null if there are no parameters with the
+     *         given key
+     */
+    public List<String> getParameter(String key) {
```
mcollovati, line 207: I wonder if an empty list could be better when there are no parameters

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
mstahv: Done, also changed the previous suggestion now compiles. Maybe I'll write quick smoke tests for the shorthands still.

mstahv:
Went and renamed the other method while doing smoke tests :-)

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -189,6 +191,37 @@ public Map<String, List<String>> getParameters() {
...
+    /**
+     * Returns query parameter values mapped with the given key.
+     * <p>
+     * Example: Calling the method with key "one" for a parameters like
+     * {@code https://example.com/?one=1&two=2&one=3} will result in the
+     * corresponding list: {@code [1, 3]}
+     *
+     * @param key
+     *            the key of query parameters to fetch
+     * @return query parameters or null if there are no parameters with the
```
mcollovati, line 203:
```suggestion
     * @return query parameters or an empty list if there are no parameters with the
```

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -189,6 +191,37 @@ public Map<String, List<String>> getParameters() {
...
+
+    /**
+     * Returns the first query parameter values mapped with the given key.
+     * <p>
+     * Example: Calling with key value "one" with
+     * {@code https://example.com/?one=1&two=2&one=3} will return {@code 1}
+     *
+     * @param key
+     *            the key of query parameters to fetch
+     * @return query parameter value
```
mcollovati, line 218:
```suggestion
     * @return query parameter value or empty if there are no parameters
     *               with the given key
```

flow-server/src/test/java/com/vaadin/flow/router/QueryParametersTest.java
```diff
@@ -221,6 +222,18 @@ public void parameterWithEmptyValue() {
...
+    public void shortHands() {
+        QueryParameters qp1 = QueryParameters.of("foo", "bar");
+        Optional<String> singleParameter = qp1.getSingleParameter("foo");
+        Assert.assertEquals("bar", singleParameter.get());
+        Assert.assertTrue(qp1.getSingleParameter("bar").isEmpty());
+
+        List<String> parameters = qp1.getParameters("foo");
+        Assert.assertEquals("bar", parameters.get(0));
+        Assert.assertEquals(1, parameters.size());
+    }
```
mcollovati, line 235: Could you also add an assertion for getting an empty list for `getParameters("bar")`?

mcollovati (changes requested):

mcollovati (approved):

knoobie:
@mstahv Just came across this in V24.1 again because I needed them.. What would you think about a method 

`public boolean contains(String key, String value)` which you can easily call like so:

```java
  if(event.getLocation().getQueryParameters().contains("action", "lorem")) {
    // do stuff because the query contains  /blub?action=lorem
  }
  

```

mstahv:
Do you actually have multiple action parameters? If not, wouldn't that be then the same as:

```
event.getLocation().getQueryParameters().getSingleParameter("action").get().equals("lorem")
```

Decent, but those damn optionals :-)

knoobie:
Optionals 👎  and this would break, because the action parameter is optional, therefore calling get() on an empty optional would throw (and I dislike throwing even more than optionals)

mstahv:
I already wrote that and a test, but not sure if I should make the PR. Noticed that there is is getParameters(String), that always returns a List. So this should work:

```
event.getLocation().getQueryParameters().getParameters("foo").contains("bar")
```

More readable and almost as compact 🤔

knoobie:
Oh, that overload isn't available in 24.1 :) I'm probably fine with that :)

mstahv:
Added in this PR, coming soon in 24.2😎

Issue #16968, opened 2023-06-06
Vaadin 23.3.13 logging Broadcaster messages repeatedly


banatom73:
### Description of the bug

After upgrade from 23.3.12 to 23.3.13 application starts to log broadcaster messages repeatedly.
However, the application seems to work properly.
```
2023-06-06 16:31:55.648  WARN VV --- [sphere-Shared-0] org.atmosphere.cpr.DefaultBroadcaster    : This message Entry{message=com.vaadin.flow.server.communication.AtmospherePushConnection$PushMessage@65b85c8b, type=RESOURCE, future=org.atmosphere.cpr.BroadcasterFuture@135a2a27} will be lost
2023-06-06 16:31:55.649  WARN VV --- [sphere-Shared-0] org.atmosphere.cpr.DefaultBroadcaster    : Failed to submit broadcast handler runnable to for Broadcaster/*

java.lang.NumberFormatException: Cannot parse null string
	at java.base/java.lang.Integer.parseInt(Integer.java:630)
	at java.base/java.lang.Integer.parseInt(Integer.java:786)
	at com.vaadin.flow.server.communication.LongPollingCacheFilter.filter(LongPollingCacheFilter.java:60)
	at org.atmosphere.cpr.BroadcasterConfig.filter(BroadcasterConfig.java:452)
	at org.atmosphere.cpr.DefaultBroadcaster.perRequestFilter(DefaultBroadcaster.java:822)
	at org.atmosphere.cpr.DefaultBroadcaster.deliverPush(DefaultBroadcaster.java:695)
	at org.atmosphere.cpr.DefaultBroadcaster.push(DefaultBroadcaster.java:572)
	at org.atmosphere.cpr.DefaultBroadcaster$1.run(DefaultBroadcaster.java:423)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:264)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1395)
	at java.base/java.util.concurrent.ForkJoinTask.doExec$$$capture(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java)
[...]

Issue #16775, opened 2023-05-11
Vaadin 24.x logging Broadcaster messages repeatedly


awillerding:
### Description of the bug

I just upgraded my Vaadin app from 23.x to 24.0.x making the necessary javax to jakarta changes and correspondingly upgrading my runtime environment to Tomcat 10.1.8.

I'm using @Push(PushMode.AUTOMATIC) and my Servlet uses the following

```
@Route("")
@PWA(name = "MyApp", shortName = "MyApp") 
@Push(value = PushMode.AUTOMATIC) 
public class Main extends AppLayout implements AppShellConfigurator {
```

I am getting the following WARN repeatedly logged and my UI is sluggish. is there something else I need to configure or need to change for the upgrade to Vaadin 24.x to fix this issue or is this a bug?

```
Atmosphere-Shared-0] WARN org.atmosphere.cpr.DefaultBroadcaster - This message Entry{message=com.vaadin.flow.server.communication.AtmospherePushConnection$PushMessage@65df019a, type=RESOURCE, future=org.atmosphere.cpr.BroadcasterFuture@4dfa5742} will be lost
[Atmosphere-Shared-0] WARN org.atmosphere.cpr.DefaultBroadcaster - Failed to submit broadcast handler runnable to for Broadcaster/*
java.lang.IllegalStateException: The request object has been recycled and is no longer associated with this facade
    at org.apache.catalina.connector.RequestFacade.checkFacade(RequestFacade.java:856)
    at org.apache.catalina.connector.RequestFacade.getHeader(RequestFacade.java:505)
    at jakarta.servlet.http.HttpServletRequestWrapper.getHeader(HttpServletRequestWrapper.java:82)
    at org.atmosphere.cpr.AtmosphereRequestImpl.getHeader(AtmosphereRequestImpl.java:351)
    at org.atmosphere.cpr.AtmosphereRequestImpl.getHeader(AtmosphereRequestImpl.java:336)
    at com.vaadin.flow.server.communication.LongPollingCacheFilter.filter(LongPollingCacheFilter.java:56)
    at org.atmosphere.cpr.BroadcasterConfig.filter(BroadcasterConfig.java:452)
    at org.atmosphere.cpr.DefaultBroadcaster.perRequestFilter(DefaultBroadcaster.java:818)
    at org.atmosphere.cpr.DefaultBroadcaster.deliverPush(DefaultBroadcaster.java:691)
[...]

PR #16996, opened 2023-06-09
fix: store LastSeenServerSyncId in atmosphere resource session

mcollovati:
## Description

Storing the header value in AtmosphereResource session should prevent issues with long polling and servlet container that recycles requests. This change stores in the resource session the value got from client during connection so that it will be available even if the underliyng request gets nullified.

Part of #16968
Part of #16775

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/LongPollingCacheFilter.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/LongPollingCacheFilter.java
@@ -17,9 +17,11 @@
 package com.vaadin.flow.server.communication;
 
 import java.io.Serializable;
+import java.util.Optional;
 
 import org.atmosphere.cache.BroadcastMessage;
 import org.atmosphere.cpr.AtmosphereResource;
+import org.atmosphere.cpr.AtmosphereResourceSession;
 import org.atmosphere.cpr.BroadcasterCache;
 import org.atmosphere.cpr.PerRequestBroadcastFilter;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/LongPollingCacheFilter.java
```diff
@@ -48,16 +50,32 @@ private static Logger getLogger() {
...
+    static void onConnect(AtmosphereResource resource) {
+        Integer syncId = Optional
+                .ofNullable(
+                        resource.getRequest().getHeader(SEEN_SERVER_SYNC_ID))
+                .map(Integer::parseInt).orElse(null);
+        if (resource.transport() == AtmosphereResource.TRANSPORT.LONG_POLLING
+                && syncId != null) {
+            AtmosphereResourceSession session = resource.getAtmosphereConfig()
+                    .sessionFactory().getSession(resource);
+            session.setAttribute(SEEN_SERVER_SYNC_ID, syncId);
```
mshabarov, line 62: Not sure I know how it works - is server sync ID value is per atmosphere request or is it per atmosphere resource ? I.e. may it be a case that this sync ID is changed between multiple requests and multiple `filter` method calls? Because usually session covers multiple requests happening during it's lifespan.

flow-server/src/test/java/com/vaadin/flow/server/communication/LongPollingCacheFilterTest.java
```diff
@@ -154,7 +163,7 @@ private void setSeenServerSyncIdHeader(int id, int... ids) {
                 .getHeader(LongPollingCacheFilter.SEEN_SERVER_SYNC_ID))
                 .thenReturn(Integer.toString(id), IntStream.of(ids)
                         .mapToObj(Integer::toString).toArray(String[]::new));
-
+        LongPollingCacheFilter.onConnect(resource);
```
mshabarov, line 166: The new codes in this test only places the sync ID into the session, but doesn't assert that the value is accessible within `filter` method call. Shall we add some checks?

flow-server/src/main/java/com/vaadin/flow/server/communication/LongPollingCacheFilter.java
mshabarov: Then it makes sense.

flow-server/src/test/java/com/vaadin/flow/server/communication/LongPollingCacheFilterTest.java
mcollovati: The tests will fail if the value cannot be retrieved.
But it makes sense to add some tests for the `onConnect` method

mshabarov: Yeah, I propose to add one or more tests, that explicitly check this attribute is there. Otherwise, if something raised up around this functionality, it would be probably harder to investigate.

mshabarov (approved):

PR #17001, opened 2023-06-12
chore: Add 9.2.0 as unsupported npm version

caalador:
Add check for known bad npm
version so that the user
knows to update if using that
exact version.


mshabarov (approved):
LGTM. Tested also npm version `9.3.1` just in case - no issues with building Flow project found.

PR #17003, opened 2023-06-12
feat: Dev tools plugin support

Artur-:
You can implement a dev tools plugin by 
1. Adding a class that implements `DevToolsMessageHandler` to handle messages from the browser
2. Annotate the class with `@JsModule(value="./my-plugin.ts", developmentOnly = true)` to load the client side part
3. Add a service loader metadata file `src/main/resources/META-INF/services/com.vaadin.base.devserver.DevToolsMessageHandler` with the name of the class

The client side part needs to register itself using
```
const plugin: DevToolsPlugin = {
  init: function (devToolsInterface: DevToolsInterface): void {
    // Add a tab using the interface, store the reference for communication later
  }
};

(window as any).Vaadin.devToolsPlugins.push(plugin);
```


mshabarov:
https://github.com/vaadin/docs/issues/2508

flow-tests/test-dev-mode/frontend/devtools-plugin.ts
```diff
@@ -0,0 +1,38 @@
+
+import type {
+  DevToolsInterface,
+  DevToolsPlugin //@ts-ignore
```
mcollovati, line 4: Can the `ts-ignore` annotation be removed now that #17019 has been merged?

mcollovati (commented):
A minor change, other than that LGTM.
A nice improvement would be allowing the message handler to push back data to the client side, for initialization stuff or to react to a server command.

Something like this on the server side
```
public interface DevToolsMessageHandler {
    handleDevToolsMessage(String command, JsonObject data, Consumer<JsonObject> reply);
}
```

and that on the client side
```
export interface DevToolsInterface {
    ....
    onMessage(command: string, data: any)
}
```

Artur-:
Pushing data from the server to the client would be nice but the API should preferably be such that you do not need to receive a message from the browser first. That's how e.g. the server info tab works: it pushes data immediately on startup. As the `DevToolsMessageHandler` instance is created by a service loader, maybe we need to have an `init` type of method also, e.g.
```
public interface DevToolsMessageHandler {
   public void init(DevToolsInterface devToolsInterface);
    public boolean handleDevToolsMessage(String command, JsonObject data);

}
```

so you can store the reference and use `devToolsInterface.send(someJson)`

Artur-:
It does not work though because there is one plugin instance per application and there can be N websocket connections (one per browser tab). So maybe `init(DevToolsInterface devToolsInterface)` should become `handleConnect` and the `DevToolsInterface` should also be passed to `handleDevToolsMessage` for responses. That method could also be just `handleMessage`

Artur-:
Then the second problem is that the dev tools window would need to somehow wait for plugins to initialize themselves before handling the initial messages from the server. `generated-flow-imports.js` which includes the dev tools plugins is loaded using a dynamic import

mcollovati:
Another potential issue is building for production when adding a custom plugin to a project downloaded from start.vaadin.com.

Since the `DevToolsMessageHandler` interface is defined in the `vaadin-dev-server` module, when building for production, the compilation fails because in the `production` profile defines an exclusion for `vaadin-dev-server`.

Maybe not a big issue, if plugins will be mostly packaged as add-ons

Artur-:
Assuming the plugin project (standalone addon or an app project) depends on `vaadin-dev-server` as `provided`, it should work as far as I understand. The project can be compiled but `vaadin-dev-server` does not end up in the production bundle nor is the service loader invoked so the plugin class is never loaded.

Artur-:
But apparently that's not how Spring Boot works... So you would need to either put the plugin in a separate project or exclude it as described in https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#packaging.examples.exclude-dependency

Artur-:
Updated to include server to client communication that queues messages on the client side if a plugin has not yet been initialized.

flow-tests/test-dev-mode/frontend/devtools-plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+    });
+  }
+}
+
+const plugin: DevToolsPlugin = {
+  init: function (devToolsInterface: DevToolsInterface): void {
+    devToolsInterface.addTab('Hello', () => html`<my-tool .devToolsInterface=${devToolsInterface}></my-tool>`);
+    devToolsInterface.addMessageListener((message: ServerMessage): boolean => {
+      // It is not clear that the tab has been rendered
+      const content = devToolsInterface.getContent();
```
Artur-, line 48: This part feels quite icky. Maybe it would be better if the plugin could render and update its own element and not care about when the dev tools window shows it or does not show it

Legioth: Agreed. What if each tab would only be identified as a custom element tag name and then future interaction would be through methods or properties on an instance of that element? In that way, `<my-tool>` would have a `receiveServerMessage` instance method that is invoked when there's a message.

mcollovati: Just a random thought. Couldn't the single plugin register an event listener on itself for the commands of interest, and then the `handleFrontendMessage` method dispatches an event to every registered plugin?

Artur-: As a user, I want to define the content for the dev tools window as a lit element. I would also want to be able to consider this a singleton and make the single instance the one handling communication with the server, for simplicity. Then I don't have to pass properties here and there or worry about an element maybe existing.

The only way I can imagine this to work is if the dev tools window creates the element instances eagerly for each tab. Then the element can already exist when the first message from the server is received.

Artur-:
We should change how/when the element is created so that somebody can understand how to build a plugin. The thing discussed above

Artur-:
Based on the doc draft here https://github.com/vaadin/docs/issues/2508#issuecomment-1689460203, I would say we should rename `handleDevToolsMessage` to `handleMessage`

Artur-:
Renamed

caalador (approved):

PR #17010, opened 2023-06-13
fix: support silent updates for tsconfig.json from Hilla 2.0

platosha:
Fixes vaadin/hilla#960

For some time between 2.0 and 2.1 version releases Hilla starters had `"moduleResolution": "nodeNext"` in the default `tsconfig.json`, while still using the same version key.

This change enables silent updates from those to the latest `tsconfig.json`.


Lodin (approved):

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig-v23.3.4-hilla.json
```diff
@@ -0,0 +1,39 @@
...
+// For more information about the configurations, please refer to http://www.typescriptlang.org/docs/handbook/tsconfig-json.html
+{
+  "flow_version": "23.3.4",
+  "compilerOptions": {
+    "sourceMap": true,
+    "jsx": "react-jsx",
+    "inlineSources": true,
+    "module": "esNext",
+    "target": "es2020",
+    "moduleResolution": "nodeNext",
```
Lodin, line 14:
```suggestion
    "moduleResolution": "bundler",
```

Lodin (changes requested):

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig-v23.3.4-hilla.json
platosha: This change will ruin the fix. The intention here is to match with the early source revision in the starter to trigger auto-update from it: https://github.com/vaadin/skeleton-starter-hilla-react/blob/b908ba644707c0d784439955a49fe5d3743a023a/tsconfig.json

Note that the latest `tsconfig.json` (which is the auto-update target) is already using `bundler` here: https://github.com/vaadin/flow/blob/main/flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json

Lodin: Ah, ok then

Lodin (approved):

PR #17029, opened 2023-06-14
feat: introduce manual validation mode for manual control of invalid state

vursen:
## Description

The PR introduces manual validation mode that disables the component's built-in constraint validation, allowing you to manually control the invalid state and error messages without conflicts.

A part of https://github.com/vaadin/flow-components/issues/5074

## Type of change

- [x] Feature


flow-server/src/main/java/com/vaadin/flow/component/HasValidation.java
```diff
@@ -25,6 +25,26 @@
  * @since 1.0.
  */
 public interface HasValidation extends Serializable {
+    /**
+     * Sets whether the component's internal validation is disabled.
+     * <p>
+     * When disabled, the component doesn't automatically validate values
+     * against constraints such as required and so on, which also leads to the
+     * invalid state not getting updated. However, it's still possible to use
+     * Binder or implement custom validation logic.
```
mshabarov, line 34: The javadoc given describes technical details for what happen if the validation is disabled.
It's not really clear what's the value for developers.
I'd make it explicit and highlight the following:
- what is an use case/problem developers can solve with this method. If I understood correct, this is a case when a custom validation logic is used, i.e. doing fields validation on value change.
- describe the scope of this configuration, i.e. is it's disabled, which validation steps are affected - only flow component constraints, affects also binder, default validator ?
- maybe also give some codes example.

flow-server/src/main/java/com/vaadin/flow/component/HasValidation.java
```diff
@@ -25,6 +25,26 @@
...
+     *
+     * @param disabled
+     *            whether the validation should be disabled.
+     */
+    void setInternalValidationDisabled(boolean disabled);
```
mshabarov, line 39: I don't really like the "internal" word in the middle. It makes me think that this API is internal and not supposed to be used by developers. `setComponentValidationDisabled` would be better IMO.

mshabarov (changes requested):
@vursen @yuriy-fix this apparently needs more discussion of what would the implementation of this API do for components. The main question is what to do with default validator and Binder, if disabled=true.

flow-server/src/main/java/com/vaadin/flow/component/HasValidation.java
vursen: We agreed on renaming `setInternalValidationDisabled` to `setManualValidation`.

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/component/HasValidation.java
mshabarov: Look very good now, thanks 👍

flow-server/src/main/java/com/vaadin/flow/component/HasValidation.java
```diff
@@ -25,6 +25,37 @@
...
+     *      } else {
+     *          field.setInvalid(false);
+     *      }
+     * });
+     * </pre>
+     *
+     * @param enabled
+     *            whether to enable manual validation mode.
+     */
+    void setManualValidation(boolean enabled);
```
mshabarov, line 58: This is a breaking change for add-on (custom components) which extend this interface. I'd avoid do it in the minor releases. Let's think about providing a default implementation. Even a default implementation is fine IMO if we state it in the javadoc. Composite component implementations can just call `setManualValidation(enabled)` for each child component.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/HasValidation.java
```diff
@@ -25,6 +25,45 @@
...
+     * The method should be implemented by components that support built-in
+     * validation.
+     *
+     * @param enabled
+     *            whether to enable manual validation mode.
+     * @throws UnsupportedOperationException
+     *             if the component doesn't support built-in validation.
+     */
+    default void setManualValidation(boolean enabled) {
+        throw new UnsupportedOperationException();
```
mshabarov, line 65: Discussed offline to make the default implementation no-op (empty), so that add-ons can safely call this method according to `setInvalid` javadoc even if an add-on doesn't implement it.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/HasValidation.java
vursen: Done.

mshabarov (approved):

mshabarov (approved):

PR #17035, opened 2023-06-15
fix: get correct vaadin-core-versions.json

### LINKED ISSUE #17017: Platform.getVaadinVersion() randomly returns Vaadin or Hilla version in hybrid apps
### Description of the bug

`Platform.getVaadinVersion()` reads the `vaadin-core-versions.json` file from classpath, parses out the `"platform"` string and returns it as the Vaadin version (e.g. 24.1.0). With Vaadin apps, the file is present in `vaadin-core.jar`.

The problem is that `hilla.jar` also contains the same `vaadin-core-versions.json` file, with `"platform"` string referring to Hilla version, e.g. 1.3.7 or 2.0.6. If both vaadin-core and hilla is present on the classpath in case of hybrid apps, this will lead to a problematic behavior.

Since `Platform.getVaadinVersion()` calls `Platform.class.getClassLoader().getResourceAsStream("vaadin-core-versions.json")`, it may parse either the file from `vaadin-core.jar` or from `hilla.jar`, whichever happens to be sooner on the classpath. That will lead to unpredictable results returned by the `Platform.getVaadinVersion()` function.

### Expected behavior

`Platform.getVaadinVersion()` should always return Vaadin version (e.g. 24.1.0), regardless of whether Hilla is or isn't on the classpath.

Quite hard to reproduce unfortunately, requires a hybrid project with two different classpath orderings.

### Versions

- Vaadin / Flow version:
- Java version:
- OS version:
- IDE (if applicable):

caalador:
Get the correct core-versions
file if both hilla and flow are available
on the classpath.

Fixes #17017

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
@@ -46,16 +46,12 @@ public class Platform implements Serializable {
         try {
             List<URL> coreVersions = Collections.list(getClassloader()
                     .getResources(Constants.VAADIN_CORE_VERSIONS_JSON));
-            if (coreVersions.size() == 1) {
-                return loadVersionForUrl(coreVersions.get(0));
-            } else {
-                Optional<URL> vaadinCoreJson = coreVersions.stream()
-                        .filter(url -> url.toString()
-                                .contains("com/vaadin/vaadin-core-internal"))
-                        .findFirst();
-                if (vaadinCoreJson.isPresent()) {
-                    return loadVersionForUrl(vaadinCoreJson.get());
-                }
+            Optional<URL> vaadinCoreJson = coreVersions.stream()
+                    .filter(url -> url.toString()
+                            .contains("com/vaadin/vaadin-core-internal"))
+                    .findFirst();
+            if (vaadinCoreJson.isPresent()) {
+                return loadVersionForUrl(vaadinCoreJson.get());
             }
         } catch (IOException ioe) {
             LoggerFactory.getLogger(Platform.class)
--- a/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
@@ -15,8 +15,13 @@
  */
 package com.vaadin.flow.server;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.Serializable;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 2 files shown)


flow-server/src/test/java/com/vaadin/flow/server/PlatformVersionTest.java
```diff
@@ -0,0 +1,98 @@
...
+            platform.when(
+                    () -> Platform.loadVersionForUrl(Mockito.any(URL.class)))
+                    .thenCallRealMethod();
+            Optional<String> vaadinVersion = Platform.getVaadinVersion();
+            Assert.assertTrue(vaadinVersion.isPresent());
+            Assert.assertEquals(version, vaadinVersion.get());
+        }
+    }
+
+}
```
mcollovati, line 98: newline missing 

```suggestion
}

```

mcollovati (changes requested):
Just a newline missing, other than that LGTM

flow-server/src/main/java/com/vaadin/flow/server/Platform.java
```diff
@@ -38,8 +43,34 @@ public class Platform implements Serializable {
      * @return the platform version or {@link Optional#empty()} if unavailable.
      */
     public static Optional<String> getVaadinVersion() {
-        try (InputStream vaadinVersionsStream = Platform.class.getClassLoader()
-                .getResourceAsStream(Constants.VAADIN_CORE_VERSIONS_JSON)) {
+        try {
+            List<URL> coreVersions = Collections.list(getClassloader()
+                    .getResources(Constants.VAADIN_CORE_VERSIONS_JSON));
+            if (coreVersions.size() == 1) {
+                return loadVersionForUrl(coreVersions.get(0));
```
Artur-, line 50: Can this be Hilla? Should the check below be done always?

flow-server/src/main/java/com/vaadin/flow/server/Platform.java
```diff
@@ -38,8 +43,34 @@ public class Platform implements Serializable {
      * @return the platform version or {@link Optional#empty()} if unavailable.
      */
     public static Optional<String> getVaadinVersion() {
-        try (InputStream vaadinVersionsStream = Platform.class.getClassLoader()
-                .getResourceAsStream(Constants.VAADIN_CORE_VERSIONS_JSON)) {
+        try {
+            List<URL> coreVersions = Collections.list(getClassloader()
+                    .getResources(Constants.VAADIN_CORE_VERSIONS_JSON));
```
Artur-, line 48: Does Hilla actually need to have a json file with this same name?

caalador: Probably should not if it is hilla specific.

flow-server/src/main/java/com/vaadin/flow/server/Platform.java
caalador: Yes. it makes it much simpler...

mcollovati (approved):

mcollovati:
As discussed internally, we will proceed reading the version from `pom.properties` (#17034), but in the meanwhile we will work on renaming version JSON files and then revisit the Platform methods implementations.

I'll temporarily close this PR, and we can reopen it or create a new one once the JSON files changes are applied on the platform.

PR #17045, opened 2023-06-16
fix: do not have server sources in client

### LINKED ISSUE #17020: flow-client-sources.jar contains source files from flow-server.jar
### Description of the bug

For example, the `VaadinServletRequest.java` is present both in `flow-client-sources.jar` and `flow-server-sources.jar`.

As a side-effect, it messes up with IDEA: https://youtrack.jetbrains.com/issue/IDEA-322302 .

### Expected behavior

The `VaadinServletRequest.java` (and all other files) should only be present in `flow-server-sources.jar` since that's where the corresponding class files are actually located.

In fact, the `flow-client-sources.jar` should be empty since `flow-client.jar` contains no Java `.class` files.

1. Download https://repo1.maven.org/maven2/com/vaadin/flow-client/24.1.0/flow-client-24.1.0-sources.jar
2. Verify that it contains the `com/vaadin/flow/server/*.java` files.

### Versions

- Vaadin / Flow version: 24.1.0

caalador:
add flow-server sources as
a sources dependency instead
of a src resource.

Closes #17020

### DIFF
--- a/flow-client/pom.xml
+++ b/flow-client/pom.xml
@@ -56,6 +56,19 @@
             <groupId>com.vaadin</groupId>
             <artifactId>flow-server</artifactId>
             <version>${project.version}</version>
+            <classifier>sources</classifier>
+            <scope>provided</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.vaadin.servletdetector</groupId>
+                    <artifactId>throw-if-servlet3</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-server</artifactId>
+            <version>${project.version}</version>
             <scope>provided</scope>
             <exclusions>
                 <exclusion>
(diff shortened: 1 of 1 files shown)


caalador:
Running with Eclipse
Version: 2023-06 (4.28.0)
Build id: 20230608-1333

I couldn't see any issues with having the sources as a dependency.

mcollovati:
Tested with IDEA, and it works fine, no need to exclude `flow-client` project anymore :+1: 

Also tested with Eclipse

```
Version: 2023-03 (4.27.0)
Build id: 20230309-1520
```

It took a while to import the project and I had to do run `Maven -> Update project` a couple of times before having the IDE not complaining about errors.
Based on my memories of working with Eclipse on maven multimodule projects, the slowness and the need for multiple "Update project"  is somehow expected.

mvysny:
This is awesome, thank you so much! This will improve the developer experience and lower the barrier of contributing to Flow :+1:

mcollovati (approved):

caalador:
Also tested with NetBeans IDE 17 without any issues (guess 18 will not have issues either)

PR #17054, opened 2023-06-19
fix: Strip ?inline from import before finding file

### LINKED ISSUE #16731: CSS import with ?inline query -> Invalid import
### Description of the bug
In one of our components we imported a Stylesheet like this:
import * as css from "./my-styles.css";

Upon starting the application Vite warns that this use of importing Stylesheets is deprecated and gives us an example to use instead:
Default and named imports from CSS files are deprecated. Use the ?inline query instead. For example: import * as css from "./my-styles.css?inline"

So as suggested we changed the line to:
import * as css from "./my-styles.css?inline"

Now after running the application again Vaadin shows an ERROR about a failed Import.

com.vaadin.flow.server.frontend.AbstractUpdateImports::resolve does not consider the "?inline" of the filename and fails to resolve a path and as such enters the InvalidPathException exception block.

The File loads fine since the styles are applied, but I dont know if it impacts something internally. 

Simply import a style in your component like this:
import * as css from "./my-styles.css?inline"

### Expected behavior
No Exception should occur and the path should be properly resolved
### Actual behavior
Path resolution fails and an Exception is thrown (handled by logging)
### Versions:
    - Vaadin / Flow version: 24.0.5
    - Java version: 17
    - OS version: Windows 11
    - IDE (if applicable): IntelliJ
    - Development or production mode: Development
    - Project from start.vaadin.com or activated in application: own application, nothing from start.vaadin.com

Artur-:
Fixes #16731

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
@@ -156,7 +156,7 @@ public final class StringUtil {
     /**
      * Strips the given suffix from the given string, if the strings end with
      * the suffix.
-     * 
+     *
      * @param string
      *            the string to scan
      * @param suffix
--- a/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/StringUtil.java
@@ -154,6 +154,25 @@ public final class StringUtil {
     }
 
     /**
+     * Strips the given suffix from the given string, if the strings end with
+     * the suffix.
+     * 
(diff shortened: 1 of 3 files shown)


Artur-:
Not sure how to really test this as an exception is only thrown on Windows

mshabarov:
@caalador could you please test this patch on Windows with project example from here https://github.com/vaadin/flow/issues/16731#issuecomment-1594946212 ? This patch is going to fix errors in `AbstractUpdateImports`, but it is not reproduced on Mac.

caalador (approved):
Tested on win11 and works as expected.

PR #17092, opened 2023-06-23
fix: add detach listener callback once for pending invocations

### LINKED ISSUE #17082: Infinite Lambda Chaining in PendingJavaScriptInvocation (Memory Leak)
### Description of the bug

Consider the minimal reproducible example below.

Once a button is attached a JS is invoked via `Button#initDisableOnClick`. Since the button `b` is not visible, that pending JS invocation will be retained internally. For those invisible components/nodes the pending JS is retained but now with every interaction with the view you can get an infinite lambda chain.

Each time you click on button `a`, during `UidlWriter#createUidl` the `UIInternals#dumpPendingJavaScriptInvocations` method is executed, which at the end is calling `UIInternals#registerDetachListenerForPendingInvocation` for each retained pending JS invocation. Finally, `invocation.then(callback, callback)` is being called which then calls

`this.successHandler = combineHandlers(this.successHandler, successHandler);`

This is called because
- Initially, `sentToBrowser` is false because this invocation has not been sent yet since its node was not visible yet.
- and the invocation is not `canceled`
- and the provided successHandler from UIInternals is not null

The provided successHandler parameter is always the same:

```java
SerializableConsumer callback = unused -> listener.onInvocationCompleted(invocation)
```

Now, during the first call the `PendingJavaScriptInvocation#successHandler` is null, so combineHandlers will just return the provided not null successHandler and store that in the field.

But, each time you click now on button `a` in the browser, next combineHandlers will have both not null, the field successHandler as well as the param successHandler. This leads to combining both to a new lambda and storing that in the field successHandler. Next time you click again and again, so you get an endless chaining of lambdas.

Screenshot 2023-06-22 at 1 09 47 PM

Remember that those chained lambda callbacks in the end are always the same ones provided by the calling method `UIInternals#registerDetachListenerForPendingInvocation`. The chaining would make sense if the callbacks were different ones, but in this case it's always the same callbacks each effectively doing a `onInvocationCompleted()` once called.

Assuming that an app has components that are initialized or implemented like in this minimal example, eventually this situation will make the app end up in an `OutOfMemoryError`, the more invisible components with pending JS there are the faster this will be reached.

### Expected behavior

There should be no memory leak in PendingJavaScriptInvocation via potential endless chaining of lambdas.

```java
@Route("")
public class View extends Div {

  public View() {
    Button a = new Button("test");
    Button b = new Button("invisible");
    b.setVisible(false);
    add(a, b);
  }
}
```

Each time you click on button `a`, the `PendingJavaScriptInvocation#successHandler` gets a new lambda chain longer than the previous one.

### Versions

- Vaadin / Flow version: 14.10.1
- Java version: 17.0.6
- OS version: macOS

mcollovati:
## Description

Currently, the detach listener callback is added to pending javascript invocation on every call to dumpPendingJavaScriptInvocations. This change will apply the callback only once, when the pending invocation is seen for the first time.

Fixes #17082

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -616,11 +616,11 @@ public class UIInternals implements Serializable {
                             node.addDetachListener(detachListener));
                     return detachListener;
                 });
-        listener.invocationList.add(invocation);
-
-        SerializableConsumer callback = unused -> listener
-                .onInvocationCompleted(invocation);
-        invocation.then(callback, callback);
+        if (listener.invocationList.add(invocation)) {
+            SerializableConsumer callback = unused -> listener
+                    .onInvocationCompleted(invocation);
+            invocation.then(callback, callback);
+        }
     }
 
     private class PendingJavaScriptInvocationDetachListener implements Command {
--- a/flow-server/src/test/java/com/vaadin/flow/component/internal/UIInternalsTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/internal/UIInternalsTest.java
@@ -25,6 +25,7 @@ import com.vaadin.flow.component.page.Push;
 import com.vaadin.flow.di.DefaultInstantiator;
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.function.DeploymentConfiguration;
+import com.vaadin.flow.function.SerializableConsumer;
 import com.vaadin.flow.internal.JsonCodec;
 import com.vaadin.flow.internal.StateNode;
 import com.vaadin.flow.internal.nodefeature.ElementChildrenList;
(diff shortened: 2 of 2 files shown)


caalador (approved):

mshabarov:
Does Flow 1.0 need this patch as well?

caalador:
No 1.0 doesn't have the same functionality. It was added in 2.x if I remember correctly

mshabarov:
Yes, `onInvocationCompleted` is not in the 1.0. Thanks!

NikolajOlejnik:
Thank you very much for the prompt fix. There is an important question. :-) Any idea when this fix will be available for Vaadin 23?

It is included in Flow 23.3.15. But Vaadin 23.3.15 uses Flow version 23.3.14 (without this fix).

We have serious problems with the load on our system and are unsure what to do - wait for the availability of this fix for Vaadin 23 or make some kind of temporary fix on our own. Thank you!

kkleinwi:
@NikolajOlejnik this appears to be fixed in 23.3.16

NikolajOlejnik:
Super, thx!

PR #17105, opened 2023-06-26
fix: add support for Java Time types for Element properties

### LINKED ISSUE #13317: Element.setPropertyBean does not work with LocalTime
### Description of the bug

If a bean has a property of type `LocalTime`, an exception gets thrown when trying to set it as a property for an element using `setPropertyBean` API.

### Expected behavior

There should not be an exception.

Workaround:

```java
// getElement().setPropertyBean("item", order);
		
// Workaround
try {
	ObjectMapper objectMapper = new ObjectMapper();
	objectMapper.registerModule(new JavaTimeModule());
	JsonObject item = Json.parse(objectMapper.writeValueAsString(order));
	getElement().setPropertyJson("item", item);
} catch (JsonProcessingException e) {
	e.printStackTrace();
}
```

### Versions

- Vaadin / Flow version: 23.0.1

mcollovati:
## Description

Allows POJOs with Java Time members to be serialized as Element JSON properties.

To avoid breaking changes, the Object mapper WRITE_DATES_AS_TIMESTAMPS feature is left as default, so some types such as LocalTime, LocaDate and LocalDateTime are serialized as arrays of numbers, as documented in Jackson JavaTimeModule.

Fixes #13317

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


Artur-:
Would somebody really want dates as numbers in separate fields? What is the backwards compatibility angle here if you were not able to send times or dates before?

mcollovati:
> What is the backwards compatibility angle here if you were not able to send times or dates before

`java.util.Date` and `java.sql.Date` works without adding the module, and the value is serialized as a number

czp13 (commented):
It can be a nice improvement to handle more cases (type) for the Element class.

Could we maybe add a covering test to `com/vaadin/flow/dom/ElementTest.java`. 
where we call the setPropertyBean method like this:
`getElement().setPropertyBean("item", <value-triggered-exception-before>);`

And check if here in the called method:

```java
    public Element setPropertyBean(String name, Object value) {
        if (value == null) {
            throw new IllegalArgumentException(USE_SET_PROPERTY_WITH_JSON_NULL);
        }
        return setPropertyJson(name, JsonUtils.beanToJson(value));
    }
```
This part works fine with the improvement `JsonUtils.beanToJson(value)` in the method.

I know we do similar checks, and tests here:
- `flow-server/src/test/java/com/vaadin/flow/internal/JsonUtilsTest.java`. 
- but I feel if someone changes the `setPropertyBean` method logic/body, then even tho the JsonUtils (and tests) work fine, and tests pass, we will not see if that part is broken for this case.
- also in the issue we could show that there is a test case for the way the error was thrown.

Not important to do, not a blocker, just my point of view :)

mshabarov (approved):
Looks good to me.
Some thoughts about other comments and propositions:
1. I agree to keep the default representation of dates, which is timestamp. I think it's more useful for developers to get the timestamp and then format it as they want, rather than work with formatted string representation. We could probably make this customisable, but Element API doesn't seem a right place to take into account such a deep detail. Better to choose a better default from the beginning.
2. Would be better to make a test for Element.setPropertyBean, this would be just more transparent.
3. Package json updates can be extracted into a separate PR.

mcollovati:
> 2. Would be better to make a test for Element.setPropertyBean, this would be just more transparent.

I'll add the test

> 3. Package json updates can be extracted into a separate PR.

They have been committed accidentally, I'll remove the changes.

mshabarov (approved):

PR #17128, opened 2023-06-28
fix: copy package-lock from bundle in hotdeploy

caalador:
Copy the package-lock.json from
the dev bundle in frontend
hotdeploy mode if no package-lock
is available.


mshabarov (changes requested):
Looks good, but needs a unit test.

caalador:
added

mshabarov (approved):

PR #17145, opened 2023-06-29
test: Add web push it test

### LINKED ISSUE #17131: End-to-end test for Web Push Notifications
We need a TestBench end-to-end test for Web Push Notifications introduced in 24.2.0.alpha1.
The test should include:
1. [Enable](https://stackoverflow.com/a/39997484) web push notifications in browser
2. Subscribe browser to notifications with `WebPush`
3. Send message with `WebPush` and check that the notification reached the browser, some insight are given [here](https://sahajamit.medium.com/automated-testing-of-web-push-notifications-ffc8a2d85e3c)
4. Unsubscribe in the clean up method.

The test should be included into nightly builds to have less load on a push server, i.e. no need to include it into PR validation and snapshot validation.

caalador:
Closes #17131

### DIFF
--- a/flow-tests/test-webpush/pom.xml
+++ b/flow-tests/test-webpush/pom.xml
@@ -1,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.vaadin</groupId>
+        <artifactId>flow-tests</artifactId>
+        <version>24.2-SNAPSHOT</version>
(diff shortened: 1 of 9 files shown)


flow-tests/test-webpush/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>test-dev-bundle</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-dev-server</artifactId>
+            <version>${project.version}</version>
```
mshabarov, line 35: Dev dependencies are not needed probably, because you run in prod mode with `build-frontend`

mshabarov (changes requested):

flow-tests/test-webpush/pom.xml
caalador: Removed. At first it was in dev mode, but that brings in extra things that we are not needing/testing.

mshabarov (approved):

Issue #17138, opened 2023-06-29
Trigger a popstate event on forward


ugur-vaadin:
### Describe your motivation

We are using `popstate` events to update the state of the new Side Nav component. Currently, when forwarding in `beforeEnter`, the update does not take the new page into account. The reason for this seems to be that there is no `popstate` event fired after forwarding. The `replaceState(...)` is sent to the client, but does not trigger the event.

### Describe the solution you'd like

It would be beneficial to trigger a `popstate` event on forwarding , or at least provide an option to do so.

### Describe alternatives you've considered

We considered wrapping the update in a `setTimeout` in the web component. This seems to work for most cases. However it is a workaround and is not guaranteed to work in any situation. Also, it does not seem to be a proper solution for this issue.

sissbruecker:
The same problem exists with `history.pushState`, for example when using `UI.navigate`.

PR #17146, opened 2023-06-29
feat: dispatch popstate event on server-side navigation

sissbruecker:
## Description

Dispatch a `popstate` event when using server-side navigation, which allows components like `vaadin-side-nav` to react to location changes. The event includes the state `{state: 'vaadin-router-ignore'}`, so that Vaadin router ignores the event.

This is similar to what Vaadin router does to notify about location changes:
https://github.com/vaadin/router/blob/f5f31e00acd542fe714ea472172d17a03ba04794/src/router.js#L733-L742

Closes https://github.com/vaadin/flow/issues/17138

## Type of change

- Feature

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/History.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/History.java
@@ -187,8 +187,8 @@ public class History implements Serializable {
         // Second parameter is title which is currently ignored according to
         // https://developer.mozilla.org/en-US/docs/Web/API/History_API
         ui.getPage().executeJs(
-                "setTimeout(() => window.history.pushState($0, '', $1))", state,
-                pathWithQueryParameters);
+                "setTimeout(() => { window.history.pushState($0, '', $1); window.dispatchEvent(new PopStateEvent('popstate', {state: 'vaadin-router-ignore'})); })",
+                state, pathWithQueryParameters);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/component/page/History.java
sissbruecker, line 1: This change might have side-effects since every `UI` and `Router` install a history change listener:
https://github.com/vaadin/flow/blob/ee0e9c9723e7e519a1e2242c535692aab05cf020/flow-server/src/main/java/com/vaadin/flow/component/UI.java#L1703-L1705
https://github.com/vaadin/flow/blob/ee0e9c9723e7e519a1e2242c535692aab05cf020/flow-server/src/main/java/com/vaadin/flow/router/Router.java#L79-L80

I could not observe any issues from testing, but I don't have a deep understanding of the routing architecture.

Alternatively we could consider dispatching a custom event, and make `vaadin-side-nav` configurable so that it also listens to custom events for location changes. The Flow component would then configure that custom event on the web component.

mcollovati: My understanding is that with the `PopStateEvent` with `vaadin-router-ignore` state is not propagated to the server side, so the change seems safe to me

mcollovati (approved):

Issue #15342, opened 2022-11-30
Gradle task ":vaadinPrepareFrontend" does not cache results


valentin-ricard:
### Describe the bug

The ":vaadinPrepareFrontend" does not cache if the source and dependencies did not change, causing increasingly long compile times on incremental builds on another project.

This shows when running the build with the `--info` flag:
```log
Task ':projectName:vaadinPrepareFrontend' is not up-to-date because:
  Task has not declared any outputs despite executing actions.
```

### Expected-behavior

When running successive builds without changing a file, the output of ":vaadinPrepareFrontend" is identical, so there is no need to execute it twice.

### Reproduction

 - Have two Vaadin applications in the same multimodule Gradle project
 - Execute two successive build commands on the project root (the default when using IntelliJ)
 - See that the ":vaadinPrepareFrontend" command is not "UP-TO-DATE" even if the files did not change

### System Info

Windows 10, Java 17, Vaadin 23.2.0
Gradle:
- Version: 7.4
- Vaadin plugin version: 23.2.0
- Build cache: Disabled
- Config cache: Disabled
- Parallel builds: Disabled

FritjofH:
This would be great to get done. It would be an improvement for sure, and with all the improvements done to caching in Gradle 8 which just released this makes sense.

tarekoraby:
No work has been started on this yet, but we are currently _considering_ adding support for this functionality soon (which would be released with Vaadin 24.2, currently scheduled for September 2023).

steve-todorov:
@tarekoraby many thanks for your reply! Are there any plans for implementing cache in `vaadinBuildFrontend` as well or only the `vaadinPrepareFrontedn` task?

tarekoraby:
I see adding caching support to `vaadinBuildFrontend` as less of an impactful enhancement since it is only used for creating production builds, which are relatively infrequently created and are usually not time-critical. Conversely, `vaadinPrepareFrontend` affects both development and production builds. So, we will prioritize the latter for now. (of course, if there is a way to address both tasks in one go during implementation, we should consider it). 

cc @mshabarov

benstpierre:
This would be an extremely valuable feature for us as we migrate our app to Gradle.  I might ping expert chat about this as well.

mshabarov:
[...]

PR #17150, opened 2023-06-29
feat: Incremental Gradle builds for vaadinPrepareFrontendTask

mshabarov:
## Description

Defines the input and outputs for `vaadinPrepareFrontendTask` of Vaadin Gradle plugin making it possible to skip the task if inputs and outputs are not changed, giving `UP-TO-DATE` result, according to [incremental builds feature](https://docs.gradle.org/current/userguide/incremental_build.html#incremental_build)

Inputs are:
- path to Node.js executable
- path to npm/pnpm executable
- configuration parameters, defined in [VaadinFlowPluginExtension](https://github.com/vaadin/flow/blob/main/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt)

Outputs are:
- `package.json`, `vite.config.js` and other files generated by Vaadin in the root of the project and in the frontend folder
- frontend folder, including generated folder

according to a prepareFrontend definition here https://github.com/vaadin/flow/blob/04a69001101aec44cb43f190ca91b4b4d1491dcf/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java#L148

How to test:
- in Flow:
`git checkout cache-prepare-frontend`
`mvn clean install -DskipTests`
`cd flow-plugins/flow-gradle-plugin`
`./gradlew clean build publishToMavenLocal -x functionalTest`
- in base-starter-gradle:
add the following to `build.gradle`
```
buildscript {
    repositories {
        mavenLocal()
        ...
    }
}

repositories {
    mavenLocal()
    ...
}
```
add the following to `gradle.properties`
```
vaadinVersion=24.2-SNAPSHOT
```
run `./gradlew clean vaadinPrepareFrontend --info`

*Open questions*:
- currently there is no input or output corresponding to changes in classpath (e.g. when a JAR dependency is added, e.g. add-on or reused theme), so that the `TaskCopyFrontendFiles` might be skipped, because the whole prepare-frontend is skipped as it doesn't see changes in classpath. Not sure if it's an issue, because in dev mode this is triggered by `DevModeInitializer` and in prod mode with the `build-frontend`.
- probably it's better to remove `@CacheableTask` for V24.2 (replace it with `@DisableCachingByDefault`) and let developers enable caching manually. The way how to disable caching for a task is described [here](https://docs.gradle.org/7.6.1/userguide/more_about_tasks.html#sec:disable-state-tracking), but I couldn't make it work neither with `tasks.register` nor with `tasks.findByName` so far -> needs more testing and we have to suggest developers how to disable caching in case if Gradle false thinks that its up-to-date when it's not.
- no tests so far

Fixes https://github.com/vaadin/flow/issues/15342

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPrepareFrontendTask.kt
```diff
@@ -18,20 +18,45 @@ package com.vaadin.gradle
...
 public open class VaadinPrepareFrontendTask : DefaultTask() {
 
+    private val extension: VaadinFlowPluginExtension
+
+    /**
+     * Defines an object containing all the inputs of this task.
+     */
+    @Nested
+    public open fun getTaskInputProperties(): PrepareFrontendInputProperties? {
+        return PrepareFrontendInputProperties(project)
```
mcollovati, line 44: The `getTaskInputProperties()` method is called multiple times. We can perhaps introduce a `PrepareFrontendInputProperties` field and instantiate it in the `init` block.

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPrepareFrontendTask.kt
```diff
@@ -18,20 +18,45 @@ package com.vaadin.gradle
...
+    }
+
+    /**
+     * Defines an object containing all the outputs of this task.
+     */
+    @Nested
+    public open fun getTaskOutputProperties(): PrepareFrontendOutputProperties? {
+        return PrepareFrontendOutputProperties(project)
```
mcollovati, line 52: The `getTaskOutputProperties()` method is called multiple times. We can perhaps introduce a `PrepareFrontendOutputProperties` field and instantiate it in the `init` block.

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/PrepareFrontendOutputProperties.kt
```diff
@@ -0,0 +1,86 @@
...
+    @OutputDirectory
+    public fun getFrontendDirectory(): File {
+        return extension.frontendDirectory
+    }
+
+    @OutputDirectory
+    public fun getGeneratedTsFolder(): File {
+        return extension.generatedTsFolder
+    }
+}
```
mcollovati, line 86: newline missing
```suggestion
}

```

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/PrepareFrontendInputProperties.kt
```diff
@@ -0,0 +1,259 @@
...
+            FrontendUtils.getVaadinHomeDirectory()
+                .absolutePath
+        }
+        settings.nodeDownloadRoot = URI(extension.nodeDownloadRoot)
+        settings.isForceAlternativeNode = extension.requireHomeNodeExec
+        settings.isUseGlobalPnpm = extension.useGlobalPnpm
+        settings.isAutoUpdate = extension.nodeAutoUpdate
+        return FrontendTools(settings)
+    }
+}
```
mcollovati, line 259: newline missing 
```suggestion
}

```

mcollovati (changes requested):

mcollovati:
> currently there is no input or output corresponding to changes in classpath (e.g. when a JAR dependency is added, e.g. add-on or reused theme), so that the TaskCopyFrontendFiles might be skipped, because the whole prepare-frontend is skipped as it doesn't see changes in classpath. Not sure if it's an issue, because in dev mode this is triggered by DevModeInitializer and in prod mode with the build-frontend.

The `GradlePluginAdapter` has a `getJarFiles()` method that could probably be used as `@Input` to track dependencies changes. But, as commented earlier, it may not be worth it, since the `TaskCopyFrontendFiles` is executed anyway. 

> probably it's better to remove @CacheableTask for V24.2 (replace it with @DisableCachingByDefault) and let developers enable caching manually. The way how to disable caching for a task is described [here](https://docs.gradle.org/7.6.1/userguide/more_about_tasks.html#sec:disable-state-tracking), but I couldn't make it work neither with tasks.register nor with tasks.findByName so far -> needs more testing and we have to suggest developers how to disable caching in case if Gradle false thinks that its up-to-date when it's not.

It seems to me that the cache annotations are not influencing the `up-to-date` state.
An alternative may be to use `doNotTrackState()` in the plugin `init` block, but make the execution conditional based on a new plugin property, e.g. `disableVaadinPrepareFrontendTaskTrackState` with default `true`.

```
init {
        if (extension.disableVaadinPrepareFrontendTaskTrackState) {
            doNotTrackState("State tracking for vaadinPrepareFronted task is disabled (default). Use the 'disableVaadinPrepareFrontendTaskTrackState' plugin property to enable the feature");
        }
}
```

This way, the new feature will not be active by default, but the developer can opt-in by setting the property in the `vaadin` section of the gradle project file.

```
vaadin {
    pnpmEnable = false
    disableVaadinPrepareFrontendTaskTrackState = false
}
```

In future releases, we can change the default value to always activate the feature active and then deprecate and remove the setting (or leave it there, it can be useful).

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/PrepareFrontendInputProperties.kt
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    @Input
+    @Optional
+    public fun getResourceOutputDirectory(): String? {
+        val resourceOutputDirectory = extension.resourceOutputDirectory
+        if (!resourceOutputDirectory.exists()) {
+            return null
+        }
+        return resourceOutputDirectory.absolutePath
```
mcollovati, line 186: The prepareFrontend task generates the `flow-build-info.json` file in this directory.
This causes the second run of the task to be considered not up-to-date.
I wonder if this directory should instead be considered a task output

mcollovati:
Also `vaadin-featureflags.properties` should be considered, since `frontend/generated/vaadin-featureflags.js` is generated and updated by the prepare frontend task, based on the contents of the file

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/PrepareFrontendInputProperties.kt
mshabarov: Using a local `TestProject()` object within the test method should work I suppose.

mcollovati: My understanding was that the temporary directory is consistent in the same test, unless you create a new instance of `TestProject`

mshabarov: Right, should be the same within the same test, weird.

mshabarov: I think you were right about moving resource dir property to an output. If we do so, Gradle would still be able to spot the re-configuring the path change and trigger a new build.

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPrepareFrontendTask.kt
mshabarov: done

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPrepareFrontendTask.kt
mshabarov: done

mcollovati (approved):

PR #17152, opened 2023-06-30
fix: Not check the license for default production build

mshabarov:
Workaround for false license checking result leading to require license for default bundle even if no commertial components are used in the project.


mshabarov:
Closed in favour of https://github.com/vaadin/flow/pull/17153

PR #17153, opened 2023-06-30
fix: Validate license only for used components

Artur-:
The package list always includes all packages (from @NpmPackages). We need to check which js modules are actually used to know which components are used


caalador (commented):
`BuildFrontendUtilTest.detectsCommercialComponents` needs mocking for `scanner.getModules` so the test returns correct data, but else seems ok.

caalador (approved):

mshabarov (approved):
Tested with Vaadin 24.1.1-based starter with this patch.
I don't see any license checkings and here what I get as a `usedPackages` for this project:
```
@vaadin/side-nav
@vaadin/horizontal-layout
@vaadin/polymer-legacy-adapter
@vaadin/notification
@vaadin/button
@vaadin/scroller
@vaadin/app-layout
@vaadin/text-field
@vaadin/vaadin-lumo-styles
@vaadin/tooltip
@vaadin/vertical-layout
```
So no commercial components.
Saw license checking before this patch.
LGTM.

Issue #17159, opened 2023-07-03
Theme and jar file watching from other Maven modules does not work in Wildfly


Artur-:
### Description of the bug

If you deploy a multi module project to Wildfly, then editing the frontend or theme files in another module than the main module will not trigger a live reload or hot deploy.

The problem is that the file watcher is not triggered on file changes. This seems to be caused by https://github.com/gmethvin/directory-watcher/issues/94

### Expected behavior

It should work like in all other servers

### Minimal reproducible example

Need to create one

### Versions

- Vaadin / Flow version: 24.1.1

Artur-:
Workaround is to add `com.sun.nio.file` as a dependency, e.g.
```
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-war-plugin</artifactId>
                <configuration>
                    <archive>
                        <manifestEntries>
                            <Dependencies>com.sun.nio.file</Dependencies>
                        </manifestEntries>
                    </archive>
                </configuration>
            </plugin>
```

AB-xdev:
Can confirm the problem for Vaadin 24, Wildfly 30 and Java 17 (Temurin)

mcollovati:
It appears that adding the manifest entry to `vaadin-dev-server` works as well.

PR #17161, opened 2023-07-03
fix: Report exception if filewatcher fails to start

Artur-:
Related to #17159

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/FileWatcher.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/FileWatcher.java
@@ -19,8 +19,12 @@ import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 
 import io.methvin.watcher.DirectoryWatcher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.function.SerializableConsumer;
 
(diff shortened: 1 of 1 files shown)


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/FileWatcher.java
```diff
@@ -61,7 +65,14 @@ public FileWatcher(SerializableConsumer<File> onChangeConsumer,
...
     public void start() {
-        watcher.watchAsync();
+        CompletableFuture<Void> future = watcher.watchAsync();
+        if (future.isCompletedExceptionally()) {
+            try {
+                future.get();
+            } catch (Exception e) {
+                getLogger().error("Error starting file watcher", e);
+            }
+        }
```
mcollovati, line 75: Can this be simplified using `CompletionStage.exceptionally()`?

```java
watcher.exceptionally(ex -> {
            getLogger().error("Error starting file watcher", ex);
            return null;
        });
```

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/FileWatcher.java
```diff
@@ -74,4 +85,8 @@ public void stop() throws IOException {
         watcher.close();
     }
 
+    private Logger getLogger() {
+        return LoggerFactory.getLogger(getClass());
+    }
+
```
mcollovati, line 91: Other classes in the same package define `getLogger` as a static method. Shall we adopt the same pattern here?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/FileWatcher.java
Artur-: Yes, much better

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/FileWatcher.java
Artur-: Sure

mcollovati (approved):

PR #17171, opened 2023-07-04
WIP: minor improvements and fixed test

mcollovati:
Minor improvements on top of #17150


mshabarov:
@mcollovati after thinking a while, I'd like to propose to make `alwaysExecutePrepareFrontend = false` by default, this seems to have benefits over `true`:
- Most of the projects would benefit from using incremental builds out-of-the-box immediately (even not being aware that this feature exists) and in the future we don't need to swap the default value - less work
- Developers who spot any issues (I expect very minor amount of cases), caused by not covering some config details, can change this flag to true, while we fix it in the maintenance release.

What do you think?

mcollovati:
The reason for having `alwaysExecutePrepareFrontend = false` as default was the following note in the original PR description

> probably it's better to remove @CacheableTask for V24.2 (replace it with @DisableCachingByDefault) and let developers enable caching manually.

But I agree to make the feature enabled by default, as it will be a good development improvement.
Furthermore, as you mentioned, if there are any issues, the feature can easily be disabled.

mshabarov:
Thanks, I've pushed the changes for it.

mshabarov (approved):

PR #17175, opened 2023-07-04
fix: prevent writing service worker on filesystem in dev mode (#17155) (CP: 23.3)

mcollovati:
When running in dev mode, the service worker should be served from memory and not written to the filesystem, to avoid potential endless restarts when Spring dev tools are in use.

This change fixes a regression introduced when service worker compression was implemented, that causes sw.js to be written to the filesystem also in dev mode.


czp13 (approved):
👍

PR #17209, opened 2023-07-10
test: fix webpush integration test

mcollovati:
Webpush integration test was working fine locally, also using selenium hub, but failed on CI env.
This change fixes the test by waiting a bit longer for the subscription to complete and handling errors received a string instead of JSON objects.


mcollovati:
Builds logs can be checked on Bender (`Flow Web Push test` job)

czp13 (approved):
👍 The changes are looking good to me.

PR #17216, opened 2023-07-13
feat: fail fast when plugin is used with an unsupported version of Gradle

### LINKED ISSUE #16149: gradle-plugin should fail if used gradle is not 7.5+
### Describe your motivation

I would like that vaadin gradle plugin checks if I'm using correct version of Gradle.

Vaadin [requires 7.5](https://github.com/vaadin/platform/issues/3720) version as spring does.

Spring Boot plugin already has a [solution](https://github.com/spring-projects/spring-boot/blob/0bba1903ea1b6e58d4498e60de7203719c8e9bfc/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/plugin/SpringBootPlugin.java#L98-L111) checking Gradle version.

When using incorrect version spring-boot throws an exception that is shown in user's console, we could use the same approach

```
./gradlew clean

FAILURE: Build failed with an exception.

* Where:
Build file '.../build.gradle' line: 9

* What went wrong:
An exception occurred applying plugin request [id: 'org.springframework.boot', version: '3.0.4']
> Failed to apply plugin 'org.springframework.boot'.
   > Spring Boot plugin requires Gradle 7.x (7.4 or later). The current version is Gradle 7.3

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

mcollovati:
## Description

Checks the gradle version when applying the plugin and fails fast if the current gradle version is unsupported.

Closes #16149

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/VaadinSmokeTest.kt
+++ b/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/VaadinSmokeTest.kt
@@ -344,7 +344,7 @@ class VaadinSmokeTest : AbstractGradleTest() {
 
         // Cannot test versions older than 7.6 because of Java version
         // incompatibilities with dependencies thar makes the build fail before
-        //the plugin is applied
+        // the plugin is applied
         for (unsupportedVersion in arrayOf("7.6")) {
             setupProjectForGradleVersion(unsupportedVersion)
             val result = testProject.buildAndFail("vaadinClean")
--- a/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/VaadinSmokeTest.kt
+++ b/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/VaadinSmokeTest.kt
@@ -15,16 +15,16 @@
  */
 package com.vaadin.gradle
 
+import java.io.File
+import kotlin.test.assertContains
+import kotlin.test.expect
 import com.vaadin.flow.server.InitParameters
 import elemental.json.JsonObject
 import elemental.json.impl.JsonUtil
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.TaskOutcome
 import org.junit.Before
-import org.junit.Ignore
 import org.junit.Test
-import java.io.File
-import kotlin.test.expect
 
 /**
  * The most basic tests. If these fail, the plugin is completely broken and all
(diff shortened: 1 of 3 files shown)


czp13 (dismissed):

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPlugin.kt
```diff
@@ -61,4 +69,16 @@ public class VaadinPlugin : Plugin<Project> {
             }
         }
     }
+
+    private fun verifyGradleVersion() {
+        val currentVersion = GradleVersion.current();
+        val supportedVersion =
+            GradleVersion.version(GRADLE_MINIMUM_SUPPORTED_VERSION)
+        if (currentVersion < supportedVersion) {
```
czp13, line 77: Minor comment.

Maybe? as `7.6.1` shall be supported as well if I understand (from the naming)
```suggestion
        if (currentVersion <= supportedVersion) {
```

flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/VaadinSmokeTest.kt
```diff
@@ -325,4 +325,33 @@ class VaadinSmokeTest : AbstractGradleTest() {
         expect(true, cssFile.toString()) { cssFile.exists() }
 
     }
+
+    @Test
```
czp13, line 329: Maybe adding two more cases just to make sure:
- one for the 7.6.1 (equals case)
- one for 7.6.2 (greater case)
- and we have the 7.6 (less case right now)

But I am not sure as this test is also ensuring the feature/development is working! 

And good job on this!

mcollovati: 7.6.1 is tested by the first build execution in the test because it is the version used by default by TestProject.

Testing 7.6.2 or even a higher version (maybe 8.x if compatible?) seems a good idea

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPlugin.kt
czp13: Ah, you are right, it was a mistake from me,  I miss see this logic :) that is why it is good that we have covering tests :D. 
We can close this, the condition works well like this! 👍 
Thanks!

flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/VaadinSmokeTest.kt
czp13: Nice, okey, indeed I missed that one too :(! Then 7.6.1, 7.6 were tested. 
8.x seems a good idea to me as well.

czp13 (approved):

PR #17228, opened 2023-07-17
fix: unregister shortcut beforeClientResponse callback on component detach

### LINKED ISSUE #17201: Serialization issue with ShortcutRegistration
### Description of the bug

We are getting the error on a cluster node which is reading a serialized session which contains a UI with a ShortcutRegistration 

similar to this one:
`Shortcuts.addShortcutListener(this, event -> testField.focus(), Key.KEY_F, KeyModifier.META);`
	
`java.lang.ClassCastException: cannot assign instance of java.lang.invoke.SerializedLambda to field com.vaadin.flow.component.ShortcutRegistration.beforeClientResponseConsumer of type com.vaadin.flow.function.SerializableConsumer in instance of com.vaadin.flow.component.ShortcutRegistration`

### Expected behavior

No serialization exception

Unfortunatelly we are not able to reproduce the issue with a small sample

### Versions

- Vaadin / Flow version: 23.3.15
- Java version:  Temurin-17.0.7+7 (build 17.0.7+7)
- OS version: docker / tomcat:9-jdk17-temurin

mcollovati:
## Description

If a component that is a shortcut lifecycle owner is attached and detached in the same roundtrip its beforeClientResponse callback is not unregistered causing also potential serialization issues.

This change unregisters the callback on component detach, and it turns the callback into an anonymous class to prevent deserialization issues due to self references.

Fixes #17201

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


czp13 (approved):
Good fix. 

I tried to fix this in other ways like:
- https://stackoverflow.com/questions/25391656/serialization-of-a-lambda-after-its-creation

But could not work to cast the lambda propery and also make it serializable, so this is the way to use an anonymous inner class and not lambda which is not serializable out of the box...

Tests are passing locally as well, seems good! 👍

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/SerializeShortcutView.java
```diff
@@ -0,0 +1,67 @@
...
+public class SerializeShortcutView extends AbstractDivView {
+    public SerializeShortcutView() {
+        Div label = new Div();
+        label.setId("message");
+
+        Input input = new Input();
+        Shortcuts.addShortcutListener(input, event -> input.focus(), Key.KEY_F,
+                KeyModifier.META);
+
+        NativeButton button1 = createButton("Add Remove Shortcut Owner",
```
taefi, line 40: Just a trivial label suggestion:
```suggestion
        NativeButton button1 = createButton("Add Shortcut Owner",
```

czp13 (approved):

PR #14647, opened 2022-09-26
fix: Cancel previous flush request if new one is forced

mshabarov:
## Description

Cancels flush request if it presents in the queue and a new request is forcibly registered.
No unit tests added, because this case is quite hard to simulate.

Regression IT test https://github.com/vaadin/flow-components/pull/3804.

Fixes #14435

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

Issue #14435, opened 2022-08-27
DataCommunicator gets stuck after browser refresh, eventually causing OutOfMemoryError


p-engl:
### Description of the bug

After upgrading to Vaadin 23.1.6, we encountered situations where the DataCommunicator of a Grid gets stuck after a browser refresh. It seems that the DataCommunicator gets stuck in some kind of flush-loop, which ultimately leads to an OutOfMemoryError. I suppose the problem was introduced in flow-data 23.1.4, this change might be relevant: https://github.com/vaadin/flow/commit/c55bdcc6dd9dac75be42291b5752d1d86d5a9997
This bug does not appear after downgrading to Vaadin 23.1.4

Note: In our application, we often use views that have a Grid on the left side and a TabSheet on the right side which can show data about the object that is selected in the left Grid. Most of the time, the data on the right side will be shown with a Grid. This type of view is affected by the problem. In general, it seems that the problem can appear in case that a component is shown, which was already used and removed before the browser refresh.

I created this issue in the flow repository, since the problem is at its core caused by the DataCommunicator.

### Expected behavior

After a browser refresh, it should be possible to show a component that contains a Grid and was shown before the browser refresh  without causing this bevahiour (this was possible before 23.1.6). The DataCommunicator should not get stuck in an infinite loop. 

### Minimal reproducible example

Below is an example that can reproduce the bug.

Steps to reproduce the bug:
Go to /test
Select "1" in left table
Click "Show Table 1"
Click "Show Table 2"
Press F5
Click "Show Table 1"

```java
import java.util.ArrayList;
import java.util.Arrays;

import com.vaadin.flow.component.Component;
import com.vaadin.flow.component.button.Button;
import com.vaadin.flow.component.checkbox.Checkbox;
import com.vaadin.flow.component.grid.Grid;
import com.vaadin.flow.component.html.Div;
import com.vaadin.flow.component.orderedlayout.HorizontalLayout;
import com.vaadin.flow.router.AfterNavigationEvent;
import com.vaadin.flow.router.AfterNavigationObserver;
import com.vaadin.flow.router.PreserveOnRefresh;
import com.vaadin.flow.router.Route;

@PreserveOnRefresh
@Route("test")
public class TestView extends Div implements AfterNavigationObserver{

	private HorizontalLayout hLayout;
		
	private Div wrapperRight;
	
	private Grid<String> tableLeft;
	
	private Grid<String> tableRightOne;
	
	private Grid<String> tableRightTwo;
	
[...]

PR #17231, opened 2023-07-18
fix: do not register requests while flush in progress

ugur-vaadin:
## Description

Similar to an [issue](https://github.com/vaadin/flow/issues/14435) that was [fixed](https://github.com/vaadin/flow/pull/14647) before, there is a new [issue](https://github.com/vaadin/flow-components/issues/4674) regarding endless loops triggered under certain conditions. The issue was originally reported for  `ComboBox`. However, the behaviour can be reproduced with a `Grid` as well, since the underlying issue is with the `DataCommunicator`. Even if the example or the test is not considered best practice, the endless loop should still be avoided.

This PR replaces the flush request canceling approach using a "flush in progress" flag. While a registered flush is triggered, until it ends, no other flush requests will be registered. With this change, any edge cases regarding registering multiple flush requests during the same round-trip should be dealt properly.

Fixes related Flow component [issue](https://github.com/vaadin/flow-components/issues/4674).

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
@@ -1129,6 +1129,9 @@ public class DataCommunicator<T> implements Serializable {
         if (forced) {
             return true;
         }
+        // New requests that are not forced are not registered while a flush
+        // is in progress. This prevents infinite loop in cases including
+        // @PreserveOnRefresh.
         return !flushInProgress && (flushRequest == null || !flushRequest.canExecute(stateNode));
     }
 
@@ -1146,6 +1149,8 @@ public class DataCommunicator<T> implements Serializable {
     }
 
     private boolean shouldRequestFlushUpdatedData() {
+        // New requests are not registered while a flush is in progress. This
+        // prevents infinite loop in cases including @PreserveOnRefresh.
         return !flushUpdatedDataInProgress && (flushUpdatedDataRequest == null ||
                 !flushUpdatedDataRequest.canExecute(stateNode));
     }
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
@@ -1132,27 +1132,27 @@ public class DataCommunicator<T> implements Serializable {
         // New requests that are not forced are not registered while a flush
         // is in progress. This prevents infinite loop in cases including
         // @PreserveOnRefresh.
-        return !flushInProgress && (flushRequest == null || !flushRequest.canExecute(stateNode));
+        return !flushInProgress && (flushRequest == null
+                || !flushRequest.canExecute(stateNode));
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1105,44 +1107,52 @@ private void requestFlush() {
...
-                        flushUpdatedDataRequest = null;
-                    });
+        if (!shouldRequestFlushUpdatedData()) {
+            return;
         }
+        flushUpdatedDataRequest = FlushRequest.register(stateNode, context -> {
+            flushUpdatedDataInProgress = true;
+            flushUpdatedData();
+            flushUpdatedDataRequest = null;
+            flushUpdatedDataInProgress = false;
```
mcollovati, line 1147: What do you think if we wrap the `flushUpdateData()` execution in a try/finally block, so we can be sure that the flag is updated even if an error occurs?

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1105,44 +1107,52 @@ private void requestFlush() {
...
             }
-            flushRequest = FlushRequest.register(stateNode, context -> {
-                if (!context.isClientSideInitialized()) {
-                    reset();
-                    arrayUpdater.initialize();
-                }
-                flush();
-                flushRequest = null;
-            });
+            flush();
```
mcollovati, line 1119: What do you think if we wrap the `flus()` execution and the `if` above, in a try/finally block, so we can be sure that the flag is updated even if an error occurs?

mcollovati (approved):
LGTM.
A couple of comments for potential, some minor, but not blocking.
It would be nice to add a test as well for the flush update request, but if it is not that easy, we can create a ticket and implement it in another PR.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
ugur-vaadin: That was the initial version, then I could not be sure if it would be overkill and removed it. But better be safe than sorry. Updating both blocks.

ugur-vaadin: Done.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
ugur-vaadin: Done.

ugur-vaadin:
> It would be nice to add a test as well for the flush update request

Sure, do you have specific cases to cover in mind?

mcollovati:
> Sure, do you have specific cases to cover in mind?

No idea ATM. Looking at the code, there should be no loop for updates.
We can add a test once we find a good case

mcollovati (approved):

Issue #17237, opened 2023-07-19
Disabling sync id checking does not work


TatuLund:
### Description of the bug

This bug fix does not take into account the case that lastSeenOnClient can be -1 when sync id checking has been disabled

https://github.com/vaadin/flow/pull/15764/files

This will result putting all messages back into cache repeatedly as alreadySeen returns always false.

This again leads to server continuously pushing.

See also trace log

```
2023-07-19 11:04:23.796 TRACE 22348 --- [nio-8080-exec-9] org.atmosphere.cpr.DefaultBroadcaster    : Checking cached message for 8fe751c1-6c2f-4f80-887b-33c424b979d2
2023-07-19 11:04:23.796 TRACE 22348 --- [nio-8080-exec-9] o.atmosphere.cache.UUIDBroadcasterCache  : Retrieved for AtmosphereResource 8fe751c1-6c2f-4f80-887b-33c424b979d2 cached messages 19
[...]

PR #17238, opened 2023-07-19
fix: Fix regression when sync id check is disabled

TatuLund:
Fixes: https://github.com/vaadin/flow/issues/17237

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/communication/LongPollingCacheFilterTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/communication/LongPollingCacheFilterTest.java
@@ -82,6 +82,19 @@ public class LongPollingCacheFilterTest {
     }
 
     @Test
+    public void filter_syncIdCheckDisabled_continueWithCurrentMessage() {
+        setTransport(AtmosphereResource.TRANSPORT.LONG_POLLING);
+        setSeenServerSyncIdHeader(-1);
+        BroadcastAction action = filter.filter("broadcasterId", resource,
+                originalMessage, message);
+        Assert.assertEquals(ACTION.CONTINUE, action.action());
+        Assert.assertSame(
+                "Message should not be altered by filter if server sync id header is missing",
+                message, action.message());
+        verifyMessageIsNotCached();
+    }
+
+    @Test
     public void filter_missingLastSeenServerSyncId_continueWithCurrentMessage() {
         setTransport(AtmosphereResource.TRANSPORT.LONG_POLLING);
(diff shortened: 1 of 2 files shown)


mcollovati (changes requested):
@TatuLund thank you very much for spotting the bug and providing the fix.
The change looks good to me, but I would like to add a test for this case

Could use please add a test to `LongPollingCacheFilterTest` class? 
Something like

```java
    @Test
    public void filter_syncIdCheckDisabled_continueWithCurrentMessage() {
        // a value of -1 means that syncId check is disabled
        setSeenServerSyncIdHeader(-1);
        setTransport(AtmosphereResource.TRANSPORT.LONG_POLLING);
        simulatePushConnection();
        BroadcastAction action = filter.filter("broadcasterId", resource,
                originalMessage, message);
        Assert.assertEquals(ACTION.CONTINUE, action.action());
        Assert.assertSame(
                "Message should not be altered by filter if syncId check is disabled",
                message, action.message());
        verifyMessageIsNotCached();
    }
```

TatuLund:
> Could use please add a test to LongPollingCacheFilterTest class?

I was already working on that

flow-server/src/test/java/com/vaadin/flow/server/communication/LongPollingCacheFilterTest.java
```diff
@@ -81,6 +81,19 @@ public void filter_notLongPollingTransport_continueWithCurrentMessage() {
...
 
+    @Test
+    public void filter_syncIdCheckDisabled_continueWithCurrentMessage() {
+        setTransport(AtmosphereResource.TRANSPORT.LONG_POLLING);
+        setSeenServerSyncIdHeader(-1);
+        BroadcastAction action = filter.filter("broadcasterId", resource,
+                originalMessage, message);
+        Assert.assertEquals(ACTION.CONTINUE, action.action());
+        Assert.assertSame(
+                "Message should not be altered by filter if server sync id header is missing",
```
mcollovati, line 92: A minor change on the assertion message
```suggestion
                "Message should not be altered by filter if syncId check is disabled",
```

mcollovati (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/communication/LongPollingCacheFilterTest.java
TatuLund: Fixed

mcollovati (approved):

Issue #14565, opened 2022-09-19
Allow modifying Query Parameters list


OlliTietavainenVaadin:
### Describe your motivation

Sometimes, you want to add or remove items from the query parameters before e.g. rerouting. This could happen e.g. if the query parameters contain information that shouldn't be displayed in the URL bar and should only be stored in the session. Currently this requires recreating the QueryParameters object, which is clunky at best:
```
QueryParameters queryParametersAfterLogin = new QueryParameters(
      parameters.entrySet().stream()
        .filter(e -> !e.getKey().equals("pwd"))
        .filter(e -> !e.getKey().equals("user"))
        .filter(e -> !e.getKey().equals("client"))
        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
```

### Describe the solution you'd like

`add(key)` and `remove(key)` methods provided in the QueryParameters object

Legioth:
There's some merit to `QueryParameters` being immutable. There could still be mutators but they would then return a new immutable object instead of modifying the existing ones.

Your usage example could then look like this:
```java
QueryParameters queryParametersAfterLogin = parameters.excluding("pwd", "user", "client");
```

Adding a single entry could be
```java
QueryParameters added = parameters.including("foo", "bar");
```

Adding multiple entries could be through accepting a `Map` (which plays nicely with the new-ish `Map::of` shorthands):
```java
QueryParameters addedMany = parameters.includingAll(Map.of("key1", "value1", "key2", "value2"));
```

PR #17259, opened 2023-07-20
feat: Allow modifying QueryParameters 

TatuLund:
Add methods proposed by Leif

Fixes https://github.com/vaadin/flow/issues/14565

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -251,11 +251,10 @@ public class QueryParameters implements Serializable {
      */
     public QueryParameters excluding(String... keys) {
         Set<String> excludedKeys = Set.of(keys);
-        Map<String, List<String>> newParameters = new HashMap<>(
-                parameters.entrySet().stream()
-                        .filter(entry -> !excludedKeys.contains(entry.getKey()))
-                        .collect(Collectors.toMap(Map.Entry::getKey,
-                                entry -> new ArrayList<>(entry.getValue()))));
+        Map<String, List<String>> newParameters = parameters.entrySet().stream()
+                .filter(entry -> !excludedKeys.contains(entry.getKey()))
+                .collect(Collectors.toMap(Map.Entry::getKey,
+                        entry -> new ArrayList<>(entry.getValue())));
         return new QueryParameters(newParameters);
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -26,6 +26,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -250,6 +251,9 @@ public class QueryParameters implements Serializable {
      * @return QueryParameters
      */
     public QueryParameters excluding(String... keys) {
+        if (keys == null || keys.length == 0) {
+            return this;
+        }
         Set<String> excludedKeys = Set.of(keys);
         Map<String, List<String>> newParameters = parameters.entrySet().stream()
                 .filter(entry -> !excludedKeys.contains(entry.getKey()))
@@ -268,6 +272,10 @@ public class QueryParameters implements Serializable {
      * @return QueryParameters.
      */
     public QueryParameters including(String key, String... values) {
+        if (key == null || key.isEmpty() || values == null
+                || values.length == 0) {
+            throw new IllegalArgumentException("Parameter missing");
+        }
         Map<String, List<String>> newParameters = new HashMap<>(parameters);
         List<String> newValues = List.of(values);
         newParameters.put(key, newValues);
@@ -282,6 +290,7 @@ public class QueryParameters implements Serializable {
      * @return QueryParameters
      */
     public QueryParameters includingAll(Map<String, List<String>> parameters) {
+        Objects.requireNonNull(parameters);
         Map<String, List<String>> newParameters = new HashMap<>(
                 this.parameters);
         newParameters.putAll(parameters);
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -244,7 +244,7 @@ public class QueryParameters implements Serializable {
 
     /**
      * Return new QueryParameters excluding given parameters by names.
-     * 
+     *
      * @param keys
      *            Names of the parameters to be excluded
      * @return QueryParameters
@@ -261,7 +261,7 @@ public class QueryParameters implements Serializable {
 
     /**
      * Return new QueryParameters including given parameters.
-     * 
+     *
      * @param key
      *            Parameter name as String
      * @param values
@@ -277,7 +277,7 @@ public class QueryParameters implements Serializable {
 
     /**
      * Return new QueryParameters including given parameters.
-     * 
+     *
      * @param parameters
      *            Map of new parameters to be included
      * @return QueryParameters
--- a/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
@@ -254,7 +254,7 @@ public class QueryParameters implements Serializable {
         Map<String, List<String>> newParameters = parameters.entrySet().stream()
                 .filter(entry -> !excludedKeys.contains(entry.getKey()))
                 .collect(Collectors.toMap(Map.Entry::getKey,
-                        entry -> new ArrayList<>(entry.getValue())));
+                        entry -> entry.getValue()));
         return new QueryParameters(newParameters);
     }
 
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -240,6 +242,52 @@ public String getQueryString() {
...
     }
 
+    /**
+     * Return new QueryParameters excluding given parameters by names.
+     *
+     * @param keys
+     *            Names of the parameters to be excluded
+     * @return QueryParameters
+     */
+    public QueryParameters excluding(String... keys) {
```
mcollovati, line 252: I wonder if it could be better to have at least one mandatory key

```suggestion
    public QueryParameters excluding(String key, String... otherKeys) {
```

Or short-circuiting to `new QueryParameters(parameters)` or `this` if the keys array is empty

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -240,6 +242,52 @@ public String getQueryString() {
...
+    /**
+     * Return new QueryParameters including given parameters.
+     *
+     * @param key
+     *            Parameter name as String
+     * @param values
+     *            Values for the parameter as Strings
+     * @return QueryParameters.
+     */
+    public QueryParameters including(String key, String... values) {
```
mcollovati, line 270: I have a concern about the method name: to me, `including` sounds like `give me a copy of the params that only includes given keys`.
But the method actually returns a new instance with the addition or replacement of a given key.

So I propose to rename it to something different like `with(key, values)` or `merging(kes, values)` 

Then we could maybe add an `including(keys...)` method that returns a copy with only the requested keys, the opposite of `excluding()`.

The same also applies for the `includingAll` method.

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
TatuLund: Done

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
TatuLund: I do not have strong opinions here. I simply used names proposed by Leif. I would expect similar concerns arising with other naming styles too.

TatuLund: Ok. Done.

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -240,6 +243,82 @@ public String getQueryString() {
...
+            throw new IllegalArgumentException("Parameter missing");
+        }
+        Map<String, List<String>> newParameters = new HashMap<>();
+        List<String> newValues = List.of(values);
+        newParameters.put(key, newValues);
+        return new QueryParameters(newParameters);
+    }
+
+    /**
+     * Return new QueryParameters including given parameters and the existing
```
mcollovati, line 286:
```suggestion
     * Return new QueryParameters adding given parameter to the existing
```

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -240,6 +243,82 @@ public String getQueryString() {
...
+     * ones.
+     *
```
mcollovati, line 288:
```suggestion
     *
     * If a parameter with the same name is already present, its values will be
     * replaced with the provided ones.
     *
```

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -240,6 +243,82 @@ public String getQueryString() {
...
+        Map<String, List<String>> newParameters = new HashMap<>(parameters);
+        List<String> newValues = List.of(values);
+        newParameters.put(key, newValues);
+        return new QueryParameters(newParameters);
+    }
+
+    /**
+     * Return new QueryParameters including given parameters and the existing
+     * ones.
+     *
```
mcollovati, line 309:
```suggestion
     *
     * Existing parameters will be replaced by the provided ones.
     *
```

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
mcollovati: I'm sorry, I think I have not explained clearly my idea of `including()`.

I was thinking it as a method that does the opposite of `excluding()`, so basically take the existing parameter map and filter it to retain only the provided keys.

So if the original parameters contain `{ "p1": X,  "p2": Y, "p3": Z }`, the `including("p1")` will return `{"p1": X}` and `including("p1", "p3")` will produce `{ "p1": X,  "p3": Z }`

flow-server/src/test/java/com/vaadin/flow/router/QueryParametersTest.java
```diff
@@ -235,6 +235,106 @@ public void shortHands() {
...
+        paramMap.put("two", Collections.singletonList("2"));
+        paramMap.put("three", Collections.singletonList("3"));
+
+        QueryParameters params = new QueryParameters(paramMap);
+        QueryParameters newParams = params.excluding();
+
+        Assert.assertEquals(params, newParams);
+    }
+
+    public void including() {
```
mcollovati, line 270: `@Test` annotation is missing
```suggestion
    @Test
    public void including() {
```

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
TatuLund: Done

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
TatuLund: Done

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
TatuLund: Done

flow-server/src/test/java/com/vaadin/flow/router/QueryParametersTest.java
TatuLund: Done

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
TatuLund: Done

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -240,6 +243,81 @@ public String getQueryString() {
...
+     */
+    public QueryParameters excluding(String... keys) {
+        if (keys == null || keys.length == 0) {
+            return this;
+        }
+        Set<String> excludedKeys = Set.of(keys);
+        Map<String, List<String>> newParameters = parameters.entrySet().stream()
+                .filter(entry -> !excludedKeys.contains(entry.getKey()))
+                .collect(Collectors.toMap(Map.Entry::getKey,
+                        entry -> entry.getValue()));
```
mcollovati, line 261: I wonder if this can be simplified by just creating a `new Map(parameters)` and the just removing given keys (e.g. `Stream.of(keys).forEach(newParamters::remove))`

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -240,6 +243,81 @@ public String getQueryString() {
...
+     * Return new QueryParameters including only the given parameters.
+     *
+     * @param key
+     *            Parameter name as String
+     * @param values
+     *            Values for the parameter as Strings
+     * @return QueryParameters.
+     */
+    public QueryParameters including(String... keys) {
+        Set<String> excludedKeys = Set.of(keys);
```
mcollovati, line 275: nit: rename variable to `includedKeys`

mcollovati: Can we add also here a short-circuit for empty keys, by returning `QueryParameters.empty()`?

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
TatuLund: Yes, done.

flow-server/src/main/java/com/vaadin/flow/router/QueryParameters.java
```diff
@@ -240,6 +243,80 @@ public String getQueryString() {
...
+     *
+     * @param includedKeys
+     *            Names of the parameters to be included
+     * @return QueryParameters.
+     */
+    public QueryParameters including(String... includedKeys) {
+        if (includedKeys == null || includedKeys.length == 0) {
+            return QueryParameters.empty();
+        }
+        Set<String> excludedKeys = Set.of(includedKeys);
```
mcollovati, line 274: Sorry again for not being clear enough.
I was suggesting to rename only the local variable `excludedKeys` in the method body, not the method argument.

mcollovati: For the method argument `keys` was good, so it is consistent with the other methos

mcollovati (approved):

PR #17261, opened 2023-07-20
chore: log classloading errors in reflections class finder

mcollovati:
Classloading errors may cause the frontend-build to fail, but without any useful information.
This change overrides the original class resolve method adding exception logging to help in troubleshooting build issues.

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/server/scanner/ReflectionsClassFinder.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/server/scanner/ReflectionsClassFinder.java
@@ -26,8 +26,11 @@ import java.util.LinkedHashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.reflections.Configuration;
 import org.reflections.Reflections;
+import org.reflections.util.ClasspathHelper;
 import org.reflections.util.ConfigurationBuilder;
+import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.server.frontend.scanner.ClassFinder;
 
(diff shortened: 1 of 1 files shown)


flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/server/scanner/ReflectionsClassFinder.java
```diff
@@ -107,4 +110,59 @@ private <T> Set<Class<? extends T>> sortedByClassName(
...
+                    type = typeName;
+                }
+
+                for (ClassLoader classLoader : ClasspathHelper
+                        .classLoaders(loaders)) {
+                    if (type.contains("[")) {
+                        try {
+                            return Class.forName(type, false, classLoader);
+                        } catch (Throwable ignored) {
+                            LoggerFactory
```
taefi, line 151: Not that it may have any effect on the performance to call `getLogger` inside a for loop, but probably it can just make the code more compact. If there is nothing against holding the logger instance in this static class, we may consider having it stored to avoid repeating the code for getting the logger twice.

taefi (changes requested):

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/server/scanner/ReflectionsClassFinder.java
mcollovati: Good point! Thanks.

mcollovati: done

taefi (approved):

PR #17262, opened 2023-07-20
chore: explictly use ant path matchers in VaadinWebSecurity

mcollovati:
Spring security configurers accepting strings matcher may fail to detect if the parameter is a Spring MVC pattern or not. This change update VaadinWebSecurity to explictly use ant path request matchers.


mcollovati:
This PR was just a test of the change against the latest spring security version.
I'll create a new PR targeting main branch

PR #17270, opened 2023-07-21
chore: prevent usage statistics failure when collecting info

mcollovati:
## Description

Handles potential failures due to JNA incompatibility when collecting statistic data, providing empty value as default.

Fixes vaadin/quarkus#136

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/DevModeUsageStatistics.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/DevModeUsageStatistics.java
@@ -118,7 +118,8 @@ public class DevModeUsageStatistics {
                 globalData.setValue(StatisticsConstants.FIELD_MACHINE_ID,
                         MachineId.get());
             } catch (Throwable ex) {
-                globalData.setValue(StatisticsConstants.FIELD_MACHINE_ID, "");
+                globalData.setValue(StatisticsConstants.FIELD_MACHINE_ID,
+                        "ERROR");
                 getLogger().debug("Cannot get Machine ID", ex);
             }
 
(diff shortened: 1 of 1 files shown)


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/DevModeUsageStatistics.java
```diff
@@ -114,8 +114,13 @@ private void trackGlobalData() {
                     ProjectHelpers.getProKey());
             globalData.setValue(StatisticsConstants.FIELD_USER_KEY,
                     ProjectHelpers.getUserKey());
-            globalData.setValue(StatisticsConstants.FIELD_MACHINE_ID,
-                    MachineId.get());
+            try {
+                globalData.setValue(StatisticsConstants.FIELD_MACHINE_ID,
+                        MachineId.get());
+            } catch (Throwable ex) {
+                globalData.setValue(StatisticsConstants.FIELD_MACHINE_ID, "");
```
mvysny, line 121: Idea: maybe we could log "ERROR" as the machine ID, to be extra-clear that this stats needs to be ignored. I wonder whether there are other cases where the machine ID may be computed as an empty string, and could be mixed with this case. However, this is just a question/hint, not a blocker.

mvysny (approved):

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/DevModeUsageStatistics.java
mcollovati: done

mvysny (approved):
Thank you, lgtm :+1:

PR #17276, opened 2023-07-24
fix: fail when adding null component event listener

mcollovati:
## Description

Currently, it is possible to add component event listeners pointing to a null reference. At runtime an NPE is thrown, but for an internal wrapper class, making difficult to spot where the null reference comes from. This change will throw immediately if null is given as a component event listener.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/component/ComponentEventBusTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/ComponentEventBusTest.java
@@ -21,7 +21,9 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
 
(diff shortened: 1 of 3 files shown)


czp13 (approved):
Seems good to me and a nice improvement! 🙇 

I was thinking about adding some more info to the Exception message, but as checked the `ComponentEventBus` class we usually do not add more, so better to be consistent on this front.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEventBusTest.java
```diff
@@ -740,4 +740,17 @@ public void addListener_eventDataExpressionsPresent_constantPoolKeyNotCreatedAft
...
+    public void addListener_nullListener_failFast() {
+        final TestButton button = new TestButton();
+        try {
+            button.addListener(ServerEvent.class, null);
+            Assert.fail(
+                    "Expecting an exception to be thrown for null listener");
+        } catch (IllegalArgumentException ex) {
+            Assert.assertTrue(
+                    ex.getMessage().contains("component event listener"));
+        }
```
caalador, line 754: This is fine as is, but would read better when using the `ExpectedException` rule e.g.
```
@Rule
public ExpectedException exceptionRule = ExpectedException.none();

... 
  public void addListener_nullListener_failFast() {
    exceptionRule.expect(IllegalArgumentException.class);
    exceptionRule.expectMessage("component event listener cannot be null");

    final TestButton button= new TestButton();
    button.addListener(ServerEvent.class, null);
  }
```

caalador (approved):

flow-server/src/test/java/com/vaadin/flow/component/ComponentEventBusTest.java
mcollovati: done

caalador (approved):

PR #17305, opened 2023-07-28
fix: handle spring boot relative path for static resource

### LINKED ISSUE #17303: View not loaded when Spring Boot static resource locations has a file system entry
### Description of the bug

If the project set a filesystem location (e.g. `file:web/static/`) in the `spring.web.resources.static-locations` setting of `application.properties` file, the application fails to load views with a not-blank route path, when they are accessed directly by their URL (e.g. `http://localhost:8080/mytest`).

When accessing the page, depending on the `static-locations` configuration, the request ends in an error that may be a `java.io.FileNotFoundException: /absolute/path/to/web/static/mytest` for absolute paths or `java.lang.IllegalArgumentException: URI is not hierarchical` for relative paths.

The reason is that `SpringVaadinServletService.getResourceURL()` returns a valid URL also for non-existing resources.
In the above link example, the URL is `file:web/static/mytest`.
Classpath locations work, because when `Resource.getUrl()` is invoked they throw an IOException that is handled by `SpringVaadinServletService`

### Expected behavior

`StaticFileServer` should ignore not-existing resources, allowing the view to be shown correctly

* Create a new Vaadin project `npx @vaadin/cli init my-app`
* Edit `application.properties` and add `spring.web.resources.static-locations=file:web/static` (it doesn't matter if the location exists or not)
* Open the browser and navigate to http://localhost:8080/about
* Verify that an error happens and the view is not shown

### Versions

- Vaadin / Flow version: 24.1, but probably all other versions are affected
- Java version: 17
- OS version: Linux
- Application Server (if applicable): Spring Boot with embedded Tomcat
- IDE (if applicable):

mcollovati:
## Description

StaticFileServer fails to handle missing static resources on Linux when relative paths are used to configure Spring Boot static locations for resources.
This change checks that the resource effectively exists and converts relative path to abdsolute form if required.

Fixes #17303

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


mcollovati:
For Flow 9.1 and 2.9 the fix could possibly be back ported to the `vaadin-spring` repository.

mshabarov (approved):

Issue #17042, opened 2023-06-16
Improve Theme Editor components coverage


MarcinVaadin:
Goal of this issue is to analyze and extend support for components in Theme Editor.

Vaadin components based on https://vaadin.com/docs/latest/components

Components coverage analysis done by @danielsarkwa-vaadiner https://docs.google.com/spreadsheets/d/1gFAxfpKOntJVXXN9DKTa2_71a2KCRE2NDBOiP3ezRUA/edit#gid=0

### Supported components

Component | Metadata | Comments
-- | -- | --
Accordion | ✅ | 
App Layout | ✅ |
Avatar | ✅ |
Badge | - | span with theme attribute (variant), no variant support for the moment
Board | ✅ | 
Button | ✅ | 
CRUD | ✅ | styling editor dialog not possible due to missing `overlayClassName` API in `vaadin-card` 
Charts | ✅ | basic block properties and data series colors 
Checkbox | ✅ | 
Combo Box | ✅ |
Confirm Dialog | ✅ |
Context Menu | - | mostly as given component feature, should be styled globally
Cookie Consent | ✅ | 
Custom Field | ✅ | 
Date Picker | ✅ | 
Date Time Picker | ✅ | confusing while picking date and time fields
Details | ✅ | 
Dialog | ✅ |
Email Field | ✅ |
Form Layout | ✅ | 
Grid | ✅ | 
Grid Pro | ✅ | cannot style Select editor overlay as it does not have classname
Horizontal Layout | ✅ |
Icon | ✅ | 
List Box | ✅ | 
Login | ✅ |
Login Overlay | ✅ | 
Map | ✅ | basic shape properties only 
Menu Bar | ✅ |
Message Input | ✅ | 
Message List | ✅ | 
Multi-Select Combo Box | ✅ |  
Notification | ✅ | 
Number Field | ✅ |
Password Field | ✅ |
Progress Bar | ✅ |
Radio Button | ✅ |
Rich Text Editor | ✅ |  
Scroller | ✅ |
Select | ✅ |
Side Navigation | ✅ | 
Split Layout | ✅ |
Spreadsheet | ✅ | basic shape properties only 
Tab, Tabs, TabSheet | ✅ | 
Text Area | ✅ |
Text Field | ✅ |
Time Picker | ✅ | overlay is closed on theming due to nested combobox events
Tooltip | - | mostly as input fields feature, should be styled globally
Tree Grid | ✅ | 
Upload | ✅ | 
[...]

PR #17316, opened 2023-07-28
Add meta data for message list

danielsarkwa-vaadiner:
### Finding on Message input

1. It will be more helpful to be able to set `border-bottom` for the message item other that border around the entire message item. shapeProperties only have `border-width`, would be helpful to have `border-bottom-width`

2. The default style values for `vaadin-message`, `::part(name)`, `::part(title)` and `::part(message)` are empty until you update the style with a new value

3. Avatar component `vaadin-message > vaadin-avatar` cannot be styled as a child component and vaadin avatar cannot be selected to be styled individually since it’s a slot not a direct shadow child element to `vaadin-message`.

### DIFF
--- a/vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-message-list.ts
+++ b/vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-message-list.ts
@@ -6,8 +6,19 @@ export default {
   displayName: 'MessageList',
   elements: [
     {
+      selector: 'vaadin-message-list',
+      displayName: 'Root element',
+      properties: [
+        shapeProperties.backgroundColor,
+        shapeProperties.borderColor,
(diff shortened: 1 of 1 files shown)


MarcinVaadin:
Part of https://github.com/vaadin/flow/issues/17042

MarcinVaadin:
> Avatar component vaadin-message > vaadin-avatar cannot be styled as a child component and vaadin avatar cannot be selected to be styled individually since it’s a slot not a child element to vaadin-message-list.

Using selectors like `vaadin-message-list vaadin-message > vaadin-avatar::part(abbr)` allows to style avatars also:
image

Please make sure you are working on updated main branch with fix for child selectors.

Update pushed.

MarcinVaadin (approved):

PR #17340, opened 2023-08-01
Add meta data for side nav

danielsarkwa-vaadiner:
### Finding on Side Nav

1. There isn’t support for `border-width-right` or `border-width-left`, it will be more helpful to set the border style on just one side of the `vaadin-side-nav` or it's parent wrapper.

2. To ensure consistent styles for multiple `vaadin-side-nav` and `vaadin-side-nav-item` elements within a single wrapper, you need to apply global styles. This may not be very helpful when there are multiple side nav (as sub navigation).

3. To make styles consistent for all `vaadin-side-nav-item`, selectors have been moved to the parent component but styles applied on the `vaadin-side-nav-item` has flow specificity and can't override parent style.

4. Some style properties on `vaadin-side-nav-item` label ends up affecting suffix content if it’s a text, eg: badge gets `italic` style, and leading icon gets the `font size` style.


MarcinVaadin (approved):

MarcinVaadin:
Part of https://github.com/vaadin/flow/issues/17042

PR #17341, opened 2023-08-01
fix: disable Java live reload if theme editor is opened

### LINKED ISSUE #16864: Disable the live reload when the theme editor is opened
### Description of the bug

When I open the theme editor and pick an item, I can assign a classname with the theme editor.
It changes the Java code then refresh the page if the live reload is enabled. (tested with JRebel).

When the page is refreshed, I'm losing my context (opened dialog for example) and also the theme editor.

### Expected behavior

The page should not be refreshed and the live reload should be disabled if the theme editor is opened

You can reproduce in any application

### Versions

- Vaadin / Flow version: Vaadin 24.1.beta1
- Java version: 17
- OS version: Mac

Tested with JRebel

szolo:
Enabled Java live reload closes theme editor which in turn leads to loosing user context.

Fixes #16864

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-dev-server/frontend/theme-editor/editor.ts
+++ b/vaadin-dev-server/frontend/theme-editor/editor.ts
@@ -214,6 +214,8 @@ export class ThemeEditor extends LitElement {
       themePreview.clear();
       this.refreshTheme();
     });
+
+    this.dispatchEvent(new CustomEvent('before-open'));
   }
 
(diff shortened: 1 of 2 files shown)


MarcinVaadin (approved):
Tested with JRebel, solution is blocking JRebel Browser reload requests keeping Vite HMR active for reloading CSS files.

PR #17355, opened 2023-08-03
feat: Record usage statistics when incremental build feature is disabled

mshabarov:
Helps to keep track the number of cases when developers want to run prepare-frontend always.

Related-to https://github.com/vaadin/flow/issues/15342


mshabarov:
Closed and another PR should be opened.

Issue #10761, opened 2021-04-22
Please document best practices for the Has* interfaces


mvysny:
Currently the javadoc of the `Has*` interfaces doesn't document the "best practices" => when it's good for your component to implement the particular `Has*` interface, and when it shouldn't. That leads to misunderstandings and discussions such as #9196 . Having this documented would be excellent both for users of the Vaadin framework, and also for the maintainers.

### Description of the bug / feature
The following `Has*` interfaces should document its usage patterns in its javadoc:
* HasComponents
* HasElement
* HasEnabled
* HasHelper
* HasLabel
* HasOrderedComponents
* HasSize
* HasStyle
* HasText
* HasTheme
* HasValidation
* HasValue
* HasValueAndElement

Some examples and general ideas:
* `HasComponents` should generally be implemented by layouts or components whose primary function is to host child components. It shouldn't be for example implemented by non-layout components such as fields.
* `HasText` should be implemented by components whose primary function is to have textual contents, such as Div, Span and HTML elements. It should not be implemented for example by layouts since layouts lay out elements and not text nodes (maybe).
* etc

### Versions:
    - Vaadin / Flow version: 14.5.3, 19.0.5

PR #17376, opened 2023-08-04
docs: Update mixin interface JavaDoc

tltv:
Improved documentation of Has* mixin interfaces. Added more details when to use some of the interfaces and when not.

Related-to: #10761

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/HasComponents.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/HasComponents.java
@@ -32,9 +32,9 @@ import com.vaadin.flow.dom.Element;
  * interface provides an explicit API for components that explicitly supports
  * adding and removing arbitrary child components.
  * <p>
- * {@link HasComponents} should generally be implemented by layouts or
- * components whose primary function is to host child components. It shouldn't
- * be for example implemented by non-layout components such as fields.
+ * {@link HasComponents} is generally implemented by layouts or components whose
+ * primary function is to host child components. It isn't for example
+ * implemented by non-layout components such as fields.
  * <p>
  * The default implementations assume that children are attached to
  * {@link #getElement()}. Override all methods in this interface if the
--- a/flow-server/src/main/java/com/vaadin/flow/component/HasComponents.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/HasComponents.java
@@ -32,6 +32,10 @@ import com.vaadin.flow.dom.Element;
  * interface provides an explicit API for components that explicitly supports
  * adding and removing arbitrary child components.
  * <p>
+ * {@link HasComponents} should generally be implemented by layouts or
+ * components whose primary function is to host child components. It shouldn't
+ * be for example implemented by non-layout components such as fields.
+ * <p>
  * The default implementations assume that children are attached to
  * {@link #getElement()}. Override all methods in this interface if the
  * components should be added to some other element.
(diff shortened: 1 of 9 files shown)


flow-server/src/main/java/com/vaadin/flow/component/HasLabel.java
```diff
@@ -23,7 +23,18 @@
  * The default implementations set the label of the component to the given text
  * for {@link #getElement()}. Override all methods in this interface if the text
  * should be added to some other element.
+ * <p>
+ * Example of a web component as a root element with a structure that supports
```
czp13, line 27: This sentence could be the same as:
- `Root element should be a web component...` for consistency and easier readability I feel.
(Or the other way around).

czp13 (commented):
I like these addition docs and examples, it makes the API usage more clean and understandable.

czp13 (dismissed):
Thanks for the new changes, Tomi! 🙇

flow-server/src/main/java/com/vaadin/flow/component/HasStyle.java
```diff
@@ -23,6 +23,8 @@
 
 /**
  * Represents {@link Component} which has class attribute and inline styles.
+ * <p>
+ * Implementation of {@link #getElement()} may return any type of element.
```
mcollovati, line 27: Could you explain the reason for this statement?
If the mixin method may fail or no have effects on some type of element, it would be good to expand the statement and maybe add an example.

flow-server/src/main/java/com/vaadin/flow/component/HasTheme.java
```diff
@@ -22,6 +22,8 @@
 
 /**
  * Represents {@link Component} which has theme attribute.
+ * <p>
+ * Implementation of {@link #getElement()} may return any type of element.
```
mcollovati, line 26: See comment on as `HasStyle` class

flow-server/src/main/java/com/vaadin/flow/component/HasStyle.java
tltv: It could fail with style object if element is a custom web component that does not extend Javascript's `HTMLElement`. For example if it just extends `Element`, then it does not have `style` and you could in theory get "style is undefined" errors.
So we need to be more specific in this statement. Like "Implementation of getElement() may return any HTML element or web component that supports [inline style attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style)"

mcollovati: `may return` or `should return`?

mcollovati (dismissed):

flow-server/src/main/java/com/vaadin/flow/component/HasComponents.java
```diff
@@ -32,6 +32,10 @@
  * interface provides an explicit API for components that explicitly supports
  * adding and removing arbitrary child components.
  * <p>
+ * {@link HasComponents} should generally be implemented by layouts or
```
mshabarov, line 35: Not blocking, but more like my personal preference: I'd suggest to avoid using "should" or "shouldn't", but make the working more mild to reader. For instance, "HasComponent is generally implemented by layouts...", or (in another place) "Implementation of {@link #getElement()} returns HTML element or web component ...".

czp13 (approved):

flow-server/src/main/java/com/vaadin/flow/component/HasComponents.java
tltv: With milder tone it is indeed a bit friendlier to read. I agree and will update.

tltv: But it only works in `HasComponents` and `HasText`. In other places it changes message so that reader may think that interface implements `getElement()` for you which it doesn't.

tltv: Also it works nicer in `HasValidation`. I'll update it as suggested in three classes.

mcollovati (approved):

mshabarov (approved):

PR #17381, opened 2023-08-07
Add meta data for Chart

danielsarkwa-vaadiner:


MarcinVaadin:
Fixed a bug when same variable name was used in different block statements.

MarcinVaadin (approved):

Issue #14553, opened 2022-09-16
Synchronize rerouteTo, forwardTo and navigate signatures


OlliTietavainenVaadin:
### Describe your motivation
Rough table of which types of parameters are accepted to each method:
<html>
<body>
<google-sheets-html-origin><style type="text/css"></style>

T parameter = target interface HasUrlParameter<T> type |   |   |   |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
  | class | class + T parameter | class + T parameter + QueryParameters | class + RouteParameters | String | String + QueryParameters | NavigationState | NavigationHandler + NavigationState | String + List<T> locationParams | String + T parameter
navigate | x | x | x | x | x | x |   |   |   |  
forwardTo | x |   |   | x | x |   | x | x | x | x
rerouteTo | x |   |   | x | x |   | x | x | x | x

</body>
</html>

`forwardTo`, `rerouteTo`, and `navigate` are all navigation methods and should therefore be usable consistently unless technically required to behave differently. So if any of them accepts a class as the navigation target, so should the others. If any one of them accepts a String to identify the target, so should the others. Most importantly, since `navigate` accepts route parameters being part of the location String, `forwardTo` and `rerouteTo` should also accept route parameters this way.

### Describe the solution you'd like

There should be a better consistency of the signatures of the methods; at the very least, Query Parameters should be supported in every one of them.

### Describe alternatives you've considered

Do we actually need three methods for in-app navigation? Could we drop one (or even two) of them?

OlliTietavainenVaadin:
Related issue: https://github.com/vaadin/flow/issues/14562

TatuLund:
> Do we actually need three methods for in-app navigation? Could we drop one (or even two) of them?

It is a good question, where the answer is most likely yes.

UI#navigate starts complete navigation cycle, and calling UI#navigate while navigation cycle is not complete is not allowed. I.e. you are not supposed to use UI#navigate in beforeEnter for example, you can however call it in afterNavigation as in that stage navigation cycle is complete. Hence rerouteTo/forwardTo is necessary as you want to have have option to divert navigation while cycle is not complete yet. And this comes in two variants, either preserving the url or not. So if reducing number of methods it is theoretically possible to join rerouteTo/forwardTo by introducing additional parameter for url preservation feature. This naturally will be breaking change, which we probably should avoid introducing at this stage.

PR #17407, opened 2023-08-09
feat: Update navigation method signatures 

tltv:
This change makes UI#navigate, BeforeEvent#rerouteTo and BeforeEvent#forwardTo methods more consistent without changing the existing API:
- adds new methods in BeforeEvent that accepts route parameters.
- adds new methods in BeforeEvent that accepts query parameters.
- improves forwardTo and rerouteTo to accept route parameters being part of the location String (same way as UI#navigate does it already).

Fixes: https://github.com/vaadin/flow/issues/14553

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/BeforeEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/BeforeEvent.java
@@ -48,6 +48,7 @@ public abstract class BeforeEvent extends EventObject {
 
     private final Class<?> navigationTarget;
     private final RouteParameters parameters;
+    private QueryParameters redirectQueryParameters;
     private final List<Class<? extends RouterLayout>> layouts;
     private NavigationState forwardTargetState;
     private NavigationState rerouteTargetState;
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/router/BeforeEvent.java
```diff
@@ -717,6 +1033,17 @@ public RouteParameters getRouteParameters() {
...
+     * Gets the query parameters for forwarded and rerouted URL. {@code null}
+     * means that query parameters of the event are preserved in the forwarded
+     * and rerouted URL.
+     *
+     * @return query parameters for forwarding and rerouting
+     */
+    public QueryParameters getRedirectQueryParameters() {
+        return redirectQueryParameters;
+    }
+
```
mcollovati, line 1046: Would it make sense to add a `hasRedirectQueryParameters()` to be consistent with the other properties?

mcollovati (approved):

mshabarov (approved):

PR #17410, opened 2023-08-09
bugfix: fixes and optimizes everything regarding event debouncing :-)

mstahv:
## Description

The event order when using debouncing sends events in wrong order. This was tried to be fixed in https://github.com/vaadin/flow/pull/14035 but apparently only issues with the core TextField was fixed (or then this was again regressed in the regression fix https://github.com/vaadin/flow/issues/15615) 🤷‍♂️

Using this trivial Element API usage like below, type in for example `asdfg` to the input field -> g-event fires before k-event and there are multiple server visits (there should be just one).

```
    Element el = new Element("input");
    el.executeJs("""
            const el = this;
            var id = 0;
            this.addEventListener('keydown', function(event) {
                id++;
                const ke = new Event("k-event");
                ke.id = id;
                el.dispatchEvent(ke);
            
                if(event.key == 'g') {
                    const ge = new Event("g-event");
                    ge.id = id;
                    el.dispatchEvent(ge);
                }
            });
    """);
    DomListenerRegistration keyreg = el.addEventListener("k-event", e -> {
        System.out.println(LocalTime.now() + " k-event" + e.getEventData().getNumber("event.id") + " phase: " + e.getPhase());
    });
    keyreg.addEventData("event.id");
    keyreg.debounce(3000);

    // Note, with multiple debounce phases, things get really weird...

    // I'm interested of g-event eagerly, so I don't debounce it
    // BUT, AFAIK, key-event should ALWAYS come in before!!
    // But it is not... :(
    DomListenerRegistration greg = el.addEventListener("g-event", e -> {
        System.out.println(LocalTime.now() + " g-event " + e.getEventData().getNumber("event.id"));
    });
    greg.addEventData("event.id");

    getElement().appendChild(el);

```

While investigating the this issue there was also a dozen of other related isseus, incomplete docs and missing integration test.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/DomEventFilterIT.java
```diff
@@ -46,52 +48,127 @@ public void filtering() {
     public void debounce() throws InterruptedException {
         open();
 
-        WebElement input = findElement(By.id("debounce"));
+        /*
+         * Note, the client side implementation "merges" all these settings in
+         * the single "Debouncer" (per element-eventtype-timeout, not per server
+         * side listenr). This is probably an issue in the original
```
mcollovati, line 54:
```suggestion
         * side listener). This is probably an issue in the original
```

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/DomEventFilterIT.java
```diff
@@ -46,52 +48,127 @@ public void filtering() {
...
+         * implementation that could maybe be refactored in some upcoming
+         * release, but then we should create identifiers for server side
+         * listeners and handle those somehow separately. My hunch is That it is
```
mcollovati, line 57:
```suggestion
         * listeners and handle those somehow separately. My hunch is that it is
```

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/DomEventFilterIT.java
```diff
@@ -46,52 +48,127 @@ public void filtering() {
...
+         * better just to documenta the limitations and focus on somethign that
```
mcollovati, line 58:
```suggestion
         * better just to document the limitations and focus on something that
```

mcollovati (changes requested):
Just some typos

mcollovati:
The overall change looks good, and the tests are more reliable.
However, since the change involves a behavior change (), we need to:

* update documentation (https://vaadin.com/docs/latest/create-ui/creating-components/events#limiting-event-frequency)
* update `DebouncePhase` javadoc

flow-server/src/main/java/com/vaadin/flow/dom/DomListenerRegistration.java
```diff
@@ -116,21 +116,67 @@ DomListenerRegistration setDisabledUpdateMode(
...
+     * if the user input has started.</dd>
+     * <dt>DebouncePhase.TIMEOUT</dt>
+     * <dd>In this case the listener is triggered only after the given timeout.
+     * If there are multiple events during that period, only the last one is
+     * reported.</dd>
+     * <dt>DebouncePhase.LEADING + DebouncePhase.TIMEOUT</dt>
+     * <dd>In this case the listener is triggered when an event burst starts,
+     * but afterwards only after the given timeout has passed. There is
+     * {@link #throttle(int)} shorthand for this mode.</dd>
+     * </dl>
```
mcollovati, line 151: It seems like `DebouncePhase.LEADING + DebouncePhase.TIMEOUT` is repeated

mstahv: fixed

mstahv:
Nope, it is fine, "just" other slightly related issues (#17430, #17429) causing obsolete server visit.

mstahv:
Checked the reference docs. Nothing urgent needs to change there. Of course more hands on examples of everything would be nice. And maybe mentioning that combining all modes is weird and may cause double calls for the same event 🤷‍♂️

mshabarov:
@mstahv With this patch I see the same behaviour as previously, without it, "g" event is triggered on the server first, and "k" goes after it:
```
14:34:24.148306 g-event 5.0
14:34:27.798952 k-event6.0 phase: TRAILING
```
How did you test it?

mstahv:
Regression in the branch 🤔 Make sure you are not causing another keydown (like shortcut to change focus to IDE) event if you are using the original case in the issue description. I tested the original case with a separate project, but then the changed the implementation still while making the IT test suit more complete 🤷‍♂️ But it still don't seem to contain test with two different event types.

Would have been easier to check two weeks ago 😤

mstahv:
Added case to DomEventFilterIT/View locally testing with Safari all seems to work fine, but can't these days be sure as the front-end bundle things are rather complex 🤓

mshabarov (approved):

mshabarov:
The same behaviour was observed because of pre-compiled dev bundle was used, which did not contain these changes. With forcing the frontend build, these changes are now shown and seems to work as expected: one request to server is sent and events go in order.
Let's apply this patch and check Flow Components tests later.
I propose to include this only to 24.2 until enough time passes by, then we can pick it to other branches.

PR #17412, opened 2023-08-10
test Jetty12

Artur-:


mshabarov:
@Artur- do you have something in mind where we should look into, to migrate to Jetty 12?

Artur-:
When Spring Boot has integrated it, tracked in https://github.com/spring-projects/spring-boot/issues/36073

PR #17470, opened 2023-08-18
chore: Update LICENSE

mshabarov:


caalador (commented):
Could we have the extra spaces at the end of the lines cleaned away so that the actual changes would be more obvious?

mshabarov:
> Could we have the extra spaces at the end of the lines cleaned away so that the actual changes would be more obvious?

Yes, definitely, let me try do this.

caalador (approved):

PR #17494, opened 2023-08-24
feat: Micrometer Observation instrumentation

marcingrzejszczak:
## Description

- adds a VaadinFilter mechanism to mimic an around-aspect
- adds a Micrometer Observation implementation
- configures VaadinServlet to automatically update the VaadinService with filters
- configures Spring to automatically update the VaadinServlet with filters

What's missing:
- Client side instrumentation
- Tests
- Documentation

Fixes #17436

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

Issue #17436, opened 2023-08-14
Micrometer Observation Support


marcingrzejszczak:
### Describe your motivation

I'm a co-maintainer of [Spring Cloud Sleuth](https://github.com/spring-cloud/spring-cloud-sleuth/) and [Micrometer](https://micrometer.io/) projects (together with @shakuzen and @jonatan-ivanov).

[Micrometer Observation](https://micrometer.io/docs/observation) is part of the Micrometer 1.10 release and [Micrometer Tracing](https://micrometer.io/docs/tracing) gives support for Tracing capabilities. The idea of Micrometer Observation is that you instrument code once but you get multiple benefits out of it - e.g. you can get tracing, metrics, logging or whatever you see fit).

I was curious if there's interest in adding Micrometer Observation support so that except for metrics, spans could be created and tracing context propagation could happen too. Via Micrometer Tracing one can use OpenTelemetry or OpenZipkin Brave Tracer, but with the handler mechanism the possibilities are endless :)

### Describe the solution you'd like

We can add a PR to instrument the RPC communication of Vaadin.

### Additional context

[Micrometer Observation Documentation](https://micrometer.io/docs/observation)

mshabarov:
Hello @marcingrzejszczak and thank you for creating this enhancement!
This sounds for me (and I bet for all the Vaadin team) more than interesting. We consider the "observability" as one of the most valuable corners of business applications. As a product owner of Vaadin Flow, I'd appreciate any work on this topic. If you have any thought or even a suggestions about actual implementation towards integrating Micrometer Observation into Vaadin, please share it in the pull request https://github.com/vaadin/flow/compare . Vaadin Flow team is happy to review, test and help driving this feature further.

mshabarov:
> please share it in the pull request https://github.com/vaadin/flow/compare

@marcingrzejszczak we discussed Micrometer integration internally with the team. We are considering adding this support to our commercial [Observability Kit](https://vaadin.com/docs/latest/tools/observability), and make a research by ourselves first. Thus, I would ask you to hold your contribution for a while. We'll get back to you once we get a vision of the solution and whether Micrometer Observation can be added into Vaadin Flow core.

mshabarov:
@marcingrzejszczak we decided to proceed with Micrometer and Vaadin integration and to try instrument the Vaadin Flow core/RPC communication. Please assign me to the pull request and/or ping me whenever you need a help or assistance. Thank you!

marcingrzejszczak:
[...]

PR #17502, opened 2023-08-25
feat: Adds a VaadinRequestInterceptor abstraction

marcingrzejszczak:
## Description

VaadinFilter simulates an around aspect around processing of a request. Prerequisite for https://github.com/vaadin/flow/pull/17494

Related to #17436 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


marcingrzejszczak:
> Quick comment on the VaadinFilter (although now in separate PR): maybe it could also cover UI.access calls coming from separate threads 🤔 And then requests & al are optional or within some context parameter 🤷‍♂️

@mstahv can you elaborate on that?

flow-server/src/main/java/com/vaadin/flow/server/VaadinFilter.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Used to provide an around-like aspect option around request processing.
+ *
+ * @author Marcin Grzejszczak
+ * @since 24.2
+ */
+public interface VaadinFilter extends Serializable {
```
mshabarov, line 27: For me the "Filter" associates to the filter chain patter and http filters: processing stops when one particular filter can handle the request.
This isn't so in our case.
I would propose to rename it to `VaadinRequestInterceptor`.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -64,26 +29,30 @@
...
 import com.vaadin.flow.shared.communication.PushMode;
-
 import elemental.json.Json;
 import elemental.json.JsonException;
 import elemental.json.JsonObject;
 import elemental.json.impl.JsonUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
```
mshabarov, line 44: Imports should be set explicitly without wildcard.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -2374,6 +2356,14 @@ public static String getCsrfTokenAttributeName() {
                 + ApplicationConstants.CSRF_TOKEN;
     }
 
+    public List<VaadinFilter> getVaadinFilters() {
```
mshabarov, line 2359: I'd prefer to have an unmodifiable collection of these filters/interceptors once it is initialised.

In `flow-server` I think we shouldn't populate any filters (another big topic), at least for observations.

These filters are supposed to be populated by Vaadin (in a separate module for observations, or for other means in the future in flow-server) or by Vaadin users in their projects.

This can be done through Service Init Listener API, as described here https://vaadin.com/docs/latest/advanced/service-init-listener .
We need an extra method in `ServiceInitEvent `, where one can add the filters before `VaadinService` is initialised. 
How sure if it's an easiest way of adding these filters, maybe for Spring-based apps it is easier to define a bean in the configuration. For non-Spring apps, it's an option.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -114,6 +83,8 @@ public abstract class VaadinService implements Serializable {
             + PushMode.class.getSimpleName() + "." + PushMode.DISABLED.name()
             + "." + SEPARATOR;
 
+    private List<VaadinFilter> vaadinFilters = new ArrayList<>();
```
mshabarov, line 86: It can be `Iterable` instead, like in the case with `RequestHandler` collection.
We only want to iterate through the filters, or is there any use case where getting by index makes sense?

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -2374,6 +2356,14 @@ public static String getCsrfTokenAttributeName() {
                 + ApplicationConstants.CSRF_TOKEN;
     }
 
+    public List<VaadinFilter> getVaadinFilters() {
+        return vaadinFilters;
+    }
+
+    public void setVaadinFilters(List<VaadinFilter> vaadinFilters) {
```
mshabarov, line 2363: Javadoc is needed, explaining when this setter and getter are used/needed.

mshabarov (changes requested):
Looks good to me in general.
Two concerns about naming of this new class and the fact that both servlet and service has these filters mutable.
If possible, I would preference to have immutable collection once it is initialised.
I will check how it can be done.

marcingrzejszczak:
@mshabarov I've checked the Service init mechanism that you mentioned and decided to rewrite my proposal to use that mechanism. You can check the details [here](https://github.com/vaadin/flow/pull/17502/commits/e5267aa97c929eea611ec16a2bef1b63688fa39b). I like it better because there is just 1 way of setting the interceptors common for spring and non-spring apps. PTAL

mstahv:
Explaining my comment regarding the UI.access & requests:

 * Not every change in Vaadin UIs happen with an active http request
 * UIs can be modified by via other users request or by external threads as well
 * UIs are then actually updated by the next XHR or via constantly open web socket communication
 * For app developers it would/could be handy if they could hook into begin and end of a "UI access", which are typically wrapped in calls passed to  UI.access calls, unless they already happen within a "UI thread". One could for example open and close user specific backend/database connection or do various profiling related tasks

I didn't look through the change proposal thoroughly, but by quickly skimming it through (mostly checkin names of new classes), this would help only with common XHR connections and user originated UI access 🤔

marcingrzejszczak:
From my perspective we need to wrap all the entry points to the backend and everything going out of the backend. If we can control what goes into and out of the UI that would be perfect. Once we have such a list of entry and exit points I can instrument them (or at least try to :) )

mshabarov:
It makes sense to me what Matti said about having hooks for `UI.access`.

`VaadinService` has `runPendingAccessTasks` where we can hook and call request interceptors before and after calling UI access pending futures. `VaadinRequestInterceptor` can be extended with extra methods for `UI.access`.

For Vaadin Push connections (communications via WebSockets), the `requestStart` and `requestEnd` methods are already being called for every request from client, see `PushHandler::callWithUi` methods.

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -1544,6 +1582,18 @@ private void handleExceptionDuringRequest(VaadinRequest request,
             vaadinSession.lock();
         }
         try {
+            try {
+                vaadinRequestInterceptors
```
mshabarov, line 1586: `handleExceptionDuringRequest` doesn't actually handle all the errors during request processing.
It catches internal errors, but errors that might happen inside lambdas provided by developers (mainly UI updates, for instance, if you pass an UI update into `beforeClientResponse`) are processed by `ErrorHandler`, so that the exception is not thrown up to the invocation hierarchy, but handled by this error handler. See https://vaadin.com/docs/latest/advanced/custom-error-handler.

This might be a default error handler or a custom error handler, which, for example shows the notification with error message or changes component's styles. 

Thus, to cover also the exceptions handled by `ErrorHandler` objects, we may add a `ErrorHandlerWrapper` that takes `ErrorHandler` in `VaadinSession::setErrorHandler` and decorates it with calls to interceptors.

flow-server/src/main/resources/META-INF/services/com.vaadin.flow.server.VaadinServiceInitListener
```diff
@@ -0,0 +1 @@
+com.vaadin.flow.server.VaadinRequestInterceptorServiceInitListener
```
mshabarov, line 1:
```suggestion
com.vaadin.flow.server.VaadinRequestInterceptorServiceInitListener

```

mshabarov (changes requested):
My previous comment are fixed, awesome work! 👍 
Three points I would like to fix and clarify before we could merge it: 
- Wrapping `UI.access` calls with vaadin request interceptor to be able to add instrumentations for it (discussed it already with Matti T. in the comments above)
- Cover `ErrorHandler` calls, so that the exceptions thrown in the UI update callbacks in the user's codes would be also wrapped and observable (see my comment below)
- Maybe more global question: should we follow the same instrumentation list as we do for [Vaadin Observability Kit](https://github.com/vaadin/observability-kit/tree/main/observability-kit-agent/src/main/java/com/vaadin/extension/instrumentation), e.g. all Vaadin-provided request handlers, static file handler, data communicator and so on. 
My vision is that we can do it iteratively, so we can apply this pull request, add instrumentation (another opened PR) for it and then extend `VaadinRequestInterceptor` step-by-step with extra methods for concrete instrumentation and egress points.
What do you think @heruan and @Legioth ?

marcingrzejszczak:
> Wrapping UI.access calls with vaadin request interceptor to be able to add instrumentations for it (discussed it already with Matti T. in the comments above)

I was thinking that we could instrument `com.vaadin.flow.server.VaadinService#runPendingAccessTasks`. But we would have to instrument using a different interceptor. The `VaadinRequestInterceptor` is really related to http. Here we could have an interceptor around executing commands. Would that make any sense?

> Cover ErrorHandler calls, so that the exceptions thrown in the UI update callbacks in the user's codes would be also wrapped and observable (see my comment below)

If we did this via the interceptor that I mentioned above we would also take care of the error handling?

> Maybe more global question: should we follow the same instrumentation list as we do for [Vaadin Observability Kit](https://github.com/vaadin/observability-kit/tree/main/observability-kit-agent/src/main/java/com/vaadin/extension/instrumentation), e.g. all Vaadin-provided request handlers, static file handler, data communicator and so on.
> My vision is that we can do it iteratively, so we can apply this pull request, add instrumentation (another opened PR) for it and then extend VaadinRequestInterceptor step-by-step with extra methods for concrete instrumentations and egress points.
> What do you think @heruan and @Legioth ?

I am all for baby steps ;)

mshabarov:
> I was thinking that we could instrument com.vaadin.flow.server.VaadinService#runPendingAccessTasks. But we would have to instrument using a different interceptor. The VaadinRequestInterceptor is really related to http. Here we could have an interceptor around executing commands. Would that make any sense?

Yes, this makes sense to me to add a new aspect for pending access invocations. Let's do it in a separate pull request.

> If we did this via the interceptor that I mentioned above we would also take care of the error handling?

`UI.access` and whatever server side code that can pass through the `ErrorHandler` are quite unrelated to each other.
`UI.access` refers to a `PUSH` functionality, i.e. asynchronous server-to-client UI updates, whereas `ErrorHandler` intercepts server-side exceptions, including (but not only limited to `UI.access`), exceptions in synchronous handlers, like "before client response" commands and RPC requests from clients. 

From the other hand, when the `ErrorHandler::error` method is called, it doesn't know from what context it's called, and this is correct. This means that we cannot easily decide which interceptor to call, when this error handler is called: request or async interceptor. If we skip intercepting error handler :: error calls, UI update codes in command would be ignored by these interceptors and not instrumented, which will make user of this API upset a lot.

A simplest solution IMO is to make a decorator for error handler, whenever it is being added to the session here
https://github.com/vaadin/flow/blob/daecc25f95d055f3e6caec10522e55d1c12c03db/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java#L429, put there a call to `VaadinRequestInterceptor::handleException`. This would intercept also errors for async updates, this might be an issue for API users and a bit confusing.

> I am all for baby steps ;)

We discussed the design internally, and our decision now is to focus on the basic request/response instrumentation, not trying to repeat all the features implemented with OpenTelemetry standard in Vaadin Observability Kits, because it would need a really deep code changes in some cases. This definitely better to do in a separate pull requests. For most of the users even having a request/response/error instrumentation is already a big feature.

flow-server/src/main/java/com/vaadin/flow/server/VaadinRequestInterceptorServiceInitListener.java
```diff
@@ -0,0 +1,63 @@
...
+ * Sets all {@link VaadinRequestInterceptor} on an {@link ServiceInitEvent}.
+ *
+ * For internal use only. May be renamed or removed in a future release.
+ *
+ * @see ServiceInitEvent
+ * @see VaadinService#init()
+ * @author Marcin Grzejszczak
+ * @since 24.2
+ */
+public class VaadinRequestInterceptorServiceInitListener
```
mshabarov, line 30: By the way, why do we need this init listener? 
Looks like doing this way, we make an extra loop forcing Flow to take a detour, whereas this can be "hardcoded" here https://github.com/vaadin/flow/blob/349cc1105a1e5901863ddd210120ff3606490e47/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java#L232 , the same way as for request handler.

This might lead to not being able to autowire the interceptors as a Spring bean, but I believe this would be fine, you can still create a `VaadinRequestInterceptorServiceInitListener` and say that it should be a `AtComponent`.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
```diff
@@ -119,6 +119,13 @@ public ServletRegistrationBean<SpringServlet> servletRegistrationBean(
         return registration;
     }
 
+    @Bean
+    public VaadinRequestInterceptorServiceInitListener vaadinRequestInterceptorServiceInitListener(
+            ObjectProvider<VaadinRequestInterceptor> interceptors) {
```
mshabarov, line 124: On a second look this seems to me as an overhead. 
We don't have a similar API/ability to add request handler and whatever objects developers can add in `ServiceInitListeners`.

Plus, projects which already have a service init listener can just add one line to their listener to add the interceptors.

See my another comment https://github.com/vaadin/flow/pull/17502/files#r1327153517.

So, if you have no big objections, I would propose to remove this from autoconfiguration for simplicity.

flow-server/src/main/java/com/vaadin/flow/server/VaadinRequestInterceptorServiceInitListener.java
marcingrzejszczak: I mean I think I was following what was being done now for `RequestHandlers`. I've actually copy pasted the whole code that the request handlers are following and did the same for interceptors. 

My analysis:

- A request interceptor `List<RequestHandler> handlers = createRequestHandlers();` is initialized. These handlers do not require any collaborators (they have default constructors only, which will not be the case when doing instrumentations - you need tracers, registries etc.)
  - I'm doing exactly the same for the interceptors `List<VaadinRequestInterceptor> requestInterceptors = createVaadinRequestInterceptors();`
- when `VaadinService` is initialized it calls ` instantiator.getServiceInitListeners().forEach(listener -> listener.serviceInit(event));` what will initialize the event with whatever the listeners want to add (e.g. custom interceptors etc.).  The custom listeners would set interceptors such as the micrometer observation interceptor
  - I'm doing the same for interceptors - I read the entries from the event and add them to an unmodifiable list
  - For the compatibility with Spring I've added a second constructor for constructor injection

I don't understand what I should actually do then if not what I've already done

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
marcingrzejszczak: >  We don't have a similar API/ability to add request handler and whatever objects developers can add in ServiceInitListeners.

I mean you do, ServiceInitListener works via SPI and it can append things to the init event. I thought that it's done exactly for this purpose.

I can do whatever you want but I don't know how you would like this to work :) How will you want to add the request interceptors to the VaadinService?

mshabarov: Here is my vision of design so far:
- `flow-server` creates an empty list of interceptors by default and makes this protected, so developers can change it if they want. In the future we might want to add some interceptors by default (for whatever purpose). `flow-server` collects interceptors from `ServiceInitEvent` and makes an immutable collection that is used for further http request interception. (this is already done and look awesome).
- `flow-micrometer` module in the Flow repo adds a `VaadinRequestInterceptorServiceInitListener` as a Spring's`AtComponent` and injects there all needed Micrometer Interceptors. No SPI registration is needed. If this module is in the classpath of the project, this listener should be picked and executed by Flow.
- Vaadin plain projects can register their own service init listeners with SPI and Vaadin Spring-based projects can register this listener as a component and autowire interceptors.
- No changes to `Instantiator` and `SpringBootAutoConfiguration` is needed for the above points to work.

flow-server/src/main/java/com/vaadin/flow/server/VaadinRequestInterceptorServiceInitListener.java
marcingrzejszczak: Looks awesome! I mean we have the same outcome by having less code I'm all for it :D So how do we proceed now?

mshabarov: I cannot push into your branch, because I have no rights.
If you can apply my patch locally, could you commit it and push (only changes in `flow-server` and `vaadin-spring`, others are use case codes and local workarounds) it?
Also, tests I think need some tweaks, if you have a chance please do, otherwise I can help you with it.

mshabarov:
> While waiting for feedback I started playing around with this concept and I got sth like this out of it (check the CommandInterceptor)

Great! Thanks for the enthusiasm 👍 🥇 
I propose to make a separate PR for it, it would be much more easier.

mshabarov:
@marcingrzejszczak So I propose to apply my last patch https://github.com/vaadin/flow/pull/17502#discussion_r1330195191, adapt the tests for it and then I propose to merge this PR, because it's going to be endless :)

Then we can consider making a new module for instrumentation and command interceptor.

marcingrzejszczak:
Yup, I'll do it next week!

marcingrzejszczak:
I've just applied the patch, let's see how the build goes...

mshabarov:
Please run `mvn formatter:format` to let the format checker step pass.

vaadin-spring/src/test/java/com/vaadin/flow/spring/service/SpringVaadinServletServiceTest.java
```diff
@@ -89,4 +95,21 @@ public void uiInitListenerAsSpringBean_listenerIsAutoregisteredAsUIInitiLietnerI
         Assert.assertSame(ui, listener.events.get(0).getUI());
     }
 
+    @Test
+    public void requestInterceptorsAreRegisteredOnTheService()
```
mshabarov, line 99: I'm not sure that I understand how this is supposed to work.
How `MyRequestInterceptor` instances get injected into `VaadinService`? I see no service init listener here.

Would it be more clear if you:
1. Move `LoggerRequestInterceptor` and `MyVaadinServiceInitListener` classes from `test-spring-boot` module to here
2. Add `MyVaadinServiceInitListener` to the Spring context (maybe this doesn't even needed if Spring test configuration can recognise the bean and add it automatically), so this listener would pass through the `VaadinService` and it's interceptors are collected and added to the list by new logic in the flow-server.
3. Pass any `VaadinRequest` to the `VaadinServlet`/`VaadinService` and check that the interceptor does changes in the request/response (like you already does in flow-server test).

Making this kind of test, we would repeat the use case of how these interceptors are added to Flow in reality.

mshabarov (changes requested):
Implementation looks good to me. 
A few comments about leftover codes and imports, and one comment about test.

vaadin-spring/src/test/java/com/vaadin/flow/spring/service/SpringVaadinServletServiceTest.java
mshabarov: Looks like this test fails with:
```
Error:  Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.172 s <<< FAILURE! -- in com.vaadin.flow.spring.service.SpringVaadinServletServiceTest
Error:  com.vaadin.flow.spring.service.SpringVaadinServletServiceTest.requestInterceptorsAreRegisteredOnTheService -- Time elapsed: 0.035 s <<< FAILURE!
org.opentest4j.AssertionFailedError: There should be 1 filter ==> expected: <1> but was: <0>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:560)
	at com.vaadin.flow.spring.service.SpringVaadinServletServiceTest.requestInterceptorsAreRegisteredOnTheService(SpringVaadinServletServiceTest.java:108)
```

marcingrzejszczak: Honestly, after the changes you've made I don't know how this is supposed to work but I'm not a Vaadin expert ;) I'm trying to make the tests work now. Will ping you once I can't figure things out

marcingrzejszczak: Actually I've made the test pass, let me see if everything else is working fine and that formatting is not broken and I will push the changes! Your comments were very helpful.

marcingrzejszczak:
Ok, hopefully everything is fine now. I do have problems with the wildcard imports and formatting because I can't seem to align with the intellij idea formatting style for this project. Anyways, hopefully that's resolved now

mshabarov (changes requested):

knoobie:
Before it is forgotten, can we get the 24.2 label? ;)

mshabarov (approved):

PR #17504, opened 2023-08-28
fix: Update UI if custom errorHandler

### LINKED ISSUE #17352: Custom ErrorHandler implementation's UI changes are ignored
### Description of the bug

We have created a customer error handler by extending the DefaultErrorHandler class.
This is added to the session, and it would create a Dialog (or later a Notification for testing).

The feature works when the exception occurs during component constructor, in general when it is a navigation to a given target.
The feature doesn't work, when the exception occurs in the lambda passed to a grid.addComponent method.

I debugged what happens, in the non-working case the exception is caught at VaadinService.handleExceptionDuringRequest

here the set errorHandler is called on line 1586
```
            if (vaadinSession != null) {
                vaadinSession.getErrorHandler().error(new ErrorEvent(t));
            }

```
During the execution everything looks ok, there is a UI instance, the ui component is created, and even attached to the ui. Nothing is visible in the browser.

on line 1594 writeUncachedStringResponse is called with the system internal error message, and it is set as the response.
Checking it in the browser, the response looks like:
```
for(;;);[
    {
        "changes": {},
        "resources": {},
        "locales": {},
        "meta": {
            "appError": {
                "caption": "Internal error",
                "url": null,
                "message": "Please notify the administrator. Take note of any unsaved data, and click here or press ESC to continue.",
                "details": null,
                "querySelector": null
            }
        },
        "syncId": -1
    }
]
```

I think the reason is, that in the above message the Vaadin server does not send the ui changes requested in the errorHandler, not even if it is wrapped in a UI.access call. These tests were running with a single browser tab open for the server, so it cannot swap the ui instance to another one.

The online documentatation does state, that the errorHandler can add a notification, or other ui changes.

https://vaadin.com/docs/latest/advanced/custom-error-handler

### Expected behavior

The requested ui changes should appear in the browser, for example Dialog open.

```
public class TestExceptionHandler extends DefaultErrorHandler {

@Override
  public void error(final ErrorEvent event) {
    Command c = () -> Notification.show("message").addThemeVariants(NotificationVariant.LUMO_ERROR);
    Optional.ofNullable(UI.getCurrent()).ifPresent((UI ui) -> ui.accessSynchronously(c));
  }
}
....
public class TestGrid extends TreeGrid<TestDto> {

       public TestGrid() {

        addComponentColumn(this::getTestColumn)
                .setHeader("Test Column")
                .setSortProperty("test);
}

    private Component getTestColumn(TestDto testDto {
        throw new NullPointerException(); // throws an exception from a columnGenerator
 }
}
```

### Versions

- Vaadin / Flow version: 23.3.16
- Java version: 17
- OS version: Linux
- Browser version (if applicable): n/a
- Application Server (if applicable): embedded jetty
- IDE (if applicable): n/a

caalador:
Enable update of UI for beforeClientResponse
execution throwing an exception when a 
custom ErrorHandler is set.

Fixes #17352


mshabarov (commented):
Not sure that I understand why `vaadinSession.getErrorHandler().error(new ErrorEvent(t));` isn't called or doesn't apply UI error-related custom changes when before client reponse callback throws.
When the exception occurs in the callback, it arises from `writeUidl` to the `handleExceptionDuringRequest`.
How it makes difference where the exception is thrown, because UIDL write is interrupted anyway if it happens a bit before or later than before client response callbacks handling.
I assume that besides `runExecutionsBeforeClientResponse` method we have few other places where exceptions potentially can happen that also would break the UIDL writing.
Will do more code digging.

caalador:
So the normal UIDL request is handled in ServerRpcHandler before write and normal exceptions happen during the invocation handling. `UidlRequestHandler::l114` -> `ServerRpcHandler::l323` -> `ServerRpcHandler::l440` -> `ServerRpcHandler::handleInvocationData` here you can see that it is handled inside a try catch block that calls the errorHandler and continues execution.
Then when the project throws during the writeUidl when executing `beforeClientResponse`s we will fall all the way up to `VaadinService::handleRequest` and execute `VaadinService::handleExceptionDuringRequest` that will run the error handler, but only send the internal error back to the client.

mshabarov (approved):
Thanks @caalador for explanations, now I think I got it. Looks good to me. Seems like there is nothing else called in `writeUidl` that might have a user code. JS invocations maybe, but they are not executed on the server anyway.

tomivirkki:
I wonder if this caused some breaking change? After this was merged, certain [`EditorImplTest`](https://github.com/vaadin/flow-components/blob/main/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/editor/EditorImplTest.java#L40) unit tests in `flow-components` started failing.

caalador:
Nothing should change if there is no custom ErrorHandler set.
It would seem spreadsheet, upload and grid add one for on attach, so that would make it not show internal error on the client side and no throw as the handler should handle that. I guess its  `editorIsInBufferedMode_closeEditorThrows` that fails?

tomivirkki:
It's actually those thats that have `@Test(expected = IllegalStateException.class)`:

```
java.lang.Exception: Unexpected exception, expected[java.lang.IllegalStateException] but was[java.lang.NullPointerException]
Caused by: java.lang.NullPointerException: Cannot invoke "Object.getClass()" because the return value of "com.vaadin.flow.server.VaadinSession.getErrorHandler()" is null
```

caalador:
Ok. That's an interesting state, I'll add a null handler test and look into it.

caalador:
Seems the underlying issue is the mock session that returns a null ErrorHandler at which it fails. Fix incoming.

caalador:
#17542

PR #17514, opened 2023-08-29
test: add test to measure reload time of plain Spring boot app

tltv:
Adds new module 'test-plain-spring-boot-reload-time' with a TestBench test that measures reload time of Hello world Spring boot application. Application is build with plain Spring boot application without any Flow dependencies other than TestBench for tests. Test result is reported to TeamCity. Data is used for comparing reload times with different sized Flow apps.


tltv:
Issues with bender are now fixed and PR is ready for review.

tepi (approved):

PR #17529, opened 2023-08-31
feat: Added methods to prevent default and stopping propagation

### LINKED ISSUE #1363: EventHandler should be able to stopPropagation and preventDefault
@EventHandler should be extended to have the possibility to stopPropagation and/or preventDefault for the handled event.

mstahv:
closes #1363


mshabarov (changes requested):
Looks good to me.
One thing I would like to include here is the same API for `@DomEvent`:
```java
@DomEvent(value = "change", stopPropagation = true, preventDefault = true)
public class ChangeEvent
        extends ComponentEvent<TextField> {
}
```
These annotation parameters can be handled `ComponentEventBus::addDomTrigger` and just passed to `DomListenerRegistration` via new methods:
```java
if (annotation.stopPropagation()) {
    registration.stopPropagation();
}
if (annotation.preventDefault()) {
    registration.preventDefault();
}
```
Because having these methods on Element API level and not on Component API level looks a bit off.

mshabarov (approved):

PR #17551, opened 2023-09-05
fix: Use released commons-fileupload2 (#17548)

### LINKED ISSUE #17269: Consider replacing "shaded" version of Apache Commons Fileupload
### Describe your motivation

Now that (finally) Apache Commons Fileupload 2 is released with Jakarta support - the shaded version based on an older snapshot version can be replaced.

https://central.sonatype.com/artifact/org.apache.commons/commons-fileupload2-jakarta/2.0.0-M1

No custom / shaded version of Apache Commmons Fileupload is in use.

caalador:
Switch to using the released
commons-fileupload2-jakarta instead
of the old self maintained
copy flow-commons-upload

Closes #17269


flow-server/pom.xml
```diff
@@ -78,9 +78,9 @@
         </dependency>
 
         <dependency>
-            <groupId>com.vaadin</groupId>
-            <artifactId>flow-commons-upload</artifactId>
-            <version>${project.version}</version>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-fileupload2-jakarta</artifactId>
+            <version>2.0.0-M1</version>
```
mshabarov, line 83: I would maybe wait until this will become a GA version, not a milestone. For upcoming 24.2 this is fine, but for already released 24.1 not sure how ready it is for production.

caalador: certainly, but at least we get the kinks ironed out now so it's easier then.

PR #17559, opened 2023-09-05
chore: Collect usage statistics for Spring-Boot

mshabarov:


czp13 (dismissed):
It is a nice improvement with the usage of the `SpringUtil` class to mark `SpringBoot` entries as well.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
```diff
@@ -56,8 +57,13 @@ public SpringInstantiator(VaadinService service,
         this.context = context;
 
         UsageStatistics.markAsUsed("flow/SpringInstantiator", null);
-        UsageStatistics.markAsUsed("SpringFramework", Optional
-                .ofNullable(SpringVersion.getVersion()).orElse("unknown"));
+        if (SpringUtil.isSpringBoot()) {
+            UsageStatistics.markAsUsed("SpringBoot",
+                    SpringBootVersion.getVersion());
+        } else {
+            UsageStatistics.markAsUsed("SpringFramework", Optional
```
Legioth, line 64: We should always mark `SpringFramework` as  used so that we capture the `SpringVersion.getVersion()` value also for Spring Boot users.

mshabarov: Good point, will fix

mshabarov: done

Legioth (approved):

PR #17560, opened 2023-09-05
fix: Allow platform dependencies to be linked

Artur-:
Do not fail if a managed version is linked to a local copy and thus the version cannot be parsed

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
@@ -328,24 +328,34 @@ public class TaskUpdatePackages extends NodeUpdater {
         assert vaadinDeps != null : "vaadin{ dependencies { } } should exist";
         assert packageJsonDeps != null : "dependencies { } should exist";
 
-        FrontendVersion packageJsonVersion;
+        FrontendVersion packageJsonVersion = null, vaadinDepsVersion = null;
         try {
-            packageJsonVersion = new FrontendVersion(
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -327,9 +327,18 @@ protected static boolean pinPlatformDependency(JsonObject packageJson,
         // packages exist at this point
         assert vaadinDeps != null : "vaadin{ dependencies { } } should exist";
         assert packageJsonDeps != null : "dependencies { } should exist";
+
+        FrontendVersion packageJsonVersion;
+        try {
+            packageJsonVersion = new FrontendVersion(
+                    packageJsonDeps.getString(pkg));
+        } catch (Exception e) {
```
mshabarov, line 335: Would it be more correct to scope the caught exception to only `NumberFormatException` ?

mshabarov (changes requested):
Also would need a unit test.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
Artur-: What should we do on another exception?

mshabarov: We don't expect other exceptions in the `FrontendVersion` CTOR, if it throws NPE or other runtime exception, then it throws.

mshabarov (approved):

PR #17572, opened 2023-09-07
fix: support downgrading dependencies based on versions.json

platosha:
For dependencies without custom versions in `package.json` (when `dependencies` and `vaadin.dependencies` agree), currently a pinned version from `versions.json` is not applied if it is for some reason older than the one specified in `package.json`. This effectively prevents downgrading npm package versions by changing the platform dependency version.

This change removes the downgrade prevention check.

Context: to speed up the upcoming release, Hilla platform 2.2 reverted to using stable React and Vaadin components 24.1.x before the beta1 release after using 24.2.0.alphaX during alpha releases. At the moment upgrading a Hilla application from 2.2.0.alpha10 to beta1 does not downgrade the component dependencies back to 24.1, while this is expected.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
@@ -82,9 +82,6 @@ class VersionsJsonFilter {
                             versions.getString(key));
                 }
                 json.put(key, userManagedDependencies.getString(key));
-            } else if (vaadinDepsVersion != null
-                    && vaadinDepsVersion.isNewerThan(version)) {
-                json.put(key, vaadinVersions.getString(key));
             } else {
                 json.put(key, versions.getString(key));
             }
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
@@ -71,9 +71,6 @@ class VersionsJsonFilter {
             final FrontendVersion userManagedVersion = FrontendUtils
                     .getPackageVersionFromJson(userManagedDependencies, key,
                             "/package.json -> { dependencies }");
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
```diff
@@ -82,9 +82,6 @@ JsonObject getFilteredVersions(JsonObject versions, String versionOrigin) {
                             versions.getString(key));
                 }
                 json.put(key, userManagedDependencies.getString(key));
-            } else if (vaadinDepsVersion != null
```
mshabarov, line 85: Code block 
```
            final FrontendVersion vaadinDepsVersion = FrontendUtils
                    .getPackageVersionFromJson(vaadinVersions, key,
                            "/package.json -> { vaadin { dependencies }}");
```
is not needed now.

mshabarov (changes requested):
This change looks fine to me.

If I understand the algorithm properly, the previous approach used the platform pinned dependency if it is newer and closes the eyes on the fact that it comes from old platform.

This in my opinion is not correct. This adds a mess into dependencies. And maybe that's at least particularly why vaadin dance is needed.

Whenever you change platform version, you expect that your project will use all the versions provided by this concrete platform version, unless you override it by yourself.

I can't imagine at the moment in what circumstances this patch could break existing applications. Anyway, vaadin dance to the rescue while we are figuring out the use case where this patch breaks an app.

mshabarov:
`runPnpmInstall_userVersionNewerThanPinned_installedOverlayVersionIsNotSpecifiedByPlatform` exactly checks that the dependency version in package.json is newer than pinned platform version and thus not touched. The expectation should be changes to opposite.

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonFilter.java
platosha: removed

mshabarov (approved):

PR #17575, opened 2023-09-08
Spring Boot DevTools reload caching

tepi:
Adds some cached data for use in case of Spring Boot DevTools reload.

- CustomResourceLoader: added a few packages to default never-scan list
- DevModeServletContextListener: Dynamically generates a white list for dev mode package scanner
- RouteServletContextListener: Dynamically generates a white list for route package scanner
- LookupInitializerListener: Cache found classes
- VaadinAppShellContextListener: Cache found classes
- CustomResourceLoader: Cache valid/skipped status of resources for later use

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
@@ -477,10 +477,6 @@ public class VaadinServletContextInitializer
                 throw new RuntimeException(
                         "Unable to initialize Vaadin DevModeHandler", e);
             }
-            // to make sure the user knows the application is ready, show
-            // notification to the user
-            ServletDeployer.logAppStartupToConsole(event.getServletContext(),
-                    true);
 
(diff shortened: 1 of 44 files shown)


tltv:
Looks good. It would be good to have system property to force disable/enable `devModeCachingEnabled`.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -572,6 +659,41 @@ public void failFastContextInitialized(ServletContextEvent event) {
      */
     public VaadinServletContextInitializer(ApplicationContext context) {
         appContext = context;
+
+        try {
+            Class.forName(
+                    "org.springframework.boot.devtools.livereload.LiveReloadServer");
+            if (appContext instanceof ConfigurableApplicationContext) {
+                String devModeCachingProperty = appContext.getEnvironment()
```
mshabarov, line 667: There is no restrictions in the code to use it in production, or I missed something?
Would be good to make the caching disabled in production mode always.

PR #17576, opened 2023-09-08
fix: forward to should update url

### LINKED ISSUE #15876: BeforeEnterEvent.forwardTo does not update URL
### Description of the bug

Forwarding to a "sub route" does not update the browser URL.

I created a little example to demonstrate the problem. We start at /routing_test_1 where a button is displayed. The button navigates to /routing_test_2. This view can have several "sub routes". In this case I want to forward to routing_test_2/default_sub_route. The problem is that the browser URL is not updated in this case.

This would not be that bad because I could still use the History API to change the URL manually. This however introduces another problem: PreserveOnRefresh does not work anymore because AbstractNavigationStateRenderer associated the components in the PreservedComponentCache with the location /routing_test_2. 

### Expected behavior

BeforeEnterEvent.forwardTo should update the browser URL in case that you forward to a sub route. In this case the PreservedComponentCache of AbstractNavigationStateRenderer should also be associated with the location of the forwarding, so that PreserveOnRefresh works.

```
@PreserveOnRefresh
public class RoutingTestParentView extends Div implements RouterLayout{

}
```

```
@Route(value = "routing_test_1", layout = RoutingTestParentView.class)
public class RoutingTest1 extends Div{

	public RoutingTest1() {
		Button button = new Button("navigate");
		button.addClickListener(e -> {
			UI.getCurrent().navigate("routing_test_2");
		});
		add(button);
	}
	
}
```

```
@Route(value = "routing_test_2/:path*", layout = RoutingTestParentView.class)
public class RoutingTest2 extends Div implements BeforeEnterObserver, AfterNavigationObserver{

	public RoutingTest2() {
		Button button = new Button("show_window");
		button.addClickListener(e -> {
			Dialog d = new Dialog();
			d.setHeight("100px");
			d.setWidth("100px");
			d.setOpened(true);
		});
		add(button);
	}
	
	@Override
		if(event.getLocation().getSegments().size() == 1) {
			event.forwardTo("routing_test_2/default_sub_route");
		}
	}

	@Override
	public void afterNavigation(AfterNavigationEvent event) {
		if(event.getLocation().getSegments().size() == 1) {
			//what I would like to do this so that the URL is correct, but this messes with PreserveOnRefresh
			//UI.getCurrent().getPage().getHistory().pushState(null, "routing_test_2/default_sub_route");
		}
	}

}
```

Steps to reproduce wrong URL:
-Go to /routing_test_1
-Click the button
-Result: Browser URL is /routing_test_2

Steps to reproduce the mentioned PreserveOnRefresh problem when using History API as workaround:
-Uncomment the line in RoutingTest2.afterNavigation
-Go to /routing_test_1
-Click the button
-Result: Browser URL is /routing_test_2/default_sub_route
-Click the button
-Dialog is shown
-Press F5
-Dialog is gone

### Versions

- Vaadin / Flow version: 23.3.2
- Java version: 17
- OS version: Windows 10

caalador:
Forward to should run even
when targeting same target
but with another URL

Fixes #15876

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
@@ -659,17 +659,17 @@ public abstract class AbstractNavigationStateRenderer
             return Optional.of(forwardToExternalUrl(event, beforeEvent));
         }
         if (beforeEvent.hasForwardTarget()
-                && !(isSameNavigationState(beforeEvent.getForwardTargetType(),
+                && (!isSameNavigationState(beforeEvent.getForwardTargetType(),
                         beforeEvent.getForwardTargetRouteParameters())
-                        || navigationState.getResolvedPath()
+                        || !navigationState.getResolvedPath()
                                 .equals(beforeEvent.getForwardUrl()))) {
             return Optional.of(forward(event, beforeEvent));
         }
 
         if (beforeEvent.hasRerouteTarget()
-                && !(isSameNavigationState(beforeEvent.getRerouteTargetType(),
+                && (!isSameNavigationState(beforeEvent.getRerouteTargetType(),
                         beforeEvent.getRerouteTargetRouteParameters())
-                        || navigationState.getResolvedPath()
+                        || !navigationState.getResolvedPath()
                                 .equals(beforeEvent.getRerouteUrl()))) {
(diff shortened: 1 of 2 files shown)


mshabarov (changes requested):
Looks good to me, but needs a test, at least for the case shown in the "Minimal reproducible example" of the original ticket.

caalador:
Added test.

mshabarov (approved):

PR #17580, opened 2023-09-09
chore: Remove sysout message

Artur-:
The message does not really offer much benefit. It is not logged with a logger for some reason and in some cases it is wrong. In most cases the browser will automatically be opened with the correct URL anyway.

With Hilla apps, it is also logged and just looks strange


caalador (dismissed):

caalador:
Seems like spring is using the log out method  `vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java:[482,28]`

czp13 (approved):

caalador (approved):

PR #17587, opened 2023-09-11
Fix: check for null resolvedPath

caalador:
Check that resolvedPath is
not null.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
@@ -658,19 +658,21 @@ public abstract class AbstractNavigationStateRenderer
         if (beforeEvent.hasExternalForwardUrl()) {
             return Optional.of(forwardToExternalUrl(event, beforeEvent));
         }
-        if (beforeEvent.hasForwardTarget()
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
```diff
@@ -658,19 +658,21 @@ protected Optional<Integer> handleTriggeredBeforeEvent(
...
         }
-        if (beforeEvent.hasForwardTarget()
-                && (!isSameNavigationState(beforeEvent.getForwardTargetType(),
-                        beforeEvent.getForwardTargetRouteParameters())
-                        || !navigationState.getResolvedPath()
-                                .equals(beforeEvent.getForwardUrl()))) {
+        if (beforeEvent.hasForwardTarget() && (!isSameNavigationState(
+                beforeEvent.getForwardTargetType(),
+                beforeEvent.getForwardTargetRouteParameters())
+                || !(navigationState.getResolvedPath() != null
```
mshabarov, line 664: Having an empty resolved path in this context looks a bit weird, why would it happen?

caalador:
Tests. The main issue that tests can have null resolved due to mocking.

mshabarov (approved):

PR #17591, opened 2023-09-11
perf: improve route registration performance

tltv:
Reduces time it takes to populate a lot of  registered routes in `AbstractRouteRegistry` by avoiding checking same path multiple times to figure out route aliases. One path can be mapped to only one `RouteTarget` based on `ConfiguredRoutes#routeMap`.


mshabarov (approved):
I would maybe refactor this long chain of methods to something like:
```java
while(!routePathMap.isEmpty) {
   // get entries by target
   // add aliases for these entries
   // remove these entries from routePathMap
}
```
But this is just my opinion, might be not a best way.

PR #17604, opened 2023-09-12
perf: optimize SpringLookup lookup time by caching

tltv:
Cache `RoutePathProvider` in `SpringLookup` for faster lookup time especially for `RouteUtil#resolve` method, but also for all default (and subclasses of default) implementations of services defined in `LookupInitializer.getDefaultImplementations()`.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
@@ -284,11 +284,7 @@ public class RouteUtil {
         RoutePathProvider provider = null;
         Lookup lookup = context.getAttribute(Lookup.class);
         if (lookup != null) {
-            provider = context.getAttribute(RoutePathProvider.class);
-            if (provider == null) {
-                provider = lookup.lookup(RoutePathProvider.class);
-                context.setAttribute(RoutePathProvider.class, provider);
-            }
+            provider = lookup.lookup(RoutePathProvider.class);
             assert provider != null;
         }
         if (provider == null) {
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
```diff
@@ -284,7 +284,11 @@ public static String resolve(VaadinContext context, Class<?> component) {
         RoutePathProvider provider = null;
         Lookup lookup = context.getAttribute(Lookup.class);
         if (lookup != null) {
-            provider = lookup.lookup(RoutePathProvider.class);
+            provider = context.getAttribute(RoutePathProvider.class);
```
mshabarov, line 287: This by-passes the Lookup API and, thus, most likely will have issues in other environments, like OSGi.
This isn't a good design as well.
If the lookup takes much time to get the service, maybe we can fix it in the lookup implementation?

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
tltv: It probably would just use lookup anyway in OSGi environment and would not speed up anything there.
If this can be updated in Lookup, that would be better indeed. I don't know the effort yet but I'll check that.

tltv: I'll prepare now update that adds caching in `SpringLookup`. It will also cache a bit more than just `RoutePathProvider` based on `LookupInitializer.getDefaultImplementations()`.

mshabarov (approved):

PR #17612, opened 2023-09-12
Generate at build time the metadata for theme editor tests

brunovianarezende:
## Description

The theme editor tests need files describing what are the properties available for the user. This PR creates such files at build time, in the vaadin-dev-server project.

## Type of change

- [ ] Bugfix
- [ X ] Feature

## Checklist


vaadin-dev-server/export-metadata-script.ts
sissbruecker, line 1: Not working on theme editor ATM, but still commenting because this looks very perplexing:
1. As this just copies files between Maven modules, why not use the original files in the first place? Is there a reason why tests can not read from the dev-server module but the dev-server build is allowed to copy files into the flow-tests module?
2. If they need to be copied, why make the dev-server build copy them instead of whatever is running the tests? It's a bit unexpected that just building the dev server updates files in a different Maven module.
3. If the files are only needed for tests, why add them to version control?

brunovianarezende: > Not working on theme editor ATM, but still commenting because this looks very perplexing:
> 
>     1. As this just copies files between Maven modules, why not use the original files in the first place? Is there a reason why tests can not read from the dev-server module but the dev-server build is allowed to copy files into the flow-tests module?
> 

It doesn't copy the files. The files are in ts format theme-editor tests expects the metadata as txt file. You can see the the TS files here: https://github.com/vaadin/flow/tree/main/vaadin-dev-server/frontend/theme-editor/metadata/components. For example: https://github.com/vaadin/flow/blob/main/vaadin-dev-server/frontend/theme-editor/metadata/components/vaadin-accordion-heading.ts.

>     2. If they need to be copied, why make the dev-server build copy them instead of whatever is running the tests? It's a bit unexpected that just building the dev server updates files in a different Maven module.
> 
>     3. If the files are only needed for tests, why add them to version control?

Right now, just because we already have a vaadin-button.txt file in the version control, so I'm just following what we have. But, I'm not sure if we should keep them in the repository.

sissbruecker: I see, it seems it's necessary then. Still, maybe you can run execute the conversion task as part of running the tests that need these files. At that point it shouldn't be necessary to have them in VCS, and there is less risk of them getting out of sync.

MarcinVaadin:
If they are generated at runtime I would not commit them into git. I would put them inside target/test-resources or similar. That way we do not have to push changes each time source TS metadata change.

They should have .json extension.

MarcinVaadin (changes requested):
If they are generated at runtime I would not commit them into git. I would put them inside target/test-resources or similar. That way we do not have to push changes each time source TS metadata change.

They should have .json extension.

brunovianarezende:
> If they are generated at runtime I would not commit them into git. I would put them inside target/test-resources or similar. That way we do not have to push changes each time source TS metadata change.
> 

I agree. I will try to move them to target/somewhere and see if I can get them, as you suggest.

> They should have .json extension.

ok

vaadin-dev-server/export-metadata-script.ts
szolo: Hello guys!

@sissbruecker this conversion was necessarily in order to extract the metadata information from files. These descriptors include much more information which can not be processed during tests execution.

@brunovianarezende awesome work in finding how to execute the script! Thanks a lot, highly appreciated :) We talked today with @MarcinVaadin that we shouldn't commit the transformed files because they will be anyway created and executed during build, hence removal is a good idea.

In case you're testing tests locally it is always possible to execute transformation manually.

szolo:
@brunovianarezende your PR looks awesome! Just remove the metadata information, make sure that it is executed during build, rebase and we are good to go!

Thanks again for a hard and productive work :)

brunovianarezende:
> @brunovianarezende your PR looks awesome! Just remove the metadata information, make sure that it is executed during build, rebase and we are good to go!
> 
> Thanks again for a hard and productive work :)

thanks! I will see if I can do as suggested by Marcin and then update this PR.

brunovianarezende:
In the last commit I:

1. removed the txt files from the repository
2. moved the created metadata files to vaadin-deve-server/src/main/resources/META-INF/metadata
3. added vaadin-deve-server/src/main/resources/META-INF/metadata to .gitignore, so the files won't be saved in the repository
4. changed the metadata files extension from '.txt' to '.json'
5. provided, in vaadin-dev-server, a utility class to read the metadata for a given component
6. changed test-theme-editor to use this utility class

szolo (approved):

MarcinVaadin (approved):

PR #17636, opened 2023-09-18
test: fix open url to use correct servlet

mcollovati:
The CSS loadin test is using a development mode servlet instead of the production mode.
This change fixes the url provided to the web driver in order to use the expected Vaadin servlet.

### DIFF
--- a/flow-tests/test-themes/src/test/java/com/vaadin/flow/uitest/ui/theme/CssLoadingIT.java
+++ b/flow-tests/test-themes/src/test/java/com/vaadin/flow/uitest/ui/theme/CssLoadingIT.java
@@ -68,4 +68,12 @@ public class CssLoadingIT extends ChromeBrowserTest {
         Assert.assertEquals(expectedMargin,
                 $(ParagraphElement.class).id(elementId).getCssValue("margin"));
     }
+
+    @Override
+    protected String getTestPath() {
+        String path = super.getTestPath();
+        String view = "view/";
+        return path.replace(view, "path/");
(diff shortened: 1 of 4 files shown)


flow-tests/test-themes/src/test/java/com/vaadin/flow/uitest/ui/theme/CssLoadingIT.java
```diff
@@ -68,4 +68,12 @@ private void assertStylesOverride(String elementId, String expectedColor,
         Assert.assertEquals(expectedMargin,
                 $(ParagraphElement.class).id(elementId).getCssValue("margin"));
     }
+
+    @Override
+    protected String getTestPath() {
```
mshabarov, line 73: It's easy to miss this overload for a new IT tests in this module. Would it be better to make a new base class for IT test
```java
public class ThemesBaseTest extends ChromeBrowserTest {
    
    @Override
    protected String getTestPath() {
        String path = super.getTestPath();
        String view = "view/";
        return path.replace(view, "path/");
    }
}
```
and extend existing tests from it?

mcollovati: Good point. I'll add the base class

mcollovati: done

mshabarov (approved):

PR #17644, opened 2023-09-18
Caching: Cache all com.vaadin.flow class instances

tepi:
Since all classes from packages under `com.vaadin.flow` come from JARs, the class instances can be cached directly.


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -526,8 +526,15 @@ public void failFastContextInitialized(ServletContextEvent event)
...
             if (devModeCachingEnabled) {
+                classes.addAll(ReloadCache.vaadinClasses);
                 ReloadCache.dynamicWhiteList = classes.stream()
-                        .map(Class::getPackageName).collect(Collectors.toSet());
+                        .map(Class::getPackageName)
+                        .filter(packageName -> !packageName
+                                .startsWith("com.vaadin.flow"))
+                        .collect(Collectors.toSet());
+                ReloadCache.vaadinClasses = classes.stream().filter(
+                        c -> c.getPackageName().startsWith("com.vaadin.flow"))
```
tarekoraby, line 536: Shouldn't we cache everything under `com.vaadin`, `org.vaadin`, and `dev.hilla`?

mshabarov, line 536: I see this as a nice improvement for reload time.
One use case where this potentially might cause problems is development of Flow add-ons with `com.vaadin.flow` namespace. I suppose that changes in classes that are in the packages `com.vaadin.flow.*` would be cached and not rescanned, resulting the changes not applied on the page. Or am I wrong and they would be spotted by Flow anyway as Spring will fire an event with changed classes?

tepi: Yes, that is a problem and that is also why one IT is currently failing with this change. The test is in `com.vaadin.flow` namespace and does a reload-causing change. This patch needs to be changed so that it can figure out which classes really are in JARs, instead of relying just on package name.

mshabarov: Good idea! Let's do it.

tarekoraby: > The test is in com.vaadin.flow namespace and does a reload-causing change

Wouldn't the solution then be to disable the devmode caching in the Flow repo, and only enable it in DevModeClassCacheIT.java?

tepi: Changed to cache classes in JARs under these three packages

tepi: I would not like to disable the caching since running our (cache-unrelated) tests with caching enabled can also point out possible issues in the caching.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -983,6 +990,19 @@ private Resource[] collectResources(String locationPattern)
                         int index = path.lastIndexOf(".jar!/");
                         if (index >= 0) {
                             String relativePath = path.substring(index + 6);
+                            if (devModeCachingEnabled
+                                    && relativePath.endsWith(".class")) {
+                                // Stores names of vaadin classes from JARs
+                                String className = relativePath.replaceAll(
```
tltv, line 996: Consider using `replace()` instead of `replaceAll()` if usage of slower regular expression is not necessary here.

tltv (approved):

tltv (dismissed):

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -990,6 +990,19 @@ private Resource[] collectResources(String locationPattern)
...
+                                        String.valueOf(File.separatorChar), ".")
+                                        .replace(".class", "");
+                                if (className.startsWith("com.vaadin.flow")
+                                        || className
+                                                .startsWith("org.vaadin.flow")
```
tarekoraby, line 1001: Why not `com.vaadin` and `org.vaadin` to include all Vaadin products, including Kits for example?

tepi: Done

tarekoraby:
Can we easily and reliably know what comes from jars?

tarekoraby:
In any case, should we merge this PR, and improve on it separately?

tltv (approved):

mshabarov:
> Why don't we cache everything that comes from jars?

Good idea, and we will discuss it within the team. I'd like to merge this patch anyway now, to be able to see the impact on the reload time on the chart.

PR #17647, opened 2023-09-19
fix: router-link postpone navigation

### LINKED ISSUE #17528: RouterLink url behaviour when using BeforeLeaveEvent.postpone
### Description of the bug

When postponing a navigation with the BeforeLeaveListeners `postpone()` method the url gets lost.

1. Open the example

Working example (`UI.getCurrent().navigate()`):
3. Click on Report 4
4. Click on the button and watch the url inside your browser
5. It should stay the same

Broken example (`RouterLink`): 
3. Click on Report 5
4. Click the link with an random uuid
5. Click yes
6. Repeat from 4 and watch the uuid disappear from the url again and again

### Expected behavior

The url should stay the same as before and not navigate to the parent.

Implementation without `RouterLink`:
https://github.com/FelixGaebler/vaadin-bugs/tree/master/src/main/java/de/dataport/bugs/views/report4

Implementation with `RouterLink`:
https://github.com/FelixGaebler/vaadin-bugs/tree/master/src/main/java/de/dataport/bugs/views/report5

### Versions

- Vaadin / Flow version: 24.1.5
- Java version: 17
- OS version: Ubuntu 22

caalador:
Fixes issue with update timing
when postponing a router-link
as the navigation goes through
the vaadin-router.

Fixes #17528

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -54,6 +54,7 @@ import com.vaadin.flow.internal.nodefeature.PollConfigurationMap;
 import com.vaadin.flow.internal.nodefeature.ReconnectDialogConfigurationMap;
 import com.vaadin.flow.router.AfterNavigationListener;
 import com.vaadin.flow.router.BeforeEnterListener;
+import com.vaadin.flow.router.BeforeLeaveEvent;
 import com.vaadin.flow.router.BeforeLeaveListener;
 import com.vaadin.flow.router.ErrorNavigationEvent;
 import com.vaadin.flow.router.ErrorParameter;
@@ -1721,7 +1722,12 @@ public class UI extends Component
             navigateToClient(getForwardToClientUrl());
             acknowledgeClient();
         } else if (isPostponed()) {
-            cancelClient();
+            BeforeLeaveEvent.ContinueNavigationAction continueNavigationAction = getInternals()
+                    .getContinueNavigationAction();
+            continueNavigationAction
+                    .setProceedCallback(this::acknowledgeClient);
+            continueNavigationAction.setCancelCallback(this::cancelClient);
+            serverPaused();
         } else {
             acknowledgeClient();
         }
@@ -1761,6 +1767,10 @@ public class UI extends Component
 
     private void cancelClient() {
         serverConnected(true);
+    }
+
+    private void serverPaused() {
+        wrapperElement.executeJs("this.serverPaused()");
     }
 
     private void serverConnected(boolean cancel) {
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -1722,11 +1722,6 @@ public class UI extends Component
             navigateToClient(getForwardToClientUrl());
             acknowledgeClient();
         } else if (isPostponed()) {
-            BeforeLeaveEvent.ContinueNavigationAction continueNavigationAction = getInternals()
-                    .getContinueNavigationAction();
-            continueNavigationAction
-                    .setProceedCallback(this::acknowledgeClient);
-            continueNavigationAction.setCancelCallback(this::cancelClient);
             serverPaused();
         } else {
             acknowledgeClient();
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1721,7 +1722,12 @@ public void connectClient(String flowRoutePath, String flowRouteQuery,
             navigateToClient(getForwardToClientUrl());
             acknowledgeClient();
         } else if (isPostponed()) {
-            cancelClient();
+            BeforeLeaveEvent.ContinueNavigationAction continueNavigationAction = getInternals()
+                    .getContinueNavigationAction();
+            continueNavigationAction
+                    .setProceedCallback(this::acknowledgeClient);
+            continueNavigationAction.setCancelCallback(this::cancelClient);
```
mcollovati, line 1729: Do I get it right that potential callbacks set by the developer on `ContinuationNavigationAction` are always overwritten by the UI callbacks?
If so, we should probably document it in `ContinuationNavigationAction` javadocs.

Or would it make sense to compose existing callbacks with the on from UI?

flow-server/src/main/java/com/vaadin/flow/router/BeforeLeaveEvent.java
```diff
@@ -77,10 +82,33 @@ public void proceed() {
...
+                    event = new NavigationEvent(event.getSource(),
+                            event.getLocation(), event.getUI(),
+                            NavigationTrigger.PROGRAMMATIC);
+                }
                 handler.handle(event);
                 setReferences(null, null);
             }
         }
+
+        public void cancel() {
```
mcollovati, line 101: Needs javadocs

flow-server/src/main/java/com/vaadin/flow/router/BeforeLeaveEvent.java
```diff
@@ -77,10 +82,33 @@ public void proceed() {
...
+            cancelCallback.execute();
+        }
+
+        public void setProceedCallback(Command proceedCallback) {
```
mcollovati, line 105: Needs javadocs

flow-server/src/main/java/com/vaadin/flow/router/BeforeLeaveEvent.java
```diff
@@ -77,10 +82,33 @@ public void proceed() {
...
+            this.proceedCallback = proceedCallback;
+        }
+
+        public void setCancelCallback(Command cancelCallback) {
```
mcollovati, line 109: Needs javadocs

flow-server/src/main/java/com/vaadin/flow/component/UI.java
caalador: The proceed and cancel callbacks are only for internal use as they are meant to notify the client and fulfill the waiting promise. 
Removed methods and use direct exec calls instead so there can be no confusion on the usage.

flow-server/src/main/java/com/vaadin/flow/router/BeforeLeaveEvent.java
```diff
@@ -40,6 +41,10 @@ public class ContinueNavigationAction implements Serializable {
 
         private NavigationHandler handler = null;
         private NavigationEvent event = null;
+        private Command proceedCallback = () -> {
+        };
+        private Command cancelCallback = () -> {
+        };
```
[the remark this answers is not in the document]
caalador: Removed the empty no op commands

flow-server/src/main/java/com/vaadin/flow/router/BeforeLeaveEvent.java
caalador: removed method

flow-server/src/main/java/com/vaadin/flow/router/BeforeLeaveEvent.java
caalador: removed method

flow-server/src/main/java/com/vaadin/flow/router/BeforeLeaveEvent.java
caalador: added javadoc

mcollovati (approved):

PR #17661, opened 2023-09-20
fix: Do not override user npm dependency version

### LINKED ISSUE #17654: @hilla/react-components is automatically downgraded after two restarts
### Description of the bug 
```
hilla init hillatest
cd hillatest
mvn
npm i @hilla/react-components@2.2.0-alpha16
mvn
# React components 2.2.0-alpha16 is used
mvn
# React components 2.1.7 are used again
```

When running `mvn` after installing `@hilla/react-components@2.2.0-alpha16` then managed dependency version is also updated, i.e. `vaadin.dependencies.@hilla/react-components` is set to 2.2.0-alpha16. This should not happen

Artur-:
Fixes #17654


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -348,6 +348,12 @@ protected static boolean pinPlatformDependency(JsonObject packageJson,
             // would be but probably it should be pinned like any other version
         }
 
+        if ((vaadinDepsVersion != null && packageJsonVersion != null)
+                && !vaadinDepsVersion.equals(packageJsonVersion)) {
+            // The user has overridden the version, use that
+            return false;
```
knoobie, line 354: Does this interfere with people downgrading? e.g. developer upgrades from 24.1 to 24.2 and notices errors, now he downgrades to 24.1 again - now I would expect that all 24.1 dependencies are used again and not those from 24.2.

Artur-: As long as the version in the `dependencies` section and the `vaadin/dependencies` match, the ´dependencies` section should be updated. So as long as a user never touches the `dependencies`, the versions are always supposed to be updated. 

What this changes is that when the user HAS changed the dependencies section manually, then the version should never be touched

caalador (approved):

Artur-:
I think the problem is that the pnpmfile for pnpm is generated based on versions.json so it does not really work properly anymore. The best solution would be #17667

caalador (approved):

PR #17667, opened 2023-09-20
refactor: Use overrides also for pnpm

Artur-:
Pnpm v5.10.1 added support for overrides, the current version is compatible with the npm syntax we use


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -93,18 +93,6 @@ public void execute() {
 
             if (modified || npmVersionLockingUpdated) {
                 writePackageFile(packageJson);
-
-                if (enablePnpm) {
```
caalador, line 97: With the removal of this the test `AbstractNodeUpdatePackagesTest::pnpmIsInUse_packageJsonModified_removePnpmLock` is wrong and should also be deleted

caalador:
This has a bigger issue with the locking of undefined dependencies that worked when using the pnpm file that locked all instances to the defined version.
Mainly this can be seen in the slow pnpm tests where `vaadin-overlay` doesn't get pinned to a version. 
Only having an override would not pin the version for the transitive package.

Artur-:
Do you have a simple example of this? Does it work differently than npm?

caalador:
TaskRunPnpmInstallTest::runPnpmInstall_overlayVersionIsPinnedViaPlatform_installedOverlayVersionIsSpecifiedByPlatform

Basically that test so in simplicity it has so far locked the overlay version, but after this change it doesn't and uses the dialog marked overlay version. I think it works basically the same as npm, but for pnpm it is a big change in how it handles itself.

Artur-:
But we want npm and pnpm to work the same?

caalador:
And in that case it's fine, but then this is a breaking change for anyone using pnpm and expecting the previous results.

Artur-:
If this is indeed so that pnpm starts to work like npm, then it sounds fine and that it would be easier for everybody in the long run

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCleanFrontendFiles.java
```diff
@@ -50,7 +50,7 @@ public class TaskCleanFrontendFiles implements FallibleCommand {
     private List<String> generatedFiles = List.of(NODE_MODULES,
             Constants.PACKAGE_JSON, Constants.PACKAGE_LOCK_JSON,
             Constants.PACKAGE_LOCK_YAML, TaskGenerateTsConfig.TSCONFIG_JSON,
-            TaskGenerateTsDefinitions.TS_DEFINITIONS, ".pnpmfile.cjs", ".npmrc",
```
caalador, line 53: Should this be left here so that when you update and already have the file it would get cleaned up as I think it will just by being there manage pnpm

Artur-: Sure, added it back

caalador: I guess the main question is should we have a auto cleanup of the pnpmfile when running a project.

caalador (approved):

PR #17670, opened 2023-09-20
test: add optional assert with reload time threshold

tltv:
Adds `vaadin.test.reload-time-assert-threshold` system property for Vaadin reload time tests. When value is set, tests will fail if reload time result goes outside the threshold.


flow-tests/vaadin-spring-tests/test-spring-boot-reload-time/src/test/java/org/vaadin/example/SpringDevToolsReloadViewIT.java
```diff
@@ -45,6 +45,13 @@ public void testSpringBootReloadTime_withNativeButton() {
         System.out.printf(
                 "##teamcity[buildStatisticValue key='%s,nativebutton,spring-boot-devtools-reload-time' value='%s']%n",
                 getVaadinMajorMinorVersion(), result);
+
+        if (getReloadTimeAssertThreshold() != null) {
+            Assert.assertTrue(Double
```
mshabarov, line 50: This assert checks the same `result` value as `optionalAssertByReloadThreshold(result);` does at line 54. So either of them is redundant I suppose.

mshabarov (changes requested):
One small comment otherwise looks good!
Can we make this threshold less than expected for one run to test this functionality?

tltv:
> Can we make this threshold less than expected for one run to test this functionality?

Yes, we can try that out.

mshabarov (approved):

Issue #17669, opened 2023-09-20
More customizable Spring configuration


oliveryasuna:
### Describe your motivation

It would be useful if the Spring configuration was more customizable.

My use case is to add a similar pattern to `@PageTitle`/`HasDynamicTitle` for the header of a view. The cleanest way, in my opinion, to validate that developers use `@Header` and `HasDynamicHeader` properly, would be to do the same check in `AbstractRouteRegistryInitializer`. In a Spring project, `VaadinServletContextInitializer` is a bean defined in `SpringBootAutoConfiguration`.

### Describe the solution you'd like

I think that this is a long-term project, but a good place to start is to make the beans defined in `SpringBootAutoConfiguration` conditional.

For example,

```java
@Bean
public ServletContextInitializer contextInitializer() {
    return new VaadinServletContextInitializer(context);
}
```

becomes,

```java
@Bean
@ConditionalOnMissingBean(ServletContextInitializer.class)
public ServletContextInitializer contextInitializer() {
    return new VaadinServletContextInitializer(context);
}
```

### Describe alternatives you've considered

There are definitely alternatives to implement my use case, but I still think the Spring configuration should be easier to customize. Right now, I would have to enable bean definition overriding and thankfully `VaadinServletContextInitializer` is not `final`.

### Additional context

The conditional-bean-in-auto-configuration pattern is common is most Spring auto configurations.

Looking into the future, it would also be helpful the implement the customizer pattern.

I'll open a PR and link it here.

oliveryasuna:
https://github.com/vaadin/flow/pull/17671

PR #17671, opened 2023-09-20
feat: Conditional spring beans

oliveryasuna:
## Description

See details of issue...

Fixes https://github.com/vaadin/flow/issues/17669

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-spring/pom.xml
+++ b/vaadin-spring/pom.xml
@@ -51,6 +51,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-testbench</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
         <!-- Spring -->
         <dependency>
             <groupId>org.springframework.boot</groupId>
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
@@ -23,6 +23,7 @@ import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.web.servlet.WebMvcAutoConfiguration;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.boot.web.servlet.ServletContextInitializer;
@@ -58,6 +59,7 @@ public class SpringBootAutoConfiguration {
     private WebApplicationContext context;
 
     @Bean
+    @ConditionalOnMissingBean(VaadinBeanFactoryInitializationAotProcessor.class)
     static VaadinBeanFactoryInitializationAotProcessor flowBeanFactoryInitializationAotProcessor() {
         return new VaadinBeanFactoryInitializationAotProcessor();
     }
@@ -68,6 +70,7 @@ public class SpringBootAutoConfiguration {
      * @return a custom ServletContextInitializer instance
      */
     @Bean
+    @ConditionalOnMissingBean(ServletContextInitializer.class)
     public ServletContextInitializer contextInitializer() {
         return new VaadinServletContextInitializer(context);
     }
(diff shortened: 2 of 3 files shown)


mshabarov:
Adding this annotation to the methods means that we expose a new API by allowing making a new, for instance, Spring Context Initialiser, custom implementation. However, this might be dangerous in some cases, because currently some of these beans are purely internal and we don't recommend to extend them or change.

Thus, I would like to, as a first step, check what exactly we can mark with this annotation, in other words, what we can allow developers to extend/change in their projects.

For your case, there is a need for custom route registry initialiser, which can be done in another way.

mshabarov:
@oliveryasuna I agree in general to make Vaadin auto-configuration more extendable.
But it seems to me that just adding this `AtConditionalOnMissingBean` annotation is not enough and that's why:
1. `AbstractRouteRegistryInitializer` isn't really open for extensions. It has a couple of protected methods, but you have to rewrite them for your case. All the private methods in it become useless then.
2. `RouteServletContextListener` has the essential logic in `failFastContextInitialized`, which uses private methods like `getRoutePackages`, `findByAnnotation` and others. So you cannot easily extend it.
3. `RouteServletContextListener` and other context listeners are `private`. This means that you need to replicate all these listeners in your project to make Vaadin work.

Thus, now I'm quite sure that the whole class `VaadinServletContextInitializer` isn't really extendable, but for internal use semantically.

Maybe if you provide your solution for `VaadinServletContextInitializer` custom implementation to me, I can see where I'm wrong.

But at the moment, I don't think it would be correct to let (and even recommend) developers to make a custom `VaadinServletContextInitializer`. This needs more design think over - more abstraction, more common logic, more protected methods/classes.

Let me check the rest of the beans you suggest to annotate.

mshabarov:
For the rest of the beans, I propose the following:
- `ServletRegistrationBean<SpringServlet>` - makes sense to annotate it 👍 , one may want a custom Spring servlet, e.g. for adding a custom Vaadin service class.
- `VaadinWebsocketEndpointExporter` - not sure if there is any use case which would need to extend Vaadin-provided `ServerEndpointExporter`. I would propose to allow having other beans of type `ServerEndpointExporter`, this would require changing the concrete Vaadin class (don't remember if it needs also specifying a name in the bean annotation):
```java
    @Bean
    public VaadinWebsocketEndpointExporter websocketEndpointDeployer() {
        return new VaadinWebsocketEndpointExporter();
    }
```
- `VaadinBeanFactoryInitializationAotProcessor` - same as for the previous point, without a good understanding where extending the existing class would be needed, I wouldn't allow to override it.
- `VaadinServletContextInitializer` - as I noted in my last comment, I see this useless of adding the annotation until we make this class more extendable (again, maybe I'm wrong, but so far there is no easy way). I think we can also specify a concrete class as a return type, so that it would be possible to add more `ServletContextInitializer` beans to the project.

mcollovati:
I agree with @mshabarov to avoid encouraging developers to override beans that are meant to be internal components.
This will prevent unexpected behaviors and lets us apply changes in patch releases, that could potentially be considered breaking if the API is made public.
On the other hand, we would love to understand uses cases for overriding Vaadin internals so that we can consider adding hooks or even make the API public, if reasonable.

In detail, I agree on not having conditional annotation on:
* `VaadinBeanFactoryInitializationAotProcessor`: it could make sense for a developer to register additional reflection hints, for example while waiting for a proper fix, but this can also be done by defining an additional `BeanFactoryInitializationAotProcessor` bean.
* `VaadinServletContextInitializer`: agree with Mikhail comments. If there are good uses cases, we can consider allowing developers to provide beans that we can look up and use in the existing components (e.g. a `RouteValidator` interface to be used in `AbstractRouteRegistryInitializer`).
* `VaadinWebsocketEndpointExporter`: honestly, I can't see a reason for a developer to override such a bean


Making `ServletRegistrationBean<SpringServlet>` conditional seems a good improvement.

vaadin-spring/pom.xml
```diff
@@ -51,6 +51,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-testbench</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
```
mcollovati, line 59: If possible, I would try to avoid adding a dependency.
But if `MockWebApplicationContext` is really needed, let's depend on `vaadin-testbench-unit-shared` instead of `vaadin-testbench`.
In addition, the `testbench.version` property should be used for version because the current project version may not match a Vaadin platform one,

mcollovati (changes requested):

mshabarov:
Closed in favour of https://github.com/vaadin/flow/pull/17883 (only scoped to ServletRegistration).

Issue #17234, opened 2023-07-18
Production Build with Vaadin 24 is very slow on a large project


simasch:
### Description of the bug

The production build with Vaadin 24 takes around 40 minutes compared to 5 minutes with Vaadin 23.

The project has around 400 Routes.

### Expected behavior

It shouldn't take longer than with previous versions.

### Minimal reproducible example

Unfortunately, I cannot share the project as it's not opensource.open source

### Versions

Vaadin 24.1.3

mcollovati:
First findings:
- During production build, class scanning is performed twice in Vaadin 24
- Byte code scanning performs the same both on V23 and V24, but V24 has an additional recursive step to collect reachable classes for every entry point. This seems to be slow, especially if the entry point can reach hundreds of classes.

mcollovati:
A first improvement consists in pre-compiling the regular expression used to detect visitable classes.
A simple test on a dummy project with 400 routes shows a significant build time reduction

```
8:17:37.387 [INFO] --- vaadin-maven-plugin:24.1.3:build-frontend (default) @ spring-skeleton ---
08:17:37.723 [INFO] Reflections took 292 ms to scan 127 urls, producing 4866 keys and 25879 values
08:17:37.743 [INFO] Scanning classes to find frontend configurations and dependencies...
08:17:46.729 [INFO] Visited 1377 classes. Took 8985 ms.
```

```
08:20:59.760 [INFO] --- flow-maven-plugin:24.2-SNAPSHOT:build-frontend (default) @ spring-skeleton ---
08:20:59.963 [INFO] Reflections took 196 ms to scan 127 urls, producing 4866 keys and 25880 values
08:20:59.969 [INFO] Scanning classes to find frontend configurations and dependencies...
08:21:02.292 [INFO] Visited 1370 classes. Took 2322 ms.
```

mcollovati:
Looking at log files for a very slow build, it seems that also `AbstractUpdateImports.process()` may take a long time during particular circumstances

gpsfl:
Is it possible to also manually exclude classes / package names? In our project it scans about 11000 classes most of which have nothing to do with vaadin.

simasch:
That's a very good question @gpsfl 

In my project it is
[INFO] Visited 16572 classes. Took 94647 ms

mshabarov:
Vaadin 24 and 23 use a slightly different algorithm to validate the license during production build.
[...]

PR #17672, opened 2023-09-20
chore: cache JS imported paths during chunk generation

mcollovati:
In a production build, for every chunk the JS files are parsed to extract the imported paths. There are checks to ensure that a module is visited only once, but this works on a chunk level.
This change caches the JS imports extraction results, so that a single JS file is visited only once during the build.

Part of #17234

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
@@ -491,7 +491,6 @@ abstract class AbstractUpdateImports implements Runnable {
     }
 
     private Set<String> getUniqueEs6ImportPaths(Collection<String> modules) {
-        long start = System.nanoTime();
         Set<String> npmNotFound = new HashSet<>();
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
```diff
@@ -781,23 +788,27 @@ private void visitImportsRecursively(Path filePath, String path,
...
+                                    + "This is probably due to it being a binary file, "
+                                    + "in which case it doesn't matter as imports are only in js/ts files.",
+                            filePath.toString(), ioe);
+                    return;
+                }
+                throw ioe;
             }
-            throw ioe;
+            ImportExtractor extractor = new ImportExtractor(content);
+            importedPathsCache.put(filePath, extractor.getImportedPaths());
```
tltv, line 808: Consider caching List of `resolvedPath`s instead of the whole file contents. `resolvedPath` is basically the end result of the each loop for the List of `importedPaths`. This would also use less memory.

tltv (dismissed):

tltv (approved):

PR #17680, opened 2023-09-21
fix: upgrade spring-boot to 3.1.4

ZheSun88:


mshabarov (dismissed):

mshabarov (approved):
LGTM, but I wonder if this is a breaking change for our Spring-based apps?

knoobie:
Probably related to Jetty https://github.com/spring-projects/spring-boot/commit/67e1ee5dcffde27439b71bdc16660134a6ecd9d5

mcollovati:
Great catch, thank you @knoobie.
Probably the test module is not completely correctly configured for a war deployment.
We have also several jetty jars in WEB-INF/lib and I am not sure if all of them are required.

Issue #11032, opened 2021-05-19
 Add ability to add devDependency with NpmPackage annotation


niiyeboah:
I would like to add an NPM package as a `devDependency`  from the Java side in order to exclude it from the frontend build.

### Example API:
`@NpmPackage(value="@types/d3", version = "6.2.0", devDependency = true)`

pleku:
The API seems simple enough. The implementation needs to change `FrontendDependencies` and `FullDependenciesScanner` so that it would separate the dev dependencies to another API declared in `FrontendDependencies` or then change the return type for `getPackages()` to include the dev-dependency information for each package. Then this logic should be taken into account when writing things to package.json.

michael-newsrx:
Is there any work around for this that doesn't involve editing the rather touchy packages.json file?

jsalts:
We would also like to see this added. It seems like Vaadin would like us to managing package.json with this annotation but is not providing all the tools to do so. Maybe manual edits are fine but it feels a bit anti-pattern.

oliveryasuna:
I am going to open a PR for this.

PR #17687, opened 2023-09-21
feat: Allow specifying `devDependencies` with `@NpmPackage`.

oliveryasuna:
## Description

This PR adds a `dev` property to the `@NpmPackage` annotation. Setting `dev = true` adds the package to `devDependencies` instead of `dependencies`. I specifically designed this to be backwards compatible with the current code. It does not affect bundles. I could repeat the motivation here, but enough has been said in the issue.

Fixes https://github.com/vaadin/flow/issues/11032

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


mshabarov:
@oliveryasuna thank you for your contribution 👍 
The enhancement you are implementing looks as a good addition to existing annotation.
Someone from Flow team will get back to you soon with a review.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTestComponents.java
```diff
@@ -54,6 +54,7 @@ public static class AllEager implements AppShellConfigurator {
     }
 
     @NpmPackage(value = "@vaadin/vaadin-button", version = "1.1.1")
+    @NpmPackage(value = "vite-plugin-pwa", version = "0.16.5", dev = true)
```
caalador, line 57: Just adding the dev package annotation doesn't test the feature.
`FullDependenciesScannerTest` should get a new test that checks that devDependencies are collected correctly and that it doesn't end up in both `getPackages` and `getDevPackages`

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -352,8 +352,8 @@ boolean updateDefaultDependencies(JsonObject packageJson) {
 
         for (Map.Entry<String, String> entry : getDefaultDevDependencies()
                 .entrySet()) {
-            added += addDependency(packageJson, DEV_DEPENDENCIES,
-                    entry.getKey(), entry.getValue());
+            added += addDependency(packageJson, DEV_DEPENDENCIES, entry.getKey(),
+                    entry.getValue());
```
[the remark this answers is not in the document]
oliveryasuna: @caalador What do you mean, "there is no change"?

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeTestComponents.java
oliveryasuna: Though, I just pushed a commit that adds a new test.

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -62,8 +62,7 @@
 import com.vaadin.flow.theme.NoTheme;
 import com.vaadin.flow.theme.ThemeDefinition;
 
-import static com.vaadin.flow.server.frontend.scanner.FrontendClassVisitor.VALUE;
-import static com.vaadin.flow.server.frontend.scanner.FrontendClassVisitor.VERSION;
+import static com.vaadin.flow.server.frontend.scanner.FrontendClassVisitor.*;
```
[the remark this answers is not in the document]
oliveryasuna: Fixed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -637,7 +644,15 @@ private void computePackages() throws ClassNotFoundException, IOException {
                         "Multiple npm versions for {} found:  {}. First version found '{}' will be considered.",
                         dependency, foundVersions, version);
             }
-            packages.put(dependency, version);
+            Set<Boolean> devs = npmPackageVisitor.getValuesForKey(VALUE,
+                    dependency, DEV);
+            devs.forEach(dev -> {
+                if (dev) {
```
caalador, line 650: The getValuesForKey only finds given values, so it doesn't check for the default value.
Either we should add a null check for dev here which would equal to false:
```suggestion
                if (dev != null && dev) {
```
Or then the getValuesForKey method in `FrontendAnnotatedClassVisitor` should have the map changed to `.map(h -> h.get(property) == null ? getDefault(property): h.get(property))` so that we would get the correct value.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
caalador: Seems that accepting defaults for the valuesForKey generate probelms in other places so the check should just have a null be the same as `dev=false`

caalador (approved):

oliveryasuna:
@caalador How do I know with which version this will be released?

mshabarov:
@oliveryasuna usually we follow this practice: `main` branch is for next minor release of Vaadin Flow. So once a feature is merged, it is included into next minor pre-release and finally in GA version. The referenced issue is automatically labeled with released version, e.g. `Released with Vaadin 24.1.12`, but this is a "post-factum" information.

Sometimes we backport some features to the older maintained major versions, but this isn't done automatically and the policy is to backport when it is sponsored.

We don't at the moment label pull requests or issues with milestones or other information about target version, but follow the policies described above.

oliveryasuna:
@mshabarov @caalador Thank you. Very useful information.

PR #17688, opened 2023-09-22
fix: mark execute as synchronized to prevent concurrent execution

taefi:
## Description

This assures that no concurrent
executions of this task result
in running multiple `npm install`
when something triggers the build
after the another build is in
progress.

Fixes vaadin/hilla#1309

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
@@ -81,7 +81,7 @@ public class TaskUpdatePackages extends NodeUpdater {
     }
 
     @Override
-    public void execute() {
+    public synchronized void execute() {
         try {
             Map<String, String> scannedApplicationDependencies = frontDeps
                     .getPackages();


mshabarov:
What triggers the reload in your case? Was it a java class change or something in Flow that triggers Spring Dev Tool to reload the server?

mshabarov:
I think even if we block the thread that tries to run `execute` concurrently,  there is no guarantee that it won't trigger another reload.

mshabarov:
Node tasks are re-created each time when `prepare-frontend` is triggered. So not sure this synchronisation would fix the original issue.

taefi:
> What triggers the reload in your case? Was it a java class change or something in Flow that triggers Spring Dev Tool to reload the server?

With the very recent version of the IntelliJ, an option is added to the Run configurations of SpringBoot applications to reload classes and resources on frame deactivation (whenever the IDE loses the focus) and it is enabled by default. This breaks Hilla (and possibly) starter application startup when `TaskUpdatePackages` and `TaskRunNpmInstall` are being called again and again whenever the user changes the focus from IntelliJ e.g. to the browser. 

Anyway, this was more of an experimentation rather than an actual PR. That's why I kept it in draft :)

PR #17712, opened 2023-09-25
In theme-editor, adds a test for vaadin-accordion component

brunovianarezende:
## Description

Adds a test for the editing of vaadin-accordion in theme-editor.

It is not possible to directly select the vaadin-accordion component directly by clicking on it, so the test is using the ComponentPicker component to navigate to the vaadin-accordion by using the keyboard.

This commit also contains:

-  a fix for a flakyness when a test tries to retrieve the CssRule related with a CssSelector
- normalization of the CssRule content inside the CssRule.equals(...) method.

## Type of change

- [ ] Bugfix
- [ X ] Feature

## Checklist

- [ X ] I have added tests to ensure my change is effective and works as intended.

### DIFF
--- a/flow-tests/test-theme-editor/src/test/java/com/vaadin/flow/uitest/ui/ThemeEditorIT.java
+++ b/flow-tests/test-theme-editor/src/test/java/com/vaadin/flow/uitest/ui/ThemeEditorIT.java
@@ -22,6 +22,7 @@ import com.vaadin.flow.component.internal.ComponentTracker;
 import com.vaadin.flow.internal.JsonUtils;
 import com.vaadin.flow.testutil.DevToolsElement;
 import com.vaadin.flow.uitest.ui.testbench.DevToolsCheckboxPropertyEditorElement;
+import com.vaadin.flow.uitest.ui.testbench.DevToolsComponentPickerElement;
 import com.vaadin.flow.uitest.ui.testbench.DevToolsThemeClassNameEditorElement;
 import com.vaadin.flow.uitest.ui.testbench.DevToolsThemeColorPropertyEditorElement;
 import com.vaadin.flow.uitest.ui.testbench.DevToolsThemeRangePropertyEditorElement;
@@ -33,7 +34,6 @@ import net.jcip.annotations.NotThreadSafe;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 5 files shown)


flow-tests/test-theme-editor/src/test/java/com/vaadin/flow/uitest/ui/ThemeEditorIT.java
```diff
@@ -117,8 +117,24 @@ public void testCheckbox() throws IOException {
         performTagTest("vaadin-checkbox", By.id("checkbox"));
     }
 
-    protected void performTagTest(String tagName, By componentSelector)
+    @Test
+    public void testVaadinAccordion() throws IOException {
```
MarcinVaadin, line 121: Is it better to have single parametrized / looping test that will go through all available metadata and look for given elements by id, like vaadin-accordion -> #accoridion? Or single test per component type?

brunovianarezende: at the moment I'm adding one test per component because not all test are the same, you can see that vaadin-accordion requires the use of ComponentPicker to select the component, it is not really possible to select it in the page through a click.

Other issues with the loop:

1. initialize the components in ThemeEditorView
2. there are metadata for components that I don't expect the user will ever instantiate directly, e.g. vaadin-accordion-heading (I might be wrong here)

szolo: What about hiding implementation details behind a common interface? Shared logic could be moved to abstract class like @MarcinVaadin proposed.

szolo:
@brunovianarezende thanks for a PR! ;) Looks good. Please format the code appropriately and feel free to proceed with a merge.

szolo (dismissed):

flow-tests/test-theme-editor/src/test/java/com/vaadin/flow/uitest/ui/ThemeEditorIT.java
szolo: I meant slightly different thing. Potentially, each component test could be slightly different and have its own specifics, like you described above. Why not to have a class which will encapsulate this difference for a potential component while common logic will be inside parent class. Something like `ButtonTest extends AbstractTest`.

brunovianarezende: got it. At  the moment, we only have two variations in the tests (i.e. the way we select the component). Let's see what happens when I proceed with this work. I'm thinking now in the next steps, maybe I will add tests for multiple components; other option is to increase what I'm checking.

PR #17735, opened 2023-09-28
chore: remove left-overs for apache httpclient

mshabarov:


flow-client/pom.xml
```diff
@@ -78,18 +78,6 @@
...
-            <version>4.5.14</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>commons-codec</groupId>
-                    <artifactId>commons-codec</artifactId>
-                </exclusion>
-            </exclusions>
-            <scope>provided</scope>
-        </dependency>
         <!-- Overrides the dependency version of httpclient -->
```
knoobie, line 81: The commons-codec provided dependency could also be removed or? Looks like Snyk doesn't like this change :)

mcollovati:
It looks like the dependency in `flow-client` was added by the PR that removed httpclient, and the reason was fixing the snyk failure (2c5f45a04f0aa521281139ce2541bbb2ec05ce65).
So, we can safely remove the dependency (they are not used at runtime).
Otherwise, we need to preserve the dependencies, as they are used by `flow-client` tests.
@mshabarov what do you think?

mshabarov:
Thanks @mcollovati for tips. I think have to preserve the dependency. I've created this PR just because I noticed an entry for https client, which, as I thought, was a left-over. But it's not.

Issue #17629, opened 2023-09-14
Spring different ViewAccessChecks


Silicat-git:
### Description of the bug

When using spring with the SpringViewAccessChecker there are 2 different methods beeing called to check for the roles, depending if a VaadinRequest exists or not.

VaadinRequest::isUserInRole checks for the "ROLE_" prefix and adds it if needed.

AuthenticationUtil::getSecurityHolderRoleChecker adds the "ROLE_" prefix no matter what.

This causes issues since the VaadinRequest is not always non null in our project which results in a reroute to our access denied view even tho the role is present. Since instead of searching for e.g. ROLE_ADMIN it will search for ROLE_ROLE_ADMIN with the AuthenticationUtil.

SpringViewAccessChecker:

    @Override
    protected Function<String, Boolean> getRolesChecker(VaadinRequest request) {
        if (request == null) {
            return AuthenticationUtil.getSecurityHolderRoleChecker();
        }
        return super.getRolesChecker(request);
    }

ViewAccessChecker:

    protected Function<String, Boolean> getRolesChecker(VaadinRequest request) {
        if (request == null) {
            return role -> false;
        }

        return request::isUserInRole;
    }

VaadinRequest::isUserInRole (HttpServletRequestWrapper) -> SecurityContextHolderAwareRequestWrapper

	public boolean isUserInRole(String role) {
	     return this._getHttpServletRequest().isUserInRole(role);
	}

SecurityContextHolderAwareRequestWrapper

	@Override
	public boolean isUserInRole(String role) {
		return isGranted(role);
	}

	private boolean isGranted(String role) {
		Authentication auth = getAuthentication();
		if (this.rolePrefix != null && role != null && !role.startsWith(this.rolePrefix)) {
			role = this.rolePrefix + role;
		}
		if ((auth == null) || (auth.getPrincipal() == null)) {
			return false;
		}
		Collection<? extends GrantedAuthority> authorities = auth.getAuthorities();
		if (authorities == null) {
			return false;
		}
		for (GrantedAuthority grantedAuthority : authorities) {
			if (role.equals(grantedAuthority.getAuthority())) {
				return true;
			}
		}
		return false;
	}

AuthenticationUtil:

    public static Function<String, Boolean> getSecurityHolderRoleChecker() {
        Authentication authentication = getSecurityHolderAuthentication();
        if (authentication == null) {
            return role -> false;
        }
        return role -> authentication.getAuthorities().stream()
[...]

PR #17739, opened 2023-09-28
fix: update role checkers to work in same way

tltv:
Added `AuthenticationUtil#getSecurityHolderRoleChecker(String rolePrefix)` method that works now in same way as `SecurityContextHolderAwareRequestWrapper#isGranted(String role)`. Role prefix is set and read in `VaadinRolePrefixHolder` bean by `SpringViewAccessChecker#getRolesChecker(VaadinRequest)`.

Fixes: https://github.com/vaadin/flow/issues/17629

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
@@ -19,6 +19,7 @@ import com.vaadin.flow.server.auth.AccessAnnotationChecker;
 import com.vaadin.flow.server.auth.ViewAccessChecker;
 import com.vaadin.flow.spring.security.RequestUtil;
 import com.vaadin.flow.spring.security.VaadinDefaultRequestCache;
+import com.vaadin.flow.spring.security.VaadinRolePrefixHolder;
 import com.vaadin.flow.spring.security.ViewAccessCheckerInitializer;
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
@@ -95,5 +96,15 @@ public class SpringSecurityAutoConfiguration {
     @Bean
     public RequestUtil requestUtil() {
         return new RequestUtil();
+    }
+
+    /**
+     * Makes role prefix holder available for security configuration.
+     *
+     * @return the role prefix holder
+     */
+    @Bean
+    public VaadinRolePrefixHolder vaadinRolePrefixHolder() {
+        return new VaadinRolePrefixHolder(null);
     }
 }
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
@@ -15,6 +15,8 @@
  */
 package com.vaadin.flow.spring;
 
+import java.util.Optional;
+
 import com.vaadin.flow.server.auth.AccessAnnotationChecker;
 import com.vaadin.flow.server.auth.ViewAccessChecker;
 import com.vaadin.flow.spring.security.RequestUtil;
@@ -28,6 +30,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityCustomizer;
+import org.springframework.security.config.core.GrantedAuthorityDefaults;
 
 /**
  * Spring boot auto-configuration class for Flow.
@@ -101,10 +104,17 @@ public class SpringSecurityAutoConfiguration {
     /**
      * Makes role prefix holder available for security configuration.
      *
+     * @param grantedAuthorityDefaults
+     *            Optional granted authority defaults bean for the default role
+     *            prefix
      * @return the role prefix holder
      */
     @Bean
-    public VaadinRolePrefixHolder vaadinRolePrefixHolder() {
-        return new VaadinRolePrefixHolder(null);
+    @ConditionalOnMissingBean
+    public VaadinRolePrefixHolder vaadinRolePrefixHolder(
+            Optional<GrantedAuthorityDefaults> grantedAuthorityDefaults) {
+        return new VaadinRolePrefixHolder(grantedAuthorityDefaults
+                .map(GrantedAuthorityDefaults::getRolePrefix).orElse(null));
     }
+
 }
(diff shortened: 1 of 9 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
```diff
@@ -96,4 +97,14 @@ public AccessAnnotationChecker accessAnnotationChecker() {
...
         return new RequestUtil();
     }
+
+    /**
+     * Makes role prefix holder available for security configuration.
+     *
+     * @return the role prefix holder
+     */
+    @Bean
+    public VaadinRolePrefixHolder vaadinRolePrefixHolder() {
```
mcollovati, line 107: I wonder if we can inject `GrantedAuthorityDefaults` as a method parameter and get the default role prefix from there.
Also, we could probably make the method `@ConditionalOnMissingBean`, so developers can easily customize it if they need.
Or is this bean supposed to be fully managed by Vaadin, to be in sync with Spring Security settings?

If it's the second case, `VaadinRolePrefixHolder` could most likely be final and have a package protected constructor.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringViewAccessChecker.java
```diff
@@ -37,8 +41,24 @@ protected Principal getPrincipal(VaadinRequest request) {
...
         }
+
+        // Update active role prefix if it's not set yet.
+        Optional.ofNullable(VaadinService.getCurrent())
+                .map(service -> service.getContext().getAttribute(Lookup.class))
+                .map(lookup -> lookup.lookup(VaadinRolePrefixHolder.class))
+                .filter(prefixHolder -> prefixHolder.getRolePrefix() == null)
+                .ifPresent(
+                        prefixHolder -> prefixHolder.resetRolePrefix(request));
+
```
mcollovati, line 61: What if we also try to initialize the `VaadinRoleprefix` role prefix in `VaadinWebSecurity`, by looking for `SecurityContextHolderAwareRequestFilter` in `securityFilterChain.getFilters(), after the call to `HttpSecurity.build()`?

Anyway, this call would still be useful if `VaadinWebSecurity` is not in use.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
tltv: `GrantedAuthorityDefaults` seems to indeed provide role prefix and its also effective in `SecurityContextHolderAwareRequestWrapper` when provided. So yes this can be done. Unfortunately bean does not exist by default so need to have it as an optional parameter.

It should be safe to add `@ConditionalOnMissingBean` for this.

tltv:
`GrantedAuthorityDefaults` could have been actually used instead of introducing `VaadinRolePrefixHolder` just by adding it in `SpringSecurityAutoConfiguration` by default with fixed default role prefix `ROLE_` and with `@ConditionalOnMissingBean`. 
But that would just move the fixed default role prefix into other place, although it would help those who uses `GrantedAuthorityDefaults` already, it would not help those who use some other way to override the default role prefix. 

This change should work now no matter how default value is changed. Usage of reflection is downside of it.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringViewAccessChecker.java
```diff
@@ -37,8 +41,24 @@ protected Principal getPrincipal(VaadinRequest request) {
...
+                            AuthenticationUtil::getSecurityHolderRoleChecker);
         }
+
+        // Update active role prefix if it's not set yet.
+        Optional.ofNullable(VaadinService.getCurrent())
+                .map(service -> service.getContext().getAttribute(Lookup.class))
+                .map(lookup -> lookup.lookup(VaadinRolePrefixHolder.class))
+                .filter(prefixHolder -> prefixHolder.getRolePrefix() == null)
+                .ifPresent(
+                        prefixHolder -> prefixHolder.resetRolePrefix(request));
```
mcollovati, line 60: Just a note, but not a blocker: if for some reason `prefixHolder.resetRolePrefix()` fails to set the prefix, or if the prefix has been configured to be `null` (however I don't know if it is really possible), we are going to execute the reflection stuff every time we need the role checker instance.
Would it make sense to set a flag in `VaadinRolePrefixHolder` when `resetRolePrefix` has been executed for the first time (or if a not null value is given to the constructor), and then have a `VaadinRolePrefixHolder.isSet()` instead of the `prefixHolder.getRolePrefix() == null` check?

mcollovati (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringViewAccessChecker.java
tltv: It's indeed possible to set it to null and I totally agree that it makes sense to add this flag to avoid resetting over and over with null prefix..

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinRolePrefixHolder.java
```diff
@@ -0,0 +1,163 @@
...
+    public boolean isSet() {
+        return rolePrefixSet;
+    }
+
+    /**
+     * Reset role prefix from the given request. Works only with a
+     * {@link VaadinServletRequest} and a chain of
+     * {@link ServletRequestWrapper}s to find
+     * {@link SecurityContextHolderAwareRequestWrapper} with the role prefix.
+     * Method doesn't do anything if role is not found.
```
mcollovati, line 83:
```suggestion
     * Method doesn't do anything if role prefix is not found.
```

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinRolePrefixHolder.java
```diff
@@ -0,0 +1,163 @@
...
+     */
+    public void resetRolePrefix(VaadinRequest request) {
+        SecurityContextHolderAwareRequestWrapper requestWrapper = findSecurityContextHolderAwareRequestWrapper(
+                request);
+        resetRolePrefix(requestWrapper);
+    }
+
+    /**
+     * Reset role prefix from the given {@link DefaultSecurityFilterChain}.
+     * Method doesn't do anything if role is not found.
```
mcollovati, line 96:
```suggestion
     * Method doesn't do anything if role prefix is not found.
```

mcollovati (changes requested):

mcollovati (approved):

PR #17745, opened 2023-09-29
feature: provide an API to make router ignore an anchor

### LINKED ISSUE #17617: Provide method to set Anchor's router-ignore flag.
### Describe your motivation

Currently, when I want to tell Vaadin to not intercept an anchor click, I need to set the `router-ignore` flag using the Element API. Since we should prevent make devs use the Element API for "normal" features, a dedicated `setRouterIgnore` method should be provided.

Also see: https://vaadin.com/docs/latest/routing/navigation#using-standard-links

Anchor should provide a `setRouterIgnore` method to improve the DX for configuring this feature.

_No response_

### Additional context

_No response_

mstahv:
## Description

Anchors in Vaadin are simply broken if the url happens to be relative. With this API that can be fixed.

Fixes #17617

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -47,6 +47,7 @@ public class Anchor extends HtmlContainer
             .optionalAttributeWithDefault("target",
                     AnchorTarget.DEFAULT.getValue());
 
+    private static final String ROUTER_IGNORE_ATTRIBUTE = "router-ignore";
     private Serializable href;
 
     /**
@@ -176,7 +177,7 @@ public class Anchor extends HtmlContainer
     /**
      * The routing mechanism in Vaadin by default intercepts all anchor elements
      * with relative URL. This method can be used make the router ignore this
-     * anchor and this way make this anchor behave normally and cause a full 
+     * anchor and this way make this anchor behave normally and cause a full
      * page load.
      *
      * @param ignore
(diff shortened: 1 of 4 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -170,10 +170,32 @@ public void removeHref() {
...
     public void setHref(AbstractStreamResource href) {
         this.href = href;
-        getElement().setAttribute(ROUTER_IGNORE_ATTRIBUTE, true);
+        setRouterIgnore(true);
         assignHrefAttribute();
     }
 
+    /**
+     * The routing mechanism in Vaadin by default intercepts all anchor elements
+     * with relative URL. This method can be used make the router ignore this
```
knoobie, line 179: I'm wondering if it would be worth to add e.g. a info-level logging in development mode on attach (like done in Label) to inform the user about this behavior when setRouterIgnore(false) and setHref(relative path)? 

I feel like most of the "pitfalls" (when they can't be changed easily, cause "senior" users are used to it by now) can be improved for new user by providing such information in the logs to notify them about the problem and possible better practice?

mstahv: More than the added phase (which might be handy as well), would be more/as handy if could intercept the phase when the state is about to be synced with the client. That I have missed a dozen of times.  There is this related UI.beforeClientResponse, but that is also component specific (and the method is in weird place and not that well documented in our reference docs).

https://github.com/vaadin/flow/blob/main/flow-server/src/main/java/com/vaadin/flow/component/UI.java#L1283-L1332

mstahv: And yeah, something derived to that could be used for some actual logic as well, not just complaining that our users have not figured out that we have bad defaults ;-)

knoobie: Yeah "beforeClientResponse".. I remember that method, everytime I see it in the flow components it feels like some hack-ish way to workaround limitations 😬

knoobie:
I agree with Matti (it was just a thought experiment), but  I trust people to missuse the anchor, like any other component 🙈

Issue #17721, opened 2023-09-26
 LitTemplateParser finds wrong content for render template


jensgoe:
### Description of the bug

LitTemplates renders wrong html content. Occurs if an additional function that returns html content exists after the render function in a lit template java script file.

### Expected behavior

A component from a lit template renders the content of its render function.

### Minimal reproducible example

Please see failing unit test of #17719 

### Versions

- Vaadin / Flow version: 24.1.9
- Java version: JDK 17
- OS version: Win/Linux
- Browser version (if applicable): firefox/chrome
- Application Server (if applicable): payara
- IDE (if applicable): eclipse

PR #17765, opened 2023-10-02
fix: improve LitTemplateParser to find correct content

tltv:
Changes regex in LitTemplateParser to find correct template content in case when lit template code contains "return html`" template also after the render function.

Fixes: #17721

### DIFF
--- a/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
+++ b/flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
@@ -42,10 +42,27 @@ public final class BundleLitParser {
             .getLogger(BundleLitParser.class);
 
     /**
-     * Lit template pattern matches the template getter
+     * Lit template pattern matches the render function
      *
      * <pre>
-     *     render() {
+     * render() {
+     *
+     * }
+     * </pre>
+     *
+     * <p>
+     * <code>render\(\)[\s]*\{</code> finds the template getter method and
+     * <code>[\s\S]*\}</code> ensures everything is captured until the last
+     * <code>}</code> character.
+     */
+    private static final Pattern LIT_TEMPLATE_PATTERN = Pattern
+            .compile("render\\(\\)[\\s]*\\{[\\s\\S]*\\}");
+
+    /**
+     * Lit template pattern for html matches the return statement with html
+     * template.
+     *
+     * <pre>
      *       return html`
      *         &lt;style&gt;
      *           .response { margin-top: 10px`; }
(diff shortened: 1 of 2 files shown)


flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
```diff
@@ -66,8 +66,11 @@ public final class BundleLitParser {
      * <code>([\s\S]*)</code> captures all text until we encounter the end
      * character with <code>\1;}</code> e.g. <code>';}</code>
      */
-    private static final Pattern LIT_TEMPLATE_PATTERN = Pattern.compile(
-            "render\\(\\)[\\s]*\\{[\\s\\S]*return[\\s]*html[\\s]*(\\`)([\\s\\S]*?)\\1;[\\s]*\\}");
+    private static final Pattern LIT_TEMPLATE_PATTERN = Pattern
```
caalador, line 69: JavaDoc explaining pattern is not correct anymore. This should be added fixed so it is clearer what is happening/done.

flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
```diff
@@ -42,10 +42,27 @@ public final class BundleLitParser {
             .getLogger(BundleLitParser.class);
 
     /**
-     * Lit template pattern matches the template getter
+     * Lit template pattern matches the render function
```
caalador, line 45: This is technically wrong as it matches from `render() {` till the last `}` in the string.

flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
```diff
@@ -87,26 +102,32 @@ public static Element parseLitTemplateElement(String fileName,
...
-        if (templateMatcher.find() && templateMatcher.groupCount() == 2) {
-            String group = templateMatcher.group(2);
-            LOGGER.trace("Found regular Lit template content was {}", group);
+        if (renderMatcher.find()) {
+            String renderGroup = renderMatcher.group(0);
+            Matcher templateMatcher = LIT_TEMPLATE_PATTERN_HTML
+                    .matcher(renderGroup);
+            // GroupCount should be at least 3 as the first group contains whole
+            // content and second group contains `|'|". Third group contains
+            // first "return html'" template contents.
```
caalador, line 113: This comment is actually wrong as per the regex pattern we only take 2 groups `(\')` and `([\s\S]*?)`
So that part shouldn't change.

flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
```diff
@@ -42,10 +42,27 @@ public final class BundleLitParser {
...
+     * <code>render\(\)[\s]*\{</code> finds the template getter method and
+     * <code>[\s\S]*\}</code> ensures everything is captured until the last
+     * <code>}</code> character.
+     */
+    private static final Pattern LIT_TEMPLATE_PATTERN = Pattern
+            .compile("render\\(\\)[\\s]*\\{[\\s\\S]*\\}");
+
+    /**
+     * Lit template pattern for html matches the return statement with html
+     * template.
```
caalador, line 63: Here we could also mark that we use this to take the first match from `render() {`

caalador (commented):
Next time there is requirements for this I guess we should move away from regex as it is not exact enough.

flow-lit-template/src/main/java/com/vaadin/flow/component/littemplate/BundleLitParser.java
tltv: True, `groupCount()` excludes 0-index. but not based on this comment. Fixing. 
Keeping the text still to point out that it's the first html template content in group(2) we are interested in. There may be more groups after.

caalador (approved):

Issue #17722, opened 2023-09-27
Web Components are not properly registered in non default Route in 24.2


manolo:
### Description of the bug

In certain projects, web components are not enhanced when changing from main route like `""` to other routes like `"/form"`.
Actually, the WC tag is in the UI, but they are not correctly registered if you inspect the DOM structure or the `window.Vaadin.registrations` object

### Expected behavior

Components are correctly registered regardless of the path

### Minimal reproducible example

1. Clone Collaboration Engine project and switch to the branch migrated to 24.2
```
git clone git@github.com:vaadin/collaboration-engine-internal.git
cd collaboration-engine-internal/
git checkout chore/bump-to-24.2
```
2. Compile and install
```
mvn clean install -DskipTests
```
3. Run the application for the ITs
```
cd collaboration-engine-test/collaboration-engine-integration-test/
mvn
```
4. Visit main page http://localhost:8080/ WC are correctly registered and UI correctly rendered
```
Vaadin.registrations.map(c => typeof c === 'function' && (new c))

[false, false, vaadin-iconset, vaadin-vertical-layout, vaadin-button, vaadin-horizontal-layout]
```

![Screenshot 2023-09-27 at 07 37 38](https://github.com/vaadin/flow/assets/161853/3d33f489-8ec1-4af2-b4dc-175f78f96258)

5. Visit another route http://localhost:8080/form only vertical layout is registered
```
Vaadin.registrations.map(c => typeof c === 'function' && (new c))

[false, false, vaadin-iconset, vaadin-vertical-layout]
```

![Screenshot 2023-09-27 at 07 37 55](https://github.com/vaadin/flow/assets/161853/79cf7064-6e7b-4449-b8e2-f8c2b82a6289)

### Versions

- Flow version: 24.2.0.beta1 24.2-SNAPSHOT

mshabarov:
Also doesn't work with Vaadin / Flow 24.1.
The root cause is that the Flow doesn't add the reachable classes from parent view `CollaborativeFormViewCommon` to `CollaborativeFormView`, because only `CollaborativeFormView ` is marked with `AtRoute`.
```java
public class CollaborativeFormView extends CollaborativeFormViewCommon {
   // has no codes
}
```
```java
@Route("form")
@PreserveOnRefresh
public class CollaborativeFormViewCommon extends VerticalLayout {
   // TextField, Checkbox and other components included
}
```
[...]

PR #17772, opened 2023-10-04
fix: Add reachable classes to child route

mshabarov:
## Description

When child entry-point has no AtRoute annotation, but it's parent class does have, the reachable classes from parent class are added to the child entry-point, so frontend modules are added as well and bundle chunk is generated with all needed web-components.

Fixes https://github.com/vaadin/flow/issues/17722

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
@@ -227,10 +227,8 @@ public class FrontendDependencies extends AbstractDependenciesScanner {
     }
 
     private void visitEntryPoints() throws IOException {
-        List<EntryPointData> sortedEntryPoints = entryPoints.values().stream()
-                .sorted(this::compareEntryPoints).collect(Collectors.toList());
-        for (EntryPointData entry : sortedEntryPoints) {
-            visitEntryPoint(entry);
+        for (Entry<String, EntryPointData> entry : entryPoints.entrySet()) {
+            visitEntryPoint(entry.getValue());
         }
 
     }
@@ -409,7 +407,12 @@ public class FrontendDependencies extends AbstractDependenciesScanner {
         // references loaded by the specific class finder loader
         Class<? extends Annotation> triggerClass = getFinder()
                 .loadClass(DependencyTrigger.class.getName());
-        for (Class<?> route : getFinder().getAnnotatedClasses(routeClass)) {
+
+        List<Class<?>> routeClasses = new ArrayList<>(
+                getFinder().getAnnotatedClasses(routeClass));
+        routeClasses.sort(this::compareEntryPoints);
+
+        for (Class<?> route : routeClasses) {
             List<String> triggerClasses = getDependencyTriggers(route,
                     triggerClass);
             boolean eager = isEagerRoute(route);
@@ -875,11 +878,10 @@ public class FrontendDependencies extends AbstractDependenciesScanner {
         return false;
     }
 
-    private int compareEntryPoints(EntryPointData entryPointData1,
-            EntryPointData entryPointData2) {
-        Annotation routeAnnotation1 = entryPointData1.getClassInfo()
+    private int compareEntryPoints(Class<?> entryPoint1, Class<?> entryPoint2) {
+        Annotation routeAnnotation1 = entryPoint1
                 .getDeclaredAnnotation(routeClass);
-        Annotation routeAnnotation2 = entryPointData2.getClassInfo()
+        Annotation routeAnnotation2 = entryPoint2
                 .getDeclaredAnnotation(routeClass);
         if (routeAnnotation1 != null && routeAnnotation2 == null) {
             return -1;
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
@@ -227,8 +227,10 @@ public class FrontendDependencies extends AbstractDependenciesScanner {
     }
 
     private void visitEntryPoints() throws IOException {
-        for (Entry<String, EntryPointData> entry : entryPoints.entrySet()) {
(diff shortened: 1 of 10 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -216,8 +216,11 @@ private void collectReachableClasses(String name, Set<String> classes) {
 
         classes.add(name);
         for (String className : visitedClass.children) {
-            if (!entryPoints.containsKey(className)) {
+            EntryPointData entryPointData = entryPoints.get(className);
+            if (entryPointData == null) {
                 collectReachableClasses(className, classes);
+            } else if (entryPointData.reachableClasses != null) {
+                classes.addAll(entryPointData.reachableClasses);
             }
```
mcollovati, line 224: I think that there's an uncovered case here: the above code works if the parent class is visited before the child one (`entryPointData.reachableClasses` will not be null), but will not work correctly if the child class is visited before the parent.
And this could happen because both classes are entry points, but I think there's no guarantee about the order they are fetched from the dependency scanner.

We could probably do some sort based on hierarchy in `collectEntryPoints()` after call to `getFinder().getAnnotatedClasses(routeClass)`.

```
        for (Class<?> route : getFinder().getAnnotatedClasses(routeClass)) {
            List<String> triggerClasses = getDependencyTriggers(route,
                    triggerClass);
            boolean eager = isEagerRoute(route);
            addEntryPoint(route, EntryPointType.ROUTE, triggerClasses, eager);
        }
```

mcollovati: An idea could be to sort in a way that annotated classes gets first (`Class.getDeclaredAnnotation(Route.class) != null)`)

mshabarov: Excellent catch and good suggestion! Those with annotation go to beginning of the collection, others go to the end.

mshabarov: And this is most likely the reason why my new test fails.

flow-server/src/test/java/com/vaadin/flow/server/frontend/scanner/FrontendDependenciesTest.java
```diff
@@ -323,6 +324,35 @@ public void shouldVisit_shouldNotMatchOnPartOfPackage() {
                 dependencies.shouldVisit("com.sun"));
     }
 
+    @Test
+    public void classScanningForChildAndParentEntryPoint_parentAnnotated_childNotAnnotated_childHasModulesFromParent() {
+        Mockito.when(classFinder.getAnnotatedClasses(Route.class))
+                .thenReturn(Stream.of(ChildRoute.class, ParentRoute.class)
+                        .collect(Collectors.toSet()));
```
mcollovati, line 331: It would be nice to collect into an `LinkedHashMap` and have two tests, with route classes in different order

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
```diff
@@ -869,4 +877,18 @@ private boolean hasThemeInfo(ClassInfo classInfo) {
...
+                .getDeclaredAnnotation(routeClass);
+        Annotation routeAnnotation2 = entryPoint2
+                .getDeclaredAnnotation(routeClass);
+        if (routeAnnotation1 != null && routeAnnotation2 == null) {
+            return -1;
+        }
+        if (routeAnnotation1 == null && routeAnnotation2 != null) {
+            return 1;
+        }
+        return 0;
```
mcollovati, line 892: I wonder if we should compare not annotated classes by hierarchy depth, to prevent issues with deep hierarchy
For example:

```
@Route
ParentRoute {}

ChildRoute < ParentRoute

GrandchildRoute < ChildRoute
```

mshabarov: Good idea, one may use this kind of hierarchy. I'll extend the implementation to take it into account.

mshabarov: This might be enough to handle this case:
```java
if (routeAnnotation1 != null && routeAnnotation2 == null) {
            return -1;
        }
        if (routeAnnotation1 == null && routeAnnotation2 != null) {
            return 1;
        }
        if (entryPoint1.isAssignableFrom(entryPoint2)) {
            return -1;
        }
        if (entryPoint2.isAssignableFrom(entryPoint1)) {
            return 1;
        }
        return 0;
```

mshabarov: Don't know if we need to check hierarchy for annotated classes. Probably not, because they get proper reachable classes always.

mcollovati: The check for annotated classes is probably not needed, but we can maybe add a test to be sure it works correctly.

```
@Route("parent")
ParentRoute {}

@Route("child")
ChildRoute < ParentRoute
```

mshabarov: Done. Added few more tests to check the edge cases (ordered and reversed) and also case for shuffled entry points.

flow-server/src/test/java/com/vaadin/flow/server/frontend/scanner/FrontendDependenciesTest.java
mshabarov: Added more tests. JUnit 4 doesn't have "per method" parameterised feature, so I made it with just three separate methods.

mcollovati (approved):

PR #17774, opened 2023-10-04
Do not passivate child keys with active&expanded parents

### LINKED ISSUE #14351: TreeGrid: DragItem null after dragging root item to another hierarchy and then somewhere else
### Description

After I dragged a root item to another hierarchy and try to move it again, the draggedItems list in the GridDragStartEvent has a null item inside.
The problem only seems to be with dragging **root items** to another hierarchy and then trying do drag **that** item again.

Side note: The refreshDataProvider is called, but maybe not in time (async)!? Calling it manually (e.g. via extra button) before dragging the just dropped item again seemed to help in my other test case.

### Expected outcome

I would expect the dragItem to never be null.

```java
public class MainView extends HorizontalLayout {

    TreeGrid<String> tree = new TreeGrid<>(String.class);
    List<String> draggedItems = null;

    public MainView() {
        setMargin(true);

        tree.removeAllColumns();
        tree.addHierarchyColumn(item -> item).setHeader("Column");
        tree.addColumn(item -> item.length() + "");
        tree.addComponentColumn(item -> VaadinIcon.ACCESSIBILITY.create());
        TreeData<String> treeData = new TreeData<>();
        TreeDataProvider<String> provider = new TreeDataProvider<>(treeData);
        tree.setDataProvider(provider);
        treeData.addRootItems("A", "B", "C");
        treeData.addItems("A", "A1", "A2");
        treeData.addItems("A1", "C1", "C2");
        treeData.addItems("C1", "D1", "D2");
        treeData.addItems("B", "B1", "B2");

        tree.expand("A");
        tree.getDataProvider().refreshAll();
        tree.getDataCommunicator().reset();
        add(tree);

        initDragAndDrop();
    }

    public void initDragAndDrop() {
        tree.setRowsDraggable(true);
        tree.setDropMode(GridDropMode.ON_TOP_OR_BETWEEN);
        tree.addDragStartListener(event -> {
            draggedItems = event.getDraggedItems();
        });
        tree.addDragEndListener(event -> {
            draggedItems = null;
        });

        tree.addDropListener(event -> {
            String dropItem = event.getDropTargetItem().orElse(null);
            if (draggedItems == null || draggedItems.size() == 0) {
                return;
            }
            handleDrop(event, dropItem);
        });
    }

    private void handleDrop(GridDropEvent<String> inEvent, String dropItem) {
        if (draggedItems == null) {
            return;
        }
        TreeData<String> treeData = tree.getTreeData();
        int dropIndex = 0;
        List<String> dropChildItems;
        boolean justAddAtTheEnd = false;
        String newParent = dropItem == null ? null : treeData.getParent(dropItem);

        if (dropItem == null || treeData.getChildren(dropItem).size() > 0 && inEvent.getDropLocation() == GridDropLocation.ON_TOP) {
            dropChildItems = treeData.getChildren(dropItem);
            dropIndex = dropChildItems.size();
            justAddAtTheEnd = true;
        } else {
            if (inEvent.getDropLocation() == GridDropLocation.BELOW && treeData.getChildren(dropItem).size() > 0 && tree.isExpanded(dropItem)) {
                // dropped on open folder below so insert into that folder
                dropChildItems = tree.getTreeData().getChildren(dropItem);
                newParent = dropItem;
                dropIndex = 0;
            } else {
                dropChildItems = tree.getTreeData().getChildren(treeData.getParent(dropItem));
                dropIndex = dropChildItems.indexOf(dropItem);
                if (inEvent.getDropLocation() == GridDropLocation.BELOW) {
                    dropIndex++;
                }
            }
        }

        try {
            for (String dragItem : draggedItems) {
                if (dragItem == null) {
                    Notification.show("drag item null"); // FIXME: shouldn't happen
                    return;
                }
                if (dragItem == dropItem) {
                    // dropped on itself
                    return;
                }
                // Make sure that the target is not in the subtree of the dragged item itself
                // ...
                if (justAddAtTheEnd) {
                    treeData.setParent(dragItem, dropItem);
                    tree.expand(dropItem);
                } else {
                    treeData.setParent(dragItem, newParent);

                    boolean toggle = true;
                    if (dropIndex >= dropChildItems.size()) {
                        dropIndex = dropChildItems.size() - 1;
                        toggle = false; // don't toggle items if dropped at the end
                    }
                    String sibling = dropChildItems.get(dropIndex);

                    treeData.moveAfterSibling(dragItem, sibling);
                    if (toggle) {
                        // toggle items with this so drop at index 0 is possible
                        treeData.moveAfterSibling(sibling, dragItem);
                    }
                }
            }
        } catch (Exception ex) {
            ex.printStackTrace();
        }
        tree.getDataCommunicator().reset();
        tree.getDataProvider().refreshAll();
    }

}
```

### Steps to reproduce

Example 1:
1. Expand B
2. Drag C in B
3. Try to drag C
-> drag item null

Example 2: 
1. Drag/drop A1 above A to root
2. Drag A1 back in A to same place
3. Drag/drop A1 above A again to root
-> drag item null

### Environment

Vaadin version(s): 23.1.4

### Browsers

Chrome, Firefox

tepi:
Fixes #14351

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
@@ -544,9 +544,22 @@ public class HierarchicalDataCommunicator<T> extends DataCommunicator<T> {
 
     @Override
     protected Set<String> getPassivatedKeys(Set<String> oldActive) {
-        return super.getPassivatedKeys(oldActive).stream()
-                .filter(key -> !isExpanded(getKeyMapper().get(key)))
-                .collect(Collectors.toCollection(HashSet::new));
+        return super.getPassivatedKeys(oldActive).stream().filter(key -> {
+            T item = getKeyMapper().get(key);
+            if (item != null) {
+                boolean passivateItem = false;
+                T parent = getParentItem(item);
+                while (parent != null) {
+                    if (!isItemActive(parent) || !isExpanded(parent)) {
+                        passivateItem = true;
+                        break;
+                    }
+                    parent = getParentItem(parent);
+                }
+                return passivateItem;
+            }
+            return !isExpanded(getKeyMapper().get(key));
+        }).collect(Collectors.toCollection(HashSet::new));
     }
 
     /**
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
@@ -547,18 +547,15 @@ public class HierarchicalDataCommunicator<T> extends DataCommunicator<T> {
         return super.getPassivatedKeys(oldActive).stream().filter(key -> {
             T item = getKeyMapper().get(key);
             if (item != null) {
-                boolean passivateItem = false;
                 T parent = getParentItem(item);
                 while (parent != null) {
                     if (!isItemActive(parent) || !isExpanded(parent)) {
-                        passivateItem = true;
-                        break;
+                        return true;
                     }
                     parent = getParentItem(parent);
                 }
-                return passivateItem;
             }
-            return !isExpanded(getKeyMapper().get(key));
+            return false;
         }).collect(Collectors.toCollection(HashSet::new));
     }
 
--- a/flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorTest.java
+++ b/flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorTest.java
@@ -16,6 +16,7 @@
 package com.vaadin.flow.data.provider.hierarchy;
 
 import java.io.Serializable;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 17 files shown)


flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorTest.java
```diff
@@ -213,6 +216,45 @@ public void replaceAll() {
...
+
+        assertTrue("SECONDROOT key is missing from KeyMapper",
+                communicator.getKeyMapper().has(secondRoot));
+    }
+
+    private void invokeFlush() {
+        try {
+            Method flush = DataCommunicator.class.getDeclaredMethod("flush");
+            flush.setAccessible(true);
+            flush.invoke(communicator);
```
mshabarov, line 252: Did you try to use the same approach as in data communicator test:
```
        dataCommunicator.setRequestedRange(0, 2);
        fakeClientCommunication();
```
this would basically triggers `flush` in the end and you don't have to use reflection.
But if it's not so easy to do with Hierarchical Data Communicator, when we can leave it as is.

mshabarov (changes requested):
Looks good to me.
Tested with all described cases from the original ticket.
Never got a null items, whereas without this patch it was thrown in many cases.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -544,9 +544,22 @@ private JsonValue generateJsonForExpandedOrCollapsedItem(T item) {
...
+                while (parent != null) {
+                    if (!isItemActive(parent) || !isExpanded(parent)) {
+                        passivateItem = true;
+                        break;
+                    }
+                    parent = getParentItem(parent);
+                }
+                return passivateItem;
+            }
+            return !isExpanded(getKeyMapper().get(key));
```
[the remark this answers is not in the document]
tepi: Simplified the whole method a bit

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorTest.java
tepi: Seems like `fakeClientCommunication();` does not work in this case.

mshabarov (approved):

PR #17791, opened 2023-10-06
feat: Show error view on action exception

### LINKED ISSUE #4715: Allow navigating to an error view when an exception occurs after a page is loaded
By implementing `HasErrorParameter<T extends Exception>`, exceptions can be handled if they occur during navigation.

However, it would sometimes be convenient to re-use them even if an exception occurs after navigating, at least to avoid showing the error notification when an unhandled exception occurs.

Even if an error handler is set on the VaadinSession, it is not possible to navigate to an error view from there. Also, the error notification is still shown.

caalador:
If there is a error view for the
thrown exception during rpc handling
update view to error view.

Fixes #4715

The feature is behind a feature flag for testing out the function.


flow-server/src/main/java/com/vaadin/flow/server/DefaultErrorHandler.java
```diff
@@ -72,6 +91,13 @@ public DefaultErrorHandler() {
     public void error(ErrorEvent event) {
         Throwable throwable = findRelevantThrowable(event.getThrowable());
         if (shouldHandle(throwable)) {
+            if (FeatureFlags.get(VaadinService.getCurrent().getContext())
+                    .isEnabled(FeatureFlags.HAS_ERROR_OUTSIDE_NAVIGATION)
+                    && throwable instanceof Exception) {
```
knoobie, line 96: See comment for considering: https://github.com/vaadin/flow/issues/4715#issuecomment-1750168450

flow-server/src/main/java/com/vaadin/flow/server/DefaultErrorHandler.java
```diff
@@ -72,6 +91,13 @@ public DefaultErrorHandler() {
     public void error(ErrorEvent event) {
         Throwable throwable = findRelevantThrowable(event.getThrowable());
         if (shouldHandle(throwable)) {
+            if (FeatureFlags.get(VaadinService.getCurrent().getContext())
+                    .isEnabled(FeatureFlags.HAS_ERROR_OUTSIDE_NAVIGATION)
```
knoobie, line 95: How should the developer opt-in when they have their own ErrorHandler implemented? Extending the DefaultErrorHandler and overwriting the error method? Most people are probably just implementing `ErrorHandler` where no such methods are available to delegate from/to

flow-server/src/main/java/com/vaadin/flow/server/DefaultErrorHandler.java
caalador: We check for exact matchin the drawErrorViewForException as at this point we do not know if there is a handler for the exact error and this check is to just be certain that the throwable is an exception and not an error.

flow-server/src/main/java/com/vaadin/flow/server/DefaultErrorHandler.java
caalador: As the feature itself is not bound to the interface it will most likely become a util class that is used. making it explicitly more opt-in as just testing this out with makes it show how intrusive the feature is when the view redraws when an exception happens.
This state was more for interested parties to try out the feature from behind the feature flag.

flow-server/src/main/java/com/vaadin/flow/server/DefaultErrorHandler.java
knoobie: I see it at line 121 - thanks for confirming the intention!

flow-server/src/main/java/com/vaadin/flow/server/DefaultErrorHandler.java
```diff
@@ -72,6 +91,15 @@ public DefaultErrorHandler() {
     public void error(ErrorEvent event) {
         Throwable throwable = findRelevantThrowable(event.getThrowable());
         if (shouldHandle(throwable)) {
+            VaadinService service = VaadinService.getCurrent();
+            if (service != null
+                    && service.getDeploymentConfiguration()
+                            .isErrorRedirectEnabled()
```
mshabarov, line 97: This new feature should be configured for sure and better to keep the default as is, so turn it off by default. But I'm not so sure that the config parameter is a best option here. 

I would prefer to have a default method in `HasErrorParameter` which returns a boolean for controlling this feature. By default it is false, so no redirecting for non-navigation exceptions. This, as a bonus, would give more flexibility, as they could configure different `HasErrorParameter` differently.

flow-server/src/main/java/com/vaadin/flow/server/DefaultErrorHandler.java
```diff
@@ -72,6 +91,15 @@ public DefaultErrorHandler() {
...
+                    && ErrorHandlerUtil
+                            .handleErrorByRedirectingToErrorView(throwable)) {
```
mshabarov, line 99: How the `ErrorHandler` API is relevant to the original issue?

Maybe I misunderstood the acceptance criteria, but I thought that we want to extend `HasErrorParameter` for non-navigation errors. So basically calling its `setErrorParameter` callback in the same cases/places in code that calls for `ErrorHandler::error()`.

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.server;
 
+import javax.annotation.OverridingMethodsMustInvokeSuper;
```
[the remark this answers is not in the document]
caalador: Yes. Cleaned.

flow-server/src/main/java/com/vaadin/flow/server/DefaultErrorHandler.java
caalador: That has the problem of having to go through the error routes for each error and instantiating the HasErrorParameter view to get the state that is it to be used or not.

flow-server/src/main/java/com/vaadin/flow/server/DefaultErrorHandler.java
caalador: This change does just that. When the error handler gets called we setup the appropriate HasErrorParameter view and call setErrorParameter with the current data and then show the error view on screen.
I don't think calling the error handler and showing the view is what is wanted as then you basically handle the error twice and controlling the errorHandler by making a custom one becomes harder.

flow-server/src/main/java/com/vaadin/flow/server/ErrorHandlerUtil.java
```diff
@@ -0,0 +1,168 @@
...
+    public static boolean handleErrorByRedirectingToErrorView(
+            Exception exception, VaadinContext context, UI ui) {
+        ApplicationRouteRegistry appRegistry = ApplicationRouteRegistry
+                .getInstance(context);
+        Optional<ErrorTargetEntry> errorNavigationTarget = appRegistry
+                .getErrorNavigationTarget(exception);
+        // Found error target and handled exception is the same as thrown
+        // exception. else do not handle
+        if (errorNavigationTarget.isPresent() && errorNavigationTarget.get()
+                .getHandledExceptionType().equals(exception.getClass())) {
```
mshabarov, line 125: How about handling also children classes of the exception?

mshabarov (changes requested):
Naming and javadoc suggestions plus question about child exceptions, otherwise looks good 👍

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/PluginAdapterBase.java
```diff
@@ -293,4 +293,13 @@ default Lookup createLookup(ClassFinder classFinder) {
...
     boolean isPrepareFrontendCacheDisabled();
+
+    /**
+     * If {@code true} navigation error views can be rendered for exceptions
+     * during RPC request handling.
+     *
+     * @return {@code true} to enable error view rendering in RPC, {@code false}
+     *         by default
+     */
+    boolean isErrorHandlerRedirect();
```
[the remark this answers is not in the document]
caalador: done

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java
```diff
@@ -442,4 +445,9 @@ public boolean skipDevBundleBuild() {
     public boolean isPrepareFrontendCacheDisabled() {
         return false;
     }
+
+    @Override
+    public boolean isErrorHandlerRedirect() {
```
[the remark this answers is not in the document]
caalador: done

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/GradlePluginAdapter.kt
```diff
@@ -189,4 +189,6 @@ internal class GradlePluginAdapter(val project: Project, private val isBeforePro
     override fun forceProductionBuild(): Boolean = extension.forceProductionBuild
 
     override fun isPrepareFrontendCacheDisabled(): Boolean = extension.alwaysExecutePrepareFrontend
+
+    override fun isErrorHandlerRedirect(): Boolean = extension.isErrorHandlerRedirect
```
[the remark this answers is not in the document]
caalador: done

flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
```diff
@@ -418,4 +418,9 @@ public boolean skipDevBundleBuild() {
     public boolean isPrepareFrontendCacheDisabled() {
         return false;
     }
+
+    @Override
+    public boolean isErrorHandlerRedirect() {
```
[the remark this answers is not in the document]
caalador: done

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
```diff
@@ -249,6 +249,12 @@ public open class VaadinFlowPluginExtension(project: Project) {
      */
     public var alwaysExecutePrepareFrontend: Boolean = false
 
+    /**
+     * If {@code true} navigation error views can be rendered for exceptions during RPC request handling.
```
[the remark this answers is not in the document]
caalador: changed

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/PluginAdapterBase.java
```diff
@@ -293,4 +293,13 @@ default Lookup createLookup(ClassFinder classFinder) {
      * `prepare-frontend` goal is supported.
      */
     boolean isPrepareFrontendCacheDisabled();
+
+    /**
+     * If {@code true} navigation error views can be rendered for exceptions
```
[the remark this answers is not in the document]
caalador: added

flow-server/src/main/java/com/vaadin/flow/server/ErrorHandlerUtil.java
caalador: The initial wish was at least for exact matches only. So as I see it extending children should not go through there. This could perhaps be seen as an enhancement to the feature if requested enough.

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/GradlePluginAdapter.kt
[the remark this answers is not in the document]
mshabarov: Not done apparently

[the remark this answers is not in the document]
caalador: I feel cheated by idea.

mshabarov (approved):

PR #17800, opened 2023-10-09
Revert "chore: update jackson to 2.15.2 (#17784)"

manolo:


mshabarov (approved):
Reverted, because this produces a few breaking changes:
- 2.15 requires Gradle 7.6+
- 2.15 requires Kotlin 1.5+

PR #17801, opened 2023-10-09
Revert "chore(deps): bump biz.aQute.bnd:bnd-maven-plugin from 6.4.0 to 7.0.0 (#17792)"

ZheSun88:
This reverts commit 236156bf232040ae6c75fa045deebaa38d400573.


mshabarov (approved):
Reverted, because this requires Maven 3.8+, which is a breaking change for our CI and might be a breaking change for users.

Issue #2963, opened 2017-11-17
JavaScript exceptions should not be swallowed in production mode


Artur-:
If you do `page.executeJavaScript("lkajsfd");` you will get a `SyntaxError` error message in the console when running in development mode (in addition to a popup) but in production mode you get absolutely nothing and have no clue that the JS failed and is causing problems.

SomeoneToIgnore:
Logging was removed due to https://github.com/vaadin/flow/issues/2028

I think even the error messages can be used to print critical information also hence this ticket should be closed.

Artur-:
#2028 was about logging what was going to be executed, I don't see any mention of exceptions in that

Legioth:
`console.error` sounds like a quite logical place of logging for exceptions, since that's where they would end up naturally if a similar error would be in a regular JS file that is included into the page.

paulroemer:
As we just got hit by the issue on vaadin.com I can tell you that logging to `console.error` isn't enough because we cannot monitor the error log of our clients.
If you would log it to the server log we could filter for those warnings/errors and get informed about it immediately.

Legioth:
It would be quite obtrusive (and has the potential of causing infinite exception handling loops) for Flow to try to submit exceptions to the server.

What Flow could do is to provide a hook for plugging in your own exception handler.

paulroemer:
We can of course start with a hook so that we can experiment with the feature and give you feedback. But on the other hand it feels crucial to me that the developer is (at least) informed about template model related issues in the server logs without the need to register a special handler.

It would make the support team's life much easier, too. Reusing the example above: I would ask for the server log, notice that there are warnings about template issues and first ask them to fix that.

Legioth:
I agree that it would help simplify many cases, but I'm also worried that it might cause quite spectacular explosions in other cases unless the implementation is flawless.

Legioth:
Could you create a separate ticket about an hook and/or possibility to send information to the server, so that this ticket could remain focused on the original request of making the errors available in the browser console?

paulroemer:
See #3708

fluorumlabs:
Any news on that? It's getting extremely hard to debug in IE11.

denis-anisimov:
No any news.
It's an enhancement and has a label `needs design` and in the staging backlog.
It means that  it has the lowest priority.

pleku:
Debugging IE11 will become much easier in 13 for project using npm since you will not need to run production mode for IE11.

[...]

PR #3858, opened 2018-04-11
Log the actual cause of errors from execJavaScript

Legioth:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/flow/3858)

PR #17805, opened 2023-10-09
Log JS exceptions in production mode

tepi:
Fixes https://github.com/vaadin/flow/issues/2963

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/Console.java
+++ b/flow-client/src/main/java/com/vaadin/client/Console.java
@@ -22,8 +22,8 @@ import com.google.gwt.core.client.JavaScriptException;
 import elemental.client.Browser;
 
 /**
- * Helper class for using window.console. Does not log anything to console if
- * production mode is enabled.
+ * Helper class for using window.console. Does not log anything except
+ * JavaScript exception traces to console if production mode is enabled.
  *
  * @author Vaadin Ltd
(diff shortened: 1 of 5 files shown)


caalador:
Only thing would be to have a test in a production mode module.
See:
https://github.com/vaadin/flow/pull/3858/files#diff-a1fe593dd03bd0c91ae9e5eb4394e31d21f6a733bb9f026dad551fcab20280e8 and https://github.com/vaadin/flow/pull/2101/files#diff-0b466c731aecc2544bbe3f4582973ddcf82366a19104cf2f246813b21e977399
that combined would sound like a good test case.
For some reason no tests are left for the logging case in production mode which is a shame as we could just mistakenly break the feature. Tested with the old `ClientSideExceptionHandlingView.java` (now not existing) in production mode and the fix does do what it should.

caalador:
`flow-misc-test` module is running production mode so that could be a good place to put the test in.

tepi:
> For some reason no tests are left for the logging case in production mode which is a shame as we could just mistakenly break the feature.

I added tests for the exception cases. Did you mean that we don't have any tests for prod mode logging? Not even to check that debug/info is not logged?

tepi:
Ok, thanks. It's a bit unrelated to this issue so I'll create a separate ticket about that

caalador (approved):

PR #17823, opened 2023-10-12
feat: Default I18N provider

### LINKED ISSUE #5917: Provide a default I18NProvider
Why don’t we provide a default I18NProvider? so that user could just put some resource bundle files like translate_fi_FI.properties and translate_en_GB.properties in src/main/resources and then just call getTranslation() on components?

caalador:
Add implementation for a I18N
provider that is loaded if there
are translation properties files
in the defined classpath folder.

If a custom I18N is set default will
never be loaded.

Fixes #5917


flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
```diff
@@ -0,0 +1,105 @@
...
+            getLogger().warn("Got lang request for key with null value!");
+            return "";
+        }
+
+        final ResourceBundle bundle = getBundle(locale);
+
+        String value;
+        try {
+            value = bundle.getString(key);
+            value = new String(value.getBytes("ISO-8859-1"), "UTF-8");
```
knoobie, line 69: With Vaadin 24.3 enforcing Java 17+, I think the usage of ISO-8859-1 with resourcebundle should not be used anymore. By default it should be UTF-8 or something else, depending on the used system property: [JavaDoc PropertyResourceBundle](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/PropertyResourceBundle.html)

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
```diff
@@ -0,0 +1,105 @@
...
+        } catch (final MissingResourceException e) {
+            getLogger().warn("Missing resource", e);
+            return "!" + locale.getLanguage() + ": " + key;
+        } catch (UnsupportedEncodingException e) {
+            getLogger().warn("Failed to parse encoding of value", e);
+            return key;
+        }
+        if (params.length > 0) {
+            value = MessageFormat.format(value, params);
```
knoobie, line 78: Would it make sense to move this in a protected method to allow overwriting this to e.g. implement a caching mechanism? MessageFormat can be quite expensive.

knoobie, line 78: Should this be changed to something like: 

```java
      MessageFormat messageFormat = new MessageFormat(value, locale);
      value = messageFormat.format(params);
```

so that the used locale is propagated to the formatter?

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        final ResourceBundle bundle = getBundle(locale);
+
+        String value;
+        try {
+            value = bundle.getString(key);
+            value = new String(value.getBytes("ISO-8859-1"), "UTF-8");
+        } catch (final MissingResourceException e) {
+            getLogger().warn("Missing resource", e);
```
knoobie, line 68: Nit: I feel like "ups you are missing a translation label" is not severe enough to justify a warn-level. I would suggest to lower it to info or even debug.

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
caalador: I don't think the default implementation brings in so much functionality that it would make sense to override it instead of just implementing your own I18NProvider as you need to anyway set it as you do at the moment.

caalador: done

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
caalador: lowered

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
knoobie: I just wanted to ask if you wanna also implement a PropertyPlaceholderResolver in here, with your comment.. I think I know the answer :D Reference of the resolver used by Spring: https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/util/PropertyPlaceholderHelper.java

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public String getTranslation(String key, Locale locale, Object... params) {
+        if (key == null) {
+            getLogger().warn("Got lang request for key with null value!");
+            return "";
+        }
+
+        final ResourceBundle bundle = getBundle(locale);
```
knoobie, line 62: this can return null -> throwing a NPE later on -> should there be a null check and return ""?

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
```diff
@@ -0,0 +1,91 @@
...
+        if (params.length > 0) {
+            value = new MessageFormat(value, locale).format(params);
+        }
+        return value;
+    }
+
+    private ResourceBundle getBundle(Locale locale) {
+        try {
+            return ResourceBundle.getBundle(BUNDLE_PREFIX, locale,
+                    I18NUtil.getClassLoader());
```
knoobie, line 80: Should this be intelligent enough so that e.g. a missing Resourcebundle for a locale fallsback to other languages? 

We have something like this:

- Passing in GERMAN
- No ResourceBundle found for GERMAN? Return Locale.ROOT (default resource bundle) 

- Passing in GERMAN
- Resourcebundle found for GERMAN
- Resourcebundle does not contain a translation for the key.. search in ResourceBundle Locale.ROOT for the key
- key in ROOT found -> return the root translation + debug/info logging 
- key ALSO not in ROOT found -> return key + warn logging

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
caalador: True if there is no default translation file available.
Returning key if no bundle available as the missing bundle will get a warn log.

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.MissingResourceException;
+import java.util.ResourceBundle;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Default i18n provider that will be initialized if
+ */
+public class DefaultI18NProvider implements I18NProvider {
```
knoobie, line 36: General Question: Should this class be registered as Spring Bean with `ConditionalOnMissingBean` so that it's also available for DI in Spring as default and people can inject / use it even without always asking the Instantiator for it.

caalador: The springInitiator actually falls back on the defaultInitiator on this if no bean found:
```
    public I18NProvider getI18NProvider() {
        int beansCount = context.getBeanNamesForType(I18NProvider.class).length;
        if (beansCount == 1) {
            return context.getBean(I18NProvider.class);
        } else {
            if (loggingEnabled.compareAndSet(true, false)) {
                LoggerFactory.getLogger(SpringInstantiator.class.getName())
                        .info("The number of beans implementing '{}' is {}. Cannot use Spring beans for I18N, "
                                + "falling back to the default behavior",
                                I18NProvider.class.getSimpleName(), beansCount);
            }
            return super.getI18NProvider();
        }
    }
```

caalador: Also having Spring annotations is a bit problematic as this is outside of spring in the normal framework.

knoobie: Thanks for the explanation! Yeah, I was wondering if it would be worth to add it to the VaadinApplicationConfiguration like a bean:

```java
  @Bean
  @ConditionalOnMissingBean(value = I18NProvider.class)
  public DefaultI18NProvider vaadinI18nProvider() {
    return new DefaultI18NProvider(/* get locales from somewhere*/);
  }

```

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
caalador: Well there is an inherent default here that if you have 
`translations_fi_FI.properties` and `translations.properties` requesting for the Finnish locale `Locale("fi","FI)` will return you the fi_FI bundle, but then requesting for instance `Locale("de")` will return the `translations.properties` which is basically then the ROOT (or default) bundle.

caalador: Recursing bundles down is perhaps not inside the feature scope for this default implementation, but could be a later enhancement

knoobie: Understandable :) (can be closed)

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
```diff
@@ -0,0 +1,88 @@
...
+    public static List<Locale> getDefaultTranslationLocales() {
+        List<Locale> locales = new ArrayList<>();
+
+        URL resource = getClassLoader()
+                .getResource(DefaultI18NProvider.BUNDLE_FOLDER);
+        if (resource == null) {
+            return locales;
+        }
+        File bundleFolder = new File(resource.getFile());
+        if (bundleFolder.exists() && bundleFolder.isDirectory()) {
```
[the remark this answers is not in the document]
mcollovati: And it seems not to work even if I put the translations in a JAR file, although I don't know if this can be considered a valid use case for this feature.

```
bundleFolder= file:/xxxxxxxxx/vaadin/starters/skeleton-starter-flow/v-i18n.jar!/vaadin-i18n
bundleFolder.exists() = false
```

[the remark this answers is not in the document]
caalador: Good Point. I'll look into it and fix the jar path. It will require using the FileSystem, but should be doable.

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
```diff
@@ -0,0 +1,94 @@
...
+
+    final List<Locale> providedLocales;
+
+    public static final String BUNDLE_FOLDER = "vaadin-i18n";
+    public static final String BUNDLE_FILENAME = "translations";
+    // Get bundles named `translations` from `translation` folder.
```
tltv, line 42: Shouldn't this say "from `vaadin-i18n` folder" instead.

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
```diff
@@ -0,0 +1,88 @@
...
+            List<File> listedFiles = Arrays.stream(bundleFolder.listFiles())
+                    .filter(file -> file.isFile())
+                    .filter(file -> file.getName().endsWith(PROPERTIES_SUFFIX))
```
tltv, line 56: This may need to also filter by `file.getName().startsWith(BUNDLE_FILENAME)` to ignore other files than translation files.

tltv (changes requested):

flow-server/src/test/java/com/vaadin/flow/i18n/DefaultInstantiatorI18NTest.java
```diff
@@ -0,0 +1,223 @@
...
+            I18NProvider i18NProvider = instantiator.getI18NProvider();
+            Assert.assertNotNull(i18NProvider);
+            Assert.assertTrue(i18NProvider instanceof DefaultI18NProvider);
+
+            Assert.assertEquals("Suomi", i18NProvider.getTranslation("title",
+                    new Locale("fi", "FI")));
+
+            Assert.assertEquals("deutsch",
+                    i18NProvider.getTranslation("title", new Locale("de")));
+
```
mcollovati, line 105: It could be nice to add an assertion to ensure that the base language bundle is taken if we provide the also country and the specific bundle file does not exist.
Of course, not a blocker

```
            Assert.assertEquals("deutsch",
                    i18NProvider.getTranslation("title", new Locale("de", "AT")));
```

flow-server/src/test/java/com/vaadin/flow/i18n/I18NUtilTest.java
```diff
@@ -0,0 +1,189 @@
...
+            util.when(() -> I18NUtil.getDefaultTranslationLocales())
+                    .thenCallRealMethod();
+            util.when(() -> I18NUtil.containsDefaultTranslation())
+                    .thenCallRealMethod();
+
+            Assert.assertFalse("Nothing should be returned for empty folder",
+                    I18NUtil.containsDefaultTranslation());
+        }
+    }
+}
```
mcollovati, line 189: nit: new line missing
```suggestion
}

```

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
```diff
@@ -0,0 +1,88 @@
...
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.stream.Collectors;
+
+import static com.vaadin.flow.i18n.DefaultI18NProvider.BUNDLE_FILENAME;
+
+public final class I18NUtil {
```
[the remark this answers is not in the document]
caalador: added javadoc. 
This is basically internal, but it is used from outside package.

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
caalador: added

flow-server/src/test/java/com/vaadin/flow/i18n/I18NUtilTest.java
caalador: added

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
caalador: fixed

flow-server/src/test/java/com/vaadin/flow/i18n/DefaultInstantiatorI18NTest.java
caalador: Added tests on this, even though it is a `ResourceBundle` feature.

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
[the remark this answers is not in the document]
caalador: Now it knows how to handle the files inside jars also.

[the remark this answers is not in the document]
mcollovati: Works great with the resource bundle in JAR!

However, is still fails with the Spring Boot fat JAR, both by adding the properties files in the project or in external JAR.

```
bundleFolder = {File@6570} "file:/xxxx/skeleton-starter-flow-spring/target/spring-skeleton-1.0-SNAPSHOT.jar!/BOOT-INF/classes!/vaadin-i18n"
pathInJar = "/BOOT-INF/classes!/vaadin-i18n"


bundleFolder = {File@8348} "file:/xxxx/skeleton-starter-flow-spring/target/spring-skeleton-1.0-SNAPSHOT.jar!/BOOT-INF/lib/i18n-1.0.jar!/vaadin-i18n"
pathInJar = "/BOOT-INF/lib/i18n-1.0.jar!/vaadin-i18n"
```

For the first case, removing the `!` from `pathInJar` makes the solution work. The second case is more tricky.
I wonder if it could be better to provide some specific implementations in the Spring add-on, or if we have to find some hack 
 to add in the current implementation  :thinking:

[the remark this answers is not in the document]
knoobie: I would suggest to go with a spring based solution in the spring module based on 

ResourcePatternResolver allowing to get all resources within a folder in a spring boot application within one line of code - foul and future proof :) relying on it for years to load our default configurations

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+        return value;
+    }
+
+    private ResourceBundle getBundle(Locale locale) {
+        try {
+            return ResourceBundle.getBundle(BUNDLE_PREFIX, locale,
+                    I18NUtil.getClassLoader());
+        } catch (final MissingResourceException e) {
+            getLogger().warn("Missing resource", e);
```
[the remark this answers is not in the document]
caalador: added

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
```diff
@@ -0,0 +1,185 @@
...
+
+                if (Files.exists(dirPath) && Files.isDirectory(dirPath)) {
+                    try (Stream<Path> jarFiles = Files.list(dirPath)) {
+                        jarFiles.filter(path -> !Files.isDirectory(path))
+                                .map(Path::toString).map(File::new)
+                                .forEach(files::add);
+                    }
+                }
+            } catch (IOException | URISyntaxException e) {
+                getLogger().debug("failed to read jar file contents", e);
```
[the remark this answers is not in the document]
caalador: As there is no need to actually handle files I removed fileSystem usage and we just handle jar entry file names as that is enough. Updated exception to contain file name

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    protected static FileSystemPair getNewOrExistingFileSystem(URI resourceURI)
+            throws IOException {
+        try {
+            return new FileSystemPair(FileSystems.newFileSystem(resourceURI,
+                    Collections.emptyMap()), false);
+        } catch (FileSystemAlreadyExistsException fsaee) {
+            getLogger().trace(
+                    "Tried to get new filesystem, but it already existed for target uri.",
```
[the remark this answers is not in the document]
caalador: removed fileSystem usage so this log doesn't exist anymore.

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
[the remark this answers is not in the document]
caalador: Seeing as this was supposed to be a small enhancement, I would propose to open a new ticket for creating Spring-Boot fat Jar support by extending the default translation with Spring parts in the spring add-on

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
```diff
@@ -0,0 +1,159 @@
...
+                entries.asIterator().forEachRemaining(entry -> {
+                    String fileName = entry.getName();
+                    if (fileName.contains(BUNDLE_FOLDER)
+                            && fileName.endsWith(PROPERTIES_SUFFIX)) {
+                        files.add(new File(fileName));
+                    }
+                });
+            } catch (IOException ioe) {
+                getLogger().debug("failed to read jar file '"
+                        + file.substring(6, file.indexOf('!')) + "'contents",
```
mcollovati, line 141:
```suggestion
                        + file.substring(6, file.indexOf('!')) + "' contents",
```

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
```diff
@@ -0,0 +1,159 @@
...
+    protected static List<File> getTranslationFiles(URL resource) {
+        List<File> files = new ArrayList<>();
+
+        File bundleFolder = new File(resource.getFile());
+
+        if ("jar".equals(resource.getProtocol())) {
+            String file = resource.getFile();
+            try {
+                Enumeration<JarEntry> entries = new JarFile(
+                        file.substring(6, file.indexOf('!'))).entries();
```
mcollovati, line 131: The substring seems not to work correctly on Linux

```
resource : jar:file:/home/marco/projects/vaadin/starters/skeleton-starter-flow/v-i18n.jar!/vaadin-i18n
file: file:/home/marco/projects/vaadin/starters/skeleton-starter-flow/v-i18n.jar!/vaadin-i18n
substring: home/marco/projects/vaadin/starters/skeleton-starter-flow/v-i18n.jar


[qtp14050562-55] DEBUG com.vaadin.flow.i18n.I18NUtil - failed to read jar file 'home/marco/projects/vaadin/starters/skeleton-starter-flow/v-i18n.jar'contents
java.nio.file.NoSuchFileException: home/marco/projects/vaadin/starters/skeleton-starter-flow/v-i18n.jar
```

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
caalador: i have that fixed already, but running the test on WSL is really slow.

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/i18n/I18NUtil.java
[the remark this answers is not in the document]
mcollovati: It looks like we need a ticket as well for the Quarkus add-on.
The class loader used to detect the bundle folder (` I18NUtil.class.getClassLoader()`) returns null, whereas it works using `Thread.currentThread().getContextClassLoader()`
Would it be possible to replace the `protected static ClassLoader getClassLoader()` method with a non-static version, so it can be overridden to provide the best class loader for each environment?
Or can we have the class loader reference as class member, provided by an additional constructor?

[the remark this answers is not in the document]
caalador: We can use current as that's not a problem. Sending a loader around becomes too complex for testing from instantiator

mcollovati (approved):

Issue #5959, opened 2019-06-21
Button click shortcut listener does not recognize value change when using control key


stefanuebe:
A simple use case: a form with some text fields, a rich text editor and a save button. The user changes stuff and wants to submit the form by using the ctrl + enter key.

The text fields are used as they are (value change mode ON_CHANGE), bound with a simple binder. To achieve the "on enter", we set a click shortcut with ENTER and CONTROL.

Now, when filling all fields and using the mouse to click the button, everything works fine.
When filling the fields and using the shortcut, the currently focused field does not get submitted.
Using a self defined shortcut listener and the button method clickInClient does not solve the problem

The problem seems to be the usage of the control key. When setting the short cut listener to ENTER only, it works finely (except for that the rich text editor focus needs to handle additional checkings)

Setting the value change mode to EAGER makes it work as a workaround.

stefanuebe:
Tested with Vaadin 14, RC1 in bower mode on a Windows server and Windows client with chrome.

```
package org.vaadin.issue;

import com.vaadin.flow.component.AttachEvent;
import com.vaadin.flow.component.Key;
import com.vaadin.flow.component.button.Button;
import com.vaadin.flow.component.dialog.Dialog;
import com.vaadin.flow.component.html.Span;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.component.textfield.TextArea;
import com.vaadin.flow.component.textfield.TextField;
import com.vaadin.flow.data.binder.Binder;
import com.vaadin.flow.router.Route;

@Route(value = "button")
public class ButtonChangeView extends VerticalLayout {

    private SimpleBean simpleBean = new SimpleBean();

    @Override
    protected void onAttach(AttachEvent attachEvent) {

        VerticalLayout info = new VerticalLayout(new Span("Results"));
        Button open = new Button("Open", event -> {
            Dialog dialog = new Dialog();

            TextField aField = new TextField("A");
            TextField bField = new TextField("B");
            TextArea cField = new TextArea("C");

            Binder<SimpleBean> binder = new Binder<>();
            binder.bind(aField, SimpleBean::getA, SimpleBean::setA);
            binder.bind(bField, SimpleBean::getB, SimpleBean::setB);
            binder.bind(cField, SimpleBean::getC, SimpleBean::setC);

            binder.readBean(simpleBean);

            Button save = new Button("Save", event1 -> {
[...]

PR #17826, opened 2023-10-12
feat: reset focus on active element in ShortcutRegistration

tltv:
Adds `setResetFocusOnActiveElement`, `resetFocusOnActiveElement` and `isResetFocusOnActiveElement` to `ShortcutRegistration` class to optionally reset focus on active element (focused element in browser). Resetting means calling `blur()` and `focus()` on the active element to ensure that input fields with `ValueChangeMode` `ON_CHANGE` will fire value change event before shortcut handler is run. Updates `ClickNotifier#addClickShortcut(Key,KeyModifier...)` to reset focus on active element by default for click shortcuts.

Fixes: #5959

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/ShortcutRegistration.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/ShortcutRegistration.java
@@ -54,16 +54,27 @@ public class ShortcutRegistration implements Registration, Serializable {
                     + "if (delegate) {"
                     + "delegate.addEventListener('keydown', function(event) {"
                     + "if (%2$s) {" // the filter text to match the key
+                    + "%3$s" // allow reset focus on active element if desired
                     + "const new_event = new event.constructor(event.type, event);"
                     + "listenOn.dispatchEvent(new_event);"
-                    + "%3$s" // the new event allows default if desired
+                    + "%4$s" // the new event allows default if desired
                     + "event.stopPropagation();}" // the new event bubbles if desired
                     + "});" // end matches filter
                     + "} else {"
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/component/ShortcutRegistration.java
```diff
@@ -54,16 +54,27 @@ public class ShortcutRegistration implements Registration, Serializable {
...
                     + "} else {"
                     + "throw \"Shortcut listenOn element not found with JS locator string '%1$s'\""
                     + "}";//@formatter:on
+    static final String FOCUS_ACTIVE_ELEMENT_JS = //@formatter:off
+        "function(){" +
+                "let ae=document.activeElement;" +
+                "while(ae&&ae.shadowRoot) ae = ae.shadowRoot.activeElement;" +
+                "return !ae || ae.blur() || ae.focus() || true;" +
+                "}()";//@formatter:on
+
```
caalador, line 72: One main thing would be can we have this script added instead to AbstractUpdateImports so that it is always pushed once to the client instead of for each shortcut.
So we would have the function available under `window.Vaadin.Flow.resetFocus`
e.g. in AbstractUpdateImports after loadOnDemand 
```
mainLines.add("window.Vaadin.Flow.resetFocus = () => { 
    let ae=document.activeElement;
    while(ae&&ae.shadowRoot) ae = ae.shadowRoot.activeElement;
    return !ae || ae.blur() || ae.focus() || true;
}");
```
Then in the build scripts we would only add `window.Vaadin.Flow.resetFocus()` instead of the full script.

caalador (approved):

PR #17833, opened 2023-10-13
test: Add tests for dev mode client-side logging; Clarify prod mode test

tepi:
Could not find a way to hook into methods of `com.vaadin.client.Console` without exporting them to JS  which we do not want. So only debug, info and error level output is checked via known log messages.

### DIFF
--- a/flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/frontend/BrowserLoggingIT.java
+++ b/flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/frontend/BrowserLoggingIT.java
@@ -33,7 +33,7 @@ import static org.hamcrest.number.OrderingComparison.greaterThan;
 public class BrowserLoggingIT extends ChromeBrowserTest {
 
     @Test
-    public void nonProductionModeHasLogEntries() {
+    public void developmentModeHasLogEntries() {
         open();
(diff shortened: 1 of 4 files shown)


flow-tests/test-dev-mode/src/test/java/com/vaadin/flow/uitest/ui/frontend/BrowserLoggingIT.java
```diff
@@ -33,15 +35,30 @@ public class BrowserLoggingIT extends ChromeBrowserTest {
...
+        findElement(By.id("exception")).click();
+
+        ArrayList<Object> logMessages = (ArrayList<Object>) executeScript(
+                "return window.allLogMessages;");
 
         assertThat(
-                "Flow in production mode should output nothing into the console",
-                getLogEntriesCount(), greaterThan(0L));
-    }
+                "Flow in non-production mode should output something into the console",
```
mcollovati, line 46: nit: I would use `development mode`, as we usually use that term
```suggestion
                "Flow in development mode should output something into the console",
```

mcollovati (commented):
Only a minor not blocking comment, other than that LGTM

mcollovati:
Just notice that the `UrlValidationIT` test is using `BrowserLoggingView` and now it fails because it looks for an element with id `elementId`

mcollovati (approved):

Issue #11335, opened 2021-06-25
Enable rerouting to a custom page if a View-based access is denied


mshabarov:
### Description of the bug / feature
View-based access control feature introduced in V21 checks whether the user's has an appropriate role to browse the secured page via `@RolesAllowed` and makes `rerouteToError(NotFoundException.class)` if it's not the case, showing `Could not navigate to ...` message (in production mode).
While this is a more secured way than report `Access Denied` message, it's not always necessary in all cases.
This enhancement considers providing an API to set the custom page the user is navigated to, if he has no access to requested page.
Here are a couple possible use cases:
- routing the users to a 'public' version of protected page.
- showing a prompt like 'Please subscribe if you want to see this protected/subscriber-only content' or other suggestion on how he could upgrade his authorities to browse the page.

Some thoughts on how it can be achieved:
- Add a new `class RouteAccessDeniedError extends Component implements HasErrorParameter<AccessDeniedException>` default error handler (or to extend the existing error handlers) and reroute to `AccessDeniedException` in the `ViewAccessChecker`. Then the developers can extend this error handler to add any prompts/styles there.
- Configure the custom route/view class in the same place where the secured annotations like `@RolesAllowed` are configured. For example, by giving an extra annotation or an extra parameter to existing annotation. It would give a flexibility that some pages would have their 'public' versions, and the rest would be reported as 'not found' by default.

chkpnt:
Yes, please reroute to `AccessDeniedException`. It's easy to configure it, that it behaves like a `NotFoundException`, but not vice versa.

tltv:
Just adding new default `RouteAccessDeniedError` page would make it easier to redirect to public page or customize error page.
 
Adding new attribute like `String unauthorizedRerouteTo` and/or `String unauthorizedForwardTo` in `RolesAllowed` annotation would make it easier as you could redirect to any route before any error page is loaded. 
Also adding `Class<? extends Exception> unauthorizedRerouteToError`  would allow overriding default `AccessDeniedException` e.g. with `NotFoundException` to fall back to how it works now.

tltv:
[PR](https://github.com/vaadin/flow/pull/17843) adds [new annotation called `AccessDeniedErrorRouter`](https://github.com/vaadin/flow/blob/2c23d70add0fb3d4898a7392902746d339c60739/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessDeniedErrorRouter.java) for route specific control over which error page to show for access denied errors.

PR #17843, opened 2023-10-17
feat: add AccessDeniedException for ViewAccessChecker

tltv:
`ViewAccessChecker` is changed to throw `AccessDeniedException` instead of `NotFoundException` by default when access is denied. Added default error page RouteAccessDeniedError for the AccessDeniedException. Page is same as `RouteNotFoundError` by default (extends `AbstractRouteNotFoundError`). Added `AccessDeniedErrorRouter` annotation to customize rerouting of access denied error per route with `AccessDeniedErrorRouter#rerouteToError`.

Fixes: #11335


flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
```diff
@@ -0,0 +1,173 @@
...
+        template = template.replace("{{additionalInfo}}", additionalInfo);
+        template = template.replace("{{path}}", path);
+
+        getElement().setChild(0, new Html(template).getElement());
+        return HttpStatusCode.NOT_FOUND.getCode();
+    }
+
+    private static Logger getLogger() {
+        return LoggerFactory.getLogger(RouteNotFoundError.class);
+    }
```
caalador, line 109: Should this be protected abstract so the actual error classes can give the correct class to the logger?

flow-server/src/test/java/com/vaadin/flow/router/RouterTest.java
```diff
@@ -2664,6 +2696,25 @@ public void custom_exception_target_should_override_default_ones() {
                 EXCEPTION_TEXT);
     }
 
+    @Test
+    public void custom_access_denied_exception_target_should_override_default_ones() {
+        setNavigationTargets(FailOnAccessDeniedException.class);
+        setErrorNavigationTargets(NonExtendingAccessDeniedTarget.class,
+                RouteNotFoundError.class);
```
caalador, line 2703:
```suggestion
                RouteAccessDeniedError.class);
```
This should be the access denied default view so we see that it the custom one overrides the default

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
tltv: Method is not needed at all actually. Used in two places, one should actually log for `AbstractRouteNotFoundError.class` and other should log for the `this.getClass()`.

caalador: Shouldn't the log outputs be for RouteAccessDeniedError and RouteNotFoundError. So `this.getClass()` should be fine enough

tltv: Static `readHtmlFile` logs about missing template file which could rather log from the exact class `AbstractRouteNotFoundError` instead of subclass. Other logging is about missing route where it makes sense to point to the subclass.

caalador (approved):

PR #17861, opened 2023-10-19
feat: Use binding object to check for Binder#hasChanges

### LINKED ISSUE #17395: Add a method to Binder to check if a specific Binding changed
### Describe your motivation

Inside our binder, we use a custom field for address input. If the values in it change, I need to persist the new address object before I can persist the binder bean. So I need to be able to find out if the address field has changed. Binder#hasChanged() only tells me if any field of the binder has changed.

The binder already has a set of changed bindings https://github.com/vaadin/framework/blob/master/server/src/main/java/com/vaadin/data/Binder.java#L1536. 
We can use this to add a method hasChanges(Binding<BEAN, ?> binding).

In the perfect word it would also remove the "change state" if the value is changed back to the original one but I don't see that this is currently a feature of the Binder.

Building the logic myself using the value change listener.

### Additional context

Is there anything else you can add about the proposal?

benedikt-roth:
## Description

Allows to check whether specific binding has been changed. So far there is only a generic Binder#hasChanges, which checks all bindings. The new method allows to check whether a specific binding has changes.

Fixes #17395

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -3035,7 +3035,7 @@ public class Binder<BEAN> implements Serializable {
     }
 
     /**
-     * Checks whether a field has uncomitted changes.
+     * Checks whether a bound field has uncomitted changes.
      *
      * @param binding
      *            Binding to be checked
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -3035,6 +3035,16 @@ public class Binder<BEAN> implements Serializable {
     }
 
     /**
+     * Checks whether a field has uncomitted changes.
+     *
+     * @param binding Binding to be checked
+     * @return true if field has been changed.
+     */
+    public boolean hasChanges(Binding<BEAN, ?> binding) {
+        return hasChanges() && changedBindings.contains(binding);
+    }
+
+    /**
      * Sets the read only state to the given value for all currently bound
      * fields.
      * <p>
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -290,7 +290,7 @@ public class Binder<BEAN> implements Serializable {
          * @throws IllegalStateException
          *             if the binding is no longer attached to a Binder.
          *
-         * @return True if field has uncommitted changes
+         * @return {@literal true} if the field the binding uses has uncommitted changes, otherwise {@literal false}.
          */
         boolean hasChanges();
     }
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -282,6 +282,18 @@ public class Binder<BEAN> implements Serializable {
          * @return A boolean value
          */
         boolean isConvertBackToPresentation();
+
+        /**
+         * Checks whether the related Binder instance has marked the field as
+         * having uncommitted changes.
+         *
+         * @throws IllegalStateException
+         *             if {@code binder} is null. Ensure that the binder has
+         *             been declared here.
+         *
+         * @return True if field has uncommitted changes
+         */
+        boolean hasChanges() throws IllegalStateException;
     }
 
     /**
@@ -1588,6 +1600,15 @@ public class Binder<BEAN> implements Serializable {
                     throw exception;
                 }
             }
+        }
+
+        @Override
+        public boolean hasChanges() throws IllegalStateException {
+            if (this.binder == null) {
+                throw new IllegalStateException();
+            }
+
+            return this.binder.hasChanges(this);
         }
     }
 
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -290,7 +290,8 @@ public class Binder<BEAN> implements Serializable {
          * @throws IllegalStateException
          *             if the binding is no longer attached to a Binder.
          *
-         * @return {@literal true} if the field the binding uses has uncommitted changes, otherwise {@literal false}.
+         * @return {@literal true} if the field the binding uses has uncommitted
+         *         changes, otherwise {@literal false}.
          */
         boolean hasChanges();
     }
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -3037,7 +3037,8 @@ public class Binder<BEAN> implements Serializable {
     /**
      * Checks whether a field has uncomitted changes.
      *
-     * @param binding Binding to be checked
+     * @param binding
+     *            Binding to be checked
      * @return true if field has been changed.
      */
     public boolean hasChanges(Binding<BEAN, ?> binding) {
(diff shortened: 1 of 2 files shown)


mcollovati:
@benedikt-roth do you mind signing the CLA and formatting the source code as suggested in https://github.com/vaadin/flow/pull/17861#issuecomment-1771359569?

benedikt-roth:
> @benedikt-roth do you mind signing the CLA and formatting the source code as suggested in [#17861 (comment)](https://github.com/vaadin/flow/pull/17861#issuecomment-1771359569)?

Sure, done that and also fixed the formatting.

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
```diff
@@ -199,6 +199,29 @@ public void clearBean_setsHasChangesToFalse() {
...
     }
 
+    @Test
+    public void writeBean_setsHasChangesBindingToFalse() {
+        var binding = binder.forField(nameField).bind(Person::getFirstName,
+                Person::setFirstName);
+
+        binder.readBean(item);
+        assertEquals("No name field value", "Johannes", nameField.getValue());
+
```
mcollovati, line 209: We could add a check that `hasChanges` is false just after reading the bean

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -3034,6 +3034,17 @@ public boolean hasChanges() {
         return !changedBindings.isEmpty();
     }
 
+    /**
+     * Checks whether a field has uncomitted changes.
```
mcollovati, line 3038:
```suggestion
     * Checks whether a bound field has uncomitted changes.
```

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
```diff
@@ -199,6 +199,29 @@ public void clearBean_setsHasChangesToFalse() {
 
     }
 
+    @Test
+    public void writeBean_setsHasChangesBindingToFalse() {
```
mcollovati, line 203: The test name doesn't clearly reflect what we are testing, since `writeBean` is never invoked.
Consider changing it to some like `bindinghasChanges_trueWhenFieldValueChanges`
Naming is hard :grin:

mcollovati (changes requested):
Thanks for the contribution, the change looks good.
I would add tests for `hasChanges` when the binding has been removed from the binder.

I also wonder if we could add the `hasChanges()` method also on the `Binding` interface, so that it would be possible to simply do a `binding.hasChanges()`.
The implementation in `BindingImpl` could use the internal `binder` field to check it `this` binding is in the `changeBindings` set.
And it will throw and `IllegalStateException` if the `binder` is null.

However, this is not mandatory for the PR to be merged

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
benedikt-roth: Ah, I changed this multiple times. You are right, this does not really explain what is happening.
Your suggestion looks good. Will use that one.

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
benedikt-roth: Good idea. The more checks the better for this logic.

benedikt-roth:
> Thanks for the contribution, the change looks good. I would add tests for `hasChanges` when the binding has been removed from the binder.
> 
> I also wonder if we could add the `hasChanges()` method also on the `Binding` interface, so that it would be possible to simply do a `binding.hasChanges()`. The implementation in `BindingImpl` could use the internal `binder` field to check it `this` binding is in the `changeBindings` set. And it will throw and `IllegalStateException` if the `binder` is null.
> 
> However, this is not mandatory for the PR to be merged

Let me look into this one. Definitely would be useful. If it is a quick change, let's include this in this PR.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -282,6 +282,18 @@ default BindingValidationStatus<TARGET> validate() {
          * @return A boolean value
          */
         boolean isConvertBackToPresentation();
+
+        /**
+         * Checks whether the related Binder instance has marked the field as
+         * having uncommitted changes.
```
benedikt-roth, line 288: @mcollovati please check if this description is a good one

mcollovati: I would not mention Binder here, as the fact we check the change status on Binder is an implementation detail.
Maybe something like "Checks whether the field the binding use has uncommitted changes"?

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -282,6 +282,18 @@ default BindingValidationStatus<TARGET> validate() {
...
+         *
+         * @throws IllegalStateException
+         *             if {@code binder} is null. Ensure that the binder has
+         *             been declared here.
```
mcollovati, line 292: I would not mention the Binder field here, as it is an implementation detail.
What about "if the binding is no longer attached to a Binder", as in the exception thrown by `validate(...)`?
```suggestion
         * @throws IllegalStateException
         *             if the binding is no longer attached to a Binder.
```

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -282,6 +282,18 @@ default BindingValidationStatus<TARGET> validate() {
...
+         *
+         * @return True if field has uncommitted changes
```
mcollovati, line 294:
```suggestion
         * @return {@literal true} if the field the binding uses has uncommitted changes, otherwise {@literal false}.
```

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -282,6 +282,18 @@ default BindingValidationStatus<TARGET> validate() {
...
+         */
+        boolean hasChanges() throws IllegalStateException;
```
mcollovati, line 296: I would skip the exception declaration, as it is a runtime exception.
I think that mentioning the exception in the javadoc is enough.

```suggestion
        boolean hasChanges() throws IllegalStateException;
```

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -1589,6 +1601,15 @@ private <T> T execute(Supplier<T> supplier) {
                 }
             }
         }
+
+        @Override
+        public boolean hasChanges() throws IllegalStateException {
+            if (this.binder == null) {
+                throw new IllegalStateException();
```
mcollovati, line 1608: We need an informative message for this exception

```suggestion
                throw new IllegalStateException("This Binding is no longer attached to a Binder");
```

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
```diff
@@ -199,6 +199,54 @@ public void clearBean_setsHasChangesToFalse() {
...
+
+        nameField.setValue("James");
+        assertTrue("Binder did not have value changes", binding.hasChanges());
+
+        binder.readBean(null);
+
+        assertFalse("Binder has changes after clearing all fields",
+                binding.hasChanges());
+    }
+
```
mcollovati, line 249: Please add also a test that checks the exception is thrown if the binding is unbound.

mcollovati (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
benedikt-roth: Fair enough

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
benedikt-roth: done ✅

mcollovati (approved):
The change looks good to me. Let's see if the validation passes.
Good job @benedikt-roth

PR #17863, opened 2023-10-20
fix: RouteAlias hides Route value

### LINKED ISSUE #9969: @Route with url parameter template doesn't work with @RouteAlias containing empty path

Navigation to a page having the url parameters template route and empty path route alias doesn't work, when the method `UI::navigate(Class<? extends Component> navigationTarget, RouteParameters parameters)` is used, as shown in the following example:

```java
@PageTitle("Hello World")
@CssImport("./styles/views/helloworld/hello-world-view.css")
@Route(value = "hello/:message?", layout = MainView.class)
@RouteAlias(value = "", layout = MainView.class)
public class HelloWorldView extends HorizontalLayout implements BeforeEnterObserver {

    private TextField name;
    private Button sayHello;

    public HelloWorldView() {
        setId("hello-world-view");
        name = new TextField("Your name");
        sayHello = new Button("Say hello");
        add(name, sayHello);
        setVerticalComponentAlignment(Alignment.END, name, sayHello);
        sayHello.addClickListener(e -> name.getOptionalValue().ifPresent(
                // This doesn't work: although the page has a
                // 'hello/:message?' navigation target, it does navigate to
                // path '' anyway.
                message -> UI.getCurrent().navigate(HelloWorldView.class,
                        new RouteParameters("message", message))));
    }

    @Override
    public void beforeEnter(BeforeEnterEvent beforeEnterEvent) {
        beforeEnterEvent.getRouteParameters().get("message").ifPresent(
                message -> Notification.show(String.format(
                        "Your message = '%s'", message),
                        3000, Notification.Position.BOTTOM_START));
    }
}
```

However, it works properly as soon as `@RouteAlias` is being removed completely, or another non-empty value is used (`@RouteAlias(value = "foo")`).

The workaround is to use `UI.getCurrent().navigate("hello/" + message);` instead. 

See attached project to reproduce 
[empty-route-alias-navigation-issue.zip](https://github.com/vaadin/flow/files/5925249/empty-route-alias-navigation-issue.zip)

1. Start the attached project `mvn spring-boot:run`
2. Navigate to localhost:8080
3. Enter any message to the text input.
4. Click say hello button.
5. Check that the browser doesn't navigate to 'hello/message'
6. Comment `@RouteAlias(value = "", layout = MainView.class)` in the code
7. Wait until restart, or restart manually
8. Do 2) - 4) and see that the browser navigates properly and the pop up notification is shown.

### Expected behavior
Browser navigates to 'hello/your-message' and pop up is shown.
### Actual behavior
Browser navigates to '' and no pop up shown.
### Versions:
    - Vaadin / Flow version: 19.0.0.alpha5, 18.0.5
    - Java version: 8, 11
    - OS version: Mac OS
    - Browser version (if applicable): Chrome 88
    - IDE (if applicable): IntellijIDEA.

caalador:
Specifically check Route annotation route 
first for match when going through templates.

Fixes #9969

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/ConfiguredRoutes.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/ConfiguredRoutes.java
@@ -16,8 +16,10 @@
 package com.vaadin.flow.router.internal;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -27,6 +29,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import com.vaadin.flow.component.Component;
+import com.vaadin.flow.router.Route;
 import com.vaadin.flow.router.RouteParameterData;
 import com.vaadin.flow.router.RouteParameterFormatOption;
 import com.vaadin.flow.router.RouteParameters;
@@ -431,8 +434,7 @@ public class ConfiguredRoutes implements Serializable {
             return null;
         }
 
-        final Collection<String> templates = model.getRoutes().keySet();
-        for (String template : templates) {
+        for (String template : getOrderedTemplates(navigationTarget, model)) {
             final T result = templateOutput.apply(template);
             if (result != null) {
                 return result;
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/internal/ConfiguredRoutes.java
```diff
@@ -442,4 +444,37 @@ private <T> T iterateTemplates(Class<? extends Component> navigationTarget,
...
+     *
+     * @param navigationTarget
+     *            target class
+     * @param model
+     *            {@link RouteModel} for navigation target
+     * @return target templates with route value before RouteAlias values
+     */
+    private static Collection<String> getOrderedTemplates(
+            Class<? extends Component> navigationTarget, RouteModel model) {
+        final Collection<String> templates = model.getRoutes().keySet();
```
tltv, line 459: It's a bit weird how `model.getRoutes()` is not already returning routes in correct order. Looks like it works as expected only with the the root (/).

Could `RouteSegment#getRoutes()` take an optional path parameter to check it in there already? https://github.com/vaadin/flow/blob/edb2bada51a733e3f101ffbec616caeea9d45f7f/flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java#L233 

So route from the `Route` annotation could be checked like this there:
`if (entry.getValue().isMainRouteSegment || (route != null && Objects.equals(route, entry.getKey()))) {`

This is optional however for considering.

tltv (approved):

flow-server/src/main/java/com/vaadin/flow/router/internal/ConfiguredRoutes.java
caalador: I would have expected that, but the mainRouteSegment doesn't seem to be set that way and I couldn't understand where and how it tried to select that. 
I feel that we should put at some point actual time into refactoring and rebuilding the templating for routing to something better/easier to handle.

Issue #3662, opened 2018-03-07
Div constructor should accept content


marcushellberg:
When I write  `new Div()`, I want to be able to pass in text content the same way I can when I call `new Span()`.

pleku:
Adding this had been debated for quite some time and many occasions, since some (am I the only one?) have argued that when doing layouting, using `new Div("my-class-name")` would be needed much more instead of `new Div("my text content").

You can do `new Div(new Text("such verbose API"))` for now as a workaround.

I think the HTML components are the first thing I would want to apply a fluent API to. Like `new Div().text("Foobar").className("text-block")`.

marcushellberg:
Yes. Ideally, I would like a fluent API. Especially when creating layouts, it's common to want to create divs with class names. 

My initial ticket was because I felt that it was inconsistent that a Span took in text content in its constructor, but a Div didn't. Both as similar as neither has any specific semantic meaning in HTML. One is just a block and the other is inline.

Legioth:
What about two different static factory methods: `Div.withText(String)` and `Div.withClass(String ...)`? If we're really eager, we could even do `Div.withHtml(String)` as well as icing on the cake.

Peppe:
I ran into an related issue. I had a bunch of HTML snippets in the backend that I wanted to show on in my UI. I wrote `add(new Html(message.getHtml());` where message is the bean from the backend.

I got the problems because some of the backend stuff was just text without root nodes like `lorem <i>ipsum</i>`, some had multiple nodes like `<div>lorem</div><div>ipsum</div>. Doing that caused my app to explode with exceptions on my screen, sayin that I either have 0 or multiple root nodes. I ended up changing my code to `add(new Html("<div>" + message.getHtml() + "</div>"));` to be able to get it working.

I know Html() is another element but it is related. I would hope it to automatically wrap my content to a div if I have zero or multiple roots with an exception in console, pointing me to another method where I can explicitely wrap it and get rid of the warning.

As with the original post, what I would have most wanted to do would have been write  `add(new Div(message.getHtml());` without me having to worry about this. And that's the same constructor @marcushellberg is talking about. Except it would render html :)

pleku:
Shortlist of strings I want to stuff into my divs in the constructor:
- plain text as block content (like the old label component)
- plain text as inline content (could wrap to newline if necessary)
- any safe HTML (like the old label component with html content)
- any unsafe HTML (because I'm bad-ass)
[...]

PR #17866, opened 2023-10-20
feat: add Div constructor that accept text content

tltv:
Fixes: #3662

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Div.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Div.java
@@ -55,7 +55,6 @@ public class Div extends HtmlContainer
      *            the text
      */
     public Div(String text) {
-        super();
         setText(text);
     }
 }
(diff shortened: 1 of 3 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/Div.java
```diff
@@ -47,4 +47,15 @@ public Div() {
...
     }
+
+    /**
+     * Creates a new div with the given text.
+     *
+     * @param text
+     *            the text
+     */
+    public Div(String text) {
+        super();
```
mcollovati, line 58: nit: call to super with no args is implicit, so I think we can remove it
```suggestion
```

mcollovati (commented):
Just a minor change, but non-blocking.

I don't know if it is worth adding a simple unit test in `DivTest` to ensure that the text is correctly set.

mcollovati (approved):

PR #17867, opened 2023-10-20
chore: fix typo when appending new line

manolo:
This fixes the following bad formatted message (notice %n after globally):

```
======================================================================================================
Your installed 'npm' version (9.2.0) is known to have problems.
Please update to a new one either:
  - by following the https://nodejs.org/en/download/ guide to install it globally%n  - or by updating your global npm installation with `npm install -g npm@latest`
  - or by running the frontend-maven-plugin goal to install it in this project:
  $ mvn com.github.eirslett:frontend-maven-plugin:1.10.0:install-node-and-npm -DnodeVersion="v20.8.0"

You can disable the version check using -Dvaadin.ignoreVersionChecks=true
======================================================================================================
```

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -1107,7 +1107,8 @@ public class FrontendTools {
             String... extraUpdateInstructions) {
         StringBuilder extraInstructions = new StringBuilder();
         for (String instruction : extraUpdateInstructions) {
-            extraInstructions.append("\n  - or ").append(instruction);
+            extraInstructions.append(System.lineSeparator()).append("  - or ")
+                    .append(instruction);
         }
         return String.format(BAD_VERSION, tool, version,
                 extraInstructions.toString(),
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -1107,7 +1107,7 @@ private String buildBadVersionString(String tool, String version,
             String... extraUpdateInstructions) {
         StringBuilder extraInstructions = new StringBuilder();
         for (String instruction : extraUpdateInstructions) {
-            extraInstructions.append("%n  - or ").append(instruction);
+            extraInstructions.append("\n  - or ").append(instruction);
```
mcollovati, line 1110: What about using `System.lineSeparator` instead of hard-coded `\n`, so it should work as `%n` in `String.format()`?

```suggestion
            extraInstructions.append(System.lineSeparator()).append("  - or ")
                    .append(instruction);
```

mcollovati (changes requested):

mcollovati (approved):

PR #17877, opened 2023-10-23
chore: Upgrade Node.js to 21.0.0

mshabarov:


knoobie:
I'm against this - NodeJS 21 is not the LTS once 24.3 is released 🙂 https://nodejs.dev/en/about/releases/

mshabarov:
> https://nodejs.dev/en/about/releases/

That's a good point. V24.3 will be maintained up to June 2024 (24.4 release in March + 3 months) and Node.js 21 will be maintained until June 2024. But Node.js 22 will come just around V24.5 release, which is inconvenient.
This means that 24.5 and 24.4 releases would stay on unmaintained version for a while until EoL.

I agree that better to stay on Node.js 20 in this case.

mshabarov:
Ok, thanks for reference. Let us discuss it within the team, but so far it seems to me a better option to keep using 20 by default. Anyway, this is configurable and one can of course use globally installed version, which is newer, if desired.

PR #17879, opened 2023-10-23
feat: Add Component/Element to ErrorEvent

### LINKED ISSUE #13044: `ErrorHandler` has no access to the origin component of the exception
### Describe your motivation

In the past (v8) it was possible to access the component that throw the exception inside the `ErrorHandler` - this was really useful to deactivate e.g. buttons that throw exceptions for some backend systems that were unavailable. This information isn't available anymore.

`ErrorEvent::getComponent/Element` which returns a `Optional<Component/Element` for the origin of the event, e.g. button or other components.

### Additional context

V8 custom error handler example:

```java

    this.setErrorHandler(event -> {
      Throwable t = event.getThrowable();

     // ....
     
      AbstractComponent abstractComponent = DefaultErrorHandler.findAbstractComponent(event);
      if (abstractComponent != null) {
        if (abstractComponent instanceof NativeButton) {
          // disable the error component where the error occurred
          abstractComponent.setEnabled(false);
        }
      }
     // ....
    });

```

caalador:
Now it is possible to query
for the element/component that
threw the exception leading
to the ErrorEvent.

Fixes #13044

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/StateTree.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/StateTree.java
@@ -400,8 +400,8 @@ public class StateTree implements NodeOwner {
                                     .equals(DefaultErrorHandler.class)) {
                                 throw e;
                             }
-                            getUI().getSession().getErrorHandler()
-                                    .error(new ErrorEvent(e));
+                            getUI().getSession().getErrorHandler().error(
+                                    new ErrorEvent(e, entry.getStateNode()));
                         }
                     });
         }
--- a/flow-server/src/main/java/com/vaadin/flow/server/ErrorEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/ErrorEvent.java
@@ -74,14 +74,7 @@ public class ErrorEvent implements Serializable {
      * @return Component that error happened for if available
      */
     public Optional<Component> getComponent() {
-        if (componentNode != null) {
-            try {
-                return Element.get(componentNode).getComponent();
-            } catch (IllegalArgumentException iae) {
-                // NO-OP return Optional.empty
-            }
-        }
-        return Optional.empty();
+        return getElement().flatMap(Element::getComponent);
     }
 
     /**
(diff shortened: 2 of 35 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/StateTree.java
```diff
@@ -400,8 +400,8 @@ public void runExecutionsBeforeClientResponse() {
                                     .equals(DefaultErrorHandler.class)) {
                                 throw e;
                             }
-                            getUI().getSession().getErrorHandler()
-                                    .error(new ErrorEvent(e));
+                            getUI().getSession().getErrorHandler().error(
+                                    new ErrorEvent(e, entry.getStateNode()));
```
Artur-, line 404: Instead of "leaking" StateNodes outside the state tree, would it be better to look up a potential element or component here?

caalador: I was trying to keep StateNode from knowing Element and Component. We could ofc do the search directly in the constructor instead of storing the node to the ErrorEvent

flow-server/src/main/java/com/vaadin/flow/server/ErrorEvent.java
```diff
@@ -46,6 +67,40 @@ public Throwable getThrowable() {
         return throwable;
     }
 
+    /**
+     * Get the Component that the error was thrown for. If not known returns
+     * empty optional.
+     *
+     * @return Component that error happened for if available
+     */
+    public Optional<Component> getComponent() {
```
mshabarov, line 76: Can be simplified as
```java
    public Optional<Component> getComponent() {
        return getElement().flatMap(Element::getComponent);
    }
```
giving less code and no need to test `getElement()` separately.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/ErrorEvent.java
caalador: done

mshabarov (approved):

Issue #17729, opened 2023-09-28
Cover more properties with the typed API in Style


mstahv:
### Describe your motivation

The Style API nowadays contains some more specific API for setting CSS than just providing string-string pairs. For example you can call `tf.getStyle().setZIndex(10);` indated of `tf.getStyle().set("z-index","10");`

The API is not complete though (and probably can't be), but was made by gut feeling to cover the most essential properties. I wrote the initial list but have already missed some properties.

### Describe the solution you'd like

We should extend the Style API whenever we find relevant missing API. Please drop in new suggestions below and let's improve it for next minor!

### Describe alternatives you've considered

Drink a six-pack of beer and carry on.

mstahv:
I was missing these recently:

        //tf.getStyle().setFontWeight("bold");
        tf.getStyle().set("font-weight", "bold");
        //tf.getStyle().setFontSize("40px");
        tf.getStyle().set("font-size", "40px");

mstahv:
Collecting some needs from our (bit ugly) [Notification examples](https://vaadin.com/docs/latest/components/notification): 

setMargin*Left/Right/Top/Bottom*
setFontWeight/Size (already cover in initial issue  though)

mstahv:
Other picks from docs:

* backgroundColor
* margin-inline-start
* margin-inline-end
* flex-wrap
* min-width
* max-width
* line-height
* max-height
* align-items

tltv:
- padding-left/-right/-top/-bottom
- border-left/-right/-top/-bottom

tltv:
Often needed flexbox properties:
- align-self
- flex-grow
- flex-shrink
- justify-content
- flex-direction
- flex-basis

PR #17882, opened 2023-10-23
feat: add more properties with the typed API in Style

tltv:
Fixes: #17729

### DIFF
--- a/flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/views/PublicView.java
+++ b/flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/views/PublicView.java
@@ -26,7 +26,7 @@ public class PublicView extends FlexLayout {
         header.getStyle().set("text-align", "center");
         add(header);
         Image image = new Image("public/images/bank.jpg", "Bank");
-        image.getStyle().setMaxWidth("100%").setMinWidth("0");
+        image.getStyle().setMaxWidth("100%").setMinHeight("0");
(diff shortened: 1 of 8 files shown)


tltv:
Set to draft for now to think about if it makes sense 
- to use `ElementConstants` in the `Style`
- to add getters (there's none other than `get(String)` at the moment)

tltv:
> - to add getters (there's none other than `get(String)` at the moment)

Dropping this idea after discussing with team. Style is only server side state that writes one way to the client. Adding all property getters could give an wrong impression of two-way synchronization.

mstahv:
I pushed in some "dogfooding" of the new API + one overload for setFontWeight.

I don't like the usage of separate constants, makes the code less readable and I don't believe it will help in maintenance either. But this really doesn't affect the end user (nor JVM), so reason to overthink this. The PR looks perfect to me 💪

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/views/PublicView.java
```diff
@@ -26,7 +26,7 @@ public PublicView() {
         header.getStyle().set("text-align", "center");
         add(header);
         Image image = new Image("public/images/bank.jpg", "Bank");
-        image.getStyle().set("max-width", "100%").set("min-height", "0");
+        image.getStyle().setMaxWidth("100%").setMinWidth("0");
```
tltv, line 29: Did you change `min-height` to `min-width` in purpose here? Does not seem to affect test results though.

tepi (approved):

PR #17886, opened 2023-10-24
chore!: Upgrade to Vite 5

Artur-:
Vite 5 is based on Rollup v4. The major breaking change there is dropping support for Node < 18 which does not affect Flow. There might be other potential plugin issues though.

Vite 5 also drops support for accidentally importing styles twice, i.e.
```ts
import styles from 'hello.css'
```

To import styles and apply globally, you should do

```ts
import 'hello.css'
```

and to import styles as a string you need to do
```ts
import styles from 'hello.css?inline';
```


mshabarov:
Waits for Vite 5 GA version release, which is expected in the beginning of November 2023.

mshabarov (approved):

PR #17889, opened 2023-10-24
fix: Show Hilla version when Hilla is available

Artur-:
Fixes https://github.com/vaadin/hilla/issues/1581

### DIFF
--- a/vaadin-dev-server/src/test/java/com/vaadin/base/devserver/ServerInfoTest.java
+++ b/vaadin-dev-server/src/test/java/com/vaadin/base/devserver/ServerInfoTest.java
@@ -15,15 +15,6 @@
  */
 package com.vaadin.base.devserver;
 
-import com.vaadin.flow.internal.hilla.EndpointRequestUtil;
-import org.junit.After;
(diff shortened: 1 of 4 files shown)


vaadin-dev-server/src/test/java/com/vaadin/base/devserver/ServerInfoTest.java
mcollovati, line 1: Why are the tests deleted? Do they no longer make sense?

Artur-: The old code sets a variable called "productName" (which is Vaadin or Hilla or "Vaadin, Hilla") and shows it using "vaadinVersion" as the version. This does not make sense as the Hilla version and the Vaadin version is not the same so the product name and related tests are removed. 

Seems like the "dash" tests could be retained though

Artur-: Added them back

mcollovati (approved):

PR #17890, opened 2023-10-24
Test bun

Artur-:
Newest bun has override support


caalador:
Seems bun doesn't get installed for the tests:
```
Vaadin is configured to use a globally installed bun, but bun was not found on your system.
```
Is on both failing it-test modules.

Artur-:
See #17913 for real test results

caalador (commented):
Looks fine, but needs passing validation.

Artur-:
This PR will not pass validation unless I make a separate PR first for unstalling bun in GHA

Artur-:
All green

flow-tests/test-frontend/pom.xml
```diff
@@ -50,6 +50,8 @@
                 <module>test-npm</module>
                 <module>test-pnpm/pom-production.xml</module>
                 <module>test-pnpm</module>
+                <module>test-bun/pom-production.xml</module>
+                <module>test-bun</module>
```
caalador, line 54: The only thing here is that this would now force dev machines to install bun to run tests as is. Also does this work on teamcity as I don't think the agents have bun installed? That would break the snapshot build (when it forces itself to run).
Should we put then bun module behind a profile and enable it by default for github actions?

Artur-: The TeamCity build can be fixed by adding a build step that installs bun. For devs who want to run all tests locally - they would have to install bun themselves. I can put it in a profile if you think it is worth it

caalador: If we get bun on the agents before merging then it s fine without I think.

caalador (approved):

Issue #17874, opened 2023-10-21
Building in CI behind proxy not working with Vaadin 24.2.0


felix-office:
### Description of the bug

Our Gitlab Runners have a proxy configuration with HTTP_PROXY and HTTPS_PROXY to work.
We've tried to update an application to vaadin 24.2.0 and the CI build fails. It seems like Vaadin 24.2.0 does not respect this env variables for npm anymore. Downgrading to vaadin 24.1.11 works as expected. We cleared all runner caches, so it seems to be a regression. Maybe there is also a change as 24.2.0 now installs node 20 by default?

Logs:
```
[INFO] using '/builds/oHy2LPvUM/0/REDACTED/REDACTED/REDACTED/.vaadin/node/node /builds/oHy2LPvUM/0/REDACTED/REDACTED/REDACTED/.vaadin/node/lib/node_modules/npm/bin/npm-cli.js --no-update-notifier --no-audit --scripts-prepend-node-path=true --ignore-scripts install' for frontend package installation
[ERROR] Command `/builds/oHy2LPvUM/0/REDACTED/REDACTED/REDACTED/.vaadin/node/node /builds/oHy2LPvUM/0/REDACTED/REDACTED/REDACTED/.vaadin/node/lib/node_modules/npm/bin/npm-cli.js --no-update-notifier --no-audit --scripts-prepend-node-path=true --ignore-scripts install` failed:
(node:148) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [TLSSocket]. Use emitter.setMaxListeners() to increase limit
npm ERR! code ETIMEDOUT
npm ERR! errno ETIMEDOUT
npm ERR! network request to https://registry.npmjs.org/yargs-parser failed, reason: 
npm ERR! network This is a problem related to network connectivity.
npm ERR! network In most cases you are behind a proxy or have bad network settings.
npm ERR! network 
npm ERR! network If you are behind a proxy, please make sure that the
npm ERR! network 'proxy' config is set properly.  See: 'npm help config'
```

For us this is a complete blocker to use vaadin 24.2.0.

### Expected behavior

The vaadin flow 24.2.0 application should build just fine.

### Minimal reproducible example

I can't provide a full example as these disclose your internal proxy configurations (which I dont even have access to).

Our Gitlab CI configures following variables globally, so they are being used in every job:
```yaml
variables:
  HTTPS_PROXY: $DATAPORT_PROXY
[...]

PR #17898, opened 2023-10-24
chore: Upgrade default npm to 10.2.1

tltv:
Fixes: #17874

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
@@ -70,7 +70,7 @@ public class FrontendTools {
     /**
      * This is the version shipped with the default Node version.
      */
-    public static final String DEFAULT_NPM_VERSION = "10.1.0";
+    public static final String DEFAULT_NPM_VERSION = "10.2.1";
 
     public static final String DEFAULT_PNPM_VERSION = "8.6.11";
 


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -70,7 +70,7 @@ public class FrontendTools {
     /**
      * This is the version shipped with the default Node version.
      */
-    public static final String DEFAULT_NPM_VERSION = "10.1.0";
+    public static final String DEFAULT_NPM_VERSION = "10.2.1";
```
caalador, line 73: This change will always fail the test `installNode_NodeIsInstalledToTargetDirectory` as the npm version supplied by node 20.8.1 (and 20.9.0) is 10.1.0
Event the current node version 21.1.0 only brings in npm 10.2.0
I don't think we have a feature for updating the npm installation after node installation.

tltv:
Need to wait that this https://github.com/npm/cli/issues/6835 gets to the next LTS node 20 release, and both node and npm default versions has to be then updated at the same time in `FrontendTools` [based on this release table](https://nodejs.org/en/download/releases).

Other options would be to update to latest node 21 but this was already decided otherwise here https://github.com/vaadin/flow/pull/17877.

mshabarov:
Waits for https://github.com/nodejs/node/issues/50376

Artur-:
This was upgraded in #18111

PR #17901, opened 2023-10-24
fix: exit disconnect if disconnection already happened

### LINKED ISSUE #17855: AtmospherePushConnectionTest.disconnect_concurrentRequests_preventDeadlocks is flaky
The `disconnect_concurrentRequests_preventDeadlocks` test on `AtmospherePushConnectionTest` fails randomly, probably because of sleeps introduced to simulate the deadlock case.
It should be fixed by adjusting the timings or finding a better way to simulate the deadlock.

```
Error:    AtmospherePushConnectionTest.disconnect_concurrentRequests_preventDeadlocks:227 Disconnection on spawned thread failed: java.lang.AssertionError
	at com.vaadin.flow.server.communication.AtmospherePushConnection.disconnect(AtmospherePushConnection.java:334)
	at com.vaadin.flow.server.communication.AtmospherePushConnectionTest.lambda$disconnect_concurrentRequests_preventDeadlocks$11(AtmospherePushConnectionTest.java:204)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

mcollovati:
Fixes #17855

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
@@ -329,18 +329,15 @@ public class AtmospherePushConnection implements PushConnection {
         }
 
         synchronized (lock) {
+            if (!isConnected() || resource == null) {
+                // Already disconnected. Should not happen but if it does,
+                // we don't want to cause NPEs
+                getLogger().debug(
+                        "Disconnection already happened, ignoring request");
+                return;
+            }
             try {
                 disconnecting = true;
-                assert isConnected();
-                if (resource == null) {
-                    // Already disconnected. Should not happen but if it does,
-                    // we
-                    // don't
-                    // want to cause NPEs
-                    getLogger().debug(
-                            "AtmospherePushConnection.disconnect() called twice, this should not happen");
-                    return;
-                }
                 if (resource.isResumed()) {
                     // This can happen for long polling because of
                     // http://dev.vaadin.com/ticket/16919


czp13 (commented):
This seems good to me and is a nice improvement, but I do not feel expert enough to approve it :).

And just a question/idea, does it make sense to add a test here:
- [AtmospherePushConnectionTest class](https://github.com/vaadin/flow/blob/main/flow-server/src/test/java/com/vaadin/flow/server/communication/AtmospherePushConnectionTest.java)

Or because this involves threads, would it be quite hard to create a situation where the synchronized block has a 'not connected' state?

mcollovati:
Tests are failing because of issue with node download.

mcollovati:
I tried to add a test, but I wasn't able to create a situation that can reproduce the condition this change should fix.
Even If I start multiple threads with a dedicated executor service, I never get the situation where two threads pass the first if condition before trying to acquire the lock.
Any suggestions for a test are more than appreciated

flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
```diff
@@ -329,14 +329,16 @@ public void disconnect() {
...
+            }
             try {
                 disconnecting = true;
-                assert isConnected();
                 if (resource == null) {
                     // Already disconnected. Should not happen but if it does,
-                    // we
-                    // don't
-                    // want to cause NPEs
+                    // we don't want to cause NPEs
```
Artur-, line 341: Should this be combined with the `isConnected` check? Can we ever be connected but still not have a resource?

Artur-:
I don't think you can make a reliable test for this without a major effort, so not worth it. It did fail here https://github.com/vaadin/flow/actions/runs/6627628044/job/18006973434 which was the reason for me to look at the problem

flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
mcollovati: Do you mean remove the `resource == null` check and move the comment in the `!isConnected` block?

Artur-: Something like
```
if (!isConnected() || resource == null) {
```
unless `isConnected` already makes sure `resource` is not null

Artur-: Right now you are leaving the `disconnecting` flag to true if `resource` is null

Artur-:
Also here https://github.com/vaadin/flow/actions/runs/6636379910/job/18028856632

flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
mcollovati: `disconnecting` is set to false in the `finally` block. Shouldn't it be executed even if you have a return statement?

mcollovati: applied the suggeston

Artur-: Ah, right, my bad. Then that's not an issue. Still having multiple checks for if we are connected does not make the code easier to read in my opinion. Especially as the debug message is wrong because the "called twice" case is already  handled by  ` if (!isConnected())`.

Artur- (approved):

PR #17912, opened 2023-10-25
refactor: Delete duplicate version comparison method

Artur-:


czp13 (approved):
Nice! As we have the:
- `comperaTo` already implemented,
- plus knowing the `Frontendversion` as well, 

we can:
 - remove the old long method (isVersionAtLeast), 
- and the parameter passing.

mshabarov (approved):

PR #17920, opened 2023-10-26
fix: page calculation when not equally divisible

### LINKED ISSUE #17750: Grid pagination broken when using count
### Description of the bug

If a count is used in the grid.setItems call, such as the following code:
`        grid.setItems(query -> samplePersonService.list(
                PageRequest.of(query.getPage(), query.getPageSize(), VaadinSpringDataHelpers.toSpringDataSort(query)))
                .stream(), query -> samplePersonService.count());`

Then the pagination that Vaadin creates gets confused for the last page if the number of items to be fetched is not a multiple of the page size (50 by default). 

What ends up happening is that the last few items (depending on the item count and page size) are duplicates of the preceding items. 

The following example is based on what gets produced by start.vaadin.com with the helloworld page changed to a Master-Detail view. Additional lines are added to data.sql to make the record count not be a multiple of 50.

Example original code (count is commented out):
`        grid.setItems(query -> samplePersonService.list(
                PageRequest.of(query.getPage(), query.getPageSize(), VaadinSpringDataHelpers.toSpringDataSort(query)))
                .stream()/*, query -> samplePersonService.count()*/);`

Result:
image

Note that the lines at the bottom contain the additional people added to data.sql.

With the following code (count is enabled):
`        grid.setItems(query -> samplePersonService.list(
                PageRequest.of(query.getPage(), query.getPageSize(), VaadinSpringDataHelpers.toSpringDataSort(query)))
                .stream(), query -> samplePersonService.count());`

Result:
image

Note that the last few lines are duplicated instead of showing the correct data.

Please note that this seems to be dependent on the user scrolling using the mouse rather than going to the bottom of the grid at once using the scroll bars. 

### Expected behavior

The resulting grid data should not depend on the whether the code makes use of the record count or if the user scrolls with the mouse, or the scroll bars or goes directly to the end of the grid. The results should be the same.

Steps to duplicate:

- Go to start.vaadin.com and change the hello world page to use a Master-Detail view.
- Download the resulting zip file, open in the IDE, get it to run, etc.
- Add 13 additional lines to the data.sql file so the record count doesn't match a multiple of the page size.
- Change the code for grid.setItems in HelloWorldView.java to:
`        grid.setItems(query -> samplePersonService.list(
                PageRequest.of(query.getPage(), query.getPageSize(), VaadinSpringDataHelpers.toSpringDataSort(query)))
                .stream(), query -> samplePersonService.count());`
- Run the application and scroll with the mouse wheel towards the end of the grid to see the duplicated items.

Attached is a zip file with the steps above in case I have missed anything.
[my-app.zip](https://github.com/vaadin/flow/files/12773087/my-app.zip)

### Versions

- Vaadin / Flow version: 23 and 24
- Java version: 11 and 17
- OS version: Linux
- IDE (if applicable): N/A

caalador:
Fix the query page and pageSize for
cases where the offset is not
equally divisible with the limit/pageSize.

Fixes #17750

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
@@ -151,7 +151,7 @@ public class Query<T, F> implements Serializable {
      * <p>
      * This is an alias for {@link #getLimit()} if the page offset can be evenly
      * divided by the limit. Else the page size will be increased to evenly
-     * divide offset so the items skip for page will go to the corrent item.
+     * divide offset so the items skip for page will go to the correct item.
      *
      * @return the page size used for data access
      */
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
@@ -147,8 +147,11 @@ public class Query<T, F> implements Serializable {
      * Returns the page size that should be returned. The amount of items can be
      * smaller if there is no more items available in the backend.
      * <p>
-     * Vaadin asks data from the backend in paged manner. This is an alias for
-     * {@link #getLimit()}. As long as the size is evenly divisible.
+     * Vaadin asks data from the backend in paged manner.
+     * <p>
+     * This is an alias for {@link #getLimit()} if the page offset can be evenly
+     * divided by the limit. Else the page size will be increased to evenly
+     * divide offset so the items skip for page will go to the corrent item.
      *
      * @return the page size used for data access
      */
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
@@ -38,6 +38,7 @@ public class Query<T, F> implements Serializable {
     private final List<QuerySortOrder> sortOrders;
     private final Comparator<T> inMemorySorting;
     private final F filter;
+    private Integer pageSize;
 
     /**
      * Constructs a Query for all rows from 0 to {@link Integer#MAX_VALUE}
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
```diff
@@ -119,24 +120,42 @@ public int getLimit() {
...
+        this.pageSize = pageSize;
     }
 
     /**
      * Returns the page size that should be returned. The amount of items can be
      * smaller if there is no more items available in the backend.
      * <p>
      * Vaadin asks data from the backend in paged manner. This is an alias for
-     * {@link #getLimit()}.
+     * {@link #getLimit()}. As long as the size is evenly divisible.
```
mcollovati, line 151: Could the "As long ..." part be reformulated?
I understand the meaning because I looked at the `getPage()` source code, but only reading the statement, I couldn't get what the size should be divisible by.

mcollovati (commented):
Just a note on a Javadoc, other than that LGTM

flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
caalador: done?

flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
```diff
@@ -119,24 +120,45 @@ public int getLimit() {
...
      * Returns the page size that should be returned. The amount of items can be
      * smaller if there is no more items available in the backend.
      * <p>
-     * Vaadin asks data from the backend in paged manner. This is an alias for
-     * {@link #getLimit()}.
+     * Vaadin asks data from the backend in paged manner.
+     * <p>
+     * This is an alias for {@link #getLimit()} if the page offset can be evenly
+     * divided by the limit. Else the page size will be increased to evenly
+     * divide offset so the items skip for page will go to the corrent item.
```
mcollovati, line 154: Maybe a typo in `the corrent item`.
Should it be `correct` or `current`?

flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
mcollovati: Great. Thanks for clarification

flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
caalador: =_=
Correct item

mcollovati (approved):

Issue #17836, opened 2023-10-16
Extend DefaultI18NProvider for Spring


caalador:
For spring there should be a Bean defined for DefaultI18NProvider that would check for default translations using `ResourcePatternResolver` e.g. `resourcePatternResolver.getResources("classpath:**/vaadin-i18n/*.properties);`

Code would be something like:
```
@Bean
@ConditionalOnMissingBean(value = I18NProvider.class)
public DefaultI18NProvider vaadinI18nProvider() {
  Resource[] translations = resourcePatternResolver.getResources("classpath:**/vaadin-i18n/*.properties);
  if(translations.length > 0) {
    List<Locale> locales = I18NUtil.collectLocalesFromFiles(Arrays.stream(translations).map(Resource::getFile).collect(Collectors.toList()));
    return new DefaultI18NProvider(locales );
  }
  return null;
}
```

I18NUtil needs extraction of the collectLocalesFromFiles.

Continuation of #17823 and #5917

PR #17933, opened 2023-10-26
feat: add default I18N provider bean for Spring

tltv:
Extends DefaultI18NProvider (#5917) for Spring. `VaadinApplicationConfiguration` defines `DefaultI18NProvider` bean and initializes it with locales based on translation properties files found in the classpath under '/vaadin-i18n' path.

Fixes: #17836


tltv:
Just noticed that Logger instance field is not needed anymore so I'll remove that.

vaadin-spring/src/main/java/com/vaadin/flow/spring/i18n/DefaultI18NProviderFactory.java
```diff
@@ -0,0 +1,82 @@
...
+     * .properties files in /vaadin-i18n folder in the classpath including JAR
+     * contents.
+     *
+     * @return location pattern
+     * @see ResourcePatternResolver
+     * @see ResourcePatternResolver#getResources(String)
+     */
+    protected String getLocationPattern() {
+        return DEFAULT_LOCATION_PATTERN;
+    }
```
caalador, line 57: This is perhaps also a bit extra as it is not used anywhere outside the class and shouldn't be overridden for the default feature.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinApplicationConfiguration.java
```diff
@@ -58,4 +61,16 @@ public ApplicationConfigurationFactory defaultApplicationConfigurationFactory()
...
+    /**
+     * Creates default {@link I18NProvider}. This is created only if there's no
+     * {@link I18NProvider} bean declared.
+     *
+     * @return default I18N provider
+     */
+    @Bean
+    @ConditionalOnMissingBean(value = I18NProvider.class)
+    public DefaultI18NProvider vaadinI18nProvider() {
+        return new DefaultI18NProviderFactory().create();
```
caalador, line 74: Doesn't this now always create a provider? It should be created only if there exists translation files.

Even for a no locales provider `createAndRegisterSession` will do some extra work on setLocale that is not done without a provider.

vaadin-spring/src/main/java/com/vaadin/flow/spring/i18n/DefaultI18NProviderFactory.java
caalador: Perhaps have it as a parameter that can be set to application.properties instead as overriding the factory and using it seems a bit too complex for just updating the pattern.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinApplicationConfiguration.java
tltv: It's now always creating bean as null will cause exception (can't remember the type atm). Need to recheck if null is possible some way.

mcollovati (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinApplicationConfiguration.java
caalador: Is it possible to cancel creation using `@Conditional`?

mcollovati: Good point about having the `@Conditional` annotation!

vaadin-spring/src/main/java/com/vaadin/flow/spring/i18n/DefaultI18NProviderFactory.java
mcollovati: Couldn't the factory be a simple utility method instead?
So the custom bean definition could be something like

```
    @Bean
    public DefaultI18NProvider vaadinI18nProvider() {
        return new DefaultI18NProviderFactory.create(myPatter);
```

tltv: All that is needed to override it, is this:
```
    @Bean
    public I18NProvider overrideI18nProvider() {
        return new DefaultI18NProviderFactory() {
            @Override
            protected String getLocationPattern() {
                return "...";
            }
        }.create();
    }
```
New parameter would make it even easier to change for sure.

tltv: > Couldn't the factory be a simple utility method instead?
This could work. Its a one liner then to change it, so new property would not make it much easier.

mcollovati: From a Spring Boot user point of view, probably the property might make more sense.

mcollovati: So we could maybe have both the factory method, and have the property value injected in `vaadinI18nProvider` method.
It should be something like `vaadinI18nProvider(@Value("${vaadin.something:the default value}"}) pattern)` and use the static factory in the method body

tltv: done

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinApplicationConfiguration.java
caalador: I would expect it to be quite well optimized on Spring side.

caalador (dismissed):

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinApplicationConfiguration.java
```diff
@@ -58,4 +63,20 @@ public ApplicationConfigurationFactory defaultApplicationConfigurationFactory()
...
+     * @return default I18N provider
+     */
+    @Bean
+    @ConditionalOnMissingBean(value = I18NProvider.class)
+    @Conditional(DefaultI18NProviderFactory.class)
+    public DefaultI18NProvider vaadinI18nProvider(
+            @Value("${vaadin.i18n.location-pattern:"
+                    + DefaultI18NProviderFactory.DEFAULT_LOCATION_PATTERN
+                    + "}") String locationPattern) {
+        return new DefaultI18NProviderFactory().create(locationPattern);
```
mcollovati, line 80: nit: not a blocker, but I still wonder if we can make`create` a static method, to avoid a useless instantiation

mcollovati (approved):

mcollovati (approved):

PR #17936, opened 2023-10-26
fix: send markAsUsed only when Theme Editor is opened first time

### LINKED ISSUE #17934: ThemeEditor usage events fired at the wrong time
### Description of the bug

ThemeEditor usage events are apparently being fired at the wrong time, leading to an inflation of usage stats: https://vaadin.slack.com/archives/C05T2D8PBTP/p1698323623072909

### Expected behavior

Only actual usage of ThemeEditor should trigger a usage event.

N/A

### Versions

N/A

MarcinVaadin:
## Description

Fixed ThemeEditor tracking to send markAsUsed only when user first time opens Theme Editor tab.

Fixed issues related to calling activate and deactivate methods on dev-tools plugins.

Fixes #17934

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-dev-server/frontend/vaadin-dev-tools.ts
+++ b/vaadin-dev-server/frontend/vaadin-dev-tools.ts
@@ -74,7 +74,6 @@ interface Tab {
   title: string;
   render: (() => TemplateResult) | string;
   element?: HTMLElement;
-  activate?: () => void;
   handleMessage?: MessageHandler;
 }
 
@@ -1373,15 +1372,15 @@ export class VaadinDevTools extends LitElement {
                 @click=${() => {
                   const currentTab = this.tabs.find((tab) => tab.id === this.activeTab);
                   if (currentTab && currentTab.element) {
-                    const deactivateMethod = (currentTab.element as any)?.deactivate;
+                    const deactivateMethod = (currentTab.element?.firstElementChild as any)?.deactivate;
                     if (deactivateMethod) {
-                      deactivateMethod.call(currentTab.element);
+                      deactivateMethod.call(currentTab.element?.firstElementChild);
                     }
                   }
                   this.activeTab = tab.id;
-                  const activateMethod = (tab.element as any).activate;
+                  const activateMethod = (tab.element?.firstElementChild as any).activate;
                   if (activateMethod) {
-                    activateMethod.call(tab.element);
+                    activateMethod.call(tab.element?.firstElementChild);
                   }
                 }}
               >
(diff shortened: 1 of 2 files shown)


vaadin-dev-server/frontend/vaadin-dev-tools.ts
```diff
@@ -1373,15 +1372,15 @@ export class VaadinDevTools extends LitElement {
...
-                    const deactivateMethod = (currentTab.element as any)?.deactivate;
+                    const deactivateMethod = (currentTab.element?.firstElementChild as any)?.deactivate;
                     if (deactivateMethod) {
-                      deactivateMethod.call(currentTab.element);
+                      deactivateMethod.call(currentTab.element?.firstElementChild);
                     }
                   }
                   this.activeTab = tab.id;
-                  const activateMethod = (tab.element as any).activate;
+                  const activateMethod = (tab.element?.firstElementChild as any).activate;
```
MarcinVaadin, line 1381: tab.element is div element wrapping dev-tools plugin component so activateMethod was always undefined.

vaadin-dev-server/frontend/vaadin-dev-tools.ts
```diff
@@ -74,7 +74,6 @@ interface Tab {
   title: string;
   render: (() => TemplateResult) | string;
   element?: HTMLElement;
-  activate?: () => void;
```
MarcinVaadin, line 77: That was never used as activate and deactivate should be called directly on plugin component.

vaadin-dev-server/frontend/vaadin-dev-tools.ts
Artur-: That is only the case for tabs that has a render function. Tabs that are defined using a tag (e.g. plugins) has no extra wrapper

MarcinVaadin:
End user can only add plugin using string tag name, so direct setting TemplateResult to render function is possible only within dev tools itself. It's not possible to write usable IT test for this change.

tarekoraby:
I'm not sure where this fix is moving the event triggering. But just a random suggestion that I'm unsure is technically feasible (feel free to ignore): it might be better to trigger the usage event when the user has edited the theme of a component for the first time.

MarcinVaadin:
@tarekoraby Currently fix is marking „as used” when user opens a tab with theme editor. Will update to first pick action.

Artur- (approved):

PR #17940, opened 2023-10-27
chore: vulnerable dependency upgrade

ZheSun88:
compare this with #17943 

netty-*: 4.1.100.Final reports [CVE-2023-4586](https://nvd.nist.gov/vuln/detail/CVE-2023-4586), but there is no new version yet. 
bcprov-jdk15on : cannot upgrade, as this is the latest version, while reports [CVE-2023-33201](https://nvd.nist.gov/vuln/detail/CVE-2023-33201)

do `mvn org.owasp:dependency-check-maven:check` can check the report.

PR #17943, opened 2023-10-30
fix!: replace outdated webpush dependency

ZheSun88:
if flow build is green, let us take this method instead of #17940

### DIFF
--- a/flow-webpush/pom.xml
+++ b/flow-webpush/pom.xml
@@ -32,12 +32,6 @@
             <artifactId>webpush-jwt-jose4j</artifactId>
             <version>6.1.1</version>
         </dependency>
-        <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>httpcore</artifactId>
-            <version>4.4.16</version>
-        </dependency>
-
 
         <!-- DefaultApplicationConfigurationFactory is declared as an OSGi
             service -->
--- a/flow-webpush/pom.xml
+++ b/flow-webpush/pom.xml
@@ -28,23 +28,16 @@
         <!-- API DEPENDENCIES -->
 
         <dependency>
-            <groupId>nl.martijndwars</groupId>
-            <artifactId>web-push</artifactId>
-            <version>5.1.1</version>
-
-            <exclusions>
-                <exclusion>
-                    <groupId>com.typesafe.netty</groupId>
-                    <artifactId>netty-reactive-streams</artifactId>
-                </exclusion>
-            </exclusions>
+            <groupId>dev.blanke.webpush.jwt</groupId>
+            <artifactId>webpush-jwt-jose4j</artifactId>
+            <version>6.1.1</version>
         </dependency>
-
         <dependency>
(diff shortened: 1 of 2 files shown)


flow-webpush/src/main/java/com/vaadin/flow/server/webpush/WebPush.java
```diff
@@ -113,8 +115,10 @@ public WebPush(String publicKey, String privateKey, String subject) {
     public void sendNotification(Subscription subscription,
             WebPushMessage message) throws WebPushException {
         try {
-            HttpResponse response = pushService
-                    .send(new Notification(subscription, message.toJson()));
+            ;
```
mshabarov, line 118:
```suggestion
```

flow-webpush/pom.xml
```diff
@@ -28,24 +28,17 @@
...
+            <artifactId>webpush-jwt-jose4j</artifactId>
+            <version>6.1.1</version>
         </dependency>
-
         <dependency>
-            <groupId>org.bouncycastle</groupId>
-            <artifactId>bcprov-jdk15on</artifactId>
-            <version>1.70</version>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpcore</artifactId>
```
mshabarov, line 37: This apparently causes the following error when the push notification is being sent.
I got this error with CRM tutorial example:

```
java.lang.ClassCastException: class jdk.internal.net.http.HttpResponseImpl cannot be cast to class org.apache.http.HttpResponse (jdk.internal.net.http.HttpResponseImpl is in module java.net.http of loader 'platform'; org.apache.http.HttpResponse is in unnamed module of loader 'app')
        at com.vaadin.flow.server.webpush.WebPush.sendNotification(WebPush.java:120) ~[flow-webpush-24.3-SNAPSHOT.jar:na]
        at com.example.application.views.MessageView.lambda$getMessageBuilder$0(MessageView.java:57) ~[classes/:na]
        at java.base/java.lang.Thread.run(Thread.java:833) ~[na:na]

```

ZheSun88: probably the implementation here needs to be updated. https://github.com/vaadin/flow/blob/7cc38fa79dace34aa47dd86bc1ff5e83eb976d47/flow-webpush/src/main/java/com/vaadin/flow/server/webpush/WebPush.java#L118

there is a cast...

mshabarov: Let me try to replace it to `jdk.internal.net.http.HttpResponse`, it needs a couple of tweaks in code.

ZheSun88: Thanks

mshabarov (approved):
Tested with https://github.com/vaadin/flow-crm-tutorial/tree/feature/webpush , works as expected. Many thanks to @ZheSun88 and @manolo 🥇

mshabarov:
This fork has changes in `Subscription` API: things like `subscription.endpoint` -> `subscription.endpoint()`.
Easy to fix however, so I propose to pick this change into `24.2` series.

PR #17952, opened 2023-10-31
fix: fix web push compilation error

mcollovati:


mshabarov (approved):

mshabarov:
Good catch, @mcollovati !

PR #17980, opened 2023-11-02
test: run production mode first for npm and pnpm test modules

mcollovati:
As in Flow 23.3


mcollovati:
For some reason, the `Flow tests in npm and production mode` module is executed by two different jobs, and it is failing on both with NPM errors.

caalador (approved):

PR #17984, opened 2023-11-03
feat: add navigation access control feature

mcollovati:
This change introduces a new component to handle access control during view navigation. The NavigationAccessControl is a transparent replacement for ViewAccessChecker, but it can be configure to perform additional custom checks, such as the out-of-the-box RoutePathAccessChecker, that verifies if a user is allowed to visit given path.

In Spring Boot projects, RoutePathAccessChecker integrates with Spring URL-based HTTP security, allowing to define protect rules for routes using ant matchers through the VaadinWebSecurity.routeAwareAntMatcher method. It can be enabled by providing a NavigationAccessControlConfigurer bean, customized as needed.

RoutePathAccessChecker integration with other security frameworks can be obtained by implementing custom AccessPathChecker.


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -142,13 +197,30 @@ public boolean isAnonymousRoute(HttpServletRequest request) {
...
-        boolean result = accessAnnotationChecker.hasAccess(targetView, null,
-                role -> false);
-        if (result) {
-            getLogger().debug(path + " refers to a public view");
+        BeforeEnterEvent event = new BeforeEnterEvent(
+                new NavigationEvent(router,
+                        new Location(path,
+                                QueryParameters
+                                        .full(request.getParameterMap())),
+                        new UI(), NavigationTrigger.CLIENT_SIDE),
```
knoobie, line 205: Is this `new UI()` really necessary? I feel like this could create an unnessary load on the server, especially when the call to instantiate new UIs is monitored / traced / oberserved

mcollovati: Good point. Actually, the `NavigationEvent` expects a UI instance because it would not make sense to have such an event without UI.
I'm not sure if we should relax this constraint.
An option could be to create a single stub UI instance in RequestUtil

mshabarov: `UI` might exist already or might not, depending on is it a first opened page in the browser tab or is it a navigation to another view from already opened page. I agree that better to avoid to create UI instance. Can we just use `UI::getCurrent` for this purpose?

mshabarov: Or, well, we cannot, as it might be `null` and event expects that it's not `null`.

mcollovati: We cannot use `UI::getCurrent` because when we evaluate if the request is for a public route, we are not in a Vaadin request, but in the Spring Security filter chain, so the returned UI will always be null.

knoobie: Looking at `NavigationContext`, it seems there is also another constructor which doesn't need a NavigationEvent.. could this help? Haven't digged deep enough.. there are is a lot of code :D

mcollovati: Exactly, a fix is on his way :smile:

mcollovati: done

mshabarov (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -217,8 +213,7 @@ protected void configure(HttpSecurity http) throws Exception {
             urlRegistry.anyRequest().authenticated();
         });
 
-        // Enable view access control
-        viewAccessChecker.enable();
+        accessControl.setEnabled(true);
```
knoobie, line 216: Should this be change to `accessControl.setEnabled(isNavigationAccessControlEnabled())` where this method is protected and returns `true` by default, allowing people to overwrite this easily? I remember multiple discussion in the past, that the default call to viewAccessChecker.enable() was cumbersome for people

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/SpringNavigationAccessControl.java
```diff
@@ -0,0 +1,62 @@
...
+import com.vaadin.flow.spring.AuthenticationUtil;
+
+/**
+ * A Spring specific navigation access control that falls back to Spring
+ * mechanisms for user retrieval and role checking, when the generic mechanisms
+ * do not work.
+ */
+public class SpringNavigationAccessControl extends NavigationAccessControl {
+
+    public SpringNavigationAccessControl() {
```
knoobie, line 22: Is this empty constructor needed?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/SpringAccessPathChecker.java
```diff
@@ -0,0 +1,73 @@
...
+     * It applies the given Vaadin servlet url mapping to the input path before
+     * delegating the check to the evaluator.
+     *
+     * @param urlMapping
+     *            Vaadin servlet url mapping
+     * @param evaluator
+     *            evaluator to check path permissions.
+     */
+    public SpringAccessPathChecker(String urlMapping,
+            WebInvocationPrivilegeEvaluator evaluator) {
```
knoobie, line 61: nit: I would personally swap the parameter in the constructor and change the first constructor to call (this(evaluator, null)

vaadin-spring/src/main/java/com/vaadin/flow/spring/AuthenticationUtil.java
```diff
@@ -1,6 +1,7 @@
 package com.vaadin.flow.spring;
 
 import java.util.function.Function;
+import java.util.function.Predicate;
```
knoobie, line 4: nit: since this class has no other changes.. this import looks wrong? :)

flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessChecker.java
```diff
@@ -0,0 +1,468 @@
...
+            return new Result(Decision.REJECT, reason);
+        }
+
+    }
+
+    /**
+     * A representation of the access check result, potentially providing deny
+     * reason.
+     */
+    final class Result implements Serializable {
```
knoobie, line 396: nit: I feel like this is such a generic name.. NavigationAccessResult? Even tho it's longer, the meaning is more clear

flow-server/src/main/java/com/vaadin/flow/server/auth/DefaultNavigationCheckDecisionResolver.java
```diff
@@ -0,0 +1,118 @@
...
+ * </pre>
+ */
+public class DefaultNavigationCheckDecisionResolver
+        implements NavigationAccessChecker.DecisionResolver {
+
+    public static final Logger LOGGER = LoggerFactory
+            .getLogger(DefaultNavigationCheckDecisionResolver.class);
+
+    @Override
+    public NavigationAccessChecker.Result resolve(
```
knoobie, line 53: Open for discussion: This is such a hot path for the server; is the extensive usage of streams nesessary? If I remember correctly streams performance isn't the best for small lists.

flow-server/src/main/java/com/vaadin/flow/server/auth/DefaultNavigationCheckDecisionResolver.java
```diff
@@ -0,0 +1,118 @@
...
+        } else {
+            // Mixed consensus
+            String summary = resultsByDecision.entrySet().stream()
+                    .map(e -> e.getKey() + " = " + e.getValue().size())
+                    .collect(Collectors.joining(", ", "Votes: ", ""));
+            if (neutralVotes > 0) {
+                summary += ", " + NavigationAccessChecker.Decision.NEUTRAL
+                        + " = " + neutralVotes;
+            }
+            LOGGER.debug("Access to view '{}' with path '{}' blocked because "
```
knoobie, line 106: should this be info? This information seem kinda important

flow-server/src/main/java/com/vaadin/flow/server/auth/AnnotatedViewAccessChecker.java
```diff
@@ -0,0 +1,89 @@
...
+    private static final Logger LOGGER = LoggerFactory
+            .getLogger(AnnotatedViewAccessChecker.class);
+
+    private final AccessAnnotationChecker accessAnnotationChecker;
+
+    /**
+     * Creates an instance using the given checker.
+     */
+    public AnnotatedViewAccessChecker() {
+        this.accessAnnotationChecker = new AccessAnnotationChecker();
```
knoobie, line 46: nit: call the other constructor with this(new AccessAnnotationChecker())

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/SpringNavigationAccessControl.java
mcollovati: It is there to simplify default usage in Spring project that does not use `VaadinWebSecurity`.
By using `new SpringNavigationAccessControl()` you get a Spring aware `NavigationAccessControl` with view access checker and default decision resolver, that's basically the current `ViewAccessChecker`

mcollovati: BTW, it misses Javadoc

knoobie: Thanks :) Finding the usage of this constructor was kinda hard with the GitHub UI. That explains it!

vaadin-spring/src/main/java/com/vaadin/flow/spring/AuthenticationUtil.java
mcollovati: It should be removed

flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessChecker.java
mcollovati: What about `AccessCheckResult`?

knoobie: I'm open for any name more meaningful :)

flow-server/src/main/java/com/vaadin/flow/server/auth/DefaultNavigationCheckDecisionResolver.java
knoobie: I'm also fine with that, I wanted to mention it cause this combination "blocked because of 'wrong' configuration" sounded like something to be fixed

mcollovati: Good point.
Sorry, I missed that this is from a "rejection" situation.
Effectively, this will then result in a `NavigationContext.reject("....")` that in development mode will throw an exception.
So, you're right.
The log could level should maybe be WARN

flow-server/src/main/java/com/vaadin/flow/server/auth/DefaultNavigationCheckDecisionResolver.java
mcollovati: That's indeed an interesting point.
I propose to leave the code as is for this PR, but then investigate in an alternative and potentially more performant solution.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/SpringNavigationAccessControl.java
mcollovati: added javadoc

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/SpringAccessPathChecker.java
mcollovati: done

vaadin-spring/src/main/java/com/vaadin/flow/spring/AuthenticationUtil.java
mcollovati: done

flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessChecker.java
mcollovati: done

flow-server/src/main/java/com/vaadin/flow/server/auth/DefaultNavigationCheckDecisionResolver.java
mcollovati: done

flow-server/src/main/java/com/vaadin/flow/server/auth/AnnotatedViewAccessChecker.java
mcollovati: done

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
mcollovati: Good suggestion.
Enabled state can be modified by providing a custom `NavigationAccessControlConfigurer` bean, but this is indeed an overkill to only disable the access control feature, when extending `VaadinWebSecurity`.
I'll add the method, but would name it `enableNavigationAccessControl()` rather than `isNavigationAccessControlEnabled()`

mcollovati: done

mshabarov (approved):

PR #17990, opened 2023-11-06
chore: Run tests with Bun in a separate profile

mshabarov:
Bun needs to be installed on system manually.
This moves Bun test modules to a separate profile to not disturb local runs and CI validation, since Bun is not installed there.


mcollovati (approved):
Changes to the workflow works correctly (see #18021)

PR #17992, opened 2023-11-06
fix: Expose Properties in VaadinFlowPluginExtension

### LINKED ISSUE #17665: Gradle project build fails if build script triggers VaadinPrepareFrontendTask init
### Description of the bug

Since Flow 23.3.4, with [this commit](https://github.com/vaadin/flow/commit/90e7fa9968726474a1f3f070bb48c8d1d51f9641), `VaadinPrepareFrontendTask` [init](https://github.com/vaadin/flow/blob/main/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPrepareFrontendTask.kt#L71-L77) method requires some values from the `VaadinFlowPluginExtension`, the values have default values as null, and are calculated in the `VaadinFlowPluginExtension`'s [autoconfigure ](https://github.com/vaadin/flow/blob/main/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt#L308-L323)method, which is only called in the [project.afterEvaluate](https://github.com/vaadin/flow/blob/main/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPlugin.kt#L54-L71). 

So that if any code in the build.gradle file triggers the `VaadinPrepareFrontendTask`'s init method. sth as simple as 
```
project.tasks.whenTaskAdded {task ->
}
```
fails the build.

Steps to reproduce:
1. Import the project linked in the Minimal reproducible example
2. IDEA already throws error 
```
A problem occurred evaluating root project 'base-starter-gradle'.
> Could not create task ':vaadinPrepareFrontend'.
   > Could not create task of type 'VaadinPrepareFrontendTask'.
      > java.lang.NullPointerException (no error message)
```
3. If not seeing the error directly in IDEA, try to run any task, e.g. `./gradlew clean`, there should be errors in the terminal.

### Expected behavior

The project should build normally

https://github.com/haijian-ccc/base-starter-gradle

It's a clone from base-starter-gradle with only [a one liner change](https://github.com/vaadin/base-starter-gradle/compare/v24...haijian-ccc:base-starter-gradle:v24)

### Versions

- Vaadin / Flow version: 23.3.4+, including 24
- Java version: 17
- OS version:
- IDE (if applicable): IntelliJ IDEA 2022.2.3 (Community Edition)

mvysny:
## Description

Refactor the `VaadinFlowPluginExtension` class to expose `Properties` instead of concrete types, as recommended by https://docs.gradle.org/current/userguide/lazy_configuration.html . This also "breaks" the cycle described at https://github.com/vaadin/flow/issues/17665#issuecomment-1792243281 .

Note that this is a potentially breaking change (since the property types changed); however the existing Gradle build scripts will continue operating as usual, since Gradle contains some magic hacks (both for Groovy and for Kotlin) so that `vaadin { foo = bar }` property assignment syntax will continue working as usual.

Fixes #17665

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


tepi (approved):
Looks good to me. Tested on a gradle project, checking prepareFrontend caching and the issue in https://github.com/vaadin/flow/issues/17665 as well.

mvysny:
Awesome, thank you @tepi for such a quick review & thorough testing. I've set the sources to auto-merge, hopefully the change will go through automatically :crossed_fingers:

Issue #16886, opened 2023-05-26
Allow to reuse the production bundle when files have not changed.


stefanuebe:
### Describe your motivation

(based on our internal discussion)

While the newly introduced production bundle (PB) is intended to prevent devs from having to install and indirectly use node / npm, it  could also help speed up integration tests on developer machines, as ITs should always be run in production mode. 

Unfortunately, there is currently no way to keep the PB, when using custom JS/TS files, even when they haven't changed. Therefore this feature is not usable for more complex projects running ITs.

### Describe the solution you'd like

It should be possible to keep the production bundle, when the respective file hashes have not changed. ~~To prevent any build server complications, this feature should be turned off by default. A configuration flag (e.g. "keep production bundle") allows to enable/disable it. Our starter application.property should however set the flag to true, so that developers have this feature enabled by default.~~
The feature should be enabled by default so that developers, who run ITs on their machines, may benefit from this new feature automatically. A configuration flag should be provided so that this feature can be disabled.

benstpierre:
This would relieve a huge pain point for us.

mshabarov:
**Acceptance criteria:**
- [ ] When default production bundle is not enough to run the application, because of JS/TS changes, Flow should re-build the prod bundle, copy it into build directory (as it works now) and additionally save the bundle in the project, in the `src/main/prod-bundle` directory, similar to `dev-bundle`
- [ ] On each subsequent prod build runs Flow should check if existing bundle in `src/main/prod-bundle` is enough for project (same checks as for dev-bundle) and re-copy the bundle from this dir to build directory
- [ ] Default is to re-use the custom bundle if possible, existing parameter `vaadin.force.production.build=true` can be used to trigger the build always
- [ ] License checking is done always for custom prod bundle.

mshabarov:
@stefanuebe I propose to have this feature by default, why do you think re-bundle by default is better?

stefanuebe:
> @stefanuebe I propose to have this feature by default, why do you think re-bundle by default is better?

I think this feature will be primarly used by devs to run ITs or test prod builds on their own machines to see, if things can be committed. Time is an important resource in that case since ITs on their own already can eat up a lot of it.

On the other hand, production builds are most times run by CI test servers, which in my opinion should not activate this feature, so that they really have the latest state of development when building things (e.g. to see if all dependencies are still available, no caches that affect builds, etc). Plus they do not have the time pressure that a developer has. 

[...]

PR #17997, opened 2023-11-07
feat: reuse the production bundle if no changes

tltv:
Makes a copy of the production bundle in `src/main/bundles/prod.bundle` zip file to make it possible to reuse it without recreating same unchanged bundle again. When `prod.bundle` exists, it will be used as a primary source for `stats.json` file.

Fixes: https://github.com/vaadin/flow/issues/16886

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -518,9 +518,11 @@ public class BuildFrontendUtil {
             File statsFile = new File(adapter.servletResourceOutputDirectory(),
                     Constants.VAADIN_CONFIGURATION + "/stats.json");
             if (!statsFile.exists()) {
-                // If no compiled bundle available check for jar-bundle
-                statsJsonContent = BundleValidationUtil
-                        .findProdBundleStatsJson(adapter.getClassFinder());
+                // If no compiled bundle available check for prod.bundle and
+                // jar-bundle
+                statsJsonContent = ProdBundleUtils.findBundleStatsJson(
+                        adapter.projectBaseDirectory().toFile(),
+                        adapter.getClassFinder());
             } else {
                 statsJsonContent = IOUtils.toString(statsFile.toURI().toURL(),
                         StandardCharsets.UTF_8);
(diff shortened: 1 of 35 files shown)


flow-server/src/test/java/com/vaadin/flow/server/frontend/BundleValidationTest.java
```diff
@@ -1822,11 +1832,8 @@ public void forceProductionBundle_bundleRequired() {
     }
 
     @Test
-    public void noDevFolder_devBundleCompressedFileExists_noBuildRequired()
+    public void noDevFolder_bundleCompressedFileExists_noBuildRequired()
```
caalador, line 1835: I would perhaps do another test for prod bundle as now the test name says noDevFolder, but also tests production bunlde that doesn't use the devFolder at all for validation.
The use case is a bit different for the 2 cases.

flow-server/src/test/java/com/vaadin/flow/server/frontend/BundleValidationTest.java
```diff
@@ -1874,11 +1887,8 @@ public void noDevFolder_devBundleCompressedFileExists_noBuildRequired()
     }
 
     @Test
-    public void noDevFolderOldDefaultBundle_devBundleCompressedFileExists_noBuildRequired()
+    public void noDevFolderOldDefaultBundle_bundleCompressedFileExists_noBuildRequired()
```
caalador, line 1890:
```suggestion
    public void noFileBundleOrJar_compressedBundleExists_noBuildRequired()
```

caalador (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/frontend/BundleValidationTest.java
tltv: Actually this test seems to be wrong, it should mock to return null URL for getResource call or otherwise it would mean that default bundle is indeed found from the jar. However, fixing this won't affect test result as existing custom bundle file prevents the result from changing anyway.

flow-server/src/test/java/com/vaadin/flow/server/frontend/BundleValidationTest.java
tltv: Done

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
```diff
@@ -429,6 +426,13 @@ public static void runFrontendBuild(PluginAdapterBase adapter)
         FrontendTools tools = new FrontendTools(settings);
         tools.validateNodeAndNpmVersion();
         BuildFrontendUtil.runVite(adapter, tools);
+        copyCompressedBundleToProdBundleLocation(adapter);
+    }
+
+    private static void copyCompressedBundleToProdBundleLocation(
```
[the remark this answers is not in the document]
tltv: Done

caalador (approved):

PR #18013, opened 2023-11-10
chore: remove mobile dnd polyfill

tomivirkki:
## Description

This pull request removes the use of ["mobile-drag-drop" polyfill](https://github.com/timruffles/mobile-drag-drop) from Flow codebase. The polyfill was initially introduced to address drag-and-drop limitations in iOS Safari versions prior to 15. With V24's support now starting from iOS version 15, which includes [native DnD support](https://caniuse.com/dragndrop), the polyfill has become obsolete and can be removed.

Related change https://github.com/vaadin/flow-components/pull/5686


czp13 (approved):

knoobie:
@tomivirkki @mshabarov Drag'n'Drop is no longer working in Vaadin 24.2.5 for MacOS with Firefox and Safari

tomivirkki:
Hi @knoobie, this seems to be an overlay-related regression from https://github.com/vaadin/web-components/pull/6850

Workaround: `dialog.getElement().executeJs("this.$.overlay._onMouseDown = () => {}");`

knoobie:
Thanks for looking into this Tomi! Sadly the workaround isn't working when executed from flow; calling it from devtools after the dialog is rendered it works

knoobie:
Ups.. my fault.. your first solution works like a charm :) Thanks again! Trying this with an application with dozen of views and dialogs.. adding this to the wrong dialog doesn't help..

PR #18019, opened 2023-11-13
fix: improve error message when a component is moved from one UI to another.

### LINKED ISSUE #9376: Add clarifying information to the error message about moving nodes to another state tree
The "Can’t move a node from one state tree to another" exception is quite often triggered by accidentally using an inappropriate DI scope for some specific component class, thus causing an instance of that component to become accidentally reused.

It would be much easier to debug that kind of problem to trace down which class has inappropriate scope if the error message would better identify the component in question. To achieve that, I would propose that the logic that throws the exception would have a special logic to check if the state node is an element node and that element has a component associated. In that case, the class name of the component would also be included in the error message.

mvysny:
## Description

Improve error message when a component is moved from one UI to another.

Fixes #9376 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
@@ -36,8 +36,12 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentUtil;
 import com.vaadin.flow.component.UI;
+import com.vaadin.flow.component.internal.ComponentTracker;
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.dom.ElementUtil;
 import com.vaadin.flow.function.SerializableConsumer;
 import com.vaadin.flow.internal.StateTree.BeforeClientResponseEntry;
 import com.vaadin.flow.internal.StateTree.ExecutionRegistration;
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
```diff
@@ -765,18 +769,51 @@ private void doSetTree(StateTree tree) {
...
+                // https://github.com/vaadin/flow/issues/9376
+
                 throw new IllegalStateException(
                         "Can't move a node from one state tree to another. "
                                 + "If this is intentional, first remove the "
                                 + "node from its current state tree by calling "
-                                + "removeFromTree");
+                                + "removeFromTree. This usually happens when a "
+                                + "component is moved from one UI to another, "
+                                + "which is not a good idea. This may be caused "
```
mcollovati, line 781: nit: I would maybe replace the "is not a good idea" with "is not recommended"

flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
```diff
@@ -765,18 +769,51 @@ private void doSetTree(StateTree tree) {
...
+        final ComponentTracker.Location createLocation = ComponentTracker
+                .findCreate(component);
+        final ComponentTracker.Location attachLocation = ComponentTracker
+                .findAttach(component);
+        if (createLocation != null || attachLocation != null) {
+            // the location.toString() includes the component class as well
+            return "created: " + createLocation + ", attached: "
+                    + attachLocation;
+        }
+        // createLocation may be null in production mode. Just return the
```
mcollovati, line 811:
```suggestion
        // createLocation is null in production mode. Just return the
```

mcollovati (changes requested):

mcollovati:
Some minor changes requested, but LGTM. Great job @mvysny

flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
mvysny: I agree, fixed.

flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
mvysny: Thanks, fixed.

mvysny:
Thank you @mcollovati , all items should now be resolved.

flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
```diff
@@ -765,18 +769,51 @@ private void doSetTree(StateTree tree) {
                     ((StateTree) getOwner()).getUI(),
                     ReplacedViaPreserveOnRefresh.class) == null;
             if (isOwnerAttached || isNotReplaced) {
+                // provide a more verbose error message:
+                // https://github.com/vaadin/flow/issues/9376
+
                 throw new IllegalStateException(
                         "Can't move a node from one state tree to another. "
```
czp13, line 776: super small: but it is maybe -I think- a bit more readable and maintainable if the exception message is extracted to a `final static String variable,` and potentially the same variable could be used from the tests to assert if it was the proper error message.

czp13 (approved):
Looks good to me as well, nice work @mvysny!

mcollovati (approved):

flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
mvysny: Ooops, sorry, I merged too hastily and missed your comment. I personally think it's better to have the string constants in-place for better readability, instead of being out-of-place in a String constant, but it's not a very strong opinion. If you'd still like me to extract the message, please reach out to me on Slack and I'll prepare another PR.

czp13: No worries, it was a smallish idea/question. Let us not focus on this or block your time or anything :).
And thank you for replying, and taking my comment nicely! 🙇

mvysny:
More docs on the topic https://mvysny.github.io/cant-move-node/

PR #18022, opened 2023-11-14
fix: support accept property in StreamReceiverHandler

tltv:
StreamReceiverHandler understands `accept` property of the target element to block unwanted mime types. `accept` property takes comma separated list of mime types like 'text/plain, text/html'.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
@@ -456,7 +456,17 @@ public class StreamReceiverHandler implements Serializable {
             return false;
         }
         for (String acceptValue : accept.split(",")) {
-            if (acceptValue.trim().equals(mimeType)) {
+            String trimmedAcceptValue = acceptValue.trim();
+            if (trimmedAcceptValue.equals("*/*")) {
+                return false;
+            }
+            if (trimmedAcceptValue.endsWith("/*")) {
+                if (mimeType.startsWith(trimmedAcceptValue.substring(0,
+                        trimmedAcceptValue.length() - 1))) {
+                    return false;
+                }
+            }
+            if (trimmedAcceptValue.equals(mimeType)) {
                 return false;
             }
         }
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
@@ -15,8 +15,6 @@
  */
 package com.vaadin.flow.server.communication;
 
-import javax.naming.SizeLimitExceededException;
-
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 import java.io.BufferedWriter;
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/StreamReceiverHandler.java
```diff
@@ -434,6 +440,29 @@ protected boolean handleFileUploadValidationAndData(VaadinSession session,
...
+            element = Element.get(stateNode);
+        } catch (IllegalArgumentException e) {
+            return false;
+        }
+        String accept = element.getProperty("accept");
+        if (accept == null || accept.isBlank()) {
+            return false;
+        }
+        for (String acceptValue : accept.split(",")) {
+            if (acceptValue.trim().equals(mimeType)) {
```
knoobie, line 459: Note: this is probably not as easy as you think :) The `accept` attribute takes some interesting combinations: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept (even the JavaDoc of the Upload has some examples with `/*` or `.pdf`. 

Note 2: Should this throw an exception or FileRejectedEvent, which mentions `setAcceptedFileTypes()` of the Upload component.

tltv: At least wildcard support need to be supported, so I'm adding that. Not so sure about other parts.

tltv: I'll add also support for file extension format like `.pdf`.

tltv: Wildcard like `text/*` and file extension like `.pdf` is now supported based on the javadoc for `setAcceptedFileTypes()`.

About that second note: I don't think that throwing `FileRejectedEvent` or other exception is necessary here as accepted file types are checked primarily on the browser already. This is secondary check in the server side.

knoobie: It depends, if you really consider the client side "checked" by the browser, I've shoot @mshabarov a comment regarding this assumption in discord privately yesterday.

I have suggested `FileRejectedEvent` because it is mentioned in the JavaDoc of the Upload Component. So when this kind of upload arrives, the end-user (developer) should be notified about this by the rejection event, so that he can handle this security event accordingly in my opinion, instead of a generic "upload failed" (The Javadoc in the end is "public API" that describes the behavior).

tltv: Ok, I see your point regarding the event throwing. It could be useful in some use cases. That could be added as a separate new feature later.

mshabarov: Technically, `FileRejectedEvent` class belongs to the Upload component, so right now we cannot use it in the Flow core, because we cannot have dependencies on components in the Flow core.
But I agree that this might be a useful feature, if we transfer this exception into Flow and reuse it in StreamResourceHandler. I agree with Tomi that this should be a separate development effort.

knoobie: Understandable arguments :) And you probably don't wanna play with Classloading and Reflection magic in the meantime :)

mshabarov (approved):

knoobie:
General question: No tests? :)

PR #18028, opened 2023-11-15
test: reset ComponentTracker.productionMode static field

mcollovati:
Prevents unexpected situation in ComponentTest caused by the ComponenTracker.productionMode flag being initialized to true by some other tests.
In particular, cannotMoveComponentsToOtherUI test may fail because the exception message is different in development and production modes.

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/component/ComponentTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/ComponentTest.java
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.component;
 
+import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


mcollovati:
An example failure from CI job
```
java.lang.AssertionError: Can't move a node from one state tree to another. If this is intentional, first remove the node from its current state tree by calling removeFromTree. This usually happens when a component is moved from one UI to another, which is not recommended. This may be caused by assigning components to static members or spring singleton scoped beans and referencing them from multiple UIs. Offending component: created: Component 'org.junit.runners.model.FrameworkMethod$1' at 'FrameworkMethod.java' (runReflectiveCall LINE 59), attached: Component 'org.junit.runners.model.FrameworkMethod$1' at 'FrameworkMethod.java' (runReflectiveCall LINE 59)
```

mvysny (approved):
I agree, it's a good idea to clean up the flag after every test, to revert the environment to a known state. Looks good!

mcollovati:
Picked manually

PR #18032, opened 2023-11-15
test: replace chrome mobile profile with custom settings

mcollovati:
Prevents PWA test failures with Chrome 119 because of 'invalid argument: Empty userAgent invalid with userAgentMetadata provided'

### DIFF
--- a/flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeDeviceTest.java
+++ b/flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeDeviceTest.java
@@ -97,7 +97,8 @@ public class ChromeDeviceTest extends ViewOrUITest {
     protected ChromeOptions customizeChromeOptions(
             ChromeOptions chromeOptions) {
         final Map<String, Object> mobileEmulationParams = new HashMap<>();
-        mobileEmulationParams.put("deviceName", "Laptop with touch");
+        mobileEmulationParams.put("deviceMetrics",
+                Map.of("width", 1280, "height", 950, "pixelRatio", 1));
 
         chromeOptions.setExperimentalOption("mobileEmulation",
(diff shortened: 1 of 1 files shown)


flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeDeviceTest.java
```diff
@@ -97,7 +97,8 @@ public void setup() throws Exception {
     protected ChromeOptions customizeChromeOptions(
             ChromeOptions chromeOptions) {
         final Map<String, Object> mobileEmulationParams = new HashMap<>();
-        mobileEmulationParams.put("deviceName", "Laptop with touch");
```
czp13, line 100: Is it okay or safe to delete the `deviceName`? 
We are not using it for any reporting or anywhere in the tests I guess.

czp13 (approved):
Seems safe to me :), only one minor thingie, but as tests were passing this should be safe! 🙇 

(I am also reviewing this as reviewed the previous PR in this topic :))

flow-test-util/src/main/java/com/vaadin/flow/testutil/ChromeDeviceTest.java
mcollovati: You cannot mix `deviceName` with other settings otherwise chromedriver will fail to create the browser session.

mcollovati: And with `deviceName` you get an empty `userAgent` that fails with the latest Chrome version

czp13: Okey, that makes sense 🙇! Cheers.

Issue #17941, opened 2023-10-27
The `vaadinPrepareFrontendTask` gradle task is not cached


steve-todorov:
### Description of the bug

We have upgrade to Vaadin `24.2.0` with the hopes to benefit from the newly introduced incremental Gradle task caching for `vaadinPrepareFrontendTask` (related PR is https://github.com/vaadin/flow/pull/17150). 

Unfortunately our local and CI builds seem to be unable to benefit from this feature.
Upon closer investigation we noticed Gradle scans show the following notice:

![image](https://github.com/vaadin/flow/assets/2303663/15c665ab-5480-4be7-a5f6-543ff16b541c)

### Expected behavior

The `vaadinPrepareFrontendTask` should show `UP-TO-DATE` or `FROM CACHE` in the gradle build scans and the task should not be computing again.

### Minimal reproducible example

Not sure how to reproduce yet, but we don't have any special setup.
We are using the `gradle vaadin plugin` and run `gradle build`.

### Versions

- Vaadin / Flow version: 24.2.0
- Java version: 17
- OS version: ubuntu-22.04 (GH CI)
- Browser version (if applicable):
- Application Server (if applicable):
- IDE (if applicable):

mshabarov:
Thanks for the issue. We will try to reproduce with our Gradle-based starters and get back to you soon.

steve-todorov:
Thanks for looking into this @mshabarov! 

Let me know if you need me to give you additional context or if you need anything more.

tepi:
Tested this out with Vaadin 24.2.0 and the [base starter](https://github.com/vaadin/base-starter-gradle). There the caching seems to work fine. This is the output I get:

steve-todorov:
Hey @tepi,

The `base starter` project does not appear to have a `package.json` and `package-lock.json` files.
In our project we have both of them. My current working theory is the `vaadinBuildFrontend` task might be corrupting the `cache`, because sometimes it updates packages in the `package.json` which were not committed into the git.

![image](https://github.com/vaadin/flow/assets/2303663/d0b3f95d-8c48-46bf-b673-4176131fd6f4)

So what appears to be happening is that the `vaadinBuildFrontend` is overriding the `package.json` file and thus changing the cache key that was initially generated by the `vaadinPrepareFrontend` task. I have been able to reproduce the problem with the base starter project.

You can have a look in my fork located here: 
https://github.com/steve-todorov/base-starter-gradle/tree/issue-17941

[...]

PR #18045, opened 2023-11-16
Do not touch package.json files unless content has changed

tepi:
In some cases, at least the `package.json` file is overwritten with the same content it already had. This causes the timestamp to be updated, which in some environments causes file fingerprint to differ. 

Fixes https://github.com/vaadin/flow/issues/17941


oliveryasuna:
@tepi Thanks for handling this so quick!

It may be quicker to simply compare the hashes, since they are already computed.

caalador (approved):

mshabarov (approved):
Nice fix, thanks 👍

PR #18054, opened 2023-11-16
feat: Implement RangeInput component

### LINKED ISSUE #3483: Simple Input Range AKA Slider
`When building UIs in Java, I want to show the user a ranged selected, so that they can select a value from a range easily and see where the current selection is on the range`

Would be quite easy add the native `<input type="range>` to `flow-html-components`.

mvysny:
## Description

Implement a component for `<input type="range">`.

Fixes #3483 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
@@ -15,7 +15,14 @@
  */
 package com.vaadin.flow.component.html;
 
-import com.vaadin.flow.component.*;
+import com.vaadin.flow.component.AbstractSinglePropertyField;
+import com.vaadin.flow.component.Focusable;
+import com.vaadin.flow.component.HasAriaLabel;
+import com.vaadin.flow.component.HasSize;
+import com.vaadin.flow.component.HasStyle;
+import com.vaadin.flow.component.PropertyDescriptor;
+import com.vaadin.flow.component.PropertyDescriptors;
+import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.data.value.HasValueChangeMode;
 import com.vaadin.flow.data.value.ValueChangeMode;
 
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2000-2023 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
 package com.vaadin.flow.component.html;
 
 import com.vaadin.flow.component.*;
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
@@ -127,7 +127,7 @@ public class RangeInput extends AbstractSinglePropertyField<RangeInput, Double>
     }
 
     /**
-     * Returns the minimum value.
+     * Gets the minimum value.
      *
      * @return the minimum value, defaults to 0.
      */
@@ -146,7 +146,7 @@ public class RangeInput extends AbstractSinglePropertyField<RangeInput, Double>
     }
 
     /**
-     * Returns the maximum value.
+     * Gets the maximum value.
      *
      * @return the maximum value, defaults to 100.
      */
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
@@ -1,0 +1,256 @@
+package com.vaadin.flow.component.html;
+
+import com.vaadin.flow.component.*;
+import com.vaadin.flow.data.value.HasValueChangeMode;
+import com.vaadin.flow.data.value.ValueChangeMode;
+
+import java.util.Arrays;
+
+/**
+ * Creates a new input element with type "range".
+ * <p>
+ * </p>
+ * Note: Slider doesn't support the read-only mode and will disable itself
+ * instead.
+ */
+@Tag(Tag.INPUT)
+public class RangeInput extends AbstractSinglePropertyField<RangeInput, Double>
+        implements Focusable<RangeInput>, HasSize, HasStyle, HasValueChangeMode,
+        HasAriaLabel {
+    private static final PropertyDescriptor<String, String> typeDescriptor = PropertyDescriptors
+            .attributeWithDefault("type", "text");
+    private static final PropertyDescriptor<String, String> minDescriptor = PropertyDescriptors
+            .attributeWithDefault("min", "0");
+    private static final PropertyDescriptor<String, String> maxDescriptor = PropertyDescriptors
+            .attributeWithDefault("max", "100");
+    private static final PropertyDescriptor<String, String> stepDescriptor = PropertyDescriptors
+            .attributeWithDefault("step", "1");
+    private static final PropertyDescriptor<String, String> orientDescriptor = PropertyDescriptors
+            .attributeWithDefault("orient", Orientation.HORIZONTAL.getValue());
+
+    /**
+     * The orientation of the range slider.
+     */
+    public enum Orientation {
+        HORIZONTAL("horizontal"), VERTICAL("vertical");
+
+        private final String value;
+
+        Orientation(String value) {
+            this.value = value;
+        }
+
+        public String getValue() {
+            return value;
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


tepi:
Not sure if `RangeInput` should/could extend `com.vaadin.flow.component.html.Input`? At least some methods are duplicated.

mvysny:
@tepi unfortunately the type of the value of the Input component is fixed to String and can not be changed in overriding class. I thought of having a CustomField, wrapping Input, but that was even more complicated. I believe the current approach to be the most straightforward one.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+        return valueChangeTimeout;
+    }
+
+    private void applyChangeTimeout() {
+        ValueChangeMode.applyChangeTimeout(currentMode, valueChangeTimeout,
+                getSynchronizationRegistration());
+    }
+
+    /**
+     * Returns the minimum value.
```
mcollovati, line 123: nit: Flow team conventions, use "gets" instead of "returns"
```suggestion
     * Gets the minimum value.
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+     *
+     * @param min
+     *            the minimum value.
+     */
+    public void setMin(double min) {
+        set(minDescriptor, "" + min);
+    }
+
+    /**
+     * Returns the maximum value.
```
mcollovati, line 142: nit: Flow team conventions, use "gets" instead of "returns"
```suggestion
     * Gets the maximum value.
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+     * @param orientation
+     *            the orientation, not null. Defaults to
+     *            {@link Orientation#HORIZONTAL}.
+     */
+    public void setOrientation(Orientation orientation) {
+        set(orientDescriptor, orientation.getValue());
+    }
+
+    /**
+     * Non-standard Attribute.
```
mcollovati, line 223:
```suggestion
     * Gets the orientation of the range slider
     * <p>
     * </p>
     * Non-standard Attribute.
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+     *
+     * @param step
+     *            the new step value, may be null.
+     */
+    public void setStep(Double step) {
+        set(stepDescriptor, step == null ? "any" : "" + step);
+    }
+
+    /**
+     * Non-standard Attribute.
```
mcollovati, line 207:
```suggestion
     * Sets the orientation of the range slider
     * <p>
     * </p>
     * Non-standard Attribute.
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+     * Range input element doesn't support the "read-only" attribute or
+     * property. We'll disable the component instead.
+     */
+    private boolean readOnly = false;
+    private boolean enabled = true;
+
+    @Override
+    public void setReadOnly(boolean readOnly) {
+        this.readOnly = readOnly;
+        updateEnabled();
```
mcollovati, line 249: What about performing the operation only if the new read-only value is different from the current one?

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+
+    @Override
+    public boolean isReadOnly() {
+        return readOnly;
+    }
+
+    @Override
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
+        updateEnabled();
```
mcollovati, line 260: What about performing the operation only if the new enabled value is different from the current one?

mcollovati (changes requested):

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+ * </p>
+ * Note: Slider doesn't support the read-only mode and will disable itself
+ * instead.
+ */
+@Tag(Tag.INPUT)
+public class RangeInput extends AbstractSinglePropertyField<RangeInput, Double>
+        implements Focusable<RangeInput>, HasSize, HasStyle, HasValueChangeMode,
+        HasAriaLabel {
+    private static final PropertyDescriptor<String, String> typeDescriptor = PropertyDescriptors
+            .attributeWithDefault("type", "text");
```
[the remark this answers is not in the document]
mvysny: The default value in this context refers to the default value of the attribute as specified by the HTML spec. According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input the "text" is the default.

If I would use `attributeWithDefault("type", "range");` and then set the PropertyDescriptor value to "range", it would actually remove the "type" attribute from the DOM, causing the input to be of type "text", which is not what we want.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.html;
+
+import com.vaadin.flow.component.*;
```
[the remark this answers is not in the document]
mvysny: Oops, thanks! Done.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+    public ValueChangeMode getValueChangeMode() {
+        return currentMode;
+    }
+
+    @Override
+    public void setValueChangeMode(ValueChangeMode valueChangeMode) {
+        currentMode = valueChangeMode;
+        setSynchronizedEvent(
+                ValueChangeMode.eventForMode(valueChangeMode, "input"));
+        applyChangeTimeout();
```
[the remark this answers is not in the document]
mvysny: This code is actually copied from the existing `Input` element. Since Input doesn't introduce this kind of optimization, I thought that this kind of optimization doesn't bring much performance benefits.

Yet, please let me know if you would still like me to perform this optimization.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    @Override
+    public void setValueChangeTimeout(int valueChangeTimeout) {
+        this.valueChangeTimeout = valueChangeTimeout;
+        applyChangeTimeout();
```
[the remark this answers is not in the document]
mvysny: This code is actually copied from the existing Input element. Since Input doesn't introduce this kind of optimization, I thought that this kind of optimization doesn't bring much performance benefits.

Yet, please let me know if you would still like me to perform this optimization.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
mvysny: Done.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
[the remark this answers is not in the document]
mcollovati: Got it, and I agree. Let's leave it as a potential future improvement.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
mvysny: Not sure about this. `HasValueAndElement.setReadOnly()` doesn't perform this kind of optimization, and this one is the default implementation for all HasValue. I'm not sure whether only this particular component should optimize on this.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
mvysny: Same as above: `HasEnabled.setEnabled()` doesn't perform this kind of optimization; I'm not sure whether only this particular component should optimize on this.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
mvysny: Done,  thank you.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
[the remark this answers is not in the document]
mcollovati: Ok, thanks. Let's do it the same way as in Input

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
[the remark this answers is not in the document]
mcollovati: Ok, thanks. Let's do it the same way as in Input

mcollovati (approved):

PR #18058, opened 2023-11-17
feat: add react router defaults

caalador:
Enabling the reactRouter feature flag will now 
add the necessary package dependencies and
generate the required frontend files
to Frontend/generated/flow

With this one should be able to run a flow project
using react router instead of vaadin-router


flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/react/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@polymer/polymer": "3.5.1",
+    "@vaadin/common-frontend": "0.0.18",
+    "construct-style-sheets-polyfill": "3.1.0",
+    "lit": "3.0.2",
```
knoobie, line 15: ``
```suggestion
    "lit": "3.1.0",
```

Artur has updated it to 3.1.0 --> updated that no mismatch happens

caalador: Need to have a shared/common dependencies so the versions do not differ.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -0,0 +1,106 @@
...
+import com.vaadin.flow.server.Version;
+import com.vaadin.open.App;
+
+import static com.vaadin.flow.server.frontend.FrontendUtils.INDEX_JS;
+import static com.vaadin.flow.server.frontend.FrontendUtils.INDEX_TS;
+import static com.vaadin.flow.server.frontend.FrontendUtils.INDEX_TSX;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * Generate <code>index.ts</code> if it is missing in frontend folder.
```
mshabarov, line 39: Should list the files actually generated, not `index.ts`.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -0,0 +1,106 @@
...
+                "flow/routes.tsx");
+        try {
+            if (!Path.of(frontendDirectory.getPath(), "App.tsx").toFile()
+                    .exists()) {
+                writeFile(appTsx, getFileContent("App.tsx"));
+            }
+
+            if (!Path.of(frontendDirectory.getPath(), "routes.tsx").toFile()
+                    .exists()) {
+                writeFile(routesTsx, getFileContent("routes.tsx"));
```
mshabarov, line 78: Just to clarify: if `routes.tsx` already exists, and Flow won't modify it, does it mean that adding `...serverSideRoutes` to it is a manual action required by developers? 

Can we automate it by adding it always if not there?

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
caalador: At the moment it means manually adding.
Automatically adding it has the problems of identifying if it is there and I at least do not know the expected setup types for the router that I could generate a good automatic add to the array/object/tree of an existing route setup.

mshabarov: Ok, fair enough, then we can consider doing this separately, as it needs a smart way of determining a proper place in `routes` to inject server-side routes.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
mshabarov, line 39:
```suggestion
 * Generates React related files <code>App.tsx</code> and <code>routes.tsx</code>
 * if they are missing in frontend folder, as well as <code>Flow.tsx</code> file,
 * that contains codes for integrating React router with Flow routes.
```

mshabarov: Or something similar.

caalador: Fixed javaDoc.

mshabarov (approved):

mshabarov (approved):
Tested with a case, when I had fresh Hilla+React starter app, added Flow views and run the app.
Then manually I was needed to add `serverSideRoutes` as follows
```javascript
import {serverSideRoutes} from "Frontend/generated/flow/Flow";
...
export const routes = [
  {
    ...
    children: [
      { path: '/', element: <HelloWorldView />, handle: { title: 'Hello World' } },
      { path: '/about', element: <AboutView />, handle: { title: 'About' } },
        ...serverSideRoutes
    ],
  },
] as RouteObject[];
...
```
Then the navigation worked as expected.
We may probably want to add `import {serverSideRoutes} from "Frontend/generated/flow/Flow";` automatically, but this is for further improvements.

PR #18072, opened 2023-11-20
fix: consider routes not-public if access control is disabled

mcollovati:
When navigation access control is disabled, Flow cannot decide whether a route is public or not.
This change makes the `isAnonymousRoute` method, used as Spring Security request matcher, always return false if the navigation access control is disabled.


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -193,14 +193,23 @@ private boolean isAnonymousRouteInternal(HttpServletRequest request) {
             return false;
         }
 
+        NavigationAccessControl navigationAccessControl = accessControl
+                .getObject();
+        if (!navigationAccessControl.isEnabled()) {
+            getLogger().debug(
+                    "Navigation Access Control disable. Cannot determine if {} refers to a public view",
```
knoobie, line 200: Would it make sense to only print this in development mode and there print it with INFO? If I disable the access control and do a production build, I hopefully know what I've done.. and in development I'm always reminded what I've done 😅

mcollovati: Using INFO level in development mode could be a good idea, to immediately spot a configuration issue.
However, I'll print the message also in production, at DEBUG level.

mcollovati: done

knoobie: Just something to consider: atLevel comes from slf4j 2.0 which could create problems if people use Vaadin in container that enforce older versions - even tho Vaadin and a lot of other libraries already depend on 2.0, using the "newer" APis could create a risk for those "older" app servers

mcollovati: Good point! Let's use a good old if statement

vaadin-spring/src/test/java/com/vaadin/flow/spring/security/RequestUtilTest.java
```diff
@@ -373,6 +373,36 @@ public void testApplyUrlMapping_nullMappedServlet_prependMapping() {
...
+
+        request = createRequest("other");
+        request.setServletPath("/");
+        Assert.assertTrue(requestUtil.isAnonymousRoute(request));
+
+        accessControl.setEnabled(false);
+        try {
+            request = createRequest("admin");
+            request.setServletPath("/");
+            Assert.assertFalse(requestUtil.isAnonymousRoute(request));
```
czp13, line 396: Maybe just for me but `Anonymous` is not the same as public (maybe it is referring to Anonymous being allowed). It would make sense in my eyes to overload this method (as renaming can be challenging for backward compatibility) with `isPublicRoute()` or `isAnonymousAllowedRoute()` (isPublicRoute is better to my eyes) method and deprecate the other.

czp13 (dismissed):
I had a nitty comment, but looks good to me. Nice work.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -193,15 +193,28 @@ private boolean isAnonymousRouteInternal(HttpServletRequest request) {
             return false;
         }
 
+        boolean productionMode = service.getDeploymentConfiguration()
+                .isProductionMode();
+        NavigationAccessControl navigationAccessControl = accessControl
+                .getObject();
+        if (!navigationAccessControl.isEnabled()) {
+            String message = "Navigation Access Control disable. Cannot determine if {} refers to a public view";
```
mshabarov, line 201:
```suggestion
            String message = "Navigation Access Control is disabled. Cannot determine if {} refers to a public view, thus access is denied. Please add an explicit request matcher rule for this URL.";
```

mshabarov (changes requested):
Message text suggestion, otherwise looks good.

vaadin-spring/src/test/java/com/vaadin/flow/spring/security/RequestUtilTest.java
mcollovati: I think the name originally related to the `@AnonymousAllowed` annotation put on a view to make it public.
I could make sense to rename it and add a deprecate method with the current name.
@mshabarov what do you think?

mshabarov: I don't see any problems with the naming here. One can argue that "anonymous" better reflects the status of user in terms of authentication and this name is better aligned with the annotation. I'd proposed to revisit it if we see any further complains from community.

mshabarov (approved):

PR #18078, opened 2023-11-20
fix: postpone commands scheduled with runWhenAttached for detached nodes

### LINKED ISSUE #18020: DataCommunicator flush not cancelled when grid is detached between registration and execution of flush
### Description of the bug

The DataCommunicator delays flushing data to the client using `UI::beforeClientResponse`.

Usually, when the Grid is removed after calling refreshAll, flushing the data is cancelled, which is as expected.

However, if removing the grid also happens inside `UI::beforeClientResponse` but before flushing the data, the data is still flushed.

In our case, this led to a NullPointerException, because we nulled out objects that are needed needed in a column's ValueProvider.

### Expected behavior

The example application below should remove the grid gracefully when clicking the button.

Instead, it currently fails with a NullPointerException, because it tries to refresh the grid after it has been detached, since one of column's ValueProviders uses a field that has been set to null on detach.

The exception should not happen because flushing the data should be cancelled when the grid is detached.

Open the following view in a browser and click on the button:

```java
public class HelloWorldView extends VerticalLayout {

    private ValueProvider<String, Integer> valueProvider = String::length;

    public HelloWorldView() {
        var grid = new Grid<String>();
        grid.addColumn(ValueProvider.identity())
            .setHeader("Col 1");
        grid.addColumn(bean -> valueProvider.apply(bean))   // Uses field that will be set to null
            .setHeader("Col 2");
        grid.addDetachListener(e -> valueProvider = null);

        GridListDataView<String> dataView = grid.setItems("one", "two", "three");

        var button = new Button("Refresh and remove grid", e -> {
            // before client response, first remove the grid before flushing
            getUI().orElseThrow().beforeClientResponse(grid, ctx -> grid.removeFromParent());
            // request flush
            dataView.refreshAll();
        });

        setMargin(true);
        add(button, grid);
    }
}
```

### Versions

- Vaadin / Flow version: 24.2.2
- Java version: Eclipse Temurin 17.0.8
- OS version: Windows 10 64 bit

mcollovati:
## Description

When detach listeners are called, the node still has a parent set, causing commands scheduled with runWhenAttached to be executed immediately. This change make runWhenAttached postpone the command by registering an attach listener, if the node has been detached during the same roundtrip.

Fixes vaadin/flow-components#1434
Fixes #18020

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
@@ -956,8 +956,9 @@ public class StateNode implements Serializable {
      *            the command to run immediately or when the node is attached
      */
     public void runWhenAttached(SerializableConsumer<UI> command) {
-
-        if (isAttached()) {
+        // Execute immediately only if the node han not been detached
+        // during current roundtrip
+        if (isAttached() && (!wasAttached || !this.hasBeenDetached)) {
             command.accept(getUI());
         } else {
             addAttachListener(new Command() {
--- a/flow-server/src/test/java/com/vaadin/flow/internal/StateNodeTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/internal/StateNodeTest.java
@@ -569,6 +569,29 @@ public class StateNodeTest {
     }
 
     @Test
+    public void runWhenAttachedNodeDetachedInSameRoundTrip() {
+        StateTree tree = createStateTree();
+        AtomicInteger commandRun = new AtomicInteger(0);
+        StateNode node = createEmptyNode();
+        setParent(node, tree.getRootNode());
+        node.removeFromTree();
+
+        node.runWhenAttached(ui -> {
+            Assert.assertEquals(tree.getUI(), ui);
+            commandRun.incrementAndGet();
+        });
+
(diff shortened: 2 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
```diff
@@ -956,8 +956,9 @@ public <T extends Serializable> T getChangeTracker(NodeFeature feature,
      *            the command to run immediately or when the node is attached
      */
     public void runWhenAttached(SerializableConsumer<UI> command) {
-
-        if (isAttached()) {
+        // Execute immediately only if the node han not been detached
+        // during current roundtrip
+        if (isAttached() && (!wasAttached || !this.hasBeenDetached)) {
```
DiegoCardoso, line 961: nit: 
```suggestion
        if (isAttached() && (!wasAttached || !hasBeenDetached)) {
```

DiegoCardoso (approved):
I can confirm that the related issue in Grid is gone after running flow-server with this change.

DiegoCardoso (approved):

mshabarov (approved):

Issue #18062, opened 2023-11-17
Websocket connection for debugWindow should only work from localhost.


czp13:
### Describe your motivation

The idea is to check if the browser web socket connection to the server is on the same (local)host. If it is not then it shall not reach the debugWindow implementation (devtools, plugins, etc).

### Describe the solution you'd like

Check the Atmosphere connection and see if we use localhost or not to make the connection.
Relevant class:
- https://github.com/vaadin/flow/blob/67a91564547e5d31a5cec3dbbbb9f33489ca634f/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java

### Describe alternatives you've considered

Fix this in the relevant project, which uses dev-server as dependency and so dev tools windows, but we discard, closed that PR and work as not optimal or feasible.

### Additional context

We need this fix to be able to fix an issue in another project.

Legioth:
One thing to note is that there might be scenarios when development mode is run on a different host name, e.g. if running the development server through Docker. There should be a configuration option to also allow other host names.

manolo:
We had to configure CI with the parameter to allow running tests in Selenoid

Artur-:
Implemented in #18079

PR #18079, opened 2023-11-20
feat: Restrict debug websocket connection to loopback addresses by default

czp13:
## Description

Adding a check to DebugWindow to see if it is connected from localhost or not. 
Additionally, it shall be configurable to allow customers to override and still use localhost if they wish.

WIP:
- need to add some option/method to customize it ✅ 
- need to add test ✅ 
- need to test it ✅ 

Fixes # (issue)
- https://github.com/vaadin/flow/issues/18062

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have added a description following the guide-line.


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -195,7 +208,23 @@ protected DevToolsInterface getDevToolsInterface(
 
     @Override
     public void onConnect(AtmosphereResource resource) {
+
         resource.suspend(-1);
```
Artur-, line 212: I don't know but I think we should not suspend the connection if it is not allowed.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -195,7 +208,23 @@ protected DevToolsInterface getDevToolsInterface(
...
+        AtmosphereRequest request = resource.getRequest();
+        String remoteAddress = request.getRemoteAddr();
+        if (!isLocalhostEnabled) {
+            try {
+                InetAddress inetAddress = InetAddress.getByName(remoteAddress);
+                if (inetAddress.isLoopbackAddress()) {
+                    return;
+                }
+            } catch (UnknownHostException e) {
+                getLogger().info("Unable to resolve remote address {}",
```
Artur-, line 223: If we are not sure it's localhost, we should probably prevent it and not allow it

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -195,7 +208,23 @@ protected DevToolsInterface getDevToolsInterface(
...
+
         resource.suspend(-1);
+
+        AtmosphereRequest request = resource.getRequest();
+        String remoteAddress = request.getRemoteAddr();
+        if (!isLocalhostEnabled) {
+            try {
+                InetAddress inetAddress = InetAddress.getByName(remoteAddress);
+                if (inetAddress.isLoopbackAddress()) {
+                    return;
```
Artur-, line 220: Isn't this the case that should be allowed, so it should run the init code below?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
czp13: Ok, let's not suspend it. Is there any way or idea how to communicate with the client. 

We have this below when we let the instructions going further and not exit from the method:
```
        resource.getBroadcaster().broadcast("{\"command\": \"hello\"}",
                resource);
               
```
 Maybe we could have something similar saying that other than localhost is not supported or just log out something and do not say anything to clients?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
czp13: Very true :), before writing tests, these kind of problems can happen :D I am fixing it!

czp13:
First test validation was flaky as I see, but this one shall be good.
(`BasicViewsITinvalidViewShowsNotFound()` was failing, some context, env was not ready my wild guess).

mcollovati:
As a side note, `BasicViewsIT` is indeed flaky, but it was not the cause of the build failure. The culprit was `FrontendLiveReloadIT.modifyThemeFile` (an issue already exists)

```
Error:  Failures: 
Error:  com.vaadin.flow.uitest.multimodule.ui.FrontendLiveReloadIT.modifyThemeFile[any_Chrome_]
Error:    Run 1: FrontendLiveReloadIT.modifyThemeFile:129 expected:<rgba([173, 216, 230, 1])> but was:<rgba([0, 0, 0, 0])>
Error:    Run 2: FrontendLiveReloadIT.modifyThemeFile:129 expected:<rgba([173, 216, 230, 1])> but was:<rgba([0, 0, 0, 0])>
Error:    Run 3: FrontendLiveReloadIT.modifyThemeFile:129 expected:<rgba([173, 216, 230, 1])> but was:<rgba([0, 0, 0, 0])>
[INFO] 
[INFO] 
Error:  Tests run: 3, Failures: 1, Errors: 0, Skipped: 0
```

If you check the logs relative to `BasicViewsIT`, you can see that they passed at the second attempt done in the same build

```
Warning:  Flakes: 
Warning:  com.vaadin.flow.servlet.BasicViewsIT.helloViewShown[any_Chrome_]
Error:    Run 1: BasicViewsIT.helloViewShown:37 » NoSuchElement No element with tag <*> found with the attributes [id='view'] using index 0
For documentation on this error, please visit: https://www.selenium.dev/documentation/webdriver/troubleshooting/errors#no-such-element-exception
Build info: version: '4.15.0', revision: '1d14b5521b'
System info: os.name: 'Linux', os.arch: 'amd64', os.version: '6.2.0-1016-azure', java.version: '17.0.9'
Driver info: driver.version: unknown
[INFO]   Run 2: PASS
[INFO] 
Warning:  com.vaadin.flow.servlet.BasicViewsIT.invalidViewShowsNotFound[any_Chrome_]
Error:    Run 1: BasicViewsIT.invalidViewShowsNotFound:44
[INFO]   Run 2: PASS
[INFO] 
Warning:  com.vaadin.flow.servlet.BasicViewsIT.rootViewShown[any_Chrome_]
Error:    Run 1: BasicViewsIT.rootViewShown:30 » NoSuchElement No element with tag <*> found with the attributes [id='view'] using index 0
For documentation on this error, please visit: https://www.selenium.dev/documentation/webdriver/troubleshooting/errors#no-such-element-exception
Build info: version: '4.15.0', revision: '1d14b5521b'
System info: os.name: 'Linux', os.arch: 'amd64', os.version: '6.2.0-1016-azure', java.version: '17.0.9'
Driver info: driver.version: unknown
[INFO]   Run 2: PASS
[INFO] 
[INFO] 
Warning:  Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Flakes: 3
```

czp13:
@mcollovati Hmm, interesting. Maybe my memory fails me, but I recall seeing a build fail due to the BasicViewsIT class's invalidViewShowsNotFound method. However, my rerun resulted in the validation passing. Perhaps there were multiple failures here due to flaky tests?

Anyway, I believe my added tests and the fixed tests should pass, so nothing related to these code changes on this PR should break the tests. 🙇 (I hope :D 🙏 )

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -145,6 +150,14 @@ public void setBackend(Backend backend) {
         this.backend = backend;
     }
 
+    public boolean isNonLocalhostEnabled() {
+        return isNonLocalhostEnabled;
+    }
+
+    public void setNonLocalhostEnabled(boolean localhostEnabled) {
```
Artur-, line 157: I do not understand who could call this and when. What I think we need instead is a system property where you can set the allowed ips, potentially using an asterisk to match many addresses

So by default only the loopback addresses would work but you can set ”192.168.1.2” to allow your other machine (or non docker machine) to connect. Alternatively you set ”192.168.*” to allow your colleagues to use the functionality.

czp13: :) Okay, we can use a system property or a system environment variable; let's go with a system property.

The requirement I read was about configurability. I did not see a lot of places in the Flow framework where the System variable option is used, so this will be something new. Therefore, I thought an API would be better for this, complete with proper JavaDoc :D, and also an update to the Flow documentation.

Artur-: There are a lot of `vaadin.something` properties, it is not something new

Artur-: They might be disguised as servlet init parameters though

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
Artur-: If there is no way to deny the connection, then we can just send `{command: 'access-denied'}` or similar and stop the connection in the browser

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
Artur-: We need to take the deny-by-default or deny-unless-we-know-it-is-safe approach to ensure it works as it should

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/DebugWindowConnectionTest.java
```diff
@@ -53,10 +64,83 @@ public void onConnect_suspend_sayHello() {
                 resource);
     }
 
+    @Test
+    public void onconnect_should_prevent_connection_from_non_localhost() {
```
Artur-, line 68: I don't think this tests anything anymore

Artur-: Actually this whole test class is kind of redundant. It should be in `IndexHtmlRequestHandlerTest`. There are only two code paths in `DebugWindowConnection` to test:
1. The correct token is present
2. The correct token is not present

czp13: okey, yap, not all tests are needed, I am happy to rename/delete some of them.

czp13: Fixed, should be better now:
- correct token
- invalid token
- no token
Just in case testing the last one too.

flow-server/src/test/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandlerTest.java
```diff
@@ -68,9 +70,7 @@
 
 import static com.vaadin.flow.server.Constants.VAADIN_WEBAPP_RESOURCES;
 import static com.vaadin.flow.server.frontend.FrontendUtils.INDEX_HTML;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertThrows;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
mcollovati, line 73: nit: please avoid wildcard imports

flow-server/src/test/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandlerTest.java
```diff
@@ -704,12 +685,42 @@ public void should_export_usage_statistics_in_development_mode()
...
+        assertTrue(isTokenPresent(indexHtml));
+
         String htmlContent = bodyInlineElements.get(0).childNode(0).outerHtml();
         htmlContent = htmlContent.replace("\r", "");
         htmlContent = htmlContent.replace("\n", " ");
         assertEquals(StringUtil.normaliseWhitespace(expected), htmlContent);
     }
 
+    // Regular expression to match a UUID in the format 8-4-4-4-12
+    private static final String UUID_REGEX = "\"token\":\"[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\"";
```
mcollovati, line 697: This could already be a Pattern instead of a String

mcollovati (changes requested):
Some minor changes for the moment. I'll continue the review tomorrow

flow-server/src/test/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandlerTest.java
czp13: Thanks, no, it is a fair comment. I remembered I changed my IntelliJ to never use wildcards, but maybe it was a project-specific configuration :(. Anyway Fixed!

flow-server/src/test/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandlerTest.java
czp13: Thanks, good point, more performant if does not need to create a Pattern all the time. Fixed.

Artur-:
It is a bit hqrd to view the changes on a phone but are all tests now missing for the logic in IndexHtmlRequestHandler? Like loopback accepted by default and nothing else + that matching with the property actually accepts and denies correctly- with and without a proxy?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -195,6 +196,17 @@ protected DevToolsInterface getDevToolsInterface(
...
+                .equals(resource.getRequest().getParameter("token"))) {
+            handleConnect(resource);
+        } else {
+            getLogger().debug(
+                    "Connection denied because of a missing or invalid token. The host is probably not on the allow list");
+            resource.resume();
+        }
+    }
+
+    public void handleConnect(AtmosphereResource resource) {
```
mcollovati, line 209: I think there is no need to have this method public

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -195,6 +196,17 @@ protected DevToolsInterface getDevToolsInterface(
 
     @Override
     public void onConnect(AtmosphereResource resource) {
+        if (IndexHtmlRequestHandler.RANDOM_DEV_TOOLS_TOKEN
+                .equals(resource.getRequest().getParameter("token"))) {
+            handleConnect(resource);
+        } else {
+            getLogger().debug(
+                    "Connection denied because of a missing or invalid token. The host is probably not on the allow list");
+            resource.resume();
```
mcollovati, line 205: Just a thought, but I'm not completely sure: if the connection should be denied, would it make sense to do a `resource.close()` instead of a `resume()`, so the underlying connection will be closed?

flow-server/src/test/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandlerTest.java
mcollovati, line 1: I may have missed something, but I can't see sensible tests related change introduced in IndexHtmlRequestHandler

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/DebugWindowConnectionTest.java
mcollovati, line 1: nit: the added tests seem very similar, probably we can refactor them to avoid copy/paste

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -358,6 +366,52 @@ private void addDevTools(Document indexDocument,
         indexDocument.body().appendChild(new Element("vaadin-dev-tools"));
     }
 
+    private boolean isAllowedDevToolsHost(AbstractConfiguration configuration,
+            VaadinRequest request) {
+        String remoteAddress = request.getRemoteAddr();
```
mcollovati, line 371: API states that `request.getRemoteAddr()` could potentially return `null`, and in this case, `InetAddress.getByName()` would return the address of the loopback interface.
Is this the intended behavior?

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -358,6 +366,52 @@ private void addDevTools(Document indexDocument,
...
+        String hostsAllowed = configuration
+                .getStringProperty("devmode.hostsAllowed", null);
+
+        if (!isAllowedDevToolsHost(remoteAddress, hostsAllowed)) {
```
mcollovati, line 375: Just a note: when documenting the `devmode.hostsAllowed` property, it will be worth it to highlight that if the pattern must also allow the proxy IP, when the server is behind a reverse-proxy

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -358,6 +366,52 @@ private void addDevTools(Document indexDocument,
...
+                return false;
+            }
+        }
+
+        return true;
+
+    }
+
+    private boolean isAllowedDevToolsHost(String remoteAddress,
+            String hostsAllowed) {
```
mcollovati, line 396: nit: maybe the method could directly get a Pattern instead of a string.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -358,6 +366,52 @@ private void addDevTools(Document indexDocument,
         indexDocument.body().appendChild(new Element("vaadin-dev-tools"));
     }
 
+    private boolean isAllowedDevToolsHost(AbstractConfiguration configuration,
```
mcollovati, line 369: Minor suggestion: if we make the method static and package protected, it may be easier to test the logic in isolation

mcollovati (changes requested):
Another round of comments, but I have not yet tested the feature

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
czp13: Okey, let's change to private, as we do not write a specific test for this only as well.

flow-server/src/test/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandlerTest.java
czp13: We have smoke tests to see if the tokens were added in correct format or not, but as you and Artur pointed out seems not enough.

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/DebugWindowConnectionTest.java
czp13: Okey, maybe the mocking part could be extracted and then remove some extra lines.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
czp13: Fair point! 👍

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
Artur-: Added code to deny if remote address is not available

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
Artur-: The reason for this is so you cannot just add a header and bypass the restriction

flow-server/src/test/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandlerTest.java
Artur-: Added tests

Artur-:
Changed the logic slightly so that localhost is always allowed. Otherwise you end up with the weird scenario with a local proxy that you would need to separately allow the local ip and the remote ip

vaadin-dev-server/src/test/java/com/vaadin/base/devserver/DebugWindowConnectionTest.java
czp13: Fixed, there should be fewer lines, hope you like this one better! 🙇

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
czp13: Made it private, thanks!

Artur-:
Changed the property from being a regexp to being a "wildcard match" using `*` and `?` to help people make fewer mistakes

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
Artur-: Not relevant anymore now that it is not a regexp

czp13: We are using the end this `FilenameUtils.wildcardMatch(remoteAddress, hostsAllowed);` 

We could rewrite it to something like this, to not use the `FilenameUtils`
```
        String regex = hostsAllowed.replace("?", ".").replace("*", ".*");
        Pattern pattern = Pattern.compile(regex);

        Matcher matcher = pattern.matcher(remoteAddress);
```

But this seems more complex and not sure if we wanna do this transformation and potentially make it less performant.

So TLDR wildcard and pattern (regexp) match is different, and we are expecting wildcard pattern from users, not sure if we wanna do an intermediate transformation to check with patterns (regexp in the end), or stick to wildcards end2end.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -195,6 +197,22 @@ protected DevToolsInterface getDevToolsInterface(
 
     @Override
     public void onConnect(AtmosphereResource resource) {
+        if (IndexHtmlRequestHandler.RANDOM_DEV_TOOLS_TOKEN
+                .equals(resource.getRequest().getParameter("token"))) {
+            handleConnect(resource);
+        } else {
+            getLogger().debug(
```
mcollovati, line 204: Could this message be at least at `WARN` level?

Artur-: Yes, makes sense now that the client does not blindly try to connect. Changed

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
Artur-: That would require work, implement the matching in an incorrect way (all special regexp chars would have special meaning) and add more code for us to mantain and test. Let's not

flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
```diff
@@ -100,6 +100,13 @@ public class InitParameters implements Serializable {
      */
     public static final String SERVLET_PARAMETER_DEVMODE_OPTIMIZE_BUNDLE = "devmode.optimizeBundle";
 
+    /**
+     * Whitelisting to specify the allowed hosts, potentially with wildcards,
+     * which can connect to the dev tools from those IP addresses. Otherwise,
+     * the non-localhost connections will be rejected.
+     */
+    public static final String HOSTS_ALLOWED = "devmode.hostsAllowed";
```
mcollovati, line 108: I would maybe rename it to `DEVTOOLS_HOST_ALLOWED`, as `HOSTS_ALLOWED` seems too generic

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -358,6 +367,59 @@ private void addDevTools(Document indexDocument,
...
+        } catch (Exception e) {
+            getLogger().debug(
+                    "Unable to resolve remote address: '{}', so we are preventing the web socket connection",
+                    remoteAddress, e);
+            return false;
+        }
+
+        if (hostsAllowed != null) {
+            // Allowed hosts set
+            return FilenameUtils.wildcardMatch(remoteAddress, hostsAllowed);
```
mcollovati, line 417: Match easier than with regexp :+1: 
nit: I wonder if we should consider allowing `hostsAllowed` to provide multiple comma separated patterns

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
czp13: Yes, agree, let's use wildcards matching from start to end, do not make it more complex.

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
czp13: Good idea, but I am not sure. 

And then I guess we need this code:
```suggestion
    String[] allowedHosts = hostsAllowed.split(",");

    for (String allowedHost : allowedHosts) {
        if (FilenameUtils.wildcardMatch(remoteAddress, allowedHost.trim())) {
            return true;
        }
    }
```

And then adding some covering tests... @Artur- I need you once more :D :( Do we wanna support that or wildcard shall be enough?

flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
czp13: Thanks, good point. Fixed this one to `SERVLET_PARAMETER_DEVMODE_HOSTS_ALLOWED` to keep the naming convention as we discussed as well!

Artur-:
Added support for many hosts to support the non-local proxy scenario

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
mcollovati: An example from a setup with a reverse-proxy (running on docker) in front of an app, connection from a local and a remote machine

Accessing from local
```
browser URL: http://localhost:9090
remote addr: 172.17.0.3 (IP from docker network)
forwarded-for: 172.17.0.1 (IP from docker network)
```

Accessing from another laptop on the same LAN
```
browser URL: http://192.168.1.99:9090
remote addr: 172.17.0.3 (IP from docker network)
forwarded-for: 192.168.1.23 
```

To make it work, I would need to set the allowed list to something like `172.17.0.*,192.168.1.23`

Artur-:
Are we done here? 
image

mcollovati (dismissed):

flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
```diff
@@ -100,6 +100,13 @@ public class InitParameters implements Serializable {
      */
     public static final String SERVLET_PARAMETER_DEVMODE_OPTIMIZE_BUNDLE = "devmode.optimizeBundle";
 
+    /**
+     * Whitelisting to specify the allowed hosts, potentially with wildcards,
```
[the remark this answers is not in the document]
czp13: Thanks fixed as we discussed same time as you did write :)

mcollovati (approved):

PR #18081, opened 2023-11-21
feat!: Implement HasPlaceholder

### LINKED ISSUE #4068: Create HasPlaceholder interface
Similar to `HasEnabled` in https://github.com/vaadin/flow/issues/3816

mvysny:
## Description

Implement the HasPlaceholder interface.

Fixes #4068 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


mvysny:
Warning: this PR brings in an incompatible change since it changes the result type of `Input.getPlaceholder()` from `Optional<String>` to `String`. I'd argue that this is OK since:
1. It unifies the return type for Input and for all other components, e.g. TextField
2. placeholder is now again a bean property
3. Returning null will not overwhelmingly cause errors since `getPlaceholder()` is used sparingly, if at all.

mcollovati (commented):
Overall LGTM, but I have a couple of questions:

* do we still need to use the property descriptor in the `Input` element?
* shouldn't also `RangeInput` implement the new interface?
* should we also update the `InputTextElement` in `flow-html-components-testbench` to make it implement `com.vaadin.testbench.HasPlaceholder`

mcollovati:
BTW, we should probably create an issue for implementing the testbench `RangeInputElement`

mvysny:
> do we still need to use the property descriptor in the Input element?

Unfortunately yes, because Input element sets the placeholder *attribute*, while all other web components and all `Has*` interfaces set the *property*. Therefore, we need to override the default implementation in `Input`.

> shouldn't also RangeInput implement the new interface?

Nope, `<input type="range">` doesn't support the `placeholder` attribute as per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range

> should we also update the InputTextElement in flow-html-components-testbench to make it implement com.vaadin.testbench.HasPlaceholder

Yup, but I can't figure out how to run IT tests in flow-html-components-testbench unfortunately.

> BTW, we should probably create an issue for implementing the testbench RangeInputElement

Agree: https://github.com/vaadin/flow/issues/18093

mcollovati (approved):

mcollovati (approved):

Issue #18082, opened 2023-11-21
Unauthenticated Access behavior changes in between V24.3 and V24.2


ZheSun88:
### Description of the bug

with Vaadin 24.3.0.alpha5 (flow 24.3.0.alpha6), run test on bakery got one failure.

![image](https://github.com/vaadin/flow/assets/31067185/91a81e8e-12d2-48aa-b355-984055991f19)

### Expected behavior

the page returns a 404 in Vaadin 24.2.3
![image](https://github.com/vaadin/flow/assets/31067185/0bab8853-851a-4827-befc-6d491ed67240)

### Minimal reproducible example

**version 24.2.3**
git clone git@github.com:vaadin/bakery-app-starter-flow-spring.git
mvn spring-boot:run 
login with `barista` 
modify the url to `http://localhost:8080/users`
you will get the view like in the expected behavior
===========================================
update the vaadin.version to 24.3.0.alpha5 and repeat the previous step
you will get the view like in the description

### Versions

- Vaadin / Flow version: 24.3.0.alpha5/24.3.0.alpha6

tltv:
It's showing the default access denied page which looks same as default `RouteNotFoundError` page. Bakery needs a custom error page for the new `AccessDeniedException` too.  It can be basically a copy of `CustomRouteNotFoundError` that extends `RouteAccessDeniedError` and overrides `setErrorParameter(BeforeEnterEvent event, ErrorParameter<AccessDeniedException> parameter)`.

ZheSun88:
okay, so this is not a bug, but a breaking change in V24.3. 

we should add some comments to the release note.

tltv:
If we can reroute to Not found error page from the default Access denied page then this would not be a braking change.

PR #18084, opened 2023-11-21
fix: reroute to NotFoundException from default RouteAccessDeniedError

tltv:
This makes `AccessDeniedException`'s default error page to work by default just like in the previous versions before `AccessDeniedException` even when `NotFoundException` is customized.

Fixes: #18082


tltv:
There's another issue with route path access checker that need to be fixed first before this can work.

mcollovati (approved):

PR #18086, opened 2023-11-21
fix: allow access to error view if not denied by access checker

mcollovati:
During a re-route to an error handler component, navigation access checkers may abstain from taking a decision because HasErrorParameter classes are usually not Flow routes with a path, but simple Flow components. For those error handling view access is allowed, if none of the checkers explicitly denies it.

This situation will never happen if AnnotatedViewAccessChecker is enabled, since it always express a decision based on the security annotation present on the error handler component.


mcollovati:
Test has been refactored to avoid using methods from deprecated class

tltv (approved):

mshabarov (approved):

PR #18097, opened 2023-11-22
chore: bump gradle wrapper to 8.4

manolo:


ZheSun88 (approved):

tltv:
Pushed in a fix to `VaadinUtils.kt` which broke with the new `Provider#filter(Predicate)` function that came with Gradle 8.4. If tests pass now this should be good to go.

PR #18100, opened 2023-11-22
fix: RangeInput orientation: add support for Chrome, Edge and Safari.…

### LINKED ISSUE #18096: RangeInput: vertical orientation only works on Firefox
### Description of the bug

According to https://caniuse.com/?search=vertical%20slider, we have to add additional styles, to make the slider truly vertical on all browsers:

1. In Firefox, `orient="vertical"` is enough
2. In Chrome+Safari, you need to add `-webkit-appearance: slider-vertical`. Maybe also `appearance: slider-vertical` as mentioned here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#creating_vertical_range_controls
3. In Edge, `writing-mode: bt-lr` is necessary.

### Expected behavior

`RangeInput.setOrientation()` should also set `-webkit-appearance` and `writing-mode` when in vertical mode.

```java
@Route("")

    public MainView() {
        final Input input = new Input();
        input.setType("range");
        input.getElement().setAttribute("orient", "vertical");
        add(input);
    }
}
```

In Firefox 120, the RangeInput is vertical, but in Chromium 119, the RangeInput is horizontal

image

Both `-webkit-appearance: slider-vertical` and also `appearance: slider-vertical` fix the issue. Need to test on Edge.

### Versions

- Vaadin / Flow version: master
- Java version: 17
- OS version:
- Browser version (if applicable): Firefox 120, Chromium 119, Edge ???

mvysny:
… Fixes #18096

## Description

According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#creating_vertical_range_controls , additional styles are required to make the Range Input vertical on Edge, Safari and Chrome. This PR adds all necessary CSS styles.

Fixes #18096

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
@@ -219,7 +219,8 @@ public class RangeInput extends AbstractSinglePropertyField<RangeInput, Double>
      * "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#non-standard_attributes">Non-standard
      * Attribute</a>. Since the vertical orientation is not standardized yet,
      * this feature is not guaranteed to work on every browser. We found this
-     * feature to work on Firefox 120+, Chromium 119+ and Edge 119+.
+     * feature to work on Firefox 120+, Chromium 119+, Edge 119+ and Safari
+     * 17.1+.
      * <p>
      * </p>
      * The orient attribute defines the orientation of the range slider. Values
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
@@ -27,6 +27,7 @@ import com.vaadin.flow.data.value.HasValueChangeMode;
 import com.vaadin.flow.data.value.ValueChangeMode;
 
 import java.util.Arrays;
+import java.util.Objects;
 
 /**
  * Creates a new input element with type "range".
@@ -216,7 +217,9 @@ public class RangeInput extends AbstractSinglePropertyField<RangeInput, Double>
      * </p>
      * <a href=
      * "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#non-standard_attributes">Non-standard
-     * Attribute</a>.
+     * Attribute</a>. Since the vertical orientation is not standardized yet,
+     * this feature is not guaranteed to work on every browser. We found this
+     * feature to work on Firefox 120+, Chromium 119+ and Edge 119+.
      * <p>
      * </p>
      * The orient attribute defines the orientation of the range slider. Values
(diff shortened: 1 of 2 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -228,7 +231,25 @@ public void setStep(Double step) {
...
     public void setOrientation(Orientation orientation) {
+        Objects.requireNonNull(orientation);
+        // Fix support for individual browsers
+        // See
+        // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#creating_vertical_range_controls
+        // for more details
+
+        // support for Firefox
         set(orientDescriptor, orientation.getValue());
+        if (orientation == Orientation.VERTICAL) {
```
knoobie, line 242: Personally, I hate this excessive usage of getStyle and the resulting inline styles, increased networt traffic and mixing of java and css. Those styles should be part of the theme / in a css file / module instead of hardcoded here.

mvysny: I agree. Unfortunately, I'm not sure whether that's possible, since RangeInput is just plain html `<input>` element, and is not even a web component, so there probably is no place to put the default styling to.

The vertical orientation is not used that often, so hopefully this added traffic won't affect our apps much.

mcollovati: I understand knoobie's point of view, but as Martin said, it actually not feasible for native component wrappers.
The other option would be to remove orientation support at all.
I propose to proceed with the solution in this PR

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
```diff
@@ -216,7 +217,9 @@ public void setStep(Double step) {
      * </p>
      * <a href=
      * "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#non-standard_attributes">Non-standard
-     * Attribute</a>.
+     * Attribute</a>. Since the vertical orientation is not standardized yet,
+     * this feature is not guaranteed to work on every browser. We found this
+     * feature to work on Firefox 120+, Chromium 119+ and Edge 119+.
```
mcollovati, line 222: We should give it a try with Safari and see if we can add it to the list

mvysny: I tested with Safari 17.1 (on iPhone and iPad only) and it worked. I'll add it to the list.

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
mvysny: Hmm, creating a css file in `/src/main/resources/META-INF/resources/frontend/my-styles/styles.css` as mentioned in the [loading resources](https://vaadin.com/docs/latest/advanced/loading-resources) documentation, table 3 add-ons? That could work!

Actually it may not work. The problems would start when someone would insert RangeInput into a ShadowDOM of some other component, then the CSS wouldn't apply unfortunately I think... Or would it? I'm lost in these details :thinking: 

I'm reading https://vaadin.com/blog/simplified-theming-in-vaadin-24 and getting more confused :-D Maybe the `::part` thingy would help; but `<input>` doesn't have any parts since it's not a web component... and at this point it feels like we're going to kill a fly with a hammer :)

I propose to go with the current solution, however inefficient, and to open a feature request to perform a refactoring in the way @knoobie proposes.

mcollovati: It is technically possible, but it may impact an application in unwanted ways, as it will be a global style.

mvysny: I agree: current solution is more self-contained.

mcollovati (approved):

flow-html-components/src/main/java/com/vaadin/flow/component/html/RangeInput.java
knoobie: Don't waste your time to create such a request :) hopefully the whole code can be removed in the near future when browser vendors get their sh*t together :)

PR #18105, opened 2023-11-22
fix: Redirect to error view by default for errors in RPC-handling

mshabarov:
Based on DX feedback, the default for https://github.com/vaadin/flow/pull/17791 is swapped to true.

### DIFF
--- a/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
+++ b/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
@@ -433,7 +433,7 @@ public class BuildDevBundleMojo extends AbstractMojo
 
     @Override
     public boolean isErrorHandlerRedirectEnabled() {
-        return false;
+        return true;
     }
 
 }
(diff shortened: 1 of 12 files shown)


flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
```diff
@@ -433,7 +433,7 @@ public boolean isPrepareFrontendCacheDisabled() {
 
     @Override
     public boolean isErrorHandlerRedirectEnabled() {
-        return false;
+        return true;
```
knoobie, line 436: Thinking out loud: if the default is swapped and the flag was only added to allow slow adoption / testing of the feature without breaking customer projects.. wouldn't it make sense to fully remove the configuration? In the end, the feature should only activate if no custom ErrorHandler is specified if I remember correctly, therefore for anyone option out would be as easy as creating s custom ErrorHandler and opt-in the vice-versa, either using the provided utility in their own Impl, or remove their impl in favor of the default

mshabarov: I almost agree to remove the config parameter. If this feature isn't needed, one can override the default error handler with `shouldHandle` returning false. Configuration would also benefit if we remove it. I've asked internally about other opinions, let us wait a bit and then, if nothing pops up, I'll remove it.

knoobie: Fight for simplicity, even in your own code base 😇

tltv (approved):

PR #18110, opened 2023-11-23
refactor: refactor navigation access control API

mcollovati:
Extracts NavigationAccessChecker inner classes to top level
Adds factory method to AccessCheckResult
Improves javadoc

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessCheckDecisionResolver.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessCheckDecisionResolver.java
@@ -64,9 +64,9 @@ public interface AccessCheckDecisionResolver extends Serializable {
      *     int diff = allow - deny;
      *
      *     if (allow == 0 && deny == 0) {
-     *         return AccessCheckResult.NEUTRAL;
+     *         return AccessCheckResult.neutral();
      *     } else if (diff > 0) {
-     *         return AccessCheckResult.ALLOW;
+     *         return AccessCheckResult.allow();
      *     } else if (diff < 0) {
      *         return AccessCheckResult.deny(String
      *                 .format("Allow %d - Deny %d. Deny wins!", allow, deny));
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessCheckDecisionResolver.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessCheckDecisionResolver.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2000-2023 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.vaadin.flow.server.auth;
+
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * Indicates a class that is responsible for taking a decisions about granting
+ * access to a target view, based on the result provided by
+ * {@link NavigationAccessChecker}s.
+ * <p>
+ * The component is used by {@link NavigationAccessControl} to compute the final
+ * decision, based on the results of all registered
+ * {@link NavigationAccessChecker}s.
+ */
+@FunctionalInterface
+public interface AccessCheckDecisionResolver extends Serializable {
+
+    /**
+     * Determines if access is granted for a specific navigation context, based
+     * on the decisions provided by {@link NavigationAccessChecker}s.
+     * <p>
+     * </p>
+     * The decision resolver should grant access or deny it by returning an
+     * appropriate {@link AccessCheckResult} object.
+     * <p>
+     * </p>
+     * The expected result of the method should be
+     * {@link AccessCheckDecision#ALLOW} or {@link AccessCheckDecision#DENY}, or
+     * {@link AccessCheckDecision#REJECT}.
+     * <p>
+     * </p>
+     * A {@link AccessCheckDecision#NEUTRAL} result does not make because it
+     * does not provide meaningful information to
+     * {@link NavigationAccessControl} to complete the access check process. For
+     * this reason, a neutral result will produce the same effect as
+     * {@link AccessCheckDecision#REJECT}, preventing the navigation and failing
+     * with an exception in development mode.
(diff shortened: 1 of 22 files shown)


flow-server/src/main/java/com/vaadin/flow/server/auth/AccessCheckDecisionResolver.java
```diff
@@ -0,0 +1,94 @@
...
+ * Indicates a class that is responsible for taking a decisions about granting
+ * access to a target view, based on the result provided by
+ * {@link NavigationAccessChecker}s.
+ * <p>
+ * The component is used by {@link NavigationAccessControl} to compute the final
+ * decision, based on the results of all registered
+ * {@link NavigationAccessChecker}s.
+ */
+@FunctionalInterface
+public interface AccessCheckDecisionResolver extends Serializable {
```
knoobie, line 32: Should this extend SerializableBiFunction to "enforce" the idea of functional interface?

mcollovati: In this case, I would prefer to avoid such a strong coupling with a platform interface because it could potentially cause more difficulties if we need/want to change the API in the future.

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessCheckResult.java
```diff
@@ -0,0 +1,128 @@
...
+     * A result instance informing that the navigation to the target view is
+     * allowed for the current user.
+     */
+    public static final AccessCheckResult ALLOW = new AccessCheckResult(
+            AccessCheckDecision.ALLOW, null);
+    /**
+     * A result instance informing that the checker cannot take a decision based
+     * on the given navigation information.
+     */
+    public static final AccessCheckResult NEUTRAL = new AccessCheckResult(
```
mshabarov, line 38: Now we have two constants and two factory methods for different results.
Maybe it's better to have only four factory methods and remove constants, so the code would look like

```java
if (hasAccess(context)) {
     return AccessCheckResult.allow();
} else {
     return AccessCheckResult.deny("Access denied");
}
```

mshabarov: Mixing constants and methods looks like incomplete API IMO.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessCheckResult.java
mcollovati: done

knoobie: Would it make sense to have a method "create()" that takes all params? So the developer isn't forced to create multiple if-else constructs where they either call accept() or reject(), but can just do create(Result, String) in the end of a method after they have checked anything and decided what the outcome is

knoobie: Another point: the AccessCheckResult is final, making it impossible to have custom implementations with for example more attributes forcing the developer to put all information in the single reason field (technical, human readable messages and so on) 

We often use wrapper classes that hold a message to be printed in the log for DevOps and an readable message for the customer to be shown on the UI as an example. 

Suggestion: AccessCheckResult could be an interface, with a SimpleAccessCheckResult like the infamous Result to allow easier extensibility for the users

(Sorry, the idea just came up.. the other PR was to large for my phone and brain 😬)

mshabarov: This `create` method is a shortcut for `new AccessCheckResult(AccessCheckDecision decision, String reason)`, right?

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/server/auth/AccessCheckResult.java
mshabarov: > Suggestion: AccessCheckResult could be an interface, with a SimpleAccessCheckResult like the infamous Result to allow easier extensibility for the users

If you think this can be useful, I believe we change to that.

knoobie: I was thinking about making this even private and have consistent static builder methods only, like your other shortcuts to have the same API for all types

mcollovati: I'm not sure about making `AccessCheckResult` as this would require additional checks on our code and would somehow block or make it more difficult for us to refactor that class, since we should consider the existence of potential custom implementations.

But we can reason about that if there are good uses cases.

@knoobie, if you could extend or have a custom `AccessCheckResult` with additional information, where would you use it?
I would say that to use them, you will have to also extend other navigation access control components.
So I wonder if it is worth it to make `AccessCheckResult` extensible, or if we can simply improve the existing instance with some additional info.

knoobie: > Interface change 
I think it's really useful and decreases the burden on you, because it allows for full flexibility for the Developer by allowing sub-classing and using instance of, if the returning result is a standard return type or some of they special type. 

Another spontaneous idea: the developer could want to return a view to delegate to once the access is denied, with a interface and custom class - the developer could just add the redirect route as another param and check if the result is of type "HasRedirect" and doesn't have to do hacks to pass the route throw Session Attributes or other means

knoobie: > So I wonder if it is worth it to make AccessCheckResult extensible, or if we can simply improve the existing instance with some additional info.

That would also work, e.g. add a map like spring did for they RFC-7807 impl to allow for additional information https://github.com/spring-projects/spring-framework/blob/fb4455b396dd8648f37917a74ab08e9ad373eef7/spring-web/src/main/java/org/springframework/http/ProblemDetail.java#L71

mcollovati: I can see a potential usage for the custom result in a custom decision resolver.
For example, having multiple checkers, based on the result type, you may want to ignore the output or maybe assign different weights to the decision and compute the final result based on that.

Adding a Map of properties could be a good idea, and also easy to implement. Potentially, we could also print the values on debug logs.

Otherwise, we consider how much refactoring the `AccessCheckResult` as an interface impacts current code. For this, I think we will need to introduce several and null consistency checks (e.g. mandatory message for some decisions type) in different places to give better advice than an NPE to the developers who incorrectly implement the interface.

mshabarov: Within this PR I propose to do: add "create" method and remove `final` from `AccessCheckResult` class, so that it can be extendable.

This PR pursues refactoring purposes and I think we can discuss further how we can improve `AccessCheckResult` API. Would be good to make a ticket for it.

mshabarov (approved):

PR #18127, opened 2023-11-24
Populate UI.getCurrent() before running ErrorHandler

### LINKED ISSUE #10533: UI.getCurrent() is null in ErrorHandler
Throwing an exception in the `ui.access()` block from bg thread will call `ErrorHandler.error()` with `UI.getCurrent()` set to null. That will fail to show any notification from the `ErrorHandler` with the following stacktrace:

```
[Thread-8] ERROR com.vaadin.flow.component.UI - UI instance is not available. It means that you are calling this method out of a normal workflow where it's always implicitly set. That may happen if you call the method from the custom thread without 'UI::access' or from tests without proper initialization.
java.lang.IllegalStateException: UI instance is not available. It means that you are calling this method out of a normal workflow where it's always implicitly set. That may happen if you call the method from the custom thread without 'UI::access' or from tests without proper initialization.
	at com.vaadin.flow.component.notification.Notification.setOpened(Notification.java:426)
	at com.vaadin.flow.component.notification.Notification.open(Notification.java:303)
	at com.vaadin.flow.component.notification.Notification.show(Notification.java:227)
	at com.vaadin.flow.component.notification.Notification.show(Notification.java:244)
	at org.vaadin.example.Bootstrap.error(Bootstrap.java:20)
	at com.vaadin.flow.component.UI$1.handleError(UI.java:520)
	at com.vaadin.flow.server.FutureAccess.handleError(FutureAccess.java:76)
	at com.vaadin.flow.server.VaadinService.runPendingAccessTasks(VaadinService.java:2045)
	at com.vaadin.flow.server.VaadinSession.unlock(VaadinSession.java:681)
	at com.vaadin.flow.server.VaadinService.ensureAccessQueuePurged(VaadinService.java:2005)
	at com.vaadin.flow.server.VaadinService.accessSession(VaadinService.java:1972)
	at com.vaadin.flow.server.VaadinSession.access(VaadinSession.java:969)
	at com.vaadin.flow.component.UI.access(UI.java:506)
	at com.vaadin.flow.component.UI.access(UI.java:489)
	at org.vaadin.example.MainView$1.run(MainView.java:52)
```

The `ErrorHandler` does not document whether it's called in Vaadin UI thread or not. If it doesn't, this needs to be clearly documented. If yes, then the Vaadin environment should be set properly - the UI.getCurrent() should not return null. In either case, this needs to be documented in the `ErrorHandler` javadoc.

Attaching the skeleton starter app which demoes the issue. In short, this view demoes the issue:
```

    public MainView() {
        // Use TextField for standard text input
        TextField textField = new TextField("Wait for it");
        add(textField);

        final UI ui = UI.getCurrent();
        new Thread() {
            @Override
            public void run() {
                try {
                    Thread.sleep(1000);
                    ui.access(() -> {
                       throw new RuntimeException("Simulated failure");
                    });
                } catch (Throwable t) {
                    t.printStackTrace();
                }
            }
        }.start();
    }
}
```
Make sure to have the ErrorHandler set; wait for 1 second then check out the console log.

[skeleton-starter-flow.zip](https://github.com/vaadin/flow/files/6237154/skeleton-starter-flow.zip)

### Expected behavior
The UI.getCurrent() should not be null.
### Actual behavior
The UI.getCurrent() is null.
### Versions:
    - Vaadin / Flow version: 14.5.1 / 2.5.1
    - Java version: 11

mvysny:
## Description

Populates UI.getCurrent() correctly before the ErrorHandler is called.

Fixes #10533 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -571,8 +571,13 @@ public void handleError(Exception exception) {
                         ErrorHandlingCommand errorHandlingCommand = (ErrorHandlingCommand) command;
                         errorHandlingCommand.handleError(exception);
                     } else if (getSession() != null) {
-                        getSession().getErrorHandler()
-                                .error(new ErrorEvent(exception));
+                        UI.setCurrent(UI.this);
+                        try {
+                            getSession().getErrorHandler()
+                                    .error(new ErrorEvent(exception));
```
knoobie, line 577: Would it make sense to also pass the UI to the event making it up aside for the Developer to access it

mvysny: I'm not sure I understand correctly - would you like the `ErrorEvent` to also have the `ui` field?

knoobie: Yeah, like in all other events 😅

knoobie: Don't get me wrong, it's also needed within the setCurrent.. looking at you Notification.show-implementation.. would just be convenient to access it from the event directly

mvysny: Makes sense... but easy for you to say, ErrorHandler is constructed in surprisingly lot of places :-D Also, not sure whether I should add a new constructor parameter for this, or simply add `ErrorHandler.ui` field which would be initialized directly via `ui = UI.getCurrent()`...

Actually, in the the `Notification.show()` case, Notification is getting the UI via `UI.getCurrent()`, but that's a nitpick.

knoobie: Well.. Let's ditch it then 😅 the most important part is the getCurrent to finally make Notification happy.. that was annoying as hell to get workarounds for that over the years

mvysny: Yes, and Yes! :-D I agree to ditch the ui field for now. Don't get me wrong - I'd love to have it; but ErrorHandler looks to be a bit of a can of worms, being called from 20 different places without  clear understanding whether the session lock is held or not... It would probably be good to have VaadinSession.callErrorHandler() which makes sure that the Session lock is held and the Current UI is set (if known/available), but that's a bit of a bigger refactoring...

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -571,8 +571,13 @@ public void handleError(Exception exception) {
                         ErrorHandlingCommand errorHandlingCommand = (ErrorHandlingCommand) command;
                         errorHandlingCommand.handleError(exception);
                     } else if (getSession() != null) {
-                        getSession().getErrorHandler()
-                                .error(new ErrorEvent(exception));
+                        UI.setCurrent(UI.this);
```
Legioth, line 574: Is there a remote possibility that another current UI is already set? It's in general a good practice to always preserve the old value and restore it later on when doing temporary assignments in framework code.

To do that, you can use `CurrentInstance.set(UI.class, UI.this);` which returns the previous value that can thus be used for restoration in `finally`.

mvysny: Great tip, thanks! Refactored.

tepi (approved):

PR #18129, opened 2023-11-24
feature: Implement multiple strategies for the `VaadinSession.checkHasLog()` call

### LINKED ISSUE #15776: Configuration for checkHasLock() behavior strategy
### Describe your motivation

Vaadin 14.9.5

We are facing an unfamiliar code base which may have cases of threads calling UI code without wrapping the call in `ui.access()`. Performing static code analysis for this kind of errors is quite hard. It would be good if I could temporarily change the strategy used by `checkHasLock()` in production, to identify and gather statistics on such incorrect usages.

At the moment, `VaadinSession.checkHasLock()` does nothing in production since assertions are disabled in production. The only alternative is to run Vaadin in development mode, however that is not ideal either: we'll get hard exceptions and things may stop to work because of that.

A perfect solution would be to have a configuration option for Vaadin on what should be done by the `checkHasLock()` function:

1. Always assert
2. Always throw IllegalStateException
3. Always log an error with a stacktrace but don't throw an exception.

It would be great to have the third option available; minor performance degradation because of frequent `hasLock()` calls are acceptable.

There should be new configuration option which should not be tied to whether Vaadin's running in production mode. It could be an enum value; if null (the default), the behavior would be as it is now: assert in production, throw in development. That would keep backward compatibility.

I would then set this new configuration option to LOG for a week; I could then analyze the log files and immediately spot incorrect API usage.

mvysny:
## Description

Implement multiple strategies for the `VaadinSession.checkHasLog()` call. The programmer is now able to reconfigure the lock checking strategy, for example to log any checkHasLock() failures into log in production mode, to figure out any problematic calls to Vaadin UI components from background thread.

Fixes #15776

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


OlliTietavainenVaadin (approved):
👍

flow-server/src/main/java/com/vaadin/flow/server/DefaultDeploymentConfiguration.java
```diff
@@ -55,6 +55,10 @@ public class DefaultDeploymentConfiguration
             + "The permitted values are \"disabled\", \"manual\",\n"
             + "and \"automatic\". The default of \"disabled\" will be used.";
 
+    public static final String WARNING_LOCK_CHECK_STRATEGY_NOT_RECOGNIZED = "WARNING: lockCheckStrategy has been set to an unrecognized value.\n"
+            + "The permitted values are \"assert\", \"log\",\n"
+            + "and \"throw\". The default of \"assert\" will be used.";
```
mcollovati, line 60: I could be good to use `LockCheckStrategy.DEFAULT.name().toLowercase()` here instead of hard-coding `assert`.
Potentially, the message could be built using `LockCheckStrategy.values()` to avoid misleading messages if something changes in the enum.

flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
```diff
@@ -55,6 +55,7 @@ public class InitParameters implements Serializable {
     public static final String SERVLET_PARAMETER_WEB_COMPONENT_DISCONNECT = "webComponentDisconnect";
     public static final String SERVLET_PARAMETER_CLOSE_IDLE_SESSIONS = "closeIdleSessions";
     public static final String SERVLET_PARAMETER_PUSH_MODE = "pushMode";
+    public static final String SERVLET_PARAMETER_LOCK_CHECK_STRATEGY = "lockCheckStrategy";
```
mcollovati, line 58: Maybe better specify `session` here; `lockCheckStrategy` seems too vague to me.
What about `sessionLockCheckStrategy`?

flow-server/src/main/java/com/vaadin/flow/server/LockCheckStrategy.java
```diff
@@ -0,0 +1,55 @@
...
+    THROW {
+        @Override
+        public void checkHasLock(VaadinSession session, String message) {
+            if (!session.hasLock()) {
+                throw new IllegalStateException(message);
+            }
+        }
+    };
+
+    public static final LockCheckStrategy DEFAULT = ASSERT;
```
mcollovati, line 42: Javadoc missing in public field

mvysny:
Awesome, thanks @OlliTietavainenVaadin !

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -556,8 +569,10 @@ public boolean hasLock() {
      * Potentially checks whether this session is currently locked by the
      * current thread, and fails with the given message if not.
      * <p>
-     * When production mode is enabled, the check is only done if assertions are
-     * also enabled. This is done to avoid the small performance impact of
+     * When production mode is enabled, the check is done according to the
+     * {@link InitParameters#SERVLET_PARAMETER_LOCK_CHECK_STRATEGY lock check
+     * strategy}. By default, the check is only done if assertions are also
+     * enabled: this is done to avoid the small performance impact of
```
mcollovati, line 575: I would mention that in development mode this always throws an exception (if I got it right)

mcollovati:
Ok, it seems like I was late to the party

PR #18139, opened 2023-11-27
chore: install also react-plugin-swc

Artur-:
It is faster and more correct.
Did not remove the old plugin as it is used in all existing projects and Flow would uninstall it if removed from here


Artur-:
The reason to switch is to get this resolved https://github.com/vitejs/vite-plugin-react/issues/266

Artur-:
Changed the PR to only add the new plugin

mshabarov (approved):
Adding this dependency looks good to me.
I honestly don't know if we have to include this plugin into Vite configuration by default.

knoobie:
What is the benefit of this plugin in flow? I just tried 24.3-rc.1 and was a little bit confused why this was added to my package.json

mshabarov:
`react-plugin-swc` has been removed in 24.4 in favour of another plugin (see PR link above).

PR #18141, opened 2023-11-27
feat: anchor handling for react-router

caalador:
Add anchor handling so we do
not loose all data for flow navigation.

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -13,9 +13,9 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-import {Flow as _Flow} from "Frontend/generated/jar-resources/Flow.js";
-import {useEffect, useRef,} from "react";
-import {useLocation, useNavigate} from "react-router-dom";
+import { Flow as _Flow } from "Frontend/generated/jar-resources/Flow.js";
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -27,11 +27,130 @@ const router = {
...
+        `vaadin-router-${type}`,
+        {cancelable: type === 'go', detail}
+    ));
+}
+
+// The list of checks is not complete:
+//  - SVG support is missing
+//  - the 'rel' attribute is not considered
+// @ts-ignore
+function vaadinRouterGlobalClickHandler(event) {
```
mshabarov, line 57: This function and `getAnchorOrigin` are borrowed from here https://github.com/vaadin/router/blob/v1.7.5/src/triggers/click.js, I propose to reference them here in the comments, as this might be helpful for next developer.

I also would propose to remove "fixme" and "istanbul ignore next" comments as this codes goes to customer projects.

caalador: Changed.
The Flow.tsx is not something that is meant for copying or modifying so basically it is a as internal and closed as click.js

mshabarov (approved):
Tested with Hilla+React app having two Flow routes, now the navigation between Flow routes doesn't reload the page 👍

PR #18175, opened 2023-12-01
chore: improve session lock strategy

mvysny:
## Description

Fixes for https://github.com/vaadin/flow/pull/18129 that was unfortunately merged too hastily.

Couple of changes:

* The LockCheckStrategy.DEFAULT was misleading since there are two defaults, one for production mode (assert), the other for development mode (throw).
* Renamed LockCheckStrategy to SessionLockCheckStrategy, to explicitly scope the lock to VaadinSession
* The init parameter got renamed to "productionSessionLockCheckStrategy", to explicitly mention that this is meant for a session lock which is only taken into account in production mode. However, this is a bit mouthful, perhaps sessionLockCheckStrategy would suffice? @mcollovati

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/SessionLockCheckStrategy.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/SessionLockCheckStrategy.java
@@ -1,9 +1,9 @@
 package com.vaadin.flow.server;
 
 /**
- * The strategy for session lock checking in production mode.
+ * Available strategies for session lock checking.
  */
-public enum LockCheckStrategy {
+public enum SessionLockCheckStrategy {
     /**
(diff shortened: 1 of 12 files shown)


mcollovati:
>  However, this is a bit mouthful, perhaps sessionLockCheckStrategy would suffice?

`sessionLockCheckStrategy ` looks good to me.
We can make explicit in the [documentation](https://vaadin.com/docs/latest/configuration/properties#properties) that this parameter affects only production mode

flow-server/src/main/java/com/vaadin/flow/server/SessionLockCheckStrategy.java
mcollovati, line 1: `SessionLockCheckStrategy` sounds better :+1:

mcollovati (approved):

PR #18265, opened 2023-12-08
fix: Use Atmosphere for debug window connection

Artur-:
This should add support for large messages, reconnect and all other features that Atmosphere comes with

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
@@ -49,6 +49,7 @@ import com.vaadin.flow.server.AbstractConfiguration;
 import com.vaadin.flow.server.AppShellRegistry;
 import com.vaadin.flow.server.BootstrapHandler;
 import com.vaadin.flow.server.Constants;
+import com.vaadin.flow.server.HandlerHelper;
 import com.vaadin.flow.server.InitParameters;
 import com.vaadin.flow.server.Mode;
 import com.vaadin.flow.server.VaadinContext;
@@ -60,6 +61,7 @@ import com.vaadin.flow.server.VaadinSession;
 import com.vaadin.flow.server.frontend.FrontendUtils;
 import com.vaadin.flow.server.frontend.ThemeUtils;
 import com.vaadin.flow.server.startup.ApplicationConfiguration;
+import com.vaadin.flow.shared.ApplicationConstants;
 
 import elemental.json.Json;
 import elemental.json.JsonArray;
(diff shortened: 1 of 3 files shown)


vaadin-dev-server/frontend/websocket-connection.ts
```diff
@@ -0,0 +1,111 @@
...
+  constructor(url: string) {
+    super();
+    if (!url) {
+        return;
+    }
+
+    const config = {
+      transport: 'websocket',
+      url,
+      //fallbackTransport: 'long-polling',
```
mshabarov, line 17: Remove this if not needed?

Artur-: Removed

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -365,6 +373,10 @@ private void addDevTools(Document indexDocument,
                     """, devToolsConf.toJson()));
 
         indexDocument.body().appendChild(new Element("vaadin-dev-tools"));
+
+        String pushUrl = BootstrapHandlerHelper.getServiceUrl(request) + "/"
+                + ApplicationConstants.VAADIN_PUSH_DEBUG_JS;
+        addScriptSrc(indexDocument, pushUrl);
```
mshabarov, line 379: Maybe not that important, but `BootstrapHandler` adds also the push script if Vaadin Push is enabled. Would it make sense to add it here only if Vaadin Push isn't enabled?

Artur-: When is BootstrapHandler used?

mshabarov: It's used for embedded web components, including Vaadin Portlet. And `includeInitialUidl` of `IndexHtmlRequestHandler` internally calls for `getPushScript` of `BootstrapHandler` and adds this scrript.

Artur-: Let's ignore portlets because we don't have support for that..

It seems like there is different logic in many places for this..
1. JavscriptBootstrapHandler writes `"pushScript"` if it includes initial UIDL and `Flow.ts` reads this and creates a script tag when initializing Flow. This happens regardless if the script has already been loaded
2. BootstrapHandler adds the push script url as a script tag to the page when push is enabled (and when BootstrapHandler happens to be used?)
3. AtmospherePushConnection in flow-client has `runWhenAtmosphereLoaded` which decides the push JS url in its own way but it checks `window.vaadinPush.atmosphere` before loading

This PR causes `IndexHtmlRequestHandler` to write a script tag, like BootstrapHandler, but so that it is always there when dev tools are used. None of the above consider the case then vaadin-dev-server is used without Flow so modifying any of them won't really help.

Suggestions for what to change?

mshabarov: I tested this patch with starter, where Push was enabled, works as expected with not errors.
Thus, I think having two same imports in dev mode is fine as it works.

I propose to unify this and call 
```java
head.appendChild(
                        createJavaScriptElement(getPushScript(context)));
```
instead. As a bonus, it will add a `v` query parameter with hash to prevent caching after version upgrade.

tepi: Yes, should not matter. The push init is only run once regardless of multiple imports.

mshabarov (approved):

PR #18268, opened 2023-12-09
fix: Support fragmented messages in debug window

Artur-:
Without this, a message larger than 16KB will cause the connection to close instead of the message being delivered

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
@@ -25,6 +25,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import org.atmosphere.cpr.AtmosphereRequest;
 import org.atmosphere.cpr.AtmosphereResource;
 import org.atmosphere.cpr.AtmosphereResource.TRANSPORT;
 import org.atmosphere.cpr.BroadcastFilterAdapter;
@@ -47,7 +48,8 @@ import elemental.json.JsonObject;
  * @author Vaadin Ltd
  * @since 1.0
  */
-public class AtmospherePushConnection implements PushConnection {
+public class AtmospherePushConnection
+        implements PushConnection, FragmentedMessageHolder {
 
     private UI ui;
     private transient State state = State.DISCONNECTED;
@@ -60,7 +62,7 @@ public class AtmospherePushConnection implements PushConnection {
     /**
      * Represents a message that can arrive as multiple fragments.
      */
-    protected static class FragmentedMessage implements Serializable {
+    public static class FragmentedMessage implements Serializable {
         private final StringBuilder message = new StringBuilder();
         private final int messageLength;
 
@@ -233,8 +235,12 @@ public class AtmospherePushConnection implements PushConnection {
      * received message, returns a {@link Reader} yielding the complete message.
      * Otherwise, returns null.
      *
+     * @param request
+     *            The atmosphere request with data
      * @param reader
-     *            A Reader from which to read the (partial) message
+     *            The request body reader
+     * @param holder
+     *            A holder for a previously received partial message
      * @return A Reader yielding a complete message or null if the message is
      *         not yet complete.
      * @throws IOException
@@ -242,22 +248,20 @@ public class AtmospherePushConnection implements PushConnection {
      * @return a Reader yielding the complete message, or {@code null} if the
      *         received message was a partial message
      */
-    protected Reader receiveMessage(Reader reader) throws IOException {
+    protected static Reader receiveMessage(AtmosphereRequest request,
+            Reader reader, FragmentedMessageHolder holder) throws IOException {
 
+        AtmosphereResource resource = request.resource();
         if (resource == null || resource.transport() != TRANSPORT.WEBSOCKET) {
             return reader;
         }
 
-        if (incomingMessage == null) {
-            // No existing partially received message
-            incomingMessage = new FragmentedMessage(reader);
-        }
-
-        if (incomingMessage.append(reader)) {
-            // Message is complete
-            Reader completeReader = incomingMessage.getReader();
-            incomingMessage = null;
-            return completeReader;
+        FragmentedMessage msg = holder.getOrCreateFragmentedMessage(request,
+                reader);
+        if (msg.append(reader)) {
+            Reader messageReader = msg.getReader();
+            holder.clearFragmentedMessage(request);
+            return messageReader;
         } else {
             // Only received a partial message
             return null;
@@ -391,6 +395,20 @@ public class AtmospherePushConnection implements PushConnection {
             state = State.DISCONNECTED;
         }
 
+    }
+
+    @Override
+    public FragmentedMessage getOrCreateFragmentedMessage(
+            AtmosphereRequest request, Reader reader) throws IOException {
+        if (incomingMessage == null) {
+            incomingMessage = new FragmentedMessage(reader);
+        }
+        return incomingMessage;
+    }
+
+    @Override
+    public void clearFragmentedMessage(AtmosphereRequest request) {
+        incomingMessage = null;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/FragmentedMessageHolder.java
```diff
@@ -0,0 +1,33 @@
...
+     * Gets the partial message that is currently being received, if any.
+     *
+     * @param request
+     *            the request to get the partial message for
+     * @param reader
+     *            the request body reader
+     * @return the partial message or null
+     */
+    public FragmentedMessage getOrCreateFragmentedMessage(
+            AtmosphereRequest request, Reader reader) throws IOException;
```
mcollovati, line 23: Just looking at the interface, it is not clear to me why this method named `getOrCreate...` should return null.
It would be good to clarify that in the Javadoc.
And maybe use `Optional` as return type, if it is not guaranteed that the method will provide an instance.

flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
```diff
@@ -233,31 +235,33 @@ protected void sendMessage(String message) {
...
+     * @param holder
+     *            A holder for a previously received partial message
      * @return A Reader yielding a complete message or null if the message is
      *         not yet complete.
      * @throws IOException
      *             if an IO error occurred
      * @return a Reader yielding the complete message, or {@code null} if the
      *         received message was a partial message
      */
-    protected Reader receiveMessage(Reader reader) throws IOException {
```
mcollovati, line 245: This could potentially be a breaking change. 
I recall at least a custom `AtmospherePushConnection` in Kubernetes KIT, but I need to check if removing the method may impact it in some way.

flow-server/src/main/java/com/vaadin/flow/server/communication/FragmentedMessageHolder.java
mcollovati, line 1: It looks like the implementation of the two methods uses the `AtmosphereRequest` only to get the `AtmosphereResource`. Could it make sense to directly pass that object instead of the request?

flow-server/src/main/java/com/vaadin/flow/server/communication/AtmospherePushConnection.java
Artur-: It is. It can probably be made compatible also, e.g. when backporting to 24.3 if needed

flow-server/src/main/java/com/vaadin/flow/server/communication/FragmentedMessageHolder.java
Artur-: Seems like that can be done and it makes it simpler, changed

flow-server/src/main/java/com/vaadin/flow/server/communication/FragmentedMessageHolder.java
Artur-: I originally tried to make minimal changes but I now moved message length reading to make this simpler and clearer

flow-server/src/main/java/com/vaadin/flow/server/communication/FragmentedMessageHolder.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.vaadin.flow.server.communication.AtmospherePushConnection.FragmentedMessage;
+
+public interface FragmentedMessageHolder extends Serializable {
+
+    /**
+     * Gets the partial message that is currently being received, if any.
+     *
+     * @param resource
+     *            the resource to get the partial message forder
```
mcollovati, line 15: Probably a typo: `forder`

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -261,6 +263,7 @@ public void onDisconnect(AtmosphereResource resource) {
                     "Push connection {} is not a live-reload connection or already closed",
                     uuid);
         }
+        fragmentedMessages.remove(resource);
```
mcollovati, line 266: Can we move the  fragmentedMessages cleanup to the top of the method?
Just to be sure that the map is purged even if the plugin throws an exception

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -67,7 +69,7 @@ public class DebugWindowConnection implements BrowserLiveReload {
     private VaadinContext context;
 
     private final ConcurrentLinkedQueue<WeakReference<AtmosphereResource>> atmosphereResources = new ConcurrentLinkedQueue<>();
-
+    private final ConcurrentHashMap<AtmosphereResource, FragmentedMessage> fragmentedMessages = new ConcurrentHashMap<>();
```
mcollovati, line 72: Just thinking out loud: could we use `AtmosphereResource.uuid()` as map key, to prevent potential leaks of the `AtmosphereResource`?

mcollovati (changes requested):

mcollovati:
Is there any chance we could add tests for partial messages? Or do we already have something?

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
Artur-: We could do that or we could merge this with the `atmosphereResources` tracking above. Which one would be better?

mcollovati: So, using a synchronized `WeakHashMap` instead of the  `ConcurrentLinkedQueue`?
Sounds good

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
Artur-: Not really relevant anymore as it was merged with the existing structure

Artur-:
> Is there any chance we could add tests for partial messages? Or do we already have something?

Added basic tests for short messages, long messages and messages equal to the limit

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -378,4 +377,22 @@ private static Logger getLogger() {
...
+    @Override
+    public FragmentedMessage getOrCreateFragmentedMessage(
+            AtmosphereResource resource) {
+        return resources.get(getRef(resource));
+    }
+
+    private WeakReference<AtmosphereResource> getRef(
+            AtmosphereResource resource) {
+        return resources.keySet().stream()
+                .filter(resourceRef -> resource.equals(resourceRef.get()))
```
mcollovati, line 389: nit: the check could be `resourceRef.refersTo(resource)`

mcollovati (dismissed):
some minor comments that can be ignored, other than that LGTM

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -378,4 +377,22 @@ private static Logger getLogger() {
...
+                .findFirst().orElse(null);
+    }
+
+    @Override
+    public void clearFragmentedMessage(AtmosphereResource resource) {
+        resources.put(getRef(resource), new FragmentedMessage());
```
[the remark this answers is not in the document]
Artur-: Added checking

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
Artur-: Updated

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -254,7 +254,7 @@ public void onDisconnect(AtmosphereResource resource) {
         for (DevToolsMessageHandler plugin : plugins) {
             plugin.handleDisconnect(getDevToolsInterface(resource));
         }
-        if (!atmosphereResources
+        if (!resources.keySet()
                 .removeIf(resourceRef -> resource.equals(resourceRef.get()))) {
```
mcollovati, line 258: nit: could use `refersTo` also here for consistency

```suggestion
                .removeIf(resourceRef -> resourceRef.refersTo(resource))) {
```

mcollovati (dismissed):

mcollovati (dismissed):

mcollovati (approved):

mcollovati (approved):

PR #18273, opened 2023-12-11
chore: update @vaadin/common-frontend

platosha:
Update to v0.0.19, which supports Lit 3.


mshabarov (commented):
Lit 3.0 is good, but honestly I barely understand the Lit related changes.
Just one concern, is this upgrade somehow a breaking change for Flow or Hilla apps?

Artur- (approved):

Lodin:
@mshabarov, well, it could be, if the specific version of decorators proposal is used (e.g., I have some troubles with dspublisher right now). However, in most cases it shouldn't. And overall, we already have this change across the components/Flow/Hilla codebase, so this is a following update

mshabarov (approved):

PR #18275, opened 2023-12-11
Highlight previous selected component when ESC is pressed

brunovianarezende:
## Description

In Theme Editor, when a user selects a component, starts component picker and then presses ESC, the theme editor will be re-opened and the previous selected component will be highlighted. The behavior is similar to what happens when the user closes the dev tools and then re-open it.

Instead of relying on the highlighting behavior to be triggered indirectly when the ThemeEditor.expand property is updated, I exposed methods that must be called to control when the previously selected component must be highlighted. This makes it easier to understand what is happening.

One thing that was not done in this PR, but might be useful, is to remove the highlighting when the dev tools is expanded, but theme editor is not visible.

## Type of change

- [ X ] Bugfix
- [ ] Feature

## Checklist

- [ X ] The issue is created in the corresponding repository and I have referenced it.

### DIFF
--- a/vaadin-dev-server/frontend/component-picker.ts
+++ b/vaadin-dev-server/frontend/component-picker.ts
@@ -110,9 +110,9 @@ export class ComponentPicker extends LitElement {
     this.dispatchEvent(new CustomEvent('component-picker-opened', {}));
   }
 
-  close() {
+  close(component?: ComponentReference) {
     this.active = false;
-    this.dispatchEvent(new CustomEvent('component-picker-closed', {}));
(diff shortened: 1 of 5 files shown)


MarcinVaadin:
Flow has been moved to Copilot, please check if it's possible to transfer changes there. Currently here https://github.com/vaadin/copilot-internal/tree/main/copilot/frontend/copilot/plugins/copilot-theme-editor

MarcinVaadin:
@brunovianarezende did you check if it's possible to move fix to Copilot?

PR #18285, opened 2023-12-13
fix: Single request navigation

caalador:
Only one request sent to the
server for flow navigation
so that the request contains
removals and adds in the same
query.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -1868,24 +1868,8 @@ public class UI extends Component
         } catch (Exception exception) {
             handleExceptionNavigation(location, exception);
         } finally {
-            FeatureFlags featureFlags;
-            try {
-                featureFlags = FeatureFlags.get(
-                        getInternals().getSession().getService().getContext());
-            } catch (NullPointerException npe) {
-                featureFlags = new FeatureFlags(null) {
-                    @Override
-                    public boolean isEnabled(Feature feature) {
-                        return false;
-                    }
-
-                    @Override
-                    public void loadProperties() {
-                        // NO-OP
-                    }
-                };
-            }
-            if (featureFlags.isEnabled(FeatureFlags.REACT_ROUTER)
+            if (getInternals().getSession().getConfiguration()
+                    .isReactRouterEnabled()
                     && getInternals().getContinueNavigationAction() != null) {
                 getInternals().clearLastHandledNavigation();
             } else {
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -1867,11 +1867,13 @@ public class UI extends Component
         } catch (Exception exception) {
             handleExceptionNavigation(location, exception);
         } finally {
-            if (!FeatureFlags
-                    .get(getInternals().getSession().getService().getContext())
-                    .isEnabled(FeatureFlags.REACT_ROUTER)) {
+            // if (!FeatureFlags
+            // .get(getInternals().getSession().getService().getContext())
+            // .isEnabled(FeatureFlags.REACT_ROUTER)) {
+            if(getInternals().getContinueNavigationAction() != null) {
                 getInternals().clearLastHandledNavigation();
             }
+            // }
         }
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/component/UI.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/UI.java
@@ -30,6 +30,7 @@ import java.util.stream.Stream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.vaadin.experimental.Feature;
 import com.vaadin.experimental.FeatureFlags;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.internal.AllowInert;
@@ -1867,13 +1868,30 @@ public class UI extends Component
         } catch (Exception exception) {
             handleExceptionNavigation(location, exception);
         } finally {
-            // if (!FeatureFlags
-            // .get(getInternals().getSession().getService().getContext())
-            // .isEnabled(FeatureFlags.REACT_ROUTER)) {
-            if(getInternals().getContinueNavigationAction() != null) {
+            FeatureFlags featureFlags;
+            try {
+                featureFlags = FeatureFlags.get(
+                        getInternals().getSession().getService().getContext());
+            } catch (NullPointerException npe) {
+                featureFlags = new FeatureFlags(null) {
+                    @Override
+                    public boolean isEnabled(Feature feature) {
+                        return false;
+                    }
+
+                    @Override
+                    public void loadProperties() {
+                        // NO-OP
+                    }
+                };
+            }
+            if (featureFlags.isEnabled(FeatureFlags.REACT_ROUTER)
+                    && getInternals().getContinueNavigationAction() != null) {
+                getInternals().clearLastHandledNavigation();
+            } else {
                 getInternals().clearLastHandledNavigation();
             }
-            // }
+
         }
     }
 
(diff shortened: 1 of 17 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -15,7 +15,8 @@
  */
 import { Flow as _Flow } from "Frontend/generated/jar-resources/Flow.js";
 import { useEffect, useRef } from "react";
-import { NavigateFunction, useLocation, useNavigate } from "react-router-dom";
+import { matchPath, matchRoutes, NavigateFunction, useLocation, useNavigate } from "react-router-dom";
+import { routes } from "Frontend/routes.js";
```
caalador, line 19: Can we get the routes some other way as this is easily breakable if the developer uses another name or doesn't export the routes.

tarekoraby: Is it enough to through an error (with a meaningful message about what needs fixing) if the developer does either of these things?

caalador: Added check that if it's note exported then Flow throws an exception.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -138,13 +139,41 @@ function navigateEventHandler(event) {
...
+                {
+                    pathname: event.detail.pathname,
+                    search: event.detail.search
+                },
+                {
+                    prevent() {
+                    },
+                    // @ts-ignore
+                    redirect: (path) => {
+                        // window.history.pushState({}, "", path);
```
mshabarov, line 160: Remove commented code

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1866,7 +1868,14 @@ private void handleNavigation(Location location,
         } catch (Exception exception) {
             handleExceptionNavigation(location, exception);
         } finally {
-            getInternals().clearLastHandledNavigation();
+            if (getInternals().getSession().getConfiguration()
+                    .isReactRouterEnabled()
```
mshabarov, line 1872: I don't get it, so now we have a feature flag and in the same time the config parameter for the same react router integration ?

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -138,13 +139,41 @@ function navigateEventHandler(event) {
...
     } else {
-        navigation(event.detail.pathname, {replace: false})
+        if (mountedContainer?.onBeforeLeave) {
+            mountedContainer?.onBeforeLeave({pathname: event.detail.pathname, search: event.detail.search}, {
+                prevent() {},
+            }, router);
+        } else {
+            mountedContainer?.parentNode?.removeChild(mountedContainer);
+            mountedContainer = undefined;
+            navigation(event.detail.pathname, {replace: false});
```
mshabarov, line 175: I barely understand the changes in this file, cannot anyhow connect the single request and this code. Better way in my opinion would be if we can discuss it together what is the idea behind :)

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1866,7 +1868,14 @@ private void handleNavigation(Location location,
...
+                    && getInternals().getContinueNavigationAction() != null) {
+                getInternals().clearLastHandledNavigation();
+            } else {
+                getInternals().clearLastHandledNavigation();
```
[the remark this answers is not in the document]
caalador: Removed in the next PR that contains some fixes for navigating and adds more tests.

flow-server/src/main/java/com/vaadin/flow/component/UI.java
caalador: DeploymentConfiguration is more stable inside server code and tests than featureFlag.
The information should be the same for both and this can then be used to revert to vaadin-router if react-router becomes the default.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
caalador: So the gist is that if we navigate to the server ('/*' route) then we run any before enter code with the new path and let the server handle navigation. Else if we are navigating somewhere else we call the server before leave to possibly postpone navigation. if no before leave we clean flow from the view and navigate to the new route using react-router.

flow-server/src/main/java/com/vaadin/flow/component/internal/JavaScriptNavigationStateRenderer.java
```diff
@@ -105,7 +106,11 @@ protected boolean shouldPushHistoryState(NavigationEvent event) {
                 || NavigationTrigger.ROUTER_LINK.equals(event.getTrigger())) {
             return true;
         } else {
-            return super.shouldPushHistoryState(event);
+            return super.shouldPushHistoryState(event)
+                    || (event.getUI().getInternals().getSession()
```
mshabarov, line 110: Are this explicit push history action and another history push in `AbstractNavigationStateRenderer.java` basically needed to be done by Flow for React router to ensure that `window.history.pushState` is called after navigation from Hilla view to Flow view? (just a clarification, maybe deserves an explanatory comment for further maintainer).

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
mshabarov: > Else if we are navigating somewhere else we call the server before leave to possibly postpone navigation

If we are on the Hilla view and navigate to something else (non-Flow view I suppose, so Hilla view), why do we need to call server-side before leave method at all ?

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -138,13 +139,41 @@ function navigateEventHandler(event) {
...
-        mountedContainer?.onBeforeLeave({pathname: event.detail.pathname, search: event.detail.search}, {
-            prevent() {},
-        }, router);
+    if(matchPath(event.detail.pathname, window.location.pathname)) {
+        return;
+    }
+    // @ts-ignore
+    let matched = matchRoutes(routes, event.detail.pathname);
+
+    if(matched?.length == 1 && matched[0].route.path === "/*") {
```
mshabarov, line 148: This perhaps deserves a comment that was gives below
```
if we navigate to the server ('/*' route) then we run any before enter code with the new path and let the server handle navigation. Else if we are navigating somewhere else we call the server before leave to possibly postpone navigation. if no before leave we clean flow from the view and navigate to the new route using react-router.
```
for further maintainers.

mshabarov (commented):
Some clarification comments, nothing buggy.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
caalador: No it's navigation to hilla (or other defined react router route). If we are in a hilla view nothing is called from this class.

flow-server/src/main/java/com/vaadin/flow/component/internal/JavaScriptNavigationStateRenderer.java
caalador: This is that as we can not get ReactRouter to update or not update the url as we would require we for react router move the update from the client to the server so that the server handles the url updating.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
mshabarov: Ah, so it's a Flow view where a user is currently on, and upon navigating to Hilla or React route, the Flow calls before leave method. Now the puzzle is complete :)

mshabarov (approved):

Issue #18088, opened 2023-11-21
Getting item's index through DataView/DataProvider API


mshabarov:
### Describe your motivation

Vaadin components with listed items can be [scrolled to item by index](https://github.com/vaadin/flow-components/blob/c15728373a7103ca31046bab06a079b3c63f9613/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java#L4528C29-L4528C29). This isn't always possible, because index of an item might be unknown (e.g. depends on the sorting and filtering).

Thus, an API to get the index by item would be helpful for such uses cases.

### Describe the solution you'd like

Proposed solution is to have a new method in `DataView` API for this.

Acceptance criteria:
1. `DataView` interface has a method `int getItemIndex(T item)` (or with a similar name) that takes items object and return the index of this item.
2. `ListDataView` reuses the existing private method `getItemIndex`.
3. `LazyDataView` has a new method for applying a callback for getting an index from backend:
   ```java
   LazyDataView::setItemIndexProvider(T item, Query<T, F> query)
   ```
   and this callback, if being set by developer, contains a logic for project's backend/database.

   This callback is a more convenient way of customising data view, than extending data view class, because then developer would need to create it's own data view class, find a proper data view base class, set up it to component and so on.

   If no callback was given, `LazyDataView` throws an `UnsupportedOperationException` or similar and highlights that it needs a callback to properly fetch the index.
4. For point 4, a practical example is provided showing how this callback can be implemented, for instance, for relational database.

### Additional context

References:
1. https://github.com/vaadin/web-components/issues/1951
2. [Implementation in Viritin add-on](https://github.com/viritin/flow-viritin/blob/4af9c1e301266609e651fc42227af3d5c26650db/src/main/java/org/vaadin/firitin/components/grid/VGrid.java#L80).

tltv:
`AbstractDataView::verifyDataProviderType` blocks `LazyDataView`'s to work for example when using data provider wrapper like `ConfigurableFilterDataProviderWrapper`, which is practical when using filtering with `DataProvider::withConfigurableFilter` call.
It's not possible to use new API for example with this example code documented here: https://vaadin.com/docs/latest/components/grid#lazy-loading.

To fix that, I'll make another PR for better support for wrappers with `LazyDataView`.

PR #18306, opened 2023-12-14
feat: add getItemIndex to DataView

tltv:
New `getItemIndex` method in `DataView` affects `ListDataView` and `LazyDataView` implementations. `LazyDataView` introduces in addition a new method `setItemIndexProvider(ItemIndexProvider)`. With `ListDataView`, `getItemIndex` works out-of-the-box with the in-memory data. With `LazyDataView`, it's required to set item index provider first with `setItemIndexProvider(ItemIndexProvider)`. Otherwise `getItemIndex` will throw `UnsupportedOperationException`. Provider can be implemented to fetch correct item index using item and `Query` parameters. `Query` object is set up to fetch all items with sorting and filter.

Fixes: #18088


tepi:
Looks good to me. Maybe wait for second opinion since this is new API.

mcollovati:
Also looks good to me.
I just wonder if we should add a detailed Javadoc on `ItemIndexProvider.apply()` to explain what we expect the method to do (e.g. apply filter and sorting) and to warn about potential pitfalls (e.g. inconsistent index upon data changes)

tepi:
> Also looks good to me. I just wonder if we should add a detailed Javadoc on `ItemIndexProvider.apply()` to explain what we expect the method to do (e.g. apply filter and sorting) and to warn about potential pitfalls (e.g. inconsistent index upon data changes)

I agree that this should be added. The former is mentioned [here](https://github.com/vaadin/flow/pull/18306/files#diff-3c170f84b70d5e0806b952e5874d04e4f513d54d362039bcef07302d73ff2297R48-R50) but would be good to repeat in `ItemIndexProvider.apply()` javadoc along with potential pitfalls.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -905,6 +907,27 @@ public void setFetchEnabled(boolean fetchEnabled) {
         this.fetchEnabled = fetchEnabled;
     }
 
+    /**
+     * Gets the item index provider for this data communicator.
+     *
+     * @return the item index provider. May be null.
+     */
+    public ItemIndexProvider<T, ?> getItemIndexProvider() {
```
mshabarov, line 915: I would recommend not to touch DataCommunicator at all and instead place the item index provider directly to `AbstractLazyDataView` class. The reason for it:
1. Let's not leak the API from `DataView` to `DataCommunicator`, it's not really necessary.
2. `DataCommunicator` shouldn't be used by developers at the first place, despite Grid has a direct access to it and for some cases it is need. ComboBox for instance has no getter for it.
If this provider was used somehow in DC, I would maybe agree, but it isn't, so no strong reasons IMO to add it to DC.

mshabarov:
I support this [comment](https://github.com/vaadin/flow/pull/18306#issuecomment-1857353436) and suggest to add an implementation example code snippet to `ItemIndexProvider`'s javadoc

flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
tltv: I started first having it in `AbstractLazyDataView`. Then I realized that data view is actually quite stateless. Everything is stored in `DataCommunicator` or component instance itself, and if you call e.g. `Grid#getLazyDataView`, you'll get new instance of data view. To preserve state, only place to put it is in `DataCommunicator`.
So even though I agree that it's not ideal place to store this particular state, it's inline with other states and also only place if we want to preserve it between `getLazyDataView`.

mshabarov: How about storing it in the Component ?

mshabarov: I suggest to use `ComponentUtil` to store the provider in the backed `Component`, like it's done, for example, in `AbstractDataView::addIdentifierProviderChangeListener`.

When data view stores something in DataCommunicator, these stored objects are used by DataCommunicator itself, but this isn't true for our use case.

tltv: Storing it to `Component` data looks promising. Better than `DataCommunicator`.

mcollovati (dismissed):

flow-data/src/main/java/com/vaadin/flow/data/provider/ItemIndexProvider.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @param <T>
+ *            the type of the item
+ * @param <F>
+ *            the type of the query filter
+ */
+@FunctionalInterface
+public interface ItemIndexProvider<T, F>
+        extends SerializableBiFunction<T, Query<T, F>, Integer> {
+    Integer apply(T item, Query<T, F> query);
```
mshabarov, line 42: Needs javadoc

mshabarov: Also "Since 24.4" might be helpful

mshabarov (changes requested):

mshabarov:
@yuriy-fix could you please verify that this new API is appropriate for components to implement scroll to item API ?

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractLazyDataView.java
```diff
@@ -131,4 +147,30 @@ public void setItemCountUnknown() {
...
+     *
+     * @return the item index provider. May be null.
+     */
+    @SuppressWarnings("unchecked")
+    protected ItemIndexProvider<T, ?> getItemIndexProvider() {
+        return (ItemIndexProvider<T, ?>) ComponentUtil.getData(component,
+                ItemIndexProvider.class);
+    }
+
+    private Query getQueryForAllItems() {
```
tltv, line 168: `getQueryForAllItems` is not returning same filter that data provider may use. E.g when setup with `DataProvider.withConfigurableFilter()`. its not required to use that filter, but I think it would be misleading to leave it without at least trying to give correct filter.

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractLazyDataView.java
```diff
@@ -131,4 +148,44 @@ public void setItemCountUnknown() {
...
+
+    private Query getFilteredQueryForAllItems() {
+        Query baseQuery = getQueryForAllItems();
+        if (DataProviderWrapper.class.isAssignableFrom(
+                dataCommunicator.getDataProvider().getClass())) {
+            DataProviderWrapper<T, ?, ?> wrapper = (DataProviderWrapper<T, ?, ?>) dataCommunicator
+                    .getDataProvider();
+            return new Query(baseQuery.getOffset(), baseQuery.getLimit(),
+                    baseQuery.getSortOrders(), baseQuery.getInMemorySorting(),
+                    wrapper.getFilter(baseQuery));
```
mshabarov, line 186: I'd propose to create a ticket for fixing the filtering this way and:
1. Implement it in the next minor release if the issue gets votes (at least 1-2 I think)
2. Implement it in the next major release instead, as this might be a behaviour change for some projects (fetch less items than expected due to the filter in wrapper data provider).

flow-data/src/main/java/com/vaadin/flow/data/provider/DataView.java
```diff
@@ -45,6 +45,16 @@ public interface DataView<T> extends Serializable {
...
 
+    /**
+     * Gets the index of the given item from the data available to the
+     * component. Data is filtered and sorted the same way as in the component.
+     *
+     * @param item
+     *            item to get index for
+     * @return index of the item or null if the item is not found
+     */
+    Integer getItemIndex(T item);
```
mshabarov, line 56: I propose to return the `Optional<Integer>` instead, this is my personal preference, but we also have `Optional` for item getters, so might be better to align with them.

mshabarov (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractLazyDataView.java
tltv: I just realized after writing almost complete new ticket that I was wrong with `DataProviderWrapper` filter not being used by `getItems`. It is as long as it uses `DataProvider#fetch` method (as it does), filter is added from the data provider properly.

`getItemIndex` is not calling fetch, it calls item index provider with the query, so there it's good to build the filter like this so that developer can use the correct query to fetch the index.

flow-data/src/main/java/com/vaadin/flow/data/provider/DataView.java
tltv: What Item getters do you mean? Nothing like that in the DataView.

tltv: Never mind, I see getNextItem/getPrevItem in ListDataView.

flow-data/src/main/java/com/vaadin/flow/data/provider/ItemIndexProvider.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+@FunctionalInterface
+public interface ItemIndexProvider<T, F>
+        extends SerializableBiFunction<T, Query<T, F>, Integer> {
+    /**
+     * Gets the index of the item in the filtered and sorted data set.
+     * <p>
+     * There will be inconsistent index if the data set for the returned index
+     * is different from the component's data set. Changing the data set of
+     * either side during this call may cause inconsistent index as a result.
```
mshabarov, line 48:
```suggestion
     * either side during this call may cause inconsistent index as a result.
     * <p>
     * The query parameter provides a filter object being set with 
     * {@link ConfigurableFilterDataProvider} or provided by a component, 
     * e.g. a string filter in ComboBox.
```

mshabarov (dismissed):
Have one clarification for javadoc, otherwise looks good to me

mshabarov (approved):

PR #18308, opened 2023-12-14
chore: Upgrade @koa/cors

Artur-:
Breaking change between 5.0 and 4.0
The default origin is set to *, if you want to keep the 4.0 behavior, you can set the origin handler like this:

app.use(cors({
  origin(ctx) {
    return ctx.get('Origin') || '*';
  },
}));

### DIFF
--- a/vaadin-dev-server/package.json
+++ b/vaadin-dev-server/package.json
@@ -9,7 +9,7 @@
     "prettier": "prettier --write frontend"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


mcollovati:
Will this change somehow impact user projects, or does it only affect `vaadin-dev-server`?

Artur-:
It is only used for testing in vaadin-dev-server

mcollovati (approved):

PR #18317, opened 2023-12-15
feat: support DataProviderWrapper better in AbstractDataView

tltv:
Adds in `AbstractDataView` a better support for wrapper data providers that implements `DataProviderWrapper`. `AbstractDataView` won't throw exception anymore if wrapped data container is supported by the data view implementation. Adds public `getWrappedDataProvider` method in `DataProviderWrapper`.

Related-to: #18088

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataProviderWrapper.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataProviderWrapper.java
@@ -54,6 +54,15 @@ public abstract class DataProviderWrapper<T, F, M>
                 "The wrapped data provider cannot be null.");
     }
 
+    /**
+     * Gets wrapped data provider.
+     *
+     * @return the wrapped data provider
+     */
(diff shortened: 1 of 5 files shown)


mshabarov (dismissed):

flow-data/src/main/java/com/vaadin/flow/data/provider/DataProviderWrapper.java
```diff
@@ -54,6 +54,15 @@ protected DataProviderWrapper(DataProvider<T, M> dataProvider) {
                 "The wrapped data provider cannot be null.");
     }
 
+    /**
+     * Gets wrapped data provider.
+     *
+     * @return the wrapped data provider
+     */
+    public DataProvider<T, M> getWrappedDataProvider() {
```
mshabarov, line 62: This might be a package-private by the way, since API users don't need the wrapped instance usually.

mshabarov (approved):

PR #18319, opened 2023-12-15
refactor!: Remove deployment conf field from VaadinSession

### LINKED ISSUE #18318: DeploymentConfiguration is retained in VaadinSession over session store-load cycle
### Description of the bug

In addition to VaadinService, also VaadinSession has a field for DeploymentConfiguration. That is not marked as transient and will be serialized along with the session. This can lead to non-valid data in the configuration if it is modified and a redeploy is done between session store and load.

### Expected behavior

DeploymentConfiguration data should always be up-to-date.

n/a

### Versions

- Vaadin / Flow version: latest

tepi:
Fixes #18318

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
@@ -139,14 +139,10 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
      *            the Vaadin service for the new session
      */
     public VaadinSession(VaadinService service) {
+        Objects.requireNonNull(service, "VaadinService can not be null.");
         this.service = service;
 
-        if (service != null) {
-            sessionLockCheckStrategy = getConfiguration().isProductionMode()
-                    ? getConfiguration().getSessionLockCheckStrategy()
-                    : SessionLockCheckStrategy.THROW;
-            assert sessionLockCheckStrategy != null;
-        }
+        refreshSessionLockCheckStrategy();
 
         resourceRegistry = createStreamResourceRegistry();
     }
@@ -345,6 +341,16 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
                 : "Cannot change the lock from one instance to another";
         assert hasLock(service, session);
         lock = service.getSessionLock(session);
+    }
+
+    /**
+     * Updates the session lock check strategy.
+     */
+    private void refreshSessionLockCheckStrategy() {
+        sessionLockCheckStrategy = getConfiguration().isProductionMode()
+                ? getConfiguration().getSessionLockCheckStrategy()
+                : SessionLockCheckStrategy.THROW;
+        assert sessionLockCheckStrategy != null;
     }
 
     /**
@@ -1102,11 +1108,7 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
         session = wrappedSession;
         service = vaadinService;
 
-        sessionLockCheckStrategy = getConfiguration().isProductionMode()
-                ? getConfiguration().getSessionLockCheckStrategy()
-                : SessionLockCheckStrategy.THROW;
-        assert sessionLockCheckStrategy != null;
-
+        refreshSessionLockCheckStrategy();
         refreshLock();
     }
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
@@ -79,11 +79,6 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
     volatile boolean sessionClosedExplicitly = false;
 
     /**
-     * Configuration for the session.
-     */
-    private DeploymentConfiguration configuration;
-
-    /**
      * Default locale of the session.
      */
     private Locale locale = Locale.getDefault();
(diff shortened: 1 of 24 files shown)


flow-server/src/test/java/com/vaadin/flow/component/InvalidUrlTest.java
```diff
@@ -80,18 +80,19 @@ private static void initUI(UI ui, String initialLocation,
...
         };
         service.setCurrentInstances(request, response);
 
         MockVaadinSession session = new AlwaysLockedVaadinSession(service);
 
-        DeploymentConfiguration config = Mockito
-                .mock(DeploymentConfiguration.class);
-        Mockito.when(config.isProductionMode()).thenReturn(false);
-
-        session.lock();
```
mcollovati, line 92: If we remove this `session.lock()`, should we also remove the `session.unlock()` at the end of the method?
Or should the `session.lock()` be preserved?

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -145,6 +140,14 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
      */
     public VaadinSession(VaadinService service) {
         this.service = service;
+
+        if (service != null) {
```
mcollovati, line 144: I wonder if we should throw an exception when service is null.
A VaadinSession without a service reference will not work correctly, and the only moment I can imagine the field to be null is on deserialization, but in that case, the constructor will not be invoked.

mcollovati (changes requested):
The change looks good and makes sense.
A minor drawback will be that it might break some test configurations on other repositories. In MPR for example, it is used in a couple of tests, and the problem here may be that we use the same MPR codebase against different Flow version.

Potentially also user projects may be affected, since `setConfiguration()` is a public API, even though I suppose `VaadinSession.setConfiguration()` is not widely used.

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
tepi: You are right, that would make sense. I'll do the change, although it will mean revisiting at least one (quite old) test, which is calling the constructor with null parameter (claiming it as an 'uninitialized' session, whatever that means).

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -1117,6 +1101,12 @@ public void refreshTransients(WrappedSession wrappedSession,
             VaadinService vaadinService) {
         session = wrappedSession;
         service = vaadinService;
+
+        sessionLockCheckStrategy = getConfiguration().isProductionMode()
+                ? getConfiguration().getSessionLockCheckStrategy()
+                : SessionLockCheckStrategy.THROW;
+        assert sessionLockCheckStrategy != null;
```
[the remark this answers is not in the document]
tepi: Will do

flow-server/src/test/java/com/vaadin/flow/component/InvalidUrlTest.java
tepi: Good catch. I'll need to check that on other tests as well. Probably the unlock is not needed.

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
mcollovati: Could it be a post deserialization simulation? Anyway, I would probably nullify the field through reflection in that case

tepi: It's this one: https://github.com/vaadin/flow/blob/main/flow-server/src/test/java/com/vaadin/flow/server/VaadinSessionTest.java#L483

mcollovati: HTTP session unboud is called when the implementing object is removed from the HTTP session, so I think that in that case the VaadinSession should have a not-null reference to VaadinService

tepi: Ok, I'll have to check how to fix it then. Initially I just changed the null to a mock VaadinService but that will fail the test.

mcollovati: Ah, my bad. The `valueUnbound` method Javadoc explains the case.
I propose to have the null-check in the constructor and nullify it with reflection (or serialize/deserialized the VaadinSession object)

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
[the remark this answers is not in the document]
tepi: Moved to a new private method

flow-server/src/test/java/com/vaadin/flow/component/InvalidUrlTest.java
tepi: Fixed, and found no other instances of this case

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
tepi: Fixed

mcollovati:
Some tests to fix, but other than that LGTM

mcollovati (approved):

tepi:
Despite approval, do not merge. This should be postponed until major release of Flow (25), since this changes public API.

mcollovati:
Would it be better to make it a draft again, to prevent unwanted merge?

tepi:
Closed in favor of https://github.com/vaadin/flow/pull/21938

PR #18321, opened 2023-12-18
test: add tests for navigating with react-router

### LINKED ISSUE #18267: Add tests for running with the react-router
Add tests that do navigation between flow views using server-side navigation and links.

caalador:
Add test for navigating using
different methods when
react-router is in use

Closes #18267


mshabarov (commented):
No big objectives on how these tests are implemented, but two comments regarding what else we should cover:
1. It's worth IMO to run the same set of tests for **production mode** from the very beginning, so I propose to add a prod pom.xml configuration to this test module. Do you see any obstacles to do it in this PR?
2. Regarding the test cases that we did manually, I think we have to include the checks for error handling, navigation to the protected Flow views and (probably) checking that all the navigation events are triggered on the server, when navigation from Hilla view to Flow view. What do you think?

caalador:
Added production pom.

Don't think we can have Hilla in the test mix, but should be fine to have a react component that is navigated to and from which is basically the same.

mshabarov:
> but should be fine to have a react component that is navigated to and from which is basically the same.

Sounds good. Let's maybe add it later in the separate PR ?

mshabarov (approved):

mshabarov (approved):

PR #18337, opened 2023-12-20
feat: rename `dev.hilla` to `com.vaadin.hilla`

cromoteca:
## Description

Hilla is undergoing a package renaming that affects some files in Flow. This PR makes the necessary changes.

Part of https://github.com/vaadin/hilla/issues/1904

## Type of change

- [ ] Bugfix
- [X] Feature

### DIFF
--- a/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/BuildFrontendMojoTest.java
+++ b/flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/BuildFrontendMojoTest.java
@@ -147,7 +147,7 @@ public class BuildFrontendMojoTest {
 
         defaultJavaSource = new File(".", "src/test/java");
         openApiJsonFile = new File(npmFolder,
-                "target/classes/dev/hilla/openapi.json");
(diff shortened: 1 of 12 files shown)


mcollovati (approved):

tarekoraby:
Is the `0.01` label the right one? This should land in Flow 24.4.

mcollovati:
> Is the 0.0.1 label the right one? This should land in Flow 24.4.

Label is applied automatically (https://teamcity.vaadin.com/viewLog.html?buildId=214295&buildTypeId=Flow_FlowPullRequestsValidationSet_ApiDiffLabeling) and the PR does not introduce changes that may give a different label.

Anyway, flow `main` is currently `24.4-SNAPSHOT` so `0.0.1` makes sense to me

caalador (approved):

mshabarov (approved):

tarekoraby:
I was just worried that `0.0.1` implies that this will be picked up in a 24.3 patch release.

caalador:
the label just shows what the target should be by minimum so by the rules this needs a patch release where as 0.1.0 would mean that it needs a minor bump and 1.0.0 needs a major bump.

Issue #18365, opened 2023-12-27
Dev bundle's stats.json is missing file hash for `@JsModule` imports


tltv:
### Description of the bug
Development bundle's stats.json doesn't add hash for resources added with `@JsModule` (or `@Javascript`) annotation.

### Minimal reproducible example
1. Create a starter Vaadin 24 Spring boot app (https://start.vaadin.com/app)
2. Add `@JsModule("./test.js")` to some route
3. Add `test.js` file in `frontend` folder. Content does not matter as long as it compiles.
4. Start the app. `Frontend/test.js` is missing from the bundle. Dev bundle is build.
5. Change something (not the test.js) to trigger a restart
6. See in logs lines: `No hash info for 'test.js'` and `Detected changed content for frontend files: - test.js`. Dev bundle is build again.
7. Check `target/dev-bundle/config/stats.json`. it doesn't contain `test.js` file with a hash in `frontendHashes`.
### Expected behavior
Dev bundle build should not be needed if `test.js` was not changed. It shouldn't say that file was changed when it wasn't.
### Actual behavior
See above.
### Versions:
    - Vaadin / Flow version: 24.3.0
    - Java version: 17
    - OS version: Windows 10
    - Browser version (if applicable): Chrome
    - Application Server (if applicable):
    - IDE (if applicable):
    - Development or production mode: dev
    - Project from start.vaadin.com or activated in application:

PR #18368, opened 2023-12-28
fix: consider local npm packages in bundle validation

### LINKED ISSUE #18106: Production bundle error when running Hilla tests
### Description of the bug

```
[INFO] --- hilla:2.5-SNAPSHOT:build-frontend (default) @ tests-spring-security-jwt ---
[INFO] Reflections took 425 ms to scan 143 urls, producing 9972 keys and 53191 values
[INFO] Scanning classes to find frontend configurations and dependencies...
[INFO] Visited 1000 classes. Took 347 ms.
[INFO] Checking if a production mode bundle build is needed
[ERROR] Error when checking if a production bundle build is needed
java.lang.NumberFormatException: 'file:../../../../ts/form' is not a valid version!
    at com.vaadin.flow.server.frontend.FrontendVersion.<init> (FrontendVersion.java:160)
    at com.vaadin.flow.server.frontend.FrontendVersion.<init> (FrontendVersion.java:128)
    at com.vaadin.flow.server.frontend.BundleValidationUtil.versionAccepted (BundleValidationUtil.java:452)
    at com.vaadin.flow.server.frontend.BundleValidationUtil.lambda$hashAndBundleModulesEqual$2 (BundleValidationUtil.java:434)
...
[INFO] Copying frontend resources from jar files ...
[INFO] Visited 143 resources. Took 38 ms.
[ERROR] Require bundle build due to missing 'config/needs-build' file.
[INFO] Running Vite ...
```

### Expected behavior

No errors

https://github.com/vaadin/hilla/pull/1755

### Versions

- Vaadin / Flow version: 24.3.0.alpha
- Java version:
- OS version:
- IDE (if applicable):

mcollovati:
## Description

If a project package.json contains references to local npm pacakges instead of parsable version, the bundle validation fails. This change update the version check in bundle validation to accept also local packages.
For those packages, a bundle re-creation is triggered only if the reference in package.json does not match the one in the stats.

Fixes #18106

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
@@ -13,6 +13,7 @@ import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -449,8 +450,30 @@ public final class BundleValidationUtil {
     }
 
     private static boolean versionAccepted(String expected, String actual) {
-        FrontendVersion expectedVersion = new FrontendVersion(expected);
-        FrontendVersion actualVersion = new FrontendVersion(actual);
+        FrontendVersion expectedVersion;
+        try {
+            expectedVersion = new FrontendVersion(expected);
+        } catch (NumberFormatException ex) {
+            expectedVersion = null;
+        }
+        FrontendVersion actualVersion;
+        try {
+            actualVersion = new FrontendVersion(actual);
+        } catch (NumberFormatException ex) {
+            actualVersion = null;
+        }
+
+        if (expectedVersion == null && actualVersion == null) {
+            return Objects.equals(expected, actual);
+        } else if (expectedVersion == null || actualVersion == null) {
+            // expected or actual version is referencing a local package
+            // while the other one is a parsable version
+            getLogger().debug(
+                    "Version '{}' cannot be parsed and compared to '{}'",
+                    expectedVersion == null ? expected : actual,
+                    expectedVersion == null ? actual : expected);
+            return false;
+        }
 
         if (expected.startsWith("~")) {
             boolean correctRange = expectedVersion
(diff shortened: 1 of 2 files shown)


mcollovati:
This patch prevents the version parsing exception.
However, currently, the bundle validation might not be able to detect if there are changes in the local package that will require a bundle rebuild.
Rebuild will be triggered if changed components from local packages are imported in fronted source, but not if they are referenced, for example, with `@JsModule` annotation.
This situation should be mitigated once #18365 will be fixed

mshabarov:
I propose for now to go with the implemented solution.

IMO this is quite a rare case for pure Flow apps when development goes in local npm package (otherwise I beileve the hotdeploy mode should be used).
Upon changes in local package, one can `mvn clean` and delete `dev.bundle` file from `src/main/bundles` and he gets a re-bundling then. 

If we get a request to make this happen automatically, we can consider a more advanced solution.

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
```diff
@@ -449,8 +450,26 @@ public static boolean hashAndBundleModulesEqual(JsonObject statsJson,
...
+        } catch (NumberFormatException ex) {
+            actualVersion = null;
+        }
+
+        if (expectedVersion == null && actualVersion == null) {
+            return Objects.equals(expected, actual);
+        } else if (expectedVersion == null || actualVersion == null) {
+            // expected or actual version is referencing a local package
+            // while the other one is a parsable version
+            return false;
```
mshabarov, line 471: Are these cases covered with the existing logging? 
If not, they need an additional INFO or DEBUG logging telling why re-bundling is happening.

mshabarov (changes requested):
Small loggin-related question, otherwise looks good.

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
mcollovati: Currently, it will log that the dependency has a wrong version in the bundle.
Adding a DEBUG level log that reports both version make sense to help in troubleshooting

mcollovati: added log

mshabarov (approved):

PR #18385, opened 2023-12-29
fix: permit access to custom PWA icons and favicon

### LINKED ISSUE #17342: Setting favicon breaks loginSuccessUrl in SpringBoot app
### Description of the bug

Setting favicon via `addFavIcon()` methods breaks the login flow. After form submit app redirects to the `http://<whatever>/icons/favicon.png`. This is because `VaadinDefaultRequestCache` stores the previous request. While setting ...

```java
http.requestCache(c -> c.requestCache(new NullRequestCache()));
```

.. solves the problem, I think Vaadin request cache should be configured accordingly.

PS: I tried _every_ favicon path specified in the doc (including `favicon.icon` which I believe is a typo, because such extension doesn't exist) and none of them working, which I think is another problem.

### Expected behavior

Exclude favicons from security request cache.

```
@SpringBootApplication
public class Application implements AppShellConfigurator {

        SpringApplication.run(Application.class, args);
    }

    @Override
    public void configurePage(AppShellSettings settings) {
        settings.addFavIcon("icon", "icons/favicon.png", "64x64");
    }
}

@EnableWebSecurity
@Configuration
public class SecurityConfiguration extends VaadinWebSecurity {

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        http.authorizeHttpRequests(auth ->
            auth.requestMatchers(new AntPathRequestMatcher("/public/**")).permitAll()
        );

        super.configure(http);

        setLoginView(http, LoginPage.class);
    }

    @Override
    public void configure(WebSecurity web) throws Exception {
        super.configure(web);
    }

    @Bean
    public UserDetailsManager userDetailsService() {
        UserDetails admin = User.withUsername("admin")
            .password("{noop}admin")
            .roles("ADMIN")
            .build();
        return new InMemoryUserDetailsManager(admin);
    }
}
```

I only added favicon method. Everything else is from [tutorial](https://vaadin.com/docs/latest/security/enabling-security).

### Versions

- Vaadin / Flow version: 24.1.4
- Java version: 17
- OS version: Fedora 37
- Spring Boot version: 3.1.2

mcollovati:
## Description

This change configures VaadinWebSecurity with an additional request matcher that grants access to everyone to custom PWA icons and favicon paths.

Fixes #17342

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


caalador (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/WebIconsRequestMatcher.java
```diff
@@ -0,0 +1,155 @@
...
+                private void registerPath(String path) {
+                    if (!path.startsWith("/")) {
+                        path = urlMapper.apply(path);
+                    }
+                    paths.add(path);
+                }
+            };
+            try {
+                vaadinService.getInstantiator().getOrCreate(appShellClass)
+                        .configurePage(settings);
```
caalador, line 117: One thing is that multiple calls to configurePage might be confusing for developers.

mcollovati: I had the same concern, but I also thought that we already instantiate a new `appShellClass` class and call the method for every UI we create, so multiple invocation on different instances should not be unexpected.

mcollovati:
I wonder if we should expand this to also handle `AppShellSettings.addLink()`; perhaps adding overloads that take a `boolen public` parameter (e.g. `public void addLink(String rel, String href, boolean public){...}`) or adding specific methods like `public void addPublicLink(String rel, String href)`.
Anyway, this could be done in a separate PR

PR #18387, opened 2024-01-02
feat!: Make react-router default

caalador:
Enable react router as default with
property to switch back to vaadin router.


flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
```diff
@@ -161,6 +161,9 @@ public class BuildDevBundleMojo extends AbstractMojo
     @Parameter(property = "npm.postinstallPackages", defaultValue = "")
     private List<String> postinstallPackages;
 
+    @Parameter(property = InitParameters.REACT_ROUTER_ENABLED, defaultValue = "true")
+    private boolean reactRouterEnabled;
```
knoobie, line 165: Are there any public available discussions about the benefits or downsides from this change for flow developers? Increased package size? Additional dependency on production to react? Performance downside or improvements? Customized `index.html` still applied?

tarekoraby: It's not a discussion per se, but there is this acceptance criteria: https://github.com/vaadin/platform/issues/4710. I believe the net effects should be neutral in terms of performance and package sizes. The main benefit is enabling mixing Flow and Hilla/React views in the same app. For pure-Flow projects, the change should be neutral.

knoobie: Understandable, I was just wondering because the default is swapped in this PR (opposite as described in the AC) and therefore it would affect all flow apps. This is why I was wondering what the benefits or downsides are for flow only users :) (except to get rid of the "abandoned" vaadin router in favor of the react router)

mshabarov: > Increased package size? Additional dependency on production to react?

Good point and we're going to estimate the bundle size. I don't expect much extra comparing to previous default, but let's see.

> Customized index.html still applied?

Yes, I expect so, haven't tested yet though. Let me add a note about it.

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/PrepareFrontendInputProperties.kt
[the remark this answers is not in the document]
caalador: I don't know. Basically it does change the package.json contents so perhaps?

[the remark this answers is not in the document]
caalador: added.

mshabarov (dismissed):

mshabarov (approved):

PR #18392, opened 2024-01-03
feat: SBOM generation

### LINKED ISSUE #14: Include SBOM generation as a kit feature
In #3 the CycloneDX Maven plugin has been identified as a proper tool to generate the SBOM from the application transitive dependency, which is currently used in the demo project.

This approach requires the developer to use the third-party plugin in their own Maven project, or alternative to provide a compatible SBOM in the application resources.

It would be more convenient to include this process as a kit feature, either by:
1. adding a new goal to the existing Vaadin Maven plugin to generate the SBOM
2. create a new Maven plugin which is part of the kit (to simplify the release process)

In both scenarios the CycloneDX API should be used to generate a compatible SBOM.

sosa-vaadin:
## Description
We want to generate SBOM files for Vaadin apps (1 file for maven dependencies, 1 file for npm dependencies).

### Approach 
We're creating a new goal for each file in the `flow-maven-plugin` subproject of `flow`. Both goals use [maven-invoker](https://maven.apache.org/shared/maven-invoker/) which then calls third party maven plugins that generate cyclonedx bom files. We tested it manually and works wonders. When shipped, this will allow Vaadin users to run 
```
mvn flow:generate-sbom
``` 
and/or 
```
mvn flow:generate-npm-sbom
``` 
to generate bom files with all the dependencies of their Vaadin apps. Our `appsec-kit` can then use these files to report vulnerabilities.

### Problem
The `maven-invoker` needs to find a maven executable. When running maven goals as you normally do, it's no problem. Because a so called `maven.home` system property is set with the address to a maven executable when running maven plugin goals. The `maven-invoker` then checks this property and uses its value. **This property is not set when running tests**, either via running `mvn test` or through your favourite IDE using JUnit.

### Solution
This is where I need ideas. The address of a maven executable can be set to the `maven-invoker` but I can't think of a way to do it reliably. I went with checking for an environment variable `MAVEN_HOME` and if set, then use that as the address to maven. An obvious issue with this is that any flow developer and any CI tool will need this property to be set as well, or have these couple of tests fail. 

We really want this feature so that sbom file generation can be part of any Vaadin app's build pipeline. But without providing reliable tests, we can't ensure the quality we (and our customers) expect from our products.

Fixes https://github.com/vaadin/appsec-kit/issues/14

## Type of change

- [x] Feature

## Checklist


flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateFrontendBOMMojo.java
```diff
@@ -0,0 +1,87 @@
...
+    public void execute() throws MojoExecutionException, MojoFailureException {
+        InvocationRequestBuilder requestBuilder = new InvocationRequestBuilder();
+        InvocationRequest request = requestBuilder.groupId(GROUP)
+                .artifactId(ARTIFACT).version(VERSION).goal(GOAL)
+                .createInvocationRequest();
+
+        Properties properties = new Properties();
+        properties.setProperty("exec.executable", "npx");
+        properties.setProperty("exec.args",
+                "@cyclonedx/cyclonedx-npm --output-file " + outputFilePath
```
knoobie, line 63: This always downloads the latest version and creates the latest version. Should this be configured to a specific version of the used Dependency as well as the CycloneDX Spec (e.g. 1.4)?

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateSBOMMojo.java
```diff
@@ -0,0 +1,194 @@
...
+    private static final String EXCLUDE_GROUP_ID = "excludeGroupId";
+    private static final String EXCLUDE_TEST_PROJECT = "excludeTestProject";
+    private static final String CYCLONEDX_VERBOSE = "cyclonedx.verbose";
+    private static final String VERBOSE = "verbose";
+    /**
+     * The component type associated to the SBOM metadata. See <a href=
+     * "https://cyclonedx.org/docs/1.4/json/#metadata_component_type">CycloneDX
+     * reference</a> for supported values.
+     */
+    @Parameter(property = PROJECT_TYPE, defaultValue = "library")
```
knoobie, line 53: Should this default be changed? The default makes sense for CycloneDX, but flow "applications" are normally "applications" and no libs.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateSBOMMojo.java
```diff
@@ -0,0 +1,194 @@
...
+    private String projectType;
+    /**
+     * The CycloneDX schema version the BOM will comply with.
+     */
+    @Parameter(property = SCHEMA_VERSION, defaultValue = "1.4")
```
knoobie, line 58: Align with the version of the npm bom?

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateFrontendBOMMojo.java
```diff
@@ -0,0 +1,87 @@
...
+     * The path to the file to be generated.
+     */
+    @Parameter(property = OUTPUT_FILE_PATH, defaultValue = "${project.build.outputDirectory}/resources/bom-npm.json")
+    private String outputFilePath;
+
+    /**
+     * The path to the package.json file to read.
+     */
+    @Parameter(property = PACKAGE_MANIFEST, defaultValue = "./package.json")
+    private String packageManifest;
```
knoobie, line 51: This only works with a package.json if I understand it correctly, ergo it won't work with the "new" modes that is used by flow by default with the development and production bundles. Should it be checked that the file is present?

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateFrontendBOMMojo.java
sosa-vaadin: Added a new parameter so that the schema version is configurable.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateSBOMMojo.java
sosa-vaadin: Made the npm bom schema configurable, just like this one.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateSBOMMojo.java
sosa-vaadin: The default in CycloneDX docs is library. However, since you are correct in that Vaadin apps are mostly apps, I changed the default value here to be "applications".

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateFrontendBOMMojo.java
sosa-vaadin: A package.json file is indeed necessary for the generation of the npm sbom. If no such file is present, the goal will fail.
@mshabarov Can you give us your input here?

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateFrontendBOMMojo.java
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Goal that generates a CycloneDX SBOM file focused on frontend dependencies.
+ */
+@Mojo(name = "generate-npm-sbom", requiresDependencyResolution = ResolutionScope.COMPILE, defaultPhase = LifecyclePhase.GENERATE_RESOURCES)
+public class GenerateFrontendBOMMojo extends AbstractMojo {
+
+    private static final String OUTPUT_FORMAT = "outputFormat";
+    private static final String OUTPUT_FILE_PATH = "outputDirectory";
+    private static final String PACKAGE_MANIFEST = "packageManifest";
+    private static final String SPEC_VERSION = "specVersion";
```
tamasmak, line 43: Should we add properties for the rest of the `cyclonedx-npm` plugin options which are listed [here](https://www.npmjs.com/package/@cyclonedx/cyclonedx-npm) to be able to use those also if needed? The `GenerateSBOMMojo` contains most of the `cyclonedx-maven-plugin` options already.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateSBOMMojo.java
```diff
@@ -0,0 +1,209 @@
...
+import org.apache.maven.shared.invoker.InvocationRequest;
+import org.apache.maven.shared.invoker.InvocationResult;
+import org.apache.maven.shared.invoker.Invoker;
+import org.apache.maven.shared.invoker.MavenInvocationException;
+
+/**
+ * Goal that generates a CycloneDX SBOM file focused on backend dependencies.
+ */
+@Mojo(name = "generate-sbom", requiresDependencyResolution = ResolutionScope.COMPILE, defaultPhase = LifecyclePhase.GENERATE_RESOURCES)
+public class GenerateSBOMMojo extends AbstractMojo {
```
tamasmak, line 37: Should the `GenerateFrontendBOMMojo` and `GenerateSBOMMojo` class names follow some common naming convention?

Example using only BOM or SBOM in the names
 `GenerateFrontendBOMMojo`
`GenerateBackendBOMMojo`
or
`GenerateNpmBOMMojo`
`GenerateMavenBOMMojo`
?

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateFrontendBOMMojo.java
tepi: I agree, the rest of the options should be added as well. Additionally, if the parameter field names in java match the parameter names, the `property` part can be left out from the `@Parameter` annotation - which means the static fields for property names can be dropped too.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateSBOMMojo.java
tepi: I would go for the latter suggestion.

tepi:
Just to be clear, does the developer who wants to use AppSec Kit need to run these two goals manually? Or how will it be done in practice?

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateSBOMMojo.java
sosa-vaadin: Done.

sosa-vaadin:
The idea is to use these new goals the same way we use the `generate-frontend` goal. Add these goals to the pom of a Vaadin app so the files are generated on every build. Of course developers can still run the goals manually to generate the files. The `generate-npm-sbom` goal can be omitted for projects with no `package.json` file. AppSec Kit can handle frontend dependencies even if there's no `package.json` file.

tepi:
> The `generate-npm-sbom` goal can be omitted for projects with no `package.json` file. AppSec Kit can handle frontend dependencies even if there's no `package.json` file.

Sounds good. Just to confirm - the developer should not get exceptions while `package.json` file is missing?

tepi:
Looks like after the change [here](https://github.com/vaadin/flow/pull/18392/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R552-R554), the tests can be run without issues using command-line maven. But running them from IntelliJ still leads to maven executable not being found. @sosa-vaadin @mshabarov any ideas for fix?

sosa-vaadin:
To run the test with IntelliJ, I had to tinker with the run configuration to add an environment variable pointing to my maven executable.
image

sosa-vaadin:
>Sounds good. Just to confirm - the developer should not get exceptions while package.json file is missing?
I think it should throw an exception. Either there must be a `package.json` file, or the goal should be removed from the pom (so as to not run it at all).

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateFrontendBOMMojo.java
tamasmak: I think the `package.json` file should be checked if it is present. If not, the goal should throw an exception as commented [here](https://github.com/vaadin/flow/pull/18392#issuecomment-1891631609), or log some warning message that it is a requirement to be present.

tepi: Ok, I agree. Just thinking out loud that in case we have starters or something that run this goal by default, I'd not like it to fail if `package.json` is missing.

mshabarov: With the pre-compiled bundles the package.json isn't usually in the project root directory indeed.

But you can get the content of it programmatically using Flow's util method `BundleValidationUtil::getPackageJson`. It will give to the package.json from project root, if it exists, or calculates the dependencies used in application and returns the content of this file to you as a JsonObject.

Also, I think we already discussed a similar problem before with @tamasmak .

There was a conclusion at that point:

```
Maybe even this function won’t be needed. In platform there is already a SBOM [built](https://github.com/vaadin/platform/blob/main/scripts/generateAndCheckSBOM.js#L485) with the same plugin, if we could get that from there that would be the best. In all other cases when the package.json is present we can use it to generate the SBOM.
```

Here is a [discussion in Slack 
](https://vaadin.slack.com/archives/C6X43FE8M/p1692957649050189?thread_ts=1692786162.763219&cid=C6X43FE8M).

tamasmak: Yes, I remember on this discussion @mshabarov. The mentioned conclusion is the solution what we use in the AppSec Kit currently. We get the already generated combined (contains the `maven` and `npm` dependencies too) SBOM json file from the platform if the `package.json` is not present, then we parse it and [filter out](https://github.com/vaadin/appsec-kit/blob/main/appsec-kit-backend/src/main/java/com/vaadin/appsec/backend/BillOfMaterialsStore.java#L76) the `maven` dependencies. If we wanted to use the same solution here, then, after filtering out the `maven` dependencies, we would need to save the result to a new SBOM json file.

The purpose of this maven goal would be to help generate a similar SBOM json file but only with `npm` dependencies. We could achieve this with the above solution or with the `BundleValidationUtil::getPackageJson` method in which case we would need to save the `JsonObject` to a `package.json` file, if not present, to be available for the `cyclonedx-npm` plugin and (probably) to delete it after generating the SBOM file, if it wasn't there before.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateMavenBOMMojo.java
```diff
@@ -0,0 +1,210 @@
...
+        properties.setProperty(EXCLUDE_TEST_PROJECT,
+                String.valueOf(excludeTestProject));
+        properties.setProperty(CYCLONEDX_VERBOSE, String.valueOf(verbose));
+        request.setProperties(properties);
+
+        Invoker invoker = new DefaultInvoker();
+        try {
+            InvocationResult result = invoker.execute(request);
+            if (result.getExitCode() != 0) {
+                throw new RuntimeException("Maven SBOM generation failed.",
```
tamasmak, line 202: Should we use `MojoExecutionException` or `MojoFailureException`, whichever is more appropriate, instead of the `RuntimeException` here and everywhere else? This `execute()` method throws those and these are used in other Mojos too.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/GenerateFrontendBOMMojo.java
tamasmak: > I think the `package.json` file should be checked if it is present. If not, the goal should throw an exception as commented [here](https://github.com/vaadin/flow/pull/18392#issuecomment-1891631609), or log some warning message that it is a requirement to be present.

We decided to proceed with a warning message that the p`package.json` file is not present for now. Later we can improve the plugin with getting the npm SBOM from elsewhere if needed.

tepi: > We decided to proceed with a warning message that the `package.json` file is not present for now. Later we can improve the plugin with getting the npm SBOM from elsewhere if needed.

Sorry if being a little thick here, but earlier [David commented](https://github.com/vaadin/flow/pull/18392#issuecomment-1891524586) that "AppSec Kit can handle frontend dependencies even if there's no package.json file." - so why do we care about the package.json file?

sosa-vaadin: Because the Maven goal will still fail without a `package.json`. The Maven goal has nothing to do with AppSec Kit. @tamasmak  meant we will update the Maven plugin goal in a future PR so that it can work by doing what the [AppSec Kit does now to get the frontend dependencies](https://github.com/vaadin/flow/pull/18392#discussion_r1452308048).

tepi: > The Maven goal has nothing to do with AppSec Kit.
Thanks! This clears it up.

tepi (approved):
I think all comments have been addressed now. Since these new goals are optional, it is ok for the `GenerateNpmBOMMojo` to throw if `package.json` is not available. So this can be merged.

PR #18435, opened 2024-01-11
Add documentation link to proxy info messages

tepi:


flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/ProxyConfig.java
```diff
@@ -66,8 +66,12 @@ public boolean isEmpty() {
      * @return Proxy if one found, else null.
      */
     public Proxy getProxyForUrl(String requestUrl) {
+        String docLink = "See https://vaadin.com/docs/latest/configuration/"
+                + "development-mode/node-js#proxy-settings-for-downloading-"
+                + "frontend-toolchain for information on proxy configuration.";
```
mcollovati, line 71: The message `No proxies configured. See .... for information on proxy configuration` looks like to me that we are telling the developer that setting the proxy is somehow mandatory.
Could we reword the sentence to make it clear that the need for proxy configuration depends on the local environment?

E.g.

`No proxies configured. If you are behind a corporate proxy server, see ....`

mcollovati: Probably the `If you are behind a corporate proxy server` is only needed in the `proxies.isEmpty()` branch

mcollovati (changes requested):

mcollovati (approved):

PR #18455, opened 2024-01-15
fix: Remove fallback middleware also after Vite restart

Artur-:
The existing hacky way of removing the middleware has two issues:

1. It removes it AFTER the first request to a non-existing file so the first request for a non-existing resources index.html and subsequent requests return 404

2. It does not work with Vite 4.0.5+ after restart, see https://github.com/vitejs/vite/issues/15589


Artur-:
You can test the problem using
```
npx @hilla/cli init hello
mvn
curl http://localhost:8080/VAADIN/static/push/vaadinPush.js # Returns vaadinPush.js
touch vite.config.ts
curl http://localhost:8080/VAADIN/static/push/vaadinPush.js # Returns index.html
```

mshabarov (approved):

PR #18462, opened 2024-01-15
Deprecate HtmlImport and Compatibility Mode

tepi:
Deprecates `@HtmlImport` and compatibility mode / bower mode

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/dependency/HtmlImport.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/dependency/HtmlImport.java
@@ -52,11 +52,18 @@ import com.vaadin.flow.shared.ui.LoadMode;
  *
  * @author Vaadin Ltd
  * @since 1.0
+ *
+ * @deprecated Bower WebJars will be deprecated soon, so this functionality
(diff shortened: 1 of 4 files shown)


czp13 (approved):
The code changes and the Javadocs look good to me. I have found some resemblances to Bower in the test classes, but I guess those need to stay until we completely get rid of it.

tepi:
> The code changes and the Javadocs look good to me. I have found some resemblances to Bower in the test classes, but I guess those need to stay until we completely get rid of it.

Yes, there are still tests remaining and also some private methods referring to bower/compatibility mode in flow code. But once all the public references are deprecated we can eventually remote the internal/private things as well without breaking anything.

czp13:
> Yes, there are still tests remaining and also some private methods referring to bower/compatibility mode in flow code. But once all the public references are deprecated we can eventually remote the internal/private things as well without breaking anything.

Thanks, Tepi! This makes 100% sense, let's move incrementally as you said.

I am happy to approve and merge, but of course, we can wait for an actual flow member and not an ex-flow member for visibility :D

mshabarov (approved):

PR #18469, opened 2024-01-16
feat: Include react plugin by default so dev build gets source information

Artur-:
Replaces the two versions of vite.config.ts that Hilla starters have used with the default to avoid having multiple react plugins in use


flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/vite/package.json
```diff
@@ -15,6 +15,7 @@
     "@vitejs/plugin-react-swc": "3.5.0",
```
mshabarov, line 15: `@vitejs/plugin-react-swc` can be now removed, right? As now only `@vitejs/plugin-react` is used in the default Vite config generated by Vaadin.

Artur-: It could but that would break any existing project using plugin-react-swc with custom configuration. Do you think we should remove it?

mshabarov: Well, it was added quite recently here https://github.com/vaadin/flow/pull/18139 and I spot a couple of mentions where react-plugin-swc produced unexpected token errors (in discord and in DX tests). Thus, IMO we can remove it.

Artur-: Removed

mshabarov (approved):

Issue #18410, opened 2024-01-08
Dynamic file creation and subsequent download doesn't work on jetty 12


xottabi4:
### Description of the bug

Hi, recently updated my app to latest versions spring boot 3.2.1 + vaadin flow 24.3.2 + jetty 12.0.5 and functionality of dynamic file creation, for user to download, stopped to work.

If I switch app to tomcat, then everything works fine.

Using com.vaadin.flow.server.InputStreamFactory for file creation.

Seems problem is in method writeModificationTimestamp of class com.vaadin.flow.server.StaticFileServer

Below is full stacktrace:

```
2024-01-08 14:49:49,157 WARN  [com.vaadin.flow.server.StaticFileServer] Error closing URLConnection input stream
java.io.FileNotFoundException: JAR entry META-INF/resources/VAADIN/dynamic/resource/0/c6e3e1e7-4c9e-47d2-ad57-7e392cae56e8/118435686.zip not found in some-app-SNAPSHOT.jar!/BOOT-INF/lib/vaadin-menu-bar-flow-24.3.2.jar
        at org.springframework.boot.loader.net.protocol.jar.JarUrlConnection.lambda$notFoundConnection$1(JarUrlConnection.java:365)
        at org.springframework.boot.loader.net.protocol.jar.JarUrlConnection.throwFileNotFound(JarUrlConnection.java:329)
        at org.springframework.boot.loader.net.protocol.jar.JarUrlConnection.getInputStream(JarUrlConnection.java:181)
        at com.vaadin.flow.server.StaticFileServer.writeModificationTimestamp(StaticFileServer.java:506)
        at com.vaadin.flow.server.StaticFileServer.serveStaticResource(StaticFileServer.java:341)
        at com.vaadin.flow.server.VaadinServlet.serveStaticOrWebJarRequest(VaadinServlet.java:430)
        at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:385)
        at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:106)
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:614)
        at org.eclipse.jetty.ee10.servlet.ServletHolder.handle(ServletHolder.java:736)
        at org.eclipse.jetty.ee10.servlet.ServletHandler$ChainEnd.doFilter(ServletHandler.java:1614)
[...]

PR #18471, opened 2024-01-16
fix: fix static resource retrieval with Jetty 12

tltv:
With Jetty 12, static resources retrieved via servlet context's `getResource(String)` may return non-null URL for non-existing resource. Jetty 11 returns null for non-existing resource. This change adds a double check for the static resource retrieved with the `getResource`, and returns null if it doesn't exist, to keep other parts work the same way as with Jetty 11.

This PR is tested manually. IT for executable JAR will be created in another task/PR.

Fixes: #18410

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
@@ -243,8 +243,7 @@ public class VaadinServletService extends VaadinService {
     @Override
     public URL getStaticResource(String path) {
         try {
-            return VaadinServletService
-                    .getStaticResource(getServlet().getServletContext(), path);
+            return getStaticResource(getServlet().getServletContext(), path);
         } catch (MalformedURLException e) {
             getLogger().warn("Error finding resource for '{}'", path, e);
         }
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/VaadinServletService.java
```diff
@@ -240,7 +243,8 @@ public String resolveResource(String url) {
     @Override
     public URL getStaticResource(String path) {
         try {
-            return getServlet().getServletContext().getResource(path);
+            return VaadinServletService
+                    .getStaticResource(getServlet().getServletContext(), path);
```
caalador, line 247: Minor nit: this can just be:
```suggestion
            return getStaticResource(getServlet().getServletContext(), path);
```
As it's a static method in the same class.

Artur-:
Is this a Jetty bug or is Jetty 12 supposed to work this way? Is there a Jetty issue?

tltv:
> Is this a Jetty bug or is Jetty 12 supposed to work this way? Is there a Jetty issue?
Issue will be created. It looks like contract in JavaDoc for `getResource(String)` is broken as it should return null for non-existing URL. However, change to Jetty 12 that lead to this may not be trivial to fix, so this PR will give time for the possible future fix in Jetty 12.

caalador (approved):

PR #17730, opened 2023-09-28
fix: Be less verbose when Vite is not running

Artur-:
Fixes #17674

PR #18484, opened 2024-01-18
fix: Abort resource handling if dev server resource handling fails

### LINKED ISSUE #17995: Low level servlet request exception if connection is broken while loading a Vite resource
### Description of the bug

1. Create a new app `npx @vaadin/cli init load`
2. Start it `cd load && mvn`
3. Open in the browser and check the url vaadin-router is loaded from, e.g. http://localhost:8080/VAADIN/@fs/some-folder/load/node_modules/@vaadin/router/dist/vaadin-router.js?v=c742f5f9
4. Put a breakpoint in `StaticFileServer.serveStaticResource`
5. Run `curl http://localhost:8080/VAADIN/@fs/some-folder/load/node_modules/@vaadin/router/dist/vaadin-router.js?v=c742f5f9`
6. Once the breakpoint is hit, press ctrl-c for the curl command
7. Let the server continue

Then you see something like
```
2023-11-06T16:06:13.695+02:00 ERROR 81361 --- [io-8080-exec-10] c.vaadin.flow.server.StaticFileServer    : Unable to load /VAADIN/@fs/some-folder/load/node_modules/@vaadin/router/dist/vaadin-router.js from the frontend dev server: java.io.IOException: Broken pipe
2023-11-06T16:06:13.699+02:00 ERROR 81361 --- [io-8080-exec-10] o.a.c.c.C.[.[.[/].[springServlet]        : Servlet.service() for servlet [springServlet] threw exception

java.lang.IllegalStateException: Cannot create a session after the response has been committed
        at org.apache.catalina.connector.Request.doGetSession(Request.java:2907) ~[tomcat-embed-core-10.1.15.jar:10.1.15]
        at org.apache.catalina.connector.Request.getSession(Request.java:2381) ~[tomcat-embed-core-10.1.15.jar:10.1.15]
        at org.apache.catalina.connector.RequestFacade.getSession(RequestFacade.java:639) ~[tomcat-embed-core-10.1.15.jar:10.1.15]
        at jakarta.servlet.http.HttpServletRequestWrapper.getSession(HttpServletRequestWrapper.java:221) ~[tomcat-embed-core-10.1.15.jar:6.0]
```

seems like sometimes the error is instead
```
java.lang.IllegalStateException: Cannot call sendError() after the response has been committed
	at org.apache.catalina.connector.ResponseFacade.checkCommitted(ResponseFacade.java:485) ~[tomcat-embed-core-10.1.15.jar:10.1.15]
	at org.apache.catalina.connector.ResponseFacade.sendError(ResponseFacade.java:329) ~[tomcat-embed-core-10.1.15.jar:10.1.15]
	at jakarta.servlet.http.HttpServletResponseWrapper.sendError(HttpServletResponseWrapper.java:89) ~[tomcat-embed-core-10.1.15.jar:6.0]
	at com.vaadin.flow.server.VaadinService.handleRequest(VaadinService.java:1528) ~[flow-server-24.2.2.jar:24.2.2]
	at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:398) ~[flow-server-24.2.2.jar:24.2.2]
	at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:106) ~[vaadin-spring-24.2.2.jar:na]
```

### Expected behavior

There is nothing in the log. The request for a static file was aborted. Nothing special about that.

As above

### Versions

- Vaadin / Flow version: 24.2.2
- Java version: 21
- OS version: macOs

Artur-:
If the failure comes from writing to the original request (e.g. because client has closed/aborted the connection), then we cannot continue trying to write another response to it. If the failure comes from the connection to the Vite server, we still don't know if it would have served a resource or not so we should abort.

Fixes #17995 

Does not fix the underlying cause for the connection being aborted though but #17995 does not have information about that


mcollovati:
A question: the try/catch was introduced by #17730 to reduce log verbosity. Now we are going to remove it, so basically reverting the previous PR and flooding the logs again when something wrong happens with Vite.

Would it make sense instead to keep the try/catch but instead of continue with the processing, just send an internal server error to the client and return true?

In the catch block we can keep the old log message and potentially add another log at debug or trace level that will report the exception stacktrace.

Artur-:
Makes sense, you mean like this new version?

Artur-:
The problem now is that we don't handle any requests if Vite is not responding properly

Artur-:
But that was probably the case with the previous version also

mcollovati (approved):
The new solution looks good to me.

Also with the previous version, the request stops if Vite is not responding, so the behavior is almost the same.

I would prefer to not completely hide the `IOException`, but log it at trace level for debugging purpose, but this is not a blocker for the PR.

flow-server/src/main/java/com/vaadin/flow/server/StaticFileServer.java
```diff
@@ -273,6 +273,20 @@ public boolean serveStaticResource(HttpServletRequest request,
...
+                        HttpStatusCode.INTERNAL_SERVER_ERROR.getCode(),
+                        "Unable to load " + filenameWithPath
+                                + " from the frontend dev server: "
+                                + e.getMessage());
+            } catch (Exception ee) {
+                // The server might have partly written an output. If so, let's
+                // just go with that
+                getLogger().trace(
+                        "Ignoring exception when writing dev server error response",
+                        ee);
```
mcollovati, line 287: I was thinking to log the trace of the `IOException` thrown by the Vite request to provide a better understanding of what happened.
I wonder if that logging the potential exception raised while we write the error response, may be less useful for troubleshooting

Artur-: I now added both :)

Artur-: While the second one might be less useful, you never know..

mcollovati (approved):

PR #18486, opened 2024-01-19
Allow moving node to another tree after removeFromTree call

### LINKED ISSUE #18390: Cannot move element to a different UI: can't change id once it has been assigned
### Description of the bug

Cannot move element to a different UI after `removeFromTree()`.

`removeFromTree()` succeeds, but after the removed element is added to a different UI (in the same session) it fails with IllegalStateException: Can't change id once it has been assigned
```
 ~[flow-server-24.2.5.jar:24.2.5]
	at com.vaadin.flow.server.communication.ServerRpcHandler.handleInvocationData(ServerRpcHandler.java:459) ~[flow-server-24.2.5.jar:24.2.5]
	at com.vaadin.flow.server.communication.ServerRpcHandler.lambda$handleInvocations$2(ServerRpcHandler.java:440) ~[flow-server-24.2.5.jar:24.2.5]
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511) ~[na:na]
	at com.vaadin.flow.server.communication.ServerRpcHandler.handleInvocations(ServerRpcHandler.java:440) ~[flow-server-24.2.5.jar:24.2.5]
	at com.vaadin.flow.server.communication.ServerRpcHandler.handleRpc(ServerRpcHandler.java:323) ~[flow-server-24.2.5.jar:24.2.5]
```
https://github.com/vaadin/flow/blob/47b643689be3d198b8b1a30e163077da9d9f1e0b/flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java#L836-L837 

This issue seems to be related to https://github.com/vaadin/flow/pull/17695, see also https://vaadin.slack.com/archives/C6X43FE8M/p1695724670429019

As a workaround, `removeFromTree(false)` seems to work, but that method is "meant for internal use only".

### Expected behavior

After `removeFromTree()` I should be able to add the element to a different UI.

1. Open two tabs of the application
2. In the first tab click "Set target"
3. In the second tab click "Send"

Expected behavior: div should be removed from the second tab and added to the first tab
Observed behavior (Vaadin 24.2): an exception is thrown

The same code works as expected in 23.3.21
```
public class HelloWorldView extends VerticalLayout {
 private static VerticalLayout targetView;

  private Div div = new Div();

  public HelloWorldView() {

    add(div);
    addAttachListener(ev -> {
      div.setText("" + UI.getCurrent().getUIId());
    });

    add(new Button("Set target", ev -> targetView = this));

    add(new Button("Send", ev -> {
      if (targetView != null && targetView != this) {
        div.getElement().removeFromTree(false);
        targetView.add(div);
        targetView.getUI().get().push();
        return;
      }
    }));
  }
 }
 ```

### Versions

- Vaadin / Flow version: 24.2.6

### PR DESCRIPTION
**Pull request by tepi**
Enables moving a component to another UI by just calling `Element.removeFromTree()` and then adding the component to the other UI.

Fixes #18390

### CODE REVIEW

File: flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java [hunk 1]
```diff
@@ -793,6 +793,11 @@ private void doSetTree(StateTree tree) {
             } else {
                 id = -1;
             }
+        } else if (id > -1 && getOwner() == NullOwner.get()) {
+            // When id is set but owner is NullOwner, removeFromTree has been
+            // called, so we should allow moving the node to another StateTree
+            id = -1;
+            wasAttached = false;
```
mshabarov, line 800: Not sure 100%, but I would suggest to try `reset(false)` instead of these two lines.
However, it mutates the node's state even more, so might cause some side effect, e.g. if a component is Grid and it's data refresh logic is sensitive to attach/detach, this might be a problem.

tepi: Well that would do no harm, since to get to this state in the first place, `reset(true)` has already been called. Calling `reset(false)` here would effectively just run these two lines, since rest of the state already matches.

tepi: That said, it's probably still safer to call `reset(false)` here, in case the reset logic ever changes.

mshabarov: Agreed, let's replace with `reset(false)` and I'm ready to approve.

tepi: Fixed

mshabarov (approved):

PR #18502, opened 2024-01-22
test: test executable jar with Jetty

tltv:
Adds a new module to run ITs for executable jar build for Spring boot with Jetty.

Related-to: #18410

### DIFF
--- a/flow-tests/vaadin-spring-tests/pom.xml
+++ b/flow-tests/vaadin-spring-tests/pom.xml
@@ -311,6 +311,7 @@
                 <module>test-spring-helpers</module>
                 <module>test-spring-common</module>
                 <module>test-spring-boot</module>
+                <module>test-spring-boot-jar</module>
                 <!-- Disabled as the resource scanner does not work with Undertow
                 <module>test-spring-boot-undertow</module>
                 -->
--- a/flow-tests/vaadin-spring-tests/test-spring-boot-jar/pom.xml
+++ b/flow-tests/vaadin-spring-tests/test-spring-boot-jar/pom.xml
@@ -12,6 +12,7 @@
 	<packaging>jar</packaging>
 	<properties>
 		<maven.deploy.skip>true</maven.deploy.skip>
+		<java.exec>java</java.exec>
 	</properties>
 
 	<dependencies>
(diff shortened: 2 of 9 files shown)


flow-tests/vaadin-spring-tests/test-spring-boot-jar/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+				<artifactId>exec-maven-plugin</artifactId>
+				<version>3.1.1</version>
+				<executions>
+					<execution>
+						<phase>pre-integration-test</phase>
+						<goals>
+							<goal>exec</goal>
+						</goals>
+						<configuration>
+							<executable>java</executable>
```
mcollovati, line 169: What about using a property instead of hard-coding `java`? It might prevent issues with environments where the Java executable is not present on path, or if you have several Java versions and the one on path is not the one you need for the build.

flow-tests/vaadin-spring-tests/test-spring-boot-jar/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+							<async>true</async>
```
mcollovati, line 170: Being the execution async, we should probably find a way to stop the process in `post-integration-test` phase.
I think there's nothing out-of-the-box in Maven.
I recall Spring Actuator has a shutdown endpoint, but we can probably do something similar on our own (a custom servlet or filter or maybe using JMX)

flow-tests/vaadin-spring-tests/test-spring-boot-jar/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+							<asyncDestroyOnShutdown>true</asyncDestroyOnShutdown>
+							<workingDirectory>${project.build.directory}</workingDirectory>
+							<commandlineArgs>-Dgosh.args="--nointeractive -s" -jar vaadin-test-spring-boot-jar-${project.version}.jar</commandlineArgs>
```
mcollovati, line 173: Isn't the `gosh` thing something related only to some OSGi container? It could probably be removed

flow-tests/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/SmokeTestIT.java
```diff
@@ -0,0 +1,81 @@
...
+        if (count == 0) {
+            throw new IllegalStateException(
+                    "URL '" + viewUrl + "' is not available");
+        }
+        URL url = new URL(viewUrl);
+        try {
+            url.openConnection().connect();
+        } catch (IOException exception) {
+            Thread.sleep(1000);
+            waitViewUrl(count - 1);
```
mcollovati, line 78: nit: this will not be a deep stack, but I wonder if it could be better to use a while or for loop, instead of recursion

mcollovati (changes requested):

flow-tests/vaadin-spring-tests/pom.xml
```diff
@@ -311,6 +311,7 @@
                 <module>test-spring-helpers</module>
                 <module>test-spring-common</module>
                 <module>test-spring-boot</module>
+                <module>test-spring-boot-jar</module>
```
knoobie, line 314: should the module name contain jetty?

mcollovati (approved):

flow-tests/vaadin-spring-tests/pom.xml
tltv: I would keep it generic like it's sibling module test-spring-boot for now, which is also jetty module. And get back to renaming when we add same test for e.g. tomcat.

PR #18532, opened 2024-01-25
feat: make it possible to controll clean-frontend

caalador:
Add options object that can be
used to controll what is removed/cleaned
in the clean-frontend goal.
By deault everything is done.

### DIFF
--- a/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/CleanFrontendMojo.java
+++ b/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/CleanFrontendMojo.java
@@ -54,30 +54,117 @@ public class CleanFrontendMojo extends FlowModeAbstractMojo {
     public static final String DEV_DEPENDENCIES = "devDependencies";
     public static final String OVERRIDES = "overrides";
 
+    public class Options {
+        private boolean cleanPackageJson = true;
+
+        private boolean removeDevBundle = true;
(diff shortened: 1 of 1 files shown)


caalador:
The main use case for this is to be able to control cleanup for hilla projects so that specifically the package-lock is not removed.
At this moment hilla plugin doesn't allow any cleaning at all due to the removal of package-lock

mshabarov (commented):
How this supposed to be used?

Soon the Hilla plugin will be wrapped into Vaadin plugin and used always.

I thought that we would need to enable `clean-frontend` in Hilla again, but add an auto-detection that tells if it's a Flow, Hilla or Hybrid app. Then `clean-frontend` mojo/task can run actual cleaning or not run depending of the nature of the application.

caalador:
So as hilla is extending the classes it should in the extending class call the runCleaning(Options)
for instance `runCleaning(new Options().withRemovePackageLock(false));` instead of throwing the exception that it is not supported.

mshabarov:
Yeah, that's how this can be used, thanks. However, the decision on what to clean up should be made in Flow, otherwise Hilla plugin will keep `package-lock.json` always for any application and this would be a step back for pure Flow apps.
This API is generally looks good for me, but the place should be not in Hilla plugin, but in the Flow plugin, who should detect if Hilla views are present in the project.

mshabarov (approved):
I haven't found the Vite generated config file in the list, but this PR isn't about it specifically, so not a blocker.

caalador:
vite.generated.js is not removed as it is always rewritten and overwrites any changes done manually.

PR #18540, opened 2024-01-25
Adds isApplied feature for Binding

tepi:
Allows developer to ignore selected bindings via providing a predicate for a binding that is tested before validating the binding, or writing the binding value to bean. Default implementation ignores bindings for which the bound field is not visible.


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2344,13 +2397,13 @@ private BinderValidationStatus<BEAN> doWriteIfValid(BEAN bean,
...
-        Collection<Binding<BEAN, ?>> currentBindings = new ArrayList<>(
-                bindings);
+        Collection<Binding<BEAN, ?>> currentBindings = bindings.stream()
+                .filter(Binding::isApplied).collect(Collectors.toList());
 
         // First run fields level validation, if no validation errors then
         // update bean
         List<BindingValidationStatus<?>> bindingResults = currentBindings
-                .stream().map(b -> b.validate(false))
+                .stream().filter(Binding::isApplied).map(b -> b.validate(false))
```
mcollovati, line 2406: Probably the `filter` operation here is not needed, since `currentBindings` is already filtered.

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
mcollovati, line 1: Does it make sense to you to add tests to ensure that the `isApplied`  is evaluated during `writeBean()` and `writeBeanAsDraft()`?

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -294,6 +294,44 @@ default BindingValidationStatus<TARGET> validate() {
...
+         * Gets predicate for testing {@link #isApplied()}. By default,
+         * non-visible components are ignored during validation and bean
+         * writing.
+         *
+         * @return predicate for testing {@link #isApplied()}
+         */
+        default SerializablePredicate<Binding<BEAN, TARGET>> getIsAppliedPredicate() {
+            return binding -> {
+                if (binding.getField() instanceof Component) {
+                    return ((Component) binding.getField()).isVisible();
```
mcollovati, line 319: nit: could use pattern matching to simplify the cast operation (unless we plan to back-port the feature)
```suggestion
                if (binding.getField() instanceof Component component) {
                    return component.isVisible();
```

mcollovati (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
tepi: True, good catch! Fixed

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
tepi: Fixed

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
tepi: Added test for these two cases

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -314,13 +314,8 @@ default boolean isApplied() {
...
         default SerializablePredicate<Binding<BEAN, TARGET>> getIsAppliedPredicate() {
-            return binding -> {
-                if (binding.getField() instanceof Component) {
-                    return ((Component) binding.getField()).isVisible();
-                } else {
-                    return true;
-                }
-            };
+            return binding -> !(binding.getField() instanceof Component c)
+                    || c.isVisible();
```
knoobie, line 318: nit: the previous version looked way more readable, this feels like brain training :D

mcollovati (approved):

PR #18546, opened 2024-01-26
fix: no serverSideRoutes check if no Routes

### LINKED ISSUE #18539: Hilla applications without any Flow routes shouldn't require server side routes entry
### Description of the bug

With Vaadin 24.4.0.alpha Hilla applications without Flow views require server side routes in `routes.tsx` at start-up, which doesn't make much sense:

```
To have working Flow routes add the following to the 'routes.tsx' file:
            - import { serverSideRoutes } from "Frontend/generated/flow/Flow";
            - route '...serverSideRoutes' into the routes definition as shown below:
```

A workaround is to add this entry as it suggests.

### Expected behavior

Flow should show this message only when Flow views/routes are actually present in the project.

Can be reproduced with Hilla [integration test](https://github.com/vaadin/hilla/tree/main/packages/java/tests/gradle/single-module) or with [hybrid app](https://github.com/vaadin/flow-hilla-hybrid-example) by removing all Flow views and removing `serverSideRoutes` entry.

### Versions

- Vaadin / Flow version: 24.4.0.alpha2

caalador:
Do not check for the
serverSideRoutes in
routes.tsx in react mode
if no `@Route` annotated
Flow routes found.

Fixes #18539

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
@@ -60,7 +60,6 @@ import static java.nio.charset.StandardCharsets.UTF_8;
  */
 public class TaskGenerateReactFiles implements FallibleCommand {
 
-    private final File frontendDirectory;
     private Options options;
     protected static String NO_IMPORT = """
             Faulty configuration of serverSideRoutes.
@@ -96,11 +95,11 @@ public class TaskGenerateReactFiles implements FallibleCommand {
      */
     TaskGenerateReactFiles(Options options) {
         this.options = options;
-        this.frontendDirectory = options.getFrontendDirectory();
     }
 
     @Override
     public void execute() throws ExecutionFailedException {
+        File frontendDirectory = options.getFrontendDirectory();
         File appTsx = new File(frontendDirectory, "App.tsx");
         File flowTsx = new File(
                 new File(frontendDirectory, FrontendUtils.GENERATED),
@@ -117,11 +116,8 @@ public class TaskGenerateReactFiles implements FallibleCommand {
             } else {
                 String routesContent = FileUtils.readFileToString(routesTsx,
                         UTF_8);
-                Pattern serverImport = Pattern.compile(
-                        "import[\\s\\S]?\\{[\\s\\S]?serverSideRoutes[\\s\\S]?\\}[\\s\\S]?from[\\s\\S]?(\"|'|`)Frontend\\/generated\\/flow\\/Flow\\1;");
-                if (!serverImport.matcher(routesContent).find()
-                        && !options.getClassFinder()
-                                .getAnnotatedClasses(Route.class).isEmpty()) {
+                if (missingServerImport(routesContent)
+                        && serverRoutesAvailable()) {
                     throw new ExecutionFailedException(
                             String.format(NO_IMPORT, routesTsx.getPath()));
                 }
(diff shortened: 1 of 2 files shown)


mshabarov:
Is this fixes https://github.com/vaadin/flow/issues/18539  ?

caalador:
yes

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -93,6 +95,7 @@ public class TaskGenerateReactFiles implements FallibleCommand {
      *            the task options
      */
     TaskGenerateReactFiles(Options options) {
+        this.options = options;
         this.frontendDirectory = options.getFrontendDirectory();
```
mshabarov, line 99: `this.frontendDirectory = options.getFrontendDirectory();` now can be removed as you have class field `options` and can access any parameter.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -116,7 +119,9 @@ public void execute() throws ExecutionFailedException {
                         UTF_8);
                 Pattern serverImport = Pattern.compile(
                         "import[\\s\\S]?\\{[\\s\\S]?serverSideRoutes[\\s\\S]?\\}[\\s\\S]?from[\\s\\S]?(\"|'|`)Frontend\\/generated\\/flow\\/Flow\\1;");
-                if (!serverImport.matcher(routesContent).find()) {
+                if (!serverImport.matcher(routesContent).find()
+                        && !options.getClassFinder()
+                                .getAnnotatedClasses(Route.class).isEmpty()) {
```
mshabarov, line 124: I suggest to simplify this if clause for better readability, e.g.:

```java
if (hasServerRoutes() && missingServerImport()) {
  // existing codes
}

private boolean hasServerRoutes(//params) {
return !options.getClassFinder()
                                .getAnnotatedClasses(Route.class).isEmpty()
}

private boolean missingServerImport(//params) {
return !serverImport.matcher(routesContent).find();
}
```

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateReactFilesTest.java
```diff
@@ -131,6 +138,9 @@ public void routesMissingImport_expectionThrown() throws IOException {
 
         FileUtils.write(routesTsx, content, StandardCharsets.UTF_8);
 
+        Mockito.when(classFinder.getAnnotatedClasses(Route.class)).thenReturn(
+                Collections.singleton(TaskGenerateReactFilesTest.class));
```
mshabarov, line 142: This also needs a test that checks that Flow doesn't throw if import is missing and no Flow routes.

mshabarov, line 142: `TaskGenerateReactFilesTest` is better to replace to e.g. `DummyRoute` private class.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
caalador: Moved

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
caalador: methodised

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateReactFilesTest.java
caalador: added

caalador: done

mshabarov (approved):

Issue #17178, opened 2023-07-04
API for disabling default validators for Binder


vursen:
### Describe your motivation

Binder currently doesn't provide a way to opt-out of the default validator feature. 

In the case of Flow components, the default validator checks component constraints. However, there are situations where the developer would prefer to define constraints on the component for their visual aspects but handle the actual validation manually using a custom validator.

### Describe the solution you'd like

A method to disable `getDefaultValidator` when binding a field to a Binder.

### Describe alternatives you've considered

Flow components don't seem to be a good place for such a public method since the API specifically targets Binder.

### Additional context

Related to https://github.com/vaadin/platform/issues/3066 and https://github.com/vaadin/flow-components/pull/3406#discussion_r909604254

vursen:
The API has been discussed in https://github.com/orgs/vaadin/discussions/4664

vursen:
The JavaDoc needs to be updated to describe that there is such a thing as default validator and that it can be disabled with the proposed API.

mshabarov:
Planned for next sprint in Vaadin Flow team.

PR #18549, opened 2024-01-26
Adds skipDefaultValidator API

tepi:
Provides means to skip default validators of bound fields. Skipping can be done either for the Binder instance (skips all), or per-binding via BindingBuilder.

Fixes https://github.com/vaadin/flow/issues/17178

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -257,28 +257,28 @@ public class Binder<BEAN> implements Serializable {
 
         /**
          * Sets up this binding to either disable or enable the default field
-         * validators (e.g. min/max validators in DatePicker). This
+         * validator (e.g. min/max validators in DatePicker). This
          * binding-level setting will override the Binder-level setting for this
          * property.
          *
          * Defaults to {@literal false}.
          *
-         * @param defaultValidatorsDisabled
-         *            {@literal true} to disable default validators for this
-         *            binding, {@literal false} to enable them, {@literal null}
+         * @param defaultValidatorDisabled
+         *            {@literal true} to disable default validator for this
+         *            binding, {@literal false} to enable it, {@literal null}
          *            to reset (fallback to Binder-level setting)
          * @see Binder#setDefaultValidatorsDisabled(boolean) for faster way to
          *      disable default validators for all bound fields.
          */
-        void setDefaultValidatorsDisabled(boolean defaultValidatorsDisabled);
+        void setDefaultValidatorDisabled(boolean defaultValidatorDisabled);
 
         /**
-         * Returns if default validators of bound field are disabled..
+         * Returns if default validator of bound field is disabled.
          *
-         * @return {@literal true} if default validators are disabled for this
-         *         binding, {@literal false} if they are enabled
+         * @return {@literal true} if default validator is disabled for this
+         *         binding, {@literal false} if it is enabled
          */
-        boolean isDefaultValidatorsDisabled();
+        boolean isDefaultValidatorDisabled();
 
         /**
          * Define whether the value should be converted back to the presentation
@@ -851,22 +851,22 @@ public class Binder<BEAN> implements Serializable {
 
         /**
          * Sets up this binding to either disable or enable the default field
-         * validators (e.g. min/max validators in DatePicker). This
+         * validator (e.g. min/max validators in DatePicker). This
          * binding-level setting will override the Binder-level setting for this
          * property.
          *
          * Defaults to {@literal false}.
          *
-         * @param defaultValidatorsDisabled
-         *            {@literal true} to disable default validators for this
-         *            binding, {@literal false} to enable them, {@literal null}
+         * @param defaultValidatorDisabled
+         *            {@literal true} to disable default validator for this
+         *            binding, {@literal false} to enable it, {@literal null}
          *            to reset (fallback to Binder-level setting)
          * @return this binding, for chaining
          * @see Binder#setDefaultValidatorsDisabled(boolean) for faster way to
          *      disable default validators for all bound fields.
          */
-        BindingBuilder<BEAN, TARGET> setDefaultValidatorsDisabled(
-                boolean defaultValidatorsDisabled);
+        BindingBuilder<BEAN, TARGET> setDefaultValidatorDisabled(
+                boolean defaultValidatorDisabled);
 
         /**
          * Sets the field to be required. This means two things:
@@ -982,7 +982,7 @@ public class Binder<BEAN> implements Serializable {
 
         private boolean asRequiredSet;
 
-        private Boolean defaultValidatorsDisabled;
+        private Boolean defaultValidatorDisabled;
 
         /**
          * Creates a new binding builder associated with the given field.
@@ -1009,7 +1009,7 @@ public class Binder<BEAN> implements Serializable {
 
             if (field instanceof HasValidator hasValidator) {
                 withValidator(
-                        (val, ctx) -> binding.isDefaultValidatorsDisabled()
+                        (val, ctx) -> binding.isDefaultValidatorDisabled()
                                 ? ValidationResult.ok()
                                 : hasValidator.getDefaultValidator().apply(val,
                                         ctx));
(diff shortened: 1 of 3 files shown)


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2709,6 +2751,14 @@ private List<ValidationResult> validateBean(BEAN bean) {
                         Collections::unmodifiableList));
     }
 
+    /**
+     * Tells Binder to skip executing the default validators (e.g. min/max
+     * validators in DatePicker) of all bound fields by default.
+     */
+    public void skipDefaultValidators() {
```
knoobie, line 2758: for consistence with the other fields I'm kinda missing setSkipDefaultValidators(boolean) 

Otherwise it's also impossible to-renable the validators

tepi: Noted. We discussed this internally earlier today and concluded to keep it as simple as possible, so (at least for now) you can do one-time setting of skip for either whole binder or individual bindings.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -255,6 +255,13 @@ default BindingValidationStatus<TARGET> validate() {
          */
         public boolean isValidatorsDisabled();
 
+        /**
+         * Returns if default validator of bound field is skipped or not.
+         *
+         * @return A boolean value.
+         */
+        boolean isSkipDefaultValidator();
```
vursen, line 263: API suggestion:

```suggestion
        boolean isDefaultValidatorSkipped();
```

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
mshabarov: @knoobie Adding this method is possible of course, but I don't understand when exactly this would be needed to change this behaviour dynamically. Swapping this for the whole binder / form sounds as too much heavy logic. I would rather recommend to put the business logic into custom Binder's validators if one really needs to change validation logic on-the-fly. However, I may miss an experience of using Binder in more advanced cases, so could you please share your thoughts how `setSkipDefaultValidators(boolean)` could be used?

knoobie: Only thing that I could think of would be highly dynamic forms or even form builder that do things based on given objects and re-create all bindings.. but not sure how widely that's used. 

I'm probably just biased because I Like side-effect free code.. or let's say, code that can be changed back once I messed-up.. now this method is a once used and it's impossible to change back (thinking unit tests for the binder for example) and it does not match all other flags the binder has, where it's possible to change them at "runtime" making it inconsistent to the API design. Personally I don't think there is much logic to consider by just replacing the direct access to the boolean with the getter from it, ensuring you are always getting the correct value the user supplied last - or do I miss something? 

TL;DR: I'm not seeing the benefits of an inconsistent API

mshabarov: We discussed with @tepi on daily and it makes sense for a sake in consistent API to add `setSkipDefaultValidators(boolean)` methods as you suggested.

knoobie: Thanks for reconsidering!

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -1247,6 +1292,8 @@ public BindingImpl(BindingBuilderImpl<BEAN, FIELDVALUE, TARGET> builder,
             this.asRequiredSet = builder.asRequiredSet;
             converterValidatorChain = ((Converter<FIELDVALUE, TARGET>) builder.converterValidatorChain);
 
+            defaultValidatorsSkipped = builder.defaultValidatorsSkipped;
```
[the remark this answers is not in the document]
knoobie: expectation: yes, it should be skipped - perfect example for a unit test :)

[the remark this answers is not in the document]
mshabarov: My expectation is either yes, need to be tested.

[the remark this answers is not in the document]
tepi: Yes, changing either the Binder's or Binding's property can be done at anytime. The final value is evaluated at validation time. BindingBuilder's value can not be changed after the binding is built, but it does not matter since it is used only as the initial value.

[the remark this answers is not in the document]
mshabarov: Good! Could you add one more unit test to cover this?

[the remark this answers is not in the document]
tepi: I'll modify the unit tests to test dynamic property changing and validation as well. Thanks for the input!

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -255,6 +255,14 @@ default BindingValidationStatus<TARGET> validate() {
          */
         public boolean isValidatorsDisabled();
 
+        /**
+         * Returns if default validator of bound field is skipped or not.
+         *
+         * @return {@literal true} to skip default validators for this binding,
+         *         {@literal false} to evaluate them
+         */
+        boolean isDefaultValidatorsSkipped();
```
vursen, line 264: I'm concerned about the two 's' in the new version of the method names. It negatively affects their readability. Also, setSomethingSkipped naming doesn't appear to be common for the existing code base.

tepi: Good point. Maybe we could go for something more like this existing API:

knoobie: I don't mind a name change, but it's probably important that Default stays in there so that people don't think this would disable all validations.. which could also be interesting for people that wanna store an un-validated draft of a form.. but that's another story..

tepi: Changed names in API to `setDefaultValidatorsDisabled` etc.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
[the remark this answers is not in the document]
tepi: Added unit tests for dynamically toggling enabled/disabled state

mshabarov (approved):

mshabarov:
Validation fails because of javadoc error:
```
/home/runner/work/flow/flow/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java:270: error: invalid use of @return
Error:           * @return this binding, for chaining
```

mshabarov (approved):

mshabarov:
@vursen I'm ready to merge it, but waiting for your feedback, does it look good to Design System team ?

mshabarov (approved):

vursen:
Could we consider making the method name singular for the Binder instance as well?

tepi:
> Could we consider making the method name singular for the Binder instance as well?

Then it might kind of clash with this existing API:
```
    /**
     * Control whether validators including bean level validators are disabled
     * or enabled globally for this Binder.
     *
     * @param validatorsDisabled
     *            Boolean value.
     */
    public void setValidatorsDisabled(boolean validatorsDisabled) {
        this.validatorsDisabled = validatorsDisabled;
    }
```

Now that I think of it, might have been better to call these `setValidationDisabled` and `setDefaultValidationDisabled` from the beginning, avoiding the singular/plural issue.

mshabarov:
> Should we use the singular form for the method name on the Binder instance too?

I personally see no problems with plurals/singulars with the current implementation: on Binder level you configure `validatorS` and on the Binding level you configure `validator`. Also, I'd keep words `default` and `validator(s)` in the method's names, cause they properly reflect the semantics and coupled to existing getDefaultValidator method.

PR #18550, opened 2024-01-26
fix: change SBOM maven goals lifecycle phase

sosa-vaadin:
## Description

- Change the lifecycle phase of sbom generator maven goals from `generate-resources` to `prepare-resources` so that it matches the `prepare-frontend` maven goal.
- Flip a boolean expression so that the `package.json` file is created if it doesn't exist when running the `prepare-frontend` maven goal.

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -150,8 +150,7 @@ public class BuildFrontendUtil {
                 .withBuildDirectory(adapter.buildFolder())
                 .withJarFrontendResourcesFolder(
                         getJarFrontendResourcesFolder(adapter))
-                .createMissingPackageJson(
-                        !new File(adapter.npmFolder(), PACKAGE_JSON).exists())
+                .createMissingPackageJson(true)
                 .enableImportsUpdate(false).enablePackagesUpdate(false)
                 .withRunNpmInstall(false)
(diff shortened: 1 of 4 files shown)


tepi:
With the latest changes, `com.vaadin.flow.server.frontend.Options#createMissingPackageJson` is always called with `true` parameter. Should we remove the whole method?

tepi:
Ok, I'll double-check it. I don't think we're using Options class from any other project either.

tepi:
We also have this interesting snippet here: https://github.com/vaadin/flow/blob/dc68b28938f08369796061f45b1a1b885d56d548/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java#L235-L239

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
```diff
@@ -150,10 +150,8 @@ public static void prepareFrontend(PluginAdapterBase adapter)
                 .withBuildDirectory(adapter.buildFolder())
                 .withJarFrontendResourcesFolder(
                         getJarFrontendResourcesFolder(adapter))
-                .createMissingPackageJson(
-                        new File(adapter.npmFolder(), PACKAGE_JSON).exists())
```
mshabarov, line 154: I think this was an incorrect leftover in some old bugfix.
Setting `true` here looks safe to me.

mshabarov (approved):
Looks reasonable to me to not run SBOM generate goal earlier than prepare-frontend goal, as the latter prepares needed frontend meta data for SBOM.

PR #18552, opened 2024-01-26
Close streams fetched from data-providers

### LINKED ISSUE #18551: Streams fetched by data-providers are never closed
### Description of the bug

Let’s say I have a backend data-provider for a stream of strings read from a `BufferedReader`, e.g.
```
public Stream<String> fetchFromBackend(Query<String> query) {
    var reader = createBufferedReader();
    return reader.lines().skip(query.getOffset()).limit(query.getLimit()).onClose(reader::close);
}
```
In this case I want to be sure to close the reader when the stream reaches a terminal operation, so I do `onClose(reader::close)`. Problem is, when the `DataCommunicator` gets that stream:
https://github.com/vaadin/flow/blob/b8a992acf42002629a932d98082af8890d32af0a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java#L1055
then it is never closed. This also applies to the `DataView` API, e.g.
https://github.com/vaadin/flow/blob/b8a992acf42002629a932d98082af8890d32af0a/flow-data/src/main/java/com/vaadin/flow/data/provider/AbstractDataView.java#L153-L155

### Expected behavior

Upon calling terminal operations, the stream is then closed.

See code snippet in the description above.

### Versions

- Vaadin / Flow version: 24.4
- Java version: 21
- OS version: macOS 14.2.1

heruan:
This is an attempt to fix #18551 for the `DataCommunicator` API by handling the streams fetched from data-providers inside try-with-resources blocks.

Unclosed streams might still leak from the `DataView` API, that should be fixed in each component using that API.


mcollovati:
The change looks good to me :+1: 
Two points we should probably consider:

* Updating `DataView` javadoc to suggest the user to use the returned stream in combination with try-with-resource to prevent leaking resources (as noted in the PR description). Possibly adding a simple code example.
* Make `HierarchicalDataCommunicator` work consistently, by applying try-with-resource where possible also in `HierarchyMapper` (e.g. `getFlatChildrenStream()`) and `HierarchicalCommunicationController` (e.g. `activate` method).

heruan:
> * Updating `DataView` javadoc to suggest the user to use the returned stream in combination with try-with-resource to prevent leaking resources (as noted in the PR description). Possibly adding a simple code example.

Updated the JavaDoc with the recommendation and code example.

> * Make `HierarchicalDataCommunicator` work consistently, by applying try-with-resource where possible also in `HierarchyMapper` (e.g. `getFlatChildrenStream()`) and `HierarchicalCommunicationController` (e.g. `activate` method).

Closing streams also for hierarchical communicators and mapper now. I'm just having a hard time understanding how unit tests work for the hierarchical API, I added one covering `HierarchicalCommunicationController::activate` and `HierarchyMapper::getChildrenStream` but I couldn't hit `HierarchyMapper::getFlatChildrenStream`.

mcollovati:
@heruan great work!
I think we can test `HierarchyMapper::getFlatChildrenStream` indirectly in `HierarchyMapperWithDataTest` by calling `HierarchyMapper.fetchHierarchyItems()`

Something like
```java
    @Test
    public void fetchHierarchyItems_streamIsClosed() {
        AtomicBoolean streamIsClosed = new AtomicBoolean();
        mapper = new HierarchyMapper<>(new TreeDataProvider<>(data) {
            @Override
            public Stream<Node> fetchChildren(HierarchicalQuery<Node, SerializablePredicate<Node>> query) {
                return super.fetchChildren(query)
                        .onClose(() -> streamIsClosed.set(true));
            }
        });
        Node rootNode = testData.get(0);
        mapper.expand(rootNode);
        mapper.fetchHierarchyItems(rootNode, Range.between(0, 10)).count();

        Assert.assertTrue(streamIsClosed.get());
    }
```

heruan:
> @heruan great work! I think we can test `HierarchyMapper::getFlatChildrenStream` indirectly in `HierarchyMapperWithDataTest` by calling `HierarchyMapper.fetchHierarchyItems()`

Oh there it is, thanks! I added two tests to `HierarchyMapperWithDataTest` for consistency, one for `getFlatChildrenStream` and one for `getChildrenStream`.

mcollovati (approved):

PR #18553, opened 2024-01-29
feat: add raw html table and related components

### LINKED ISSUE #15475: Add html table component et al
### Describe your motivation

Using html table (et al) would sometimes make more sense than to use e.g. Grid component. For example much more flexible when styling. As we support raw html table as well like we do support UnorderedList, H1 etc.

There should be Table,Tr,Td,THead and TBody components in flow-html-components. Names of the components should probably be something more meaningful than in html (e.g. `TableCell` instead of `Td`);

Keep using add-ons.

### Additional context

The [Html Table add-on ](https://vaadin.com/directory/component/html-table) by @stefanuebe could be the basis for official components. See also [BeanTable](https://vaadin.com/directory/component/beantable) by @TatuLund that use <table> element to render DTOs.

sosa-vaadin:
## Description

Add component for `<table>` element and related components:
- `<caption>`
- `<thead>`
- `<tbody>`
- `<tfoot>`
- `<tr>`
- `<th>`
- `<td>`

Fixes #15475


mshabarov:
Before Flow team starts making the review, I'd like to ping @rolfsmeds and @yuriy-fix to ask if they have immediate complains against this feature in general, e.g. naming, design or the whole idea about having this component on Flow side.

tepi:
Not sure if I'm getting something wrong but seems to me this patch does not really provide any added value in terms of convenience APIs and testing facilities. Only convenience API I could find is adding some components to here and there within the table or its sub-components.

For example, comparing this to an add-on that was mentioned in the linked issue: [`HtmlTable` in this patch](https://raw.githubusercontent.com/vaadin/flow/128289c54597485cdfe4b23e0bc927d474e969d2/flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlTable.java) vs [`Table` in add-on](https://raw.githubusercontent.com/stefanuebe/vaadin-html-table/master/addon/src/main/java/org/vaadin/stefan/table/Table.java).

Also, the testbench elements for Table and its parts are all empty. 

I have a hard time what value this brings as such? Saves you from typing the tag names when creating the elements?

sosa-vaadin:
I focused on just adding bare bones components. I did not check the add-on and only based the components on other already existing html flow components. Since no API requirements or specifications were given in the issue (which I found only after I was done coding what is now in the PR) I submitted it as is.

I thought of adding a few convenience methods, but decided against it.

yuriy-fix:
I don't have any objections to proceeding with this contribution. My only suggestion would be to opt for `Native` prefix instead of `Html` for better alignment. I also agree with @tepi, as I don't see real benefit in providing empty classes without methods.

Regarding the component name `Table` (without prefixes), we would like to preserve it for the DS.

sosa-vaadin:
I've added a bunch of methods to match most of the features provided in the add-on. I still need to write tests for those so once I'm done (next CF) I will change it back from draft.

AlainaFaisal (commented):
The code logic seems sound and I could not find any discrepancies.

tepi (approved):

PR #18569, opened 2024-01-30
fix: Uncaught SyntaxError: Unexpected identifier 'RefreshRuntime'

Artur-:
Newlines cannot be removed from JavaScript as they work as ;

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
@@ -573,8 +573,8 @@ public class WebComponentBootstrapHandler extends BootstrapHandler {
                 .replace("//<![CDATA[", "/*<![CDATA[*/")
                 .replace("//]]>", "/*]]>*/")
                 // get rid of all the unnecessary white-space
-                .replaceAll("\\s{2,}", "").replace("\t", "")
-                .replace("\n", "\\n").replace("\r", "");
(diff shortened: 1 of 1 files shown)


caalador (approved):

caalador (changes requested):
This actually doesn't seem to add new lines as I still get the same exception on the console and see that the output for inline is gotten as a single line.

flow-server/src/main/java/com/vaadin/flow/server/communication/WebComponentBootstrapHandler.java
```diff
@@ -573,8 +573,8 @@ private static String inlineHTML(String html) {
                 .replace("//<![CDATA[", "/*<![CDATA[*/")
                 .replace("//]]>", "/*]]>*/")
                 // get rid of all the unnecessary white-space
-                .replaceAll("\\s{2,}", "").replace("\t", "").replace("\n", "")
-                .replace("\r", "");
+                .replaceAll("\\s{2,}", "").replace("\t", "")
+                .replace("\n", "\\n").replace("\r", "");
```
caalador, line 577:
```suggestion
                .replace("\r", "").replaceAll("\\s{2,}", "")
                .replace("\t", "").replace("\n", "\\n");
```
Needs to be changed so that any \r is removed first as on windows its \r\n for new line and that falls into the \s for 2 characters or more

caalador (changes requested):

caalador (approved):

Issue #18505, opened 2024-01-22
Auto-detect if React or Lit is used in a project


mshabarov:
### Describe your motivation

Flow should auto-detect if React or Lit is used in a project to automatically set the flag `reactRouterEnabled` to true or false, which in turn will use either Lit deps or React deps and `vaadin-router` or `react-router`.

### Describe the solution you'd like

Detection is based on the files:
- if `routes.tsx` is present and `index.ts` imports the `App.tsx` or `routes.tsx`, then the `reactRouterEnabled` should be true.
- if `index.ts` is present and it imports vaadin-router, i.e. `import {Router} from '@vaadin/router';`, then `reactRouterEnabled` should be false.
- if no `routes.tsx` and `index.ts` are present, then the `reactRouterEnabled` should be true, meaning that this might be a plain Flow application.

### Additional context

Vaadin/Hilla 24.4

PR #18571, opened 2024-01-30
feat: Auto-detect if react-router or vaadin-router should be used

mshabarov:
If no value for `enable.react.router` parameter is set, looks into `frontend/index.ts` file content for `@vaadin/router` imports. If it presents, when sets the config parameter to false. In all other cases sets it to true. 

Fixes https://github.com/vaadin/flow/issues/18505

### DIFF
--- a/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
+++ b/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
@@ -161,8 +161,8 @@ public class BuildDevBundleMojo extends AbstractMojo
     @Parameter(property = "npm.postinstallPackages", defaultValue = "")
     private List<String> postinstallPackages;
 
-    @Parameter(property = InitParameters.REACT_ROUTER_ENABLED, defaultValue = "true")
-    private boolean reactRouterEnabled;
+    @Parameter(property = InitParameters.REACT_ROUTER_ENABLED, defaultValue = "${null}")
+    private String reactRouterEnabled;
 
     @Override
     public void execute() throws MojoFailureException {
@@ -435,7 +435,7 @@ public class BuildDevBundleMojo extends AbstractMojo
     }
 
     @Override
-    public boolean isReactRouterEnabled() {
+    public String isReactRouterEnabled() {
         return reactRouterEnabled;
     }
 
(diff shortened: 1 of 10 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -305,6 +308,9 @@ public class FrontendUtils {
 
     public static final String BRIGHT_BLUE = "\u001b[94m%s\u001b[0m";
 
+    public static final Pattern IMPORT_VAADIN_ROUTER = Pattern.compile(
+            "import[\\s\\S]+\\{[\\s\\S]+}[\\s\\S]+from[\\s\\S]+([\"'`])@vaadin/router\\1;");
+
```
knoobie, line 313: I know you guys love your regex(s) but isn't it enough to check for `file.contains("@vaadin/router")` to simplify this and all possible ways people could have imported it.. I would highly anticipate that a file containing that string uses the router in some way, making the detection a little bit easier

flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
```diff
@@ -436,7 +436,10 @@ public boolean isPrepareFrontendCacheDisabled() {
 
     @Override
     public boolean isReactRouterEnabled() {
-        return reactRouterEnabled;
+        if (reactRouterEnabled != null) {
+            return reactRouterEnabled;
+        }
+        return FrontendUtils.isReactRouterRequired(frontendDirectory());
```
caalador, line 442: For the devBundle plugin mojo we should always just do the react build as that is the default and the usage is only internal. 
So this does not need the change at and should just run with the default value.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
caalador: Sounds like it would be much better to just use contains for `@vaadin/router` only thin I would perhaps run the string through first is `StringUtil.removeComments(indexTsContent)` to not mistakenly find a mention in a comment block.

caalador (changes requested):
This PR should remove all 
```
<configuration>
  <reactRouterEnabled>false</reactRouterEnabled>
</configuration>
```
blocks from test modules as that will also work as a validation that the correct mode is chosen.

flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
mshabarov: All the changes in this file are reverted. So it's true by default.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
mshabarov: Thank you guys, good comment, I was thinking about it at the beginning. Changed.

caalador (approved):

Issue #18564, opened 2024-01-30
Include Hilla components automatically depending on React/Lit


mshabarov:
### Describe your motivation

Hilla project has `hilla-react` and `hilla-lit` that basically differ in what they import with `AtNpmPackage` annotation, e.g. [CRUD component](https://github.com/vaadin/hilla/blob/main/packages/java/hilla-react/src/main/java/com/vaadin/hilla/HillaReactCrud.java).

This mean that one need to exclude `hilla-react` artifact from `vaadin` and add `hilla-lit` artifact to have a proper component in node_modules.

But this can be simplified and avoided by adding these components into [pre-filled package.json](https://github.com/vaadin/flow/tree/main/flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies) on Flow side. Then Flow based on `reactRouterEnabled` flag chooses a proper package.json file and that's it (this works like this already).

### Describe the solution you'd like

There should be only one artifact `hilla` that depends on `endpoints` JAR and what else needed, but doesn't contain `NpmPackage` annotations for components. These should be added by Flow automatically, dependening on React or Lit is used in the project. If Hilla isn't used at all, none of these packages should be added by Flow.

### Additional context

Vaadin 24.4.

PR #18597, opened 2024-02-01
feat: Use React/Lit auto-detection to add Hilla components

mshabarov:
Adds Hilla components to `package.json` automatically in build time, so that developers don't need to swap `hilla-react` and `hilla-lit` artifacts, like this

```xml
<dependency>
   <groupId>com.vaadin</groupId>
   <artifactId>vaadin-spring-boot-starter</artifactId>
   <version>${vaadin.version}</version>
   <exclusions>
       <exclusion>
            <groupId>com.vaadin.hilla</groupId>
            <artifactId>hilla-react</artifactId>
        </exclusion>
      </exclusions>
</dependency>
<dependency>
      <groupId>com.vaadin.hilla</groupId>
      <artifactId>hilla-lit</artifactId>
      <version>${vaadin.version}</version>
</dependency>
```

Related-to https://github.com/vaadin/flow/issues/18564, https://github.com/vaadin/hilla/pull/2015

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -176,8 +176,8 @@ public abstract class NodeUpdater implements FallibleCommand {
             VersionsJsonConverter convert = new VersionsJsonConverter(
                     Json.parse(
                             IOUtils.toString(content, StandardCharsets.UTF_8)),
-                    options.isReactRouterEnabled()
-                            && EndpointRequestUtil.isHillaAvailable());
+                    options.isReactRouterEnabled() && FrontendUtils
+                            .isHillaUsed(options.getFrontendDirectory()));
             versionsJson = convert.getConvertedJson();
             versionsJson = new VersionsJsonFilter(getPackageJson(),
                     DEPENDENCIES)
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -575,10 +575,21 @@ public abstract class NodeUpdater implements FallibleCommand {
         return versionsJson;
     }
 
+    /**
+     * Adds Hilla components to package.json if Hilla is used in the project.
+     *
+     * @see <a href=
+     *      "https://github.com/vaadin/hilla/tree/main/packages/java/hilla/src/main/resources/com/vaadin/flow/server/frontend/dependencies/hilla/components</a>
+     * @param dependencies
+     *            to be added into package.json
+     * @param id
+     *            the relative path location of template files
+     */
     private static void putHillaComponentsDependencies(
             Map<String, String> dependencies, String id) {
         if (EndpointRequestUtil.isHillaAvailable()) {
-            dependencies.putAll(readDependencies(id, "dependencies"));
+            dependencies.putAll(
+                    readDependencies("hilla/components/" + id, "dependencies"));
         }
     }
 }
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -37,7 +37,6 @@ import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.vaadin.experimental.FeatureFlags;
 import com.vaadin.flow.internal.hilla.EndpointRequestUtil;
 import com.vaadin.flow.server.Constants;
 import com.vaadin.flow.server.frontend.scanner.ClassFinder;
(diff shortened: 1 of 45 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -573,4 +574,22 @@ private JsonObject generateVersionsFromPackageJson(JsonObject packageJson)
...
+     *
+     * @see <a href=
+     *      "https://github.com/vaadin/hilla/tree/main/packages/java/hilla/src/main/resources/com/vaadin/flow/server/frontend/dependencies/hilla/components</a>
+     * @param dependencies
+     *            to be added into package.json
+     * @param id
+     *            the relative path location of template files
+     */
+    private static void putHillaComponentsDependencies(
+            Map<String, String> dependencies, String id) {
```
caalador, line 589: Should this be extended with `String packageJsonKey` so that you can use it to also add dev dependencies from `getDefaultDevDependencies()` if any set?

flow-server/src/test/resources/com/vaadin/flow/server/frontend/dependencies/hilla/components/react/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@vaadin/hilla-react-auth": "24.4.0-alpha2",
+    "@vaadin/hilla-react-crud": "24.4.0-alpha2",
+    "@vaadin/hilla-react-form": "24.4.0-alpha2",
+    "@vaadin/hilla-lit-form": "24.4.0-alpha2"
```
caalador, line 15: Should the lit form be with the react components?
Looks strange that get react has both lit and react form, also the test now doesn't say that lit was not read with react, only that react was not read with lit.

mshabarov: `hilla-lit-form` exports some types for these Hilla react components.
To me this also looks strange, but this is how Hilla components are designed.
We can discuss this with @platosha .

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
mshabarov: Yes, it makes sense. Will add.

flow-server/src/test/resources/com/vaadin/flow/server/frontend/dependencies/hilla/components/react/package.json
caalador: for testing it doesn't matter what the actual content is. Test should just be clear that what is tested and that it works.
Main beef here is that it's not clear as react and lit have the same common package for both.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
caalador: Will we add the devDependencies adding here or will you create a ticket?

mshabarov: Yes, added it here. And also added new Hilla views detection method.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
```diff
@@ -467,6 +470,88 @@ public void testGetPlatformPinnedDependencies_VaadinAndVaadinCoreVersionsArePres
         Assert.assertTrue(pinnedVersions.hasKey("@vaadin/vaadin-grid-pro"));
     }
 
+    @Test
+    public void getDefaultDependencies_reactRouterIsUsed_addsHillaReactComponents() {
+        boolean reactRouterEnabled = options.isReactRouterEnabled();
+        try (MockedStatic<EndpointRequestUtil> mock = Mockito
+                .mockStatic(EndpointRequestUtil.class)) {
+            mock.when(EndpointRequestUtil::isHillaAvailable).thenReturn(true);
```
caalador, line 478: This is not checked anymore and the test fails as nothing is added.
```
Error:  Failures: 
Error:    NodeUpdaterTest.getDefaultDependencies_reactRouterIsUsed_addsHillaReactComponents:485 React component should be added when react-router is used
Error:    NodeUpdaterTest.getDefaultDependencies_vaadinRouterIsUsed_addsHillaLitComponents:515 Lit component should be when vaadin-router is used
[INFO] 
Error:  Tests run: 3595, Failures: 2, Errors: 0, Skipped: 10
```

mshabarov: Yes, it needs mocking for the new util method.

mshabarov: Fixed.

Issue #18401, opened 2024-01-05
Use custom Flow components in Hilla views


mshabarov:
### Describe your motivation

Hilla framework gives you an opportunity to develop UI using Lit or React libraries using corresponding Vaadin components: [React](https://github.com/vaadin/react-components) or [Lit](https://github.com/vaadin/web-components)-based.

For unification reasons, Vaadin should also allow to use custom Flow components on the Hilla views, like follows (example for Lit):

```javascript
@customElement('hilla-view')
export class HillaView extends LitElement {

  render() {
    return html`
      <vaadin-vertical-layout>
          <h1>Hilla view</h1>
          <login-form userlbl='Username' pwdlbl='Password'>Login Form</login-form>
      </vaadin-vertical-layout>
    `;
  }
}
```
Where the `login-form` is a Flow server-side component:
```java
public class LoginForm extends Div {
    private TextField userName = new TextField();
    private PasswordField password = new PasswordField();
    // implementation omitted
}
```

### Describe the solution you'd like

The Flow component can be exported using `WebComponentExporter` API:
```java
public class LoginFormExporter
        extends WebComponentExporter<LoginForm> {
    public LoginFormExporter() {
        super("login-form");
        addProperty("userlbl", "")
                .onChange(LoginForm::setUserNameLabel);
        addProperty("pwdlbl", "")
                .onChange(LoginForm::setPasswordLabel);
    }
}
```
and adding the web-component's script into `index.ts`:

```
const script = document.createElement('script');
script.src = '/web-component/login-form.js';
document.head.appendChild(script);
```

However, this doesn't work out-of-the-box with both Lit and React, the following errors are shown up in the browser console preventing Hilla to render the component:
```
VM475:1 Uncaught TypeError: window.Vaadin.Flow.initApplication is not a function
    at <anonymous>:1:8053
    at <anonymous>:1:8098
    at web-component-bootstrap.js?url=http://localhost:8080/web-component/web-component-bootstrap.js:6:8583
    at web-component-bootstrap.js?url=http://localhost:8080/web-component/web-component-bootstrap.js:10:3
(anonymous) @ VM475:1
(anonymous) @ VM475:1
(anonymous) @ web-component-bootstrap.js?url=http://localhost:8080/web-component/web-component-bootstrap.js:6
(anonymous) @ web-component-bootstrap.js?url=http://localhost:8080/web-component/web-component-bootstrap.js:10

[...]

PR #18601, opened 2024-02-02
Feat: add loadComponentScript method

caalador:
Add the loadComponentScript method
to Flow.tsx to be used when
WebComponents are loaded for
Hilla views.

part of #18401

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -353,3 +353,21 @@ export const serverSideRoutes = [
         }
     });
 })();
+
+
+/**
+ * Load the script for an exported WebComponent with the given tag
+ *
+ * @param tag name of the exported web-component to load
+ */
+export const loadComponentScript = (tag: String) => {
+    useEffect(() => {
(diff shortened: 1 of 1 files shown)


mshabarov:
Can we merge this function and `React.createElement(LOGIN_COMPONENT);`, so to have them both incapsulated in 
```
function MyLoginView() {
    return /*create and load the component*/;
}
```
?

caalador:
Yes we can have it as
```
import {VerticalLayout} from "@vaadin/react-components/VerticalLayout.js";
import React, {useEffect} from "react";
import { loadComponentScript } from "Frontend/generated/flow/Flow";

function MyLoginView() {
    const LOGIN_COMPONENT = "my-flow-component";
    loadComponentScript(LOGIN_COMPONENT);
    return React.createElement(LOGIN_COMPONENT);
}

export default function HillaView() {
    return (
        <>
            <VerticalLayout className={'centered-content'}>
                <h3>Hilla View</h3>
                <MyLoginView/>
            </VerticalLayout>
        </>
    );
}
```

mshabarov:
Ok, let's leave it as is, because `React.createElement` may have a more advanced usage (e.g. with attributes), so better to leave this part for developers.

PR #18604, opened 2024-02-02
feat: create and load WebComponent

caalador:
Add method to create and
load the WebComponent using
one method if nothing
specific wanted or needed
for createElement.

Part of #18401


caalador:
With this we can now simplify the adding of a WebComponent to:
```
import {VerticalLayout} from "@vaadin/react-components/VerticalLayout.js";
import { createWebComponent } from "Frontend/generated/flow/Flow";

function MyLoginView() {
    return createWebComponent("my-flow-component");
}

export default function HillaView() {
    return (
        <>
            <VerticalLayout className={'centered-content'}>
                <h3>Hilla View</h3>
                <MyLoginView/>
            </VerticalLayout>
        </>
    );
}
```

mshabarov (approved):

Issue #18608, opened 2024-02-02
Binder throws NPE when binding field after setting bean


sissbruecker:
### Description of the bug

Binding a field after setting a bean throws the following exception:
```
com.vaadin.flow.data.binder.BindingException: An exception has been thrown inside binding logic for the field element [invalid='false', value='foo']
	at com.vaadin.flow.data.binder.Binder$BindingImpl.execute(Binder.java:1670)
	at com.vaadin.flow.data.binder.Binder$BindingImpl.doConversion(Binder.java:1399)
	at com.vaadin.flow.data.binder.Binder$BindingImpl.lambda$initFieldValue$1(Binder.java:1460)
	at com.vaadin.flow.data.binder.Binder$BindingImpl.execute(Binder.java:1656)
	at com.vaadin.flow.data.binder.Binder$BindingImpl.initFieldValue(Binder.java:1455)
	at com.vaadin.flow.data.binder.Binder$BindingBuilderImpl.bind(Binder.java:1034)
	at com.vaadin.flow.data.binder.Binder.bind(Binder.java:2047)
	at com.vaadin.flow.component.grid.it.UpdateEditorComponentPage.lambda$new$7b976870$1(UpdateEditorComponentPage.java:38)
	at com.vaadin.flow.component.ComponentEventBus.fireEventForListener(ComponentEventBus.java:239)
	at com.vaadin.flow.component.ComponentEventBus.handleDomEvent(ComponentEventBus.java:488)
	at com.vaadin.flow.component.ComponentEventBus.lambda$addDomTrigger$dd1b7957$1(ComponentEventBus.java:298)
	at com.vaadin.flow.internal.nodefeature.ElementListenerMap.lambda$fireEvent$2(ElementListenerMap.java:447)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at com.vaadin.flow.internal.nodefeature.ElementListenerMap.fireEvent(ElementListenerMap.java:447)
	at com.vaadin.flow.server.communication.rpc.EventRpcHandler.handleNode(EventRpcHandler.java:62)
	at com.vaadin.flow.server.communication.rpc.AbstractRpcInvocationHandler.handle(AbstractRpcInvocationHandler.java:76)
	at com.vaadin.flow.server.communication.ServerRpcHandler.handleInvocationData(ServerRpcHandler.java:466)
[...]

PR #18623, opened 2024-02-05
Fix regression NPE in binder

tepi:
This patch fixes a regression in Binder introduced by adding skipDefaultValidator API (https://github.com/vaadin/flow/pull/18549). This issue was missed since it seems to be impossible/difficult to trigger via unit testing using the mock field components in Flow project.

Affected flow-components tests:
[UpdateEditorComponentIT](https://github.com/vaadin/flow-components/blob/main/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/UpdateEditorComponentIT.java#L34)
[DynamicEditorKBNavigationIT](https://github.com/vaadin/flow-components/blob/main/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DynamicEditorKBNavigationIT.java#L37)
[GridViewEditorIT.dynamicNotBufferedEditor_closeEditorUsingKeyboard](https://github.com/vaadin/flow-components/blob/main/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewEditorIT.java#L96)
[GridViewEditorIT.dynamicNotBufferedEditor](https://github.com/vaadin/flow-components/blob/main/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewEditorIT.java#L404)
[GridViewEditorIT.dynamicNotBufferedEditor_navigateUsingKeyboard](https://github.com/vaadin/flow-components/blob/main/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewEditorIT.java#L466)

Fixes https://github.com/vaadin/flow/issues/18608

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -1024,6 +1024,12 @@ public class Binder<BEAN> implements Serializable {
 
             BindingImpl<BEAN, FIELDVALUE, TARGET> binding = new BindingImpl<>(
                     this, getter, setter);
+            // Setting the binding field value has been moved up here to fix a
+            // regression NPE https://github.com/vaadin/flow/issues/18608 which
+            // breaks tests of Grid component: UpdateEditorComponentIT,
+            // DynamicEditorKBNavigationIT and
+            // GridViewEditorIT.dynamicNotBufferedEditor*
+            this.binding = binding;
 
             // Remove existing binding for same field to avoid potential
             // multiple application of converter
@@ -1042,7 +1048,6 @@ public class Binder<BEAN> implements Serializable {
             if (getBinder().incompleteBindings != null) {
                 getBinder().incompleteBindings.remove(getField());
             }
-            this.binding = binding;
 
             return binding;
         }
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -1006,9 +1006,12 @@ public class Binder<BEAN> implements Serializable {
             this.statusHandler = statusHandler;
 
             if (field instanceof HasValidator hasValidator) {
-                withValidator((val, ctx) -> binding.isDefaultValidatorDisabled()
-                        ? ValidationResult.ok()
-                        : hasValidator.getDefaultValidator().apply(val, ctx));
+                withValidator((val,
+                        ctx) -> binding != null
+                                && binding.isDefaultValidatorDisabled()
+                                        ? ValidationResult.ok()
+                                        : hasValidator.getDefaultValidator()
+                                                .apply(val, ctx));
             }
         }
 


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -1024,6 +1024,12 @@ public Binding<BEAN, TARGET> bind(ValueProvider<BEAN, TARGET> getter,
 
             BindingImpl<BEAN, FIELDVALUE, TARGET> binding = new BindingImpl<>(
                     this, getter, setter);
+            // Setting the binding field value has been moved up here to fix a
+            // regression NPE https://github.com/vaadin/flow/issues/18608 which
+            // breaks tests of Grid component: UpdateEditorComponentIT,
+            // DynamicEditorKBNavigationIT and
+            // GridViewEditorIT.dynamicNotBufferedEditor*
+            this.binding = binding;
```
mshabarov, line 1032: As an addition we probably should check agains null here:
```
            if (field instanceof HasValidator hasValidator) {
                withValidator((val, ctx) -> binding.isDefaultValidatorDisabled()
                        ? ValidationResult.ok()
                        : hasValidator.getDefaultValidator().apply(val, ctx));
            }
```
so that it becaomes
```
            if (field instanceof HasValidator hasValidator) {
                withValidator((val, ctx) -> binding != null && binding.isDefaultValidatorDisabled()
                        ? ValidationResult.ok()
                        : hasValidator.getDefaultValidator().apply(val, ctx));
            }
```

as `withValidator` method has this NPE check.

tepi: Good point, thank you! Fixed now.

mshabarov (approved):

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -1024,6 +1027,12 @@ public Binding<BEAN, TARGET> bind(ValueProvider<BEAN, TARGET> getter,
...
```
knoobie, line 1035: Could this variable creationg be inlined? like this.binding = new? 

Would it make sense to add a unit test in flow about this? I was stumbled by the issue that there isn't any test for such a case; thinking that I'm literally using this (setBean.. adding / removing bindings later on) in every project with a more complex form with dynamic fields.

tepi: It could but would require either changing the field type from Binding to BindingImpl, or doing an ugly cast a few lines down. @mshabarov Any thoughts should we do either?

Screen Shot 2024-02-05 at 14 28 47

knoobie: doesn't sound worth it :)

mshabarov: This issue is unfortunately throws only with Flow component (Grid IIRC): `binding` object is accessed by component during creating a binding itself, so even if we make some unit test in Flow, this would look absolutely synthetic.

tepi: Small clarification: Only Grid editor tests use the binding mechanism in a way that it fails, so I guess the use case is not overly common. It can be made to fail with e.g. standalone TextField as well. But as said, could not find a way to make it fail with a unit test.

PR #18625, opened 2024-02-05
fix(React): install React components even when Hilla is absent

platosha:
Using basic React features should not require having Hilla on the class
path.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -177,8 +177,7 @@ public abstract class NodeUpdater implements FallibleCommand {
             VersionsJsonConverter convert = new VersionsJsonConverter(
                     Json.parse(
                             IOUtils.toString(content, StandardCharsets.UTF_8)),
-                    options.isReactRouterEnabled()
-                            && EndpointRequestUtil.isHillaAvailable());
+                    options.isReactRouterEnabled());
             versionsJson = convert.getConvertedJson();
             versionsJson = new VersionsJsonFilter(getPackageJson(),
                     DEPENDENCIES)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -177,8 +177,7 @@ private JsonObject getFilteredVersionsFromResource(URL versionsResource,
             VersionsJsonConverter convert = new VersionsJsonConverter(
                     Json.parse(
                             IOUtils.toString(content, StandardCharsets.UTF_8)),
-                    options.isReactRouterEnabled()
-                            && EndpointRequestUtil.isHillaAvailable());
+                    options.isReactRouterEnabled());
```
knoobie, line 180: So how do I exclude React then as flow only developer while still using the "new default" router? Which is enabled by default?

Artur-: If you use React router, you also use React. This PR does not change that in any way, as far as I understand

knoobie: This would also add react-/pro vaadin components poluting the package.json if I remember the old PR correctly

knoobie: Don't clutter flow with that. Period. Flow should only include JavaScript the developer has explicitly opt-in. Flow Developer that wanna use react based Vaadin Components need a Java companion anyway, meaning that the class they use to connect both side with need an NPM annotation anyway. Automatic including yet another dependency with no added value feels contra productive, especially now that SBOMs are highly enforced in multiple sectors and every added dependency has to be approved, inspected and what not. Or is the SBOM Generation of flow intelligent enough to not just include the package.json's dependencies in the crested SBOM.. but only the real real used stuff after an production build?

mshabarov: It sounds more logical, if, having react-router and core react libs by default already, we add also react wrappers for Vaadin Web components.

Artur-: @knoobie using https://github.com/janbiasi/rollup-plugin-sbom?tab=readme-ov-file#usage-with-vite it should be possible to get the actual dependencies... but maybe we are discussing in the wrong issue

knoobie: > but maybe we are discussing in the wrong issue

That's probably right, but it's one of the problems that raise when this is added. 

> It sounds more logical, if, having react-router and core react libs by default already, we add also react wrappers for Vaadin Web components.

What are the benefits of this for flow users? I don't really see it or better say.. I don't know your bigger picture here. 
 - react-router : understandable, because it replaces the vaadin router (not really maintained in my eyes -> good)
 - react : understandable, because it's a direct dependency of the react-router 
 - vaadin react components: That part I don't understand. When people use flow with hilla it makes sense -> already done without any change. If flow people don't use them, they don't need them -> already done. If flow people wanna use them, they add the React Component Adapter and add the vaadin react components like any other javascript library. 
 
Switching back to reactRouterEnabled=false does not feel like a good long term solution, that's why I'm hesitant..

tarekoraby: To my mind, React router support goes hand in hand with support for using React Vaadin components. It thus makes sense to me to use `reactRouterEnabled` to ensure that React Vaadin components are added as well.

Perhaps we should rename `reactRouterEnabled` to `reactEnabled`.

mshabarov: One of the goals of unification is to provide possibly a best DX for cases, when you want to start a development in React in your Flow project. Adding React component by default help to achieve this: you get the auto-completion immediately, you don't need to add manually any dependencies or re-configure your project.

As a side note, lacking React components by default raises npm installation issues and we need this to complete the puzzle.

mshabarov: We agreed to rename the flag to `reactEnabled` as Tarek commented. This should be more consistent with the concept.

PR #18630, opened 2024-02-06
fix: Catch exceptions when package.json template doesn't exist

mshabarov:
If a package.json template cannot be found for any reason, Flow catches it and logs an error message.


mshabarov:
Tested with the hybrid application starter and saw no NPE.

tltv (approved):

manolo (approved):

Issue #18578, opened 2024-01-31
Add "auto" mode to vaadin.frontend.hotdeploy config parameter


mshabarov:
### Describe your motivation

`vaadin.frontend.hotdeploy` should be a string parameter now, and a default value provider for this parameter should call the auto-detection mechanism (from #18577) instead of `EndpointRequestUtil.isHillaAvailable()`. 

If in a project the parameter is
- missing or `auto` -> auto-detection is triggered, 
- if it's blank or `true` -> hot-deploy mode is ON. 
- If it's false -> bundle mode is ON. 

For other values, like `foo`, Flow throws an exception.

### Additional context

Vaadin 24.4.

caalador:
Does it actually need to be a String? Wouldn't the `${null}` value for Boolean used in `reactRouterEnabled` and `productionMode` suffice to see if we should use auto detection, hotdeploy or bundle mode?

mshabarov:
The only reason for String is an ability to `auto` value for this parameter that triggers auto-detection, as well as absence of this parameter.

caalador:
But isn't this still the same, except you need to do string comparisons.
If auto is default then it shouldn't be a problem to have the 3 step version with a boolean.

mshabarov:
Yeah, with String it would need extra complexity of course.
Well, I was thinking that it is more convenient to swap `true` or `false` to `auto` and vice versa, rather than deleting and adding the whole parameter.
But perhaps this isn't something that needs a change often. Once you set it, you use it alway. Thus, I'm almost convinced to go with `Boolean`.

PR #18635, opened 2024-02-07
chore: change hotdeploy property to null by default

tltv:
null means automatic mode.

Fixes: #18578

### DIFF
--- a/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java
+++ b/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java
@@ -221,8 +221,8 @@ public abstract class FlowModeAbstractMojo extends AbstractMojo
      * <p>
      * By default, the frontend server is not used.
      */
-    @Parameter(property = InitParameters.FRONTEND_HOTDEPLOY, defaultValue = "false")
-    private boolean frontendHotdeploy;
+    @Parameter(property = InitParameters.FRONTEND_HOTDEPLOY, defaultValue = "${null}")
+    private Boolean frontendHotdeploy;
 
     @Parameter(property = InitParameters.SKIP_DEV_BUNDLE_REBUILD, defaultValue = "false")
     private boolean skipDevBundleRebuild;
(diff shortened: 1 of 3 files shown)


flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java
```diff
@@ -221,8 +221,8 @@ public abstract class FlowModeAbstractMojo extends AbstractMojo
      * <p>
      * By default, the frontend server is not used.
      */
-    @Parameter(property = InitParameters.FRONTEND_HOTDEPLOY, defaultValue = "false")
-    private boolean frontendHotdeploy;
+    @Parameter(property = InitParameters.FRONTEND_HOTDEPLOY, defaultValue = "${null}")
+    private Boolean frontendHotdeploy;
```
mshabarov, line 225: This also needs a change in `VaadinFlowPluginExtension`, as now it falls back to false:
```
    val frontendHotdeploy: Property<Boolean> = extension.frontendHotdeploy
        .convention(false)
```
So you need to call `FrontendUtils.isHillaUsed` in the convention callback.

flow-plugins/flow-maven-plugin/src/test/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojoTest.java
```diff
@@ -157,6 +157,22 @@ public void tokenFileShouldExist_noHotdeployTokenVisible()
                 buildInfo.get(SERVLET_PARAMETER_PRODUCTION_MODE));
     }
 
+    @Test
+    public void tokenFileShouldExist_hotdeployIsTrueTokenVisible()
+            throws IOException, MojoExecutionException, MojoFailureException,
+            IllegalAccessException {
+        ReflectionUtils.setVariableValueInObject(mojo, "frontendHotdeploy",
+                Boolean.TRUE);
```
mshabarov, line 165: Why this is TRUE? I expect a null value here and check that the isHillaUsed method to be called when it's null.

mshabarov (changes requested):

mshabarov (approved):

PR #18636, opened 2024-02-07
feat: Add method to write only the changed properties to the bean

### LINKED ISSUE #18583: Binder: Add method to write only the changed properties to the bean
It is not uncommon that setters of the bean may have side effects that application developer would not like to get triggered when writing bean with value which is not changed. E.g. This is how jOOQ tracks changes on records and will then generate minimal update statements.
​
Current implementation is
​
```
​public void writeBean(BEAN bean) throws ValidationException {
    BinderValidationStatus<BEAN> status = doWriteIfValid(bean, bindings);
    if (status.hasErrors()) {
        throw new ValidationException(status.getFieldValidationErrors(),
                status.getBeanValidationErrors());
    }
}
```
​
1. One implementation would be to add a new method `updateBean`.

```
​public void updateBean(BEAN bean) throws ValidationException {
    BinderValidationStatus<BEAN> status = doWriteIfValid(bean, changedBindings);
    if (status.hasErrors()) {
        throw new ValidationException(status.getFieldValidationErrors(),
                status.getBeanValidationErrors());
    }
}
```

2. Another implementation could be refactor and add method boolean parameter
​
```
​public void writeBean(BEAN bean) throws ValidationException {
    writeBean(bean, false);
}
​
​public void writeBean(BEAN bean, boolean changedFields) throws ValidationException {
    if (changedFields) {
       BinderValidationStatus<BEAN> status = doWriteIfValid(bean, bindings);
    } else {
        BinderValidationStatus<BEAN> status = doWriteIfValid(bean, changedBindings);
    }
    if (status.hasErrors()) {
        throw new ValidationException(status.getFieldValidationErrors(),
                status.getBeanValidationErrors());
    }
}
```
​
Unfortunately `changedBindings` is private and you can just extend Binder

3.  One could consider also another approach of having methods like

```
Set<Binding<BEAN, ?>> getChangedBindings() {
    return changedBindings;
}
```

```
​public void writeBean(BEAN bean) throws ValidationException {
    writeBean(bean, bindings);
}
​
​public void writeBean(BEAN bean, Set<Binding<BEAN, ?>> bindingsToWrite) throws ValidationException {
    BinderValidationStatus<BEAN> status = doWriteIfValid(bean, bindingsToWrite);
    if (status.hasErrors()) {
        throw new ValidationException(status.getFieldValidationErrors(),
                status.getBeanValidationErrors());
    }
}

​public void updateBean(BEAN bean) throws ValidationException {
    writeBean(bean, changedBindings);
}
```

This would allow greater flexibility to write any subset of bindings for other edge cases i am not able to figure out now. Having getter for `changedBindings` could also serve other purposes (see also https://github.com/vaadin/flow/issues/15266).

What you would vote for 1., 2. or 3.?

TatuLund:
fixes #18583

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -2405,7 +2405,8 @@ public class Binder<BEAN> implements Serializable {
      * @throws ValidationException
      *             if some of the bound field values fail to validate
      */
-    public void writeChangedBindingsToBean(BEAN bean) throws ValidationException {
+    public void writeChangedBindingsToBean(BEAN bean)
+            throws ValidationException {
         writeBean(bean, getChangedBindings());
     }
 
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -2322,6 +2322,7 @@ public class Binder<BEAN> implements Serializable {
      * thrown.
      *
      * @see #writeBeanIfValid(Object)
+     * @see #writeBean(Object, Collection)
      * @see #readBean(Object)
      * @see #setBean(Object)
      *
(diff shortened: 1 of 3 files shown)


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2332,13 +2333,86 @@ public void refreshFields() {
...
+
+    /**
+     * Get the Set of changed bindings.
+     *
+     * @see #hasChanges()
+     *
+     * @return Set of bindigns.
+     */
+    public Set<Binding<BEAN, ?>> getChangedBindings() {
+        return changedBindings;
```
mcollovati, line 2413: I would return an immutable copy of `changedBindings`, to prevent unwanted modifications to the collection.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2332,13 +2333,86 @@ public void refreshFields() {
...
+     * @see #readBean(Object)
+     * @see #setBean(Object)
+     *
+     * @param bean
+     *            the object to which to write the field values, not
+     *            {@code null}
+     * @throws ValidationException
+     *             if some of the bound field values fail to validate
+     */
+    public void updateBean(BEAN bean) throws ValidationException {
```
mcollovati, line 2401: On the linked issue, there was a remark about not adding this method. Ping @mshabarov 

Anyway, as also suggested on the issues, I would change the name to something more explicit like `writeChangedBindingToBean` or `writeChangesToBean`

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2332,13 +2333,86 @@ public void refreshFields() {
...
+     * @param bean
+     *            the object to which to write the field values, not
+     *            {@code null}
+     * @param bindingsToWrite
+     *            Collection of bindings to use in writing the bean
+     * @throws ValidationException
+     *             if some of the bound field values fail to validate
+     */
+    public void writeBean(BEAN bean,
+            Collection<Binding<BEAN, ?>> bindingsToWrite)
```
mcollovati, line 2366: Since we allow providing a collection of bindings from the outside, I wonder if we should add a sanity check to ensure that the given bindings belong to this Binder

mcollovati (changes requested):

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
mcollovati: If the collection contains _foreign_ bindings, we could either throw an exception or just log a warnin message and ignore them.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
mshabarov: `writeChangedBindingToBean` sounds self-explanatory, I'd go with this name and keep this method.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
TatuLund: I prefer throwing IllegalArgumentException.

mcollovati (approved):

PR #18661, opened 2024-02-09
fix: delay initial poller call

### LINKED ISSUE #18645: Attributes are not resent to the server when coming back to a view with a WebComponent.
When using a WebComponent with client set attributes in a hilla view navigating out and the back to the view client will not set the attrubutes again and the server component will not have the same data that was available before navigation.

Simplest test is to create a Hilla application with 2 views where one brings in the sample WebComponent from [Flow Component in Hilla](https://github.com/vaadin/docs/pull/3170) documentation

caalador:
Delay the first poller call
for registerElement as if
we have already a live
client and the required method
react won't be fast enough
to set the attributes before
they are read.

Fixes #18645


mshabarov:
@caalador could you share an example project where you tested, so I can do testing also ?

mshabarov (changes requested):
Works well with this fix 👍 

Could you please add a comment line explaining why we need to postpone the poll for future maintenance?

caalador:
Done

mshabarov (approved):

PR #18668, opened 2024-02-12
fix: Log debug connection issues as debug

Artur-:
There is no value in seeing a lot of messages about denied connections if you e.g. leave browser tabs open. The warning is logged in the browser so you can understand why the connection is not working

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
@@ -201,8 +201,8 @@ public class DebugWindowConnection implements BrowserLiveReload {
                 .equals(resource.getRequest().getParameter("token"))) {
             handleConnect(resource);
         } else {
-            getLogger().warn(
-                    "Connection denied because of a missing or invalid token. The host is probably not on the allow list");
+            getLogger().debug(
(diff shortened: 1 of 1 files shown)


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -201,8 +201,8 @@ public void onConnect(AtmosphereResource resource) {
                 .equals(resource.getRequest().getParameter("token"))) {
             handleConnect(resource);
         } else {
-            getLogger().warn(
-                    "Connection denied because of a missing or invalid token. The host is probably not on the allow list");
+            getLogger().debug(
+                    "Connection denied because of a missing or invalid token. Either the host is not on the 'vaadin.devmode.hosts-allowed' list or it is using an outdated token");
```
mcollovati, line 205: nit: should we also suggest restarting the server to resolve the issue?

mcollovati (approved):
I added a comment about the log message, but the PR looks good to me also as is.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
Artur-: I am not sure that makes sense as it would be logged when I share a link to you, to my local dev environment. Then there is nothing to resolve, it's just that you are not allowed to use the debug connection to change my files

PR #18669, opened 2024-02-12
chore: Hide dev tools by default

Artur-:
Copilot will be shown instead


mshabarov (approved):

mshabarov:
Just for understanding, this only hides the dev tools windows, but do we target the whole dev tools movement to Copilot in 24.4, or will it be next minor/major?

Artur-:
Functionality will move later

PR #18677, opened 2024-02-12
fix: do not run generator without hilla

caalador:
Do not add the generators
if not hilla used in project.

Fixes vaadin/platform#5016

### DIFF
--- a/flow-plugins/flow-plugin-base/src/test/java/com/vaadin/flow/plugin/base/BuildFrontendUtilTest.java
+++ b/flow-plugins/flow-plugin-base/src/test/java/com/vaadin/flow/plugin/base/BuildFrontendUtilTest.java
@@ -98,7 +98,7 @@ public class BuildFrontendUtilTest {
 
         // no hilla no lookup call
         Mockito.verify(lookup, Mockito.never())
-                lookup(EndpointGeneratorTaskFactory.class);
+                .lookup(EndpointGeneratorTaskFactory.class);
         Mockito.verify(lookup, Mockito.never())
                 .lookup(TaskGenerateOpenAPI.class);
         Mockito.verify(lookup, Mockito.never())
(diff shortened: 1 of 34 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -315,7 +315,9 @@ private void addEndpointServicesTasks(Options options) {
         EndpointGeneratorTaskFactory endpointGeneratorTaskFactory = lookup
                 .lookup(EndpointGeneratorTaskFactory.class);
 
-        if (endpointGeneratorTaskFactory != null) {
+        if (endpointGeneratorTaskFactory != null && FrontendUtils
```
mcollovati, line 318: Could the whole `addEndpointServicesTasks` method be skipped if Hilla is not used?

caalador: Well there is a comment that it is skipped inside the method. Moved the isUsed check to be first.

flow-plugins/flow-plugin-base/src/test/java/com/vaadin/flow/plugin/base/BuildFrontendUtilTest.java
```diff
@@ -95,7 +96,9 @@ public void should_notUseHilla_inPrepareFrontend()
 
         BuildFrontendUtil.prepareFrontend(adapter);
 
-        Mockito.verify(lookup).lookup(EndpointGeneratorTaskFactory.class);
+        // no hilla no lookup call
+        Mockito.verify(lookup, Mockito.never())
+                lookup(EndpointGeneratorTaskFactory.class);
```
mcollovati, line 101:
```suggestion
                .lookup(EndpointGeneratorTaskFactory.class);
```

mcollovati (changes requested):
There's a compilation error in one test, but other than that LGTM

flow-plugins/flow-plugin-base/src/test/java/com/vaadin/flow/plugin/base/BuildFrontendUtilTest.java
caalador: Done.

mcollovati (approved):

PR #18680, opened 2024-02-12
feat: add translation file request handler

### LINKED ISSUE #18641: Implement an API for getting translation properties file
### Describe your motivation

We would need a way of getting translation properties file contents for locale bundle generation to be used within Hilla views.

Implement Spring `Controller` or [`RequestHandler`](https://vaadin.com/docs/latest/advanced/service-init-listener) for accessing the translation property file.
There should be a logic in the Flow side to get all of the available translation property files. (to be verified)

_No response_

### Additional context

Further can be enhanced with [Hilla Endpoint](https://hilla.dev/docs/react/guides/endpoints#creating). Not in the initial implementation as it's difficult to apply it for this use-case.

ugur-vaadin:
## Description

This PR adds a request handler to serve translation properties files. For the request to be handled by this handler, the path info should be equal to `i18n`. There are two optional parameters to construct the locale: `lang` and `country`. If a translation properties file that corresponds to the locale, the response contains the translation entries in JSON format.

Fixes #18641

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                request.getParameter(COUNTRY_PARAMETER_NAME), "");
+        return new Locale(language, country);
+    }
+
+    private ResourceBundle getTranslationPropertyFile(Locale locale) {
+        try {
+            return ResourceBundle.getBundle(DefaultI18NProvider.BUNDLE_PREFIX,
+                    locale, I18NUtil.getClassLoader());
+        } catch (MissingResourceException e) {
+            getLogger().warn("Missing resource bundle for "
```
knoobie, line 87: Even tho this is a direct copy of https://github.com/vaadin/flow/blob/cd0c36f2b4d6a001e1b147ffb4e73d980160f7f3/flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java#L91 - I would suggest to lower the severity of the logging here to info or lower. 

The other provider is instantiated with the used locales ok the server side, making a missing resource bundle defined by a server side locale by the developer more severe than a possible (random) locally supplied by a user via http headers.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    private void writeFileToResponse(VaadinResponse response,
+            ResourceBundle translationPropertyFile) throws IOException {
+        JsonObject json = Json.createObject();
+        translationPropertyFile.keySet().forEach(
+                key -> json.put(key, translationPropertyFile.getString(key)));
+        response.getWriter().write(json.toJson());
+    }
+
+    private boolean isTranslationFileRequest(VaadinRequest request) {
+        return request.getPathInfo().equals(TRANSLATION_FILE_REQUEST_PATH_INFO);
```
mcollovati, line 71: I don't know if it has already been discussed, but what about using the Flow request type parameter (`ApplicationConstants.REQUEST_TYPE_PARAMETER`)  instead of path info?
Using path info could potentially "steal" a valid route path.

We are currently using the request type parameter for Flow "services" (e.g. push, heartbeat, ...).
See `RequestType` enum in `HandlerHelper` class.

ugur-vaadin: Good catch. I added an enum `RequestType.TRANSLATION_FILE` with the constant `ApplicationConstants.REQUEST_TYPE_TRANSLATION_FILE`, and updated the handler and the tests to use them.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                key -> json.put(key, translationPropertyFile.getString(key)));
+        response.getWriter().write(json.toJson());
+    }
+
+    private Locale getLocale(VaadinRequest request) {
+        String language = Objects.requireNonNullElse(
+                request.getParameter(LANGUAGE_PARAMETER_NAME), "");
+        String country = Objects.requireNonNullElse(
+                request.getParameter(COUNTRY_PARAMETER_NAME), "");
+        return new Locale(language, country);
```
sissbruecker, line 74: Was there some special consideration for using separate identifiers for language and country instead of the language tag? In the frontend you would usually just want to have a single identifier for a language, such as `"en-US"`. So that means we would have to split up the tag on the frontend and then join it together again on the server. Not a big deal, but might not be necessary. In general I think we're pretty flexible with this and can change this later on.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,92 @@
...
+    public boolean handleRequest(VaadinSession session, VaadinRequest request,
+            VaadinResponse response) throws IOException {
+        if (!HandlerHelper.isRequestType(request,
+                HandlerHelper.RequestType.TRANSLATION_FILE)) {
+            return false;
+        }
+        Locale locale = getLocale(request);
+        ResourceBundle translationPropertyFile = getTranslationPropertyFile(
+                locale);
+        if (translationPropertyFile == null) {
```
sissbruecker, line 43: From the perspective of the client we need to have one set of translations in order to for an app to display something to the user. Returning a not found is problematic:
- The client would first make a request for the identified language
- Gets a 404, and then does what?
- Maybe tries to load a fallback language, which would result in an additional request. That would mean it takes even longer before the app can display something to the user. Also the fallback language configured in the client might not exist either.

I was thinking the server should return a "default" set of translations whenever it doesn't find a set of translations for the specified language. That means there would only ever be one request before the app can display something. In terms of Java resource bundles, that would probably mean we send the contents of the default / root properties file. The response would then also need to contain some information for which language the set of translations are actually for, so the client knows that it received the fallback instead of the thing it requested.

ugur-vaadin: Updated the PR to include the language tag for the retrieved bundle. The handler returns the default file if it is available, and file not found response is only the case when there is no default to return. The language tag is added to the response with the header name [Retrieved-locale](https://github.com/vaadin/flow/pull/18680/files#diff-bbee435e8e962ebcf4be9f7c0043f7fc1f41193a3b8381cb76087f57116019a4R31).

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: I initially lowered the severity. However, with the final implementation, we want the handler to return any available default bundle if the requested bundle is not found. So maybe it could be better to have a warning log in this case, since it also means that there is no default translation too. I mentioned that there is no default bundle to fall back on in the log message.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
sissbruecker: So it does return a default bundle now, though it might not be the one you expect. What seems to happen now is:
- If there is a bundle for the server locale (`Locale.getDefault()`), it sends that
- Otherwise it sends the root bundle (`translation.properties`)

Since Flow's `DefaultI18nProvider` uses the same method call, I would assume it works the same way, but it would be good to double check. Otherwise we might want to have a discussion what the fallback should be. Should it consider the server locale as fallback, or should it always use the root bundle? If we consider the server locale it could return different results in different environments. On the other hand I don't know what the best practice is for the root bundle. Would it be a safe assumption that the root bundle contains translations for the default language? An alternative could be to let developers explicitly specify a fallback language of which they know that a bundle exists.

ugur-vaadin: Added tests with mocked system default locale that confirm this behaviour. I will modify the tests based on the decisions on whether to introduce a specified fallback locale and the use of system default locale in general.

sissbruecker: Seems to work well, also still handles fallbacks such as from `de` -> `de_DE` and vice versa.

sissbruecker (dismissed):
LGTM, Flow team should have a final look at the changes.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.i18n;
+
+import com.vaadin.flow.server.*;
```
mcollovati, line 18: Please, avoid wildcard imports.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,106 @@
...
+import elemental.json.Json;
+import elemental.json.JsonObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.*;
```
mcollovati, line 25: Please, avoid wildcard imports

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,106 @@
...
+
+public class TranslationFileRequestHandler implements RequestHandler {
+
+    static final String LANGUAGE_TAG_PARAMETER_NAME = "langtag";
+
+    static final String RETRIEVED_LOCALE_HEADER_NAME = "Retrieved-Locale";
```
mcollovati, line 31: I wonder if it would be better to use a more customized header name, such as `X-Vaadin-Retrieved-Locale` or `Vaadin-Retrieved-Locale`

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,106 @@
...
+        response.getWriter().write(json.toJson());
+    }
+
+    private Locale getLocale(VaadinRequest request) {
+        String languageTag = Objects.requireNonNullElse(
+                request.getParameter(LANGUAGE_TAG_PARAMETER_NAME), "");
+        return Locale.forLanguageTag(languageTag);
+    }
+
+    private ResourceBundle getTranslationPropertyFile(Locale locale) {
```
mcollovati, line 79: It looks to me that this method assumes that the `DefaultI18NProvider` is always in use, but an application may provide its own implementation of `I18NProvider`.
I wonder if we need to revisit `I18NProvider` interface to make this handler work with custom implementations.

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
sissbruecker: I feel there isn't an easy answer here. Another feature that is supposed to be implemented soon is support for pushing and pulling translations to/from Control Center. It's not really clear how that is supposed to work together with a custom I18n provider either, right now all discussions have revolved around using the default translations resource bundle. 

Personally I would assume that these features are not intended to work together and just make the default setup work. Let's see if @rolfsmeds can clarify.

ugur-vaadin: Good catch. Do you think it makes sense to add APIs to get/set the translation folder and bundle file names?

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: Updated the PR to use explicit imports.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: Updated the PR to use `X-Vaadin-Retrieved-Locale` as the header name.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
mcollovati: > Do you think it makes sense to add APIs to get/set the translation folder and bundle file names

I'm not sure about that. Making it configurable will require additional code to read values from config or the developer to extend the default implementation class. In addition, it would probably break native builds (or require changes to support the configurable values)

mcollovati: So, the request handler logic should be executed only if the default I18N provider is in use (`vaadinSession.getService().getInstantiator().getI18NProvider() instanceof DefaultI18NProvider`)

ugur-vaadin: Added the default i18n provider check to the handler.

sissbruecker: What kind of response is generated now if the default I18n provider isn't used? Does it just return the index HTML?

IMO we should specifically generate an error response stating that loading translations is not supported when using a custom I18n provider.

knoobie: Security Stand Point: 
- Production mode: 404
- Development mode: more information

mcollovati: Good point @sissbruecker.
@ugur-vaadin this means that the request handler should always be installed, so the check in VaadinService we talk earlier should be removed.
Or, as an alternative, we could instantiate the handler with a constructor argument indicating whether it is enabled or not, e.g `new TranslationFileRequestHandler(getInstantiator().getI18NProvider() instanceof DefaultI18NProvider)` or just `new TranslationFileRequestHandler(getInstantiator().getI18NProvider())` and handle logic in the handler (having a reference to I18N provider could also be useful for next related features).

mcollovati: In addition, if the request handler stores a reference to the I18N provider, we could make  `DefaultI18NProvider.getBundle(...)` package protected and reuse it, if it could help to prevent code duplication

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -40,12 +40,24 @@ public class TranslationFileRequestHandler implements RequestHandler {
...
     @Override
     public boolean handleRequest(VaadinSession session, VaadinRequest request,
             VaadinResponse response) throws IOException {
-        if (!shouldHandle(session, request)) {
+        if (!HandlerHelper.isRequestType(request,
+                HandlerHelper.RequestType.TRANSLATION_FILE)) {
             return false;
         }
+        if (!(i18NProvider instanceof DefaultI18NProvider)) {
+            response.sendError(HttpStatusCode.METHOD_NOT_ALLOWED.getCode(),
```
mcollovati, line 57: `METHOD_NOT_ALLOWED` seems not the best status code to me.
I would go either with a `400 - Bad Request` or perhaps `501 Not Implemented`.
Also, we should consider Knoobie's remark about security and return a `404` in production mode.

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: Updated the PR to return `404` for production and `400` otherwise when a custom i18n provider is used.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,135 @@
...
+        response.setStatus(HttpStatusCode.NOT_FOUND.getCode());
+    }
+
+    private void handleCustomI18NProvider(VaadinSession session,
+            VaadinResponse response) throws IOException {
+        String errorMessage = "Loading translations is not supported when using a custom i18n provider.";
+        if (session.getConfiguration().isProductionMode()) {
+            response.setStatus(HttpStatusCode.NOT_FOUND.getCode());
+        } else {
+            response.sendError(HttpStatusCode.BAD_REQUEST.getCode(),
```
mcollovati, line 89: Thinking again about the status code, I would change it to `501 - NOT_IMPLEMENTED`.
Bad request means that the client sent a malformed request, but this is not the case. The request is correct, but on the server side we cannot fulfill it.

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,135 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Locale;
+import java.util.MissingResourceException;
+import java.util.Objects;
+import java.util.ResourceBundle;
+
+public class TranslationFileRequestHandler implements RequestHandler {
```
mcollovati, line 35: Since we now access VaadinSession to check production mode, this should extend `SynchronizedRequestHandler`, otherwise at runtime we get `java.lang.IllegalStateException: Cannot access state in VaadinSession or UI without locking the session.`
As an alternative, the production mode flag can probably be computed once in the constructor

flow-server/src/test/java/com/vaadin/flow/i18n/TranslationFileRequestHandlerTest.java
```diff
@@ -0,0 +1,252 @@
...
+            throws IOException {
+        testResponseContentWithMockedDefaultLocale("es-ES", true, "en-US",
+                "{\"title\":\"Root bundle lang\"}", "und");
+    }
+
+    @Test
+    public void withoutRootBundle_requestedLocaleBundleNotAvailable_responseIsEmpty()
+            throws IOException {
+        testResponseContentWithMockedDefaultLocale("es-ES", false, "en-US", "",
+                null);
```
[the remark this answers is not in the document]
ugur-vaadin: Added the verification.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: Updated the PR to make `DefaultI18NProvider.getBundle` package protected and reuse it. Also, added the i18n provider to the handler constructor to hold the reference the it.

flow-server/src/test/java/com/vaadin/flow/i18n/TranslationFileRequestHandlerTest.java
```diff
@@ -0,0 +1,252 @@
...
+            String requestedLanguageTag, String expectedResponseContent,
+            String expectedResponseLanguageTag) throws IOException {
+        createTranslationFiles(withRootBundleFile);
+        try (MockedStatic<I18NUtil> util = Mockito.mockStatic(I18NUtil.class,
+                Mockito.CALLS_REAL_METHODS)) {
+            util.when(I18NUtil::getClassLoader).thenReturn(urlClassLoader);
+            setRequestParams(requestedLanguageTag,
+                    HandlerHelper.RequestType.TRANSLATION_FILE.getIdentifier());
+            Assert.assertTrue(
+                    handler.handleRequest(session, request, response));
```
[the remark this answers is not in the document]
ugur-vaadin: Added failure messages to response content assertions.

flow-server/src/test/java/com/vaadin/flow/i18n/TranslationFileRequestHandlerTest.java
```diff
@@ -0,0 +1,252 @@
...
+    public void withRootBundle_languageTagWithoutCountryAvailable_responseIsCorrect()
+            throws IOException {
+        testResponseContent(true, "fi", "{\"title\":\"Suomi\"}", "fi");
+    }
+
+    @Test
+    public void withRootBundle_languageTagWithoutCountryNotAvailable_responseIsRootBundle()
+            throws IOException {
+        testResponseContent(true, "es", "{\"title\":\"Root bundle lang\"}",
+                "und");
```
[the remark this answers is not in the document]
mcollovati: To provide at least a compatible bundle, we can get the list of available locales from `DefaultI18NProvider.getProvidedLocales()` (values are computed based on the existing translation files) and the collection to get `Locale`s with the same `language` and take the first one, if any.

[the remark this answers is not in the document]
ugur-vaadin: Updated the PR to find and return the best locale match among the bundles. Also added support for underscore delimiters in language tag.

flow-server/src/test/java/com/vaadin/flow/i18n/TranslationFileRequestHandlerTest.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    @Test
+    public void withoutRootBundle_languageTagWithoutCountryNotAvailable_responseIsEmpty()
+            throws IOException {
+        testResponseContent(false, "es", "", null);
```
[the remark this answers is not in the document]
ugur-vaadin: Updated the PR to find and return the best locale match among the bundles.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    private ResourceBundle getTranslationPropertyFile(Locale locale) {
+        Locale bestMatchLocale = getBestMatchLocale(locale);
+        if (!locale.equals(bestMatchLocale)) {
+            getLogger().debug("Missing resource bundle for "
+                    + DefaultI18NProvider.BUNDLE_PREFIX + " and locale "
+                    + locale.getDisplayName() + ".");
+        }
+        if (bestMatchLocale == null) {
```
mcollovati, line 127: I would try to swap the two conditions and use an `else if`.
It should prevent the "Missing..." message to being printed twice in the case of `bestMatchLocale` is `null`.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,169 @@
...
+        if (languageAndCountryMatch.isPresent()) {
+            return languageAndCountryMatch.get();
+        }
+        Optional<Locale> languageMatch = providedLocales.stream()
+                .filter(providedLocale -> providedLocale.getLanguage()
+                        .equals(locale.getLanguage()))
+                .findAny();
+        if (languageMatch.isPresent()) {
+            return languageMatch.get();
+        }
```
mcollovati, line 159: nit: minor performance improvement: what about doing a first filtering only for `language` and then searching for `language + country` on the reduced collection? I should prevent looping twice two over the full list of provided locales (although the list may be small)

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,169 @@
...
+        JsonObject json = Json.createObject();
+        translationPropertyFile.keySet().forEach(
+                key -> json.put(key, translationPropertyFile.getString(key)));
+        response.getWriter().write(json.toJson());
+    }
+
+    private Locale getLocale(VaadinRequest request) {
+        String languageTag = Objects.requireNonNullElse(
+                request.getParameter(LANGUAGE_TAG_PARAMETER_NAME), "");
+        if (languageTag.contains("_")) {
```
mcollovati, line 110: Support for parameters like `en_US` looks good. We need to tests for this variant.

flow-server/src/test/java/com/vaadin/flow/i18n/TranslationFileRequestHandlerTest.java
mcollovati, line 1: I would add tests for unsupported languages as well. e.g. 'it', 'it_IT' and 'it-IT'

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,169 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Locale;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.ResourceBundle;
+import java.util.Set;
+
+public class TranslationFileRequestHandler extends SynchronizedRequestHandler {
```
mcollovati, line 36: Needs Javadoc.
In addition to scope, behavior and parameters, I would mention how language resolution works.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: I considered that but thought that the clarity of the flow might be more favourable than the performance gains. Still, I will refactor it.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: I already added a test for resolving `es_ES`. Do you think we should add any more?

flow-server/src/test/java/com/vaadin/flow/i18n/TranslationFileRequestHandlerTest.java
ugur-vaadin: Do you mean locales that have no corresponding resource bundles present, or are there explicitly unsupported languages that are specified somewhere?

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
mcollovati: Sorry, my bad. I did not notice it.

flow-server/src/test/java/com/vaadin/flow/i18n/TranslationFileRequestHandlerTest.java
mcollovati: The first one. Locales that have no translation file

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
mcollovati: You can leave the code as is, if you think it is easier to read and maintain.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -0,0 +1,169 @@
...
+    private final I18NProvider i18NProvider;
+
+    public TranslationFileRequestHandler(I18NProvider i18NProvider) {
+        this.i18NProvider = i18NProvider;
+    }
+
+    @Override
+    public boolean synchronizedHandleRequest(VaadinSession session,
+            VaadinRequest request, VaadinResponse response) throws IOException {
+        if (!(i18NProvider instanceof DefaultI18NProvider)) {
```
knoobie, line 53: Just asking: 

`DefaultI18NProvider` is public and non-final (as well as all methods in it), allowing me to do e.g.: `MyI18nProvider extends DefaultI18NProvider` -> which would also match here.. is that intended?

mcollovati: We are mostly using package protected methods of `DefaultI18NProvider`, but, yeah, the check could probably be on exact class. IIRC you are always instantiating `DefaultI18NProvider` programmatically and do not rely on DI for it.

knoobie: My question came from something like this; where a user could theoretically just extend the default with their own bundle magic (without knowing that this could literally expose all their translation keys / values):

```java

  public class MyI18nProvider extends DefaultI18NProvider {

    public MyI18nProvider() {
      super(List.of(Locale.GERMAN));
    }

    @Override
    public String getTranslation(String key, Locale locale, Object... params) {
      var flowResult = super.getTranslation(key, locale, params);
      if (key.equals(flowResult)) {
        // flow did not find a translation;
        return "Do our own translation magic";
      } else if (flowResult.equals("!" + locale.getLanguage() + ": " + key)) {
        // flow did not find a translation;
        return "Do our own translation magic";
      } else {
        return flowResult;
      }
    }
  }

```

mcollovati: Good point. @ugur-vaadin what do you think? Would it be safer to make an exact class match?

ugur-vaadin: I think it makes sense to go for the safer option for now. We can consider adding support for more custom implementations later if necessary.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: Added Javadoc including the language matching logic.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: Updated the check and added tests for a provider extending `DefaultI18NProvider`.

flow-server/src/test/java/com/vaadin/flow/i18n/TranslationFileRequestHandlerTest.java
ugur-vaadin: The unsupported `it` and `it-IT` cases were already covered with tests. I added more tests that also cover the `it_IT` type tags.

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
ugur-vaadin: I updated the logic to check for the `null` case first.

mcollovati (dismissed):

mcollovati (approved):

Issue #18367, opened 2023-12-28
Jackson 2.16.1 breaks Gradle builds


mcollovati:
jackson-core 2.16.1 is a multi-version artifact, containing classes for Java 21.

```
META-INF/versions/21/
META-INF/versions/21/com/
META-INF/versions/21/com/fasterxml/
META-INF/versions/21/com/fasterxml/jackson/
META-INF/versions/21/com/fasterxml/jackson/core/
META-INF/versions/21/com/fasterxml/jackson/core/io/
META-INF/versions/21/com/fasterxml/jackson/core/io/doubleparser/
META-INF/versions/21/com/fasterxml/jackson/core/io/doubleparser/FastDoubleSwar.class
META-INF/versions/21/com/fasterxml/jackson/core/io/doubleparser/FastIntegerMath.class
```

This causes Gradle validation failure because of the usage of Gradle `7.6`, that is the declared as the minimum supported version for Flow. 

However, the build works with Jackson 2.16.0, since it contains classes up to Java 19, that is supported by Gradle 7.6

```
META-INF/versions/19/
META-INF/versions/19/com/
META-INF/versions/19/com/fasterxml/
META-INF/versions/19/com/fasterxml/jackson/
META-INF/versions/19/com/fasterxml/jackson/core/
META-INF/versions/19/com/fasterxml/jackson/core/io/
META-INF/versions/19/com/fasterxml/jackson/core/io/doubleparser/
META-INF/versions/19/com/fasterxml/jackson/core/io/doubleparser/FastDoubleSwar.class
META-INF/versions/19/com/fasterxml/jackson/core/io/doubleparser/FastIntegerMath$UInt128.class
META-INF/versions/19/com/fasterxml/jackson/core/io/doubleparser/FastIntegerMath.class
```

https://docs.gradle.org/current/userguide/compatibility.html

Currently, the jackson bump [PR](#18360) has been [reverted](#18366).

Latest [spring-boot-dependencies](https://central.sonatype.com/artifact/org.springframework.boot/spring-boot-dependencies/3.2.1) POM defines `<jackson-bom.version>2.15.3</jackson-bom.version>`.

To be able to upgrade Jackson to 2.16.1+, we need to bump the minimum supported Gradle version at least to 8.4

mcollovati:
Closing as completed in #18683 and #18686

PR #18683, opened 2024-02-13
chore: upgrade minimum supported Gradle version to 8.4

tarekoraby:
Close https://github.com/vaadin/platform/issues/5027.

Related to https://github.com/vaadin/flow/issues/18367.

### DIFF
--- a/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/MiscSingleModuleTest.kt
+++ b/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/MiscSingleModuleTest.kt
@@ -24,40 +24,6 @@ import kotlin.test.expect
 import org.junit.Ignore
 
 class MiscSingleModuleTest : AbstractGradleTest() {
-    /**
-     * Tests https://github.com/vaadin/vaadin-gradle-plugin/issues/26
-     */
-    @Test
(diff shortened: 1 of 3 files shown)


tarekoraby:
I keep forgetting what the `+0.0.1` label is supposed to signify. But just to clarify, this change should be in the next Minor (not 24.3)

mcollovati (changes requested):
`VaadinSmokeTest.pluginShouldFailWithUnsupportedGradleVersion` should be updated as well

tarekoraby:
There is a failure at https://github.com/vaadin/flow/blob/43aab7c96b56d8fcec73a4774b9854197522449c/flow-plugins/flow-gradle-plugin/src/functionalTest/kotlin/com/vaadin/gradle/MiscSingleModuleTest.kt#L31. I don't understand what's wrong nor how to fix it.

mcollovati:
> I don't understand what's wrong nor how to fix it.

The test ensures that gradle plugin works in MPR, in combination with the [com.devsoap.plugin.vaadin](https://github.com/johndevs/gradle-vaadin-plugin) gradle plugin that is for Vaadin 8.
However, the Vaadin 8 compatible plugin is not maintained anymore, and it looks like it cannot run with Gradle 8.

So I think it means that after this upgrade we cannot support gradle anymore in MPR projects

mcollovati (approved):

PR #18687, opened 2024-02-13
chore: Fix typo in default value

Artur-:


caalador (dismissed):

caalador:
`JavaLiveReloadIT:overlayShouldRender()` is failing on the final assert
```
        WebElement window2 = devTools.$("*")
                .attributeContains("class", "dev-tools").first();
        Assert.assertTrue(window2.isDisplayed());
```
So the change seems to be correct, but the test wrong in some way against the new state

Artur-:
It tried to click on the button that is no longer shown and verify that the dev tools window is shown

caalador (approved):

PR #18695, opened 2024-02-14
fix: expose effective plugin configuration at vaadin.effective.

### LINKED ISSUE #18572: Gradle: vaadin.productionMode.get() returns false regardless of -Pvaadin.productionMode
### Description of the bug

In Vaadin 24.2, `vaadin.productionMode` would be a boolean value which would be true if ` -Pvaadin.productionMode` would be supplied on the command line.

Things changed in Vaadin 24.3 - the vaadin.productionMode is a `Property<Boolean>` now. That would be okay however the problem is that the property is always false. The actual effective value of that property is calculated in `PluginEffectiveConfiguration.productionMode`. Unfortunately, the class is internal and the app build.gradle script has no way of accessing that (except for a crazy reflection).

### Expected behavior

Either `vaadin.productionMode.get()` should be made to return the proper effective value, or we should make PluginEffectiveConfiguration public. We could maybe achieve the former by populating the extension object ourselves:
```kotlin
public class VaadinFlowPluginExtension {
  public val productionMode: Property<Boolean> = ObjectFactory.property(Boolean::class.java)
}
```
However, care must be taken that the property keeps its name and origin: that `VaadinFlowPluginExtension.productionMode.toString()` returns `extension 'vaadin' property 'productionMode'` - and I have no idea how to achieve that. Therefore, perhaps it's better to make PluginEffectiveConfiguration public.

`build.gradle.kts`:

```kotlin
plugins {
    java
    application
    id("com.vaadin")
}

defaultTasks("clean", "build")

repositories {
    mavenCentral()
}

val vaadinVersion: String by extra

dependencies {
    // Vaadin
    implementation("com.vaadin:vaadin-core:$vaadinVersion") {
        afterEvaluate {
            println(vaadin.productionMode)
            println(vaadin.productionMode.javaClass)
            println("!!!" + vaadin.productionMode.get())
            if (vaadin.productionMode.get()) {
                exclude(module = "vaadin-dev")
            }
        }
    }

    // Vaadin-Boot
    implementation("com.github.mvysny.vaadin-boot:vaadin-boot:12.2")

    implementation("org.jetbrains:annotations:24.0.1")

    // logging
    // currently we are logging through the SLF4J API to SLF4J-Simple. See src/main/resources/simplelogger.properties file for the logger configuration
    implementation("org.slf4j:slf4j-simple:2.0.9")

    // Fast Vaadin unit-testing with Karibu-Testing: https://github.com/mvysny/karibu-testing
    testImplementation("com.github.mvysny.kaributesting:karibu-testing-v24:2.1.1")
    testImplementation("org.junit.jupiter:junit-jupiter-engine:5.9.3")
    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
}

java {
    sourceCompatibility = JavaVersion.VERSION_17
    targetCompatibility = JavaVersion.VERSION_17
}

tasks.withType<Test> {
    useJUnitPlatform()
    testLogging {
        exceptionFormat = TestExceptionFormat.FULL
    }
}

application {
    mainClass = "com.vaadin.starter.skeleton.Main"
}
```

run it with `./gradlew clean -Pvaadin.productionMode` and it will still print false.

### Versions

- Vaadin / Flow version: 24.3.3

mvysny:
## Description

Expose the plugin effective configuration, so that users are able to easily exclude vaadin-dev.jar from production builds:
```kotlin
    // Vaadin
    implementation("com.vaadin:vaadin-core:$vaadinVersion") {
        afterEvaluate {
            if (vaadin.effective.productionMode.get()) {
                exclude(module = "vaadin-dev")
            }
        }
    }
```

Fixes #18572 

## Type of change

Not sure whether this is a bugfix or a feature, personally leaning towards a bugfix since `vaadin.productionMode` used to provide effective values in earlier Vaadins (24.2 for sure; it was before we moved to Properties and build caching).

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
+++ b/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
@@ -29,8 +29,9 @@ import org.gradle.api.provider.ListProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import java.io.File
+import javax.inject.Inject
 
-public abstract class VaadinFlowPluginExtension {
+public abstract class VaadinFlowPluginExtension @Inject constructor(private val project: Project) {
     /**
      * Whether we are running in productionMode or not. Defaults to false.
      * Responds to the `-Pvaadin.productionMode` property.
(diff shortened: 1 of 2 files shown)


flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
```diff
@@ -29,8 +29,9 @@ import org.gradle.api.provider.ListProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import java.io.File
+import javax.inject.Inject
```
mcollovati, line 32: Shouldn't this be `jakarta.inject.Inject`?

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/GradlePluginAdapter.kt
```diff
@@ -198,5 +198,5 @@ internal class GradlePluginAdapter(
 
     override fun isPrepareFrontendCacheDisabled(): Boolean = config.alwaysExecutePrepareFrontend.get()
 
-    override fun isReactEnabled(): Boolean = config.reactEnable.get()
+    override fun isReactEnabled(): Boolean = config.reactEnabled.get()
```
mshabarov, line 201: Parameter name should be `reactEnable`, as like `npmEnable` or `bunEnable`.

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
mcollovati: It looks like Gradle internally is still using the `javax.inject` package

mcollovati (changes requested):
We should probably revert the "fix variable names" commit, as per Mikhail [comment](https://github.com/vaadin/flow/pull/18695/files#r1494505739). Other than that, LGTM

caalador:
now the variable names are wrong and all instances should be updated to `reactEnable` as there are some that are using `reactEnable` and some that are using `reactEnabled` making the build fail

mcollovati:
@mvysny to pick this change, should also https://github.com/vaadin/flow/pull/17992 be back-ported?

mvysny:
Yup, that’s a prerequisite. If that’s a problem, we can postpone this to 24.4

mcollovati:
I'm trying to pick locally, but there are several conflicts to resolve

mvysny:
Awesome, thank you, I verify that the setting works correctly:
```
dependencies {
    // Vaadin
    implementation(libs.vaadin.core) {
        if (vaadin.effective.productionMode.get()) {
            exclude(module = "vaadin-dev")
        }
    }
}
```

PR #18698, opened 2024-02-14
feat: Wrap the JS snippet in async function

### LINKED ISSUE #18693: Support await keyword in Element.executeJs
### Describe your motivation

While hard to figure out (should be fixed with JavaDocs separately), the current API already supports returning a JS Promise, causing the PendingJavaScriptResult to be notified later on when the value is resolved. However, most code examples on earth utilizing Promise using _await_ keyword, which doesn't work with executeJs without additional hacks.

The executeJs should automatically wrap the JS body into anonymous async function, if it contains await keyword.

Provide an alternative method like Element.asyncJs that does this explicitly.

### Additional context

Prototyping in Viritin: 

https://github.com/viritin/flow-viritin/blob/3c3d603e26b2086337b7bba048da345a8956b739/src/test/java/org/vaadin/firitin/JsPromiseView.java#L124-L134

mstahv:
Wondering if anything else would change for users but await would just magically start to work 🤔 Let's see what the test say...

Quick prototype for #18693, not at all tested...

Fixes #18693

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/ExecJavaScriptView.java
+++ b/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/ExecJavaScriptView.java
@@ -20,6 +20,7 @@ import java.io.Serializable;
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.html.Input;
 import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.component.html.Span;
 import com.vaadin.flow.internal.JsonUtils;
 import com.vaadin.flow.router.Route;
 import com.vaadin.flow.uitest.servlet.ViewTestLayout;
@@ -42,6 +43,23 @@ public class ExecJavaScriptView extends AbstractDivView {
                 "console.log($0)", JsonUtils.createArray(
                         Json.create("Hello world"), Json.create(true)));
 
+        NativeButton awaitButton = createButton("await button", "awaitButton",
+                e -> e.getSource().getElement().executeJs("""
+                            const response = await fetch("/plain-script.js");
+                            const text = await response.text();
+                            return text.indexOf("Hello from src script") != -1;
+                        """).then(Boolean.class, success -> {
+                    Span span = new Span();
+                    span.setId("awaitResult");
+                    span.setText("await result: " + success);
+                    add(span);
+                }, error -> {
+                    Span span = new Span();
+                    span.setId("awaitResult");
+                    span.setText("await error: " + error);
+                    add(span);
+                }));
+
         NativeButton createElementButton = createButton(
                 "Create and update element", "createButton", e -> {
                     Input input = new Input();
@@ -51,7 +69,8 @@ public class ExecJavaScriptView extends AbstractDivView {
                     add(input);
                 });
 
-        add(alertButton, focusButton, swapText, logButton, createElementButton);
+        add(alertButton, focusButton, swapText, logButton, createElementButton,
+                awaitButton);
     }
 
     private NativeButton createJsButton(String text, String id, String script,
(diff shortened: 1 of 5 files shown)


Legioth:
This seems like a good and safe approach at least from a theoretical point of view. Let's hope the tests agree 😄

mstahv:
All checks passed, what could possibly go wrong🤷‍♂️ 6 characters, coded with browser and a great DX enhancment 😎

mstahv:
Some simple integration test to verify await now works, and more essentially, keeps working in the future would be great still.

mstahv:
Somethign like this ought to be enough, but @vaadin/flow team probably knows better where to put it and can do it in a fraction of what it would take for me:

            getElement().executeJs("""
                        const response = await fetch("/");
                        const text = await response.text();
                        return text.indexOf("html") != -1 
                    """).then(Boolean.class, success -> {
                        if(!success) {
                            // fail test here
                        }
            });

mcollovati (commented):
`Page.executeJs()` does not wrap the expression. Should it also be changed to support the `await` keyword, or in that case, is it up to the developer to handle it in some way?

mcollovati:
I added a simple IT test. @mstahv please take a look to check if it makes sense to you

mstahv:
The test looks good to me! Never used executeJs from Page IIRC, might need something here if we want same feature there:

https://github.com/vaadin/flow/blob/main/flow-server/src/main/java/com/vaadin/flow/server/communication/UidlWriter.java#L360-L367

mcollovati:
Good catch! If you don't mind, I can apply the change and add a test for `Page.executeJS()`

mstahv:
Go ahea! I don't even have this branch checked out, did the original change with GitHub directly 😁

mstahv:
Here is an updated JavaDoc that also highlights that Promise's can be returned. This is currently too well hidden:


```
Asynchronously runs the given JavaScript expression in the browser in the
context of this element. The expression is executed in an <code>async</code> JavaScript method so you can utilize <code>await</code> syntax when consuming JavaScript API returning a Promise. The returned
<code>PendingJavaScriptResult</code> can be used to retrieve the
<code>return</code> value from the JavaScript expression. If a Promise is returned in the JavaScript expression, PendingJavaScriptResult will report the resolved value once it becomes available. If no return
value handler is registered, the return value will be ignored.
```

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/ExecJavaScriptView.java
```diff
@@ -42,6 +44,41 @@ protected void onShow() {
                 "console.log($0)", JsonUtils.createArray(
                         Json.create("Hello world"), Json.create(true)));
 
+        NativeButton elementAwaitButton = createButton("Element await button",
+                "elementAwaitButton",
+                e -> e.getSource().getElement().executeJs("""
+                            const response = await fetch("/plain-script.js");
```
Legioth, line 50: Could it be less fragile to use `setTimeout` instead so that the tested logic wouldn't depend on yet another network request?

The script could then be along the lines of this:
```
const result = new Promise((resolve) => {
  setTimeout(() => resolve(42), 10);
});
return await result;
```

mcollovati: Thanks Leif, I'll update the test

mstahv: Yup, less moving parts that way.

mcollovati: done

mstahv:
Thanks @mcollovati, looks very good to me now 💪 Would rubber stamp this if I only could, but GH considers this my PR now 🤷‍♂️

mcollovati:
It's yours PR :smile: 
I'll ask for an additional review from someone else from the Flow team

tepi (approved):

PR #18700, opened 2024-02-14
feat: Add methods for checking roles and authorities

peholmst:
This PR adds methods for checking the user's roles and authorities. The purpose is to make the `AuthenticationContext` more in line with the `useAuth` hook in Hilla and reduce the need for implementing a custom "security service". In many cases, it should be enough to inject and use the `AuthenticationContext` directly.


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
```diff
@@ -63,6 +69,8 @@ public class AuthenticationContext {
 
     private CompositeLogoutHandler logoutHandler;
 
+    private static final String ROLE_PREFIX = "ROLE_";
```
knoobie, line 72: That value is not really static :) See https://github.com/vaadin/flow/blob/a130d64a9077cdedcbe41fa3eba4d781a0146e35/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java#L194

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
```diff
@@ -140,6 +147,165 @@ public void logout() {
...
+     *         roles, otherwise {@literal false}.
+     * @throws IllegalArgumentException
+     *             if the given array is empty.
+     */
+    public boolean hasAnyRole(String... roles) {
+        return hasAnyRole(Stream.of(roles));
+    }
+
+    private boolean hasAnyRole(Stream<String> roles) {
+        return hasAnyAuthority(roles.map(role -> getRolePrefix() + role));
```
knoobie, line 191:
```suggestion
        var rolePrefix = getRolePrefix();
        return hasAnyAuthority(roles.map(role -> rolePrefix + role));
```

hasAnyRole and hasAllRoles can be quite the hot-path; therefore I would reduce the amount of calculation needed to re-create the role prefix multiple times in the same request/check - especially with hundreds of roles..

While thinking about hot-path: Is the usage of stream here really necessary? (Up for others to decide)

peholmst: That was a good catch, I hadn't noticed it. I've now rewritten the checks to be more efficient.

vaadin-spring/src/test/java/com/vaadin/flow/spring/security/AuthenticationContextTest.java
```diff
@@ -130,6 +130,202 @@ public void getPrincipalName_loggedUser_getsAuthenticationName() {
...
+    public void hasAnyRole_anonymous_false() {
+        Assert.assertFalse(authContext.hasAnyRole("ANONYMOUS"));
+        Assert.assertFalse(authContext.hasAnyRole(List.of("ANONYMOUS")));
+    }
+
+    @Test
+    @WithMockUser(roles = { "USER", "ADMIN" })
+    public void hasAnyRole_hasRole_true() {
+        Assert.assertTrue(authContext.hasAnyRole("USER"));
+        Assert.assertTrue(authContext.hasAnyRole(List.of("ADMIN")));
```
mcollovati, line 174: What about adding a test with both assigned and not assigned roles?
e.g.
```
Assert.assertTrue(authContext.hasAnyRole(List.of("ADMIN", "SUPERADMIN")));
```

vaadin-spring/src/test/java/com/vaadin/flow/spring/security/AuthenticationContextTest.java
```diff
@@ -130,6 +130,202 @@ public void getPrincipalName_loggedUser_getsAuthenticationName() {
...
+        Assert.assertFalse(authContext.hasAnyAuthority("ROLE_ANONYMOUS"));
+        Assert.assertFalse(
+                authContext.hasAnyAuthority(List.of("ROLE_ANONYMOUS")));
+    }
+
+    @Test
+    @WithMockUser(authorities = { "AUTH_READ", "AUTH_WRITE" })
+    public void hasAnyAuthority_hasAuthority_true() {
+        Assert.assertTrue(authContext.hasAnyAuthority("AUTH_READ"));
+        Assert.assertTrue(authContext.hasAnyAuthority(List.of("AUTH_WRITE")));
```
mcollovati, line 268: What about adding a test with both assigned and not assigned authorities?
e.g.

```
Assert.assertTrue(authContext.hasAnyAuthority(List.of("AUTH_WRITE", "AUTH_MANAGE")));
```

vaadin-spring/src/test/java/com/vaadin/flow/spring/security/AuthenticationContextTest.java
mcollovati, line 1: I would add some tests as well for `getGrantedRoles()` and `getGrantedAuthorities()`, since there's some logic involved.
However, they may be already indirectly verified by the other tests.

mcollovati (changes requested):
Some nits, but other than that, look good to me

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
```diff
@@ -140,6 +143,207 @@ public void logout() {
...
+                    "Must provide at least one authority to check");
+        }
+        var grantedAuthorities = getGrantedAuthorities().stream()
+                .map(GrantedAuthority::getAuthority)
+                .collect(Collectors.toSet());
+        for (var authority : authorities) {
+            if (grantedAuthorities.contains(authority)) {
+                return true;
+            }
+        }
```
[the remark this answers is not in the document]
peholmst: My first version of this PR used that, but it was questioned by @knoobie . Which one do you want, the one with the loop or the one with the stream?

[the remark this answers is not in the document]
knoobie: I think he meant that it is probably unnecessary to use a for-loop after the stream was terminated.. so now we are creating s stream, closing it and afterwards iterating over it. The closing (toSet) and for-loop could theoretically be replaced with anyMatch if I understand Marco's intention correctly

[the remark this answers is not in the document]
mcollovati: Exactly. @knoobie explained clearly my ambiguous comment :smile:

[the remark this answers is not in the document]
peholmst: Roger that.

mcollovati (approved):
LGTM. Thanks @peholmst for adding this feature!

mcollovati:
I'll create a ticket on docs repository to add a note about this new utility methods in https://vaadin.com/docs/latest/security/enabling-security#security-utilities

PR #18701, opened 2024-02-14
fix: Possibly missing packages should not show error

caalador:
When getting outside package json
files we should not log errors for
them if they are missing as they may
or may not be availabe.

Fixes vaadin/platform#5023

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -363,8 +363,8 @@ public abstract class NodeUpdater implements FallibleCommand {
         defaults.putAll(readDependencies("vite", "devDependencies"));
         putHillaComponentsDependencies(defaults, "devDependencies");
         if (options.isReactEnabled()) {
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -337,21 +336,35 @@ JsonObject readPackageJson(String id) throws IOException {
...
+
     Map<String, String> getDefaultDevDependencies() {
         Map<String, String> defaults = new HashMap<>();
         defaults.putAll(readDependencies("default", "devDependencies"));
         defaults.putAll(readDependencies("vite", "devDependencies"));
         putHillaComponentsDependencies(defaults, "devDependencies");
         if (options.isReactEnabled()) {
-            defaults.putAll(
-                    readDependencies("react-router", "devDependencies"));
+            defaults.putAll(readDependenciesIfAvailable("react-router",
```
mshabarov, line 366: This isn't an external package.json, why is this needed?

mshabarov (dismissed):

mshabarov (approved):

Issue #18684, opened 2024-02-13
Vaadin 24.4 adds Vaadin web-components into package.json by default


mshabarov:
### Description of the bug

Vaadin 24.4 should add only react-components into `package.json` and shouldn't add web-components there by default until one set's up `react.enable=false` flag. 
However, this isn't true with the latest version, both web-compoents and react-components are added, causing [auto-completion problems ](https://github.com/vaadin/flow/issues/18638#issuecomment-1938354810).

### Expected behavior

React-components only should be added by default and only web-components should be added in `react.enable=false`.

### Minimal reproducible example

https://github.com/vaadin/flow-hilla-hybrid-example/tree/v24.4

### Versions

- Vaadin / Flow version: 24.4 pre-release.

mshabarov:
Here is a short summary on what we're going to implement to ensure the needed components are present in the `package.json` and IDE suggests only what is in `package.json`:
1. Add a feature in Flow that would exclude a given list of packages from resulted `package.json` if any package entry in `versions.json` with `npmName` has this list under `exclusions` key.
2. Add an `exclusions` key to `@react-components` JSON in platform’s `versions.json` as follows:
```
"react": {
        "react-components": {
            "jsVersion": "24.4.0-alpha13",
            "npmName": "@vaadin/react-components",
            "mode": "react",
            "exclusions": [
                  "@vaadin/a11y-base",
                  "@vaadin/accordion",
                  ...
            ]
        }
    }
```
This drops the idea to remove `@NpmPackage` annotations from Flow components, as this might be a breaking changes, see e.g. this [comment](https://github.com/vaadin/flow-components/pull/6079#discussion_r1500443133).

Then, if a development of frontend goes with Lit, one can fallback to `reactEnable=false` and have only web-components in `package.json` ensuring only them are suggested.

If one wants to develop with both React and Lit, he fallbacks to `reactEnable=false` and adds the following package:
```
@NpmPackage(value = "@vaadin/react-components", version = "24.4.0")
```

PR #18707, opened 2024-02-14
fix: include only react-components in package.json

tltv:
When React is enabled, package.json update includes only @vaadin/react-components and excludes all other @vaadin/* web-components pinned by platform versions (core and pro). React components are wrappers for web-components. When React is disabled, only web-components are updated in the package.json.

Fixes: #18684

Note! [This PR in platform](https://github.com/vaadin/platform/pull/5076) must be merged before this PR.


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -221,9 +232,12 @@ private boolean updatePackageJsonDependencies(JsonObject packageJson,
         /*
          * #10572 lock all platform internal versions
          */
-        List<String> pinnedPlatformDependencies = new ArrayList<>();
-        final JsonObject platformPinnedDependencies = getPlatformPinnedDependencies();
         for (String key : platformPinnedDependencies.keys()) {
+            if (options.isReactEnabled() && isReactModuleAvailable()
+                    && !FrontendUtils.isPackageKeyReactComponents(key)
+                    && pinnedPlatformDependencies.contains(key)) {
```
caalador, line 238: the contains(key) is always true here as we are iterating the same set of keys as added to the pinnedPlatformDependencies

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -205,8 +206,18 @@ private boolean updatePackageJsonDependencies(JsonObject packageJson,
             Map<String, String> applicationDevDependencies) throws IOException {
         int added = 0;
 
+        List<String> pinnedPlatformDependencies = new ArrayList<>();
+        final JsonObject platformPinnedDependencies = getPlatformPinnedDependencies();
```
caalador, line 210: Now that the platformPinnedDependencies already filters out the react-components if they are not needed this place only needs to check if the react-components key  `@vaadin/react-components` is available in the Json.
If yes then ignore old components else do not ignore.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -205,8 +206,18 @@ private boolean updatePackageJsonDependencies(JsonObject packageJson,
...
+        Collections.addAll(pinnedPlatformDependencies,
+                platformPinnedDependencies.keys());
+
         // Add application dependencies
         for (Entry<String, String> dep : applicationDependencies.entrySet()) {
+            if (options.isReactEnabled() && isReactModuleAvailable()
+                    && !FrontendUtils.isPackageKeyReactComponents(dep.getKey())
+                    && pinnedPlatformDependencies.contains(dep.getKey())) {
+                continue;
+            }
```
caalador, line 220: Trusting the `platformPinnedDependencies` content as `boolean reactComponents = pinnedPlatformDependencies.contains("@vaadin/react-components");` would make this:
```suggestion
            if (reactComponents 
                    && pinnedPlatformDependencies.contains(dep.getKey())) {
                continue;
            }
```

Also is this no here just to not add the `@NpmPackage` annotations from server Component definitions?

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -1323,4 +1323,15 @@ private static boolean mayHaveClientSideRoutes(String routesContent) {
...
+    /**
+     * Checks if the given key is a package key for a Vaadin react components.
+     *
+     * @param key
+     *            the key to check
+     * @return {@code true} if the key is a package key for a Vaadin react
+     *         components, {@code false} otherwise
+     */
+    static boolean isPackageKeyReactComponents(String key) {
+        return key.equals("@vaadin/react-components");
```
caalador, line 1335: As a string check this sounds a bit easy to fail.
A better way would be to collect all the possibilities in the `versions.json` found under the `react` (perhaps in the future `react-pro`) object which is formatted as:
```
    "react": {
        "react-components": {
            "jsVersion": "24.4.0-alpha13",
            "npmName": "@vaadin/react-components"
        }
    },
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
tltv: Yes, basically so. Without this just having vaadin-core dependency would end up adding all web-components. Or having any dependency with `@NpmPackage` with core components.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -286,6 +298,16 @@ && pinPlatformDependency(packageJson,
                 || !oldHash.equals(newHash);
     }
 
+    private boolean isReactModuleAvailable() {
```
caalador, line 301: This is now not needed as it's not used anywhere in the class

caalador (commented):
After removing the unused method it looks fine.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
tltv: removed

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
```diff
@@ -221,9 +233,11 @@ private boolean updatePackageJsonDependencies(JsonObject packageJson,
         /*
          * #10572 lock all platform internal versions
          */
-        List<String> pinnedPlatformDependencies = new ArrayList<>();
-        final JsonObject platformPinnedDependencies = getPlatformPinnedDependencies();
         for (String key : platformPinnedDependencies.keys()) {
+            if (reactComponents
+                    && !FrontendUtils.isPackageKeyReactComponents(key)) {
+                continue;
+            }
```
mshabarov, line 240: This looks dangerous and probably not correct, as it skips adding everything from platform's versions json that isn't `@vaadin/react-components`. Besides web-component, versions json has lumo, polymer, bundles, statistics and potentially other "non-component" dependencies that are skipped with this codes.

Also, if I run this patch with the existing package.json containing the web-components, this patch doesn't remove them from, probably because the logic below in this method isn't running, which is incorrect.

But from the other side, how Flow can only extract semantically web-components entries...Should we change the structure of `versions.json` in the platform, so it would store web-components in a separated sub-list?

mshabarov (changes requested):
The case where package.json isn't exist looks working well, but if package.json exists before starting the app and it has WCs, then they aren't removed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java
tltv: It's not removing existing dependency as that's how the package.json update works in general. but if you think, it's safe to remove existing, it could be done too. Just not adding it is probably safer.

mshabarov: Well, retaining them would require package.json removal during migration to effectively trigger this feature, but this is probably fine. Maybe it's indeed safer to only not adding them.

Having a sub-list is discussible, see the corresponding discussion in Slack.

mshabarov: My main concern vs merging this change is these messages in the logs:
```
2024-02-19T13:28:30.526+02:00  INFO 10210 --- [onPool-worker-1] c.v.f.server.frontend.TaskUpdateImports  : 

  Failed to find the following imports in the `node_modules` tree:
      - @vaadin/polymer-legacy-adapter/style-modules.js
  If the build fails, check that npm packages are installed.

  To fix the build remove `package-lock.json` and `node_modules` directory to reset modules.
  In addition you may run `npm install` to fix `node_modules` tree structure.

@vaadin/bundles npm package is not found, Vaadin component dependency bundles are disabled.

 Error: The following dependencies are imported but could not be resolved:
2024-02-19T13:28:36.541+02:00  INFO 10210 --- [v-server-output] c.v.b.devserver.DevServerOutputTracker   : 
2024-02-19T13:28:36.541+02:00  INFO 10210 --- [v-server-output] c.v.b.devserver.DevServerOutputTracker   :   @vaadin/polymer-legacy-adapter/style-modules.js
```

These shouldn't be logged at least if we think that there is no impact on applications.

mshabarov: Let's put on hold this PR until we decide about a solution to add/remove only components.

mshabarov:
We have to do the following:

1. Add a new parameter/option into entries in `version.json` in platform with values `react` / `lit` / `all` (or empty) and key "mode", e.g.:
```
       "accordion": {
            "javaVersion": "{{version}}",
            "jsVersion": "24.4.0-alpha13",
            "npmName": "@vaadin/accordion",
            "mode": "lit"
        }
```
2. Remove `@NpmPackage` annotations from Flow components importing web-components packages.
3. Extend `TaskUpdatePackages` with a logic that dynamically adds the components depending on the project config (React/Lit).

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
```diff
@@ -68,6 +73,9 @@ private void collectDependencies(JsonObject obj) {
                 continue;
             }
             JsonObject json = (JsonObject) value;
+            if (!isIncludedByMode(json)) {
```
mshabarov, line 76: Now with this new method the old approach above `(!reactEnabled && isReactObject(key))` isn't needed anymore.
Flow should add or not to add the npm dependency based on this mode value.

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
```diff
@@ -38,6 +38,11 @@ class VersionsJsonConverter {
     private static final String JS_VERSION = "jsVersion";
     private static final String NPM_NAME = "npmName";
     private static final String NPM_VERSION = "npmVersion";
+    private static final String MODE = "mode";
```
mshabarov, line 41: Better to make it private so platform's codes can reuse it if necessary.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -1323,4 +1325,15 @@ private static boolean mayHaveClientSideRoutes(String routesContent) {
...
 
+    /**
+     * Checks if the given key is a package key for a Vaadin react components.
+     *
+     * @param key
+     *            the key to check
+     * @return {@code true} if the key is a package key for a Vaadin react
+     *         components, {@code false} otherwise
+     */
+    static boolean isPackageKeyReactComponents(String key) {
```
mshabarov, line 1336: Looks like this isn't needed anymore.

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
mshabarov: Also IMO it would be better to structure it as:
```java
if (isIncludedByMode(json)) {
            if (json.hasKey(NPM_NAME)) {
                addDependency(json);
            } else {
                collectDependencies(json);
            }
}
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
```diff
@@ -76,6 +84,20 @@ private void collectDependencies(JsonObject obj) {
...
+        if (!json.hasKey(MODE)) {
+            return true;
+        }
+        String mode = json.getString(MODE);
+        if (mode == null || mode.isBlank() || MODE_ALL.equalsIgnoreCase(mode)) {
+            return true;
+        } else if (reactEnabled) {
+            return MODE_REACT.equalsIgnoreCase(mode);
+        } else {
+            return MODE_LIT.equalsIgnoreCase(mode);
```
mshabarov, line 97: I propose to add a debug logging to each cases, showing the package name and it's version.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskUpdatePackagesNpmTest.java
```diff
@@ -668,7 +785,13 @@ private void createVaadinVersionsJson(String dialogVersion,
                         + "    \"jsVersion\": \"" + overlayVersion + "\",\n"
                         + "    \"npmName\": \"" + VAADIN_OVERLAY + "\",\n"
                         + "    \"releasenotes\": true\n"
-                        + "}}},\n";//@formatter:on
+                        + "}},\n"
+                        + "\"react\": {\n" +
+                        "        \"react-components\": {\n" +
+                        "            \"jsVersion\": \"24.4.0-alpha13\",\n" +
+                        "            \"npmName\": \"@vaadin/react-components\"\n" +
```
mshabarov, line 792: this needs a `react` mode ?

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
tltv: Current code does not require "mode: react" for react dependencies. If we change this so that "mode: react" is needed in version.json, then we need to add it in platform versions.json and wait for next snapshot before we merge this PR (this takes more time naturally).
If we keep it as it is, we can merge this/platform independently. Which path should we take?

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
tltv: I assume you mean public.

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
tltv: Discussed and decided to add react mode in platform and this means that [platform PR needs to be merged first](https://github.com/vaadin/platform/pull/5076) before this PR can be merged (after related changes). In other words, we will not depend on "react" and "react-pro" keys anymore. We use "mode: react" feature instead.

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
tltv: done (refactored)

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
mshabarov: Ah yes! Public :)

mshabarov (changes requested):
LGTM, except that for some reason this PR gets not-related changes from other commits.
Needs a clean-up and rebase.

tltv:
> LGTM, except that for some reason this PR gets not-related changes from other commits. Needs a clean-up and rebase.

It's just how github shows commits _after_ rebasing. So it's already rebased.

mshabarov:
I see some other commits in your branch. I propose to revert them and squash all the commits. Then merge latest main.
Also this PR has some conflicts.

mshabarov (approved):

PR #18710, opened 2024-02-15
feat: Move ReactComponentAdapter to flow-react

### LINKED ISSUE #18699: Extract all React related logic into a new module flow-react
### Describe your motivation

Currently, flow-server module has React-specific logic (ReactAdapter, ReactAdapterComponent and other files), this deserves a separate module in Flow, so that you can exclude it if you want.

New flow-react maven module in Flow, containing React Component wrapper classes.

### Additional context

Vaadin 24.4.

caalador:
Move the React component code
to new module `flow-react`

Fixes #18699


tltv:
Looks good to me. `flow-react` module doesn't look like it's included by default anywhere now, is this also in purpose? 
It could be added by default in platform.

caalador:
It is added by default into `flow` module which is added in platform into the versions. Platform will get a Pr for adding the module to core-internal

tltv (approved):

PR #18717, opened 2024-02-15
feat: Only add react-components with flow-react

### LINKED ISSUE #18676: Exclude react components from package.json if not needed
### Description of the bug

When I develop a flow application without any usage of react components, I expect no react components to be added to my package.json - this includes the free and the commercial components. Even tho it's possible to disable react completely by using reactEnabled=false, this would result in some kind of legacy mode which is not wanted in the long run. 

### Expected behavior

When I exclude hilla and only use flow's default, no react components should be present in the application's package.json for compliance reasons. 

Suggestions: 
- add the react components annotation to a flow- react package which I can exclude 
- add the react components to the hilla-react spring boot starter which I can exclude 

Any 24.4 version. 

### Versions

- Vaadin / Flow version: 24.4+

caalador:
If flow-react module is not
available do not add react-components
into package.json

Fixes #18676

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
@@ -175,13 +175,23 @@ public abstract class NodeUpdater implements FallibleCommand {
             VersionsJsonConverter convert = new VersionsJsonConverter(
                     Json.parse(
                             IOUtils.toString(content, StandardCharsets.UTF_8)),
-                    options.isReactEnabled());
+                    options.isReactEnabled() && isReactModuleAvailable());
             versionsJson = convert.getConvertedJson();
             versionsJson = new VersionsJsonFilter(getPackageJson(),
                     DEPENDENCIES)
                     .getFilteredVersions(versionsJson, versionsOrigin);
         }
         return versionsJson;
+    }
+
+    private boolean isReactModuleAvailable() {
+        try {
+            options.getClassFinder().loadClass(
+                    "com.vaadin.flow.component.react.ReactAdapterComponent");
+            return true;
+        } catch (ClassNotFoundException e) {
+            return false;
+        }
     }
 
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -175,7 +175,7 @@ private JsonObject getFilteredVersionsFromResource(URL versionsResource,
             VersionsJsonConverter convert = new VersionsJsonConverter(
                     Json.parse(
                             IOUtils.toString(content, StandardCharsets.UTF_8)),
-                    options.isReactEnabled());
+                    options.isReactEnabled() && isReactModuleAvailable());
```
knoobie, line 178: Probably out of scope; just in case: The VersionJsonConverter has weird naming for reactEnabled vs. collectReactComponents -> and there is still react and react-pro

caalador: Well for the converter the boolean says that should it collect react based components (not react-components per say) or not.
Also at least for the moment the versions contains
```
    "react": {
        "react-components": {
            "jsVersion": "24.4.0-alpha13",
            "npmName": "@vaadin/react-components"
        }
    },
```
and to my understanding the components might get singled out (they would be named under react) or at least put into core and pro so `react` and `react-pro`.

knoobie: You might be right, you are closer to the informations 🙂 Just remembered the ticket that free and commercial react components were merged https://github.com/vaadin/react-components/issues/235 and I was wondering if this was just a left over / over sight in the class in flow

caalador: Well there has been no information on it making any changes to the platform versions file so it doesn't concern this part as long as there are no decisions on changing the structure.

tltv (approved):

Issue #18656, opened 2024-02-08
Default development and production bundle ignores react.router.enabled=false


tltv:
Vaadin's default development and production bundle packages always `index.ts` by copying content from Flow's `index-react.ts` with react router. This is expected so as bundle can have just one router option at the moment. Setting `react.router.enabled=false` doesn't change this. React router is enabled even when disabled with this property.

At minimum, new local bundle should be built automatically when _react router is disabled and default bundle would be used otherwise_. This would ensure that vaadin-router is used instead as very likely is expected with `react.router.enabled=false`.

mshabarov:
A solution would be:
1. To re-build the bundle to not have a react-related content there, or
2. Generate index.ts content conditionally based on this flag and not include into bundle

PR #18718, opened 2024-02-15
chore: rebuild bundle if React is disabled

tltv:
Default production and development bundles are packaged with react router only. With this change, bundle is always rebuild if React is disabled and default bundle would be used otherwise. Rebuilding ensures that Vaadin router is used instead of the React router.

Fixes: #18656

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
@@ -117,6 +117,7 @@ public final class BundleValidationUtil {
             getLogger().info("No dev-bundle found.");
             return true;
         }
+
         if (!DevBundleUtils
                 .getDevBundleFolder(npmFolder, options.getBuildDirectoryName())
                 .exists() && compressedDevBundle.exists()) {
@@ -125,6 +126,7 @@ public final class BundleValidationUtil {
                             new File(npmFolder,
                                     options.getBuildDirectoryName()),
                             Constants.DEV_BUNDLE_LOCATION));
+
         }
 
         if (options.isSkipDevBundle()) {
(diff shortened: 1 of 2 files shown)


flow-server/src/test/java/com/vaadin/flow/server/frontend/BundleValidationTest.java
```diff
@@ -2121,6 +2152,37 @@ public void defaultProdBundleExists_noCompressedProdBundleFile_noBuildRequired()
                 needsBuild);
     }
 
+    @Test
+    public void defaultProdBundleExists_noCompressedProdBundleFile_reactDisabled_noBuildRequired()
+            throws IOException {
```
caalador, line 2157:
```suggestion
    public void defaultProdBundleExists_noCompressedProdBundleFile_reactDisabled_buildRequired()
            throws IOException {
```
should be build required. Perhaps `prodBuildRequired`

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
```diff
@@ -135,6 +137,15 @@ private static boolean needsBuildDevBundle(Options options,
             return false;
         }
 
+        if (!DevBundleUtils
+                .getDevBundleFolder(npmFolder, options.getBuildDirectoryName())
+                .exists() && !options.isReactEnabled()) {
+            // Using default dev bundle in a Jar.
+            // Default dev bundle is packaged with react router only. With react
+            // disabled, let's rebuild bundle to get vaadin router instead.
```
caalador, line 145: could have an info log that `Bundle build required for non default router.` or something like that

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
```diff
@@ -152,6 +163,15 @@ private static boolean needsBuildDevBundle(Options options,
     private static boolean needsBuildProdBundle(Options options,
             FrontendDependenciesScanner frontendDependencies)
             throws IOException {
+
+        if (!options.isReactEnabled() && !ProdBundleUtils
+                .getProdBundle(options.getNpmFolder()).exists()) {
+            // Using default prod bundle in a Jar.
+            // Default prod bundle is packaged with react router only. With
+            // react disabled, let's rebuild bundle to get vaadin router
+            // instead.
```
caalador, line 172: could have an info log that `Bundle build required for non default router.` or something like that

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
tltv: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/BundleValidationUtil.java
tltv: done

caalador (approved):

PR #18733, opened 2024-02-16
fix: Fix compile errors for router import

mshabarov:
Should be `{ router } ` instead of ` router `.

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/index-react.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/index-react.tsx
@@ -13,7 +13,7 @@
 import { createElement } from 'react';
 import { createRoot } from 'react-dom/client';
 import { RouterProvider } from 'react-router-dom';
-import { router } from "Frontend/routes";
+import { router } from 'Frontend/routes.js';
 
 function App() {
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/index-react.tsx
```diff
@@ -13,7 +13,7 @@
 import { createElement } from 'react';
 import { createRoot } from 'react-dom/client';
 import { RouterProvider } from 'react-router-dom';
-import router from 'Frontend/routes.js';
+import { router } from 'Frontend/routes.js';
 
 function App() {
     return <RouterProvider router={router} />;
```
tltv, line 19: If import can't be `.js` file, what is there to import from the routes.ts other than `routes` variable? And what `RouterProvider` expects to get in, is it any object with a `routes` variable or just routes `variable`?

mshabarov: No, I was mistaken, it can be .js, but should be in curly brackets.

mshabarov:
Depending on how `routes` are exported in `routes.tsx` (const or default), the import may or may not require the brackets.
Will double check further what is the usual use case.

PR #18749, opened 2024-02-19
feat: Call copilot init feature when starting from an empty app

Artur-:


mcollovati (commented):
I tested both Flow and Hilla view generation and got source code generated, but I noticed the following (perhaps expected?):

- starting application from the IDEA, after creating Flow view, the server is restarted, browser page is reloaded, and I can see the new view
- starting application from the IDEA, after creating Hilla view the server is not restarted, the browser page reloads, but still `NoRoutesError_dev.html` is shown. After manual restart, I can see the view.
- starting application with `mvn` from terminal, server is not restarted neither for Flow nor for Hilla view. After manual restart, I can see the view

Artur-:
>starting application from the IDEA, after creating Flow view, the server is restarted, browser page is reloaded, and I can see the new view

Sounds like this is working as it should

> starting application from the IDEA, after creating Hilla view the server is not restarted, the browser page reloads, but still NoRoutesError_dev.html is shown. After manual restart, I can see the view.

There might be several cases here and I am not sure which one you had or what project you started from:
1. If you have an empty project from start.spring.io then it does not even have spring boot dev tools and will not restart at all automatically.
2. If you have an empty project + dev tools then it should restart automatically as a Theme annotation is added to the application class
3. If you have a project that already has a theme, it is possible that it does not change any Java files and that Flow will not pick up frontend changes as it is running with a dev bundle(?)

> starting application with mvn from terminal, server is not restarted neither for Flow nor for Hilla view. After manual restart, I can see the view

This also depends on the case. If you don't have any IDE running, then nothing will compile the files and no restart will take place.

mcollovati:
I created a test project with `npx @vaadin/cli init --empty my-app` and removed the example view.
I'll test again, removing also the theme stuff.

mcollovati:
Starting with an application without a theme and creating a Hilla view works, but it took three restarts.
After the first one, the page was reloaded showing the `NoRoutesError_dev.html`, then after a couple of seconds (and two restarts) I could see the "frontend build" page and eventually the Hilla view.

Artur-:
Indeed, it is not optimal but a start

mcollovati:
It would be nice to improve `NoRoutesError_dev.html` telling the developer that the action make take a couple of seconds, and that the server should be manually restarted if not running from IDE (or if nothing happens after e.g. 5 seconds)

Artur-:
Yes, but I think those changes should go on the copilot side. Similarly to how it adds a message if you don't have spring boot dev tools in the project

PR #18759, opened 2024-02-20
fix: use Vite @vitejs/plugin-react also in production

platosha:
Vite's esbuild JSX output in production has some incompatibility with the Preact Signals library. As a workaround, this change applies JSX transformation provided by `@vitejs/plugin-react` in production mode.

Required by: https://github.com/vaadin/hilla/pull/2043

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -308,7 +308,7 @@ function statsExtracterPlugin(): PluginOption {
 
       const isFrontendResourceCollected = (id: string) =>
           !id.startsWith(themeOptions.frontendGeneratedFolder.replace(/\\/g, '/'))
-          || isThemeComponentsResource(id) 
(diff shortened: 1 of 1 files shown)


Artur- (approved):

flow-server/src/main/resources/vite.generated.ts
```diff
@@ -772,14 +772,16 @@ export const vaadinConfig: UserConfigFn = (env) => {
         ]
       }),
       // The React plugin provides fast refresh and debug source info
-      devMode && reactPlugin({
+      reactPlugin({
```
mshabarov, line 775: Just for clarification to me, random concern: wouldn't this lead to a leaking of live-reload or debugging tools somehow to end users in production?

platosha: Should not be the case. The only concern is our custom additions, so I have disabled the transformation added by Flow in production.

Issue #18681, opened 2024-02-12
Vaadin stateless jwtAuth should allow custom claims to be added instead of just subject and roles.


ilyas2016:
### Describe your motivation

One of use case to use JWT token to skip extra DB call for each request. like getting user id, or some other details liked. let say i am working for a hotel management system and for the manager role. i need the hotel id, then i will prefer to set hotel id in token, because It will prevent extra db call for each operation.

### Describe the solution you'd like

we can modify JwtSecurityContextRepository##encodeJwt method  and add extra claim just below the roles with Key "details" and pass the Authentication##getDetails. following this way programmer can create his own sub class of org.springframework.security.core.userdetails.UserDetails and returning it from UserDetailsService##loadUserByUsername method. whichever custom fields user created in his class will be available in jwt toke.

PR #18769, opened 2024-02-21
Added userDetails object to jwt claims

ilyas2016:
## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/JwtSecurityContextRepository.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/JwtSecurityContextRepository.java
@@ -60,6 +60,7 @@ import org.springframework.security.web.context.SecurityContextRepository;
 class JwtSecurityContextRepository implements SecurityContextRepository {
     private static final String ROLES_CLAIM = "roles";
     private static final String ROLE_AUTHORITY_PREFIX = "ROLE_";
+    private static final String DETAILS_CLAIM = "details";
     private final Log logger = LogFactory.getLog(this.getClass());
     private final SerializedJwtSplitCookieRepository serializedJwtSplitCookieRepository;
     private final JwtAuthenticationConverter jwtAuthenticationConverter;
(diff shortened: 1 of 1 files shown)


mcollovati:
Releated to #18681

mcollovati:
Hi @ilyas2016, thank you for creating the ticket and for the contribution.

We discussed internally about the feature and the proposed change, and we think that always adding authentication details claim may not be the best option for all projects.

Adding custom claims is indeed a useful feature, so I propose a different design that should work for different needs:

* allow `JwtSecurityContextRepository` to be configured with a component that creates claims based on `Authentication` object (e.g. a `Function<Authentication, Map<String, Object>>`) and use it in `encodeJwt` method
* add the method `withCustomClaims(Function<Authentication, Map<String, Object>>)` in `VaadinStatelessSecurityConfigurer`. The value will be propagated to `JwtSecurityContextRepository` in `VaadinStatelessSecurityConfigurer.configure()` (as we currently do for `expiresIn`, for example)
* add the method `VaadinWebSecurity.setStatelessAuthentication(HttpSecurity, Customizer<VaadinStatelessSecurityConfigurer>)` in `VaadinWebSecurity`, so the developer will have more control on `VaadinStatelessSecurityConfigurer`
* refactor existing `setStatelessAuthentication` methods to use the version with the `Customizer`

In addition, we should add necessary tests (e.g. in `JwtSecurityContextRepositoryTest`) to ensure the feature works as expected.

I know this is quite a bigger effort in respect to the current PR, but it will provide more flexibility and freedom for the developer to add only the claims he needs.
If you are interested in implementing the feature, we will be glad to support you both with guidance and code contribution.
If you instead prefer not to go ahead, we can take it over, depending on the Flow team planning.

mcollovati:
I'll close this PR as the change will require additional efforts to be completed, and the related issue currently has a not high priority in the team backlog.
I'll update the ticket to link the proposal solution.

@ilyas2016 thank you again for creating the issue and the PR.
Please let us know with a comment here if you decide to proceed with the implementation.

PR #18772, opened 2024-02-21
feat: Support top-level await

### LINKED ISSUE #18575: Unable to use top level await in projects
### Description of the bug

According to https://github.com/vaadin/platform/releases/tag/24.0.0 24.x supports Safari 15.4+ and evergreen versions of all other browsers. All of these support top level await but if you try to use it in a project, you get
```
VAADIN/build/indexhtml-!~{001}~.js:5066:27: ERROR: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
```

The target environment seems a bit off, with the wrong safari major and old versions of all the other browsers

### Expected behavior

You can use top level await.

Like you can today by manually adding
```
  build: {
    target: ["esnext", "safari15"],
  },
```
to vite.config.ts

Create a new project, add e.g. 
```
await import("./other.js");
```
in index.js

### Versions

- Vaadin / Flow version: 24.4
- Java version:
- OS version:
- IDE (if applicable):

Artur-:
Fixes #18575

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -719,6 +719,7 @@ export const vaadinConfig: UserConfigFn = (env) => {
       outDir: buildOutputFolder,
       emptyOutDir: devBundle,
       assetsDir: 'VAADIN/build',
+      target: ["esnext", "safari15"],
       rollupOptions: {
         input: {
           indexhtml: projectIndexHtml,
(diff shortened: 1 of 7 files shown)


Artur-:
Top level await does not work properly even with this because of https://github.com/evanw/esbuild/issues/253. I wonder if it would be better to revert it so you do not accidentally use it and it seems to work even though it does not

Issue #18793, opened 2024-02-23
Browser page crash when starting hybrid app with Vaadin 24.4.0.alpha5


mshabarov:
### Description of the bug

I get my browser page crashed when [upgrade](https://github.com/vaadin/flow-hilla-hybrid-example/pull/22) the hybrid app starter, run it and open in browser.

The problematic codes seems to be [here](https://github.com/vaadin/flow/blob/206493322a3979745593acdd9605892400473f96/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx#L295), causing the following error if I wrap it as
```javascript
try { 
  // declare listener
  // @ts-ignore
  this._addEventListener(type, listener, useCapture);

  // @ts-ignore
  if (!this.eventListenerList) this.eventListenerList = {};
  // @ts-ignore
  if (!this.eventListenerList[type]) this.eventListenerList[type] = [];

  // add listener to  event tracking list
  // @ts-ignore
  this.eventListenerList[type].push({type, listener, useCapture});
} catch (e) {
  console.error(e)
}
```

```
hilla:5 TypeError: Cannot read properties of undefined (reading '_addEventListener')
    at cst.addEventListener (Flow.tsx:295:26)
    at new ConnectClient (@vaadin_hilla-frontend.js?v=c5d8b4f5:3236:5)
    at connect-client.default.ts:2:18
```

### Expected behavior

No errors

### Minimal reproducible example

See the PR mentioned above

### Versions

- Vaadin / Flow version: 24.4.0.alpha5

PR #18795, opened 2024-02-23
fix: Use closures to store listeners and fallback to window

mshabarov:
Doesn't use `this` reference directly to store a list of event listeners, but uses closures.

Calls the original `addEventListener` functions via `.call(this, arguments)` to avoid explosions due to `this` being `undefined`.

Fixes https://github.com/vaadin/flow/issues/18793

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -295,6 +295,7 @@ export const serverSideRoutes = [
             }
             function removeEventListener(type: string, listener: EventListener,
                                          useCapture: boolean = false) {
+                // @ts-ignore
                 if (!eventListenerList[type]) {
                     return;
                 }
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -281,74 +281,74 @@ export const serverSideRoutes = [
...
+            function getEventListener(type: string) {
+                if (!eventListenerList) return [];
+                if (type == undefined) return eventListenerList;
+                // @ts-ignore
+                return eventListenerList[type];
+            }
+
+            // save the original methods before overwriting them
+            const originalAddEventListener = cst.prototype.addEventListener;
+            const originalRemoveEventListener = cst.prototype.removeEventListener;
```
caalador, line 335: We had the issue basically earlier also, but now looking at this if it is run 2 times or more we will have broken handling as we will start to handle more and more listener lists for each add and remove as the new add will use the previous add etc...  previously it would just have broken in that it looses the original addEventListener.

caalador (approved):

Issue #18436, opened 2024-01-11
Consider moving /frontend directory under /src/main


tarekoraby:
### Describe your motivation

The standard directory layout of Maven/Gradle projects involves adding source files under `src/main`. However, Vaadin (especially when used with Hilla) currently requires adding JS/TS source files under `/frontend`.

### Describe the solution you'd like

Consider moving the  `/frontend` folder under `src/main` as a default. For backward compatibility, the current location of the `/frontend` folder at the project root can remain supported.

mshabarov:
We currently have the parameter `vaadin.frontend.frontend.folder` for setting a relative custom path for frontend folder, but I'm pretty sure that it's broken as in some parts of Flow the frontend folder path is hardcoded. Needs to be checked.

PR #18803, opened 2024-02-26
feat!: Move default frontend to src/main

caalador:
Move frontend folder from root
to src/main.

Closes https://github.com/vaadin/flow/issues/18436


flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig-v23.3.4-hilla.json
```diff
@@ -34,6 +34,7 @@
     "types.d.ts"
   ],
   "exclude": [
-    "frontend/generated/jar-resources/**"
+    "frontend/generated/jar-resources/**",
+    "src/main/frontend/generated/jar-resources/**"
```
mshabarov, line 38: This probably shouldn't be here, as in 23.3 the frontend dir is in the root always.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig-v23.3.4.json
```diff
@@ -34,6 +34,7 @@
     "types.d.ts"
   ],
   "exclude": [
-    "frontend/generated/jar-resources/**"
+    "frontend/generated/jar-resources/**",
+    "src/main/frontend/generated/jar-resources/**"
```
mshabarov, line 38: Same

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
```diff
@@ -85,11 +85,11 @@ class ThemeLiveReloadPlugin {
             "Custom theme component style sheet '" +
               deletedComponentStyleFile +
               "' has been deleted.\n\n" +
-              "You should also delete './frontend/generated/theme-" +
+              "You should also delete './src/main/frontend/generated/theme-" +
```
mshabarov, line 88: This probably need to mention both locations of frontend dir, as for most of the existing Customers it would be always old location ./frontend/...

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
```diff
@@ -85,11 +85,11 @@ class ThemeLiveReloadPlugin {
...
               themeName +
               ".generated.js' (simultaneously) with the component stylesheet'.\n" +
               "Otherwise it will cause a webpack compilation error 'no such file or directory', as component style sheets are referenced from " +
-              "'./frontend/generated/theme-" +
+              "'./src/main/frontend/generated/theme-" +
```
mshabarov, line 92: Same

mshabarov (changes requested):
Some minot comments.

Plus the following leftovers for "frontend" are still there:
- In `ExternalDependencyWatcher` class the folder is hardcoded
- `DeploymentConfigurationFactoryTest` the same, "frontend" hardcoded

Also, I tried this change with the hybrid and Flow apps, and it doesn't break them. Nothing is changed and they work as previously.
So in what cases this is a breaking change?
Can we merge this immediately without smart detection?

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -112,7 +112,7 @@ public class FrontendUtils {
      * By default it is <code>/frontend</code> in the project folder.
```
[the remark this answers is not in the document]
caalador: added

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig-v23.3.4-hilla.json
caalador: removed

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
caalador: added other dir

flow-server/src/main/resources/plugins/theme-live-reload-plugin/theme-live-reload-plugin.js
caalador: changed to reference file not full directory

caalador:
ExternalDependencyWatcher mentions  frontend for metaInfFolder e.g META-INF and not the project frontend folder.
For DeploymentConfigurationFactoryTest the tests mentioning frontend are for the token file contents so the actual folder doesn't matter.

For the unified execution check that the plugin is using the correct code for the frontend folders. it should not work with having frontend in ./frontend

mshabarov (approved):
LGTM, but I propose to rebase the second PR onto this, and merge the second PR.

Let me test it locally.

PR #18804, opened 2024-02-26
chore: upgrade spring framework to 5.3.32 (23.4)

ZheSun88:
default spring-framework version coming with sb 2.7 is 5.3.31.  currently, we cannot find the date when next sb 2.7 will come. so let us use it this to resolve the cve
https://www.cve.org/CVERecord?id=CVE-2024-22243

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,6 +84,9 @@
 
         <!-- Dependencies -->
         <spring.boot.version>2.7.18</spring.boot.version>
+        <!-- spring-framework 5.3.32 has security fixes -->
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -84,6 +84,9 @@
 
         <!-- Dependencies -->
         <spring.boot.version>2.7.18</spring.boot.version>
```
knoobie, line 86: > we cannot find the date when next sb 2.7 will come

FYI: There will be no new boot version, because it's Open Source EoL: https://spring.io/projects/spring-boot#support

mcollovati (approved):

PR #18805, opened 2024-02-26
fix: Binder should not set non-error fields invalid

tepi:
Set field as invalid only if `ValidationResult.isError()` returns true. Add tests to check validation with each ErrorLevel.

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/DefaultBinderValidationErrorHandler.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/DefaultBinderValidationErrorHandler.java
@@ -48,7 +48,8 @@ public class DefaultBinderValidationErrorHandler
 
     /**
      * Handles a validation error emitted when trying to write the value of the
-     * given field.
+     * given field. The field will be set to invalid state only if
+     * {@code result.isError()} returns {@literal true}.
      *
      * @param field
      *            the field with the invalid value
@@ -59,7 +60,7 @@ public class DefaultBinderValidationErrorHandler
     public void handleError(HasValue<?, ?> field, ValidationResult result) {
         if (field instanceof HasValidation) {
             HasValidation fieldWithValidation = (HasValidation) field;
-            fieldWithValidation.setInvalid(true);
+            fieldWithValidation.setInvalid(result.isError());
             fieldWithValidation.setErrorMessage(result.getErrorMessage());
         }
         setErrorTheme(field, result);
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/binder/DefaultBinderValidationErrorHandler.java
```diff
@@ -59,7 +60,7 @@ public class DefaultBinderValidationErrorHandler
     public void handleError(HasValue<?, ?> field, ValidationResult result) {
         if (field instanceof HasValidation) {
             HasValidation fieldWithValidation = (HasValidation) field;
-            fieldWithValidation.setInvalid(true);
+            fieldWithValidation.setInvalid(result.isError());
             fieldWithValidation.setErrorMessage(result.getErrorMessage());
```
knoobie, line 64: Just to make sure: Does this interfere in any kind with how the web components handle the error? e.g. they don't show the error message because invalid=false? Otherwise those https://github.com/vaadin/flow/pull/11384#issuecomment-879309037 would fail and we heavily rely on custom error level to show e.g. "Information" for the user, which are NOT allowed to block the submit; but should be visible for the user once they input "trash"

tepi: Good point, this probably needs more thinking. We might need another flag(s) besides invalid to fix this properly (hasInfo, hasWarning?). With current implementation you can have a Binder which is valid, with all of its fields being invalid, just because isError is checked in some places but not in others.

You mentioned you're showing information to the user, but are you showing warnings also? Seems to me with the current single invalid flag that is impossible?

knoobie: > With current implementation you can have a Binder which is valid, with all of its fields being invalid, just because isError is checked in some places but not in others.

I totally understand your concern here. This needs definitly a solution where multiple level of errors can be shown; some making the bean & field invalid; some done. 

We are currently doing it like so:
 - field blue / "INFO: Text" -> Field invalid + Bean valid  --> Send Form allowed
 - field red / "Error Text" -> Field invalid + Bean invalid --> Send Form forbidden
 
grafik

(Sorry for the bad example texts.. had to remove sensitive informations)

tepi: Yes, the use case is definitely valid. And also just noticed that we are sort of using the error theme name to 'communicate' the error level to the components. Not sure if components are actually using that info for anything though.

knoobie: We also have this in a multi-tiered Combobox where we say something like this: "Hey, you have selected B.. B isn't a good value. When you send the form, we are going to replace B with A." 

There is also a "recent" discussion regarding this, were the flow-component broke in v24 some of our implemtations: https://github.com/vaadin/flow-components/issues/4804

knoobie: See me edit; we use exactly that name to apply different styling on the client :)

PR #18806, opened 2024-02-26
test: set minimum supported Gradle version to 7.6.4

tltv:


flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinPlugin.kt
```diff
@@ -29,7 +29,7 @@ import org.gradle.util.GradleVersion
  */
 public class VaadinPlugin : Plugin<Project> {
     public companion object {
-        public const val GRADLE_MINIMUM_SUPPORTED_VERSION: String = "7.6"
+        public const val GRADLE_MINIMUM_SUPPORTED_VERSION: String = "7.6.1"
```
tltv, line 32:
```suggestion
        public const val GRADLE_MINIMUM_SUPPORTED_VERSION: String = "7.6.4"
```

mshabarov (approved):

PR #18807, opened 2024-02-27
fix: Read Hilla version from correct file

Artur-:
Fixes https://github.com/vaadin/platform/issues/5014

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/Platform.java
@@ -29,7 +29,7 @@ import org.slf4j.LoggerFactory;
  * @since 23.0
  */
 public class Platform implements Serializable {
-    static final String HILLA_POM_PROPERTIES = "META-INF/maven/com.vaadin.hilla/hilla/pom.properties";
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/Platform.java
```diff
@@ -87,15 +88,15 @@ public static Optional<String> getHillaVersion() {
         // immutable and thread-safe.
         if (hillaVersion == null) {
             try (final InputStream hillaPomProperties = Thread.currentThread()
-                    .getContextClassLoader().getResourceAsStream(
-                            "META-INF/maven/com.vaadin.hilla/hilla/pom.properties")) {
+                    .getContextClassLoader()
+                    .getResourceAsStream(HILLA_POM_PROPERTIES)) {
```
mshabarov, line 92: I don't get the idea: properties are red from the same place `"META-INF/maven/com.vaadin.hilla/hilla/pom.properties"` , what is the difference ?

Artur-: Uh, what happened... the correct path is `com.vaadin` without `.hilla`. Fixed

mshabarov (approved):
Tested with the starter project and couldn't see the error record in the logs anymore.

PR #18815, opened 2024-02-28
chore: Use SerializableConsumer in public methods of react adapter

mshabarov:


mcollovati (approved):

mcollovati (changes requested):
Function should become SerializableFunction, since it is captured in a lambda (for  `PropertyChangeListener`)

mshabarov:
Excellent catch! Change to serializable function.

mcollovati (approved):

Issue #14086, opened 2022-06-28
Consider adding interfaces to listen to session and UI initializations, and to session destructions


tarekoraby:
### Describe your motivation

Many use cases require applying some logic to all sessions or to all UIs. For example, to implement a [custom error handler](https://vaadin.com/docs/latest/advanced/custom-error-handler), one needs to listen to session initialization events and add the custom handler to each session. Similarly, certain use cases require listening to all UI initializations (for example, to check user's credentials) and all session destructions (for example, to free up some resources). 

Currently, one can listen to those three events via `VaadinService`'s `addUIInitListener`, `addSessionInitListener`, and `addSessionDestroyListener` methods. However, it might not be obvious to developers that `VaadinService` is the way to access these three events. Moreover, listening to these three events currently requires some boilerplate to [listen to init events of the `VaadinService` itself](https://vaadin.com/docs/latest/advanced/session-and-ui-init-listener). 

### Describe the solution you'd like

Consider providing convenient interfaces to enable listening to UI initializations and session initializations and destructions. Such interfaes would enable one to do something like the following:

```
public class MySessionListener implements VaadinSessionInitListener {
    
    @Override
    public void sessionInit(SessionInitEvent event) {
        event.getSession().setErrorHandler(new MyErrorHandler());
    }
}
```

Legioth:
I guess you would suggest that the design would be in sync with `VaadinServiceInitListener` with regards to how you register the listener, i.e. marking as `@Component` with Spring or adding a file to `META-INF/services` when using vanilla servlets.

tarekoraby:
Yes, that's what I had in mind.

knoobie:
> Wouldn't be simpler if Flow just scans for those listener during startup?

Flow can already use listener registered as Component or via META-INF - auto-registering would be something new (so not simpler). 

Additionally this would be a regression, because as a developer, I can decide with spring annotations like `@Profile` that the given listener should only be used with the given profile. Auto-registering would break this. I don't think the flow team would want to deal with that :)

tarekoraby:
> auto-registering would be something new (so not simpler).

A new approach can be simpler than the old way of doing things. Doing something automatically is, to my mind, simpler than manually doing so.  

> Additionally this would be a regression ....

I would call this a breaking change, not a regression. 

> I can decide with spring annotations like @Profile that the given listener should only be used with the given profile.

[...]

PR #18819, opened 2024-02-28
Publish ServiceInitEvent as Spring event

mstahv:
Partial "regression safe" solution to #14086. Only publishes the ServiceInitEvent at this point. Shouldn't expose any new ways how people can use even in wrong way and shouldn not pose any performance issues (event is essentially run once during application startup.

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
```diff
@@ -67,9 +68,12 @@ public SpringInstantiator(VaadinService service,
 
     @Override
     public Stream<VaadinServiceInitListener> getServiceInitListeners() {
```
tltv, line 70: Needs some Javadoc with the template from the `Instantiator` interface plus mention EventListener support.

tltv (changes requested):
Change looks good to me. Missing JavaDoc and one integration test for this in `flow-tests/vaadin-spring-tests/test-spring-common` and it's good to go as a first step for the main ticket.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
mstahv: This method? Why would it now start to need it? In general the whole class is an implementation detail 🤔

I think the only sane place to document that ServiceInitListener is now published as a spring event is the reference docs. IIRC the related, existing, CDI feature is documented there as well.

mstahv:
Added an IT blindly to flow-tests/vaadin-spring-tests/test-spring-common, I have no idea how that project is supposed to be used. mvn verify didn't seem to do anything.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
mstahv: I don't understand why 

* the current behaviour don't need JavaDoc 
* why would somebody find this classess JavaDoc

mstahv: Without understanding those, I really don't have anything reasonable to write to the javadoc

mstahv:
Some unit test was failing after I pulled in changes from master but apparently re-initiating them fixed it 🤷‍♂️

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
tltv: SpringInstantiator is public class and it's constructed at least in SpringVaadinServletService. But could be used somewhere else too. This is one reason to have proper JavaDoc.

For example something like this would be enough:

```
    /**
     * Gets all service init listeners to use. Registers by default an init
     * listener that publishes ServiceInitEvent to application context.
     *
     * @return stream of service init listeners, not <code>null</code>
     */
```

mstahv: I still don't get who will ever be reading this or why the original behaviour don't need to be mentioned in the doc, but I guess the cheapest way to get forward to do something that makes a difference is to copy-paste this into the class 🤷‍♂️

tltv: Repeating whole original just didn't seem necessary, so trimmed it a bit to the point. Who will ever read this? I guess there's possibility that some projects customizes this even outside of the Flow maintenance team.

tltv (approved):
Looks good.

Tip: To run IT locally:

```
mvn verify -pl :vaadin-test-spring-helpers,:vaadin-test-spring-common,:vaadin-test-spring-boot -Dit.test=ComponentAddedVia*
```

tltv:
I can add new article for this feature (probably in spring/events page) in docs unless @mstahv you are already on it?

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
mstahv: Yeah, and now somebody could claim the interface doc was more informative and now hidden. And it still don't mention the the previous (auto registration of ServiceEventListener beans) feature. But as said, I'm fine with or without JavaDocs, there are dozends of more important places to document, I really hope nobody needs to touch this implementation detail of our Spring  integration...

PR #18822, opened 2024-02-28
feat: Add interface for Hilla routes to add public matchers

### LINKED ISSUE #2051: [FS router] add options to enable Spring Security
Todo:
- Add roles allowed to client and server-side
- Add access checks for hilla views by view metadata (roles allowed) 
- Add Hilla public routes as anonymous allowed to VaadinWebSecurity - 
- Filter routes by access and return only allowed routes to MainLayout navigation. 
- Apply access checks as soon as routes are available on server side
- Add tests and documentation/examples.

For client-side access control - existing implementation should still work (as both are needed for complete access control implementation in Hilla)

krissvaa:
Adds support for Hilla to use VaadinWebSecurity with file-system routes and to secure them.

Fixes https://github.com/vaadin/hilla/issues/2051


mcollovati (changes requested):
Some minor comments on Javadoc and a proposal to simply the interface.

In addition, we need some tests (could maybe be added to `RequestUtilTest`, or in a new test class if it will simplify maintenance).

I also realized that we are missing tests to verify the usage of `RequestUtil.isAnonymousEndpoint`. It would be good if we could also add these.

mcollovati:
Additional note: we should also check that the new request matchers work fine with `SpringAccessPathChecker`.
Basically, enabling route path access checker, Hilla view access should be automatically checked as well for Flow navigation (e.g. `UI.navigate("hilla-view")`).

In particular, we need to verify if `RequestUtil.isAnonymousHillaRoute()` needs a check similar to `RequestUtil.isAnonymousRoute()` to prevent endless recursion.

cromoteca:
Closing for now as we don't need this unless we want to enable the feature by default one day.

PR #18823, opened 2024-02-28
fix: also scan changed files for routes

### LINKED ISSUE #18794: Route changes are not consistently detected when redeploying through Spring Boot Devtools
### Description of the bug

There's something weird going on with updating the route registry when redeploying through Spring Boot Devtools. For some classes, the route registry is updated only after a full server restart whereas a Devtools redeploy is enough for other classes.

### Expected behavior

All view changes should be detected with a regular redeploy.

1. Download skeleton-starter-flow-spring-24 from the quick download button on start.vaadin.com
2. Launch using `mvn` or by running the `main` method
3. Add `Test.java` with the following content
```
@Route
public class Test extends VerticalLayout {
    public Test() {
        add(new Text("Test"));
    }
}
```
4. Notice that the server is redeployed but that navigating to `/test` gives an error page
5. Restart the server. Notice that you can now navigate to `/test`.
6. Comment out `@Route` in `Test.java`. Wait for the redeploy and notice that `/test` still renders the view.
7. Restore the `@Route` annotation in `Test.java` and comment out `@Route` in `MainView.java`. Wait for the redeploy and notice that you cannot navigate to `/`.
8. Restore the annotation in `MainView.java`, wait for the redeploy and notice that you can again navigate to `/`.

### Versions

- Vaadin / Flow version: Tested with 24.3.5 and 24.4.0.alpha4
- Java version: 21
- IDE (if applicable): Tested with Eclipse and VS Code

tepi:
Do Route scanning on packages of changed files also. Depending on the package/class structure, it some cases such change was not correctly scanned earlier.

Fixes #18794

### DIFF
--- a/flow-tests/test-redeployment/src/test/java/com/vaadin/flow/uitest/ui/DevModeClassCacheIT.java
+++ b/flow-tests/test-redeployment/src/test/java/com/vaadin/flow/uitest/ui/DevModeClassCacheIT.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import java.util.List;
 
 import net.jcip.annotations.NotThreadSafe;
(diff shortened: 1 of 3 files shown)


flow-tests/test-redeployment/src/test/java/com/vaadin/flow/uitest/ui/DevModeClassCacheIT.java
```diff
@@ -96,10 +112,28 @@ public void testDevModeClassCachePopulated() {
...
+        waitForElementPresent(By.id("last-span"));
+        allSpans = $("span").all();
+        Assert.assertEquals("Unexpected cached route packages.",
+                "com.vaadin.flow.uitest.ui,com.vaadin.flow.uitest.ui.reloadaddedviews",
+                allSpans.get(5).getText().split(":")[1]);
     }
 
     // create class on the fly
-    protected void createNewViewReloadAndWait() {
+    protected void createNewViewReloadAndWait(boolean withRoute) {
```
mcollovati, line 136: nit: we can probably change the method name to make test more clear, since now it is used both to create and update the view. What about `createOrUpdateViewReloadAndWait()`?

mcollovati (changes requested):
Just a nit on the test, other than that looks good to me.
When adding, modifying or removing a route, the changes are effective after the automatic redeploy.

mcollovati (approved):

PR #18827, opened 2024-02-28
feat: check custom header for dev tools connection verification

mcollovati:
Allows to specify a custom header name to get the client address that is verified to allow access to dev tools.
In addition, validates all hops in the X-Forwaded-For chain.


flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -420,15 +433,17 @@ static boolean isAllowedDevToolsHost(AbstractConfiguration configuration,
...
+            // or if local addresses must be rejected and there are no host
+            // rules defined
             return false;
         }
         // Always allow localhost
         try {
             InetAddress inetAddress = InetAddress.getByName(remoteAddress);
-            if (inetAddress.isLoopbackAddress()) {
+            if (inetAddress.isLoopbackAddress() && allowLocal) {
                 return true;
```
Artur-, line 447: If it IS a local address and local is NOT allowed, should this not immediately return false, i.e. if it is a loopback address, always return `allowLocal`

mcollovati: Good point

mcollovati: done

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinConfigurationProperties.java
```diff
@@ -185,6 +185,13 @@ public static class Devmode {
          */
         private String hostsAllowed = "";
 
+        /**
+         * The name of the custom HTTP header that contains the client IP
```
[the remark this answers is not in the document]
mcollovati: done

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -420,16 +447,18 @@ static boolean isAllowedDevToolsHost(AbstractConfiguration configuration,
     }
 
     private static boolean isAllowedDevToolsHost(String remoteAddress,
-            String hostsAllowed) {
-        if (remoteAddress == null) {
-            // No remote address available so we cannot check...
+            String hostsAllowed, boolean allowLocal) {
+        if (remoteAddress == null || remoteAddress.isBlank()
+                || (hostsAllowed == null && !allowLocal)) {
```
Artur-, line 452: this interprets hostsAllowed == "" different from null while the code below interprets them as the same

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
```diff
@@ -397,21 +401,44 @@ static boolean isAllowedDevToolsHost(AbstractConfiguration configuration,
...
+                SERVLET_PARAMETER_DEVMODE_HOSTS_ALLOWED, null);
 
-        if (!isAllowedDevToolsHost(remoteAddress, hostsAllowed)) {
+        if (!isAllowedDevToolsHost(remoteAddress, hostsAllowed, true)) {
             return false;
         }
-        String forwardedFor = request.getHeader("X-Forwarded-For");
-        if (forwardedFor != null) {
+        String remoteHeaderIp = configuration.getStringProperty(
+                SERVLET_PARAMETER_DEVMODE_REMOTE_ADDRESS_HEADER, null);
```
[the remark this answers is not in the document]
mcollovati: Yes, the assumption is a single header with a single address, as this is supposed to be safely handled by the proxy server.
I'll update the docs

flow-server/src/main/java/com/vaadin/flow/server/communication/IndexHtmlRequestHandler.java
mcollovati: Is it ok to have the same behavior for null, empty and blank?
If so, I would directly trim to null the value read from the configuration

Artur-: I would say it is better to have the same behavior and fewer cases

mcollovati: done

Artur- (approved):

PR #18828, opened 2024-02-28
feat: enable frontend fallback

caalador:
Enable falling back to the
original frontend folder if
the new one doesn't exist.


mshabarov:
Something is missing, as now if I move the frontend folder into src/main by hand, and start the app:
- when running with IDE's play button, the frontend location resolved well and I get app started and running.
- when I run with mvn, Flow looks into ./frontend still and explodes because theme is missing in ./frontend

mshabarov:
It was my mistake, I haven't changed the plugin from `vaadin-maven-plugin` to `flow-maven-plugin` and only built the Flow.
So after corrections, I got the following configurations working well with starter project:
1. Legacy folder in root
2. Folder in src/main
3. Custom location "./my-frontend" and parameter `<frontendDirectory>./my-frontend</frontendDirectory>`.

I think we can merge it.

mshabarov (approved):

Issue #18813, opened 2024-02-28
Change the default content of routes.tsx to respect FS router


mshabarov:
### Describe your motivation

Vaadin Flow generates a `routes.tsx` file in `frontend` with a default codes containing only server-side routes.
This default codes should take into account the File System-based routing feature and, in the same time, it should allow easily configure the explicit routing configurataion.

### Describe the solution you'd like

Here is the codes that takes into account FS router (should be double checked with the latest Hilla versions):
```javascript
import views from 'Frontend/generated/views.js';
import { serverSideRoutes } from "Frontend/generated/flow/Flow";
import {createBrowserRouter} from "react-router-dom";

views.children?.push(...serverSideRoutes);

// Or use explicit routes configuration
// export const routes = [
//     { path: 'react-components',  element: <ReactComponents/> },
//     ...serverSideRoutes
// ] as RouteObject[];

export const routes = views;

export default createBrowserRouter(views, {basename: new URL(document.baseURI).pathname });
```

This file should be generated on adding Hilla views as previously but it should contain a template for explicit routes config if developers want to use this approach, but in the same time it should allow to use FS router without any manual changes.

### Additional context

Vaadin 24.4.

mshabarov:
`views.js` is generated by `@vaadin/hilla-file-router/vite-plugin.js` . The plugin should be added manually to the Vite config (for now).
Thus, even if the routes.tsx contains `views.js`, this would not compile unless you add a plugin.

mshabarov:
We need to add the plugin by default and move serverSideRoutes into views.js, see https://github.com/vaadin/flow/pull/18830#discussion_r1507090757.

Then probably we can figure out a unified content of this file that allows easily add a path and work with FS router out-of-the-box.

mshabarov:
This should be done first https://github.com/vaadin/flow/issues/18832.

mshabarov:
Here what we discussed for the next steps https://github.com/vaadin/hilla/issues/2146#issuecomment-1982919750.

PR #18830, opened 2024-02-28
chore: Update the routes.tsx template to give hint about FS router

mshabarov:
Related-to https://github.com/vaadin/flow/issues/18813


flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -6,3 +19,9 @@ export const routes = [
 ] as RouteObject[];
 
 export default createBrowserRouter(routes);
+
+// An example of File System-based routing auto-configuration
+// import views from 'Frontend/generated/views.js';
+// views.children?.push(...serverSideRoutes);
```
caalador, line 25: This is not specifically for the change, but what if for some reason views doesn't contain children so that we end up without any serverSideRoutes and the app doesn't navigate.... should the generated views.js routes have a helper method that should be called instead that would correctly setup the routes array so that the given route (serverSideRoutes for our case) would always 100% be added.

mshabarov: Yes, that would be more convenient of course.
Ideally, as a developer I'd prefer as less as possible codes in the default file, but in the same time an easy way to:
- swap FS routing and explicit routing
- add a new route

mshabarov: My gut feeling is that it makes sense to complete the following pre-requisites before we change the content to something useful for users:
- add Vite plugin to default Vite config, so it will be added each time when frontend build is needed.
- add this plugin to package.json default deps, so it gets installed automatically.
- include serverSideRoutes into generated `views.js` and probably other boilerplate codes.
Then I think we can update this template file with a more or less final codes.

PR #18834, opened 2024-02-29
chore: Upgrade react deps to latest

mshabarov:
Related-to https://github.com/vaadin/hilla/issues/2147


caalador (approved):
Else this looks good, but hilla uses ^ and Flow wants to specify the actual version so whenever there is a update this will be out of sync again.
Should hilla use the defined version or would writing the overrides file with the specific flow used version work?

mshabarov:
This is a general question regarding npm versions handling in general.
I'd pin the defined version everywhere to get rid of sudden npm errors.
But asking for @platosha opinion, which is better for Hilla ?

PR #18874, opened 2024-03-04
fix: propagate error message in default access denied handler

### LINKED ISSUE #18870: Flow should indicate that access was denied on the "not found" view when in development mode
### Description of the bug

With access control enabled, when you add a view and you cannot navigate to it, you have no idea what is wrong.
The error page only states
```
Could not navigate to 'myview'
```

### Expected behavior

It should state the reason that you cannot navigate to `myview`, e.g. "Access is denied"

npx @vaadin/cli init authtest --auth
 
+ remove `@PermitAll` from Hello world

### Versions

- Vaadin / Flow version: 24.3.
- Java version:
- OS version:
- IDE (if applicable):

mcollovati:
Fixes #18870

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/RouteAccessDeniedError.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/RouteAccessDeniedError.java
@@ -36,7 +36,8 @@ public class RouteAccessDeniedError extends Component
     @Override
     public int setErrorParameter(BeforeEnterEvent event,
             ErrorParameter<AccessDeniedException> parameter) {
-        event.rerouteToError(NotFoundException.class);
+        event.rerouteToError(NotFoundException.class,
+                parameter.getCustomMessage());
         return HttpStatusCode.NOT_FOUND.getCode();
     }
 }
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/RouteAccessDeniedError.java
```diff
@@ -36,7 +36,8 @@ public class RouteAccessDeniedError extends Component
     @Override
     public int setErrorParameter(BeforeEnterEvent event,
             ErrorParameter<AccessDeniedException> parameter) {
-        event.rerouteToError(NotFoundException.class);
+        event.rerouteToError(NotFoundException.class,
```
knoobie, line 39: Is this message always printed? in Production and Development Mode? (Just in case)

mcollovati: It depends on the logic that does the re-route.
`AccessDeniedException` by default has always null message.
If the re-route is triggered by `NavigationAccessControl` the parameter message is present only in development mode.
If it is the application code that re-routes, it is up to the developer to provide a message or not.

tepi (approved):

Issue #18808, opened 2024-02-27
ReactAdapterElement auto completion in VS Code doesn't use the right types


Legioth:
### Description of the bug

VS Code doesn't use the declared types when auto completing the abstract `render` method in `ReactAdapterElement`.

The method gets completed like this:
```typescript
    protected render(hooks: Readonly<{ useState: <T>(key: string, initialValue?: T | undefined) => [value: T, setValue: Dispatch<T>]; useCustomEvent: <T = undefined>(type: string, options?: CustomEventInit<T>) => T extends undefined ? () => boolean : (value: T) => boolean; }>): ReactElement<any, string | JSXElementConstructor<any>> | null {
        throw new Error("Method not implemented.");
    }
```

### Expected behavior

Expected to get the same type signature as in the abstract method:
```typescript
    protected render(hooks: RenderHooks): ReactElement | null {
        throw new Error("Method not implemented.");
    }

```
Note that there's a difference both in the type of the `hooks` parameter and in the return type.

### Minimal reproducible example

1. Type `class MyClass extends ReactAdapterElement {}` into a .tsx file in VS Code
2. Add an import for `ReactAdapterElement`
3. Select the Quick Fix option "Implement inherited abstract class"

### Versions

- Vaadin / Flow version: 24.4.0.alpha5
- IDE (if applicable): VS Code 1.86.2

tepi:
Same result with IDEA, although it's interesting that immediately after IDEA implements the abstract method, it suggests to remove the explicit type argument for the return type:

<img width="740" alt="Screenshot 2024-03-04 at 09 45 42" src="https://github.com/vaadin/flow/assets/1527900/3a22e14b-2062-45b9-b0df-47bae36a44a9">

Also noticed that RenderHooks is wrapped with `Readonly<>`. If that is removed, the result is more or less what's expected (on IDEA):

<img width="849" alt="Screenshot 2024-03-04 at 09 46 18" src="https://github.com/vaadin/flow/assets/1527900/1b52c133-3186-40d0-96a3-d03acd0a0ae6">

Unfortunately could not find a reason or a fix for either behavior yet.

PR #18875, opened 2024-03-04
fix: move readonly keywords for better autocomplete

tepi:
Fixes https://github.com/vaadin/flow/issues/18808 (partially)


caalador (approved):

mshabarov (approved):
Fixes the issue for me.
Now VS Code gives
```
    protected render(hooks: RenderHooks): ReactElement<any, string | JSXElementConstructor<any>> | null {
        throw new Error("Method not implemented.");
    }
```
We might want to make `ReactElement` generated proper types, but this is another issue.

PR #18876, opened 2024-03-04
fix: navigation to client route

### LINKED ISSUE #18867: [24.4] Navigating from a Flow view to a React view using UI.navigate is not working
### Description of the bug

When I try to navigate from a Flow view to a route associated with a React view using `UI.getCurrent().navigate`, it is not working. The target url is correctly set to the browser, but it seems that the React router is not reacting to the changed url:

https://github.com/vaadin/flow/assets/70698630/f1a7bfa1-007f-4800-bf9a-84a4283c0b5f

### Expected behavior

To be able to navigate from Flow views to React views and vice versa.

I used [hilla-file-router-example](https://github.com/Lodin/hilla-file-router-example) as a base to add a React view with required param "id" (so it is not visible in the automatically populated menu) but it is registered with the route `/person/:id` (it can be verified by inspecting the value of `Vaadin.server.views` in the browser console). 

My playground is shared in the `taefi/navigate-from-flow-issue` branch of the same repository. 

### Versions

- Vaadin version: 24.4.0.alpha8

caalador:
Fix the client route naviagtion
by adding client event also
when not preventing navigation.

Fixes #18867

### DIFF
--- a/flow-client/src/main/frontend/Flow.ts
+++ b/flow-client/src/main/frontend/Flow.ts
@@ -50,6 +50,7 @@ export interface NavigationParameters {
 
 export interface PreventCommands {
   prevent: () => any;
+  continue?: () => any;
 }
 
 export interface PreventAndRedirectCommands extends PreventCommands {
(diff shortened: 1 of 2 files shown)


tltv (approved):
Tested with hilla-file-router-example and it works.

Issue #18832, opened 2024-02-29
Add FileSystem-based router plugin by default when frontend build is needed


mshabarov:
### Describe your motivation

Hilla's [FS-based router plugin](https://github.com/vaadin/hilla/tree/main/packages/ts/hilla-file-router) is supposed to be added by hand:
- run "npm i @vaadin/hilla-file-router"
- add the "vitePluginFileSystemRouter" Vite plugin into vite.config.ts:
```
       const customConfig: UserConfigFn = (env) => ({
          plugins: [vitePluginFileSystemRouter()]
       });
```
This should be done by default, so whenever Flow starts the frontend build (Vite dev server or new bundle build), the plugin is installed and applied.
This would also trigger a file `views.js` that is needed for `routes.tsx`.

### Describe the solution you'd like

Flow adds this plugin into generated Vite config and adds it's package to package.json.

### Additional context

Vaadin 24.4.

knoobie:
> This should be done by default, so whenever Flow starts the frontend build (Vite dev server or new bundle build), the plugin is installed and applied.

Only if Hilla is used, right? :)

mshabarov:
Well, yes. Not needed for bundle. Let me correct.

PR #18878, opened 2024-03-04
feat: add file system router plugin for vite

tltv:
Adds vitePluginFileSystemRouter to Vite plugins by default when React and Hilla is in use.

Fixes: #18832


mshabarov (approved):
Tested, Vite generated now has this plugin and `views.js` is generated.

Issue #18816, opened 2024-02-28
[Draft] WebComponentExporter improvements for React integration


mshabarov:
### Describe your motivation

The following point were found within DX testing of Flow-React integration in Vaadin 24.4:

- [ ] WebComponentExporter can automatically add a script for loading a web-component, currently one should do it manually or use `loadComponentScript` from `Flow.tsx`
- [ ] Note about must have public CTOR in WebComponentExporter
- [ ] Add additional method parameters to `createWebComponent` method, so that it can obtain the component's attributes.
- [ ] Web-component loading takes time. We have to provide a convenient API to hook into web-component loading process, e.g.:
   - [ ] add a second method parameter `loadComponentScript = (tag: String, onLoad?: () => void)` that helps to make changes in UI on load of the script
   - [ ] return `Promise` in `loadComponentScript`, so that the it would possible to wait for loading with this promise.
- [ ] Update docs with an example how to add attributes into component, e.g.
```javascript
import {loadComponentScript} from "Frontend/generated/flow/Flow";
import {VerticalLayout} from "@vaadin/react-components/VerticalLayout.js";
import React from "react";

type CustomComponentProps = {
    greeting: string
}

function CustomComponent(props: CustomComponentProps) {
    loadComponentScript("custom-component");
    return React.createElement("custom-component", {
         // greeting is a property defined on the server-side also        
         greeting: props.greeting
    });
}

export default function HillaView() {
    return (
        <>
            <VerticalLayout className={'centered-content'}>
                <h3>Hilla View</h3>
                <CustomComponent greeting={"hello from Hilla"} />
            </VerticalLayout>
        </>
    );
}
```
- [ ] Routing and Placeholder components from Flow-Hilla React starter project can be included into some helper in Flow.

### Additional context

Vaadin 24.4.0.alpha5

PR #18890, opened 2024-03-06
feat: Extends createWebComponent

caalador:
Add optional properties to
createWebComponent to be able to use
it instead of loadComponentScript
and createElement when adding properties.

Add fucntion hook to listen to onLoad for
the webcomponent script.

part of #18816


mshabarov (changes requested):
This works as expected to me and looks good.
Tested with 
```
```

caalador:
The problem is that I think it is not possible in any way to export the same method name with different parameters in js

mshabarov (approved):

Issue #18869, opened 2024-03-03
Binder setter method is not invoked until the entire form is completed


alexanoid:
### Description of the bug

With recent updates to the form validation process, the binder setter method is now not invoked until the entire form is completed.

For example, in the following code, I set the name value to DTO:

```
        bindingBuilder.bind(dto -> value, (dto, name) -> {
            name = StringUtils.isNoneBlank(name) ? name.trim() : null;
            dto.setName(name);
        });

```
And then, in another field, I check the name for some custom validation. For example:

```
binder.forField(skillComboBox).withValidator(entity -> {
                    JobableDto dto = binder.getBean();
                    if(dto.getName ... some logic
```

Now I'm unable to get the dto.getName() value until I completely fill out the entire form and press Save button with `binder.writeBean(dto)`. Previously, I was able to see these values without that.

Is there a way to achieve this with the current implementation?

This is not very convenient, as it requires the user to wait until the entire form is completed before being notified of any additional validation errors

### Expected behavior

The binder setter method should be invoked immediately after the field value is added and validated on the filed level.

### Minimal reproducible example

n/a

### Versions

- Vaadin / Flow version: 24.4.0.alpha8
- Java version: 17

Legioth:
> **Expected behavior**
> The binder setter method should be invoked immediately after the field value is added.

The intention is that values are written to the bean only when that would leave the bean in a consistent state, i.e. when all validations pass.

There are two ways around this:
* If you're concerned about recent validation changes causing more strict checking, then you could test the newly introduced `setDefaultValidatorsEnabled(false)` API either for some specific binding or for the entire binder.
* If you explicitly want to write out the values that pass validation while ignoring invalid values, then you can use the `writeBeanAsDraft` API

knoobie:
> The intention is that values are written to the bean only when that would leave the bean in a consistent state, i.e. when all validations pass. 

I haven't tested 24.4-alpha; but once beta is released, I'm going to test it with out applications with some advanced binder usages, but just looking at this I would say it would break our apps because we rely on getBean() to always returns the up-to-date instance.. why should I use setBean if the values aren't written instantly? If I want to have a consistent bean I would have used read/writeBean in the first place. 

At least in the past it was like this if setBean is used: 
[...]

PR #18891, opened 2024-03-06
fix: fix setBean single binding write and add tests

tepi:
Fixes https://github.com/vaadin/flow/issues/18869

Adds missing test to confirm that setBean works as before.

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -1916,11 +1916,15 @@ public class Binder<BEAN> implements Serializable {
     }
 
     /**
-     * Informs the Binder that a value in Binding was changed. This method will
-     * only validate the changed binding and ignore state of other bindings.
+     * Informs the Binder that a value in Binding was changed.
      *
-     * If {@link #setBean(Object)} was used, the changed value will be written
-     * to the bean immediately, assuming that Binder-level validators also pass.
+     * If {@link #readBean(Object)} was used, this method will only validate the
+     * changed binding and ignore state of other bindings.
+     *
+     * If {@link #setBean(Object)} was used, all pending changed bindings will
+     * be validated and non-changed ones will be ignored. The changed value will
+     * be written to the bean immediately, assuming that Binder-level validators
+     * also pass.
      *
      * @param binding
      *            the binding whose value has been changed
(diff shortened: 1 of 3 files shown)


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -2229,9 +2217,13 @@ public <FIELDVALUE> Binding<BEAN, FIELDVALUE> bindReadOnly(
...
-     * Any change made in one of the fields also runs validation for all the
-     * fields {@link Binding} and bean level validation for this binder (bean
-     * level validators are added using {@link Binder#withValidator(Validator)}.
+     * Note: Any change made in one of the bound fields runs validation for only
+     * the changed {@link Binding}, and additionally any bean level validation
+     * for this binder (bean level validators are added using
+     * {@link Binder#withValidator(Validator)}. As a result, the bean set via
+     * this method is not guaranteed to always be in a valid state. If bean
+     * validity is required at all times, {@link #readBean(Object)} and
+     * {@link #writeBean(Object)} should be used instead.
```
mcollovati, line 2230: Should we also add a note about `StatusChangeListener` and `BinderValidationStatusHandler` being invoked with validation successful events, even if the bean may be in an invalid stated?

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
mcollovati, line 1: Related to the comment about validation events, should we add assertions (or separated tests) to verify that valid status events are fired with `setBean` even if other fields are invalid?

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
tepi: Good point, I'll add it

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
tepi: That makes sense, I'll add those too

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
tepi: Done

flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
tepi: Done

mcollovati (approved):

PR #18892, opened 2024-03-06
feat: Pass Java class info to browser in dev mode

Artur-:


flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
```diff
@@ -222,6 +229,14 @@ private native void publishJavascriptMethods(String applicationId,
     private Node getDomElementByNodeId(int id) {
         StateNode node = registry.getStateTree().getNode(id);
         return node == null ? null : node.getDomNode();
+
+    }
+
+    private String getJavaClass(int id) {
+        StateNode node = registry.getStateTree().getNode(id);
+        return node == null ? null
+                : node.getMap(NodeFeatures.ELEMENT_DATA).getProperty("jc")
```
caalador, line 238: Just asking as I can't remember if it is a problem, but could we just use here the JAVA_CLASS constant?

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
```diff
@@ -245,6 +246,18 @@ && isAttachedToParent(component.getElement(), parent)) {
                         component.getElement().isEnabled());
             }
         }
+
+        boolean isUI = ui.isPresent() && ui.get() == component;
+        if (!isUI && !attachEvent.getUI().getSession().getConfiguration()
```
Artur-, line 251: This is probably wrong as the UI does not necessarily have a session and not having a session is OK..

Artur-:
As this info is only needed in real development mode, I changed the patch to ignore all other situations (missing session, missing configuration etc). Then it should not even require any test changes..

flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
caalador: Seems to work so it was reverted?
We should aim for not using magic strings as we create problems for ourselves with those.

caalador (commented):
Else I guess it would be fine.

flow-client/src/main/java/com/vaadin/client/ApplicationConnection.java
Artur-: Seems to have gotten lost while reverting test fixes and fixing stuff again. Now it's back again. Quick, do something before it disappears!

Issue #18894, opened 2024-03-06
Theme not found when using Flow alpha 17


Artur-:
### Description of the bug

When trying to get the theme folder using `ThemeUtils.getThemeName` then it tries to use `src/main/frontend` even when the theme is in `frontend/themes` and fails to find the theme name.

Potentially related to the fact that the code is called through the push channel(?)

```
	at com.vaadin.copilot.plugins.themeeditor.ThemeModifier.lambda$getThemeFolder$3(ThemeModifier.java:206) ~[classes/:na]
	at java.base/java.util.Optional.orElseThrow(Optional.java:403) ~[na:na]
	at com.vaadin.copilot.plugins.themeeditor.ThemeModifier.getThemeFolder(ThemeModifier.java:205) ~[classes/:na]
	at com.vaadin.copilot.plugins.themeeditor.ThemeModifier.<init>(ThemeModifier.java:58) ~[classes/:na]
	at com.vaadin.copilot.plugins.themeeditor.ThemeEditorMessageHandler.<init>(ThemeEditorMessageHandler.java:60) ~[classes/:na]
	at com.vaadin.copilot.Copilot.setupCommands(Copilot.java:75) ~[classes/:na]
	at com.vaadin.copilot.Copilot.handleConnect(Copilot.java:64) ~[classes/:na]
	at com.vaadin.base.devserver.DebugWindowConnection.handleConnect(DebugWindowConnection.java:222) ~[vaadin-dev-server-24.4.0.alpha17.jar:na]
	at com.vaadin.base.devserver.DebugWindowConnection.onConnect(DebugWindowConnection.java:202) ~[vaadin-dev-server-24.4.0.alpha17.jar:na]
	at com.vaadin.flow.server.communication.PushHandler.lambda$onConnect$3(PushHandler.java:599) ~[flow-server-24.4.0.alpha17.jar:24.4.0.alpha17]
	at java.base/java.util.Optional.ifPresent(Optional.java:178) ~[na:na]
	at com.vaadin.flow.server.communication.PushHandler.lambda$onConnect$4(PushHandler.java:598) ~[flow-server-24.4.0.alpha17.jar:24.4.0.alpha17]
	at com.vaadin.flow.server.communication.PushHandler.callWithServiceAndSession(PushHandler.java:227) ~[flow-server-24.4.0.alpha17.jar:24.4.0.alpha17]
[...]

PR #18902, opened 2024-03-06
fix: Frontend folder is always correctly checked

caalador:
Fixes frontend folder check
so that it is correctly gotten
from configuration even when no
buildInfo is available.

Fixs #18894


mshabarov (approved):
Changes look good to me. Have you tested with the Copilot Flow project ?

Artur-:
Seems to fix the issue in copilot

PR #18904, opened 2024-03-06
feat: default routes.tsx for fs router

tltv:
Generate default routes.tsx with a support for File System router. FS router plugin is always installed by default when react is enabled.

Fixes: #18813


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -111,13 +111,20 @@ private void doExecute() throws ExecutionFailedException {
...
         try {
-            writeFile(flowTsx, getFileContent(FLOW_TSX));
+            writeFile(flowTsx, getFlowTsxContent());
             if (fileAvailable(REACT_ADAPTER_TSX)) {
                 writeFile(reactAdapterTsx, getFileContent(REACT_ADAPTER_TSX));
             }
 
             if (!routesTsx.exists()) {
-                writeFile(routesTsx, getFileContent(FrontendUtils.ROUTES_TSX));
+                if (FrontendUtils.isHillaUsed(options.getFrontendDirectory(),
```
mshabarov, line 120: The content should be the same for all cases.
As we move it to `frontend/generated`, it can be generated even for plain Flow app.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -175,6 +182,31 @@ private void cleanup() throws ExecutionFailedException {
...
+                            """
+                                    import { toReactRouter } from '@vaadin/hilla-file-router/runtime.js';
+                                    import views from 'Frontend/generated/views.js';
+                                    """)
+                    .replace("//#getDefaultReactRoutesFunction#", """
+                            export const getDefaultReactRoutes = () => {
+                                  // @ts-ignore
+                                  const route = toReactRouter(views);
+                                  if(route.children) {
+                                      route.children?.push(...serverSideRoutes);
```
mshabarov, line 199: Only `route.children?.push(...serverSideRoutes);` can be used when we add `children` property to route object in Hilla.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes-fs.tsx
```diff
@@ -0,0 +1,7 @@
+import { createBrowserRouter, RouteObject } from 'react-router-dom';
+import { serverSideRoutes } from "Frontend/generated/flow/Flow";
+import { getDefaultReactRoutes } from "Frontend/generated/flow/Flow";
+
+
+export const routes = getDefaultReactRoutes();
```
mshabarov, line 6: This would be better to explicitly configure in this file, because:
1. Import routes from Flow looks semantically incorrect
2. For non standard cases developer would need to remove this method and rewrite anyway.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -83,6 +83,7 @@ public class TaskGenerateReactFiles implements FallibleCommand {
     private static final String REACT_ADAPTER_TSX = "ReactAdapter.tsx";
     static final String FLOW_FLOW_TSX = "flow/" + FLOW_TSX;
     static final String FLOW_REACT_ADAPTER_TSX = "flow/" + REACT_ADAPTER_TSX;
+    static final String FRONTEND_VIEWS_FOLDER = "views";
```
mshabarov, line 86: This already exists in `FrontendUtils::HILLA_VIEWS_PATH`

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
```diff
@@ -136,8 +138,8 @@ private void createGeneratedConfig() throws IOException {
     }
 
     private String updateFileSystemRouterVitePlugin(String template) {
-        if (options.isReactEnabled() && FrontendUtils.isHillaUsed(
-                options.getFrontendDirectory(), options.getClassFinder())) {
+        if (options.isReactEnabled() && EndpointRequestUtil
```
mshabarov, line 141: Why we need to check for hilla class?

tltv: We can't actually. Tests are failing probably due to this. Need to remove it.

tltv: Fixed

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
tltv: Done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
tltv: Rewritten for new spec.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes-fs.tsx
tltv: Rewritten for new spec.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
tltv: Rewritten for new spec.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/react-router/package.json
```diff
@@ -15,6 +15,7 @@
   },
   "devDependencies": {
     "@types/react": "18.2.60",
-    "@types/react-dom": "18.2.19"
+    "@types/react-dom": "18.2.19",
+    "@vaadin/hilla-file-router": "24.4.0-alpha10"
```
caalador, line 19: This looks like it will create issues with updating the version. Could we instead have the default views.ts not use the hilla-file-router at all and return a empty object 
 `{ path: "", module: undefined, children: [] }` 
 
 And move the `toReactRouter` from routes.tsx to views.ts that we would always extend with a method?
 e.g. we would in routes.tsx only have `import buildRoutes from 'Frontend/generated/views.js';` with the call `export const routes = buildRoutes(...serverSideRoutes)` which would do all the heavy lifting with imports and return routes.

tltv: Same version update maintenance issue as with already existing dependencies. We need this for the Vite plugin installation. So adding it in case when hilla is not used otherwise, but still using react. To get the views.ts generated.

views.ts exports routes which is `AgnosticRoute` type, so not sure if it's good idea to hide this type completely. Also, hilla generates this same content when views folder exists and doesn't have any views.

Auto-detection of hilla views based on the `const routes = [...serverSideRoutes]` is still in use. If we change this format in the routes.tsx, then auto-detection should also understand the new format. Or as discussed before, we don't necessarily need to check routes.tsx content at all anymore, but somehow we need to know when there are only server side routes when using routes.tsx. So the content based check is still used because it works right now. If we do this suggested change, how do we check this instead?

caalador: It's not quite the same as it should be up to date with the vaadin release version. (this should be fine when having the platform versions.json I think)
Also now we have flow dependency towards hilla functionality that is always there even when hilla is explicitly excluded.

tltv: Actually, after adding this fallback routes.ts in Flow, we can probably revert a bit back to only install plugin when hilla is used. Because why would you now need the plugin to generate that "blank" views.ts file, if we do it anyway in Flow. Original idea was to let Hilla plugin generate views.ts always, but we can generate "blank" just as well in Flow side, imho. And in practice we have to do that as a fallback method, it seems. If views folder doesn't exist at all, hilla plugin doesn't generate anything.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -1,8 +1,27 @@
...
-export const routes = [
-  ...serverSideRoutes
-] as RouteObject[];
+// @ts-ignore
+const route = toReactRouter(views);
+export const routes = [...serverSideRoutes] as RouteObject[]
+if(route.children && route.children.length > 0) {
+  route.children.push(...serverSideRoutes);
+  routes.pop();
+  routes.push(route);
```
mshabarov, line 12: This doesn't work if I simply have `frontend/views/my-view.tsx` without a main layout.
It's because if you have undefined "element", it makes troubles for react router and it cannot navigate to Flow view.
This works for me:
```
if(route.children && route.children.length > 0) {
  route.children.push(...serverSideRoutes);
  routes.pop();
  if (route.element) {
    routes.push(route);
  } else {
    routes.push(...route.children);
  }
}
```
So we have to check if the root element is actually defined.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
```diff
@@ -136,8 +136,7 @@ private void createGeneratedConfig() throws IOException {
     }
 
     private String updateFileSystemRouterVitePlugin(String template) {
-        if (options.isReactEnabled() && FrontendUtils.isHillaUsed(
-                options.getFrontendDirectory(), options.getClassFinder())) {
+        if (options.isReactEnabled()) {
```
mshabarov, line 139: Vite generated file is now generated even for Flow app with default bundle.
Shouldn't be there.

mshabarov: Probably we miss an additional "is bundle" check.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
tltv: If we revert this back to the original, then we should be good with bundle too. Check out this thread https://github.com/vaadin/flow/pull/18904#discussion_r1517605620

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -1,8 +1,27 @@
 import { createBrowserRouter, RouteObject } from 'react-router-dom';
 import { serverSideRoutes } from "Frontend/generated/flow/Flow";
+import { toReactRouter } from '@vaadin/hilla-file-router/runtime.js';
+import views from 'Frontend/generated/views.js';
 
-export const routes = [
-  ...serverSideRoutes
-] as RouteObject[];
+// @ts-ignore
+const route = toReactRouter(views);
```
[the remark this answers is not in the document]
tltv: If block uses `route`: `if(route.children && route.children.length > 0) {`.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
mshabarov: I propose to revert, as now I get confused. With this change:
1. Flow + default bundle - Vite config and empty views are created. Routes.tsx and Flow.tsx aren't created. I believe we shouldn't generate Vite config and views.ts at all in this mode.
2. Flow + custom bundle - Vite config, empty views.ts, routes.tsx and Flow.tsx are generated. I don't know why we need the latter 3 files. I propose to not generate them at all and not add the Vite plugin. Also, application doesn't work with this new change - UI isn't rendered at all.
3. Flow + Hilla views in views folder - Vite config, views.ts, routes.tsx and Flow.tsx have to be generated. But in this case views.ts is generated by Hilla plugin.

All of these give me a clue that Flow shouldn't generate `views.ts` at all. It shouldn't generate routes.tsx and Flow.tsx either for bundles modes as well.

Instead, only generate all of these files when Hilla is used.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/react-router/package.json
tltv: Tests fails now to `import { toReactRouter } from '@vaadin/hilla-file-router/runtime.js';` in routes.tsx after reverting. So that import has to be dynamic, if we don't require the plugin.

mshabarov: Seems like `views.tsx`, as well as `Flow.tsx` and `routes.tsx` are NOT needed for bundle modes.
See my comment below.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
mshabarov: Well, this isn't that easy as I described. I missed the fact that `index.tsx` imports routes and default bundle also have index and routes already generated.

mshabarov:
We don't need to check `serverSideRoutes` for `isHillaUsed`:
```
        var files = List.of(FrontendUtils.INDEX_TS, FrontendUtils.INDEX_TSX,
                FrontendUtils.ROUTES_TS, FrontendUtils.ROUTES_TSX);
        for (String fileName : files) {
          ....
        }
```
Can be removed, as it switched to Vite for custom bundle.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/react-router/package.json
tltv: Latest changes adds more dynamic import for hilla-file-router and new `buildRoute` function in Flow.ts. A bit like Mikael suggested, but not in views.ts. Let's see how it works and let's improve it if needed. Point is to not require hilla-file-router and let Flow.ts handle the import automatically.

tltv:
> We don't need to check `serverSideRoutes` for `isHillaUsed`:
> 
> ```
>         var files = List.of(FrontendUtils.INDEX_TS, FrontendUtils.INDEX_TSX,
>                 FrontendUtils.ROUTES_TS, FrontendUtils.ROUTES_TSX);
>         for (String fileName : files) {
>           ....
>         }
> ```
> 
> Can be removed, as it switches to Vite for custom bundle with the new content in `routes.tsx`

We still need to check it to also detect the case when routes in `routes.tsx` is build manually without any views in Frontend/views folder.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -418,3 +419,4 @@ export const createWebComponent = (tag: string, props?: Properties, onload?: ()
     return React.createElement(tag);
 };
 
+//%buildRouteFunction%
```
caalador, line 422: no new line.
Just for usability I would perhaps have here 

```suggestion
export const buildRoute = (views: any): any => {
  //%buildRouteFunction%
}
```
 and the replace would just be `return views;` or `return toReactRouter(views);` to have as little coding in strings as possible. 
 
 Didn't see a need for `// @ts-ignore`

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -1,8 +1,34 @@
...
 ] as RouteObject[];
+if(route.children && route.children.length > 0) {
+  route.children.push(...serverSideRoutes);
+  routes.pop();
+  if (route.element) {
+    routes.push(route);
+  } else {
+    routes.push(...route.children);
+  }
+}
```
caalador, line 20: I would drop the serverSideRoutes check and move adding of serverSideRoutesTo the buildRoute method so that the default routes.tsx would not import serverSideRoutes nor views, only buildRoute

so it would look like:
```
export const routes = buildRoutes();
```

or  for the manual case
```
let routing = [
    {
        element: <MainLayout />,
        handle: { title: 'Main' },
        children: [
            { path: '/', element: <AboutView />, handle: { title: 'Welcome' } },
            { path: '/hilla', element: <HillaView />, handle: { title: 'Hilla' } }
        ],
    },
] as RouteObject[];

export const routes = buildRoute(routing, routing[0].children);
```


Then the buildRoute method when views is not used would look like:
```

export const buildRoute = (routes?: RouteObject[], serverSidePosition?: RouteObject[]): RouteObject[] => {
    const combinedRoutes = [] as RouteObject[];
    if(!routes) {
        const route: RouteObject = toReactRouter(views);
        if(route.children && route.children.length > 0) {
            serverSidePosition = route.children;
            if (route.element) {
                routes = [route];
            } else {
                routes = route.children;
            }
        }
    }
    if(serverSidePosition) {
        serverSidePosition.push(...serverSideRoutes);
    } else {
        combinedRoutes.push(...serverSideRoutes);
    }
    if(routes) {
        combinedRoutes.push(...routes);
    }
    return combinedRoutes;
};
```
where 
```
    if(!routes) {
        const route: RouteObject = toReactRouter(views);
        if(route.children && route.children.length > 0) {
            serverSidePosition = route.children;
            if (route.element) {
                routes = [route];
            } else {
                routes = route.children;
            }
        }
    }
``` 
would be the fill for `//%buildRouteFunction%` if hilla automatic views is used.

This would also remove the need for the routes export and circular link as we would be generating the routes in Flow.tsx and could just store it in the method call to be used.

caalador: Also the need for creating the empty views file would go away.

tltv: Moving logic out from the routes.tsx makes sense. This is what was suggested to be moved in views.ts instead, but for now we could move it to Flow.ts before the final design for default routes.tsx is decided.

Having both views and serverSideRoutes imported in routes.tsx was one point to keep them there as a reference. Now both are hidden behind this buildRoutes function. I assume imports can make a comeback later in some other form.

We still need to generate empty views file for apps using react without Hilla views. Just to get rid of that hilla-file-router import when it's not installed.

Missing serverSideRoutes also means that isHillaUsed check needs to do something else for routes.tsx. Or probably additional check just for routes.tsx based on the usage of buildRoute. This might be the biggest change in this all.

caalador: Without hilla views we would just need to not have 
```
    if(!routes) {
        const route: RouteObject = toReactRouter(views);
        if(route.children && route.children.length > 0) {
            serverSidePosition = route.children;
            if (route.element) {
                routes = [route];
            } else {
                routes = route.children;
            }
        }
    }
```
in the `buildRoute` method

So ideally it would be written as
```
export const buildRoute = (routes?: RouteObject[], serverSidePosition?: RouteObject[]): RouteObject[] => {
    const combinedRoutes = [] as RouteObject[];
    //%buildRouteFunction%
    if(serverSidePosition) {
        serverSidePosition.push(...serverSideRoutes);
    } else {
        combinedRoutes.push(...serverSideRoutes);
    }
    if(routes) {
        combinedRoutes.push(...routes);
    }
    return combinedRoutes;
};
```
Where the `%buildRouteFunction%` would be replaced with "" if no hilla views are available.

tltv: I understood that part. But import to views.js is still always there. I wouldn't touch that before we have some final design. It's better this way that you can always import it and Flow generates it if it's not there. But if it will be never used after all, then let's remove it and import dynamically.

tltv: This is done now except dependency to routes.tsx in Flow.ts is still there. It's tricky to remove so I suggest to do it separately after rest is done.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -84,6 +98,16 @@ public class TaskGenerateReactFiles implements FallibleCommand {
...
 
+    static final String VIEWS_TS_FALLBACK = """
+            const routes = { path: "", module: undefined, children: [] };
+            export default routes;
+            """;
+
+    private static Pattern SERVER_IMPORT_PATTERN = Pattern.compile(
+            "import[\\s\\S]?\\{[\\s\\S]?serverSideRoutes[\\s\\S]?\\}[\\s\\S]?from[\\s\\S]?(\"|'|`)Frontend\\/generated\\/flow\\/Flow(\\.js)?\\1;");
+    private static Pattern BUILDROUTE_PATTERN = Pattern.compile(
+            "import[\\s\\S]?\\{[\\s\\S]?buildRoute[\\s\\S]?\\}[\\s\\S]?from[\\s\\S]?(\"|'|`)Frontend\\/generated\\/flow\\/Flow(\\.js)?\\1;");
```
caalador, line 109: The regex is a bit bad  in that if you import 2 things from Flow then it will always fail e.g. if you have ` import { serverSideRoutes, buildRoute} from Flow`.
We should at a later point update this so that it is one regex that checks for serverSideRoutes or buildRoute and accepts other imports inside the `{}` block.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -418,3 +421,16 @@ export const createWebComponent = (tag: string, props?: Properties, onload?: ()
     return React.createElement(tag);
 };
 
+export const buildRoute = (routes?: RouteObject[], serverSidePosition?: RouteObject[]): RouteObject[] => {
```
caalador, line 424: JavaDoc could be useful for the parameters and what they are used for.
e.g. 
- routes are for adding own route definition, giving routes will skip FS routes
- serverSidePosition is the position where server routes should be put. If non given they go to the root of the routes []

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
caalador: The idea was to just either add a Flow.tsx parameter the same way we have 
let prevNavigation: string;
let popstateListener: { type: string, listener: EventListener, useCapture: boolean };
( this works fine for the routes as I did test it)
Or then you could store it on the window.Vaadin object as the value we want is the endresult from buildRoute

tltv: Split to new ticket: https://github.com/vaadin/flow/issues/18955

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -175,15 +204,44 @@ private void cleanup() throws ExecutionFailedException {
         }
     }
 
+    private String getFlowTsFileContent() throws IOException {
+        String content = getFileContent(FLOW_TSX);
+        if (FrontendUtils.isHillaUsed(options.getFrontendDirectory(),
+                options.getClassFinder())) {
+            return content.replace("//%toReactRouterImport%",
+                    "import { toReactRouter } from '@vaadin/hilla-file-router/runtime.js';")
```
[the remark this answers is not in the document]
tltv: Change to import it only if used.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -175,15 +204,44 @@ private void cleanup() throws ExecutionFailedException {
...
+                                                routes = [route];
+                                            } else {
+                                                routes = route.children;
+                                            }
+                                        }
+                                    }
+                                    """);
+        } else {
+            return content.replace("//%toReactRouterImport%", "")
+                    .replace("//%buildRouteFunction%", "");
```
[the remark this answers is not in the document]
tltv: Done

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
tltv: Done

tltv:
I'll add some new tests regarding buildRoute function but in other PR. No need to wait for those as cases are already tested manually with starters.

caalador (approved):

mshabarov (approved):
Works perfect for both cases, when I have default generated `routes.tsx` and if I uncomment explicit routing.

https://github.com/vaadin/flow/issues/18954 is a bit annoying indeed if I try to add serverSideRoutes by hand, but thanks to `buildRoutes` function I don't care anymore.

PR #18916, opened 2024-03-07
Update headers and add license checker for Flow 23.5

mcollovati:
**DO NOT SQUASH**: headers update and license checker addition are done in two separated commits


flow-client/src/main/frontend/index.ts
```diff
@@ -1 +1,10 @@
+///
+/// Copyright (C) 2000-2024 Vaadin Ltd
+///
+/// This program is available under Vaadin Commercial License and Service Terms.
+///
+/// See <https://vaadin.com/commercial-license-and-service-terms> for the full
+/// license.
+///
```
tepi, line 8: Not going to check all files, but is there a reason `.ts` files have `///` instead of `/* */`

mcollovati: The updates are done by license-maven-plugin, but I could not find the reason triple slash is used for `.ts` files 

https://github.com/mathieucarbou/license-maven-plugin/blob/master/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/document/DocumentType.java#L96C3-L96C42

The plugin allows custom mapping, so I suppose we can change it to `/* */` for typescript files

tepi: Alright, maybe it should not be changed then. I am not sure what the convention is for `.ts` files. Just caught my eye that it's a bit weird looking.

mcollovati: I'm not sure the triple slash is correct: https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html

tepi: Definitely looks not correct to use them for header

tepi (approved):

flow-client/src/main/frontend/index.ts
mcollovati: Now the headers are duplicate :joy: 
Let me clean up and push the changes

tepi: Right 😆 Wasn't that obvious the commit..

mcollovati: Now it should be fixed

tepi (approved):

PR #18919, opened 2024-03-08
feat: loadComponentScript returns promise

caalador:
Update loadComponentScrip to use
a promise for the script onload
and onerror for better handling.
Update createWebComponent to take
onerror callback for script loading.

part of #18816

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -403,7 +403,7 @@ interface Properties {
  * @param onload optional callback to be called for script onload
  * @param onerror optional callback for error loading the script
  */
-export const createWebComponent = (tag: string, props?: Properties, onload?: () => void, onerror?: (err) => void) => {
+export const createWebComponent = (tag: string, props?: Properties, onload?: () => void, onerror?: (err:any) => void) => {
     loadComponentScript(tag).then(() => onload?.(), (err) => {
         if(onerror) {
             onerror(err);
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -395,9 +401,17 @@ interface Properties {
  * @param tag custom web-component tag name.
  * @param props optional Properties object to create element attributes with
  * @param onload optional callback to be called for script onload
+ * @param onerror optional callback for error loading the script
  */
-export const createWebComponent = (tag: string, props?: Properties, onload?: () => void) => {
-    loadComponentScript(tag, onload);
+export const createWebComponent = (tag: string, props?: Properties, onload?: () => void, onerror?: (err) => void) => {
```
mshabarov, line 406: Can we have Promise for this method as well?

caalador: I don't think that is possible as it needs to return the React element

mshabarov: Well, true, when you need to return something, it's not convenient.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -395,9 +401,17 @@ interface Properties {
  * @param tag custom web-component tag name.
  * @param props optional Properties object to create element attributes with
  * @param onload optional callback to be called for script onload
+ * @param onerror optional callback for error loading the script
  */
-export const createWebComponent = (tag: string, props?: Properties, onload?: () => void) => {
-    loadComponentScript(tag, onload);
+export const createWebComponent = (tag: string, props?: Properties, onload?: () => void, onerror?: (err:any) => void) => {
```
mshabarov, line 406: Should be maybe `(err: Event | string)` ?

mshabarov (approved):

PR #18925, opened 2024-03-11
fix: add or update package.json type

### LINKED ISSUE #18924: PiT 24.4: flow does not update entry `"type": "module"` in `package.json`
### Description of the bug

Discovered when [testing](https://github.com/vaadin/hilla/issues/2193) `skeleton-starter-hilla-lit-gradle` in PiT, we have realised that flow since 24.1 is adding the `"type": "module"` when creating a new `package.json` but it is not updating existing ones, causing errors like 

```
Startup of Vite failed. Output was:
✘ [ERROR] "@vaadin/hilla-file-router/vite-plugin.js" resolved to an ESM file. ESM file cannot be loaded by `require`. See https://vitejs.dev/guide/troubleshooting.html#this-package-is-esm-only for more details. [plugin externalize-deps]

    node_modules/esbuild/lib/main.js:1373:27:
      1373 │         let result = await callback({
           ╵                            ^
```

[Slack discussion](https://vaadin.slack.com/archives/C065KMY4D2M/p1709901333027389)

### Expected behavior

projects upgraded to 24.4.0.alpha9+ that use the `hilla-file-router` do not fail if they don't have the entry already in their package.json

checkout branch https://github.com/vaadin/skeleton-starter-hilla-lit-gradle/tree/update-to-v24.4 and run ./gradlew bootRun twice 

### Versions

- Vaadin / Flow version: 24.4.0.alpha0

caalador:
If package json type is missing
or not set to module when upgrading
and modifying package.json
add/update type to module.

Fixes #18924

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGeneratePackageJson.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGeneratePackageJson.java
@@ -46,6 +46,15 @@ public class TaskGeneratePackageJson extends NodeUpdater {
             JsonObject mainContent = getPackageJson();
             modified = updateDefaultDependencies(mainContent);
             if (modified) {
+                if (!mainContent.hasKey("type")
+                        || !mainContent.getString("type").equals("module")) {
+                    mainContent.put("type", "module");
+                    log().info(
+                            """
+                                    Adding package.json type as module to enable ES6 modules which is now required.
+                                    With this change sources need to use 'import' instead of 'require' for imports.
+                                    """);
+                }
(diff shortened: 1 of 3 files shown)


tepi:
Should this check also be done in `com.vaadin.flow.server.frontend.NodeUpdater#getPackageJson`? There the type:module is added if package.json does not exist at all, but could not find a check for existing file.

caalador:
I feel we should only touch the type if we make modifications on the package.json, so added the check to TaskGeneratePackageJson also so that if we modify it there we update it. updating in the getPackageJson doesn't guarantee that it will be written to the file.

tepi (approved):

Issue #18871, opened 2024-03-04
A frontend/routes.tsx is always created


Artur-:
### Description of the bug

When starting the app when using Vite, I get a `frontend/routes.tsx` containing
```
import { createBrowserRouter, RouteObject } from 'react-router-dom';
import { serverSideRoutes } from "Frontend/generated/flow/Flow";

export const routes = [
  ...serverSideRoutes
] as RouteObject[];

export default createBrowserRouter(routes);
```

Is this intended? What should I do with the file? It looks auto generated

### Expected behavior

There is no file in `frontend` that I do not need

### Minimal reproducible example

```
npx @vaadin/cli init --pre testflow
cd testflow
echo "vaadin.frontend.hotdeploy=true" >> src/main/resources/application.properties
mvn
```

### Versions

- Vaadin / Flow version: 24.4 alpha 8

tltv:
It's intended when react is enabled then default `frontend/routes.tsx` is generated. I agree that it's not really wanted there in many cases.
Here are couple improvement options:
- deault `routes.tsx` template file could be copied to `frontend/generated` and `index-react.ts` template file changed to import it instead if project's `Frontend/routes.tsx` doesn't exist. 
- code in the default `routes.tsx` template could be moved to default `index-react.ts` template instead, but only if `Frontend/routes.tsx` doesn't exist.

mshabarov:
Before moving this file anywhere from frontend's root, we need to detect the location of this file (in root, or in generated folder or somewhere else) at build time and dynamically update the import in `Flow.tsx`, as `Flow.tsx` import routes from this file for it's needs.

PR #18936, opened 2024-03-11
feat: generate routes.tsx in Frontend/generated

tltv:
Generate default routes.tsx in Frontend/generated folder instead of Frontend/. Imports to routes.js are now dynamic in Flow.ts and index-react.ts.

Fixes: #18871


mshabarov:
@tltv could you please rebase on main branch?

mshabarov (approved):
Works as expected, when:
1. view is added, file is generated in generated folder
2. if I move it to frontend root, it is used
3. if I uncomment explicit routing, it is used
4. If I remove my view and remove all generated files, it's re-generated again in generated.

PR #18968, opened 2024-03-14
feat: do not require routes import

### LINKED ISSUE #18955: Avoid export of combinedRoutes in routes.tsx
### Describe your motivation

`routes.tsx` exports `routes` object so that `Flow.tsx` can import it and use for React router integration.
This can be avoided if `combinedRoutes` in `Flow.tsx` are stored outside of the method, so it can be reused in another place.
See [this comment](https://github.com/vaadin/flow/pull/18904#discussion_r1522630271).

### Additional context

Vaadin 24.4.
See this [pull request](https://github.com/vaadin/flow/pull/18904)

caalador:
Do not import routes from
routes.tsx if not needed.

Fixes tests so that files are not generated into the flow-server folder.

Fixes #18955

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
@@ -209,22 +209,7 @@ public class TaskGenerateReactFiles implements FallibleCommand {
             throws IOException, ExecutionFailedException {
         String content = getFileContent(FLOW_TSX);
 
-        // To accept usage of serverSideRoutes without using buildRoute by the
-        // user handle import.
-        if (customRoutesTsx.exists()) {
-            String customRoutes = StringUtil.removeComments(
-                    FileUtils.readFileToString(customRoutesTsx, UTF_8));
-            // If there is a manual serverSideRoutes used, but no buildRoute,
-            // import routes from routes file.
-            if (customRoutes.contains("serverSideRoutes")
-                    && !customRoutes.contains("buildRoute")) {
-                content = content
-                        .replace("//%routesImport%",
-                                "import { routes } from 'Frontend/routes.js';")
-                        .replace("let routes: RouteObject[];", "")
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -207,42 +205,71 @@ private void cleanup() throws ExecutionFailedException {
...
+                    && !customRoutes.contains("buildRoute")) {
+                content = content
+                        .replace("//%routesImport%",
+                                "import { routes } from 'Frontend/routes';")
+                        .replace("let routes: RouteObject[];", "")
+                        .replace("routes = combinedRoutes;", "");
+            }
+        } else if (options.isFrontendHotdeploy()
+                && !options.isProductionMode()) {
+            // For frontendHotdeploy we read the routes from file as re-loading
```
tltv, line 263: So this condition is needed now to fix reload. I don't understand exactly how reload can't work with having `let routes`. Could routes be stored in sessionStorage or localStorage instead? Those should survive reloads. Or maybe there's method to read routes directly from the React router.

tltv (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
caalador: Even using sessionStorage doesn't fix the issue. I don't understand why the update reloads Flow.tsx, but not routes.

tltv: Seems that for Vite the dependency is the key here to make the reload affect also routes.tsx. Maybe we could just keep the dependency there for the reload, but not use it actually, and with a comment telling that.

caalador: Only having the import did not make it work in the reload tests

Issue #18896, opened 2024-03-06
With a React based main layout, navigating to a Flow view from the route not found page does not work


Artur-:
### Description of the bug

When your main layout is in React, then the 404 page for an invalid URL is shown inside the mainlayout like this

![image](https://github.com/vaadin/flow/assets/260340/22a027d5-1161-4aff-b6df-1516fb68cd3f)

If you click on /flow then you end up on an empty view

![image](https://github.com/vaadin/flow/assets/260340/d0a9752c-e6fb-4d80-9dc5-44e442448500)

If you refresh, you get the actual view

![image](https://github.com/vaadin/flow/assets/260340/aba91c94-ebce-4ed9-ab2b-ba02daea3f2c)

### Expected behavior

You should see the view

### Minimal reproducible example

https://github.com/vaadin/flow-hilla-hybrid-example/tree/v24.4

### Versions

- Vaadin / Flow version: 24.4 alpha 17
- Java version:
- OS version:
- Browser version (if applicable):
- Application Server (if applicable):
- IDE (if applicable):

Issue #18966, opened 2024-03-14
@Route("") is not rendered if views/$layout.tsx is present


Legioth:
### Description of the bug

A Flow view with `@Route("")` is not rendered if the application has a client-side main layout (`views/$layout.tsx`). Flow views with other routes are rendered and the `""` view is also rendered if `$layout.tsx` is removed.

### Expected behavior

The view should render

### Minimal reproducible example

1. Get a starter with Vaadin 24.4.0.alpha14 and remove all Flow and/or Hilla views
2. Add a simple Flow view with `@Route("")`
3. Add a simple Hilla view in `views/hilla.tsx`
4. Launch and see that you can navigate to both views
5. Add a simple `views/$layout.tsx` file.
6. Observe that the Flow view no longer renders
7. Change the Flow view to e.g. `@Route("flow")` and notice that it renders again
8. Change back to `@Route("")` and disable the layout by renaming it to `_$layout.tsx`. Notice that the view renders again.

You can use this simple content for `$layout.tsx`
```
import { Outlet } from 'react-router-dom';

export default function MainLayout() {
    return <Outlet />
}
```

### Versions

- Vaadin / Flow version: Vaadin 24.4.0.alpha14
- Java version: 21

caalador:
So it seems that Hilla file routes always adds the `$layout.js` with the path `""` even though as a general layout it shouldn't have a path to it at all. So as this is registered to / then the Flow route / is not resolved as going for / will render the $layout and not flow which is under `/*`.

This actually caused another issue in the routing when instead of getting only match for `/*` we got a match for `/`and `/*` which was unexpected.

caalador:
Will wait for the result from vaadin/hilla#2238 if that handles the layout route correctly

caalador:
Right now running with 24.4.-SNAPSHOT just having a hilla view without a defined `@layout.js` will generate `createRoute("", [createRoute("hilla", Page0)]);` which overrides the server  / route removing the first `createRoute` to get `createRoute("hilla", Page0)` makes flow / route render as expected.

mshabarov:
Just tested as well with the 24.4.0.alpha21 and noticed no changes, since last testing: if I add `AtRoute("")` Flow view, having `@layout.tsx`, then I get a blank page when navigate to `/`. Changing the mapping to `/foo` makes the content rendered.
Thus, the issue is still valid.

mshabarov:
[...]

PR #18971, opened 2024-03-14
fix: fix anchor navigation with react router

tltv:
Fixed navigation via anchor to Flow views not rendering content. Removed unnecessary if-block from Flow.tsx that stopped rendering the view.

Fixes: #18896

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -245,9 +245,6 @@ export default function Flow() {
                 window.addEventListener('popstate', popstateHandler);
             }
         }
-        if(lastNavigation === pathname) {
-            return;
-        }
         flow.serverSideRoutes[0].action({pathname, search}).then((container) => {
             const outlet = ref.current?.parentNode;
             if (outlet && outlet !== container.parentNode) {
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -158,8 +158,11 @@ function navigateEventHandler(event) {
 
     // if navigation event route targets a flow view do beforeEnter for the
     // target path. Server will then handle updates and postpone as needed.
-    if(matched?.length == 1 && matched[0].route.path === "/*") {
+    if(matched && matched.length > 0 && matched[matched.length - 1].route.path === "/*") {
         if (mountedContainer?.onBeforeEnter) {
+            // ensure that menu highlighting is updated by calling navigation
+            navigation(event.detail.pathname, {replace: false});
+            // onBeforeEvent call will handle the Flow navigation
             mountedContainer.onBeforeEnter(
                 {
                     pathname: event.detail.pathname,
@@ -274,7 +277,7 @@ export default function Flow() {
 
             // if router force navigated using 'Link' we will need to remove
             // flow from the view
-            if(matched && matched[0].route.path !== "/*") {
+            if(matched && matched[matched.length - 1].route.path === "/*") {
                 mountedContainer?.parentNode?.removeChild(mountedContainer);
                 mountedContainer = undefined;
             }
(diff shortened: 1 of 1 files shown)


tltv:
Removed if-block was added originally along [test PR](https://github.com/vaadin/flow/pull/18321/files#diff-0287966eb47153e839f9ed5cc011d6d126737e69b299516a380188fbbbf8cc25).
Related integration tests works without this if-block (locally. CI runs tests atm so don't know the result at the time writing this).

@caalador do you remember what was the original purpose of this if-block check?

caalador:
That was when react router always generated 2 roundtrips to the server so we blocked the second navigation round.
Check that one navigation in flow only generates a single request that gets all the changes back and not 2 for a navigation. Not that this is about a pure flow application and hilla/react views should not be in the mix.

caalador:
Also check that the `vaadinRouterGlobalClickHandler(event)` is called for the link click and which path it takes.

tltv:
> That was when react router always generated 2 roundtrips to the server so we blocked the second navigation round. Check that one navigation in flow only generates a single request that gets all the changes back and not 2 for a navigation. Not that this is about a pure flow application and hilla/react views should not be in the mix.

There are two requests _only when navigating via anchor to Flow view_. First one leaves just blank page, second one updates it. Is this expected behavior or not? Should it be able to do it with one request instead?

> Also check that the vaadinRouterGlobalClickHandler(event) is called for the link click and which path it takes.

I'm not sure what path you mean exactly, but it reads the target navigation path from the correct anchor element there.

tltv:
image

First anchor click event is not detected as flow view here. That's why the second request is needed it seems. But could we detect this as a flow view in the first place here?

tltv:
Sorry to flood with comments @caalador but if I change the if-block in the previous comment's screenshot to this: `if(matched?.length > 0 && matched[matched.length - 1].route.path === "/*")` then anchor works with just one request.

Can't detect any side effects yet.

caalador:
So the main issue is that the matcher returns targets  / and /* where we expect only /*
Why would it suddenly start to return different results...

tltv:
fs-routing is now always used with hilla. buildRoute() combines server side routes in root "/" route's children array by default which means that `matchRoutes` function returns also the root path "/".

caalador:
I guess the fs-routing always setting the / is the reason we can't have Route("") in flow working

tltv:
My last change breaks navigation menu's automatic highlighting somehow. It worked before when it still though it was updating client side route. Now that it updates for server side route, highlighting is not working.

tltv:
Fixed highlighting issue. Also fixed wrong "!==" to "===" in code that is supposed to run special cleanup task only for Flow page.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -274,7 +277,7 @@ export default function Flow() {
 
             // if router force navigated using 'Link' we will need to remove
             // flow from the view
-            if(matched && matched[0].route.path !== "/*") {
+            if(matched && matched[matched.length - 1].route.path === "/*") {
```
caalador, line 280: This should be if not navigating to flow view. if we are targeting a flow view we should not clear the container.

tltv: Ok. It's confusing because it's not checking target path there but the currently shown Flow view path that is cleaned up. Looks like it's broken anyway.

caalador: The point here is that if we navigate to a flow view then the return message will contain everything and the view contents will be updated as they shoud, but if we are going to a non Flow view then we need to clean the Flow view from the dom as we will not be getting a uidl response.

[Fix outside navigation](https://github.com/vaadin/flow/pull/18387/commits/bac5a2ac3f3d05368823fda2d294337bf3838320)

tltv: Thanks for explaining. Updated code + added more comments to make it clear.

caalador (commented):
Looks ok.
not certain I like that fs routing always adds a / route as it makes things complicated.

tltv:
Highlighting NavLink in the react menu automatically when navigating via anchor in Flow view to another Flow view is now broken but I think that can be fixed separately. Problem is that now we never call react router's navigation and for some reason react router is not reacting to the location change to refresh the menu.

Afaik usage of "/" path for the main layout is React router's feature. We have also this one related to that https://github.com/vaadin/flow/issues/18966.

caalador (approved):

PR #18975, opened 2024-03-15
chore: Remove theme editor and show-in-code

Artur-:
It is now in Copilot


vaadin-dev-server/package.json
sissbruecker, line 1: - `web-test-runner.config.mjs` should also be removed
- You're removing the`@open-wc/dev-server-hmr` dependency which is used in `web-dev-server.config.mjs`
- There's also a `patch-application.js` and related NPM script that is only useful if the dev server and HMR is used.
- Looks like non of the Vaadin components in the dev / non-dev dependencies are used anymore, except for @vaadin/overlay.

sissbruecker: These dependencies also seem to be unused now: `vanilla-colorful`, `sinon`, `@types/sinon`, `@types/chai`

sissbruecker: A left-over module from the theme editor: `src/main/frontend/vendor/@vaadin/overlay/src/vaadin-overlay-position-mixin.js`

Artur-: Updated. Leaving web-dev-server there for now

MarcinVaadin (approved):

Issue #18986, opened 2024-03-19
PiT 24.4: flow modifications to types.d.ts throws an exception but does not stop server


manolo:
### Description of the bug

When upgrading projects created by a previous version of flow, it updates the types.d.ts file and it warns the user in the console showing an exception stacktrace, but it does not stop the server.
Messages in console is an endless stream of the same warning and stacktrace:

```

java.lang.IllegalStateException: com.vaadin.flow.server.ExecutionFailedException:

***************************************************************************
*  The TypeScript type declaration file 'types.d.ts' has been updated     *
*  to the latest version by Vaadin. Previous content has been backed up   *
*  on 'types.d.ts.flowBackup' file. Please verify that the updated        *
*  'types.d.ts' file contains configuration needed for your project, and  *
*  then delete the backup file.                                           *
***************************************************************************



	at com.vaadin.base.devserver.AbstractDevServerRunner.getCause(AbstractDevServerRunner.java:772) ~[vaadin-dev-server-24.4.0.alpha22.jar:na]
	at com.vaadin.base.devserver.AbstractDevServerRunner.getCause(AbstractDevServerRunner.java:768) ~[vaadin-dev-server-24.4.0.alpha22.jar:na]
	at com.vaadin.base.devserver.AbstractDevServerRunner.handleRequestInternal(AbstractDevServerRunner.java:631) ~[vaadin-dev-server-24.4.0.alpha22.jar:na]
	at com.vaadin.base.devserver.AbstractDevServerRunner.handleRequest(AbstractDevServerRunner.java:617) ~[vaadin-dev-server-24.4.0.alpha22.jar:na]
	at com.vaadin.flow.server.VaadinService.handleRequest(VaadinService.java:1577) ~[flow-server-24.4.0.alpha22.jar:24.4.0.alpha22]
	at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:398) ~[flow-server-24.4.0.alpha22.jar:24.4.0.alpha22]
	at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:106) ~[vaadin-spring-24.4.0.alpha22.jar:na]
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:658) ~[tomcat-embed-core-10.1.18.jar:6.0]
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:205) ~[tomcat-embed-core-10.1.18.jar:10.1.18]
[...]

PR #18987, opened 2024-03-19
Fix: update tsconfig and log message

### LINKED ISSUE #18984: PiT 24.4: when migrating a project to 24.4 modifications to tsconfig.json and types.d.ts should throw once
### Description of the bug

When frontend configuration files require updating, Flow does not perform these updates in a batch. Instead, it executes separate modifications and throws errors, necessitating the user to run the application multiple times after the migration.

NOTE: it makes several projects run with the PiT script fail because it only considers one extra run

### Expected behavior

It should throw just once

The attached project is a 24.3 project with the needed modifications to run in 24.4, but types.d.ts, tsconfig.json and package.json have not been modified.
Unzip the project and run mvn, you need to run 3 times

[latest-lit-top.zip](https://github.com/vaadin/flow/files/14647570/latest-lit-top.zip)

### Versions

- Vaadin / Flow version: 24.4.0.alpha15

caalador:
Instead of updating and throwing
exception for 'tsconfig.json'.
Generate .bak file,
update file and log info if the
file is updated.

Fixes #18984

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
@@ -44,7 +44,7 @@ public class TaskGenerateTsConfig extends AbstractTaskClientGenerator {
      * Keeps track of whether a warning update has already been logged. This is
      * used to avoid spamming the log with the same message.
      */
-    private static boolean warningEmitted = false;
+    protected static boolean warningEmitted = false;
 
     private static final String COMPILER_OPTIONS = "compilerOptions";
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
@@ -52,15 +52,17 @@ public class TaskGenerateTsConfig extends AbstractTaskClientGenerator {
             "v23.3.0", "v23.2", "v23.1", "v22", "v14", "osgi", "v23.3.4",
             "v23.3.4-hilla" };
 
-    //@formatter:off
-    static final String ERROR_MESSAGE =
-            "%n%n**************************************************************************"
-            + "%n*  TypeScript config file 'tsconfig.json' has been updated to the latest *"
-            + "%n*  version by Vaadin. Please verify that the updated 'tsconfig.json'     *"
-            + "%n*  file contains configuration needed for your project (add missing part *"
-            + "%n*  from the old file if necessary) and restart the application.          *"
-            + "%n**************************************************************************%n%n";
-    //@formatter:on
+    static final String ERROR_MESSAGE = """
+
+            **************************************************************************
+            *  TypeScript config file 'tsconfig.json' has been updated to the latest *
+            *  version by Vaadin. Please verify that the updated 'tsconfig.json'     *
+            *  file contains configuration needed for your project (add any missing  *
+            *  parts from the old file if necessary) and restart the application.    *
+            *  Old configuration is stored as a '.bak' file.                         *
+            **************************************************************************
+
+            """;
 
     private Options options;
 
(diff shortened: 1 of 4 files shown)


mshabarov:
I'd suggest to also align/unify how back-up file is created and which format is used for `tsconfig.json` and `types.d.ts`.
Also, it would make sense to log/throw only once in the end, giving a bunch of messages to the user.

See my comments here https://github.com/vaadin/flow/issues/18984#issuecomment-2006594242.

mshabarov:
Collecting the logs can be done separately, as now we have the [issue](https://github.com/vaadin/flow/issues/18986) for it :)

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
```diff
@@ -52,15 +58,17 @@ public class TaskGenerateTsConfig extends AbstractTaskClientGenerator {
...
-    static final String ERROR_MESSAGE =
-            "%n%n**************************************************************************"
-            + "%n*  TypeScript config file 'tsconfig.json' has been updated to the latest *"
-            + "%n*  version by Vaadin. Please verify that the updated 'tsconfig.json'     *"
-            + "%n*  file contains configuration needed for your project (add missing part *"
-            + "%n*  from the old file if necessary) and restart the application.          *"
-            + "%n**************************************************************************%n%n";
-    //@formatter:on
+    static final String ERROR_MESSAGE = """
+
```
mshabarov, line 62: nit: This has one line break before and after, whereas the similar banner for types.d.ts has two line breaks, would it be better to align them so they both have one or two lines each?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsDefinitions.java
```diff
@@ -171,8 +176,10 @@ private void updateIfContentMissing() throws ExecutionFailedException {
                         Files.writeString(tsDefinitions.toPath(),
                                 uncommentedDefaultContent,
                                 StandardOpenOption.APPEND);
-                        if (updateMode == UpdateMode.UPDATE_AND_THROW) {
-                            throw new ExecutionFailedException(UPDATE_MESSAGE);
+                        if (updateMode == UpdateMode.UPDATE_AND_BACKUP
+                                && !warningEmitted) {
+                            log().warn(UPDATE_MESSAGE);
```
mshabarov, line 181: Misses String.format: `log().warn(String.format(UPDATE_MESSAGE));`
Or `TaskGenerateTsConfig` has redundant String.format, hmm

mshabarov (dismissed):
Works as expected to me: messages are logged one time, each has backup file, each has the same format of the file and unique ID.

Couple of nits, feel free to merge, if you think they are non-sense.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsConfig.java
caalador: Aligned both to have the same single as used normally.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsDefinitions.java
caalador: Removed unneeded string format

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsDefinitions.java
```diff
@@ -43,7 +42,6 @@ public class TaskGenerateTsDefinitions extends AbstractTaskClientGenerator {
     private static final String DECLARE_CSS_MODULE = "declare module '*.css?inline' {";
     static final String UPDATE_MESSAGE = """
 
-
```
mshabarov, line 46: This message should contains old "flowBackup" suffix. Should contain the same text, as for tsconfig: "Old configuration is stored as a '.bak' file."

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsDefinitions.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsDefinitions.java
```diff
@@ -43,21 +42,18 @@ public class TaskGenerateTsDefinitions extends AbstractTaskClientGenerator {
...
             ***************************************************************************
             *  The TypeScript type declaration file 'types.d.ts' has been updated     *
             *  to the latest version by Vaadin. Previous content has been backed up   *
-            *  on 'types.d.ts.flowBackup' file. Please verify that the updated        *
-            *  'types.d.ts' file contains configuration needed for your project, and  *
-            *  then delete the backup file.                                           *
+            *  as a '.bak' file.. Please verify that the updated 'types.d.ts' file    *
```
mshabarov, line 48:
```suggestion
            *  as a '.bak' file. Please verify that the updated 'types.d.ts' file    *
```

mshabarov (approved):

Issue #18988, opened 2024-03-19
Comment in generated routes.tsx is confusing


Legioth:
### Description of the bug

There's a comment in the generated `frontend/generated/routes.tsx` file about making your own customizations to the file. The problem is that the user should only make changes to `frontend/routes.tsx` but not to the file in `frontend/generated/`.

### Expected behavior

The comment should first guide the user to create a copy of the file and only the modify that copy.

### Minimal reproducible example

N/A

### Versions

- Vaadin / Flow version: 24.4.0.alpha15

mshabarov:
Comments added in https://github.com/vaadin/flow/pull/19020.

PR #18992, opened 2024-03-19
chore: Clarify that the file should be copied before modification

mshabarov:
Fixes https://github.com/vaadin/flow/issues/18988


flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -1,3 +1,16 @@
+/******************************************************************************
+ * This file is auto-generated by Vaadin.
+ * If you want to customize the routes manually, you have to:
+ * - copy this file or create your own `routes.tsx` in your frontend directory;
+ * - modify the copied or created file - uncomment the example code below and
+ *   customize it according to your routes.
```
caalador, line 6: I guess this part is a bit extra as they want to modify it. better would perhaps be to describe the `buildRoute` method usage instead

mshabarov: the original missing part was the instruction to copy the file, if one wants to customise it. I propose to drop words "uncomment..." and give an instruction for buildRoute, saying that the example is given below.

mshabarov: I've added more description for buildRoutes. 

However, I don't understand at the moment that what would be a profit of adding server side routes to an arbitrary level of hierarchy, what would it give to me?

Would it mean that if I add them onto one level down in the hierarchy, I could have e.g. `/admin/*` that is mapped to server routes. So I could have `AtRoute("flow")` and then navigate to `/admin/flow` successfully, but then `/flow` wouldn't be available?

I thought that I can set something like this:
```
function Inner() {
    return <div>Hello</div>;
}

let routing = [
    {
        element: <Layout />,
        handle: { title: 'Main' },
        children: [
            { path: '/', element: <Public />, handle: { title: 'Public' } },
            { path: '/admin', children: [{ path: '/hello', element: <Inner/>} ] }
        ],
    },
    { path: '/login', element: <Login />, handle: { title: 'Login' } }
] as RouteObject[];

export const routes = buildRoute(routing, routing[0].children[1].children);
```
And then navigate to `/admin/flow` and this gives me my Flow UI, but this doesn't work.
So I'm a bit lost.

caalador: It's the place in the layout where flow is rendered that is defined.
The path building should work, but is the issue here that admin is missing the `element`?
Navigation inside Flow in the nested path might not work as expected though.

mshabarov: This configuration at least doesn't crash the browser:
```
let routing = [
    {
        element: <Layout />,
        handle: { title: 'Main' },
        children: [
            { path: '/', element: <Public />, handle: { title: 'Public' } },
            { path: '/admin', element: <InnerLayout/>, routes: [{ path: '/hello', element: <Inner/>} ] }
        ],
    }
] as RouteObject[];
```
it, however, cannot find `/admin/flow`, nor `/admin/hello`. It renders `/flow` fine, but without any layout.
Almost the same config, but with `children`, crashes the page without any errors:
```
let routing = [
    {
        element: <Layout />,
        handle: { title: 'Main' },
        children: [
            { path: '/', element: <Public />, handle: { title: 'Public' } },
            { path: '/admin', element: <InnerLayout/>, children: [{ path: '/hello', element: <Inner/>} ] }
        ],
    }
] as RouteObject[];
```
Maybe I just misconfigured the routes, because even if I inject server side routes into `routing[0].children`, the pages is crashed.

mshabarov: Maybe we shouldn't recommend at this point to inject Flow server-side entry to any arbitrary level deeper than `routing[0].children` ?

mshabarov: This API however is going to be improved soon via https://github.com/vaadin/hilla/issues/2238, so should maybe postpone this PR.

PR #18997, opened 2024-03-20
fix: prevent runWhenAttached command to execute immediately by detach listeners

### LINKED ISSUE #18983: Grid-Component : Errors with Drag-And-Drop and Item-Selection
### Description of the bug

A View has a grid for which a Drag-And-Drop with DragMode.BETWEEN is configured and that will navigate to another view on item click. Clicking on an item leads to several javascript erros in the browser console.
In our application this had the effect, that some checkboxes could not be clicked afterwards (something got broken), but I could not reproduce this in the showcase.
if DragAndDrop is deactivated (see if(true)-Statement in class PersonGridDragDrop no errors occur.
[example4.zip](https://github.com/vaadin/flow/files/14329862/example4.zip)

### Expected behavior

I expect that no javascript errors occur on the console

See attached example

### Versions

- Vaadin / Flow version: 23.3.33
- Java version: 17
- OS version: Windows
- Browser version (if applicable): Firefox
- Application Server (if applicable): Tomcat
- IDE (if applicable): Eclipse

mcollovati:
Fixes #18983

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java
@@ -344,17 +344,23 @@ public class StateNode implements Serializable {
         List<StateNode> nodes = new ArrayList<>();
         visitNodeTreeBottomUp(nodes::add);
         nodes.forEach(StateNode::handleOnDetach);
-        for (StateNode node : nodes) {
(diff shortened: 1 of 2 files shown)


sissbruecker (approved):
Tested with the reproduction from the issue and can confirm that it fixes the problem

Issue #18999, opened 2024-03-20
Browser back navigation doesn't work after navigating on server-side


sissbruecker:
### Description of the bug

When using `UI.navigate` and browser back navigation multiple times, the view doesn't update accordingly. For example:
- Button click -> `UI.getCurrent().navigate(AboutView.class)`
- Shows about view
- Navigate back
- Shows previous view (correct)
- Button click -> `UI.getCurrent().navigate(AboutView.class)`
- Shows about view
- Navigate back
- Still shows about view (wrong)

While the URL updates correctly, Flow doesn't seem to do any updates. There's no UIDL request, no DOM updates, the page title doesn't update.

### Expected behavior

It should always show the previous view when navigating back in the browser.

### Minimal reproducible example

```java
@PageTitle("Hello World")
@Route(value = "hello", layout = MainLayout.class)
@RouteAlias(value = "", layout = MainLayout.class)
public class HelloWorldView extends Div {
    public HelloWorldView() {
        Button navigate = new Button("Navigate to About");
        navigate.addClickListener(e -> {
            UI.getCurrent().navigate(AboutView.class);
        });
        add(navigate);
    }
}
```

### Versions

- Vaadin / Flow version: 24.4.0.alpha14 (also tested with snapshot)
- Browser: Tested with Chrome + FF

PR #19007, opened 2024-03-21
fix: preserve last navigation path with react

tltv:
Fixes: #18999


flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/PopStateHandlerIT.java
```diff
@@ -160,8 +160,7 @@ public void testEmptyHash_quadrupleBack_noHashServerToServer() {
 
         verifyPopStateEvent(FORUM);
         // NOTE: see https://github.com/vaadin/flow/issues/10865
-        // verifyInsideServletLocation(isClientRouter() ? FORUM : EMPTY_HASH);
```
tltv, line 163: FYI @caalador I'm afraid that this test has been false-green for some time. I reverted the commented line back as with this fix for the back-navigation, it should work now again as before.

caalador (approved):

Issue #19009, opened 2024-03-21
Maven targets use wrong jars when Flow version has been overridden


Artur-:
### Description of the bug

If you create an application using Vaadin 24 alpha 15 and override the Flow version from alpha22 to alpha23, then `mvn vaadin:build-frontend` will generate package.json using Flow alpha22 and not alpha23 (e.g. `@babel/preset-react` will be `^7` and not `7.23.3`).

This apparently happens because that the class loader created using the URLs from the project compile path (these URLs are correct and include Flow alpha 23) is set to use the context class loader as the parent class loader here: https://github.com/vaadin/flow/blob/ba28266fc9156fb652e0cf086c397dfa309dcf1f/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/server/scanner/ReflectionsClassFinder.java#L59-L60

Now the context class loader apparently contains the plugin dependencies and as the plugin was built using Flow alpha22, it contains a reference to that. Class loader specs also say that any class loader should first try the parent class loader, so when trying to find e.g. `com/vaadin/flow/server/frontend/dependencies/vite/package.json` then the created class loader will first consult the parent, which finds flow-server alpha22 and returns the URL to the resource.

### Expected behavior

The project classpath would be used, not the classpath used to create the plugin

### Minimal reproducible example

New project + Flow override, as described

### Versions

- Vaadin / Flow version:
- Java version:
- OS version:
- Browser version (if applicable):
- Application Server (if applicable):
- IDE (if applicable):

PR #19010, opened 2024-03-21
fix: Classes should not be loaded from the plugin

Artur-:
Related-to https://github.com/vaadin/flow/issues/19009


Artur-:
Does @mcollovati, @platosha  or somebody else know why this should not be done? The current implementation finds Flow classes the plugin was built with and not from the project. No tests here fail, does Hilla rely on this somehow?

mcollovati:
The only reason I can see against `null` parent is that the provided URLs might not cover all possible location for parent classes, so a specific class cannot be loaded because the classloader cannot find one ancestor.

Looking at the flow maven plugin code, it looks like we provide all runtime dependencies (`compile` + `runtime` scopes) but we filter out `provided` and `system` to only get portlet and servlet-api.
However, according to the [documentation](https://maven.apache.org/guides/mini/guide-maven-classloading.html#Plugin_Classloaders), the plugin class loader should not contain the project dependencies, so the execution may fail anyway for dependencies excluded from URL. Unless the same dependencies are defined as plugin dependencies.

I wonder if we should provide to `ReflectionsClassFinder` also a class loader, and create that one based on all `MavenProject` dependencies.
Or temporarily change the context class loader during mojo execution.

mcollovati:
The Surefire plugin for in-process execution uses a custom class loader, temporarily set as the current thread context class loader. The classloader is built on project dependencies (` getProject().getArtifacts()`).

```java
        @Override
        public RunResult invoke(Object forkTestSet) throws ReporterException, InvocationTargetException {
            ClassLoader current = swapClassLoader(testsClassLoader);
            try {
                Method invoke = getMethod(providerInOtherClassLoader.getClass(), "invoke", INVOKE_PARAMETERS);
                Object result = invokeMethodWithArray2(providerInOtherClassLoader, invoke, forkTestSet);
                return (RunResult) surefireReflector.convertIfRunResult(result);
            } finally {
                if (System.getSecurityManager() == null) {
                    Thread.currentThread().setContextClassLoader(current);
                }
            }
        }

        private ClassLoader swapClassLoader(ClassLoader newClassLoader) {
            ClassLoader current = Thread.currentThread().getContextClassLoader();
            Thread.currentThread().setContextClassLoader(newClassLoader);
            return current;
        }
```

Artur-:
and presumably the built test classloader does not use the plugin classloader as its parent?

mcollovati:
If I got it correctly, it uses the platform classloader as parent.

https://github.com/apache/maven-surefire/blob/4281a9bdd41a5cc4f17f4ed101ee9e8a1a089aa5/surefire-booter/src/main/java/org/apache/maven/surefire/booter/Classpath.java#L120-L136

https://github.com/apache/maven-surefire/blob/master/surefire-booter/src/main/java/org/apache/maven/surefire/booter/IsolatedClassLoader.java

Artur-:
Added the platform classloader as parent..

Artur-:
Should we merge this for 24.6 or try to figure out if it breaks something or is there some better approach?

mcollovati:
Building a hybrid application with this change results in an error

```
[ERROR] Failed to execute goal com.vaadin:vaadin-maven-plugin:24.6-SNAPSHOT:prepare-frontend (default) on project spring-petclinic-vaadin-flow: Could not execute prepare-frontend goal.: Error occured during goal execution: class com.vaadin.hilla.internal.EndpointGeneratorTaskFactoryImpl cannot be cast to class com.vaadin.flow.server.frontend.EndpointGeneratorTaskFactory (com.vaadin.hilla.internal.EndpointGeneratorTaskFactoryImpl is in unnamed module of loader java.net.URLClassLoader @2b20caed; com.vaadin.flow.server.frontend.EndpointGeneratorTaskFactory is in unnamed module of loader org.codehaus.plexus.classworlds.realm.ClassRealm @78d92eef)
```

mcollovati:
I think the missing piece here is to make the Flow mojos set the `ClassFinder` as thread context class loader.
At least, this seems to work for Flow projects: for example, if the project uses Vaadin 24.5.0 but sets flow-maven-plugin to version 24.6-SNAPSHOT, the resources are then correctly collected from the 24.5 `flow-server` jar.

I wonder if the previous error can be fixed by applying the same pattern to the Hilla `configure` mojo, that seems to be executed before `prepare-fronted`, and probably is loading the `EndpointGeneratorTaskFactory` into a shared classloader.

mcollovati:
Isolating the execution so that flow-server classes are not loaded into the plugin class loader seems to be more complicated than expected.

mcollovati:
Changing the thread context class loader during mojo execution does not work. flow-server classes are still loaded into the plugin class loader, causing a `ClassCastException` when Hilla or Copilot classes are loaded.

mcollovati:
Also, making `flow-server` as provided dependency seems not to work. In addition, we might end up in situations where the plugin might fail or misbehave because of incompatibilities between the `flow-server` version expected by the plugin and the one used by the end-user project.

mcollovati:
Superseded by #20465

PR #19020, opened 2024-03-22
feat: Use a new RouterBuilder API for routes.tsx

mshabarov:
Removes the `buildRoutes` method and replaces it with a new [RouterBuilder](https://github.com/vaadin/hilla/pull/2250) API that automatically injects server side routes where needed and adds the auth level.

Related-to https://github.com/vaadin/hilla/issues/2238

Fixes https://github.com/vaadin/flow/issues/18988


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -96,15 +84,22 @@ public class TaskGenerateReactFiles implements FallibleCommand {
...
-            const routes = { path: "", module: undefined, children: [] };
-            export default routes;
+    static final String FILE_ROUTES_TS_FALLBACK = """
+            const fileRoutes = { path: "", module: undefined, children: [] };
+            export default fileRoutes;
+            """;
+
+    static final String SERVER_ROUTES_TS = """
+            import { serverSideRoutes } from "Frontend/generated/flow/Flow";
+            export const serverRoute = serverSideRoutes;
```
caalador, line 98: Why do we  have this extra file that just exports an import from Flow that could be just imported from Flow?
If it is the name I think we should just change it in Flow.tsx
Same goes for Flow.tsx which can just be renamed instead of hidden.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -141,12 +138,18 @@ private void doExecute() throws ExecutionFailedException {
             if (fileAvailable(REACT_ADAPTER_TSX)) {
                 writeFile(reactAdapterTsx, getFileContent(REACT_ADAPTER_TSX));
             }
-            if (!viewsTs.exists()) {
-                writeFile(viewsTs, VIEWS_TS_FALLBACK);
+            if (!fileRoutesTs.exists()) {
+                writeFile(fileRoutesTs, FILE_ROUTES_TS_FALLBACK);
+            }
```
caalador, line 143: Do we still need the routes fallback as I don't see a must link to it anywhere?

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateReactFilesTest.java
```diff
@@ -92,90 +100,33 @@ public void routesContainImport_serverSideRoutes_noExceptionThrown()
...
-
-        FileUtils.write(routesTsx, content, StandardCharsets.UTF_8);
-
-        TaskGenerateReactFiles task = new TaskGenerateReactFiles(options);
-
-        task.execute();
-    }
-
-    @Test
-    public void routesContainMultipleFlowImports_noExceptionThrown()
```
caalador, line 154: This should be left and test that if you need/want multiple imports from Flow (or the other file) that it does pass the requirement of the regex.

tltv:
`FrontendUtils#isRoutesTsxContentUsingHillaViews()` has regex checks that should be changed now to support new format. Maybe it's enough to check if `.withReactRoutes(...)` is called. 
Also part in `FrontendUtils#isHillaViewsUsed(File)` that check if there are any views in the Frontend/views/ should now check if `.withFileRoutes(...)` is actually called in addition to having views files.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
mshabarov: This pursues the goal to have similar imports for FS router and server routes, i.e. server routes import name and a file name from where it is imported: `file-routes.js` and `server-routes.js`.
This is just a matter of better naming for objects, publicly exposed in routes.tsx.

@platosha what do you think? Was there anything more complex behind this new import name, rather than just a "unified" name?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
mshabarov: Yes, seems like it's not needed anymore, as Hilla uses internally in RouterBuilder and Flow apps don's use it.
Will remove it.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateReactFilesTest.java
mshabarov: Well, no, the regexp now checks for exactly one import (serverRoute), multiple imports aren't expected.
Do you think we should allow the regexp to have multiple imports still ?

caalador: I think we should as we should not be limiting, but requiring that the expected import is there.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
mshabarov: Decided to revert back to `import { serverSideRoutes } from 'Frontend/generated/flow/Flow';` to simplify this PR.

`serverSideRoutes` was used with older versions of Hilla, and now I'm confident that it's a good idea to change it.

We can discuss offline and rename it in the separate PR, if needed.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
mshabarov: Removed.

flow-server/src/test/java/com/vaadin/flow/server/frontend/TaskGenerateReactFilesTest.java
mshabarov: Brought back.

mshabarov:
> FrontendUtils#isRoutesTsxContentUsingHillaViews() has regex checks that should be changed now to support new format

This now checks that routes.tsx has `withFileRoutes` or `withReactRoutes` method that means Hilla.

mshabarov:
Tested with different configurations and routes.tsx content is as expected, also triggers an error if manual changes are incorrect.
Ready for review I think and should wait for RouterBuilder.

caalador (approved):

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -1,18 +1,56 @@
...
+    // .withReactRoutes([
+    //     {
+    //         element: <MainLayout />,
+    //         handle: { title: 'Main' },
+    //         children: [
+    //             { path: '/hilla', element: <HillaView />, handle: { title: 'Hilla' } }
+    //         ],
+    //     },
+    // ])
+    .withServerFallback(serverSideRoutes)
```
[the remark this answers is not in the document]
caalador: Sounds like a strange handling from the builder. So we can't give the server path?
This sounds like it will be a bit more fragile when the path definition is in 2 places and is easily changed.

This also makes routes-flow and this differ considerably and here we should then just import Flow and give <Flow/> to the withServerFallback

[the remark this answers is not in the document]
mshabarov: [This fix](https://github.com/vaadin/flow/pull/19063) covers the problem with `/*`, thanks @caalador 👍 

About naming, as we discussed offline, it makes sense to have method with a name `withFallbackComponent` that takes a component (not necessarily Flow route) and fallbacks React Router to it, if no client-side route is matched. E.g. this can be a view that shows custom 404 page.

So having two options
- `import Flow from 'Frontend/generated/flow/Flow';` or
- `import { serverSideRoutes } from 'Frontend/generated/flow/Flow';`

might be fine, depending if it's Hilla or Flow.
We need to count this in the routes.tsx content validation.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -1,18 +1,56 @@
...
+ * explicit routes configuration with the server-side routes.
+ *
+ * It has the following methods:
+ * - `withFileRoutes` enables the File System-based routes autoconfiguration;
+ * - `withReactRoutes` adds manual explicit route hierarchy. Allows also to add
+ * an individual route, which then merged into File System-based routes,
+ * e.g. Log In view;
+ * - `withServerFallback` adds server-side routes automatically to either
+ * autoconfigured routes, or manually configured routes;
+ * - `protect` adds an authentication later to the routes.
```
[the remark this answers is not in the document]
mshabarov: Makes sense, will add more explanations.
And, once we'll have online docs available, it makes sense also to put a link.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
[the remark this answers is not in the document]
caalador: So the regex should be something like: `import[\s\S]?(?:{[\s\S]*serverSideRoutes[\s\S]*}|Flow)+[\s\S]?from[\s\S]?("|'|`)Frontend\/generated\/flow\/Flow(\.js)?\1` so that it matches both imports as a either or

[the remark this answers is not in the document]
mshabarov: I'd make it so that anything that imported from Flow.tsx with a default import, should match what is passed into `.withServerRoutes`. As dev can change it from Flow to something else.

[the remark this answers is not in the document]
caalador: Sounds like something that should not be done in a single regex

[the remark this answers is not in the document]
mshabarov: Tested with `import\\s+(\\w+)\\s+from\\s+(\"|'|`)Frontend\\/generated\\/flow\\/Flow(\\.js)?\\2;[\\s\\S]+withFallbackComponent\\(\\s*\\1\\s*\\)")`, seems to do the job.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -249,7 +216,16 @@ private boolean fileAvailable(String fileName) {
...
+        routesContent = StringUtil.removeComments(routesContent);
+        if (usesRouterBuilder(routesContent)) {
+            return !FALLBACK_COMPONENT_PATTERN.matcher(routesContent).find();
+        } else {
+            return !SERVER_ROUTE_PATTERN.matcher(routesContent).find();
+        }
+    }
+
+    private boolean usesRouterBuilder(String routesContent) {
+        return routesContent.contains("new RouterBuilder(");
```
caalador, line 228: Should we just always check both matchers e.g.
```suggestion
        routesContent = StringUtil.removeComments(routesContent);
        return !FALLBACK_COMPONENT_PATTERN.matcher(routesContent).find() && !SERVER_ROUTE_PATTERN.matcher(routesContent).find();
```
Also fallback component pattern checks more than server route as it checks that there is a with method call.
Perhaps SERVER_ROUTE_PATTERN should also check that there is a `...serverSideRoutes` used in the routes file.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -1,18 +1,60 @@
...
+    //         children: [
+    //             { path: '/hilla', element: <HillaView />, handle: { title: 'Hilla' } }
+    //         ],
+    //     },
+    // ])
+    .withFallbackComponent(Flow)
+    // To add individual routes, use the following code:
+    // .withReactRoutes([
+    //     { path: '/login', element: <Login />, handle: { title: 'Login' } },
+    // ])
```
caalador, line 53: Should the withReactRoutes samples be one and the same saying that you can define manually and/or add individual routes.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -1,18 +1,60 @@
...
+    .protect();
```
caalador, line 54: Also protect should probably be marked as optional to have?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
mshabarov: Yep, no complaints against this suggestion.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
mshabarov: Yep, I can merge them into one sample, but still believe that these are two different cases - having overall manual config and adding in the individual route.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
mshabarov: Makes sense.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
caalador: Can you run with multiple withReactRoutes calls and have all the routes get added?

mshabarov: From the quick implementation check, it seems like I can call this method multiple times, never tested though.

I don't think that in practice you'll need to call them twice. You either have, for instance, full routes config with login path included, or file rotes + single login path.

tltv (changes requested):

flow-server/src/main/resources/com/vaadin/flow/server/frontend/routes.tsx
```diff
@@ -1,18 +1,62 @@
...
-//    handle: { title: 'Main' },
-//    children: [
-//      { path: '/hilla', element: <HillaView />, handle: { title: 'Hilla' } }
-//    ],
-//  },
-// ] as RouteObject[];
-// export const routes = buildRoute(routing, routing[0].children);
-
-export default createBrowserRouter(routes);
+export const { router, routes } = new RouterConfigurationBuilder()
```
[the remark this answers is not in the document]
mshabarov: Changed to `import { router } ...` in `generated `index.tsx` file.
`Flow.tsx` has already the `import { routes } ...`, that is fine.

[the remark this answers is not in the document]
mshabarov: default routes content for Flow was changed to:
```
import { createBrowserRouter, RouteObject } from 'react-router-dom';
import { serverSideRoutes } from 'Frontend/generated/flow/Flow';

function build() {
    const routes = [...serverSideRoutes] as RouteObject[];
    return {
        router: createBrowserRouter(routes),
        routes
    };
}
export const { router, routes } = build()
```

to have the same exports as for Hilla.

tltv:
I have this error in browser console with `flow-hilla-hybrid-example`, based on your branch `use-routes-builder`. Main page looked ok with one public view listed. Then sign in page gives me this when clicking Log in button:
image

mshabarov:
I don't see this error, when logged in. This looks as something related to the LoginOverlay component.
Could you try in incognito mode or clear browser cache?

tltv:
> UPD: This might be related to `AuthContext` from `hilla-react-auth`, let me check.

Not a cache issue as I get same error with Edge/Firefox. Yeah AuthContext is not setup for some reason. Maybe I run with old `hilla-react-auth`.

mshabarov:
Try to clean up the project and re-run. I've just tested with the latest Hilla snapshot.

mshabarov:
@caalador could you please also test?

tltv:
> @tltv could you please also double check if you have `AuthProvider` in the index.tsx?
> ...
> This is required.

Ok I didn't use the custom index.tsx. It's not there in the generated index.tsx and somehow I expected it to be correct automatically. After copying default index.tsx and adjusting like that, it works. Shouldn't Flow auto-generate this part too?

mshabarov:
This is problematic, as `AuthProvider` requires other codes that needs user information. You can check `auth.tsx` in the starter project. So we cannot easily add it to any project, because it is project-specific.

mshabarov:
Why don't you use `frontend/index.tsx` file by the way? It should be in the https://github.com/vaadin/flow-hilla-hybrid-example/pull/29

mshabarov:
And even in the v24.4 branch https://github.com/vaadin/flow-hilla-hybrid-example/blob/v24.4/src/main/frontend/index.tsx.

tltv:
> Why don't you use `frontend/index.tsx` file by the way? It should be in the [vaadin/flow-hilla-hybrid-example#29](https://github.com/vaadin/flow-hilla-hybrid-example/pull/29)

I'm testing that generated files works too.

mshabarov:
These are good comments and something that we should investigate.
But not in this PR ofc, as this PR is about using the builder API and adapting to it.

PR #19021, opened 2024-03-22
fix: reorder unhandled imports in CSS bundler

sissbruecker:
Currently `CssBundler` just replaces `@import`s with inlined file contents and returns the result. That can be problematic if an unprocessed import, for example to an external URL, comes after an inlined import. Now the unprocessed import comes after a regular CSS rule, which is invalid and ignored by the browser:
```css
/* inlined import content */
body { background: blue; }
/* unprocessed import */
@import url('https://cdn.jsdelivr.net/fontsource/css/aclonica@latest/index.css');
```

This change reorders unhandled imports so that they are at the top of the file:
```css
@import url('https://cdn.jsdelivr.net/fontsource/css/aclonica@latest/index.css');
body { background: blue; }
```

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
@@ -3,6 +3,8 @@ package com.vaadin.flow.server.frontend;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.regex.MatchResult;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -112,6 +114,19 @@ public class CssBundler {
             }
             return Matcher.quoteReplacement(result.group());
         });
+
+        // Move unhandled @import statements to the top, as they would be
+        // ignored by the browser if they appear after regular CSS rules
+        Matcher remainingImportMatcher = importPattern.matcher(content);
+        List<String> remainingImports = new ArrayList<>();
+        while (remainingImportMatcher.find()) {
+            remainingImports.add(remainingImportMatcher.group());
+        }
+        content = remainingImportMatcher.replaceAll("");
+        String remainingImportsString = String.join("\n", remainingImports);
+        boolean addNewLine = !content.isEmpty()
+                && !remainingImportsString.isEmpty();
+        content = remainingImportsString + (addNewLine ? "\n" : "") + content;
 
         return content;
     }
--- a/flow-server/src/test/java/com/vaadin/flow/server/frontend/CssBundlerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/frontend/CssBundlerTest.java
@@ -135,6 +135,30 @@ public class CssBundlerTest {
                 .inlineImports(themeFolder, getThemeFile("styles.css")));
     }
 
+    @Test
+    public void unhandledImportsAreMovedToTop() throws IOException {
+        writeCss("body {background: blue};", "other.css");
+        writeCss(
+                """
+                        @import url('https://cdn.jsdelivr.net/fontsource/css/inter@latest/index.css');
+                        @import url('other.css');
+                        @import url('https://cdn.jsdelivr.net/fontsource/css/aclonica@latest/index.css');
+                        """,
+                "styles.css");
+
+        Assert.assertEquals(
+                """
+                        @import url('https://cdn.jsdelivr.net/fontsource/css/inter@latest/index.css');
+                        @import url('https://cdn.jsdelivr.net/fontsource/css/aclonica@latest/index.css');
+
+                        body {background: blue};
+                        """
+                        .trim(),
+                CssBundler
+                        .inlineImports(themeFolder, getThemeFile("styles.css"))
+                        .trim());
+    }
+
     private boolean createThemeFile(String filename) throws IOException {
         File f = getThemeFile(filename);
         f.getParentFile().mkdirs();


flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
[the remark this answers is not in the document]
mcollovati: According to the documentation, stylesheets from npm Packages should be loaded using with the `documentCss` key in `theme.json`.
Currently, `CSSBundler` resolves and inlines only CSS files in the source folder.

So the test module should be fixed by removing the line-awesome CSS import from `styles.css` and adding the `importCss` key in theme.json
```
  "importCss": [
    "line-awesome/dist/line-awesome/css/line-awesome.min.css"
  ],
```

If we want to allow to asset CSS imports directly in `styles.css` (for example, to have imports in a specific order) we need to modify `CSSBundler` to rewrite relatives imports appending `VAADIN/themes`, so that `StaticFileServer` can handle it. 
But this should probably deserve a separate ticket.

```java
            if (potentialFile.exists()) {
                 ....
            } else if (url.startsWith("./")) {
                // potential reference to theme asset file
                String relativePath = themeFolder.getParentFile().toPath()
                        .relativize(potentialFile.toPath()).toString()
                        .replaceAll("\\\\", "/");
                return Matcher.quoteReplacement(
                        "url('VAADIN/themes/" + relativePath + "')");
            }
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
```diff
@@ -113,6 +115,19 @@ public static String inlineImports(File themeFolder, File cssFile)
...
+        List<String> remainingImports = new ArrayList<>();
+        while (remainingImportMatcher.find()) {
+            remainingImports.add(remainingImportMatcher.group());
+        }
+        content = remainingImportMatcher.replaceAll("");
+        String remainingImportsString = String.join("\n", remainingImports);
+        boolean addNewLine = !content.isEmpty()
+                && !remainingImportsString.isEmpty();
+        content = remainingImportsString + (addNewLine ? "\n" : "") + content;
+
```
mcollovati, line 130: The change looks good, but I wonder if we can spare a regex match by filling `remainingImports` and replacing imports with blank in the previous `replaceAll()` block.
For example,
```
        List<String> remainingImports = new ArrayList<>();
        Matcher importMatcher = importPattern.matcher(content);
        content = importMatcher.replaceAll(result -> {
           .....
            String url = getNonNullGroup(result, 3, 4, 5, 7, 8);
            if (url == null || !url.trim().endsWith(".css")) {
                // return Matcher.quoteReplacement(result.group());
                remainingImports.add(result.group());
                return "";
            }
            ....
     }
```

Then the above block should only have to concat strings
```java

        if (!remainingImports.isEmpty()) {
            content = String.join("\n", notInlined)
                    + (content.isEmpty() ? "" : "\n" + content);
        }
```

mcollovati (changes requested):
@sissbruecker do you mind fixing the test module?

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
sissbruecker: Thanks, updated the logic. I actually didn't include the part you quoted, it tests for invalid URLs...I guess? Added it for the other two cases though (layer rule or media query, not a local file), and extended the test case to cover those.

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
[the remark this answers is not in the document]
sissbruecker: Thanks, it seems to work now.

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
mcollovati: Looks like I chose the worst example :smile: 

Anyway, sounds good to ignore that case for now.

mcollovati:
Doing some additional tests, I found that there is another use case that we might consider: a CSS URL with query string.
For example, if I add `@import url('https://cdn.jsdelivr.net/fontsource/css/aclonica@latest/index.css?someCacheKillerParam');` to the `unhandledImportsAreMovedToTop` test, after processing it is put at the bottom.

sissbruecker:
> Doing some additional tests, I found that there is another use case that we might consider: a CSS URL with query string.
For example, if I add @import url('https://cdn.jsdelivr.net/fontsource/css/aclonica@latest/index.css?someCacheKillerParam'); to the unhandledImportsAreMovedToTop test, after processing it is put at the bottom.

@mcollovati  Changed it so that it handles query strings. Now it basically moves all unhandled imports to the top.

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
```diff
@@ -126,6 +138,13 @@ private static String getNonNullGroup(MatchResult result, int... groupId) {
         return null;
     }
 
+    private static String sanitizeUrl(String url) {
+        if (url == null) {
+            return null;
+        }
+        return url.trim().split("\\?")[0];
```
mcollovati, line 145: nit: looking at hierarchical URI composition (`[scheme:][//authority][path][?query][#fragment]`), we should probably take care also of `#fragment` to be safe.

mcollovati (approved):
LGTM, but please look at the comment on `sanitizeUrl()` method and see if it makes sense to you.

There might still be some corner cases where a stylesheet that cannot be inlined is supposed to overwrite rules from a local theme stylesheet, so changing the order might break styling.
But let's merge this fix, and then we will see whether the above is a real use case or not.

flow-server/src/main/java/com/vaadin/flow/server/frontend/CssBundler.java
mcollovati: @sissbruecker what do you think? Should we take care of the fragment part, or do you think handling the query string is enough?

mcollovati (approved):

PR #19025, opened 2024-03-24
fix: pass deployment mode to FS Router vite plugin

taefi:
## Description

This passes the deployment mode to the FS Router vite plugin so that it can decide where to generate needed file at runtime. This is to prevent file addition/modification under `classes` dir in dev mode to prevent spring-boot devtools
cause a restart during the startup. 

Related to: vaadin/hilla#2234


mshabarov:
This needs also a correction in test `TaskUpdateViteTest::generatedTemplate_reactAndHillaUsed_correctFileRouterImport` to expect the parameter.

This also should be merged after https://github.com/vaadin/hilla/pull/2234 I suppose.

tltv (approved):

PR #19038, opened 2024-03-26
feat: Add method for checking if annotation based security is used

Artur-:


flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessControl.java
```diff
@@ -401,4 +401,15 @@ private static Logger getLogger() {
...
     }
 
+    /**
+     * Checks if an {@link AnnotatedViewAccessChecker} is in use, i.e. route
+     * annotations are read.
+     *
+     * @return {@code true} if an {@link AnnotatedViewAccessChecker} is in use,
+     *         {@code false} otherwise
+     */
+    public boolean isUsingAnnotatedViewAccessChecker() {
```
mcollovati, line 411: Could this be a more generic method like `isUsingAccessChecker(Class<? extends NavigationAccessChecker> clazz)`?

I wonder if there might be a use case for checking for `RoutePathAccessChecker`, for example.
If possible (and it is reasonable) I would prevent having this class filled with `isUsingSometing()` methods.

If it does not make sense, I'm ok with this change.

Artur-: Made it a more generic

flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessControl.java
```diff
@@ -401,4 +401,14 @@ private static Logger getLogger() {
...
 
+    /**
+     * Checks if an access checker of the given type is in use.
+     *
+     * @return {@code true} if a checker is in use, {@code false} otherwise
+     */
+    public boolean hasAccessChecker(
+            Class<? extends NavigationAccessChecker> type) {
+        return checkerList.stream()
+                .anyMatch(checker -> checker.getClass() == type);
```
mcollovati, line 412: The strict class equality check may fail in some setups. For example, if the checker instance is a Spring bean proxy.
Another example is Quarkus, that in development mode always creates proxies for beans.
Could it be an `instanceof` check instead? Or is it too relaxed?

flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessControl.java
mcollovati: > Then the question is, should it be ”is” or ”get”

`get` because there may be a use case where the checker instance could be used outside the NavigationAccessControl?

flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessControl.java
Artur-: Sure, changed

mcollovati (approved):

PR #19047, opened 2024-03-26
fix: do not deactivate expanded item keys (#19006)

### LINKED ISSUE #19006: Newly generated keys no longer match expandedItems after the old keys are removed from KeyMapper
### Description of the bug

As @tomivirkki discovered in https://github.com/vaadin/web-components/issues/7097#issuecomment-1943321098, Flow HierarchicalCommunicationController can [passivate](https://github.com/vaadin/flow/blob/a130d64a9077cdedcbe41fa3eba4d781a0146e35/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java#L142) some of the item keys when the items are expanded recursively and request KeyMapper to [remove](https://github.com/vaadin/flow/blob/a130d64a9077cdedcbe41fa3eba4d781a0146e35/flow-data/src/main/java/com/vaadin/flow/data/provider/KeyMapper.java#L131) the keys for the passivated items. Since TreeGrid no longer uses `DataProvider.getId()` for the item identity after https://github.com/vaadin/flow-components/pull/3251, the items get fresh new keys that no longer match the old ones in the `expandedItems` array.

### Expected behavior

Not sure if it would be better to revert the https://github.com/vaadin/flow-components/pull/3251 in TreeGrid (to have it use DataProvider.getId() for the item identity again) or fix the passivating logic in HierarchicalCommunicationController.

```java
@Route("repro")
public class Repro extends Div {

    public Repro() {
        var grid = new TreeGrid<String>();
        grid.setPageSize(10);
        grid.addHierarchyColumn(String::toString).setHeader("Item");

        var data = new TreeData<String>();
        data.addItems(null, "Granddad");
        for (int i = 0; i < 20; i++) {
            data.addItems("Granddad", "Dad " + i);
            for (int j = 0; j < 20; j++) {
                data.addItems("Dad " + i, "Son " + i + "/" + j);
            }
        }

        grid.setDataProvider(new TreeDataProvider<>(data));

        add(grid);

        add(new NativeButton("Expand all",
            e -> grid.expandRecursively(data.getRootItems(), 1)));

        add(new NativeButton("Print \"Dad 0\" key",
            e -> System.out.println("Dad 0 key: " + grid.getDataCommunicator().getKeyMapper().key("Dad 0"))));
    }
}
```

...and the stack trace after clicking "Expand all":

```
at com.vaadin.flow.data.provider.KeyMapper.remove(KeyMapper.java:140)
at com.vaadin.flow.data.provider.hierarchy.HierarchicalCommunicationController.lambda$doUnregister$4(HierarchicalCommunicationController.java:324)
at java.base/java.lang.Iterable.forEach(Iterable.java:75)
at com.vaadin.flow.data.provider.hierarchy.HierarchicalCommunicationController.doUnregister(HierarchicalCommunicationController.java:320)
at java.base/java.lang.Iterable.forEach(Iterable.java:75)
at com.vaadin.flow.data.provider.hierarchy.HierarchicalCommunicationController.unregisterPassivatedKeys(HierarchicalCommunicationController.java:312)
at com.vaadin.flow.data.provider.hierarchy.HierarchicalCommunicationController.flush(HierarchicalCommunicationController.java:145)
at com.vaadin.flow.data.provider.hierarchy.HierarchicalDataCommunicator.lambda$requestFlush$e15592a2$1(HierarchicalDataCommunicator.java:118)
```

This is followed by an additional request from Grid due to HierarchicalCommunicationController clearing one of the parent items:

```
at com.vaadin.flow.component.treegrid.TreeGrid$TreeGridUpdateQueue.clear(TreeGrid.java:125)
at com.vaadin.flow.data.provider.hierarchy.HierarchicalCommunicationController.clear(HierarchicalCommunicationController.java:219)
at com.vaadin.flow.data.provider.hierarchy.HierarchicalCommunicationController.collectChangesToSend(HierarchicalCommunicationController.java:173)
at com.vaadin.flow.data.provider.hierarchy.HierarchicalCommunicationController.flush(HierarchicalCommunicationController.java:135)
at com.vaadin.flow.data.provider.hierarchy.HierarchicalDataCommunicator.lambda$requestFlush$e15592a2$1(HierarchicalDataCommunicator.java:118)
```

### Versions

- Vaadin / Flow version: 24.4 and earlier
- Java version: 17
- OS version: Mac OS

tepi:
Fixes #19006 

[PR for test](https://github.com/vaadin/flow-components/pull/6164) - should be reviewed together

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
@@ -79,6 +79,8 @@ public class HierarchicalCommunicationController<T> implements Serializable {
     // Update ids that have been confirmed since the last flush
     private final HashSet<Integer> confirmedUpdates = new HashSet<>();
 
+    private boolean hasUniqueKeyProviderSupplier;
+
     /**
      * Constructs communication controller with support for hierarchical data
      * structure.
@@ -319,7 +321,8 @@ public class HierarchicalCommunicationController<T> implements Serializable {
         if (passivated != null) {
             passivated.forEach(key -> {
                 T item = keyMapper.get(key);
-                if (item != null && !mapper.isExpanded(item)) {
+                if (item != null && (hasUniqueKeyProviderSupplier
+                        || !mapper.isExpanded(item))) {
                     dataGenerator.destroyData(item);
                     keyMapper.remove(item);
                 }
@@ -339,6 +342,11 @@ public class HierarchicalCommunicationController<T> implements Serializable {
         json.put("key", keyMapper.key(item));
         dataGenerator.generateData(item, json);
         return json;
+    }
+
+    public void setHasUniqueKeyProviderSupplier(
+            boolean hasUniqueKeyProviderSupplier) {
+        this.hasUniqueKeyProviderSupplier = hasUniqueKeyProviderSupplier;
     }
 
     private static final void withMissing(Range expected, Range actual,
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
@@ -319,7 +319,7 @@ public class HierarchicalCommunicationController<T> implements Serializable {
         if (passivated != null) {
             passivated.forEach(key -> {
                 T item = keyMapper.get(key);
-                if (item != null) {
+                if (item != null && !mapper.isExpanded(item)) {
                     dataGenerator.destroyData(item);
                     keyMapper.remove(item);
                 }
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
@@ -196,7 +196,8 @@ public class HierarchicalDataCommunicator<T> extends DataCommunicator<T> {
                                         .startUpdate(getDataProviderSize()),
                                 (pkey, range) -> mapper.fetchChildItems(
                                         getKeyMapper().get(pkey), range)));
-
+        controller.setHasUniqueKeyProviderSupplier(
+                uniqueKeyProviderSupplier.get() != null);
         Range range = computeRequestedRange(start, length);
         controller.setRequestRange(range.getStart(), range.length());
         requestFlush(controller);


flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
```diff
@@ -319,7 +319,7 @@ private void doUnregister(Integer updateId) {
         if (passivated != null) {
             passivated.forEach(key -> {
                 T item = keyMapper.get(key);
-                if (item != null) {
+                if (item != null && !mapper.isExpanded(item)) {
```
tltv, line 322: Wouldn't this keep passivated item data also when using unique key provider which doesn't have the issue as key is same then?
Suggestion: to make this more conditional, how about passing `uniqueKeyProviderSupplier` from `HierarchicalDataCommunicator` to `HierarchicalCommunicationController` and checking if `uniqueKeyProviderSupplier` is used. If it is, then it's ok to clean item data. Otherwise not.

Final version would look something like this then:
`if (item != null && (uniqueKeyProviderSupplier.get() != null || !mapper.isExpanded(item))) {`

tepi: True, this would make sense, I'll change it. Thanks!

tepi: Seems `HierarchicalCommunicationController` and its constructor is public API (although really mostly internal I think), so probably can't add it as a param there.

tltv: Other alternative is to add `boolean isPermanentUniqueKey()` in `KeyMapper` with default impl returning false. then implement it in `HierarchicalDataCommunicator#KeyMapperWrapper` to return `uniqueKeyProviderSupplier.get() != null`.

tepi: Did slightly different solution, did not want to mess with `DataKeyMapper` interface

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
```diff
@@ -341,6 +344,11 @@ public JsonValue generateJson(T item) {
         return json;
     }
 
+    public void setHasUniqueKeyProviderSupplier(
```
tltv, line 347: So many public methods without Javadoc that it makes me wonder if this class should be documented as internal API.

tltv:
There's no unit tests in Flow, but there's IT in flow-components which looks good to me. Have you run this locally with flow-components tests to see if it's working there?

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicationController.java
tepi: My thoughts exactly. Most likely it should be. Not sure if relevant to this issue though, but I'm pretty sure no-one is using this as public API

tepi:
> There's no unit tests in Flow, but there's IT in flow-components which looks good to me. Have you run this locally with flow-components tests to see if it's working there?

Building a unit test in Flow proved to be quite cumbersome - I couldn't figure out how to reproduce all the steps TreeGrid is doing to surface the issue. The IT in flow-components is sufficient, it turns up the issue with changing IDs when this patch is not applied, and passes fine once the patch is applied.

tltv (approved):

PR #19050, opened 2024-03-27
fix: show available client routes

tltv:
Adds client views in the list of available routes when viewing "Route not found" page in development mode. Adds ClientRoutesProvider interface to get available client routes. Client routes means mainly Hilla views, but not limited to.

Fixes: #18857

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
@@ -42,6 +42,7 @@ import com.vaadin.flow.di.Lookup;
 import com.vaadin.flow.router.internal.ClientRoutesProvider;
 import com.vaadin.flow.server.HttpStatusCode;
 import com.vaadin.flow.server.VaadinService;
+import com.vaadin.flow.server.frontend.FrontendUtils;
 
 /**
  * This is abstract error view for routing exceptions.
@@ -141,13 +142,8 @@ public abstract class AbstractRouteNotFoundError extends Component {
     }
 
     private List<Element> getClientRoutes() {
-        return Optional.ofNullable(VaadinService.getCurrent())
-                .map(VaadinService::getContext)
-                .map(ctx -> ctx.getAttribute(Lookup.class)
-                        .lookup(ClientRoutesProvider.class))
-                .stream()
-                .flatMap(provider -> provider.getClientRoutes().stream())
-                .filter(Objects::nonNull).map(this::clientRouteToHtml).toList();
+        return FrontendUtils.getClientRoutes().stream()
+                .map(this::clientRouteToHtml).toList();
     }
 
     private Element routeTemplateToHtml(String routeTemplate,
--- a/flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
@@ -143,7 +143,8 @@ public abstract class AbstractRouteNotFoundError extends Component {
 
     private List<Element> getClientRoutes() {
         return FrontendUtils.getClientRoutes().stream()
-                .filter(route -> !route.startsWith("$"))
+                .filter(route -> !route.contains("$layout"))
+                .map(route -> route.replace("$index", ""))
                 .map(this::clientRouteToHtml).toList();
     }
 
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
```diff
@@ -79,7 +82,7 @@ public int setRouteNotFoundErrorParameter(BeforeEnterEvent event,
         boolean productionMode = event.getUI().getSession().getConfiguration()
                 .isProductionMode();
         String template;
-        String routes = getRoutes(event);
+        String routes = getServerRoutes(event);
```
Artur-, line 85: This should not use only server routes in the end. The logic below requires routes not to be empty if there are server or client views

tltv: Right, looks like I forgot to rename getServerRoutes back to getRoutes.

caalador:
Does and should `ClientRoutesProvider` also handle user defined routes setup from routes.tsx?
Is this feature only for file router setups?

tltv:
> Does and should `ClientRoutesProvider` also handle user defined routes setup from routes.tsx? Is this feature only for file router setups?

It's meant for both. So far I've tested it with Hilla by letting `ClientRouteRegistry` to implement `ClientRouteProvider` and it returns both.

I noticed that $index.tsx and $layout.tsx are still listed. $layout.tsx should be always hidden as it's not a real path afaik, but $index.tsx path should show up to cover up cases when it's in other places than root.

tltv:
There was discussion in Flow daily meeting about the Spring security to be taken in account with this, but as `ClientRoutesProvider` returns a full list of available routes to build a link list for development mode, no access checking is needed in this use case. For other use case that needs security checking, we can add another method in the interface for a list of accessible paths if needed.

flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
```diff
@@ -127,14 +133,21 @@ private String getRoutes(BeforeEnterEvent event) {
...
 
+        routeElements.addAll(getClientRoutes());
         return routeElements.stream().map(Element::outerHtml)
                 .collect(Collectors.joining());
     }
 
+    private List<Element> getClientRoutes() {
+        return FrontendUtils.getClientRoutes().stream()
+                .filter(route -> !route.contains("$layout"))
+                .map(route -> route.replace("$index", ""))
```
caalador, line 147: The test reads paths as "/" or "/hilla/hilla", but looking at the code the expectation would be "" and "/hilla/hilla/" where is the string updated?

flow-server/src/main/java/com/vaadin/flow/router/internal/ClientRoutesProvider.java
```diff
@@ -0,0 +1,35 @@
...
+public interface ClientRoutesProvider extends Serializable {
+
+    /**
+     * Get a list of client side routes.
+     *
+     * @return a list of client side routes
+     */
+    default List<String> getClientRoutes() {
+        return List.of();
+    }
```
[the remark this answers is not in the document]
tltv: Yeah, no need for the default impl anymore. It was there for a reason first.

flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -1434,4 +1435,20 @@ public static boolean isReactModuleAvailable(Options options) {
...
+     * collected from all {@link ClientRoutesProvider} implementations found
+     * with Vaadin {@link Lookup}.
+     *
+     * @return a list of available client routes
+     */
+    public static List<String> getClientRoutes() {
+        return Optional.ofNullable(VaadinService.getCurrent())
+                .map(VaadinService::getContext).stream()
+                .flatMap(ctx -> ctx.getAttribute(Lookup.class)
+                        .lookupAll(ClientRoutesProvider.class).stream())
```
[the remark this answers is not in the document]
tltv: I don't know. Only use case for supporting multiple is when FS routing and manual routes need a different beans. But if `ClientRouteRegistry` is enough in Hilla side as it seems to be, then there's no real use for support for multiple.

flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
caalador: Missed that. That makes testing for the root route `"/"` not that functional as `/` is in every route and only that is not checked. Should the routes have a starting and ending space so that we would require a route that is a more exact match?

tltv: I noticed that too. I'll refactor the test better and also there is indeed extra / in the expected list.

caalador (approved):

PR #19059, opened 2024-03-27
refactor: import file-routes.js instead of views.js in Flow.tsx

taefi:
Follow up for vaadin/hilla#2271

NOTE: the vaadin/hilla#2271 also keeps the generation of views.ts temporarily to prevent possible snapshot build failures.
Please feel free to add any changes needed for a smooth transition, if needed.
After these two PRs (on both sides) merged successfully, the generation of views.ts would be removed from Hilla codebase.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
@@ -223,7 +223,7 @@ public class TaskGenerateReactFiles implements FallibleCommand {
             return content.replace("//%toReactRouterImport%",
                     "import { toReactRouter } from '@vaadin/hilla-file-router/runtime.js';")
                     .replace("//%viewsJsImport%",
-                            "import views from 'Frontend/generated/views.js';")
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -223,7 +223,7 @@ private String getFlowTsxFileContent(boolean frontendRoutesTsExists)
             return content.replace("//%toReactRouterImport%",
                     "import { toReactRouter } from '@vaadin/hilla-file-router/runtime.js';")
                     .replace("//%viewsJsImport%",
-                            "import views from 'Frontend/generated/views.js';")
+                            "import views from 'Frontend/generated/file-routes.js';")
```
mshabarov, line 226: This wouldn't be needed after we switch to `RouterBuilder` API, because `views.js` won't be used by Flow anymore.

taefi: OK, so we can close this PR for good.

PR #19063, opened 2024-03-28
fix: more robust Server route check

caalador:
Make server route validation
not use path but element instead.

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -311,7 +311,7 @@ export default function Flow() {
             // If we are going to a non Flow view then we need to clean the Flow
             // view from the dom as we will not be getting a uidl response.
             // @ts-ignore
-            if(matched && matched.filter(path => path.route?.element?.type?.name === Flow.name).length >= 1) {
+            if(matched && matched.filter(path => path.route?.element?.type?.name === Flow.name).length == 0) {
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -309,7 +310,8 @@ export default function Flow() {
             // flow from the view
             // If we are going to a non Flow view then we need to clean the Flow
             // view from the dom as we will not be getting a uidl response.
-            if(matched && matched[matched.length - 1].route.path !== "/*") {
+            // @ts-ignore
+            if(matched && matched.filter(path => path.route?.element?.type?.name === Flow.name).length >= 1) {
```
tltv, line 314:
```suggestion
            if(matched && matched.filter(path => path.route?.element?.type?.name === Flow.name).length == 0) {
```

tltv (changes requested):

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
mshabarov: Right, this should pass, if it's not a Flow route.

tltv: As this can happen so easily, maybe it would make sense to just add a test for this case. One that checks that old content is removed.

tltv (approved):

mshabarov (approved):

PR #19087, opened 2024-04-03
fix: Query in path

### LINKED ISSUE #19080: In Vaadin 24.4.0.alpha18, RouterLink does not update the browser URL
### Description of the bug

In Vaadin 24.4.0.alpha17 everything is working fine, but in Vaadin 24.4.0.alpha18, clicking on RouterLinks does not update the browser URL.

For example, I stay on

http://localhost:8080/notifications?status=pending

Then, clicking on the RouterLink with the following target:

http://localhost:8080/?lang=en

After that, I successfully moved to http://localhost:8080/?lang=en, but in the browser address bar, I still see http://localhost:8080/notifications?status=pending

In Vaadin 24.4.0.alpha17, everything is working fine

### Expected behavior

browser url in address bar should change to the correct one

n/a

### Versions

- Vaadin / Flow version: Vaadin 24.4.0.alpha18
- Java version: 17

caalador:
Fix mistaken add of query
parameters into the route
breaking route resolution
when navigating thrhough link.

fixes #19080

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -131,12 +131,9 @@ function vaadinRouterGlobalClickHandler(event) {
     }
 
     // if none of the above, convert the click into a navigation event
-    const {pathname, href, baseURI, search, hash} = anchor;
+    const {href, baseURI, search, hash} = anchor;
     // Normalize away base from pathname. e.g. /react should remove base /view from /view/react
-    let normalizedPathname = href.slice(0, baseURI.length) == baseURI ? href.slice(baseURI.length) : pathname;
-    if(normalizedPathname.indexOf('?') !== -1) {
-        normalizedPathname = normalizedPathname.slice(0, normalizedPathname.indexOf('?'));
-    }
+    let normalizedPathname = href.replace(baseURI, '').replace(search,'').replace(hash, '');
     normalizedPathname = normalizedPathname.startsWith("/") ? normalizedPathname : "/" + normalizedPathname;
     if (fireRouterEvent('go', {pathname: normalizedPathname, search, hash, clientNavigation: true})) {
         event.preventDefault();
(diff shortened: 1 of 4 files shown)


tltv (commented):
Makes sense. Create also new IT for the main issue (can be in separated PR).
Shouldn't this also remove hash from the baseURI?

caalador:
Actually this might be a problem in the server side route resolver instead of a problem getting the query. I can only see a couple tests that test the href for routerlink, but nothing that would test actually navigating with the query and that is what fails now.

caalador:
Now it should be ok and there are tests in place for the navigation.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -131,10 +131,10 @@ function vaadinRouterGlobalClickHandler(event) {
     }
 
     // if none of the above, convert the click into a navigation event
-    const {pathname, href, baseURI, search, hash} = anchor;
+    const {href, baseURI, search, hash} = anchor;
     // Normalize away base from pathname. e.g. /react should remove base /view from /view/react
-    let normalizedPathname = href.slice(0, baseURI.length) == baseURI ? href.slice(baseURI.length) : pathname;
-    normalizedPathname = normalizedPathname.startsWith("/") ? normalizedPathname: "/" + normalizedPathname;
+    let normalizedPathname = href.replace(baseURI, '').replace(search,'').replace(hash, '');
```
tltv, line 136: `search` could in some case contain baseURI (e.g. `https://vaadin.com/view?param=https://vaadin.com`), which could leave invalid query part as a result (`/view?param=`). 
Calling `replace(search,'')` before `replace(baseURI, '')` will fix that.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -158,9 +158,10 @@ function navigateEventHandler(event) {
         event.preventDefault();
     }
     // Normalize path against baseURI if href available.
-    let normalizedPathname = event.detail.href && event.detail.href.slice(0, document.baseURI.length) == document.baseURI ?
-        event.detail.href.slice(document.baseURI.length) : event.detail.pathname;
-    normalizedPathname = normalizedPathname.startsWith("/") ? normalizedPathname: "/"+normalizedPathname;
+    let normalizedPathname = event.detail.href ?
+        event.detail.href.replace(document.baseURI, '').replace(event.detail.search ,'').replace(event.detail.hash, '') :
```
tltv, line 162: Same here.

tltv (changes requested):

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
caalador: This shouldn't be an issue as per the documentation `If pattern is a string, only the first occurrence will be replaced` so to make it replace all then it would need to be a regex with `/ig` ending.
But switched them anyway

tltv (approved):

PR #19088, opened 2024-04-03
chore: upgrade testbench to 9.2.5

ZheSun88:


mcollovati:
Closing in favor of #19090 that also has the fix for the compilation error

PR #19092, opened 2024-04-03
fix: normalize and replace line separator in node task lock file

### LINKED ISSUE #19091: `.flow-node-tasks.lock` can contain more than the two expected lines due to CLI args with line-breaks
### Description of the bug

If the command-line to start a dev Flow application contains literal
line-breaks, these line-breaks are written to `.flow-node-tasks.lock`;
when in the lock handling the file is read again later, the error `Error
releasing lock file...` is show, which hides the error `Invalid lock
file. It should contain 2 rows but contains...`: the lock file is read
into an array of strings and the size is checked to be 2, but contains
more lines because the literal line-breaks are written to that file.

This all sounds like an odd problem to have, but it becomes quite
regular when using Gradle (maybe due to the use of the
Springboot-Plugin), since it adds `-Dline.separator=<NEWLINECHAR>` to
the arguments to the task.

### Expected behavior

The locking must anticipate the command line to contain line-breaks (or any "separator"): quote the separator or compare the lock content with what would be written

Use any fresh V24.3 Gradle project (yet, this sould be reproduceable agnostic of the build-tool)
```console
./gradlew :bootRun -Dintro="Hello
World"
```
The applications logs: `... c.vaadin.flow.server.frontend.NodeTasks  : Error releasing lock file ...`

Then check the log:

```console
% cat .flow-node-tasks.lock
446462
$HOME/.sdkman/candidates/java/21.0.2-zulu/bin/java -Dfile.separator=/ -Dintro=Hello
World -Djava.class.path=$HOME/.gradle/wrapper/dists/gradle-8.6-bin/afr5mpiioh2wthjmwnkmdsd5w/gradle-8.6/lib/gradle-launcher-8.6.jar ... -Dline.separator=
 -Dnative.encoding=UTF-8 -Dorg.gradle.appname=gradlew -Dorg.gradle.daemon=true -Dos.arch=amd64 ...
% wc -l .flow-node-tasks.lock 
4 .flow-node-tasks.lock
```

(`...` and `$HOME` replace lengthy unrelated output and are not part of
the original file)

### Versions

- Vaadin / Flow version: 24.3
- Java version: OpenJDK 64-Bit Server VM Zulu21.32+17-CA (build 21.0.2+13-LTS, mixed mode, sharing)
- OS version: Debian/SID Linux 6.6 amd64
- IDE (if applicable): N/A

mcollovati:
Normalizes and replaces line separators potentially present in the process command line before writing the lock file, so that it will contain only two lines. Same normalization is applied when getting the command line for the process identified by the pid in the lock file before comparing it with the value stored in the lock file.

Fixes #19091

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -363,7 +363,7 @@ public class NodeTasks implements FallibleCommand {
                         .of(lockInfo.pid());
 
                 if (processHandle.isPresent()
-                        && processHandle.get().info().commandLine().orElse("")
+                        && normalizeCommandLine(processHandle.get().info())
                                 .equals(lockInfo.commandLine())) {
                     if (!loggedWaiting) {
                         getLogger().info("Waiting for a previous instance of "
@@ -442,9 +442,14 @@ public class NodeTasks implements FallibleCommand {
     private void writeLockFile() throws IOException {
         ProcessHandle currentProcess = ProcessHandle.current();
         long myPid = currentProcess.pid();
-        String commandLine = currentProcess.info().commandLine().orElse("");
+        String commandLine = normalizeCommandLine(currentProcess.info());
         List<String> lines = List.of(Long.toString(myPid), commandLine);
         Files.write(lockFile, lines, StandardCharsets.UTF_8);
+    }
+
+    private String normalizeCommandLine(ProcessHandle.Info processInfo) {
+        return processInfo.commandLine()
+                .map(line -> line.replaceAll("\\r?\\n", " \\\\n")).orElse("");
     }
 
     private Logger getLogger() {


mcollovati:
Manually tested this way:

Without the patch, run maven with a parameter that spans over multiple rows, e.g. 
 ```
mvn -Dxxxx="sdfdsf
dfgdfgdf
dfgdfgdfg
dfgdfg
" process-resources
 ```
In the console log, you should see the error `Error releasing lock file ....` and the lock file is not deleted.

With the patch, the error is not shown.

Additional test:

1. Put a break point in `NodeTasks.writeLockFile` on the `Files.write()` instruction
2. Run the above maven command in debug mode (`mvnDebug`)
3. When the debugger reaches the breakpoint, proceed until the lock file is written and then wait
4. On another terminal, run the same maven command, without debug mode. On the console, you should see something like `Waiting for a previous instance of NodeTasks (pid: 84845) to finish...`
5. Detach the debugger so that the command on the first terminal can complete
6. Check that also the command launched on the other terminal completes without errors and the lock file is deleted

mcollovati:
So far, I have no good idea on how to implement a proper test for this change

caalador (approved):

caalador:
I don't see a good way to test this either...

Issue #19112, opened 2024-04-05
Speedup startup and reload time of 24.4


tltv:
### Describe your motivation

Spring application start up time has increased with 24.4 compared to 24.3 when using Hilla and Copilot. It's increased even when not actually using Hilla. Just having dependencies in the classpath is enough to slow down startup time and also reload time.

### Describe the solution you'd like

Speed up startup time and reload time of Vaadin 24.4 Spring application. More information on how to do that in https://github.com/vaadin/flow/issues/18812.

Proposed solution is to add following in the `VaadinServletContextInitializer#CustomResourceLoader` to filter jars that are not needed to be read file-by-file by the class scanner:
```
        @Override
        protected Set<Resource> doFindPathMatchingJarResources(Resource rootDirResource, URL rootDirUrl, String subPattern) throws IOException {
            String path = rootDirResource.getURI().toString();
            if(DEFAULT_SCAN_NEVER_JAR.stream().anyMatch(path::contains)) {
                return Set.of();
            }
            return super.doFindPathMatchingJarResources(rootDirResource, rootDirUrl, subPattern);
        }

        @Override
        protected Set<Resource> doFindAllClassPathResources(String path) throws IOException {
            var result = super.doFindAllClassPathResources(path);
            result.removeIf(res -> {
                try {
                    return DEFAULT_SCAN_NEVER_JAR.stream()
                            .anyMatch(res.getURI().toString()::contains);
                } catch (IOException e) {
                    return false;
                }
            });
            return result;
        }
```

### Describe alternatives you've considered

Just excluding Hilla dependencies speeds up start time, but not to the same level as with 24.3.

### Additional context

https://github.com/vaadin/flow/issues/18812

PR #19113, opened 2024-04-05
fix: filter hilla and copilot packages better

tltv:
Filter "com/vaadin/hilla" and "com/vaadin/copilot" packages by default from the class scanner in `VaadinServletContextInitializer` and include only what is really needed: `com.vaadin.copilot.startup` and `com.vaadin.hilla.startup`. This speeds up mostly reload time and also startup time a bit.

Related-to: #19112

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
@@ -130,7 +130,9 @@ public class VaadinServletContextInitializer
             "com/vaadin/external/gwt", "javassist/", "io/methvin",
             "com/github/javaparser", "oshi/", "io/micrometer", "jakarta/",
             "com/nimbusds", "elemental/util", "elemental/json",
-            "org/reflections", "org/aopalliance", "org/objectweb")
+            "org/reflections", "org/aopalliance", "org/objectweb",
+
+            "com/vaadin/hilla", "com/vaadin/copilot")
             .collect(Collectors.toList());
 
     /**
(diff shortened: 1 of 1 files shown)


caalador (approved):
The only thing here is how do we bring up the issue that if copilot adds some other scan required parts?
Hilla would seem to use the startup package which makes it a non issue, but copilot does not. Should we ask copilot to also use a dedicated pacakge for startup and scanning?

tltv:
> The only thing here is how do we bring up the issue that if copilot adds some other scan required parts? Hilla would seem to use the startup package which makes it a non issue, but copilot does not. Should we ask copilot to also use a dedicated pacakge for startup and scanning?

https://github.com/vaadin/flow/issues/19117 has a solution for that. It would let copilot and hilla to define what packages or classes to include/exclude.

tltv:
Since we don't have https://github.com/vaadin/flow/issues/19117 done yet and no schedule for it, having a dedicated package like `com.vaadin.copilot.startup` for `CopilotIndexHtmlLoader` and `CopilotLoader` can make sense. No need to touch Flow then when adding new loaders in Copilot. Just add them in `com.vaadin.copilot.startup` and Flow scans them. Otherwise new loaders (or other classes that Flow should scan) would need to be added to Flow's `VaadinServletContextInitializer`.

But is it worth to refactor package names in Copilot? If there's no plans to add new classes that Flow should scan, then waiting that https://github.com/vaadin/flow/issues/19117 is done could make more sense. Pinging @MarcinVaadin for better insight regarding Copilot.

Artur-:
Makes sense to refactor copilot to use `com.vaadin.copilot.startup` rather than list classes here

tltv:
Added `com.vaadin.copilot.startup` already. Refactoring copilot can happen anytime and no need to get back to change Flow. But we need a ticket for that in Copilot.

Artur-:
I can fix copilot as part of the next Flow upgrade

Artur-:
You can remove the copilot classes also already

caalador (approved):

Issue #19062, opened 2024-03-28
Live reload declares java unavailable 


emaayan:
### Description of the bug

live reload doesn't declare java active 

### Expected behavior

Live Reload should identify java saying java active 

### Minimal reproducible example

configure the application for live reload 
using the following basic maven config 

1. download  and install the JBR 17.0.10
2. download the 1.4.2 version of Hotswap agent jar and place the jar in the directory <JBR-install-directory>/lib/hotswap 
3. configured my IDE  to use that JDK.
4. configure run configuration so it added the following JVM parameters: -XX:+AllowEnhancedClassRedefinition -XX:HotswapAgent=fatjar
5. start by executing mvn jetty:run 
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <groupId>org.example</groupId>
    <artifactId>VaadinApp</artifactId>
    <version>1.0-SNAPSHOT</version>
    <name>My Application</name>
    <packaging>war</packaging>

    <properties>
        <maven.compiler.source>17</maven.compiler.source>
        <maven.compiler.target>17</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <failOnMissingWebXml>false</failOnMissingWebXml>
        <vaadin.version>24.3.8</vaadin.version>
    </properties>

    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>com.vaadin</groupId>
                <artifactId>vaadin-bom</artifactId>
                <type>pom</type>
                <scope>import</scope>
                <version>${vaadin.version}</version>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <dependencies>
        <dependency>
            <groupId>com.vaadin</groupId>
            <!-- Replace artifactId with vaadin-core to use only free components -->
            <artifactId>vaadin</artifactId>
        </dependency>

        <!-- Added to provide logging output as Flow uses -->
        <!-- the unbound SLF4J no-operation (NOP) logger implementation -->
        <dependency>
            <groupId>org.slf4j</groupId>
            <artifactId>slf4j-simple</artifactId>
        </dependency>

        <dependency>
            <groupId>jakarta.servlet</groupId>
[...]

PR #19120, opened 2024-04-05
fix: show java livereload status with shared ws

tepi:
Fixes https://github.com/vaadin/flow/issues/19062


mshabarov (approved):
Status works for me as expected.

Issue #18940, opened 2024-03-11
Print a warning when an event is blocked by a modal dialog


mvysny:
### Describe your motivation

The recommended way to notify the server is to fire CustomEvent DOM events on the <body> element and then hook DOM listeners server-side to the UI.getCurrent().getElement(). However, those DOM listeners are not being called under specific conditions; if one does not understand those conditions then it's really hard to see what's going on.

For example, Viritin notifies current UI of the GPS location. However, if Viritin's Geolocation is called when a Modal dialog is being open, then the listeners are not being called, as can be seen in https://github.com/viritin/flow-viritin/issues/47 .

### Describe the solution you'd like

If a DOM event targeting current UI is caught and ignored by Vaadin server (e.g. because of modal dialog being opened), Vaadin should log a warning message clarifying the situation.

mvysny:
Related to #15614

mcollovati:
Flow should already print a message, but it is at TRACE level.
@mvysny could you try to set TRACE level for category `com.vaadin.flow.server.communication.rpc` and check if you can see something like

```
Ignored RPC for invocation handler 'XXXXXX' from the client side for an inert node id='YYYY'
```

If this could be the message you expect to see, we can then discuss the better level and if the information is enough or should be improved.

mvysny:
Steps to reproduce: just paste this example to skeleton starter
```java
@Route("")
public class MainView extends VerticalLayout {

    public MainView() {
        add(new Button("Open dialog", e -> {
            final Dialog dlg = new Dialog();
            dlg.setModal(true);
            dlg.open();
            dlg.add(new Button("fire event on UI", e2 -> fireEventWithTimeout(UI.getCurrent())));
            dlg.add(new Button("fire event on modal element", e2 -> fireEventWithTimeout(UI.getCurrent().getInternals().getActiveModalComponent())));
        }));
    }

    private static void fireEventWithTimeout(Component target) {
        target.getElement().addEventListener("foo", e -> System.out.println(target + ": DOM Event fired: " + e));
        target.getElement().executeJs("let self = this; setTimeout(() => self.dispatchEvent(new CustomEvent('foo')), 1000);");
    }
}
```

@mcollovati yup you're right, enabling trace level revealed the error log:
```
[...]

PR #19121, opened 2024-04-05
feat:allow explicitly reveiving certain dom events for inert elements

### LINKED ISSUE #19116: event listeners need to support bypassing modality checks
### Description of the bug

Events are a way more powerful mechanism to send data from client to the server. For example supporting debouncing etc. Sinse the server side modality changes, this is handicapped for certain scenarios, where events are not actual component events but somethign regarding browser/ui level. Like geolocation API, browser window size etc.

### Expected behavior

The API should support allowInert() or similar annotation to bypass the checks for selected events.

Check the hacks needed for Viritin's geolocation API to work when a dialog is open.

### Versions

- Vaadin / Flow version: 24.3

mstahv:
Added API to allow bypassing inert check for DOM events.

Fixes #19116

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist


mstahv:
There is bit ugly API changes, but luckily only for internal classes (marked properly with documentation). Some code smell, but don't have good ideas how to refactor as I don't understand all the details related to the implementation.

mcollovati:
A comment after a quick and not in-depth look: do we really need to change `handleNode()` and `fireEvent()` to add the `inert` flag?
I mean, the flag seems to be used only by `ElementListenerMap.fireEvent()` and I wonder if we could get the same information from the event as we do for `isEnabled`  (e.g. `event.getSource().getNode().isInert()`).
This way, we may probably be able to also remove the hard-coded `instanceof  EventRpcHandler` in `AbstractRpcInvocationHandler`; then `EventRpcHandler` should override `allowInert(...)` to return `true` (similar to `PublishedServerEventHandlerRpcHandler`) since the check will be performed later in `fireEvent`.

I apologize in advance if the above statements are incorrect :grin:

knoobie:
I would follow Marco's suggestion if it works to reduce the possibility of any breaking change 😅

mstahv:
The "breaking changes" are only in classes marked with "For internal use only. May be renamed or removed in a future release.", so I think those ought to be rather safe. 

I spent quite a while looking into a way to get the right details, but I sure might have missed some opportunity due to the complexity of the code. I have an hour or two in car so I can give it an another look now, but my my true wish is that somebody who actaully understands this code would step in :-) 

@mcollovati is there any kind of "developer docs" for this internals? Like some kind of brief overview or description of what these classes do or even just some class/sequence diagrams? Currently it is very hard to understand how the thing are supposed to work and why it is build in this way...

mstahv:
If we can assume that this piece in AbstractRpcInvocationHandler: 

```
StateNode node = ui.getInternals().getStateTree()
                .getNodeById(getNodeId(invocationJson));
```

... returns the same instance as `event.getSource().getNode()` in ElementListenerMap, then passing around the inert flag could be omitted as well. I have no idea, don't really even understand the purpose of StateNode 😬

mstahv:
Refactored, the current state isn't tested in any ways though now...

mstahv:
There is now a unit test IT will be there throught the resize event improvement. I think this is ready. Annotation support could be added later if needed, I don't need it.

mcollovati:
> If we can assume that this piece in AbstractRpcInvocationHandler: ...

The node get from UI in `AbstractRpcInvocationHandler` is the same used to create the`Element` object for the `DomEvent`, so `event.getSource().getNode()` returns the exactly same instance.

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/ElementListenerMap.java
```diff
@@ -114,6 +114,7 @@ private static class DomEventListenerWrapper
         private int debounceTimeout = 0;
         private EnumSet<DebouncePhase> debouncePhases = NO_TIMEOUT_PHASES;
         private List<SerializableRunnable> unregisterHandlers;
+        private boolean allowIntert;
```
mcollovati, line 117: Typo here: `allowIntert`

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/ElementListenerMap.java
```diff
@@ -436,6 +445,15 @@ public void fireEvent(DomEvent event) {
 
         List<DomEventListener> listeners = new ArrayList<>();
         for (DomEventListenerWrapper wrapper : typeListeners) {
+            if (inert && !wrapper.allowIntert) {
+                // drop as inert
+                LoggerFactory.getLogger(ElementListenerMap.class.getName())
+                        .info("Ignored listener invocation from "
+                                + "the client side for an inert {} element",
+                                event.getSource().getTag());
```
mcollovati, line 453: I wonder if we can provide some additional information in the message. For example, the event type.

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/BrowserWindowResizeView.java
```diff
@@ -15,7 +15,9 @@
  */
 package com.vaadin.flow.uitest.ui;
 
+import com.vaadin.flow.component.UI;
```
mcollovati, line 18: unused import
```suggestion
```

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/BrowserWindowResizeView.java
```diff
@@ -15,7 +15,9 @@
...
 import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
```
mcollovati, line 20: unused import
```suggestion
```

flow-server/src/test/java/com/vaadin/flow/dom/ElementTest.java
```diff
@@ -382,6 +383,33 @@ public void listenerReceivesEvents() {
         Assert.assertEquals(1, listenerCalls.get());
     }
 
+    @Test
+    public void listenerReceivesEventsWithAllowIntert() {
```
mcollovati, line 387: Typo
```suggestion
    public void listenerReceivesEventsWithAllowInert() {
```

mcollovati: I would add a similar test to `EventRpcHandlerTest` to prevent regression because of changes in the `allowInert()` override.

mcollovati (changes requested):

mcollovati:
It looks like this change is breaking navigation on modal components (`InertComponentIT.modalComponentAdded_routerLinkClicked_navigation`).
Probably `ElementListenerMap.fireEvent()` should allow navigation events (`BrowserNavigateEvent.EVENT_NAME` and `BrowserLeaveNavigationEvent.EVENT_NAME`) regardless of the inert node flag, as `AbstractRpcInvocationHandler.isNavigationInvocation()` does.

mcollovati:
@mstahv if you are busy, I can apply the changes and then continue the discussion

mstahv:
@mcollovati Go ahead if you have these in IDE already! Otherwise I can open this again at some point when I can't get text out of my brain 😁

mstahv:
Nah, if you have good ideas @mcollovati, go for it. My idea was to use the new API for those navigation events, but they are some "special cases" and it started to look complicated...

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -62,12 +64,12 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
         // ignore RPC requests from the client side for the nodes that are
         // invisible, disabled or inert
         if (node.isInactive()) {
-            getLogger().trace("Ignored RPC for invocation handler '{}' from "
+            getLogger().info("Ignored RPC for invocation handler '{}' from "
```
tepi, line 67: Would not go for `info` level here. Debug or trace should be fine.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -62,12 +64,12 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
                     + "the client side for an inactive (disabled or invisible) node id='{}'",
                     getClass().getName(), node.getId());
             return Optional.empty();
         } else if (!allowInert(ui, invocationJson) && node.isInert()) {
-            getLogger().trace(
+            getLogger().info(
```
tepi, line 72: Would not go for `info` level here. Debug or trace should be fine.

tepi (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
mstahv: I very much disagree. This is kind of "should not even happen" case (typically browser side/components should block it as well. If it does, it is most likely either somebody trying to do something evil or issue in our components (or in custom addons), and cant be very odd for the developer as the "exception" is eaten. There is a separate bug & PR for that part: https://github.com/vaadin/flow/issues/18940

flow-server/src/test/java/com/vaadin/flow/dom/ElementTest.java
mstahv: done

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
tepi: Fair enough. In normal use this should not happen unless there's a bug somewhere. Developers can of course make it happen with their own code, and in that case info level could make sense. So let's go with info level, and fix possible upcoming issues separately (e.g. something causing these in excess and filling the logs).

tepi:
Looks good to me! @mcollovati did you have any remaining comments about this one?

mcollovati:
Looks good to me too. There's still a pending discussion about the message log level, but if we want to differentiate between dev and prod, we can do it in another PR

mcollovati (approved):

tepi (approved):

mstahv:
Excellent! I migth even be so bold that I would throw illegal state exception, at least in dev mode.

Issue #19094, opened 2024-04-03
PiT 24.4: Regression in alpha19 NoClassDefFoundError and redirection to line-awesome.min.css on login


manolo:
### Description of the bug

There is an exception error when running the project in my app

```
java.lang.NoClassDefFoundError: org/jsoup/safety/Safelist
	at com.vaadin.flow.router.AbstractRouteNotFoundError.setRouteNotFoundErrorParameter(AbstractRouteNotFoundError.java:80) ~[flow-server-24.4.0.alpha26.jar:24.4.0.alpha26]
	at com.vaadin.flow.router.RouteNotFoundError.setErrorParameter(RouteNotFoundError.java:36) ~[flow-server-24.4.0.alpha26.jar:24.4.0.alpha26]
	at com.vaadin.flow.router.internal.ErrorStateRenderer.notifyNavigationTarget(ErrorStateRenderer.java:121) ~[flow-server-24.4.0.alpha26.jar:24.4.0.alpha26]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.notifyNavigationTarget(AbstractNavigationStateRenderer.java:616) ~[flow-server-24.4.0.alpha26.jar:24.4.0.alpha26]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.sendBeforeEnterEvent(AbstractNavigationStateRenderer.java:593) ~[flow-server-24.4.0.alpha26.jar:24.4.0.alpha26]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.sendBeforeEnterEventAndPopulateChain(AbstractNavigationStateRenderer.java:494) ~[flow-server-24.4.0.alpha26.jar:24.4.0.alpha26]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.createChainIfEmptyAndExecuteBeforeEnterNavigation(AbstractNavigationStateRenderer.java:466) ~[flow-server-24.4.0.alpha26.jar:24.4.0.alpha26]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.handle(AbstractNavigationStateRenderer.java:211) ~[flow-server-24.4.0.alpha26.jar:24.4.0.alpha26]
	at com.vaadin.flow.router.internal.ErrorStateRenderer.handle(ErrorStateRenderer.java:106) ~[flow-server-24.4.0.alpha26.jar:24.4.0.alpha26]
[...]

PR #19140, opened 2024-04-09
fix: fix URL rewrite for nested CSS

### LINKED ISSUE #19045: static resources can not be imported from parent theme
### Description of the bug

When I import a theme from another jar via `"parent": "parent-theme"` in theme.json, I'm getting a 404 in the browser.

NPM Dependencies are loaded via Dependency class in jar module and svg sources are imported via theme.json and assets notation.

### Expected behavior

the keyboard icon in Hello World View should be visible.

See https://github.com/SebastianKuehnau/demo-theme-app

import to IntelliJ and run Demo Spring Application

### Versions

- Vaadin / Flow version: 24.3.8
- Java version: 21
- OS version: MacOS 14.4.1
- IDE (if applicable):

mcollovati:
Fixes URL rewrites in nested CSS, in both Vite theme plugin and CssBundler.

In addition, CssBundler now tries to guess if a theme relative path that do not match a physical file might be a reference to a theme asset, and in this case it replaces the URL with one resolvable by StaticFileHandler.

Fixes #19045
Refs #19094


tltv:
Looks like there is an issue with paths with Windows. E.g. `ReusableThemeIT` gives this error:

```
[ERROR] com.vaadin.flow.uitest.ui.theme.ReusableThemeIT.cssWithAssetRelativePaths_urlPathIsNotRelative[any_Chrome_] -- Time elapsed: 8.449 s <<< FAILURE!
java.lang.AssertionError: Expecting relative asset URL to be resolved as http://localhost:8888/path/VAADIN/static/themes/reusable-theme/fortawesome/icons//keyboard.svg but was url("http://localhost:8888/path/VAADIN/themes/reusable-theme/fortawesome/icons/keyboard.svg")
        at org.junit.Assert.assertTrue(Assert.java:42)
        at com.vaadin.flow.uitest.ui.theme.ReusableThemeIT.cssWithAssetRelativePaths_urlPathIsNotRelative(ReusableThemeIT.java:157)
```

mcollovati:
I have seen a similar error when running the test more than once without clean: first time the runtime URL has `VAADIN/static/themes/`, second time it has `VAADIN/themes/`.
Deep cleaning (bundles, generated stuff, target, ...) the project made it work again.
@tltv can you check if this holds also for you? If so, there might be some mistake on project configuration 

In addition, there's an extra `/` in the error message that should be fixed (however, the assertion is correct).

tltv:
>@tltv can you check if this holds also for you? If so, there might be some mistake on project configuration

Unfortunately cleaning off all generated files didn't change the result. But tests seems to fail same way even without these changes so it could be just local issue if not OS specific. Problem seems to be related to assets not being copied at all for some reason.

I'm approving this due to code review is otherwise OK, no change requests. But as I can't verify this locally myself, I'll request others to test it.

tltv (dismissed):
LGTM. Could you @caalador run `ReusableThemeIT` in Windows to see it works for you. I had bad luck with this test so could not verify that.

caalador:
I can't get it to execute correctly either on windows. Even with the current master branch I get the same exceptions thrown and the 2 images do not load.

mcollovati:
There is definitely something incorrect in the module configuration, but probably unrelated to the mentioned issue on Windows.
If I run the test without a prod bundle, then the bundle is created, but the application deployed on Jetty does not use it.
If I run it again, with the prod bundle already there, then the application uses it and some image URLs are not correct.
The reason seems to be that with a development or production bundle, Flow does not inline the theme CSS, but it loads it directly with a `style` tag, causing relative paths to be wrong.
For example, the `style` tag points to `http://localhost:8888/path/VAADIN/themes/reusable-theme/styles.css`, that has a relative URL ` url("./test/path/monarch-butterfly.jpg")` resolved as `http://localhost:8888/path/VAADIN/themes/reusable-theme/test/path/monarch-butterfly.jpg`, resulting in a 404.

tltv (approved):

Issue #19109, opened 2024-04-04
It is not possible to specify Hilla menu information for Flow views


Artur-:
### Describe the bug

You can specify the menu order and icon for client side views e.g. as
```
export const config: ViewConfig = {
    menu: {"order":1,"icon":"file"} ,
};
```

There is no way to define this for Flow views but they are still included in the menu.
The menu class already exists but is always set to null here https://github.com/vaadin/hilla/blob/12282623322771e0351fe3ca5e34f55641a34ddf/packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java#L172

### Expected-behavior

```
@Menu(order=1, icon="file")
@Route("someroute")
public class MyView extends VerticalLayout {
```

### Reproduction

-

### System Info

Vaadin 24.4 alpha20

mshabarov:
Would it be better to have another name, e.g. `AtMenuItem` ? `AtMenu` makes me think that the annotated class is a main menu itself.

Legioth:
It might be confusing if you would see `@Menu` on its own but it will always be used together with `@Route` and I assume the association is clear enough when you see both next to each other.

mshabarov:
Acceptance Criteria:
- [ ] Add a new annotation `AtMenu` in Flow with the same parameters, as in [ViewConfig](https://github.com/vaadin/hilla/blob/a3258b4451215d8b4d2ce4c537c3576d749a9ca3/packages/ts/file-router/src/types.d.ts#L3) API in Hilla, except these three - `rolesAllowed`, `loginRequired`, `route` as they are configured with the existing Flow annotations.
- [] modify the implementation for [collectServerViews](https://github.com/vaadin/hilla/blob/a3258b4451215d8b4d2ce4c537c3576d749a9ca3/packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java#L152) and use `AtMenu` annotation data to fill in the `AvailableViewInfo` object.
- [ ] Use `AtMenu` title parameter for menu item text. If title param is absent - fallback to `AtPageTitle` annotation value. Limitation: `HasDynamicTitle` cannot be used, as view instance may not have been instantiated at that point of time.
- [ ] Quarkus may have problems with not inheritable annotations, but as Hilla is not supported in Quarkus apps this can be left as a limitation for now.

PR #19147, opened 2024-04-11
feat: add Menu annotation

tltv:
Menu annotation has information like title, order and icon for automatically populated menu. Matches with ViewConfig menu API in Hilla.

Related-To: #19109

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/Menu.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/Menu.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2000-2024 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.vaadin.flow.router;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Defines menu information for a route for automatically populated menu.
+ *
+ * @see Route
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+@Documented
+public @interface Menu {
+
+    /**
+     * Title to use in the menu. Falls back the page title if not defined.
+     *
+     * @return the title of the item in the menu. Empty String by default.
+     */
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/router/Menu.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Set to true to explicitly exclude a view from the automatically populated
+     * menu.
+     *
+     * @return true to exclude the view from the menu, false otherwise
+     */
+    boolean exclude() default false;
+
+    /**
+     * Icon to use in the menu.
```
knoobie, line 64: nit: What Icon? Which format? What is used as default "icon source"?

tltv: Format is still open and will be figured out in next steps in other PRs. Value of it is supposed to work together with ViewConfig's menu icon property and format depends on that. Flow can do some auto-conversion for paths, if it's a path. This annotation is needed also by other features so making PR for it a bit early for that reason.

Legioth: The logic that actually populates the menu based on this data will be in the main layout file that is part of the application so application logic can do whatever it wants with it. The default main layout implementation that we ship for Hilla does `{icon ? <Icon src={icon} slot="prefix"></Icon> : <></>}` so I guess the corresponding thing in the eventual Flow implementation would be to pass the value to the `Icon(String)` constructor.

knoobie: Those defaults should probably mentioned here, once those are move finalized - otherwise it gets pretty hard for a developer to know where he can get the icon name from and so on

flow-server/src/main/java/com/vaadin/flow/router/Menu.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Defines menu information for a route for automatically populated menu.
```
caalador, line 27: Could define here also is it for client and/or server side menu, also 
what menu or is it for Menu Bar that is a component that we have.
Dows it get picked automatically or is something needed to get it into use.

flow-server/src/main/java/com/vaadin/flow/router/Menu.java
tltv: > The default main layout implementation that we ship for Hilla does `{icon ? <Icon src={icon} slot="prefix"></Icon> : <></>}` so I guess the corresponding thing in the eventual Flow implementation would be to pass the value to the `Icon(String)` constructor.

@Legioth Will it be `<Icon src={icon}` or `<Icon icon={icon}`? `icon` attribute takes in icon group with name like "vaadin:file", but src would take in something else.

Legioth: The current template in the preview branch for start.vaadin.com uses `<Icon src={icon}>` in combination with configuration such as `icon: 'line-awesome/svg/lock-open-solid.svg'` for views.

tltv: Alright, might as well mention both icon and src attributes of Icon with examples in Javadoc.

caalador (approved):

PR #19170, opened 2024-04-15
fix: allow css properties in react components (#19077)

### LINKED ISSUE #19077: Using css properties in inline styles in React components should not trigger TypeScript errors
### Description of the bug

If you write
```
      <HorizontalLayout style={{ '--foo': 'red' }}>
```

you get "TS2353: Object literal may only specify known properties, and '--foo' does not exist in type Properties<string | number, string & {}>"

### Expected behavior

It should be allowed, e.g. by adding
```
declare module 'csstype' {
  interface Properties {
    // Allow any CSS Custom Properties
    [index: `--${string}`]: any;
  }
}

```

to `types.d.ts`

      <HorizontalLayout style={{ '--foo': 'red' }}>

### Versions

- Vaadin / Flow version: 24.4 alpha
- Java version:
- OS version:
- IDE (if applicable):

tepi:
Fixes #19077


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsDefinitions.java
```diff
@@ -200,6 +245,12 @@ private UpdateMode computeUpdateMode(String content)
                 // Current content has been written by Flow, can be replaced
                 return UpdateMode.REPLACE;
             }
+            templateContent = getTemplateContent(".v2");
+            if (compareIgnoringIndentationAndEOL(content, templateContent,
+                    String::equals)) {
+                // Current content has been written by Flow, can be replaced
```
mshabarov, line 251: What happens if current content has Hilla content + `*.css?inline` default block (basically any V24.4 app that has Hilla) ?
I suspect the `.hilla` content check will fail and status will be `UpdateMode.REPLACE`. Then Flow will replace  the current content with the default content, which has no Hilla stuff.

Thus, I'd suggest to add one more template `.hilla.v2` that has previous `.hilla` and `*.css?inline` block.

Yeah, this turns into too complicated logic unfortunately.

P.S. I haven't tested yet, maybe it works without these tweaks, let me check.

tepi: In that case it should go to `UpdateMode.UPDATE_AND_BACKUP` mode and add whichever of the two blocks is missing, but please do test and let me know if it fails.

mshabarov: Yep, it shows a warning `UPDATE_MESSAGE` message and back-ups the content in case of hybrid app (for Hilla app I suppose the same).

It should just update the content without backup and without this message.

So we need one more hilla template (or merge .hilla and v2 templates)

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsDefinitions.java
mshabarov: > In that case it should go to UpdateMode.UPDATE_AND_BACKUP

Hybrid starter has nothing extra. In this case it should't make a backup, nor show a warning message. Just update silently.

tepi: Should be fixed now

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateTsDefinitions.java
mshabarov: Yeah, works as expected, well done!

PR #19196, opened 2024-04-17
fix: make property change listener trigger on removeProperty

### LINKED ISSUE #3994: Element.removeProperty should fire PropertyChangeEvent if property was removed

akhodyko:
Removing element property by calling `com.vaadin.flow.internal.nodefeature.AbstractPropertyMap#removeProperty` does not trigger property change listener because of `com.vaadin.flow.internal.nodefeature.ElementPropertyMap#remove` method is skipped due to `AbstractPropertyMap#removeProperty` method calls super:
```
    public void removeProperty(String name) {
        super.remove(name);
    }
```

Fixes #3994

### DIFF
--- a/flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
+++ b/flow-data/src/test/java/com/vaadin/flow/data/binder/BinderTest.java
@@ -2613,8 +2613,7 @@ public class BinderTest extends BinderTestBase<Binder<Person>, Person> {
                     throw new NullPointerException();
                 }, name -> name))
                 .bind(Person::getFirstName, Person::setFirstName)
-                .read(new Person());
-
+                .read(Person.createTestPerson1());
     }
 
     @Test(expected = BindingException.class)
--- a/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/AbstractPropertyMap.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/AbstractPropertyMap.java
@@ -81,7 +81,7 @@ public abstract class AbstractPropertyMap extends NodeMap {
      *            the name of the property to remove
      */
     public void removeProperty(String name) {
-        super.remove(name);
+        remove(name);
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/ElementPropertyMap.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/ElementPropertyMap.java
@@ -193,7 +193,7 @@ public class ElementPropertyMap extends AbstractPropertyMap {
         Serializable oldValue = super.remove(key);
 
         fireEvent(new PropertyChangeEvent(Element.get(getNode()), key, oldValue,
-                true));
+                false));
 
         return oldValue;
     }
--- a/flow-server/src/test/java/com/vaadin/flow/component/AbstractSinglePropertyFieldTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/AbstractSinglePropertyFieldTest.java
@@ -33,9 +33,6 @@ import com.vaadin.flow.dom.DomListenerRegistration;
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.function.SerializableRunnable;
 import com.vaadin.flow.internal.JsonUtils;
-import com.vaadin.flow.server.VaadinService;
-import com.vaadin.flow.server.VaadinSession;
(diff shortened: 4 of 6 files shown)


mshabarov (dismissed):

akhodyko:
> @akhodyko thanks for you contribution! Looks good to me! May I transfer this pull request to "ready for review" state?

yes, you can, but there are some trouble with `com.vaadin.flow.data.binder.BinderTest.readBean_converterThrows_exceptionHandlerSet_bindingExceptionIsThrown` test yet.

mshabarov:
> > @akhodyko thanks for you contribution! Looks good to me! May I transfer this pull request to "ready for review" state?
> 
> yes, you can, but there are some trouble with `com.vaadin.flow.data.binder.BinderTest.readBean_converterThrows_exceptionHandlerSet_bindingExceptionIsThrown` test yet.

That seems wrong, if the Binder doesn't throw while the exception happens in the converter. Weird side effect. We will take a look, thanks!

akhodyko:
@mshabarov 

I think `readBean_converterThrows_exceptionHandlerSet_bindingExceptionIsThrown` test is not fully correct.

I will try to explain:
1. Before this PR, setting `null` value to `com.vaadin.flow.component.AbstractSinglePropertyField` was not triggered property change listener on removing property here `com.vaadin.flow.component.AbstractSinglePropertyField.TypeHandler#createWriter`. 
2. Because of point 1, `com.vaadin.flow.component.internal.AbstractFieldSupport#pendingValueFromPresentation` was not updated to actual value taking into account default value of field. This caused the field value change event to fire. 
3. After that, value change event caused validation at `com.vaadin.flow.data.binder.Binder#handleFieldValueChange` followed with conversion field value to model.

According to that, because of we have in that test `com.vaadin.flow.data.binder.testcomponents.TestTextField` accepting null values with default empty string value, null value must be treated as empty string and should not trigger the change listener and toModel conversion. 

Replacing bean Person to another with non null firstName field fixed the test.

mshabarov (changes requested):
As the property removal is now considered as non-client originated event, we have to change to `Assert.assertFalse(event.get().isUserOriginated());` in `ElementPropertyMapTest::removeProperty_fireEvent_listenerIsNotNotified`

mshabarov:
> According to that, because of we have in that test com.vaadin.flow.data.binder.testcomponents.TestTextField accepting null values with default empty string value, null value must be treated as empty string and should not trigger the change listener and toModel conversion.
Replacing bean Person to another with non null firstName field fixed the test.

This makes sense, okay.

mshabarov:
Requested another review from @caalador , just in case, if I missed something important, the overall fix looks good to me and correct.

akhodyko:
> As the property removal is now considered as non-client originated event, we have to change to `Assert.assertFalse(event.get().isUserOriginated());` in `ElementPropertyMapTest::removeProperty_fireEvent_listenerIsNotNotified`

done

caalador (commented):
Looks good to me.

mshabarov:
@akhodyko could you please run `mvn formatter:format` and push? Then we can merge this change.

akhodyko:
> @akhodyko could you please run `mvn formatter:format` and push? Then we can merge this change.

done

mshabarov (approved):

mshabarov:
This has been reverted due to a regression in Vaadin Flow components (to be more precise, in `RadioButtonGroup`) that blocks our releases, see https://github.com/vaadin/flow/pull/19310.

For the record, the failed test is this:
```
[setIdentifierProviderOnId_setItemWithNullId_shouldFailToSelectExistingItemById] java.lang.NullPointerException: Cannot invoke "com.vaadin.flow.component.radiobutton.CustomItem.getId()" because the return value of "com.vaadin.flow.component.radiobutton.RadioButtonGroup.getValue()" is null
	at com.vaadin.flow.component.radiobutton.RadioButtonGroupTest.setIdentifierProviderOnId_setItemWithNullId_shouldFailToSelectExistingItemById(RadioButtonGroupTest.java:441)
```

I.e. the component reset the value to `null`, if `setValue` is called with an item, that is not present in the items list.
We plan to reconsider the logic in RadioButtonGroup regards invalid items and re-apply this patch again.

akhodyko:
Are there any plans to release this with 24.5.0?

mshabarov:
@akhodyko no plans unfortunately, this was reverted and we haven't got back to it again afterwards, because of other priorities. This patch opened the can of worms in Flow components, we tried to fix that, but it turned into complicated problem that even more could bring a behaviour change in components, see the description above.

PR #19199, opened 2024-04-18
chore: return removed methods

caalador:
Add back the removed methods
as deprecated for compatibility.

Methods are not exposed for
calling from the client.

Closes vaadin/platform#5244


flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1759,6 +1759,31 @@ public BrowserNavigateEvent(UI source, boolean fromClient,
...
+     * @param historyState
+     *            client side history state value
+     * @param trigger
+     *            navigation trigger
+     *
+     * @deprecated method is not enabled for client side anymore and
+     *             connectClient is triggered by DOM event, to be removed in
+     *             next major 25
+     */
+    public void connectClient(String flowRoutePath, String flowRouteQuery,
```
knoobie, line 1781: Add `@Deprecated(forRemoval=true)` to both methods?

mcollovati: Should we also add the Java `@Deprecated` annotation?

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1834,6 +1859,27 @@ private void replaceStateIfDiffersAndNoReplacePending(String route,
...
+     *
+     * @param route
+     *            the route that is navigating to.
+     * @param query
+     *            route query string
+     * @deprecated(forRemoval=true) method is not enabled for client side
+     *                              anymore and leave navigation is triggered by
+     *                              DOM event, to be removed in next major 25
+     */
+    public void leaveNavigation(String route, String query) {
```
mcollovati, line 1878: Should we also add the Java `@Deprecated` annotation?

mcollovati (approved):

Issue #19181, opened 2024-04-17
Format requirement for routes.tsx is too strict


Legioth:
### Description of the bug

The requirement that `routes.tsx` contains `export const { router, routes } = new RouterConfigurationBuilder()` is too strict in case I want to manually modify the `routes` value before it's exported. In that case, I would want to structure my code like this:

```
const { router, originalRoutes } = new RouterConfigurationBuilder();
    .withFallback(Flow)
    .build();

const routes = originalRoutes; // This is where the modification would be done
export { router, routes }
```

If I use this structure, then Flow will complain that `routes.tsx` doesn't export `routes` even though it's indeed doing that.

### Expected behavior

There will be an obvious way of customizing the routes beyond what's easily doable through `RouterConfigurationBuilder`.

### Minimal reproducible example

1. Get any starter from anywhere
2. Add a `routes.tsx` file with the content described above
3. `mvn`

### Versions

- Vaadin / Flow version: Vaadin 24.4.0.alpha22

PR #19203, opened 2024-04-18
fix: Use less strict regexp for export validation

mshabarov:
Fixes https://github.com/vaadin/flow/issues/19181

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
@@ -112,7 +112,7 @@ public class TaskGenerateReactFiles implements FallibleCommand {
             "import\\s+(\\w+)\\s+from\\s+(\"|'|`)Frontend\\/generated\\/flow\\/Flow(\\.js)?\\2;[\\s\\S]+withFallback\\(\\s*\\1\\s*\\)");
 
     private static final Pattern ROUTES_EXPORT_PATTERN = Pattern.compile(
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -112,7 +112,7 @@ function build() {
             "import\\s+(\\w+)\\s+from\\s+(\"|'|`)Frontend\\/generated\\/flow\\/Flow(\\.js)?\\2;[\\s\\S]+withFallback\\(\\s*\\1\\s*\\)");
 
     private static final Pattern ROUTES_EXPORT_PATTERN = Pattern.compile(
-            "export\\s+const\\s+(\\{[\\s\\S]*(router[\\s\\S]*routes|routes[\\s\\S]*router)[\\s\\S]*}|routes)");
+            "export\\s+(const\\s+)?(\\{[\\s\\S]*(router[\\s\\S]*routes|routes[\\s\\S]*router)[\\s\\S]*}|routes)");
```
caalador, line 115: What about a case where the setup would be 
```
...

export { router }
// Some other code here
export { routes }
```

It still exposes bot required exports, but they just are not in the same export clause

mshabarov: By the way, this case is covered by the existing regexp. I added a couple of unit tests that check the case above (no exception) and with only a single export `export { routes }` (throws).

caalador (approved):

PR #19205, opened 2024-04-18
chore: Delete UI elements from dev tools

Artur-:
Fixes https://github.com/vaadin/copilot-internal/issues/1498


czp13 (commented):
This seems good to me from eyeball check, but let's wait for another pair of flow eyes.
(package.json seems a bit unrelated, but it's ok)

MarcinVaadin:
Why do we still need `@Element("vaadin-dev-tools")
public class DevToolsElement extends TestBenchElement {` ?

MarcinVaadin (approved):

flow-tests/test-frontend/vite-basics/package.json
caalador, line 1: The updated versions are not up to date.

PR #19220, opened 2024-04-22
chore: Add type to FlowContainer

Artur-:
Makes it possible for Copilot to detect it without importing the file. Importing the file triggers the esbuild import order bug


Artur-:
Let's see that this is in sync with https://github.com/vaadin/hilla/pull/2349 before merging it

PR #19037, opened 2024-03-25
fix: delete only unnecessary generated files

mcollovati:
Prevents built tool plugins to to delete all frontend genreated files as first step in the prepare-frontend task, moving the login in BuildFrontendUtil and changing it to only delete files that are no longer created by node tasks.

Part of #18907

PR #19254, opened 2024-04-25
fix: Generate flow imports in the same folder

mshabarov:


mcollovati:
Just a note: since #19037 has been merged, handling of the webcomponent imports file could be added in  `TaskRemoveOldFrontendGeneratedFiles.isKnownUnhandledFile`

caalador (dismissed):

mshabarov:
> Just a note: since #19037 has been merged, handling of the webcomponent imports file could be added in `TaskRemoveOldFrontendGeneratedFiles.isKnownUnhandledFile`

Added to `isKnownUnhandledFile`.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRemoveOldFrontendGeneratedFiles.java
```diff
@@ -122,11 +123,15 @@ public FileVisitResult postVisitDirectory(Path dir,
     }
 
     private Predicate<Path> isKnownUnhandledFile() {
+        File frontendFolder = frontendGeneratedFolder.getParent().toFile();
```
caalador, line 126: Perhaps just put the Options in a variable and use it here to get the frontend folder from options instead of using the generated/parent as there is a minor chance that the user has moved the generated folder to some other path as that is allowed

mshabarov: Uses frontend folder from Options now.

PR #19257, opened 2024-04-25
feat: add check for Hilla routes in VaadinWebSecurity

cromoteca:
This PR adds a check for Hilla routes so that developers (and our Start) do not have to do themselves.

As the check happens in a Spring bean, nothing changes if Hilla is not on the classpath, as the implementation will be added in Hilla.

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
@@ -209,6 +209,9 @@ public abstract class VaadinWebSecurity {
             // Public endpoints are OK to access
             urlRegistry.requestMatchers(requestUtil::isAnonymousEndpoint)
                     .permitAll();
+            // Checks for known Hilla views
+            urlRegistry.requestMatchers(requestUtil::isAllowedHillaView)
+                    .permitAll();
             // Public routes are OK to access
             urlRegistry.requestMatchers(requestUtil::isAnonymousRoute)
                     .permitAll();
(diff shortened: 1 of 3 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -209,6 +209,9 @@ protected void configure(HttpSecurity http) throws Exception {
             // Public endpoints are OK to access
```
knoobie, line 209: @mshabarov Generel question: Should there be a check in `VaadinWebSecurity` for Hilla availability and not in every RequestUtil method, so that the `urlRegistry.requestMatchers()` is only called if Hilla is available to reduce adding unnessary requestmatcher? Each of those costs resources / over head.

tepi: I'm not confident we have that information available at the point when security filter chain is being registered/configured.

mshabarov: Yeah, I'm not sure either. `EndpointRequestUtil.isHillaAvailable()` can probably be used, but I don't know now reliable it is for runtime checks.

A better way IMO would be to store the result of `FrontendUtils.isHillaUsed` to the token file (or to another place where we have a build information) during build time and restore this value in runtime through the util method. This logic isn't implemented though yet.

I don't think that it would have a big impact on performance for Flow apps, as it returns false immediately and for Hilla apps this check should be done anyway.

mshabarov (approved):

Issue #16879, opened 2023-05-25
Support for serialization of record object to Json


rmuller:
Vaadin 24 currently does not support serialization of record objects to JSON. Which is a pity because of the great DX record type bring on the table. 

Serialization is done by `com.vaadin.flow.internal.JsonSerializer`. It is easy to extend it with support for record types (depending on `com.vaadin.external.gwt:gwt-elemental:2.8.2.vaadin2` for the real work).

Instead of 

        try {
            JsonObject json = Json.createObject();
            BeanInfo info = Introspector.getBeanInfo(bean.getClass());
            for (PropertyDescriptor pd : info.getPropertyDescriptors()) {
                if ("class".equals(pd.getName())) {
                    continue;
                }
                Method reader = pd.getReadMethod();
                if (reader != null) {
                    json.put(pd.getName(), toJson(reader.invoke(bean)));
                }
            }

            return json;
        } catch (Exception e) {
            throw new IllegalArgumentException(
                    "Could not serialize object of type " + bean.getClass()
                            + " to JsonValue",
                    e);
        }

Use

        try {
            JsonObject json = Json.createObject();
            var type = bean.getClass();
            
            if (type.isRecord()) {
                for (var rc : type.getRecordComponents()) {
                    json.put(rc.getName(), toJson(rc.getAccessor().invoke(bean)));
                }
                return json;
            }            
            
            BeanInfo info = Introspector.getBeanInfo(type);
            for (PropertyDescriptor pd : info.getPropertyDescriptors()) {
                if ("class".equals(pd.getName())) {
                    continue;
                }
                Method reader = pd.getReadMethod();
                if (reader != null) {
                    json.put(pd.getName(), toJson(reader.invoke(bean)));
                }
            }
            return json;
        } catch (Exception e) {
            throw new IllegalArgumentException(
                "Could not serialize object of type %s to JsonValue".formatted(bean.getClass()), e);
        }

Legioth:
Something like that would indeed make sense starting from Vaadin 24 since we no longer need to support Java versions without `Class::getRecordComponents` and similar APIs.

Legioth:
Should probably also do something similar for `BeanPropertySet` so that you could do `new Grid<>(MyRecordType.class)` to automatically configure columns based on record components.

rmuller:
> Should probably also do something similar for BeanPropertySet so that you could do new Grid<>(MyRecordType.class) to automatically configure columns based on record components.

[...]

PR #19258, opened 2024-04-25
feat: record support in json serializer (#16879)

tepi:
Fixes https://github.com/vaadin/flow/issues/16879


flow-server/src/test/java/com/vaadin/flow/internal/JsonSerializerTest.java
```diff
@@ -45,6 +45,9 @@ public static enum SomeEnum {
         SOME_VALUE_1, SOME_VALUE_2;
     }
 
+    public record SomeRecord(String name, int age) {
```
knoobie, line 48: Should there be a test for a record with a record or a record with an object as parameter to ensure all those combination work as expected?

tepi: Good point, thanks. I'll add one tomorrow.

tepi:
Still need record support in `com.vaadin.flow.internal.JsonSerializer#toObject(java.lang.Class<T>, elemental.json.JsonValue)` so marking as draft until that is in.

flow-server/src/test/java/com/vaadin/flow/internal/JsonSerializerTest.java
tepi: Added - and that was a good thing. The test revealed that `JsonSerializer.toObject` also needed updating to handle records.

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/internal/JsonSerializer.java
```diff
@@ -85,14 +85,23 @@ public static JsonValue toJson(Object bean) {
 
         try {
             JsonObject json = Json.createObject();
-            BeanInfo info = Introspector.getBeanInfo(bean.getClass());
-            for (PropertyDescriptor pd : info.getPropertyDescriptors()) {
-                if ("class".equals(pd.getName())) {
-                    continue;
+            var type = bean.getClass();
```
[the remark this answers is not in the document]
tepi: True. Maybe I'll change those vars since they are not used in this class otherwise.

caalador (approved):

PR #19259, opened 2024-04-25
feat: record support in beanpropertyset

tepi:
Related to https://github.com/vaadin/flow/issues/16879


caalador (commented):
Seems a bit strange that we have no tests for BeanUtil in the flow-server module, only in flow-data.
Should we have some simple test in server also specifically for the methods we have?

tepi:
> Seems a bit strange that we have no tests for BeanUtil in the flow-server module, only in flow-data. Should we have some simple test in server also specifically for the methods we have?

Yeah, noticed the same problem. I think there should be some tests in flow-server as well. Although not sure if tests should be added in context of this issue or in a separate PR.

tepi:
https://github.com/vaadin/flow/issues/19264

PR #19266, opened 2024-04-26
Revert "fix: prevent embedded styles to leak main document"

mshabarov:
MERGE, NOT SQUASH

This reverts https://github.com/vaadin/flow/pull/19221, as it introduced a delay in applying styles for Hilla apps and makes Hilla apps broken in Safari, where styles aren't applied at all.

Fixes https://github.com/vaadin/flow/issues/19265


caalador (approved):

knoobie:
This was also released in 23.3 & 23.4

mshabarov:
> This was also released in 23.3 & 23.4

Yes, we most likely have to revert them also, however, I don't know at the moment, how much these versions are affected. Flow apps have no such a problem, but only Hilla apps do. We'll discuss it internally shortly.

PR #19270, opened 2024-04-26
feat: Open tab in dev mode less often

### LINKED ISSUE #18393: Reuse the existing tab in browser when Vaadin app is started in dev mode
### Describe your motivation

When Vaadin application is started in dev mode, it always opens a new tab in browser.
This quickly leads to many tabs opened in browser, which is annoying.
Would be good, if Vaadin reused the existing tab in the browser, if it exists.

-

### Additional context

See [BrowserLauncher](https://github.com/vaadin/flow/blob/main/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/BrowserLauncher.java) (Flow repo) and [Open library](https://github.com/vaadin/open) (Separate Vaadin repo).

caalador:
As we can not know if a tab
for the application is open
we add a file that is checked
for modification time and after
default 30 min we open a new tab on
server restart.

If server is restarted again before
timeout file timestamp is updated.

Clearing build folder will also
cause new tab to open on next restart.

Closes #18393

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
@@ -278,5 +278,5 @@ public class InitParameters implements Serializable {
      *
      * Time is given in minutes.
      */
-    public static final String LAUNCH_BROWSER_DELAY = "launch-browser.delay";
+    public static final String LAUNCH_BROWSER_DELAY = "launch-browser-delay";
 }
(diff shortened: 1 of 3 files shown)


caalador:
Time can be modified with the `vaadin.launch-browser.delay` system parameter

mcollovati (commented):
It works like a charm.
I would add the property also to `VaadinConfigurationProperties` so it can be autocompleted in the IDE. Only thing I'm not completely sure if naming can clash with the existing `launch-browser` property.

caalador:
Added and changed to all - instead of - and .

mcollovati (approved):

Issue #19268, opened 2024-04-26
Wrong import is suggested in IntelliJ


jcgueriaud1:
### Describe the bug

In 24.4.0.alpha24, IntelliJ is suggesting these 2 imports. I guess the correct one is `./ReactAdapter` but is there a way to not have the `Frontend/generated/flow/ReactAdapter`.

### Expected-behavior

IntelliJ should suggest only one import.

### Reproduction

I'm using a project from start.vaadin.com. Run the project and add this code:
```
import { ReactElement } from "react";
import { ReactAdapterElement, RenderHooks } from "./generated/flow/ReactAdapter";
import { Button } from "@vaadin/react-components";

class ReactExample extends ReactAdapterElement {
    protected render(hooks: RenderHooks): ReactElement {
        const [value] = hooks.useState("value", "initial");
        const sendEvent = hooks.useCustomEvent("example-event", {detail: ""});
        return <Button onClick={() => sendEvent("Bar")}>{value}</Button>
    }
}

customElements.define("react-example", ReactExample);
```

Remove the import and try to auto-import.

### System Info

MacOS, Java 21

tepi:
Flow test code does `import { ReactAdapterElement } from 'Frontend/generated/flow/ReactAdapter.js';` in `test-react-adapter` module. Not sure if correct or not.

jcgueriaud1:
Here is what IntelliJ is suggesting
https://github.com/vaadin/flow/assets/51313578/71bf79a7-5831-44a2-bf28-e1719978965c

PR #19275, opened 2024-04-29
fix: export reactadapter as .template, not .tsx

tepi:
Renames ReactAdapter.tsx to ReactAdapter.template when building flow-server jar. This prevents newer versions of IDEA trying to offer this file for import. File will be renamed back to .tsx when it is written to generated files folder.

Fixes https://github.com/vaadin/flow/issues/19268


flow-react/pom.xml
```diff
@@ -48,6 +48,24 @@
...
+                <executions>
+                    <execution>
+                        <id>copy-and-rename-file</id>
+                        <phase>compile</phase>
+                        <goals>
+                            <goal>rename</goal>
+                        </goals>
+                        <configuration>
+                            <sourceFile>${project.build.outputDirectory}/com/vaadin/flow/server/frontend/ReactAdapter.tsx</sourceFile>
+                            <destinationFile>${project.build.outputDirectory}/com/vaadin/flow/server/frontend/ReactAdapter.template</destinationFile>
```
caalador, line 64: I guess this works, but do we lose something if we would just rename the file to template instead of tsx.
The adapter tsx file is probably not that often edited and can on development be renamed if needed.

tepi: That would be nicer option in my mind as well. What we lose is IDEA identifying file type and offering syntax, completions etc.

But fair point, should be ok to just temporarily rename it if needed.

caalador: Also it seems idea allows selecting filtype:
`IDEA allows temporarily setting a file type : open "Actions" menu (Ctrl+Shift+A), and look for "override file type" item.`

caalador (approved):

Issue #19117, opened 2024-04-05
Configurable class scanning filters for JAR packages


tltv:
### Describe your motivation

In Spring applications that uses `VaadinServletContextInitializer`, reading JARs and scanning classes takes significant amount of time on startup and reload. Packages to scan can be already filtered with `vaadin.blocked-packages` environment property, but it takes action from the developer. In addition, it would be convenient if JAR packages already has that information, and Flow could read it to include or exclude packages dynamically.

### Describe the solution you'd like

Design a new file format to put optionally in a JAR package that has three optional properties:
- excluded packages
- included packages
- exclude whole jar

Rules:

- Other properties are ignored if excluding whole jar.
- Everything is included by default.
- Included packages overrides excluded packages. E.g. excluding "com.vaadin.hilla" and including "com.vaadin.hilla.startup" would ignore everything else under "com.vaadin.hilla" than "com.vaadin.hilla.startup" and also include everything else in the JAR.
- Having only included packages inverts the default to exclude everything else than the included packages.
- Asterix `*` or `*.*` should not be needed here. Everything is filtered with _contains_ condition.

Reading the file would be done in `VaadinServletContextInitializer#CustomResourceLoader`.

### Describe alternatives you've considered

Only alternative is to keep using hardcoded jar/package filtering as currently and keep maintaining it.

### Additional context

For reference: https://github.com/vaadin/flow/issues/18812, https://github.com/vaadin/flow/issues/19112

tltv:
New suggestion of the format, name and location for the configuration file:

Location and name inside a JAR: `META-INF/VAADIN/package.properties`
Example format, where each is optional:
```
vaadin.blocked-packages=com/vaadin/foo,com/vaadin/bar
vaadin.allowed-packages=com/vaadin/foo/bar,com/vaadin/bar/foo
vaadin.blocked-jar=false
```

if `vaadin.blocked-jar=true`, then JAR content shouldn't be scanned at all other than the `package.properties` file content.

Property naming pattern follows [what is already used](https://vaadin.com/docs/latest/integrations/spring/configuration#configure-the-scanning-of-packages) in project's application properties. You can't mix allowed-packages and blocked-packages there and it should be same with the `package.properties` too to avoid confusion, so this part would be a change to the ticket's description. It should say now "You should use either allowed-packages or blocked-packages — not both. If both have values, though, blocked-packages is ignored." as documented already.

PR #19299, opened 2024-05-02
feat: filterable class scanning of JAR

tltv:
This change improves optimization of class scanning of `VaadinServletContextInitializer` by reading optional configuration file within a JAR dependency. Configuration file is a regular properties file that can contain `vaadin.allowed-packages` or `vaadin.blocked-packages` to exclude/include packages being scanned by `VaadinServletContextInitializer`. Package exclusion/inclusion effects only classes in the JAR/module where `META-INF/VAADIN/package.properties` is located at.

Notice that this is only affecting the performance of the class scanning of Vaadin specific annotations and types and this feature is not meant to be used for excluding packages from the whole application. E.g. excluding `VaadinServiceInitListener` in a Spring application is not stopping Spring and Vaadin initializing it anyway, although some time-consuming actions like `DevModeServletContextListener` would still exclude it for optimization.

Fixes: #19117

### DIFF
--- a/flow-tests/vaadin-spring-tests/test-spring-filter-packages/lib-exclude/pom.xml
+++ b/flow-tests/vaadin-spring-tests/test-spring-filter-packages/lib-exclude/pom.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+~ Copyright 2000-2024 Vaadin Ltd.
+~
+~ Licensed under the Apache License, Version 2.0 (the "License"); you may not
+~ use this file except in compliance with the License. You may obtain a copy of
+~ the License at
+~
+~ http://www.apache.org/licenses/LICENSE-2.0
+~
+~ Unless required by applicable law or agreed to in writing, software
+~ distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+~ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+~ License for the specific language governing permissions and limitations under
+~ the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<artifactId>vaadin-test-spring-filter-packages</artifactId>
+		<groupId>com.vaadin</groupId>
+		<version>24.4-SNAPSHOT</version>
+	</parent>
+	<artifactId>vaadin-test-spring-filter-packages-lib-exclude</artifactId>
+	<name>Library with vaadin.blocked-jar property
+    </name>
+
+	<packaging>jar</packaging>
+	<properties>
+		<maven.deploy.skip>true</maven.deploy.skip>
+	</properties>
+
+	<dependencies>
+		<dependency>
+			<groupId>com.vaadin</groupId>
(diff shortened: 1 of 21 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
```diff
@@ -0,0 +1,371 @@
...
+     *         properties,
+     */
+    protected boolean isAllowedByPackageProperties(String rootPath,
+            String targetPath, boolean defaultValue) {
+        Properties properties = propertiesCache.get(rootPath);
+        if (properties == null) {
+            return defaultValue;
+        }
+
+        List<String> allowedPackages = Stream.of(properties
```
mcollovati, line 358: The properties file computation is executed several times, and could potentially slow a bit in the startup.
Would it make sense to put in the `propertiesCache` an object with all information got from the properties file?
I mean, parse the properties file once and put into the cache something like `record PackageInfo(Set<String> allowedPackages, List<String> blockedPackages,...)`

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
```diff
@@ -0,0 +1,371 @@
...
+     * included.
+     *
+     * @param resource
+     *            the resource to check
+     * @return {@code true} if the resource is a blocked jar, {@code false}
+     *         otherwise
+     */
+    protected boolean isBlockedJar(Resource resource) {
+        // placeholder to handle case of package.properties with
+        // vaadin.blocked-jar=true
```
mcollovati, line 244: Is the `blocked-jar` feature implemented? I couldn't find it anywhere

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
```diff
@@ -0,0 +1,371 @@
...
+                        ? PropertiesLoaderUtils.loadProperties(resource)
+                        : null;
+                propertiesCache.put(rootPath, properties);
+                getLogger().trace("Caching package.properties of directory {}",
+                        rootPath);
+            }
+
+        } catch (IOException e) {
+            getLogger().warn("Failed to find " + PACKAGE_PROPERTIES_PATH
+                    + " for path " + res, e);
```
mcollovati, line 229: nit: could use placeholders instead of string concatenation

flow-tests/vaadin-spring-tests/test-spring-filter-packages/lib-exclude/pom.xml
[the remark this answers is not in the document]
tltv: I planned originally to implement also support for a property to block whole jar, but as allowed/blocked packages feature is big one alone, and we have another ticket for the whole jar blocking (https://github.com/vaadin/flow/issues/19112), this module is just a placeholder for upcoming tests when 19112 gets started.

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
mcollovati: Already answered in https://github.com/vaadin/flow/pull/19299#discussion_r1590640258

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
tltv: Makes sense, and with a record it's less boilerplate.

mcollovati (approved):

PR #19309, opened 2024-05-06
fix!: rename createWebComponent method

caalador:
Rename the createWebComponent function
as the create gives the impression that
the function is heavy.

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -446,7 +446,7 @@ interface Properties {
  * @param onload optional callback to be called for script onload
  * @param onerror optional callback for error loading the script
  */
-export const createWebComponent = (tag: string, props?: Properties, onload?: () => void, onerror?: (err:any) => void) => {
+export const webComponentTSX = (tag: string, props?: Properties, onload?: () => void, onerror?: (err:any) => void) => {
     loadComponentScript(tag).then(() => onload?.(), (err) => {
         if(onerror) {
             onerror(err);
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -446,7 +446,7 @@ interface Properties {
  * @param onload optional callback to be called for script onload
  * @param onerror optional callback for error loading the script
  */
-export const createWebComponent = (tag: string, props?: Properties, onload?: () => void, onerror?: (err:any) => void) => {
+export const webComponentTSX = (tag: string, props?: Properties, onload?: () => void, onerror?: (err:any) => void) => {
```
tltv, line 449: Is this limited to work with TSX only, or could it even be just short as `webComponent`?

caalador: Well it returns a `React element` and not a `webComponent`.  Another that was was `getWebComponent`, but that also feels missleading.

Should it actually be `flowReactElement`? This would not mix in webComponent into the name as it would explicitly name what is done.

tltv: or just `reactElement` then. It's imported from a file called Flow, so context is already known by the caller.

mshabarov: `reactElement` sounds good to me as well.
`TSX` postfix is not needed IMO.

mshabarov:
This has a chance to be included into 24.4 for better DX, so I've added the cherry-pick label for 24.4.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
caalador: renamed.

TSX was only for if it was webComponent so you would know what to expect as you are not getting a webComponent.

mshabarov (approved):

mshabarov:
Let's wait for another opinions for a while.

marcushellberg:
I like `reactElement`, simple and clear.

PR #19322, opened 2024-05-07
fix: add explicit dependency to JDK classes in vaadin-dev-server

### LINKED ISSUE #17159: Theme and jar file watching from other Maven modules does not work in Wildfly

mcollovati:
Adds the Dependencies manifest entry to the vaadin-dev-server artifact to prevent runtime errors raised by FileWatcher when application is deployed in Wildfly.

Fixes #17159
Fixes vaadin/skeleton-starter-flow-cdi#636

### DIFF
--- a/vaadin-dev-server/pom.xml
+++ b/vaadin-dev-server/pom.xml
@@ -102,6 +102,17 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Dependencies>jdk.unsupported</Dependencies>
+                        </manifestEntries>
+                    </archive>
+                </configuration>
+            </plugin>
+            <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>exec-maven-plugin</artifactId>
                 <version>3.2.0</version>


vaadin-dev-server/pom.xml
```diff
@@ -101,6 +101,17 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
```
mshabarov, line 106: Does this cover also `war` cases? I mean, the [workaround](https://github.com/vaadin/flow/issues/17159#issuecomment-1617703993) mentioned `maven-war-plugin`.

mcollovati: This works at the artifact level, the workaround was at the application level.
Adding the manifest entry to `vaadin-dev-server` jar, would resolve the issue without having to add the entry to the application WAR.

mcollovati: I mean, the issue happens because a `vaadin-dev-server` dependency is using JDK internal classes. So I think it makes sense to add the manifest entry there.

mshabarov (approved):

Issue #19200, opened 2024-04-18
Flow page title is not propagated to the top part of the app layout in a Hilla main layout


Legioth:
### Description of the bug

When using a Hilla main menu from start.vaadin.com, the page title of a Flow view (set using `@PageTitle` or `HasDynamicTitle`) is applied to `document.title` but the same value is not shown in the top area above the view  (the `<h2 slot="navbar">` element). Instead, the title in the navbar falls back to the application name. This is inconsistent with how it works for Hilla views in the same application and with how it works in an application with a main layout implemented in Flow.

Even though this should probably be ultimately fixed in Hilla or Start, I'm filing it against Flow since I expect that some new feature will have to be added to Flow first and then only enabled through Hilla or Start.

### Expected behavior

Expected that the Flow view title is shown in the same way as Hilla page titles.

### Minimal reproducible example

1. Create an application at start.vaadin.com with one Hilla view and one Flow view. Give each view a distinct name (e.g. "Hilla" and "Flow")
2. `mvn`
3. Navigate to the Hilla view. Observe that that the name of the view is shown in the top of the main layout.
4. Navigate to the Flow view (not that it's not in the menu so you need to type the URL or go thorough the dev mode 404 view). Observe that the value shown in the top of the main layout is the name of the app rather than the name of the view.

### Versions

- Vaadin / Flow version: Vaadin 24.4.0.alpha22 (or alpha21 which is what you get from Start today)

mshabarov:
I propose to retest this issue after we implement https://github.com/vaadin/flow/issues/19127.

Legioth:
The menu title does not necessarily have anything to do with the page title, especially in the case of `HasDynamicTitle` and views that aren't even directly shown in the menu.

caalador:
This is implemented in the layout component as `currentTitle = useRouteMetadata()?.title` so that would mean that the Flow element should implement the metadata for `handle` function with the correct content for the current route: https://reactrouter.com/en/main/hooks/use-matches#breadcrumbs

I think for Hilla the generated `file-routes.json` is responsible for the metadata.

Legioth:
The metadata is known at compile time for Hilla views but it's only known after a round trip for Flow views. (Though there's also the question on how to implement dynamic page titles for Hilla views...)

tltv:
[...]

PR #19329, opened 2024-05-08
fix: set page title to documentTitleSignal

tltv:
`UIInternals#setTitle(String)` sets page title via JavaScript to `document.title` and optionally `window.Vaadin.documentTitleSignal.value` where documentTitleSignal is expected but not limited to be Signal<string> type with a value field. This allows Hilla main layout, when used, being kept in sync with the Flow page title even when set via PageTitle annotation or HasDynamicTitle interface.

Fixes: #19200


mshabarov (approved):
Can confirm that with this patch and [main layout changes](https://github.com/vaadin/flow/issues/19200#issuecomment-2100459743) the Flow page title is shown on the navigation bar in the same way as Hilla shows it.

Issue #19244, opened 2024-04-24
Reload/refresh router content without reloading the full page 


chvndb:
I am using Vaadin 24 with a `RouterLayout` and trying to refresh the router content without refreshing all the other components of the RouterLayout. 

Using `UI.getCurrent().getPage().reload();` does a full page refresh which is not the UX I want. If I use `UI.getCurrent().navigate(...)`, then it simply reuses the previous router component. This means I need to implement logic that checks if it is a refresh or not and update accordingly. 

I want to avoid this overhead and simply have a new instance of the router component so that the content is recreated and acts as a refresh.  I tried manipulating the router through `UI.getCurrent().getInternals().getRouter()`, but could not find a solution.

Is this in any way possible? I can only find issues talking about preserve on refresh.

In the end, I would like something like `UI.getCurrent().getPage().refreshRoute();` or `UI.getCurrent().getRoute().refresh();` that would make sure a new instance of the route component is created and `RouterLayout.showRouterLayoutContent(HasElement content)` is called again.

There should also not be any new entry in the browser session history (History). So that navigating back goes to a previous page and not again to the same page.

chvndb:
I found a workaround. It is a hack as I have to use the `getInternals()` method which is not a stable API and should not really be used as I understood.

I basically clear the `routerTargetChain` in `UIInternals` using a fake location. Then I do the actual navigation using the `NavigationTrigger.PROGRAMMATIC` trigger to avoid a browser history entry.

```
Location location = ...;  // the location corresponding to the route I want to refresh
UI.getCurrent().getInternals().showRouteTarget(new Location("random"), new Div(), null);
UI.getCurrent().getInternals().getRouter().navigate(UI.getCurrent(), location, NavigationTrigger.PROGRAMMATIC);
```

This results in the behaviour that I need. Did I miss something that can do the same, but in a clean way? Or is this a feature that could be added?

mshabarov:
> Or is this a feature that could be added?

Yes, Vaadin Flow team is going to implement this feature soon and hopefully this will target Vaadin 24.5 (or maybe even 24.4) release.

knoobie:
@mshabarov I just came across a use case for this as well. I know it's pretty late for 24.4.. but this looks like a neat and easy addition; some Users and even your Copilot would love to have 😉

mshabarov:
@knoobie good point, let's see how development goes, I'm okay to include it into 24.4 even though it's in beta already. Copilot experience was the main point, indeed.

[...]

Issue #19262, opened 2024-04-26
Refresh the view instead of reloading the page when hotswapping takes place


Artur-:
### Describe your motivation

When a class has been hotswapped, Flow currently refreshes the whole page. This reloads all JS files etc which is typically not needed, causes flickering and takes time. 

### Describe the solution you'd like

Flow could re-render the component tree for the current view instead of reloading. The implementation needed for this sounds like exactly what is requested in #19244 although that is a different use case

tepi:
Needs https://github.com/vaadin/flow/issues/19261 to be implemented first

mcollovati:
Releated PR https://github.com/vaadin/flow/pull/19650

PR #19331, opened 2024-05-08
feat: API for reloading current route target

tepi:
Fixes https://github.com/vaadin/flow/issues/19244

Prepares implementation of https://github.com/vaadin/flow/issues/19262


tepi:
Draft for now - just wanted to see existing tests pass fine with the change. Still needs tests to be added for the refresh functionality.

tepi:
Something to consider is `@PreserveOnRefresh` routes - what should we do with them? This PR basically bypasses them, so even if explicit reload API is called, the cached instance will still be used. We could recreate those as well if it seems appropriate.

mcollovati:
> Something to consider is `@PreserveOnRefresh` routes - what should we do with them? This PR basically bypasses them, so even if explicit reload API is called, the cached instance will still be used. We could recreate those as well if it seems appropriate.

My opinion is that the new API should recreate components even if the route is marked with `@PreserveOnRefresh`. The new method performs an action that is different from a normal page reload or navigation, so if a developer is explicitly calling it, I think it would expect route component(s) to be recreated, even if preserve on refresh is in use.

tepi:
Logic in this PR recreates the whole layout chain, which might not be wanted in all cases. Working on a change to provide means to control this in the API.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -1042,6 +1045,17 @@ public Location getLastHandledLocation() {
      */
     public void setLastHandledNavigation(Location location) {
         lastHandledNavigation = location;
+        if (location != null) {
+            locationForRefresh = location;
+        }
+    }
+
+    public void refreshCurrentRoute(boolean refreshRouteChain) {
+        if (locationForRefresh != null) {
```
mcollovati, line 1054: What about throwing an exception if `locationForRefresh` is null? Or at least logging a debug message so that the developer can understand why nothing happens when the method is called.

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
```diff
@@ -483,7 +490,9 @@ private Optional<Integer> sendBeforeEnterEventAndPopulateChain(
 
         try {
             for (Class<? extends HasElement> elementType : typesChain) {
-                HasElement element = getRouteTarget(elementType, event);
+                HasElement element = getRouteTarget(elementType, event,
+                        typesChain.indexOf(elementType) == typesChain.size()
```
mcollovati, line 494: I wonder if `typesChain.indexOf(elementType)` could get a wrong result if the same class is present multiple times in the chain. This is probably something that will never happen, but, to be sure, what about using and old for loop with index?

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
```diff
@@ -526,14 +535,8 @@ private Optional<Integer> sendBeforeEnterEventToExistingChain(
...
-        Optional<Integer> result = sendBeforeEnterEvent(chainEnterHandlers,
-                event, beforeNavigation, chain);
-
-        if (result.isPresent()) {
-            return result;
-        }
-
-        return Optional.empty();
+        return sendBeforeEnterEvent(chainEnterHandlers, event, beforeNavigation,
+                chain);
```
mcollovati, line 539: Nice catch :smile:

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/RefreshCurrentRouteIT.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private String getString(String id) {
+        waitForElementPresent(By.id(id));
+        return findElement(By.id(id)).getText();
+    }
+
+    private int getInt(String id) {
+        return Integer.parseInt(getString(id));
+    }
+}
```
mcollovati, line 100: nit: new line missing
```suggestion
}

```

flow-server/src/test/java/com/vaadin/flow/router/internal/NavigationStateRendererTest.java
mcollovati, line 1: Do you think we can add some unit tests, for faster feedback?

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
tepi: Added logging

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
tepi: Done

flow-server/src/test/java/com/vaadin/flow/router/internal/NavigationStateRendererTest.java
tepi: Makes sense, will add unit tests

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/RefreshCurrentRouteView.java
```diff
@@ -0,0 +1,102 @@
...
+import com.vaadin.flow.component.UI;
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.router.AfterNavigationEvent;
+import com.vaadin.flow.router.AfterNavigationObserver;
+import com.vaadin.flow.router.BeforeEnterEvent;
+import com.vaadin.flow.router.BeforeEnterObserver;
+import com.vaadin.flow.router.BeforeLeaveEvent;
+import com.vaadin.flow.router.BeforeLeaveObserver;
+import com.vaadin.flow.router.PreserveOnRefresh;
```
mcollovati, line 16: unused import
```suggestion
```

flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1245,6 +1245,19 @@ public void navigate(String locationString,
         }
     }
 
+    /**
+     * Re-navigates to the current route. Also re-instantiates the route target
+     * component, but not if {@link com.vaadin.flow.router.PreserveOnRefresh} is
+     * used.
```
mcollovati, line 1251: We can probably remove the `@PreserveOnRefresh` note now.
Or (maybe better) make it explicit that components are re-instantiated even if the target route is annotated with `@PreserveOnRefresh`

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -1042,6 +1045,29 @@ public Location getLastHandledLocation() {
...
+        if (location != null) {
+            locationForRefresh = location;
+        }
+    }
+
+    /**
+     * Re-navigates to the current route. Also re-instantiates the route target
+     * component, even if {@link com.vaadin.flow.router.PreserveOnRefresh} is
+     * used.
+     *
```
mcollovati, line 1057: See comment about `@PreserveOnRefresh` note on `UI.refreshCurrentRoute` method

tepi: Fixed

flow-server/src/main/java/com/vaadin/flow/component/UI.java
tepi: Fixed

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/RefreshCurrentRouteView.java
tepi: Removed unused import

flow-server/src/test/java/com/vaadin/flow/router/internal/NavigationStateRendererTest.java
tepi: Unit tests added for `@PreserveOnRefresh` case and "normal" case.

mcollovati: Excellent! Thank you!

mcollovati (approved):

mwcodeslave:
in case someone looking for the solution:

`UI.getCurrent().refreshCurrentRoute(false); // refresh AppLayout only`

Issue #19320, opened 2024-05-07
@Menu order should be a double


Legioth:
### Description of the bug

The properties in the `@Menu` annotation should correspond to the `menu` property in Hilla's `ViewConfig` type where `order` is a `number` which corresponds to a `double` in Java. The purpose of allowing floating point values here is to enable fractional indexing so that you can squeeze in a menu entry between 1 and 2 by setting the order to 1.5 rather than having to re-index all existing menu entires.

### Expected behavior

The type of `order` in `@Menu` is `double`

### Versions

- Vaadin / Flow version: 24.4.0.beta1

knoobie:
@Legioth little issue hijack.. 

I was wondering if it would be feasible to also add a parent double param to the annotation and the hilla side pushing the menu annotation from a "just for demo purpose" to an "oh that could fit a lot of use cases". The parent double would match with a corresponding index to find the appropriate view parent while building the menu; allowing lose coupling by an ID (order has to be unique anyway) instead of the hard references like layout = Class we've seen previously in Flow.

Legioth:
We have considered hierarchy support for the automatic menu but it hasn't been high enough priority to be implemented yet. I don't think it's a good idea to speculatively add some specific API before we have built the functionality that uses that API.

What we should maybe do instead is to make sure it's easy for applications to add their own custom properties to the menu entry that they can then use in their own menu rendering logic.

knoobie:
Your second approach sounds also reasonable!

PR #19332, opened 2024-05-08
fix: change type of Menu order to double

tltv:
Changing `Menu#order` type double and `MenuData#order` type to Double to match it with Hilla's `ViewConfig` type `number`.

Fixes: #19320


tltv:
Merging this will break Hilla's `ClientViewMenuConfig` record which has Long type for order, so that has to be changed to Double too. PR is prepared here https://github.com/vaadin/hilla/pull/2401.

mcollovati (approved):

PR #19340, opened 2024-05-08
fix: Fix failure when Vite sends many messages in sequence

Artur-:
Sending multiple messages at the same time resulted in

```
java.lang.IllegalStateException: The remote endpoint was in state [TEXT_FULL_WRITING] which is an invalid state for called method
```

which originates in calling `browserSession.getAsyncRemote().sendText` a second time before the complete callback for the first call has been called.

Now the getBasicRemote() method is used instead which blocks until the message has been sent.

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketProxy.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketProxy.java
@@ -54,15 +54,14 @@ public class ViteWebsocketProxy implements MessageHandler.Whole<String> {
     public ViteWebsocketProxy(Session browserSession, Integer vitePort,
             String vitePath) throws InterruptedException, ExecutionException {
         viteConnection = new ViteWebsocketConnection(vitePort, vitePath,
-                browserSession.getNegotiatedSubprotocol(),
-                msg -> browserSession.getAsyncRemote().sendText(msg, result -> {
-                    if (result.isOK()) {
-                        getLogger().debug("Message sent to browser: {}", msg);
-                    } else {
+                browserSession.getNegotiatedSubprotocol(), msg -> {
+                    try {
+                        browserSession.getBasicRemote().sendText(msg);
+                    } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


mcollovati (approved):

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/viteproxy/ViteWebsocketProxy.java
```diff
@@ -54,15 +54,14 @@ public class ViteWebsocketProxy implements MessageHandler.Whole<String> {
...
-                        getLogger().debug("Message sent to browser: {}", msg);
-                    } else {
+                browserSession.getNegotiatedSubprotocol(), msg -> {
+                    try {
+                        browserSession.getBasicRemote().sendText(msg);
+                    } catch (IOException e) {
                         getLogger().debug("Error sending message to browser",
-                                result.getException());
+                                e);
                     }
```
czp13, line 63: The only side effect of this PR (I just checked because of copilot Slack discussion), that we will not have debug logs anymore for the sent messages. Potentially it could be added back in case of need.

```
browserSession.getNegotiatedSubprotocol(), msg -> {
                    try {
                        browserSession.getBasicRemote().sendText(msg);
                         getLogger().debug("Message sent to browser: {}", msg);
                    } catch (IOException e) {
                        getLogger().debug("Error sending message to browser",
                                result.getException());
                                e);
                    }
```

mcollovati: My bad. I thought it was intentionally removed.

mcollovati: Thanks Peter for pointing out

czp13: Sorry to intervene with your PR review, Marco. I just checked it because we had internal conversations about this issue in the Copilot channel. Thanks for the review and your kindness, as you always show. 🙇 💙

PR #19389, opened 2024-05-15
feat: Filter client routes by authentication and role

caalador:
Add filtering of menu client
routes by checking authentication
and roles.


flow-server/src/main/java/com/vaadin/flow/router/AbstractRouteNotFoundError.java
```diff
@@ -136,14 +137,16 @@ private String getRoutes(BeforeEnterEvent event) {
...
+        routeElements.addAll(getClientRoutes(event));
         return routeElements.stream().map(Element::outerHtml)
                 .collect(Collectors.joining());
     }
 
-    private List<Element> getClientRoutes() {
-        return FrontendUtils.getClientRoutes().stream()
-                .filter(route -> !route.contains("$layout"))
+    private List<Element> getClientRoutes(BeforeEnterEvent event) {
+        return MenuRegistry
```
tltv, line 146: Is `MenuRegistry#getClientRoutes` and `FrontendUtils#getClientRoutes` same thing now? I see `MenuRegistry` something needed for the Menu only and `FrontendUtil`'s method is needed to get all client routes. 
It shouldn't do any access checking in the error view, it should list all routes for the developer to be aware of them. Just like with server routes.

Therefore deprecating `ClientRoutesProvider` is not necessary as it is meant for different use case. I think we need to keep both and keep using `FrontendUtil.getClientRoutes()` here.

caalador: This is the unfiltered client routes. 
`MenuRegistry.getClientRoutes(true, config)` would give filtered results.

The issue with ClientRoutesProvider is that it is provided by Hilla only and the routes are read and compiled from the same data as MenuRegistry uses.

We were looking at this and no one had any usecase for having the provider specifically.

tltv: Ah, it's same then. It's collecting fs-routes but not custom client routes then. And If it's same, then `ClientRoutesProvider` becomes jobless. And I'm still not aware of any new use case for that provider other than adding your custom client routes in the picture by implementing the provider. Not so sure if it's enough to keep the provider, maybe not, as you suggest. 

Calling `MenuRegistry#getClientRoutes(false, config)` is just a bit misleading. I wonder if it could be called from the existing `FrontendUtils.getClientRoutes` where `DeploymentConfiguration` is read from the active VaadinSession.

caalador: changed.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
```diff
@@ -162,16 +172,59 @@ private static Map<String, RouteParamType> getParameters(RouteData route) {
...
+        return collectClientMenuItems(filterClientViews,
+                deploymentConfiguration, vaadinRequest);
+    }
+
+    /**
+     * Get registered client routes. Possible to have all routes or only
+     * accessible routes.
+     *
+     * @param filterClientViews
+     *            {@code true}if inaccessible client routes should be filtered
```
tltv, line 198: Add space between `{@code true}` and `if` or change it to be inline with other methods. Could say "true to filter routes by authentication status" like others.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
```diff
@@ -60,22 +66,26 @@ public class MenuRegistry {
...
-        return new MenuRegistry().getMenuItems();
+        return new MenuRegistry().getMenuItems(true);
     }
 
     /**
      * Collect views with menu annotation for automatic menu population. All
      * client views are collected and any accessible server views.
      *
+     * @param filterClientViews
+     *            {@code true} to filter client views by login information
```
tltv, line 77: Not inline with all other methods. All should say same message. Either this or the other option.

flow-server/src/main/java/com/vaadin/flow/router/internal/ClientRoutesProvider.java
```diff
@@ -19,9 +19,16 @@
 import java.io.Serializable;
 import java.util.List;
 
+import com.vaadin.flow.function.DeploymentConfiguration;
+
 /**
  * Interface for providing client side routes.
+ *
+ * @deprecated Provider is deprecated, use
+ *             {@link com.vaadin.flow.server.menu.MenuRegistry#getClientRoutes(boolean, DeploymentConfiguration)}
```
tltv, line 28: Should link to `FrontendUtils#getClientRoutes` instead now.

tltv (changes requested):

tltv (approved):

PR #19391, opened 2024-05-15
chore: add feature flag for Full-stack Signals

taefi:
## Description

Fixes vaadin/hilla#2430


Legioth (dismissed):

caalador (commented):
Dash in the ID generates a `invalid assignment left-hand side` which seems to be the reason for rollup failing the build.
image

caalador (approved):

PR #19395, opened 2024-05-16
feat: speedup startup time by excluding jars

tltv:
Optimize class scanning of VaadinServletContextInitializer by excluding some commonly used jars and add support for vaadin.blocked-jar boolean property in META-INF/VAADIN/package.properties to exclude whole jar content from the class scanner.

Fixes: #19112


vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
```diff
@@ -174,42 +202,55 @@ protected Set<Resource> doFindAllClassPathResources(String path)
             throws IOException {
         var result = super.doFindAllClassPathResources(path);
         result.removeIf(res -> {
-            cachePackageProperties(res);
-            return isBlockedJar(res);
+            try {
+                String resourcePath = res.getURI().getPath();
+                if (resourcePath != null && DEFAULT_SCAN_NEVER_JAR.stream()
```
mcollovati, line 207: Testing locally, it seems that `resourcePath` is always `null` for JAR entries (e.g. `"URL [jar:file:/home/user/.m2/repository/com/vaadin/open/8.5.0/open-8.5.0.jar!/]").
Some additional considerations:
* if `resourcePath` matches one of the `DEFAULT_SCAN_NEVER_JAR` entries, shouldn't the resource be removed, by returning `true`?
* Wouldn't it make sense to apply the `DEFAULT_SCAN_NEVER_JAR` check only if `resourcePath` is effectively a JAR URL?
* Couldn't the `contains` filter cause a false positive if the `resourcePath` is an absolute path that accidentally contains one of the tokens in the `DEFAULT_SCAN_NEVER_JAR` list (e.g. `/home/michael-jackson-laptop/.m2/repository/my/org/myjar/1.0/myjar-1.0.jar`). Would it be feasible to try to determina the exact JAR file name and check against it (e.g. `jar:file:/home/user/.m2/repository/com/vaadin/open/8.5.0/open-8.5.0.jar!/` -> `open-8.5.0.jar`)

mcollovati (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
```diff
@@ -68,11 +69,32 @@ public class FilterableResourceResolver
...
+            "hibernate-validator", "jboss-logging", "/selenium-", "slf4j",
+            "/spring-", "org/webjars/bowergithub", "snakeyaml",
+
+            "javax.", "jakarta.", "kotlin-",
+
+            "gwt-elemental", "javassist", "javaparser-core",
+            "javaparser-symbol", "oshi-core", "micrometer-", "nimbus-jose-jwt",
+
+            "/hilla-engine-core-", "/hilla-engine-runtime-",
+            "/hilla-parser-jvm-", "/hilla-runtime-plugin-").toList();
```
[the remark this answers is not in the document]
mcollovati: We can probably add `jooq` and `jooq-` to the list, since JOOQ is a quite popular library.
And also some additional Vaadin transitive dependency, such as directory-watcher, classgraph, jsoup, throw-if-servlet3, ph-css, ph-commons, gentyref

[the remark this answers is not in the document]
tltv: Pattern is checked with `contains` function. Concern with these patterns is that someone may add custom-hilla-engine-core.jar, which may or may not have Vaadin classes. Starting with slash ensures that we don't accidentally exclude too much. Some of the names are less risky than others. Your suggestions seems safe except jsoup, which could be `/jsoup`.

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
tltv: Good catch. Yes, I've noticed `.getURI().toPath()` returning null in some cases. it should check also `.getURI().toString()`.

Yes, it should return true instead. How I missed this... well hard-coded jars I tested only manually in one test application. Need to check if this could be verified better with IT too.

Other points are good points too, I'll improve matching to only match with the jar file-name. It's a bit too wide atm. matching with whole path.

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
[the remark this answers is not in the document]
tltv: matching to jar file name only will simplify this.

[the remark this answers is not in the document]
tltv: I had to take that back. We still need the slash in the beginning e.g. for vaadin-spring-*.jar which caused issues as it was blocked.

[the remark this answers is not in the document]
mcollovati: I see

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
tltv: Done

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
```diff
@@ -174,42 +207,73 @@ protected Set<Resource> doFindAllClassPathResources(String path)
             throws IOException {
         var result = super.doFindAllClassPathResources(path);
         result.removeIf(res -> {
-            cachePackageProperties(res);
-            return isBlockedJar(res);
+            try {
+                String jarName = resolveJarName(res.getURI());
+                if (DEFAULT_SCAN_NEVER_JAR.stream()
```
mcollovati, line 212: I wonder if we can avoid this check if the resource is not a JAR and continue with the lambda logic.

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
```diff
@@ -68,11 +71,34 @@ public class FilterableResourceResolver
...
+            "aspectjweaver", "hibernate-core", "hibernate-commons",
+            "hibernate-validator", "jboss-logging", "selenium-", "slf4j",
+            "/spring-", "snakeyaml",
+
+            "javax.", "jakarta.", "kotlin-",
+
+            "gwt-elemental", "javassist", "javaparser-core",
+            "javaparser-symbol", "oshi-core", "micrometer-", "nimbus-jose-jwt",
+            "jooq.", "jooq-", "directory-watcher", "classgraph.", "jsoup.",
+            "throw-if-servlet3", "ph-css", "ph-commons", "gentyref.",
```
mcollovati, line 93: What's the reason for the patterns ending with `.`? I noticed that `gentyref` and `jsoup` are not excluded at runtime, since the JAR name is like `jsoup-1.17.2.jar`.

There are other libraries that can probably be skipped, based on running the test: `asm`, `asm-`, `jetty-`, `tomcat-`, `classmate`, `reflections`, `jna-`, `jsr305`, `jcip-annotations`, `activation`, `httpcore5`, `httpclient5`.

tltv: That was my mistake. Version in the filename breaks it so I'll remove it. These short patterns combined with `contains` are not working so reliably and we can't really add those without changing the check to more specific. regex could be a bit too slow here. But basic startsWith, indexOf and substring would be fast.

Changing `contains` to `startsWith`, and splitting pattern by asterix `*` would let us define `jsoup-*.*.*.jar` pattern that would work better with short filenames. Idea is to use `startsWith` and/or `indexOf` for each part in the splitted array in order, doing substring on the way.  `jsoup-*.*.*.jar` pattern would only match something like `jsoup-XX.YY.ZZ.jar` name.

tltv: done

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
```diff
@@ -113,9 +161,17 @@ protected boolean isJar(String path) {
         return path.lastIndexOf(JAR_KEY) != -1;
     }
 
+    private List<String> getDefaultBlockedJarList() {
+        return customBlockedJarsList != null ? customBlockedJarsList
+                : DEFAULT_SCAN_NEVER_JAR;
```
mcollovati, line 166: Couldn't we always initialize `customBlockedJarsList` in `initCustomBlockedJars` (fallback to `DEFAULT_SCAN_NEVER_JAR`) and simply its value here? Or directly use the field instead of the method?

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
```diff
@@ -174,42 +235,125 @@ protected Set<Resource> doFindAllClassPathResources(String path)
...
+     * "spring-*-*.*.*.jar" pattern matches to "spring-foo-1.0.0.jar",
+     * "spring-foo-bar.1.0.jar" but NOT "spring-foo-bar-1.0.0.jar" or
+     * "spring-1.0.0.jar".<br/>
+     * <br/>
+     * "spring-*_*.*.jar" match "spring-abc.1_0.0.jar" due to the order '*' is a
+     * substring part by part from left to right. <br/>
+     * <br/>
+     * Method is not using much regex to get optimal performance.
+     */
+    private boolean patternMatch(String jarName, String pattern) {
```
mcollovati, line 278: nit: it would be nice to have a unit test for this method

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
tltv: Probably good idea to simplify like that.

vaadin-spring/src/main/java/com/vaadin/flow/spring/io/FilterableResourceResolver.java
tltv: I agree. It turned out into more complex so unit test would make sense.

tltv: Done

mcollovati (approved):
Looks good. Great job @tltv

PR #19406, opened 2024-05-17
fix: support route parameters in menu

### LINKED ISSUE #19392: navigation issue with hybrid application
### Describe the bug

Using `24.4.0.beta2` in a hybrid application causes a blank UI after navigating between different routes
image

### Expected-behavior

UI are loaded correctly. 

### Reproduction

```
download the app using this preset https://start.vaadin.com/dl?&preset=test-hybrid-react&projectName=test-hybrid-react&preset=partial-prerelease
unzip
./mvnw
navigate to localhost:8080 [1]
click Hello Hilla
click Hello Flow [2]
```

- [1] 
image

- [2] the url goes to about:blank#blocked and leave a blank screen
image

### System Info

Vaadin: 24.4.0.beta2

tltv:
Fixes server menu item url for routes with route parameters. Routes with only required parameters are excluded automatically from the menu. Fixes #19392 for 24.5+.

RelatedTo: #19392


flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
```diff
@@ -260,8 +260,7 @@ public List<RouteData> getRegisteredAccessibleMenuRoutes(
 
     private Stream<RouteData> getMenuRouteCandidates() {
         return getRegisteredRoutes().stream()
-                .filter(route -> route.getMenuData() != null)
-                .filter(route -> !route.getMenuData().isExclude());
+                .filter(route -> route.getMenuData() != null);
```
mshabarov, line 263: why `.filter(route -> !route.getMenuData().isExclude());` was removed? This looks suspicious as this flag should be still check, shouldn't it?

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
```diff
@@ -307,6 +306,11 @@ private void populateRegisteredRoutes(ConfiguredRoutes configuration,
         List<Class<? extends RouterLayout>> parentLayouts = getParentLayouts(
                 configuration, template);
 
+        var parameters = configuration.getParameters(template);
+        // exclude route from the menu if it has any required parameters
+        boolean excludeFromMenu = parameters != null && !parameters.isEmpty()
+                && parameters.values().stream()
+                        .anyMatch(param -> !param.getTemplate().contains("?"));
```
caalador, line 313:
```suggestion
                && parameters.values().stream().map(ParameterInfo::new).
                        .anyMatch(param -> !param.isOptional() && !param.isVarargs());
```
Wildcard routes should by definition work with no parameters.
Also we should use the ParameterInfo class instead of a magic char.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
tltv: It's excluded here to let user still navigate to the route with correct parameters. If it's excluded here, Hilla will not add this route in the header script, which breaks the navigation. Hilla hides the excluded menu item.

But now I realize that this would not work for server side menu and we may need another method that excludes the menu item anyway. Or just leave the filtering to the "client" like for Hilla.

In summary, no RouteData means no access at all. `excluded=true` means excluding route from the menu, but route is still accessible via browser's address bar/links.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
```diff
@@ -123,11 +124,36 @@ private static void addMenuRoutes(Map<String, AvailableViewInfo> menuRoutes,
...
+     * @param route
+     *            route to get url for
+     * @return url for the route
+     */
+    private static String getRouteUrl(RouteData route) {
+        if (route.getRouteParameters() != null
+                && !route.getRouteParameters().isEmpty()) {
+            String editUrl = "/" + route.getTemplate();
+            for (RouteParameterData param : route.getRouteParameters().values()
+                    .stream().filter(param -> param.getTemplate().contains("?"))
```
caalador, line 147: This should also probably be for both optional and varargs

Perhaps extending `RouteData` with a method that returns route parameters as List<ParameterInfo> which is never null would make checks simpler as then it would only be a isEmpty check instead of also a null check and the template handling would be cleaner.

Another could be to give RouteParameterData the isOptional and isVarargs methods

caalador (approved):

caalador (approved):

mshabarov (approved):

PR #19430, opened 2024-05-22
chore: Upgrade to Spring-Boot 3.3.0

mshabarov:
Related to https://github.com/vaadin/platform/issues/5251

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@
         <failOnMissingWebXml>false</failOnMissingWebXml>
 
         <!-- Dependencies -->
-        <spring.boot.version>3.2.5</spring.boot.version>
+        <spring.boot.version>3.3.0-RC1</spring.boot.version>
         <gwt.version>2.9.0</gwt.version>
         <hibernate.validator.version>8.0.1.Final</hibernate.validator.version>
         <slf4j.version>2.0.13</slf4j.version>
@@ -172,6 +172,13 @@
                 <enabled>false</enabled>
             </snapshots>
         </repository>
+        <repository>
+            <id>spring-milestones</id>
+            <url>https://repo.spring.io/milestone/</url>
+            <snapshots>
+                <enabled>true</enabled>
+            </snapshots>
+        </repository>
     </repositories>
 
     <pluginRepositories>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -85,7 +85,7 @@
         <failOnMissingWebXml>false</failOnMissingWebXml>
 
         <!-- Dependencies -->
-        <spring.boot.version>3.2.5</spring.boot.version>
+        <spring.boot.version>3.3.0-RC1</spring.boot.version>
```
knoobie, line 88: FYI:

> [SEVERE] http://localhost:8888/my/login/page - Refused to apply style from 'http://localhost:8888/VAADIN/build/indexhtml-BOFKPHa4.css' because its MIME type ('') is not a supported stylesheet MIME type, and strict MIME checking is enabled.

There is a bug in Spring Boot RC1 about Mime Types: https://github.com/spring-projects/spring-boot/issues/40860

Fixed in 3.3.0 to be released today

ZheSun88: seems things are going ok with the final version. @mshabarov

Artur- (approved):

ZheSun88:
do we want to merge this one? and announce that SB 3.3 is the minimal support on V24.5?

knoobie:
> do we want to merge this one? and announce that SB 3.3 is the minimal support on V24.5?

My two cents: 

Vaadin should support IMHO "all" supported Spring Boot Versions; their Support Timeframe is not soo long, giving enterprise people (your target audience) some time to upgrade because some Spring Boot Minor Versions can be nasty.. This should not stop you from already merging SB 3.3 - but there should also be some kind of tests that SB 3.2 is going to work for the next months. 

Example: 
 - Vaadin 24.4: Should work on 3.2 / 3.3 (3.1 is EoL since 2024-05-18)
 - Vaadin 24.5 (released in.. dunno.. August / September?): Should work on 3.2 / 3.3) 
 - Vaadin 24.6 (released in.. dunno.. Dezember?): Should work on 3.3 / 3.4)  (3.2 is EoL since 2024-11-23)

https://spring.io/projects/spring-boot#support

mshabarov:
@knoobie Vaadin usually supports the latest minor version of Spring/Spring-Boot for latest Vaadin versions. Vaadin 24.4 will support SB 3.2 during it's lifespan. We think that this should be enough to users to upgrade to 3.3, which is going to be a minimal support version in Vaadin 24.5. 
We also don't have any plans to intentionally break the support of 3.2, it should work, however, we don't want to continuously test it with latest Vaadin/Flow versions, e.g. 24.5 and further and, in general, we don't want to accept this maintenance burden and test older SB versions with latest Vaadin releases.

Issue #19261, opened 2024-04-26
Integrate hotswap support from the Hotswap Agent plugin


Artur-:
### Describe your motivation

The Hotswap agent plugin at https://github.com/HotswapProjects/HotswapAgent/tree/master/plugin/hotswap-agent-vaadin-plugin implements support for hot reloading of various parts of Flow but it has the issue that it tries to support all Flow versions at the same time, and that it is an external project so it is always lagging behind when Flow gets new features. As it tries to support many Flow versions, the code becomes quite complex.

### Describe the solution you'd like

Implement the same approach already tested in Hilla here https://github.com/vaadin/hilla/blob/main/packages/java/endpoint/src/main/java/com/vaadin/hilla/Hotswapper.java#L34 and here https://github.com/vaadin/hilla/blob/main/packages/java/hilla-dev/src/main/java/com/vaadin/hilla/devmode/hotswapagent/HillaPlugin.java

In other words, make the Hotswap Agent plugin call a Flow method when hotswapping has taken place without doing any filtering or trying to figure out what needs to be done. Let the `onHotswap` method in Flow decide what needs to be done due to reloading of the given class(es).

This would have many benefits:
1. The Hotswap Agent plugin does not need to support multiple Flow versions. It only needs to find `onHotswap` and different Flow versions can implement the method in different ways
2. There is no need for strange reflection code to clear caches etc. The code in `onHotswap` will become much simpler than what is in the hotswap agent plugin
3. The JRebel plugin can use the same method so that hotswapping will work equally well regardless of which product you use

mshabarov:
**Acceptance Criteria**:
- [ ] Flow has an API for updating routes, its implementation does the same as the existing logic does in [hotswap-agent-vaadin-plugin](https://github.com/HotswapProjects/HotswapAgent/blob/master/plugin/hotswap-agent-vaadin-plugin/src/main/java/org/hotswap/agent/plugin/vaadin/VaadinIntegration.java)
- [ ] `hotswap-agent-vaadin-plugin` has no `VaadinIntegration` anymore, but instead it calls a single `onHotswap` method, which is Vaadin's public API, so all the HA related API (annotations, base classes) stays in the HA plugin in the HA project and everything related to Vaadin is moved into Vaadin repos
- [ ] Hilla's HA plugin `HillaPlugin` is removed
- [ ] Platform (or Hilla, should be checked which is better) has one single public method `onHotswap` that is called from `hotswap-agent-vaadin-plugin`.
[...]

PR #19431, opened 2024-05-22
chore: improve route registry update logics

mcollovati:
Part of #19261


flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
```diff
@@ -311,35 +313,104 @@ public static String resolve(VaadinContext context, Class<?> component) {
...
+        // - not annotated classes manually added to the registry:
+        // if the class is modified the route will be removed from the registry
+
+        boolean isSessionRegistry = registry instanceof SessionRouteRegistry;
+
+        // routes for modified classes should be removed
+        // - @Route annotation has been removed
+        // - @Route annotation is present but registerAtStartup=false
+        // for session registries a modified route should never be removed
+        // because we don't know how it has been registered
```
tepi, line 363: Quite a few corner cases here, and some of them seem quite annoying. Is there something we could add e.g. to `RouteData` class or somewhere else to retain information on how the route has been registered?

I think at least any programmatically added routes should remain in the registry unless explicitly removed.

mcollovati: > Is there something we could add e.g. to RouteData class or somewhere else to retain information on how the route has been registered

A was thinking the same. I'll try to see if it is doable without huge changes.

mcollovati: `RouteData` seems not an option, since instance are computed on-demand. So when we need to check the previous attributes, the instances will be created based on the update class version.
I will try to look at `RouteTarget` and `RouteModel`

mcollovati: Added some metadata on `RouteTarget` and implemented a couple of rules to preserve manual registered routes.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteTarget.java
```diff
@@ -94,4 +103,27 @@ public List<Class<? extends RouterLayout>> getParentLayouts() {
         return parentLayouts;
     }
 
+    /**
+     * Gets if the route navigation target is a {@link Route} annotated class or
+     * not.
+     *
+     * @return {@literal true} if the route was registered at application
+     *         startup, otherwise {@literal false}.
```
tepi, line 111: Wrong description here

mcollovati: Good catch, thanks!

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
```diff
@@ -299,6 +304,24 @@ public static String resolve(VaadinContext context, Class<?> component) {
      * Updates route registry as necessary when classes have been added /
      * modified / deleted.
      *
+     * Registry Update rules:
+     * <ul>
+     * <li>a route is preserved if the class does not have a {@link Route}
+     * annotation, but had not it also at registration time</li>
```
tepi, line 310: nit: "a route is preserved if the class does not have a {@link Route} annotation and did not have it at registration time" might be a bit clearer

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
```diff
@@ -299,6 +304,24 @@ public static String resolve(VaadinContext context, Class<?> component) {
...
+     * <li>a route is preserved if the class is annotated with {@link Route} and
+     * registerAtStartup=false and the the flag has not changed</li>
+     * <li>new classes are not automatically added to session registries</li>
+     * <li>existing routes in session registries are not removed in case of
+     * class modification</li>
+     * </ul>
+     *
+     * Not handled cases:
+     * <ul>
+     * <li>Registered route class is not more a Flow Component</li>
```
tepi, line 320: nit: "Registered route class is no longer a Flow Component" might be a bit clearer

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
```diff
@@ -311,35 +334,131 @@ public static String resolve(VaadinContext context, Class<?> component) {
...
+        RouteConfiguration routeConf = RouteConfiguration.forRegistry(registry);
+
+        // remove potential routes for classes that are no more components
+        // not all agents/JVMs support reload on class hierarchy change
+        deletedClasses.stream()
+                .filter(clazz -> !Component.class.isAssignableFrom(clazz))
+                .forEach(clazz -> routeConf.removeRoute((Class) clazz));
+        modifiedClasses.stream()
+                .filter(clazz -> !Component.class.isAssignableFrom(clazz))
+                .forEach(clazz -> routeConf.removeRoute((Class) clazz));
```
tepi, line 373: Javadoc of this method says:

```
     * Not handled cases:
     * <ul>
     * <li>Registered route class is not more a Flow Component</li>
```

but seems such case is handled here

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
mcollovati: I removed the "not handled cases" section, as it is outdated.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
mcollovati: Remove the note from Javadoc. Also move the `routeConf.removeRoute` calls into the `registry.update(...)` block.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
mcollovati: It looks much better, thanks

mcollovati: updated

tepi (approved):

PR #19434, opened 2024-05-23
feat: Flow components as children for ReactAdapterComponent

### LINKED ISSUE #19379: Support content elements in React adapter
Add support for adding Flow components as children to ReactAdapterComponent.

Implementation idea:

- add `useContent(name: string)` ReactAdapterElement hook, supports rendering a DOM placeholder in React
- add the corresponding `protected Element getContentElement(String name)` Java ReactAdapterComponent method, that creates and returns the named container element.

Prototype: https://github.com/vaadin/flow-hilla-hybrid-example/tree/proto/flow-react-children

The first implementation could use two DOM wrapper elements, each with style "display: contents;".

caalador:
Add support for having Flow Components
as children in a ReactAdapoterComponent

Fixes #19379


flow-react/src/main/java/com/vaadin/flow/component/react/ReactAdapterComponent.java
```diff
@@ -49,6 +56,8 @@
  * @since 24.4
  */
 public abstract class ReactAdapterComponent extends Component {
+    private Map<String, Element> contentMap = new HashMap<>();
```
knoobie, line 59: Nit: Micro optimization: should this be null by default and only created if needed? I would think not so many react components going to have flow children

flow-client/src/main/java/com/vaadin/client/ReactUtils.java
```diff
@@ -0,0 +1,63 @@
...
+     * @param element
+     *            react component element
+     * @param runnable
+     *            callback function runnable
+     */
+    public static native void addReadyCallback(Element element,
+            Runnable runnable)
+    /*-{
+            if(element.addReadyCallback){
+                element.addReadyCallback(
```
Legioth, line 45: No `$entry` needed here?

flow-client/src/main/java/com/vaadin/client/ReactUtils.java
```diff
@@ -0,0 +1,63 @@
...
+              function() {
```
Legioth, line 46: Is the wrapping `function` necessary or could this be simplified as `element.addReadyCallback(runnable.@java.lang.Runnable::run(*).bind(runnable))``

flow-client/src/main/java/com/vaadin/client/ReactUtils.java
caalador: I guess it should. Added.

flow-client/src/main/java/com/vaadin/client/ReactUtils.java
caalador: :+1: removed extra function

flow-react/src/main/java/com/vaadin/flow/component/react/ReactAdapterComponent.java
caalador: done.

tltv (dismissed):
LGTM. Works as expected. Manually tested even ReactAdapterComponent inside a ReactAdapterComponent case works.

flow-react/src/main/resources/com/vaadin/flow/server/frontend/ReactAdapter.template
```diff
@@ -118,6 +129,10 @@ export abstract class ReactAdapterElement extends HTMLElement {
         this.#maybeRenderRoot();
     }
 
+    public addReadyCallback(id: string, readyCallback: ReadyCallbackFunction) {
```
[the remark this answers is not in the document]
caalador: done

platosha (approved):

Issue #19333, opened 2024-05-08
Add an alternative API for getTranslation


tomivirkki:
### Describe your motivation

The current `getTranslation` API is convenient when [used inside a parent layout](https://vaadin.com/docs/latest/advanced/i18n-localization#using-localization-in-application) because it's inherited from `Component`. However, if you need to get a translation for a key somewhere where this API isn't inherited there aren't convenient options available. One such case is a static factory method for creating components:

```java
private static TextField createField() {
    return new TextField("Your name");
}
```

In this scenario, you could access the `getTranslation` API, for example, via `VaadinService.getCurrent().getInstantiator().getI18NProvider()` but a helper method would be required to avoid verbosity.

Another approach could be not setting the field's label in the constructor and using the instance itself to access `getTranslation`.

```java
var field = new TextField();
field.setLabel(field.getTranslation("yourname.label"));
```

However, this method requires referencing a component instance, even though obtaining a localized string isn't a task inherently tied to a specific instance.

### Describe the solution you'd like

Introduce a method for obtaining translations through an API not bound to a component instance but available universally. 

I don't have a well-thought-out API suggestion at this time, but perhaps a static method would work here, so the usage could be similar to that of `@vaadin/hilla-react-i18n`:

```java
import static com.vaadin.flow.i18n.I18NProvider.translate;

...

private static TextField createField() {
    return new TextField(translate("yourname.label"));
}
```

### Additional context

The absence of such an API poses challenges with the Copilot i18n plugin, which automatically transforms static strings to `getTranslate` calls. Since it's not context-aware, in the worst-case scenario, it might produce a view class that doesn't compile. See [the workaround under consideration](https://github.com/vaadin/copilot-internal/issues/1652) for more details.

knoobie:
Alternative suggestion: Add a constructor/methods that takes in some kind of custom class e.g. "Translateable(String template, Object...)" so that the components can translate it on demand. 

```java

private static TextField createField(User user) {
    return new TextField(Translatable.of("yourname.label", user.getName()));
}

```

tltv:
I suggest to add two new static methods in `I18NProvider` interface:

```
static String translate(String key, Object... params)
static String translate(String key, Locale locale, Object... params)
```
These methods would only work within an active Vaadin service. Without it, these would either throw exception or return empty String.

[...]

PR #19461, opened 2024-05-27
feat: add alternative translate methods to I18NProvider

tltv:
Adds static `I18NProvider#translate` methods for alternative way to get translation. Compares to calling getTranslation via `VaadinService.getCurrent().getInstantiator().getI18NProvider()`. `translate` methods throws IllegalStateException without active VaadinService.

Fixes: #19333

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/i18n/I18NProvider.java
+++ b/flow-server/src/main/java/com/vaadin/flow/i18n/I18NProvider.java
@@ -19,6 +19,8 @@ import java.io.Serializable;
 import java.util.List;
 import java.util.Locale;
 
+import com.vaadin.flow.internal.LocaleUtil;
+
 /**
  * I18N provider interface for internationalization usage.
  *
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/i18n/I18NProvider.java
```diff
@@ -67,4 +69,43 @@ public interface I18NProvider extends Serializable {
...
+     * @param key
+     *            translation key
+     * @param params
+     *            parameters used in translation string
+     * @return translation for key if found
+     * @throws IllegalStateException
+     *             thrown if no I18NProvider found from the VaadinService
+     */
+    static String translate(String key, Object... params) {
+        return translate(LocaleUtil.getLocale(LocaleUtil::getI18NProvider), key,
```
mcollovati, line 87: nit: `LocaleUtil.getLocale(LocaleUtil::getI18NProvider)` looks a bit weird because of `LocaleUtil` used for both getting the locale and the potential provider. To reduce boilerplate, couldn't we add a `LocaleUtil.getLocale()` method, that does not require arguments? 

```java
    public static Locale getLocale() {
        return getLocale(LocaleUtil::getI18NProvider);
    }
```

flow-server/src/main/java/com/vaadin/flow/i18n/I18NProvider.java
```diff
@@ -67,4 +69,43 @@ public interface I18NProvider extends Serializable {
...
+     * @param params
+     *            parameters used in translation string
+     * @return translation for key if found
+     * @throws IllegalStateException
+     *             thrown if no I18NProvider found from the VaadinService
+     */
+    static String translate(Locale locale, String key, Object... params) {
+        return LocaleUtil.getI18NProvider()
+                .orElseThrow(() -> new IllegalStateException(
+                        "No I18NProvider found from the VaadinService"))
```
mcollovati, line 108: This error message could be misleading in some cases. For example, if the method is called outside a Vaadin request, the real problem is that the `VaadinService` is not available, rather than the i18n provider not being found. We can maybe provide a specific error or have a generic message telling all potential culprits for the failure.

flow-server/src/test/java/com/vaadin/flow/i18n/DefaultInstantiatorI18NTest.java
```diff
@@ -78,6 +78,7 @@ public void translationFileOnClasspath_instantiateDefaultI18N()
 
         VaadinService service = Mockito.mock(VaadinService.class);
         mockLookup(service);
+        VaadinService.setCurrent(service);
```
mcollovati, line 81: It would be good to clear the `VaadinService` thread local in an `@After` method, to prevent potential unexpected issues

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/i18n/I18NProvider.java
tltv: It can't be specific due to getI18NProvider is safe to call. If it returns empty Optional, we just know that VaadinService, initializer or I18NProvider is null. Probably message should say so.

mcollovati (approved):

Issue #19447, opened 2024-05-24
With hybrid app onDetach method is not called when navigating from Flow to Hill view


krissvaa:
### Description of the bug

When using Flow with Hilla, Flow view's detach event doesn't happen when navigating to Hilla view. 
After navigating to some other Flow view the onDetach method is called. 

### Expected behavior

onDetach should be called when navigating to Hilla view

### Minimal reproducible example

Create a new project with 3 views (Hilla view and 2 Flow views) from starter.vaadin.com
Add onDetach() method on one of the Flow views. 
Test:
open the Flow view with onDetach method:
navigate to the Hilla view  - onDetach() not called
navigate to other Flow view - finally onDetach() is called

Also can be tested with: https://github.com/vaadin/hackathon-24-4/tree/kriss
Open Data-Grid: http://localhost:8080/data-grid 
Navigate to  HillaHello http://localhost:8080/ - onDetach() not called
Navigate to Dashboard   - finally onDetach() is called

### Versions

- Vaadin / Flow version: 24.4.0.beta5
- Java version: 21
- OS version: Windows

PR #19462, opened 2024-05-27
fix: fix navigation between flow and react views

caalador:
fixes https://github.com/vaadin/flow/issues/19447

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -150,6 +150,7 @@ let lastNavigation: string;
 let lastNavigationSearch: string;
 let prevNavigation: string;
 let popstateListener: { type: string, listener: EventListener, useCapture: boolean };
+let beforeEnterCalled: boolean = false;
 
 // @ts-ignore
 function navigateEventHandler(event) {
@@ -170,7 +171,7 @@ function navigateEventHandler(event) {
     // target path. Server will then handle updates and postpone as needed.
     // @ts-ignore
     if(matched && matched.filter(path => path.route?.element?.type?.name === Flow.name).length >= 1) {
-        if (mountedContainer?.onBeforeEnter) {
+        if (mountedContainer?.onBeforeEnter && !beforeEnterCalled) {
             // onBeforeEvent call will handle the Flow navigation
             mountedContainer.onBeforeEnter(
                 {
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -288,35 +288,42 @@ function Flow() {
...
+                            prevent() {
+                            },
+                            continue() {
+                                window.document.removeEventListener('click', vaadinRouterGlobalClickHandler);
+                                window.removeEventListener('vaadin-router-go', navigateEventHandler);
+                                if (popstateListener) {
+                                    window.removeEventListener('popstate', popstateHandler);
+                                    window.addEventListener('popstate', popstateListener.listener, popstateListener.useCapture);
+                                }
+                                mountedContainer?.parentNode?.removeChild(mountedContainer);
```
tltv, line 313: This looks like it happens a bit too late causing flicker. How about moving `removeChild` calls to the beginning before `onBeforeLeave`, and add it back with `appendChild` in `prevent()` function if it's called.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -288,35 +288,42 @@ function Flow() {
                 outlet.append(container);
             }
             mountedContainer = container;
-            if (container.onBeforeEnter) {
```
tltv, line 291: Why it's necessary to remove `onBeforeEnter` call here? Could cause some tests fail.

caalador: That is because at the current version there is a onNavigateEvent (probably from the mouse click? It seems that the click event is still propagated even as the listener is added during the useEffect.) so that this is actually not executed, but the onBeforeEnter on line 185 is executed 2 times. 
As we can only have one onBeforeEnter callback so the later one overrides the first.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -292,31 +293,57 @@ function Flow() {
...
             // view from the dom as we will not be getting a uidl response.
             // @ts-ignore
-            if(matched && matched.filter(path => path.route?.element?.type?.name === Flow.name).length == 0) {
-                mountedContainer?.parentNode?.removeChild(mountedContainer);
-                mountedContainer = undefined;
+            if (matched && matched.filter(path => path.route?.element?.type?.name === Flow.name).length == 0) {
+                if (mountedContainer?.onBeforeLeave) {
+                    mountedContainer?.onBeforeLeave(
+                        {pathname, search}, {
+                            prevent() {
```
mshabarov, line 324: Postpone feature seems a bit broken: if I have a confirmation dialog like this https://vaadin.com/docs/latest/routing/lifecycle#postpone, I see both Flow and Hilla contents when trying navigate outside of Flow view.

Screenshot 2024-05-28 at 9 44 42

mshabarov (changes requested):

PR #19488, opened 2024-05-31
feat:Improve Binder change detection

### LINKED ISSUE #19260: Binder: Improve change detection
### Describe your motivation

Currently, when the value of a bound field changes, the binding is considered as "changed" and is included in the `changedBindings`, regardless of whether the user reverts to the initial value. For example, if the value of a name field is changed from "foo" to "foo bar" and then reverted to "foo", the binding still appears in `changedBindings`, causing `binder.hasChanges()` to return true.

It's a common use case to enable the save button only if there is any changes in the form. This way it would be possible to enable the save button only if there are "real" changes.

A potential solution is to store the initial value of each binding when the `readBean` method is invoked. Subsequently, after each value change, this initial value can be compared to the new value. If the new value matches the initial one, the binding would not be marked as "changed," thereby ensuring that changedBindings only includes entries where the current value differs from the initial value. 

An alternative solution could involve extending this mechanism to maintain a list of all changes, which would facilitate an "undo" feature. By tracking each change as an entry in a history list, users can easily revert the most recent modifications or even all changes back to the original state.

onuridrisoglu:
## Description

Fixes #19260

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -1929,18 +1931,41 @@ public static <BEAN> Binder<BEAN> withPropertySet(
      *
      * @param binding
      *            the binding whose value has been changed
+     * @param event
+     *            the value change event
      */
-    protected void handleFieldValueChange(Binding<BEAN, ?> binding) {
-        changedBindings.add(binding);
-
+    protected void handleFieldValueChange(Binding<BEAN, ?> binding, ValueChangeEvent<?> event) {
```
tepi, line 1937: Please avoid changing this method signature since it would be a breaking change and require waiting for the next major. Instead, one option could be to deprecate the current one, and if that is used, just skip the new change detection logic and function as before.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -1929,18 +1931,41 @@ public static <BEAN> Binder<BEAN> withPropertySet(
...
+     * reverted to the initial value in bean
+     * 
+     * @param binding
+     *            the binding whose value has been changed
+     * @param newValue
+     *            the new value if the field
+     * @return true if the field value is reverted
+     */
+    protected boolean isRevertedToInitialValue(Binding<BEAN, ?> binding, Object newValue) {
+		return Objects.equals(bindingInitialValuesMap.get(binding), newValue);
```
tepi, line 1965: The logic here relies on `equals` of the value type being implemented properly. That is something the Binder has not required before, so it's potentially a breaking change.

tepi:
@mshabarov This change will introduce a slight change in behavior of Binder change tracking due to requiring `equals` to be implemented properly by value types. As well as reverting the tracked change if the value is reverted, which is not currently done (that is, the binding is considered as changed anyway). Do you think we'll need an enable/disable switch for this new change tracking?

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
tepi: Also, it migh make more sense to add Binding-level API for overriding the logic for checking if value is reverted. That way if custom logic is needed, one does not need to extend Binder to do it.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -949,6 +952,18 @@ BindingBuilder<BEAN, TARGET> asRequired(
...
+         * Sets the {@code equalityPredicate} used to compare the current value of a field with its initial value.
+         * <p>
+         * The default implementation uses {@link Objects#equals(Object, Object)}, but a custom equality predicate
+         * can be provided for cases where the value type does not support the standard {@code equals} method.
+         * </p>
+         * 
+         * @param equalityPredicate the predicate to use for equality comparison
+         * @return this {@code BindingBuilder}, for method chaining
+         */
+        public BindingBuilder<BEAN, TARGET> withEqualityPredicate(SerializableBiPredicate<Object, Object> equalityPredicate);
```
knoobie, line 966: Vote 4 default implementation; otherwise this is a breaking change

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
onuridrisoglu: Added the current `handleFieldValueChange` back as deprecated

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
onuridrisoglu: introduced an equality predicate (I'm open for naming suggestions) that uses `Objects::equals` as default, but it can be overridden in binding-level

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
onuridrisoglu: Done

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -1810,6 +1850,7 @@ void setIdentity() {
 
     private BindingExceptionHandler exceptionHandler = new DefaultBindingExceptionHandler();
 
+    private Map<Binding<BEAN, ?>, Object> bindingInitialValuesMap = new HashMap<>();
```
mshabarov, line 1853: What if the initial values would be a class field in the `BindingImpl` instead and have a getter method in `Binding`? Then you can call newly added method `Binding::getInitialValue`.

mshabarov (changes requested):

mshabarov:
> @mshabarov This change will introduce a slight change in behavior of Binder change tracking due to requiring `equals` to be implemented properly by value types. As well as reverting the tracked change if the value is reverted, which is not currently done (that is, the binding is considered as changed anyway). Do you think we'll need an enable/disable switch for this new change tracking?

I'd propose to set the equality predicate `null` by default (or some other value that marks it not defined), that means the change detection feature is not enabled by default. Then if this predicate is set by `BindingBuilder::withEqualityPredicate`, then this feature is enabled. In this way this would be backwards compatible.

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -320,6 +321,8 @@ default BindingValidationStatus<TARGET> validate() {
          *         changes, otherwise {@literal false}.
          */
         boolean hasChanges();
+        
+        SerializableBiPredicate<Object, Object> getEqualityPredicate();
```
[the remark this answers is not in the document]
onuridrisoglu: done

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
onuridrisoglu: I refactored the changes and removed `bindingInitialValuesMap`. Now the initial value is kept in Binding level which is set during readBean.

onuridrisoglu:
> > @mshabarov This change will introduce a slight change in behavior of Binder change tracking due to requiring `equals` to be implemented properly by value types. As well as reverting the tracked change if the value is reverted, which is not currently done (that is, the binding is considered as changed anyway). Do you think we'll need an enable/disable switch for this new change tracking?
> 
> I'd propose to set the equality predicate `null` by default (or some other value that marks it not defined), that means the change detection feature is not enabled by default. Then if this predicate is set by `BindingBuilder::withEqualityPredicate`, then this feature is enabled. In this way this would be backwards compatible.

I did the changes, however my only concern is now `withEqualityPredicate ` should be called for each binding in order to activate this feature.

mshabarov (approved):
Looks good to me, but I haven't tested it.

I agree with the concern about having this API only on a binding level, would be good to allow doing the same for the whole form.

mshabarov:
Just discussed with @tepi, the Binder-level API could be done in a way:
1. Adding a method on Binder's level that enables/disables usage of these predicates on bindings level.
2. Then, if enabled, Binder will use a configured predicate for a binding, if it exists, otherwise it uses `equals` as a fallback. If disabled, nothing is used (default).

mshabarov (changes requested):
Setting to request changes

tepi:
> Just discussed with @tepi, the Binder-level API could be done in a way:
> 
>     1. Adding a method on Binder's level that enables/disables usage of these predicates on bindings level.
> 
>     2. Then, if enabled, Binder will use a configured predicate for a binding, if it exists, otherwise it uses `equals` as a fallback. If disabled, nothing is used (default).

@onuridrisoglu fyi, I'll pick this up and do the changes

mshabarov (approved):

PR #19505, opened 2024-06-04
fix: do not use popstate event

### LINKED ISSUE #19513: Additional navigation events caused by firing popstate events
### Description of the bug

Flow fires extra history event, which triggers extra navigation event.
Additionally, React Router complains about [popstate event being fired by Flow](https://github.com/vaadin/flow/blob/f4bf965a54f5fa4c77d54c560f69f6ac95bf715f/flow-server/src/main/java/com/vaadin/flow/component/page/History.java#L189C9-L191C18) :

```
You are trying to use a blocker on a POP navigation to a location that was not created by @remix-run/router. This will fail silently in production. This can happen if you are navigating outside the router via `window.history.pushState`/`window.location.hash` instead of using router navigation APIs.  This can also happen if you are using createHashRouter and the user manually changes the URL.
```

which in turn causes the blocker function in Flow.tsx not to be used in production, so it will be unable to block navigation away (e.g. for postpone feature) or call beforeLeave.

### Expected behavior

No React Router warnings, no extra history event and postpone works.

Extra event can be reproduced with `NavigationTriggerIT`.
Postpone problem can be reproduced with [Navigation postpone example](https://vaadin.com/docs/latest/routing/lifecycle#postpone) and Flow-Hilla hybrid example application.

### Versions

- Vaadin / Flow version: 24.4.0.beta5

caalador:
When pushing or replacing history
send a custom event 'vaadin-navigate'
instead of popstate event.

Fixes #19513

### DIFF
--- a/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/PopStateHandlerIT.java
+++ b/flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/PopStateHandlerIT.java
@@ -25,12 +25,10 @@ public class PopStateHandlerIT extends ChromeBrowserTest {
         pushState(FORUM);
 
         verifyInsideServletLocation(FORUM);
-//        verifyNoServerVisit();
 
         pushState(ANOTHER_PATH);
 
         verifyInsideServletLocation(ANOTHER_PATH);
-//        verifyNoServerVisit();
 
         goBack();
     }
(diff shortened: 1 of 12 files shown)


Artur-:
> React Router would be more happy, if the event is wired through navigate

Where "more happy" means that back and forward navigation would work. Without that, we would need to preserve the `idx` value in the state, which sounds quite hacky

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/PopStateHandlerIT.java
```diff
@@ -62,23 +60,20 @@ public void testSamePathHashChanges_noServerSideEvent() {
...
 
         verifyInsideServletLocation(FORUM_SUBCATEGORY2);
-        verifyNoServerVisit();
 
         goBack();
     }
 
     @Test
-    @Ignore("Ignored because of the issue https://github.com/vaadin/flow/issues/10825")
+//    @Ignore("Ignored because of the issue https://github.com/vaadin/flow/issues/10825")
```
Artur-, line 76:
```suggestion
```

caalador: done

tltv (commented):
Just tested that this fixes also all the cases fixed in [other PR](https://github.com/vaadin/flow/pull/19507). So that one can be closed.

platosha (commented):
LGTM

platosha (approved):

mshabarov (approved):

PR #19520, opened 2024-06-05
fix: Add usage stats for react adapter

mshabarov:
this was removed in https://github.com/vaadin/flow/pull/19434 (maybe accidentally? )


caalador (approved):
Seems like a merge mistake that was not noted.

Issue #19525, opened 2024-06-06
[PiT 24.4.0 RC1] business-app-starter-flow is broken with V24.4.0.rc1


ZheSun88:
### Description of the bug

use the attached app,  there are infos missing from the view after checking the account details
![business-app-24-4-rc1](https://github.com/vaadin/flow/assets/31067185/1f6cb539-b99d-40af-90f9-c90d3da6cc0c)

(an observation is that : there is a blink after single click the data in the account view, and the data is loaded in the accountDetails view, then another event triggers a reload, data got missing)

### Expected behavior

data are loaded correctly without blinking.

### Minimal reproducible example

use the project attached

```
mvn -Pproduction
navigate to localhost:8080 and check the accounts
click the first row of data
there is a blink then the info of availability  and account info  are missing the next view.
refresh the page
info got loaded correctly.
```

[business-app-starter-flow.zip](https://github.com/user-attachments/files/15630465/business-app-starter-flow.zip)

### Versions

- Vaadin / Flow version: 24.4.0.rc1

PR #19539, opened 2024-06-07
fix: preserve navigated.current value

tltv:
Fixes: #19525


caalador (changes requested):
Seems ok.
HistoryIT needs to be updated as the asdf pushState is not a history change event.

caalador (approved):

Issue #19527, opened 2024-06-06
Include `.scss` and `.sass` in `stats.json` hashes


oliveryasuna:
### Describe your motivation

We use SASS (`.scss`) stylesheets under `/frontend/styles`. These are imported and fully supported by Vite. It seems that the `vaadin:stats` Vite plugin does not generate hashes for them and store in `stats.json#frontendHashes`. Because of this, even with `prod.bundle`, the Vaadin Maven/Gradle plugins think there are new files, and thus, do not use the existing bundle.

### Describe the solution you'd like

`const projectFileExtensions = ['.js', '.js.map', '.ts', '.ts.map', '.tsx', '.tsx.map', '.css', '.css.map'];` needs to also support `.scss` and `.sass`.

Is there a reason those file extensions are not already included? If not, happy to contribute this simple PR.

### Describe alternatives you've considered

I could write a Vite plugin that computes and adds the hashes.

I could create a Gradle task that compiles the SASS files and outputs them to `/frontend/styles`.

Both options seem messy.

### Additional context

I discovered this while investigating ways to speed up our builds, in lieu of Vaadin's proper Gradle caching support. https://github.com/vaadin/flow/issues/17941

This idea came from our recent endeavour to speed up builds. Relevant issues:
* https://github.com/vaadin/flow/issues/17941
* https://github.com/gradle/gradle/issues/27696
* https://github.com/vaadin/flow/issues/19527
* https://github.com/vaadin/flow/issues/19528
* https://github.com/vaadin/docs/issues/3477
* https://github.com/vaadin/flow/issues/19543

steve-todorov:
In addition to `scss` and `sass` extensions, people might also use `less` so might be beneficial to add it to the extension list as well.

oliveryasuna:
Maybe extensions should be customizable with an include/exclude pattern.

oliveryasuna:
I opened a draft PR to demonstrate my proposal: https://github.com/vaadin/flow/pull/19550

PR #19550, opened 2024-06-11
Added support to customizable the `projectFileExtensions`.

oliveryasuna:
## Description

Please see the issue.

Fixes https://github.com/vaadin/flow/issues/19527

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [ ] I have added tests to ensure my change is effective and works as intended. _NOT APPLICABLE?_


caalador:
The dev-bundle-plugin also needs to implement the new `projectFileExtensions()` method
```
Error:  /home/runner/work/flow/flow/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java:[80,8] com.vaadin.flow.plugin.maven.BuildDevBundleMojo is not abstract and does not override abstract method projectFileExtensions() in com.vaadin.flow.plugin.base.PluginAdapterBase
```

Also it feels like the extensions should be an addition to the existing and not replacing the list of extensions.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
```diff
@@ -127,7 +127,11 @@ private void createGeneratedConfig() throws IOException {
                 .replace("#webComponentTags#",
                         webComponentTags == null || webComponentTags.isEmpty()
                                 ? ""
-                                : String.join(";", webComponentTags));
+                                : String.join(";", webComponentTags))
+                .replace("#projectFileExtensions#",
+                        options.getProjectFileExtensions().stream()
+                                .map(ext -> "'" + ext + "'")
```
mcollovati, line 133: Since the values potentially come from the outside, it would be good to escape them

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
```diff
@@ -127,7 +127,11 @@ private void createGeneratedConfig() throws IOException {
...
+                                .reduce((a, b) -> a + ", " + b).orElse(""));
```
mcollovati, line 134: nit: a different option is to use the `joining` collector instead of `reduce().orElse`, e.g. `collect(Collectors.joining(", "))`

Up to you to apply the suggestion or leave the original code as is.

mcollovati (changes requested):
A couple of comments:
* I agree with @caalador that the configurable extensions should be added to the default list and should not replace them.
* it would be better to define the default extensions in a single place instead of having duplication in Maven and Gradle plugin. `FrontendUtils` class could be a good host.
* I know that in the Vite configuration file, the constant is named `projectFileExtensions`, but IMO it doesn't sound like a good name for a plugin setting. What about `frontendFileExtensions`? Could it be a better name?

oliveryasuna:
> A couple of comments:
> 
>     * I agree with @caalador that the configurable extensions should be added to the default list and should not replace them.
> 
>     * it would be better to define the default extensions in a single place instead of having duplication in Maven and Gradle plugin. `FrontendUtils` class could be a good host.
> 
>     * I know that in the Vite configuration file, the constant is named `projectFileExtensions`, but IMO it doesn't sound like a good name for a plugin setting. What about `frontendFileExtensions`? Could it be a better name?

I agree with everything and will make changes later. Thanks for considering this!

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
```diff
@@ -127,7 +131,17 @@ private void createGeneratedConfig() throws IOException {
...
                         webComponentTags == null || webComponentTags.isEmpty()
                                 ? ""
-                                : String.join(";", webComponentTags));
+                                : String.join(";", webComponentTags))
+                .replace("#extraProjectFileExtensions#", Optional
+                        .ofNullable(options.getExtraProjectFileExtensions())
+                        .orElse(Collections.emptyList()).stream().map(ext -> {
+                            try {
+                                return "'" + StringEscapeUtils.escapeJava(ext)
+                                        + "'";
```
mcollovati, line 140: It looks like to me that `StringEscapeUtils.escapeJava` does not escape single quotes, that is the only replacement that we need (e.g. `ext.replace("'", "\\'")`).

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
```diff
@@ -127,7 +131,17 @@ private void createGeneratedConfig() throws IOException {
...
+                            } catch (Exception e) {
+                                throw new RuntimeException(e);
+                            }
+                        }).collect(Collectors.joining(", ")));
```
mcollovati, line 144: The result of this string seems not correct; it will be appended to the existing list without a coma.

```
const projectFileExtensions = ['.js', '.js.map', '.ts', '.ts.map', '.tsx', '.tsx.map', '.css', '.css.map''.svg', '.ico'];
```


An option could be the following, for example
```
        String extraExtension = "";
        if (options.getExtraProjectFileExtensions() != null
                && !options.getExtraProjectFileExtensions().isEmpty()) {
            extraExtension = 
                    Optional.ofNullable(options.getExtraProjectFileExtensions())
                            .orElse(Collections.emptyList()).stream()
                            .map(ext -> ext.replace("'", "\\'"))
                            .collect(Collectors.joining("', '", ", '", "'"));
        }
        template = template.replace("#extraProjectFileExtensions#", extraExtension);
```

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
```diff
@@ -278,6 +278,13 @@ public abstract class VaadinFlowPluginExtension @Inject constructor(private val
 
     public abstract val cleanFrontendFiles: Property<Boolean>
 
+    /**
+     * The list of extra file extensions that are considered project files.
+     * Hashes are calculated for these files as part of detecting if a new prod
+     * bundle should be generated.
+     */
+    public abstract val extraProjectFileExtensions: ListProperty<String>
```
mcollovati, line 286: I'm still not convinced by the setting name. It still seems too generic to me.
I mean, the purpose is to add extension for **frontend** files that should be considered, not overall project files.
I propose something like `extraFrontendFileExtensions`.

@mshabarov what do you think?

mcollovati (changes requested):
The implementation seems broken because of a missing comma in placeholder replacement. 
I also suggest adding a couple of tests in `TaskUpdateViteTest` like the following one

```java
    @Test
    public void generatedTemplate_extraFrontendExtension_addedToViteConfiguration()
            throws IOException {
        options.withExtraProjectFileExtensions(List.of(".svg", ".ico"));
        TaskUpdateVite task = new TaskUpdateVite(options, null);
        task.execute();

        File configFile = new File(temporaryFolder.getRoot(),
                FrontendUtils.VITE_GENERATED_CONFIG);

        String template = IOUtils.toString(configFile.toURI(),
                StandardCharsets.UTF_8);

        Assert.assertTrue(
                "Extra frontend extensions should be added to vite configuration, but was not.",
                template.contains(
                        "const projectFileExtensions = ['.js', '.js.map', '.ts', '.ts.map', '.tsx', '.tsx.map', '.css', '.css.map', '.svg', '.ico'];"));
    }

```

mcollovati (commented):
Additional note: it would probably make sense that configuration setting does not require the extension to start with a dot. I would prefer a configuration like 

```xml
<extraProjectFileExtensions>
   <extraProjectFileExtension>svg</extraProjectFileExtension>
   <extraProjectFileExtension>ico</extraProjectFileExtension>
</extraProjectFileExtensions>
```
rather than 

```xml
<extraProjectFileExtensions>
   <extraProjectFileExtension>.svg</extraProjectFileExtension>
   <extraProjectFileExtension>.ico</extraProjectFileExtension>
</extraProjectFileExtensions>
```

It would be better if the dot would be automatically added at runtime, if not already defined in configuration.

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/PluginAdapterBase.java
```diff
@@ -326,4 +326,11 @@ default Lookup createLookup(ClassFinder classFinder) {
      *         router and excluding React dependencies
      */
     boolean isReactEnabled();
+
+    /**
+     * Get the list of project file extensions.
+     *
+     * @return list of project file extensions
```
mcollovati, line 333: We should add a note on the expected extension format here (e.g. no initial dot) with a simple example list.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java
```diff
@@ -232,6 +233,9 @@ public abstract class FlowModeAbstractMojo extends AbstractMojo
     @Parameter(property = InitParameters.REACT_ENABLE, defaultValue = "${null}")
     private Boolean reactEnable;
 
+    @Parameter(defaultValue = "${null}")
```
mcollovati, line 236: I would add a `property` attribute named for example `frontend.extraFileExtensions` so that it could be specified also on the command line

mcollovati: Also, the field needs Javadoc. IIRC they are used for maven goal documentation

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
```diff
@@ -278,6 +278,13 @@ public abstract class VaadinFlowPluginExtension @Inject constructor(private val
 
     public abstract val cleanFrontendFiles: Property<Boolean>
 
+    /**
+     * The list of extra file extensions that are considered project files.
+     * Hashes are calculated for these files as part of detecting if a new prod
+     * bundle should be generated.
+     */
```
mcollovati, line 285: We should add a note on expected extension format and a simple example list.

flow-server/src/main/java/com/vaadin/flow/server/frontend/Options.java
```diff
@@ -954,4 +956,26 @@ public Options withCleanOldGeneratedFiles(boolean clean) {
     public boolean isCleanOldGeneratedFiles() {
         return cleanOldGeneratedFiles;
     }
+
+    /**
+     * Sets the extra project file extensions.
+     *
```
mcollovati, line 962: We need to provide additional detail on the expected extension format and perhaps  a simple example

mcollovati (changes requested):

mcollovati:
@oliveryasuna let me know if you need help in setting up the tests or if you have any doubts

mcollovati:
@oliveryasuna are you still interested in/do you have time to complete this PR?
Otherwise, the Flow team can take care of and mention you as a co-author.

oliveryasuna:
> @oliveryasuna are you still interested in/do you have time to complete this PR? Otherwise, the Flow team can take care of and mention you as a co-author.

Hey @mcollovati, sorry, it is completely fell off my radar. If the flow team could finish it off, that would be great! This will allow us and many to simplify our build process.

caalador:
Opened new PR with commits from this one. See vaadin/flow#20397

Issue #19512, opened 2024-06-04
Flow app might redirect to /login page when code is compiling in the background.


abdullahtellioglu:
### Description of the bug

When I do some operations using Copilot, the login page might appear from time to time. It is annoying and confusing for users especially after such an operation 

https://github.com/vaadin/flow/assets/100126447/8f778e97-0da5-4077-aa93-dfa768442341

This also happens when there is a code compilation error (the example above is not one of them).

Comment from @Legioth in internal Slack conversation: https://vaadin.slack.com/archives/C6X43FE8M/p1717057898635429

It has been noticed but we have had problems reproducing. That compilation error hint might be useful there since it's probably technically the same conditions as while a compilation is happening in the background with the main difference being that it never completed due to the error. 

### Expected behavior

Login page should not appear. 

### Minimal reproducible example

You can create a flow application and enable flow modifications for the copilot. Then, drag some templates from palette into view and you will encounter it about after 5-10 tries. 

It happens 1 in 5 times when I drop an element from the palette(which causes server restart) if the server startup time is longer than 3~4 seconds when there is no hot-swap agent or JRebel

### Versions

- Vaadin / Flow version: 24.5.0.alpha1
- Java version: Oracle OpenJDK 22.0.1 - aarch64
- OS version:
- Browser version (if applicable):
- Application Server (if applicable):
- IDE (if applicable):

tltv:
I'm not able to reproduce this exact same effect with 24.5.0.alpha1 or 24.5-SNAPSHOT. Actually i don't see server restart to happen at all after source file is updated because there's no compilation triggered. 
@Artur-, just a quick question, is Copilot supposed to also compile Java file after it modifies Java file, or should compilation happen manually by developer or automatically by IDE?

abdullahtellioglu:
Do you have the vaadin plugin installed? Idea plugin compiles the java file. https://plugins.jetbrains.com/plugin/23758-vaadin

Artur-:
Copilot never compiles Java files but your IDE will. IntelliJ if you have the plugin, VS code will always do it automatically I think

tltv:
With https://github.com/vaadin/intellij-plugin/issues/44 being fixed now, I tested this case with the plugin and I see that it has the same double restart issue.
[...]

PR #19564, opened 2024-06-11
fix: improve restart with copilot

tltv:
Reloads page if vite-devmode.ts fails to load.
Fixes ReloadEvent by supporting slash in the class name when extracting package name. Helps with ReloadCache to preserve route packages better when restarting server rapidly via deleted and added route class.

Fixes: #19512


vaadin-spring/src/main/java/com/vaadin/flow/spring/ReloadEvent.java
```diff
@@ -51,7 +51,7 @@ public Set<String> getChangedPackages() {
     }
 
     private Set<String> extractPackageNames(Set<String> classNames) {
-        return classNames.stream()
+        return classNames.stream().map(name -> name.replace("/", "."))
```
tepi, line 54: Is this replacing for path separator character? That should be handled already here https://github.com/vaadin/flow/blob/main/vaadin-spring/src/main/java/com/vaadin/flow/spring/ReloadListener.java#L64-L71

tltv: ReloadListener replaces File.separatorChar with a "." but here it also has to replace "/" with ".". I'm testing with Windows. I think it would make sense to move this to ReloadListener instead to always replace "/" with a ".". Spring dev tools seems to return path with "/" even in Windows.

tepi: > Spring dev tools seems to return path with "/" even in Windows.

Well of course it does :(. In that case I guess it's better to change the logic in ReloadListener to always replace "/" and replace `File.separatorChar` if its != "/".

PR #19578, opened 2024-06-13
fix: add useContent to renderHooks

### LINKED ISSUE #19572: useContent is missing from RenderHooks
### Description of the bug

The `useContent` hook is only available through the web component (i.e. `this.useContent`) which is inconsistent with the other `ReactAdapterElement` hooks that are primarily accessed through `RenderHooks` (e.g. `hooks.useContent`).

### Expected behavior

Expected that all hooks are in `RenderHooks`.

N/A

### Versions

- Vaadin / Flow version: Vaadin 24.5.0.alpha2

caalador:
fixes #19572

### DIFF
--- a/flow-react/src/main/resources/com/vaadin/flow/server/frontend/ReactAdapter.template
+++ b/flow-react/src/main/resources/com/vaadin/flow/server/frontend/ReactAdapter.template
@@ -86,6 +86,13 @@ export type RenderHooks = {
      * @protected
      */
     readonly useCustomEvent: ReactAdapterElement["useCustomEvent"]
+
+    /**
+     * A hook helper to generate the content element for binding Flow element.
+     *
+     * @param name - The name attribute of the element
+     */
+    readonly useContent:  ReactAdapterElement["useContent"]
 };
 
 interface ReadyCallbackFunction {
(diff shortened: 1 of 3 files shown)


flow-react/src/main/resources/com/vaadin/flow/server/frontend/ReactAdapter.template
```diff
@@ -86,6 +86,13 @@ export type RenderHooks = {
      * @protected
      */
     readonly useCustomEvent: ReactAdapterElement["useCustomEvent"]
+
+    /**
+     * A hook helper to generate the content element for binding Flow element.
```
mshabarov, line 91: The javadoc may be more descriptive, e.g.:
```
A hook helper to bind a server-side Flow component with a given string identifier and inject it to this React component.
To be used together with {@link ReactAdapterComponent::getContentElement} to let server-side component aware of the binding.
When bound, can be rendered as follows:
return <>
            <span>Server-side components:</span>
            {content}
</>;
```

mshabarov: `protected useContent(name: string):` implementation method doesn't have a javadoc, would be good to add it either.

mshabarov (changes requested):

flow-react/src/main/resources/com/vaadin/flow/server/frontend/ReactAdapter.template
caalador: javadoc and jsdoc were added in #19574 as I expected that would have gone in quite some time ago.

caalador: Updated, added 19574 into this pr

mshabarov:
Need formatting, otherwise LGTM.

caalador:
merge should fix the formatting as it was all outside of the files touched by the PR

mshabarov (approved):

PR #19631, opened 2024-06-25
fix: handle clicks outside container

### LINKED ISSUE #19630: React Router Full Page Load while clicking link in dialog
### Description of the bug

When I click a RouterLink within a Dialog I expect that the already rendered MainLayout is shown / preserved and no white background / full page reload happens.

### Expected behavior

No full page load.

https://drive.google.com/file/d/1Pb2Rm-7CZPKU0HwGtYdvbNJheXPa0l2W/view?usp=sharing

I’ve added coloring and thread.sleep to make the problem more obvious.

Clicking the link in the MainView: ok
Clicking the link in the Dialog of the MainView: not okay

### Versions

- Vaadin / Flow version: 24.4.3

caalador:
Add the navigationEventHandler
to window to be able to handle
link clicks into dialogs that
are outside of the container element.

Fixes #19630

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -269,11 +269,13 @@ function Flow() {
                                     blocker.reset();
                                 } else {
                                     blocker.proceed();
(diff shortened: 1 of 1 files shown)


mshabarov (approved):
Works as expected with this patch.

PR #19632, opened 2024-06-25
docs: Update javadoc for ValueContext constructors

### LINKED ISSUE #19211: ValueContext constructors javadoc does not agree with code
### Description of the bug

In `ValueContext` class, there's some disagreement between javadoc and code in the constructors.

Javadoc states component can be null, but code does a null check anyway:
```
    /**
     * Constructor for {@code ValueContext}.
     *
     * @param component
     *            The component related to current value. Can be null. If the
     *            component implements {@link HasValue}, it will be returned by
     *            {@link #getHasValue()} as well.
     */
    public ValueContext(Component component) {
        Objects.requireNonNull(component,
                "Component can't be null in ValueContext construction");
```

Same issue exists for the constructor with signature `public ValueContext(Component component, HasValue<?, ?> hasValue)`.

However the last constructor (`public ValueContext(Component component, HasValue<?, ?> hasValue, Locale locale)`) has no null check for component parameter.

### Expected behavior

Expect these to be consistent.

n/a

### Versions

- Flow version: 24.4-SNAPSHOT, probably older versions as well.

AlainaFaisal:
Documentation Update

Fixes #19211
## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


caalador (dismissed):
At least the javadoc is the same opinion as the code, but I'm not 100% on the nonNull requirement as we have many other constructors that use component as null, starting directly with `public ValueContext(Binder<?> binder) {`

mshabarov (changes requested):
I think `null` check is redundant (not needed) in constructors, as semantically component object can be `null` - getter method assumes it can be `null` and some constructors assume that it can be `null`.

I propose to do the opposite - remove the "requires not null" check.

mshabarov (approved):

PR #19638, opened 2024-06-26
feat: React 19

Artur-:


Artur-:
This should not be merged before at least signals are working with React 19

PR #19646, opened 2024-06-27
chore: Vite 5.3.3

Artur-:


flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/vite/package.json
```diff
@@ -10,7 +10,7 @@
   "license": "Apache-2.0",
   "dependencies": {},
   "devDependencies": {
-    "vite": "5.3.1",
+    "vite": "5.3.2",
```
knoobie, line 13: Off topic; just wanted to mention it here cause perfect place.. 

I noticed that when running clean install vaadin downgrades vite in 24.4 from 5.3.1 to 5.2.13 again.. once I start the application it's 5.3.1 again but it is really weird and always creates stale changes in git.

Artur-: Sounds completely off topic but is it `vaadin:prepare-frontend` that downgrades it? Sounds a bit like you somehow get an old Flow version on the classpath for the build part but when running the app you have the correct version

knoobie: Yeah prepare-frontend runs default. You are definitly correct with your assumption, only thing that made me question myself: No addons; just Vaadin 24.4 and dependency:tree looks good. Just wanted to mention it

Artur-: Might be that `mvn dependency:resolve-plugins` would show something else

knoobie: Sadly nothing really obvious; only thing that looked a bit off is the 24.4.1 of flow in 24.4.3 instead of the 24.4.2 mentioned in the release notes:

grafik

mshabarov (approved):

PR #19666, opened 2024-07-03
fix: Pass route data through window.Vaadin.routesConfig instead of an import

### LINKED ISSUE #19658: Circular imports in routes.tsx + Flow.tsx
### Description of the bug

When you start an app using Hilla FS routes and then do
```
touch src/main/frontend/generated/file-routes.ts
```

Two things can happen in the browser:

1.
```
Warning: You are calling ReactDOMClient.createRoot() on a container that has already been passed to createRoot() before. Instead, call root.render() on the existing root instead if you want to update it.
```
This is covered by #19657

2. An additional
```
(index):5 ReferenceError: Cannot access 'Flow' before initialization
    at routes.tsx:61:19
```
This is related to the Vite logs
```
[hmr] /generated/flow/Flow.tsx failed to apply HMR as it's within a circular import. Reloading page to reset the execution order. To debug and break the circular import, you can run `vite --debug hmr` to log the circular dependency path if a file change triggered it.
```
and with `--debug hmr`
```
2024-07-03T13:24:08.956+03:00  INFO 70104 --- [v-server-output] c.v.b.devserver.DevServerOutputTracker   : 2024-07-03T10:24:08.956Z vite:hmr circular imports detected: /generated/routes.tsx -> /generated/flow/Flow.tsx -> /generated/routes.tsx
```

### Expected behavior

No circular import errors

npm init vaadin test-hmr
> Hilla example views
cd test-hmr
mvn
touch src/main/frontend/generated/file-routes.ts

### Versions

- Vaadin / Flow version: 24.4.4

Artur-:
Instead of importing `routes.js/routes.ts/routes.tsx` from the project into `Flow.tsx` and creating a circular dependency, this moves importing of `routes.js` to a new file that is always included when react is used `vaadin-react.tsx`. This file publishes the routes as `window.Vaadin.routesConfig` so that `Flow.tsx` can read them from there without directly depending on `routes.js`

Fixes #19658

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -237,7 +237,7 @@ function Flow() {
                 return;
             }
             const {pathname, search} = blocker.location;
-            const routes = (window as any)?.Vaadin?.routesConfig || [];
+            const routes = ((window as any)?.Vaadin?.routesConfig || []) as AgnosticRouteObject[];
             let matched = matchRoutes(Array.from(routes), window.location.pathname);
 
             // Navigation between server routes
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -21,6 +21,7 @@ import {
     useBlocker,
     useLocation,
     useNavigate,
+    AgnosticRouteObject
 } from "react-router-dom";
 
 const flow = new _Flow({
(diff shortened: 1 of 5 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -238,7 +237,8 @@ function Flow() {
                 return;
             }
             const {pathname, search} = blocker.location;
-            let matched = matchRoutes(Array.from(routes), window.location.pathname);
+            const routes = (window as any)?.Vaadin?.routesConfig || [];
+            let matched = matchRoutes(routes, window.location.pathname);
```
mshabarov, line 241: Shouldn't this be `matchRoutes(Array.from(routes), window.location.pathname)` , like previously ?

Artur-: Maybe but why? It already is an array

mshabarov (approved):
Tested the hmr for react views and server-side live reload with the hybrid app - works as expected. The codes looks good as well.

PR #19669, opened 2024-07-04
feat: add application identifier to build info

mcollovati:
Gets or computes an application identifier during production build and stores it into flow-build-info.json

### DIFF
--- a/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
+++ b/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
@@ -15,10 +15,6 @@
  */
 package com.vaadin.flow.plugin.maven;
 
-import static com.vaadin.flow.server.Constants.VAADIN_SERVLET_RESOURCES;
-import static com.vaadin.flow.server.Constants.VAADIN_WEBAPP_RESOURCES;
-import static com.vaadin.flow.server.frontend.FrontendUtils.FRONTEND;
-
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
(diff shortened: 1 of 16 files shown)


flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
```diff
@@ -459,4 +457,8 @@ public boolean isReactEnabled() {
         return reactEnable;
     }
 
+    @Override
+    public String applicationIdentifier() {
+        return project.getGroupId() + ":" + project.getArtifactId();
```
mshabarov, line 462: This will be `com.vaadin:vaadin-dev-bundle` always, but I believe this will never be used, as Vaadin plugin will make a new token file anyway with the real app id, right?

mcollovati: Exactly. And the token is updated by build-frontend goal, regardless the bundle was precompiled or just built.

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
```diff
@@ -405,43 +407,59 @@ public class PluginEffectiveConfiguration(
...
 
     public val reactEnable: Provider<Boolean> = extension.reactEnable
         .convention(FrontendUtils.isReactRouterRequired(BuildFrontendUtil.getFrontendDirectory(GradlePluginAdapter(project, this, true))))
-        .overrideWithSystemProperty(InitParameters.REACT_ENABLE)
+        .overrideWithSystemPropertyFlag(InitParameters.REACT_ENABLE)
 
     public val cleanFrontendFiles: Property<Boolean> = extension.cleanFrontendFiles
             .convention(true)
+
+    public val applicationIdentifier: Provider<String> = extension.applicationIdentifier.convention(project.name)
```
mshabarov, line 433: Maybe project name is enough, but for future we can use `project.group` instead, if project.name="web" (some default value for multi module cases, I suppose)

mshabarov (approved):

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
mcollovati: My understanding of gradle is that sub-project name is usually the name of the subfolder or it is defined by developer in `settings.gradle`.

On the other hand, `project.group` will be empty for single module project and will be the name of the parent module for multimodule projects

mcollovati: But, yeah, we can change the default later

PR #19716, opened 2024-07-23
chore: updgrade pnpm in validation workflow

mcollovati:


mcollovati:
Verified in https://github.com/vaadin/flow/pull/19714, on the `flow-test-multi-war2` module.
Other test failures depend on selenium and are fixed in another PR.

PR #19722, opened 2024-07-25
fix: prevent style leaking with exported webcomponents

mcollovati:


mcollovati:
This is an experimental attempt to try to fix some exported web-components issues:
- styles leaking in the embedding document
- styles leaking between web-components from different applications
- JavaScript import errors when embedding from different applications

mcollovati:
It looks like this experiment is breaking everything

Issue #19635, opened 2024-06-25
React Router blocks navigation to the current view


knoobie:
### Description of the bug

When I click on a link that targets the current page no server-side round-trip takes place and no navigation events are triggered.

1. Click prints (when I wanna enter the route for the first time):
```
Was I called 0?
Was I called 1?
Was I called 2?
```
2. Click prints (when I'm already on the route):

```
```

### Expected behavior

Like in previous versions: The Navigation Observer gets triggered so that I can "reload" the page for people clicking the same link in a menu.

1. Click prints (when I wanna enter the route for the first time):
```
Was I called 0?
Was I called 1?
Was I called 2?
```
2. Click prints (when I'm already on the route):
```
Was I called 1?
Was I called 2?
```

### Minimal reproducible example

```java
@Route(value = "test", layout = MainLayout.class) // Your Main Layout of choice where you place a router link to this view
public class TestView extends Div implements BeforeEnterObserver, AfterNavigationObserver {

  public TestView() {
    System.out.println("Was I called 0?");
  }

  @Override
  public void beforeEnter(BeforeEnterEvent event) {
    System.out.println("Was I called 1?");
  }

  @Override
  public void afterNavigation(AfterNavigationEvent afterNavigationEvent) {
    System.out.println("Was I called 2?");
  }
}

```

### Versions

- Vaadin / Flow version: 24.4.x

mshabarov:
React Router behaves in a different way than our Vaadin Router in that case.
But nevertheless, this is a bug as the Flow navigation does not happens as it did with Vaadin Router.
We have to investigate, meanwhile the workaround may be to fallback to Vaadin Router unfortunately.

knoobie:
Created a BFP (VS-5299) for this (just in case Jira has problems)

mcollovati:
The issue was triaged and currently added to the backlog priority queue for further investigation

PR #19730, opened 2024-07-29
fix: do not block navigation to current url

tepi:
Lets navigation cause server round-trip, when navigation is triggered by clicking a link.

Fixes https://github.com/vaadin/flow/issues/19635

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -233,10 +233,6 @@ function Flow() {
 
     useEffect(() => {
         if (blocker.state === 'blocked') {
-            if(navigated.current) {
-                blocker.proceed();
-                return;
-            }
             const {pathname, search} = blocker.location;
             const routes = ((window as any)?.Vaadin?.routesConfig || []) as AgnosticRouteObject[];
             let matched = matchRoutes(Array.from(routes), window.location.pathname);
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -184,6 +184,8 @@ function Flow() {
         }
 
         navigated.current = false;
+        // When navigation is triggered by click on a link, fromAnchor is set to true
+        // in order to get a server round-trip even when navigating to the same URL again
         fromAnchor.current = true;
         navigate(path);
     }, [navigate]);
@@ -234,6 +236,7 @@ function Flow() {
 
     useEffect(() => {
         if (blocker.state === 'blocked') {
+            // Do not skip server round-trip if navigation originates from a click on a link
             if (navigated.current && !fromAnchor.current) {
                 blocker.proceed();
                 return;
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -170,7 +170,7 @@ function Flow() {
     });
     const {pathname, search, hash} = useLocation();
     const navigated = useRef<boolean>(false);
-
+    const fromAnchor = useRef<boolean>(false);
     const containerRef = useRef<RouterContainer | undefined>(undefined);
 
     const navigateEventHandler = useCallback((event: MouseEvent) => {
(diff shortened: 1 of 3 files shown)


flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
```diff
@@ -184,6 +184,7 @@ function Flow() {
         }
 
         navigated.current = false;
+        fromAnchor.current = true;
```
caalador, line 187: The only thing I would like for this fix is a comment on the fromAnchor that clearly states that the requirement it to `get a server roundtrip for clicks on links even when the target is the current url`.

tepi: Added comments

caalador:
Perhaps we should add the TestView from the issue with a RouterLink targeting self and use a messageList (like we have in [DomEventFilterView](https://github.com/vaadin/flow/blob/f81d6bddd603cc5a659bcddb9cbd2273ae7e40c0/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DomEventFilterView.java#L67)) and check that messages are added correctly so that it is not broken by mistake later on.

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/NavigationEventsIT.java
```diff
@@ -0,0 +1,82 @@
...
+import com.vaadin.flow.component.html.testbench.AnchorElement;
+import com.vaadin.flow.testutil.ChromeBrowserTest;
+
+public class NavigationEventsIT extends ChromeBrowserTest {
+
+    @Test
+    public void assertNavigationToSelfProducesNavigationEvents() {
+        open();
+
+        assertMessages(0, "BeforeEnterEvent", "AfterNavigationEvent");
```
knoobie, line 35: Shouldn't this already be 2 because of the initial visit to the view?

tepi:
Added test to check correct navigation events happening

flow-tests/test-root-context/src/test/java/com/vaadin/flow/uitest/ui/NavigationEventsIT.java
```diff
@@ -0,0 +1,82 @@
...
+        messages = messages.subList(skip, messages.size());
+
+        if (messages.size() < expectedTail.length) {
+            Assert.fail("Expected " + expectedTail.length
+                    + " messages, but there are only " + messages.size() + ". "
+                    + joinMessages(messages));
+        }
+
+        for (int i = 0; i < expectedTail.length; i++) {
+            Assert.assertEquals("Unexpected message at index " + i,
```
caalador, line 60: Should this be `i + skip`? As now the fail will possibly be with an offset that doesn't match the screenshot.
Not really important, but this should perhaps give the position in the actual messages list instead of the subList

knoobie: Just my two cents: This tests looks IMHO more complicated than necessary? Why not a simple

```java
assertThat(messages.getText()
.as("Events should be executed initial")
.isEqualsTo("EventA_1 & EventB_1");

openAgain();

assertThat(messages.getText()
.as("Events should be executed initial + once now")
.isEqualsTo("EventA_1 & EventA_1 & EventA_2 & EventB_2");

openAgain();

assertThat(messages.getText()
.as("Events should be executed initial + twice now")
.isEqualsTo("EventA_1 & EventA_1 & EventA_2 & EventB_2 & EventA_3 & EventB_3");
```

caalador: Long string equals checks make the actual exception hard to spot.
For tests we should always make the exceptions easy to spot even if that makes the test code longer or a bit more complex.

knoobie: Understandable argument when using normal junit! You might be interested in [AssertJ](https://assertj.github.io/doc/) to get also more readable tests for easier maintainability for you guys :)

tepi: Simplified the test a bit, I hope it's ok now

caalador: Perhaps missing an assertion message on all 3 equals assertions. Perhaps mostly for the size as just seeing `3 is not 4` requires looking into the sources to figure out what it is about.

caalador (approved):

PR #19736, opened 2024-08-01
feat: push and replace without navigation callback

### LINKED ISSUE #19613: History#replaceState trigger navigation event with React router
### Description of the bug

When using React router, History#pushState and #replaceState trigger navigation event (trigger HISTORY, coming from a dom event "ui-navigate"). Even if only query parameters are changed. 

With Flow Router, no navigation event occurs, not matter which part of the path has changed.

May relate to 
* #19580 Also has something to do with breaking behavior change in History#replaceState with React Router
* #19540 Also has something to do with undesired navigation event
* #19494 Also has something to do with undesired HISTORY navigation event if the target URL does not exactly match the current one
* #19539 This sounds like a similar fix, but only applied to pushState. Indeed, pushState work as Flow Router and also as I would expect.

### Expected behavior

No navigation events on History#pushState and #replaceState. If a navigation event is desired, I would use UI.navigate.

I modified the example from a  comment in #19580 

```
@Route(value = "test/:path?")
public class TestView extends VerticalLayout implements BeforeEnterObserver, BeforeLeaveObserver {

    public TestView() {
        add(new Button("replace state - no query params",
                event -> UI.getCurrent().getPage().getHistory().replaceState(null, "test")));
        // no navigation only if the current URL is the same including query params
        add(new Button("replace state - push query param",
                event -> UI.getCurrent().getPage().getHistory().replaceState(null, "test?new")));
        // no navigation only if the current URL is the same including query params
        add(new Button("replace state - change path param",
                event -> UI.getCurrent().getPage().getHistory().replaceState(null, "test/new")));
        // no navigation only if the current URL is the same including query params
        add(new Button("push state - no query params",
                event -> UI.getCurrent().getPage().getHistory().pushState(null, "test")));
        // no navigation
        add(new Button("push state - query param",
                event -> UI.getCurrent().getPage().getHistory().pushState(null, "test?new")));
        // no navigation
        add(new Button("push state - change path param",
                event -> UI.getCurrent().getPage().getHistory().pushState(null, "test/new")));
        // no navigation
    }

    @Override
        add(new Div("beforeEnter"));
    }

    @Override
    public void beforeLeave(BeforeLeaveEvent beforeLeaveEvent) {
        add(new Div("beforeLeave"));
    }
}
```

### Versions

- Vaadin / Flow version: 24.4.3
- Java version: 17
- OS version: Windows
- Application Server (if applicable): Spring Boot
- IDE (if applicable):

caalador:
Add feature to support replace and
push without getting a callback
to the server for the change.

fixes #19613

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/History.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/History.java
@@ -194,7 +194,7 @@ public class History implements Serializable {
 
     /**
      * Invokes <code>history.pushState</code> in the browser with the given
-     * parameters.
+     * parameters. Does not make a callback to the server by default.
      *
      * @param state
      *            the JSON state to push to the history stack, or
@@ -281,7 +281,7 @@ public class History implements Serializable {
 
     /**
      * Invokes <code>history.replaceState</code> in the browser with the given
-     * parameters.
+     * parameters. Does not make a callback to the server by default.
      *
      * @param state
      *            the JSON state to push to the history stack, or
@@ -291,7 +291,7 @@ public class History implements Serializable {
      *            to only change the JSON state
      */
     public void replaceState(JsonValue state, Location location) {
-        replaceState(state, location, true);
+        replaceState(state, location, false);
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/History.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/History.java
@@ -173,8 +173,8 @@ public class History implements Serializable {
     /**
      * Invokes <code>history.pushState</code> in the browser with the given
      * parameters. This is a shorthand method for
-     * {@link History#pushState(JsonValue, Location)}, creating {@link Location}
-     * from the string provided.
+     * {@link History#pushState(JsonValue, Location, boolean)}, creating
+     * {@link Location} from the string provided.
      *
      * @param state
      *            the JSON state to push to the history stack, or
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/component/page/History.java
```diff
@@ -172,7 +172,29 @@ public void pushState(JsonValue state, String location) {
 
     /**
      * Invokes <code>history.pushState</code> in the browser with the given
-     * parameters.
+     * parameters. This is a shorthand method for
+     * {@link History#pushState(JsonValue, Location)}, creating {@link Location}
```
tepi, line 176: This should point to the new 3-parameter method

flow-server/src/main/java/com/vaadin/flow/component/page/History.java
```diff
@@ -217,7 +258,30 @@ public void replaceState(JsonValue state, String location) {
 
     /**
      * Invokes <code>history.replaceState</code> in the browser with the given
-     * parameters.
+     * parameters. This is a shorthand method for
+     * {@link History#replaceState(JsonValue, Location)}, creating
```
tepi, line 262: This should point to the new 3-parameter method

caalador: done

flow-server/src/main/java/com/vaadin/flow/component/page/History.java
caalador: done

tepi:
Looks good to me. @mshabarov should check and ok the new API next week.

mshabarov (approved):
Needs to be documented

Issue #19700, opened 2024-07-17
WebComponentExporter: @CSSImport styles are registered in the head


roelnoten:
### Description of the bug

When exposing a part of a vaadin application as a webcomponent via the WebComponentExporter, then all styles loaded via the java @CSSImport annotation (without themefor attribute) are registered in the head of the parent document instead of within the bounds of the webcomponent.
Not only does this leak styles in to the parent document and can thus alter it, it also leads to the absence of those styles in the webcomponent.

The issue can be circumvented by not using @CSSImport and instead importing the css files through @import statements in the themes styles.css file.
But, when using addons (form the vaadin addon directory) those typically also reference their styles via @CSSImport and thus their styling is missing when used in a WebComponentExporter.

### Expected behavior

Styles loaded via @CSSImport should be registered within the webcomponent when running as a WebComponentExport webcomponent; and should not be registered in the head of the parent document.

### Minimal reproducible example

Try for instance adding the addon `<dependency><groupId>org.vaadin.tatu</groupId><artifactId>twincolselect</artifactId<version>2.5.2</version></dependency>` to a WebComponentExporter.
This addon does a @CSSImport at https://github.com/TatuLund/TwinColSelect/blob/ec21083c1502ec8f150a603185be1f6151e45d9a/src/main/java/org/vaadin/tatu/TwinColSelect.java#L73 of https://github.com/TatuLund/TwinColSelect/blob/master/frontend/themes/mytheme/styles.css.

You'll notice these styles of that styles.css are registered in the head of the document and not active within the WebComponentExporter webcomponent.

### Versions

- Vaadin / Flow version: 23.3.34
- Java version: any
- OS version: any
- Browser version (if applicable): any
- Application Server (if applicable): any
- IDE (if applicable): any

mcollovati:
The issue was triaged and currently added to the backlog priority queue for further investigation

mcollovati:
The issue was assigned to a developer and is currently being investigated

PR #19740, opened 2024-08-01
fix: apply CssImport to exported webcomponent

tltv:
CssImport annotation with just a `value` attribute can be injected automatically to shadow root of all exported web components (embedded applications) using Constructable StyleSheets. `WebComponentExporter` should have a theme to make automation work properly in `theme-generator.js`. Theme property `"autoInjectGlobalCssImports": true` in `theme.json` enable auto injection. Disabled by default.

Fixes: https://github.com/vaadin/flow/issues/19700

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
@@ -78,7 +78,6 @@ abstract class AbstractUpdateImports implements Runnable {
             + " tpl.innerHTML = block;\n"
             + " document.head.appendChild(tpl.content);\n" + "}";
     private static final String IMPORT_INJECT = "import { injectGlobalCss } from 'Frontend/generated/jar-resources/theme-util.js';\n";
-    private static final String IMPORT_COMPOSE_LOAD_ON_DEMAND = "import { composeLoadOnDemand } from 'Frontend/generated/jar-resources/theme-util.js';\n";
     private static final String IMPORT_WC_INJECT = "import { injectGlobalWebcomponentCss } from 'Frontend/generated/jar-resources/theme-util.js';\n";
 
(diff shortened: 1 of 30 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
```diff
@@ -76,6 +78,8 @@ abstract class AbstractUpdateImports implements Runnable {
             + " tpl.innerHTML = block;\n"
             + " document.head.appendChild(tpl.content);\n" + "}";
     private static final String IMPORT_INJECT = "import { injectGlobalCss } from 'Frontend/generated/jar-resources/theme-util.js';\n";
+    private static final String IMPORT_COMPOSE_LOAD_ON_DEMAND = "import { composeLoadOnDemand } from 'Frontend/generated/jar-resources/theme-util.js';\n";
```
mcollovati, line 81: The `composeLoadOnDemand` function seems not be used anymore. I suppose we can remove both the constant and the function definition in `theme-utils`

flow-server/src/main/java/com/vaadin/flow/server/frontend/AbstractUpdateImports.java
```diff
@@ -292,7 +314,7 @@ private Map<File, List<String>> process(Map<ChunkInfo, List<CssData>> css,
             start = System.nanoTime();
 
             chunkLoader.add("");
-            chunkLoader.add("const loadOnDemand = (key) => {");
+            chunkLoader.add("const loadOnDemand = (key, isWebcomponent) => {");
```
mcollovati, line 317: The `isWebcomponent` parameter seems not required anymore

tltv: I leaved it in just in case. But as it's not used after all, I'll remove it.

mcollovati (approved):

PR #19747, opened 2024-08-05
feat: Add CompletableFuture WebStorage API

### LINKED ISSUE #19743: Improve WebStorage API
### Describe your motivation

I'm programming in kotlin and heavily using coroutines. The WebStorage API only defines static functions with callbacks, which can only be extended as suspending in Kotlin with:

```kotlin
suspend fun getWebstorageItem(key: String): String? {
    return suspendCoroutine { continuation ->
        WebStorage.getItem(key) { continuation.resume(it) }
    }
}
```

Because these are static methods I couldn't do:

```kotlin
suspend fun WebStorage.getItem(key: String): String? {
    return suspendCoroutine { continuation ->
        getItem(key) { continuation.resume(it) }
    }
}
```

That's why I suggest to add overload methods that return a CompletableFuture instead of taking a callback. This way I can easily await on the CompletableFuture. Like:

```java
    public static CompletableFuture getItem(String key) {
        // Impl Detail
    }
```

```kotlin
     WebStorage.getItem("key").await()
```

This way also other reactive frameworks would benefit from it.
The API could be expanded like this without breaking the existing one.

Also I would suggest to refactor the API to use objects. So that the calls could be:

```java
   WebStorage.getLocalStorage().setItem("key")
   WebStorage.getSessionStorage().setItem("key")
   
   WebStorage.getLocalStorage().getItem("key")
   WebStorage.getSessionStorage().getItem("key")
```

For spsecific UI:

```java   
   UI ui = UI.getCurrent()
   WebStorage.getLocalStorage(ui).setItem("key")
   WebStorage.getSessionStorage(ui).setItem("key")
   
   WebStorage.getLocalStorage(ui).getItem("key")
   WebStorage.getSessionStorage(ui).getItem("key")
```
Or as alternative construct it directly as object:

```java   
   new LocalStorage(UI.getCurrent()).getItem("key")
   new SessionStorage(UI.getCurrent()).getItem("key")   
```

All my suggestions would allow to keep the current static approach without breaking.

Dudeplayz:
## Description

Adds CompletableFuture methods for `WebStorage.getItem`.

Fixes #19743

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/WebStorage.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/WebStorage.java
@@ -223,10 +223,8 @@ public interface WebStorage extends Serializable {
      */
     public static void getItem(UI ui, Storage storage, String key,
             Callback callback) {
-        ui.getPage()
-                .executeJs("return window[$0].getItem($1);", storage.toString(),
-                        key)
-                .then(String.class, callback::onValueDetected, s -> {
+        requestItem(ui, storage, key).then(String.class,
+                callback::onValueDetected, s -> {
                     LoggerFactory.getLogger(WebStorage.class.getName()).debug(
                             "Error while getting value for key '{}' from storage '{}': {}",
                             key, storage, s);
@@ -296,10 +294,13 @@ public interface WebStorage extends Serializable {
      */
     public static CompletableFuture<String> getItem(UI ui, Storage storage,
             String key) {
-        return ui
-                .getPage().executeJs("return window[$0].getItem($1);",
-                        storage.toString(), key)
-                .toCompletableFuture(String.class);
+        return requestItem(ui, storage, key).toCompletableFuture(String.class);
+    }
+
+    private static PendingJavaScriptResult requestItem(UI ui, Storage storage,
+            String key) {
+        return ui.getPage().executeJs("return window[$0].getItem($1);",
+                storage.toString(), key);
     }
 
 }
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/WebStorage.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/WebStorage.java
@@ -18,6 +18,8 @@ package com.vaadin.flow.component.page;
 import java.io.Serializable;
 import java.util.concurrent.CompletableFuture;
 
+import org.slf4j.LoggerFactory;
+
 import com.vaadin.flow.component.UI;
 
 /**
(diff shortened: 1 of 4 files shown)


Dudeplayz:
Questions:
- The callback style methods could also use the CompletableFuture one and be called with then?
- CompletableFuture would return an exception in error cases instead of null, how is the behavior intended here?

mcollovati:
> The callback style methods could also use the CompletableFuture one and be called with then?

I would leave the callback method as is. I think there's no need to create a `CompletableFuture` there.

> CompletableFuture would return an exception in error cases instead of null, how is the behavior intended here?

My opinion is that the `CompletableFuture` version should behave in the same way as the `Callback` one. So, probably the returned `CompletableFuture` should have an `exceptionally(...)` stage that returns null (and maybe logging the error at DEBUG level). It would also be good to mention the behavior on the Javadoc.

flow-server/src/main/java/com/vaadin/flow/component/page/WebStorage.java
```diff
@@ -215,17 +219,88 @@ public static void getItem(Storage storage, String key, Callback callback) {
...
+     * while holding the session lock since the request handling thread that
+     * makes the result available will also need to lock the session. <br>
+     * See {@link PendingJavaScriptResult#toCompletableFuture} for more
+     * information.
+     *
+     * @param key
+     *            the key for which the value will be fetched
+     * @return a CompletableFuture that will be completed with the value once
+     *         transferred from the client side or <code>null</code> if the
+     *         value was not set
```
mcollovati, line 249: I would use `available` as in the `Callback` javadoc
```suggestion
     *         value was not available.
```

flow-server/src/main/java/com/vaadin/flow/component/page/WebStorage.java
```diff
@@ -215,17 +219,88 @@ public static void getItem(Storage storage, String key, Callback callback) {
...
+     * See {@link PendingJavaScriptResult#toCompletableFuture} for more
+     * information.
+     *
+     * @param storage
+     *            the storage
+     * @param key
+     *            the key for which the value will be fetched
+     * @return a CompletableFuture that will be completed with the value once
+     *         transferred from the client side or <code>null</code> if the
+     *         value was not set
```
mcollovati, line 269:
```suggestion
     *         value was not available.
```

flow-server/src/main/java/com/vaadin/flow/component/page/WebStorage.java
```diff
@@ -215,17 +219,88 @@ public static void getItem(Storage storage, String key, Callback callback) {
...
+     *
+     * @param ui
+     *            the UI for which the storage is related to
+     * @param storage
+     *            the storage
+     * @param key
+     *            the key for which the value will be fetched
+     * @return a CompletableFuture that will be completed with the value once
+     *         transferred from the client side or <code>null</code> if the
+     *         value was not set
```
mcollovati, line 293:
```suggestion
     *         value was not available.
```

mcollovati (changes requested):

mcollovati (approved):

PR #19760, opened 2024-08-08
Provide an API to find a Java class for component with a given tag

### LINKED ISSUE #19353: Provide an API to find a Java class for component with a given tag
### Describe your motivation

Vaadin Copilot needs a way to find which Java component corresponds to e.g. `vaadin-horizontal-layout`. Vaadin Flow to provide an API to give this information, i.e. give a `Class<? extends Component>` for the given `String` tag.

To include scanning for `AtTag` in the scanner that finds `AtJsModule` at a build time and produce a map `<String, Class<? extends Component>>`, which is available with a public API.

### Additional context

Limited to only DEV mode, not needed for PROD.

Targets Vaadin 24.5.

AlainaFaisal:
## Description
This pull request adds functionality to map HTML tags to Java component classes for Vaadin Copilot. Includes updates to ComponentMetaData and AnnotationReader.

Fixes #19353

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


caalador (commented):
Collecting the tags in ComponentMetaData only collects them for that MetaData component as we also see in the test that you can't get the `mock-tag` from `metaDataAnotherSample`.

I would guess the better way to go would be to take `@Tag` in `DevModeStartupListener` `@HandlesTypes` and collect the tag-to-components into a map inside a Registry.
Artur (or Marcin) probably could give a hint where they would want to have the Registry so they could easily get it.

Artur-:
Isn't this something that is application wide, so it can be a static method anywhere to either get the full map or the mapping for a given tag? It does not have to be discoverable so something like `ComponentUtil.getSomethingSuitable` would work

flow-server/src/main/java/com/vaadin/flow/component/internal/ComponentMetaData.java
```diff
@@ -266,4 +266,4 @@ private static void doCollectSynchronizedProperties(Class<?> clazz,
                     propertyName, eventNames, annotation.allowUpdates()));
         }
     }
-}
+}
```
caalador, line 269: This one could be fully reverted to not remove the end new line.

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
```diff
@@ -59,6 +62,24 @@ public class ComponentUtil {
     static ReflectionCache<Component, ComponentMetaData> componentMetaDataCache = new ReflectionCache<>(
             ComponentMetaData::new);
 
+    private static final Map<String, Set<Class<? extends Component>>> tagToComponentsMap = new ConcurrentHashMap<>();
+
+    public static void registerComponentClass(String tag,
```
caalador, line 67: JavaDocs missing for the public methods.
Also remember to add the information that the mapping is only populated in development mode.

caalador (changes requested):

caalador (approved):

Issue #19728, opened 2024-07-26
Root items are never removed from the key mapper


vursen:
### Description of the bug

Root items are never removed from the key mapper when scrolling through a large TreeGrid. This is likely a regression from https://github.com/vaadin/flow/pull/17774.

### Expected behavior

Root items should be removed from the key mapper once they move out of the requested range.

### Minimal reproducible example

```java
TreeGrid<String> grid = new TreeGrid<>();
grid.addHierarchyColumn(item -> item).setHeader("Name");

List<String> rootItems = IntStream.range(0, 1500).mapToObj(i -> "Item " + i).toList();
grid.setItems(rootItems, root -> Collections.emptyList());

add(grid);
```

Try to scroll this grid down to the end and see that the root items that are no longer in the viewport stay in the key mapper.

### Versions

- Vaadin / Flow version: 24.5 and earlier
- Java version: 17
- OS version: Mac OS

PR #19764, opened 2024-08-08
fix: Allow keymapper to remove root items

tepi:
Fixes https://github.com/vaadin/flow/issues/19728

Also fixes https://github.com/vaadin/flow/issues/14351 in a more sane way and hopefully does not cause a regression which the original fix did.


tepi:
Note to self: If the fix is viable, it will also need a test in flow-components to confirm that the root items are properly removed from the key mapper when scrolling them out of the requested range.

PR #19766, opened 2024-08-08
feat: Strongly cache Copilot files

Artur-:
Speeds up every reload as there is no need to check if the files have changed


Artur-:
This needs a change to copilot or here as the entry point, currently `generated/copilot/copilot.js` cannot be strongly cached

Artur-:
Probably better to fix that in Copilot

mshabarov (approved):

PR #19771, opened 2024-08-09
feat: add support for Daily Active User feature

mcollovati:
Integrates with License Checker to add support for Daily Active User feature for DAU enabled production build artifacts.


flow-server/pom.xml
```diff
@@ -69,7 +69,6 @@
         <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>license-checker</artifactId>
-            <scope>provided</scope>
```
knoobie, line 72: big thumbs down!

flow-server/src/main/java/com/vaadin/flow/server/dau/DAUUtils.java
```diff
@@ -0,0 +1,387 @@
...
+     * {@code creationTime} expressed as number of milliseconds from the epoch
+     * of 1970-01-01T00:00:00Z. The cookie creation time is required on
+     * subsequent requests to detect active users. By default, the cookie
+     * expires after 24 hours.
+     *
+     * @param request
+     *            http request from browser
+     * @return http cookie to be used to count application's end-users daily
+     */
+    public static Cookie generateNewCookie(VaadinRequest request) {
```
knoobie, line 88: Storing cookies is in my opinion a big no-no because each cookie has to be present in e.g. the privacy page of an application.. which has to be approved by (sometimes) multiple (even external) stakeholders...

flow-server/pom.xml
mshabarov: LicenseChecker API is now used by Flow server at runtime, it's better to use compile scope thus.

knoobie: I understand why it's done but I still think this is really bad. flow-server should NOT depend on commercial things. If the DAU feature or other things require to be run by flow.. shouldn't those be discovered by flow via service discovery or any other means? Now open-source flow depends on something that is "open source" by license, but NOT open source accessible making it a really big problem for audits.

flow-server/src/main/java/com/vaadin/flow/server/dau/DAUUtils.java
Legioth: Would localStorage be any better? They are technically equivalent but are there any legal differences?

knoobie: I don't think so (but I'm no lawyer) - just as an example: https://www.3sat.de/datenschutz/cookies-und-local-storage-100.html

Legioth: That's what I suspected as well. It basically boils down to the fact that the new subscription models that we're planning that are based on counting the number of daily active users will not be a good fit for all potential customers. Those customers will instead need to negotiate a custom subscription deal with us where pricing is based on other factors. But we will still assume daily active users as the basis for pricing will be a good match for many (or at least better than the number of named developers).

mshabarov: Let me give a bit more context.

This cookie will be only generated/used by applications of Vaadin customers who use the new (non-yet-sold) subscription type.
It won't impact open-source Vaadin users in any way, nor customers with the existing subscriptions.

Additionally, this cookie doesn't carry any personal data of end-users, it's only used to know the number of daily users.

flow-server/pom.xml
mshabarov: We have to test if the `provided` scope works in runtime if any commercial Vaadin product is used.
My assumption that yes, because if one uses `com.vaadin:vaadin` artifact, the license-checker dependency is fetched anyway, e.g. by Collaboration Engine.

If it's not the case, it think we can go with plan B consisting in using something like service loader or reflection.

Brilliant comment by the way!

flow-server/src/main/java/com/vaadin/flow/server/dau/DAUUtils.java
knoobie: Just asking: Why would you need a cookie or local storage when the application developer / server supplies an `UserIdentitySupplier`? Wouldn't this be enough to ensure that each user has its unique identify for DAU tracking? When no `UserIdentitySupplier` the cookie solution can be used as fallback.. but once this thing is implemented by an application the developer should be in charge to get you the current logged-in user, or? Or do you also intend to track "anonmyous" DAUs?

simasch: I don't like the cookie solution as well but for another reason. Some of my customers are using multiple browsers or the browser in incognito mode. Would this count also multiple users?

Legioth: The cookie is basically a fallback for when there's no `UserIdentitySupplier` and for users that don't log in (but note that just opening the login form but not proceeding will not pass the threshold to be counted as a user). All cookie values that get the same user id through `UserIdentitySupplier` are counted as the same user even if they use multiple browsers.

knoobie: Wouldn't that also mean that a single (and even small) DDOS attack would render my application useless because every anonymous user would count as DAU?

flow-server/pom.xml
mcollovati: I restored the provided scope and split some classes so that a normal application will not fail if license checker not is not present at runtime.

flow-server/src/main/java/com/vaadin/flow/server/dau/DAUUtils.java
knoobie: When you say it like this.. I've got a bad feeling that the DAUs aren't calculated in my application.. but somewhere in the cloud and my application requires access to the internet to send some statistics to your server.. instead of having the limit available in the license key/file so that it can be calculated offline.

Legioth: Offline requirements is another reason for customers to opt for a model not based on DAU that I already mentioned. We're not making DAU the only option but we expect it to be the simplest and "fairest" option in enough cases to justify the development cost.

Offline calculation would make a bunch of other things more complicated (e.g. persistence, renewals, clustering, notifying owners for temporary overruns, grace periods). We might develop a solution for all those challenges later but we're starting with this simpler but not universally applicable approach.

knoobie: Before we go overboard in this PR: while I understand the reasoning behind the decision; it feels inappropriate for strict regulated customer like banks and so on where this would put your business modell at a disadvantage.. yes there are different license models.. but probably with a way higher costs compared to a 10-20 DAU license making Vaadin even less interesting for internal applications. We can discuss this in deeper detail at Vaadin Create in person!

mshabarov (approved):

PR #19775, opened 2024-08-12
Create Fieldset - Legend Java API

### LINKED ISSUE #18213: Create Fieldset - Legend Java API
### Describe your motivation

I would like to create a `<fieldset>` and `<legend>` html tag without using a custom Java class to group my fields.

I would like a Java API for these 2 tags, they are linked together so the structure can be similar to the NativeDetails/Summary.

Create my own API, use the aria attribute but it's less accessible and there is no Java API for it. (So I have to use the Element API).

AlainaFaisal:
## Description

Java API to create a <fieldset> and <legend> html tag without using a custom Java class to group fields.

Fixes #18213

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/FieldSet.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/FieldSet.java
@@ -131,6 +131,15 @@ public class FieldSet extends HtmlContainer implements HasAriaLabel {
     }
 
     /**
+     * Gets the text of the legend.
+     *
+     * @return the text of the legend, or null if no legend is present.
+     */
+    public String getLegendText() {
+        return (legend != null) ? legend.getText() : null;
+    }
+
+    /**
      * Returns the content of the fieldset.
      *
      * @return the content component, can be null.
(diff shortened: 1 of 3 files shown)


mshabarov:
@rolfsmeds @yuriy-fix just double checking to avoid further confusion, does Design System have any concerns about this feature or reasons for not doing this in Flow, e.g. if you have plans to implement something similar as a standalone components ?

flow-html-components/src/main/java/com/vaadin/flow/component/html/FieldSet.java
```diff
@@ -0,0 +1,151 @@
...
+    public void setLegendText(String text) {
+        legend.setText(text);
+    }
+
+    /**
+     * Returns the content of the fieldset.
+     *
+     * @return the content component, can be null.
+     */
+    public Component getContent() {
```
jcgueriaud1, line 133: Should the FieldSet be a `HtmlContainer` since it's supposed to have a list of fields inside?

rolfsmeds:
> @rolfsmeds @yuriy-fix just double checking to avoid further confusion, does Design System have any concerns about this feature or reasons for not doing this in Flow, e.g. if you have plans to implement something similar as a standalone components ?

No current plans – we will probably introduce a Form Section component at some point, which could potentially have some overlap with this, but I don't see that as a reason not to introduce this as API for native HTML fieldset.

PR #19781, opened 2024-08-13
fix: use URLClassLoader with ResourceProvider in Maven plugins

platosha:
Fixes vaadin/hilla#2637

ResourceProvider ignored the actual project's resources when using the default ClassLoader. This resulted in loading `vaadin-featureflags.properties` from the vaadin-dev-bundle jar, even though there is a file in the project.

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/utils/ResourceProviderImpl.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/utils/ResourceProviderImpl.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.List;
 
 import com.vaadin.flow.di.ResourceProvider;
+import com.vaadin.flow.server.frontend.scanner.ClassFinder;
 
 /**
  * {@link ResourceProvider} for use with plugin execution.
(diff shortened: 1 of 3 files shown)


flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/utils/ResourceProviderImpl.java
```diff
@@ -31,6 +32,16 @@
...
+        this(ResourceProviderImpl.class.getClassLoader());
+    }
+
+    ResourceProviderImpl(ClassLoader classLoader) {
+        this.classLoader = classLoader;
+    }
+
     @Override
     public URL getApplicationResource(String path) {
         return ResourceProviderImpl.class.getClassLoader().getResource(path);
```
caalador, line 47: I guess all of the `ResourceProviderImpl.class.getClassLoader()` should be changed to `classLoader` instead

Artur-:
Does #19010 resolve this, but in a different wat?

platosha:
@Artur- probably not, here the issue is in the `ResourceProviderImpl` not being wired to the the ClassFinder or ClassLoader.

caalador (approved):

jcgueriaud1:
Is it supposed to be backported in 24.4 ?

Issue #19772, opened 2024-08-09
Estimated size methods should be marked as unsupported for HierarchicalDataCommunicator


vursen:
### Description of the bug

`HierarchicalDataCommunicator` does not seem to support estimated sizes (undefined sizes). However, the related methods from the parent `DataCommunicator` class are still available for use, namely:

- `setItemCountEstimate`
- `getItemCountEstimate`
- `setItemCountEstimateIncrease`
- `getItemCountEstimateIncrease`
- `setDefinedSize`
- `getDefinedSize`

### Expected behavior

These methods should be marked as unsupported.

### Minimal reproducible example

-

### Versions

-

PR #19784, opened 2024-08-13
fix: HierarchicalDataCommunicator doesn't support estimates

tepi:
Fixes https://github.com/vaadin/flow/issues/19772


tltv (approved):
I assume that previously it only could have worked with root items if none of them got expanded child items. Which means it was quite useless even if it worked somehow that way.

PR #19791, opened 2024-08-15
fix: use Thread context ClassLoader for loading I18n ResourceBundle

platosha:
- **fix: use Thread context ClassLoader for loading I18n ResourceBundle**
- **fix: make sure DefaultInstantiator also uses thread context ClassLoader for menu and i18n resources**

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/server/auth/DefaultInstantiatorMenuAccessControlTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/auth/DefaultInstantiatorMenuAccessControlTest.java
@@ -19,9 +19,12 @@ package com.vaadin.flow.server.auth;
 import java.lang.reflect.Field;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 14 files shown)


flow-server/src/test/java/com/vaadin/flow/server/auth/DefaultInstantiatorMenuAccessControlTest.java
```diff
@@ -30,18 +33,34 @@
 import com.vaadin.flow.server.InvalidMenuAccessControlException;
 import com.vaadin.flow.server.VaadinContext;
 import com.vaadin.flow.server.VaadinService;
+import org.mockito.invocation.InvocationOnMock;
 
 import static org.junit.Assert.assertThrows;
 
 public class DefaultInstantiatorMenuAccessControlTest {
```
caalador, line 40: This should perhaps be marked with `@NotThreadSafe` as we use a set current in the test

caalador (approved):

PR #19817, opened 2024-08-22
testing

tepi:
## Description

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
@@ -727,7 +727,7 @@ public abstract class AbstractNavigationStateRenderer
         NavigationEvent newNavigationEvent = getNavigationEvent(event,
                 beforeNavigation);
         newNavigationEvent.getUI().getPage().getHistory().replaceState(null,
-                newNavigationEvent.getLocation(), true);
+                newNavigationEvent.getLocation(), false);
 
         return handler.handle(newNavigationEvent);
     }


flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
```diff
@@ -724,10 +724,25 @@ private int forwardToExternalUrl(NavigationEvent event,
...
+
+        NavigationRouteTarget target = router.getRegistry()
+                .getNavigationRouteTarget(navigationState.getResolvedPath());
+
+        List<Class<? extends RouterLayout>> routerLayoutTypes = new ArrayList<>();
+        if (target.hasTarget()) {
+            routerLayoutTypes = target.getRouteTarget().getParentLayouts();
+        }
+
+        boolean preserveOnRefreshTarget = isPreserveOnRefreshTarget(
```
caalador, line 739: So is the issue that the target route is preserve on refresh or the forward route is preserve on refresh?
This feels like it mixes the 2 routes as the forward target has it's own parentLayouts, but here we are using the navigation target which is the current route.

If we just need to know if the original target is preservOnRefresh we could  just store the result from line 179 so that we can read it here as needed and do not need to do any extra checks and collections.

caalador: For the forward target we could perhaps just use:
```
        List<Class<? extends RouterLayout>> parentLayouts = RouteUtil.getParentLayouts(
                event.getUI().getRouter().getRegistry(),
                beforeNavigation.getForwardTargetType(),
                beforeNavigation.getForwardUrl());
```

tepi: Good catch, thanks! Fixed

caalador (approved):

PR #19835, opened 2024-08-27
Revert "fix: Make outlet tsx generated (#19824)"

caalador:
This reverts commit 1aa9c1586551af78a2e45952ddbc82a44dfc7ed6.


mshabarov:
So is this reverted now because https://github.com/vaadin/flow/pull/19826 is the only proper fix ?

caalador:
Yes. So the issue with the generated file is that it is generated after the bundle check has ben executed so the bundle can never find the file if the frontend generated folder has been cleaned.

mshabarov (approved):

PR #19845, opened 2024-08-28
fix: React file not added when reactEnabled false

### LINKED ISSUE #19842: ReactRouterOutletElement.tsx breaks build of Flow application using vaadin-router
### Description of the bug

If the Flow application is configured to use `vaadin-router` instead of `react-router` (e.g. by setting `reactEnable=false` in vaadin-maven-plugin), the production build fails on Vite execution because `react-router-dom` import cannot be resolved.

```
[ERROR] [vite]: Rollup failed to resolve import "react-router-dom" from "/xxxxx/vaadin/starters/skeleton-starter-flow-spring/src/main/frontend/generated/jar-resources/ReactRouterOutletElement.tsx".
[ERROR] This is most likely unintended because it can break your application at runtime.
[ERROR] If you do want to externalize this module explicitly add it to
```

The workaround is to exclude the `flow-react` dependency in the POM file.

If it is expected that `flow-react` should be excluded when using `vaadin-router`, it should be explicitly documented. In the [upgrade guide](https://vaadin.com/docs/latest/upgrading#hilla-react-dependencies) it looks like an optional step.

### Expected behavior

A Flow application using `vaadin-router` should not fail during build even if `flow-react` dependency is present.

- Clone https://github.com/vaadin/skeleton-starter-flow-spring/
- Update `vaadin-maven-plugin` configuration in POM file adding `<reactEnable>false</reactEnable>`
- Do a production build `mvn clean package -Pproduction`
- Check Vite build error in the console

### Versions

- Vaadin / Flow version: 24.5-SNAPSHOT
- Java version: 17

caalador:
Do not add the outlet
file if react is not
enabled as the required
react node files are not
available.

Fixes #19842
Fixes vaadin/start#3159

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -579,7 +579,8 @@ public class BuildFrontendUtil {
         }
 
         FrontendDependenciesScanner scanner = new FrontendDependenciesScanner.FrontendDependenciesScannerFactory()
-                .createScanner(false, adapter.getClassFinder(), true, null);
+                .createScanner(false, adapter.getClassFinder(), true, null,
+                        adapter.isReactEnabled());
         List<Product> commercialComponents = findCommercialFrontendComponents(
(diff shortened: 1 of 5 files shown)


mshabarov (approved):
A unit test would be good to have, but it probably require efforts to mock things that are mostly private.

PR #19873, opened 2024-09-03
fix: Validate layout values

### LINKED ISSUE #19872: Having multiple Layout with same value should throw
When having multiple `@Layout` classes having 2 or more with the same value should throw an exception.

caalador:
Check Layout annotations
for duplicate values.
If found throw exception.

Fixes #19872

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
@@ -17,7 +17,10 @@ package com.vaadin.flow.server.startup;
 
 import jakarta.servlet.annotation.HandlesTypes;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -116,6 +119,31 @@ public class RouteRegistryInitializer extends AbstractRouteRegistryInitializer
                     .collect(Collectors.joining(","));
             throw new InvalidRouteLayoutConfigurationException(
                     String.format(message, faultyLayoutsString));
+        }
+        Map<String, List<Class<?>>> layoutsMap = new HashMap<>();
+        for (Class<?> routeClass : routesSet) {
+            if (routeClass.isAnnotationPresent(Layout.class)) {
+                String layoutValue = routeClass.getAnnotation(Layout.class)
+                        .value();
+                layoutsMap.computeIfAbsent(layoutValue, k -> new ArrayList<>());
+                layoutsMap.computeIfPresent(layoutValue, (value, classList) -> {
+                    classList.add(routeClass);
+                    return classList;
+                });
+            }
+        }
+        Set<List<Class<?>>> collect = layoutsMap.values().stream()
+                .filter(entry -> entry.size() > 1).collect(Collectors.toSet());
+        if (!collect.isEmpty()) {
+            StringBuilder messageBuilder = new StringBuilder(
+                    "Found duplicate @Layout values in classes:");
+            for (List<Class<?>> classes : collect) {
+                messageBuilder.append("\n").append(" - ")
+                        .append(classes.stream().map(clazz -> clazz.getName())
+                                .collect(Collectors.joining(" - ")));
+            }
+            throw new InvalidRouteLayoutConfigurationException(
+                    messageBuilder.toString());
         }
     }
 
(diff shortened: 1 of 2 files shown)


tepi:
LGTM otherwise, but should the duplicates-not-allowed also be mentioned in `Layout` javadoc? At least I could not see it there.

flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
```diff
@@ -117,6 +120,31 @@ public static void validateLayoutAnnotations(Set<Class<?>> routesSet)
...
+        Map<String, List<Class<?>>> layoutsMap = new HashMap<>();
+        for (Class<?> routeClass : routesSet) {
+            if (routeClass.isAnnotationPresent(Layout.class)) {
+                String layoutValue = routeClass.getAnnotation(Layout.class)
+                        .value();
+                layoutsMap.computeIfAbsent(layoutValue, k -> new ArrayList<>());
+                layoutsMap.computeIfPresent(layoutValue, (value, classList) -> {
+                    classList.add(routeClass);
+                    return classList;
+                });
```
mcollovati, line 132: nit: can this part be simplified as  `layoutsMap.computeIfAbsent(layoutValue, k -> new ArrayList<>()).add(routeClass);`?

flow-server/src/test/java/com/vaadin/flow/server/startup/RouteRegistryInitializerTest.java
```diff
@@ -893,6 +906,24 @@ public void layout_annotation_on_non_routelayout_throws()
                 servletContext);
     }
 
+    @Test
+    public void same_layout_annotation_values_throws() {
+        StringBuilder messageBuilder = new StringBuilder(
+                "Found duplicate @Layout values in classes:");
+        messageBuilder.append("\n").append(" - ")
+                .append(AnnotatedParentLayout.class.getName()).append(" - ")
+                .append(AnotherAnnotatedParentLayout.class.getName());
```
mcollovati, line 915: I've got this test failing randomly because of the order of the classes in the error message is not predictable, most likely because of `HashSet`s are used in `RouteRegistryInitializer`

flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
```diff
@@ -117,6 +120,31 @@ public static void validateLayoutAnnotations(Set<Class<?>> routesSet)
...
+        }
+        Set<List<Class<?>>> collect = layoutsMap.values().stream()
+                .filter(entry -> entry.size() > 1).collect(Collectors.toSet());
+        if (!collect.isEmpty()) {
+            StringBuilder messageBuilder = new StringBuilder(
+                    "Found duplicate @Layout values in classes:");
+            for (List<Class<?>> classes : collect) {
+                messageBuilder.append("\n").append(" - ")
+                        .append(classes.stream().map(clazz -> clazz.getName())
+                                .collect(Collectors.joining(" - ")));
```
mcollovati, line 143: nit: a newline could make the message more readable
```suggestion
                                .collect(Collectors.joining("\n - ")));
```

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
caalador: This would perhaps work if you expected there to never be more than one pair or one conflicting layout.
When there is multiple I was thinking they should all be on one line so for 2 different conflicting layouts now would give:
```
 - Class1 - Class2 - Class5
 - Class3 - Class4
 ```
Where as with a new line it would become:
``` 
 - Class 1
 - Class 2
 - Class 5
 - Class 3
 - Class 4
```
Perhaps having something noting the different conflicting sets...

caalador:
> LGTM otherwise, but should the duplicates-not-allowed also be mentioned in `Layout` javadoc? At least I could not see it there.

Yes should be noted in the documentation. Basically you can have duplicates but only one will be then used which would lead to confusion on the layout chosen.

flow-server/src/main/java/com/vaadin/flow/server/startup/RouteRegistryInitializer.java
mcollovati: Got it. Definitely single line makes more sense

mcollovati (approved):

Issue #19831, opened 2024-08-26
PiT 24.5: BeforeEnterObserver is called in a Flow view when navigating to a Hilla view


manolo:
### Description of the bug

When the attached starter we navigate from a Flow view implementing BeforeEnterObserver to a Hilla view, ane exception happens because BeforeEnterObserver is called causing an exception since the route path of the next view is sent as a numeric parameter to the view.

```
java.lang.NumberFormatException: For input string: "hello-world"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67) ~[na:na]
	at java.base/java.lang.Long.parseLong(Long.java:711) ~[na:na]
	at java.base/java.lang.Long.parseLong(Long.java:834) ~[na:na]
	at java.base/java.util.Optional.map(Optional.java:260) ~[na:na]
	at com.example.application.views.masterdetail.MasterDetailView.beforeEnter(MasterDetailView.java:146) ~[classes/:na]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.sendBeforeEnterEvent(AbstractNavigationStateRenderer.java:620) ~[flow-server-24.5.0.alpha10.jar:24.5.0.alpha10]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.sendBeforeEnterEvent(AbstractNavigationStateRenderer.java:597) ~[flow-server-24.5.0.alpha10.jar:24.5.0.alpha10]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.sendBeforeEnterEventAndPopulateChain(AbstractNavigationStateRenderer.java:514) ~[flow-server-24.5.0.alpha10.jar:24.5.0.alpha10]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.createChainIfEmptyAndExecuteBeforeEnterNavigation(AbstractNavigationStateRenderer.java:480) ~[flow-server-24.5.0.alpha10.jar:24.5.0.alpha10]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.handle(AbstractNavigationStateRenderer.java:224) ~[flow-server-24.5.0.alpha10.jar:24.5.0.alpha10]
	at com.vaadin.flow.component.internal.JavaScriptNavigationStateRenderer.handle(JavaScriptNavigationStateRenderer.java:78) ~[flow-server-24.5.0.alpha10.jar:24.5.0.alpha10]
[...]

PR #19875, opened 2024-09-03
fix: prevent Flow navigation for Hilla route

tepi:
Prevents Flow navigation actions for cases where a client route exists and is navigated to, but there is also a Flow route with a parameter that could potentially match the target URL.

Fixes https://github.com/vaadin/flow/issues/19831

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
@@ -406,7 +406,7 @@ public class MenuRegistry {
     }
 
     /**
-     * See if there is a client route available for give route path.
+     * See if there is a client route available for given route path.
      *
      * @param route
      *            route path to check
@@ -416,6 +416,17 @@ public class MenuRegistry {
         return hasClientRoute(route, false);
     }
 
+    /**
+     * See if there is a client route available for given route path, optionally
+     * excluding layouts (routes with children) from the check.
+     *
+     * @param route
+     *            route path to check
+     * @param excludeLayouts
+     *            {@literal true} to exclude layouts from the check,
+     *            {@literal false} to include them
+     * @return true if a client route is found.
+     */
     public static boolean hasClientRoute(String route, boolean excludeLayouts) {
         if (VaadinSession.getCurrent() == null || route == null) {
             return false;
--- a/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
@@ -318,10 +318,12 @@ public class MenuRegistry {
             return getClassLoader().getResource(FILE_ROUTES_JSON_PROD_PATH);
         }
         try {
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
```diff
@@ -318,10 +318,12 @@ public static URL getViewsJsonAsResource(
             return getClassLoader().getResource(FILE_ROUTES_JSON_PROD_PATH);
         }
         try {
-            Path fileRoutes = configuration.getFrontendFolder().toPath()
-                    .resolve(GENERATED).resolve(FILE_ROUTES_JSON_NAME);
-            if (fileRoutes.toFile().exists()) {
-                return fileRoutes.toUri().toURL();
+            if (configuration.getFrontendFolder() != null) {
```
mshabarov, line 321: How frontend folder can be null ? Which scenario does this cover ?

mshabarov: The return value is implicitly not null and all the caller codes in Flow rely on it to be not null. I propose to mock it in the test.

Otherwise the PR looks good to me.

tepi: Fixed

mshabarov (approved):

mshabarov (changes requested):
You need to free the static mocks (e.g. use try-with-resources) for MockStatic, when mocking MenuRegistry.

Without it, we get:

```
Error:  Errors: 
Error:    JavaScriptBootstrapUITest.should_update_pushState_when_navigationHasBeenAlreadyStarted:441 Mockito 
For com.vaadin.flow.server.menu.MenuRegistry, static mocking is already registered in the current thread

To create a new mock, the existing static mock registration must be deregistered
```

mshabarov (approved):

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
```diff
@@ -408,13 +408,28 @@ public static ClassLoader getClassLoader() {
...
+     * excluding layouts (routes with children) from the check.
+     *
+     * @param route
+     *            route path to check
+     * @param excludeLayouts
+     *            {@literal true} to exclude layouts from the check,
+     *            {@literal false} to include them
+     * @return true if a client route is found.
+     */
+    public static boolean hasClientRoute(String route, boolean excludeLayouts) {
```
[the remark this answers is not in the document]
tepi: I found no way of testing this since it's a static method that calls other static methods, meaning I would have to mockstatic them all, which at the end does not test anything really. Most of the logic is from the original `hasClientRoute` method which had no unit tests either, probably because of the same reason.

tepi:
Seems there was no `MenuRegistry` class in 24.4 so CP is not possible.

caalador:
Also I guess the issue in https://github.com/vaadin/start/issues/3169 is also from this as the "" client route which comes from the `@layout.tsx` (this should be a layout not a route, but as there are not children)  is now seen as a actual client route which cancels the navigation in flow.

PR #19886, opened 2024-09-04
test: assert after browser resize event listener completes (2.11)

mcollovati:
Waits for the browser resize event listener to update the UI before asserting on the window width.


mcollovati:
Fixes Flow 2.11 nightly builds

tepi (approved):

Issue #12711, opened 2022-01-12
Frontend file copy should NOT copy the readonly bit


kklein-celera:
### Description of the bug

When copying files from the application's frontend resource directory, the readonly status of the copied files should NOT also be copied. Otherwise, you gets lots of these errors the next time the frontend is copied:

java.lang.IllegalArgumentException: File parameter 'destFile is not writable: 

The files in my application's src/main/resources/META-INF/frontend folder are read-only because they are managed by source control.

### Expected behavior

All files copied into the target/flow-frontend folder should be writable, regardless of the writable status in the source folder or jar.

### Minimal reproducible example

1. Create a CSS file in src/main/resources/META-INF/frontend and mark it as read-only using your underlying OS tools.
2. Run the application using "jetty run"
3. Notice that the CSS file copied to target/flow-frontend is also read-only.

### Versions

- Vaadin / Flow version: 22.0.2
- Java version: 11.0.2
- OS version: Windows 10
- Browser version (if applicable): N/A
- Application Server (if applicable): N/A
- IDE (if applicable): N/A

PR #19894, opened 2024-09-05
feat: Add option to not modify readonly files during copy, add logging of time

### LINKED ISSUE #19866: TaskCopyLocalFrontendFiles#copyLocalResources(File, File, String... ) takes a long time using Windows-OS
### Description of the bug

I'm using a windows pc and have noticed, that starting up our app takes longer than my colleges with mac.
The method com.vaadin.flow.server.frontend.TaskCopyLocalFrontendFiles#copyLocalResources takes up to a minute on windows and only 5 seconds on OSX.

We have found out, that the part that takes long ist what was added with https://github.com/vaadin/flow/issues/12711 . To be precise the code lines:
`try (Stream<Path> fileStream = Files
.walk(Paths.get(target.getPath()))) {
// used with try-with-resources as defined in walk API note
fileStream.filter(file -> !Files.isWritable(file)).forEach(
filePath -> filePath.toFile().setWritable(true));
}`

### Expected behavior

Could you please add an option/property to deactive the functionality of setting all files writeable.
None of our files are readonly and this is costing our developers lots of time.

not sure

### Versions

- Vaadin / Flow version: 24.4.4
- Java version: 17.0.9_9
- OS version: Windows 10 Pro 10.0.19045
- IDE (if applicable): IDEA 2024.1.4

TatuLund:
Allows to skip setting the writable flag on copied files, by providing the vaadin.frontend.disableWritableFlagCheckOnCopy system property.
This may improve performance in certain scenarios with Windows OS.

Fixes #19866

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyLocalFrontendFiles.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyLocalFrontendFiles.java
@@ -43,9 +43,9 @@ import org.slf4j.LoggerFactory;
 public class TaskCopyLocalFrontendFiles
         extends AbstractFileGeneratorFallibleCommand {
 
-    public static boolean PREVENT_READONLY_FILES = true;
-
     private final Options options;
+
+    private static boolean performReadonlyCheck = true;
 
     /**
      * Copy project local frontend files from defined frontendResourcesDirectory
@@ -55,6 +55,8 @@ public class TaskCopyLocalFrontendFiles
      */
     TaskCopyLocalFrontendFiles(Options options) {
         this.options = options;
+        performReadonlyCheck = !Boolean.valueOf(System.getProperty(
+                "vaadin.frontend.disableReadonlyCheckOnCopy", "false"));
     }
 
     @Override
@@ -103,7 +105,7 @@ public class TaskCopyLocalFrontendFiles
                     .getFilesInDirectory(source, relativePathExclusions));
             FileUtils.copyDirectory(source, target,
                     withoutExclusions(source, relativePathExclusions));
-            if (PREVENT_READONLY_FILES) {
+            if (performReadonlyCheck) {
                 try (Stream<Path> fileStream = Files
                         .walk(Paths.get(target.getPath()))) {
                     // used with try-with-resources as defined in walk API note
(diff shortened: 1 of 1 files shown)


thomasharre:
@TatuLund you have changed the default behaviour, I only wanted a possibility to turn it of.
 I suggest changing:
`PREVENT_READONLY_FILES = true;`

TatuLund:
@thomasharre Thanks for spotting that. I was testing this locally changing the value back and forth, it was wrong value by accident.

mcollovati:
What's the idea behind the usage of the static field? Is the developer supposed to change it in the application code?
If so, this will work for development mode, but will not help when the task is executed by Maven or Gradle.
Could a system property be a better option?

TatuLund:
> Could a system property be a better option?

Yes it would.

thomasharre:
We are only experiencing problems in certain developer environements.
Therefor the static field is the most pragmatic solution for us.

A system property would also work, but the current solution with public static field would be a fast way to help us.

mcollovati:
I still would prefer a system property, instead of the static field. Something like `vaadin.frontend.disableReadonlyCheckOnCopy` (feel free to suggest a better name), to make explicit that we are disabling an operation that was added to fix application failures (for example #12711)

thomasharre:
A system property would also work fine for me.

So the code in copyLocalResources would be something like?
`boolean performReadonlyCheck =  !Boolean.valueOf(System.getProperty("vaadin.frontend.disableReadonlyCheckOnCopy", "false"));
if (performReadonlyCheck){
...
}
`

@TatuLund would that suit you? If yes, could you please commit that?

thomasharre:
@TatuLund the member 'performReadonlyCheck' should not be static

TatuLund:
> the member 'performReadonlyCheck' should not be static

@thomasharre No, it must be as the method where it is used is static

thomasharre:
setting a static member in the constructor can't be right
then i would delete the member and introduce a new static method:

`  private static boolean performReadonlyCheck(){
    return !Boolean.valueOf(System.getProperty("vaadin.frontend.disableReadonlyCheckOnCopy", "false"));
  } `

thomasharre:
@TatuLund thanks

@mcollovati could you take another look and descide if this can be merged?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyLocalFrontendFiles.java
```diff
@@ -55,6 +55,11 @@ public class TaskCopyLocalFrontendFiles
         this.options = options;
     }
 
+    private static boolean isperformReadonly() {
```
mcollovati, line 58: nit: I would rename this method to something more explicit, like `shouldApplyWriteableFlag()`.
Maybe also rename the system property to `vaadin.frontend.disableWritableFlagCheckOnCopy`

Not a blocker, anyway.

mcollovati (dismissed):
Just a comment about naming, other than that LGTM

mcollovati (approved):

PR #19919, opened 2024-09-10
feat: add Vaadin i18n plugin for translation analysis

platosha:


caalador:
Seems fine, but needs running the formatter on the flow-server module.
Perhaps someone with better rollup knowledge could also have a look.

platosha:
Obsolete by #21480

PR #19933, opened 2024-09-11
feat: generate layout.json

caalador:
Genrate a layout.json for
Hilla to automatically
enable server layouts

part of vaadin/hilla#2709


tltv (commented):
- test is failing because `DevModeClassFinderTest` needs the Layout.class in knownClasses List
- cleanup should probably delete layout.json file?
- unit test for the new layout.json file and the expected content

caalador:
done.

PR #19934, opened 2024-09-11
chore: hash default application identifier

tltv:


flow-server/src/main/java/com/vaadin/flow/internal/MessageDigestUtil.java
```diff
@@ -75,6 +76,32 @@ public static byte[] sha256(String string, byte[] salt, Charset charset) {
...
+     *            the string to hash
+     *
+     * @return the hash as a hex string
+     */
+    public static String sha256ToHex(String string, Charset charset) {
+        byte[] bytes = MessageDigestUtil.sha256(string, charset);
+        return bytesToHex(Base64.getEncoder().encode(bytes));
+    }
+
+    private static String bytesToHex(byte[] hash) {
```
mcollovati, line 93: There's a similar private method in `StringUtil`. Would it make sense to make it public and reuse it instead of duplicating the code?

mcollovati (approved):
Overall LGTM.
I added a comment about potential code duplicate, but not a blocker.
Also, the `app-name` prefix looks a bit verbose. Could it simply be `app-`?

mcollovati:
Maybe the Javadoc of `PluginAdapterBase.applicationIdentifier()` should also mention that the default computed should be hashed.

flow-server/src/main/java/com/vaadin/flow/internal/MessageDigestUtil.java
tltv: Good point. I missed this one, so let's try to reuse it.

tltv: We can actually just use existing `StringUtil.getHash(string, charset)`. No real need for base64 encoding for this.

tltv: Done

mcollovati (approved):

PR #19846, opened 2024-08-28
chore: add premiumFeatures.enable token to build info

tltv:
Adds `premiumFeatures.enable` token to production build info.

Will not compile until license checker is updated.

PR #19935, opened 2024-09-11
fix: Propagate Premium Features flag value to configuration

mshabarov:
## Description

This adds missing codes into app configuration factory that propagates the PREMIUM_FEATURES flag value from token file to app config. This value can be retrieved from deployment configuration in runtime.

Related-to https://github.com/vaadin/flow/pull/19846

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
+++ b/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
@@ -30,6 +30,7 @@ import java.util.stream.Stream;
 import com.vaadin.flow.component.dependency.JavaScript;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
+import com.vaadin.flow.internal.StringUtil;
 import com.vaadin.flow.plugin.base.BuildFrontendUtil;
 import com.vaadin.flow.plugin.base.PluginAdapterBase;
 import com.vaadin.flow.plugin.base.PluginAdapterBuild;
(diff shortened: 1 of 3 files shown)


flow-server/src/test/java/com/vaadin/flow/server/startup/DefaultApplicationConfigurationFactoryTest.java
```diff
@@ -156,6 +156,28 @@ public void create_propertiesAreReadFromContext() throws IOException {
                 configuration.getStringProperty("foo", null));
     }
 
+    @Test
+    public void create_tokenFileWithPremiumFlag_premiumFlagIsPropagatedToDeploymentConfiguration()
```
tltv, line 160: Is this test adding anything that other tests here don't already test? This seems to just test same thing with different constant.

flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
```diff
@@ -30,6 +30,7 @@
 import com.vaadin.flow.component.dependency.JavaScript;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
+import com.vaadin.flow.internal.StringUtil;
```
tltv, line 33:
```suggestion
```

tltv (changes requested):

flow-server/src/test/java/com/vaadin/flow/server/startup/DefaultApplicationConfigurationFactoryTest.java
mshabarov: Apparently not, because no test was failing before, though this property wasn't populated :) 

Seems we don't have tests that checks that all possible properties are propagated from token file to config. I'll try to do such a test.

tltv: It's a good idea to add such a test. But it will not help with new properties, unless there is for example a list of properties that has to be updated for new properties. It could be in Constants for example.

mshabarov: This is problematic, as `AbstractConfigurationFactory` uses a subset of properties that are declared in `Constants` and `InitParameters`, thus, we need to refactor the properties names to group them accordingly, e.g. in separate classes or by marking with comments, if it works. Looks like a bigger work, than just to write one test.

PR #19941, opened 2024-09-12
chore: exclude transitive JNA dependency from org.eclipse.core.resources

mcollovati:


mcollovati:
Sonar check fails because it is executed with Java 17. We can probably ignore it for this PR

ZheSun88 (approved):

PR #19947, opened 2024-09-12
Revert "feat: Support top-level await"

Artur-:
Reverts vaadin/flow#18772


mshabarov (approved):

mshabarov:
Reverted as top level await causes problems that hard to diagnose and solve.

Issue #19905, opened 2024-09-06
MenuRegistry should have method to get List of ordered menu items


caalador:
MenuRegistry should have a method to get the available menu items as a list
that is ordered instead of getting only the current `Map<String, AvailableViewInfo>`.

This would mean that the list would be `List<AvaiableViewInfo>` and that the route
in the AvailableViewInfo should be fixed so it is the correct route and not only the view route for client routes.

Ordering would be done according to the AvailableViewInfo.Menu.order and/or alphabetically using `Collator.getInstance()` to get the same result as Hilla menu ordering

Legioth:
> Ordering would be done according to the AvailableViewInfo.Menu.order or alphabetically?

Same ordering as the [corresponding Hilla API](https://github.com/vaadin/hilla/blob/2eb52a85dbb6d4e67660455cca8eb04fe87fce3f/packages/ts/file-router/src/runtime/createMenuItems.ts#L30-L33). First order by the `order` property and then resolve ties by the alphabetic order of the route paths.

PR #19952, opened 2024-09-13
feat: add collectMenuItemsList() to MenuRegistry

tltv:
fixes: #19905

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
@@ -76,7 +76,13 @@ public class MenuRegistry {
      * @return routes with view information
      */
     public static Map<String, AvailableViewInfo> collectMenuItems() {
-        return new MenuRegistry().getMenuItems(true);
+        Map<String, AvailableViewInfo> menuRoutes = new MenuRegistry()
+                .getMenuItems(true);
+        menuRoutes.entrySet()
+                .removeIf(entry -> Optional.ofNullable(entry.getValue())
+                        .map(AvailableViewInfo::menu).map(MenuData::isExclude)
+                        .orElse(false));
+        return menuRoutes;
     }
 
     /**
--- a/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
@@ -21,12 +21,15 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.Path;
+import java.text.Collator;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -41,6 +44,7 @@ import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.router.BeforeEnterListener;
+import com.vaadin.flow.router.MenuData;
 import com.vaadin.flow.router.PageTitle;
 import com.vaadin.flow.router.RouteConfiguration;
 import com.vaadin.flow.router.RouteData;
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
```diff
@@ -75,6 +79,24 @@ public static Map<String, AvailableViewInfo> collectMenuItems() {
...
+     */
+    public static List<AvailableViewInfo> collectMenuItemsList() {
+        return collectMenuItems().entrySet().stream().map(entry -> {
+            AvailableViewInfo value = entry.getValue();
+            return new AvailableViewInfo(value.title(), value.rolesAllowed(),
+                    value.loginRequired(), entry.getKey(), value.lazy(),
+                    value.register(), value.menu(), value.children(),
+                    value.routeParameters(), value.flowLayout());
+        }).sorted(getMenuOrderComparator(
+                Collator.getInstance(Locale.forLanguageTag("en-US")))).toList();
```
knoobie, line 97: This might be a problem with GraalVM and their idea to have as small as possible executables.. `Collator.getInstance()` might be more random.. but probably safer

tltv: Am I understanding this right that you mean that usage of Collator instance for en-US could increase memory consumption compared to just using the default provided with `Collator.getInstance()`?

flow-server/src/test/java/com/vaadin/flow/server/menu/MenuRegistryTest.java
```diff
@@ -258,6 +260,13 @@ public void testWithLoggedInUser_userHasRoles() throws IOException {
 
         Assert.assertEquals(5, menuItems.size());
         assertClientRoutes(menuItems, true, true);
+
+        // Verify that getMenuItemsList returns the same data
+        List<AvailableViewInfo> menuItemsList = MenuRegistry
+                .collectMenuItemsList();
+        Assert.assertEquals(5, menuItemsList.size());
+        assertOrder(menuItemsList, new String[] { "", "/about", "/hilla",
+                "/hilla/sub", "/login" });
```
caalador, line 269: This test actually now shows that it is build wrong. /login is marked with `menu { "exclude": true }` which should lead to it not being in the menu items list at all.

tltv: This makes me wonder if this was so intentionally. If not, then I think I can add that missing filter for both collect* methods.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
knoobie: Correct, because GraalVM only includes locales / languages that are required by the application - if the application is e.g. completely in german, the above language might not be available - or GraalVM adds it if needed (this part I'm not sure.. I just remember seeing issues regarding this topic in the past)

flow-server/src/test/java/com/vaadin/flow/server/menu/MenuRegistryTest.java
caalador: Yes. I think menu items should not contain `menu: excluded` items as they are specifically excluded.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuRegistry.java
tltv: en-US matches with how [Hilla is doing it in Typescript here](https://github.com/vaadin/hilla/blob/2eb52a85dbb6d4e67660455cca8eb04fe87fce3f/packages/ts/file-router/src/runtime/createMenuItems.ts#L17)

We should sort same way in both places and currently it's hard coded in Hilla. If this causes trouble, it can be avoided by adding optional Locale parameter in the API. Default can be en-US. I'll add another method `collectMenuItemsList(Locale)`.

knoobie: Just FYI where my concern comes from:

> It is also possible to specify which locales should be included in the executable and which should be the default. For example, to switch the default locale to Swiss German and also include French and English, use the following options:

> native-image -Duser.country=CH -Duser.language=de -H:IncludeLocales=fr,en

> The locales are specified using [language tags](https://docs.oracle.com/javase/tutorial/i18n/locale/matching.html). You can include all locales via -H:+IncludeAllLocales, but note that it increases the size of the resulting executable.

https://www.graalvm.org/latest/reference-manual/native-image/dynamic-features/Resources/#locales

flow-server/src/test/java/com/vaadin/flow/server/menu/MenuRegistryTest.java
tltv: Done

flow-server/src/test/java/com/vaadin/flow/server/menu/MenuRegistryTest.java
```diff
@@ -257,7 +259,14 @@ public void testWithLoggedInUser_userHasRoles() throws IOException {
                 .getMenuItems(true);
 
         Assert.assertEquals(5, menuItems.size());
-        assertClientRoutes(menuItems, true, true);
+        assertClientRoutes(menuItems, true, true, false);
+
+        // Verify that getMenuItemsList returns the same data
+        List<AvailableViewInfo> menuItemsList = MenuRegistry
+                .collectMenuItemsList();
+        Assert.assertEquals(4, menuItemsList.size());
```
caalador, line 267: Nit. This could have a message that `Login is not expected` or  something that explains why the same data is one item less in this case.

caalador (approved):

Issue #19822, opened 2024-08-24
Multiple redundant invocations of the method com.vaadin.flow.router.HasUrlParameter#setParameter in Vaadin 24.4.9


alexanoid:
### Description of the bug

I have an AdministrationView:

```
@Route(value = "administration", layout = AdministrationLayout.class)
@PermitAll
public class AdministrationView extends BaseView implements BeforeEnterObserver {

    @Override
    public void beforeEnter(BeforeEnterEvent event) {

        removeAll();

        event.forwardTo(AdministrationRecruitersView.class);
    }

}
```

Available at the following URL:

http://localhost:8080/administration

As you can see, in the `beforeEnter `method, there is a forward to `AdministrationRecruitersView.class`.

The `AdministrationRecruitersView.class` itself is implemented as:

```
@Route(value = "administration/recruiters", layout = AdministrationLayout.class)
@PermitAll
public class AdministrationRecruitersView extends CatalogContainer<DecisionMatrix> implements HasUrlParameter<String> {

    @Override
    public void setParameter(BeforeEvent event, @OptionalParameter String jobIdParameter) {
    }

}
```

So, when navigating to AdministrationView (by http://localhost:8080/administration) and subsequently auto forwarding to AdministrationRecruitersView, the method AdministrationRecruitersView.setParameter() is called **3 times in a row**. If I navigate directly to AdministrationRecruitersView, i.e., to http://localhost:8080/administration/recruiters, the method AdministrationRecruitersView.setParameter() is correctly called only once, as it should be.

This issue started appearing in version 24.4.9, while in version 24.4.0.beta5 everything works correctly. I haven't changed anything in my code. 

Also, in version 24.4.9, the bug from https://github.com/vaadin/flow/issues/18736 has reappeared, but it’s not present in version 24.4.0.beta5, where everything works fine.

### Expected behavior

no redundant invocation of com.vaadin.flow.router.HasUrlParameter#setParameter

### Minimal reproducible example

n/a

### Versions

- Vaadin / Flow version: 24.4.9
- Java version: 17

alexanoid:
Sorry to bother you, but is there a chance this will be fixed anytime soon?

tepi:
Seems this is another issue with the `react-router` performing the full navigation cycle on server side when forwarding is used, and after that doing a `ui-navigate` event callback to the server which triggers the `setParameter` for a second time. That said, in my testing I only got two calls to `setParameter`, not three.

[...]

Issue #19813, opened 2024-08-21
replaceState does not update the URL in all scenarios when using PreserveOnRefresh


jameskerrtaskize:
### Description of the bug

I previously raised https://github.com/vaadin/flow/issues/19701 with a minimal reproducible example. That example was fixed but our actual use case requires a slightly more complicated reproducible example which I have attached to this ticket.

This issue is stopping us from being able to upgrade to 24.4

The example provided uses replaceState to update the URL. I think the issue also occurs when using forwarding. We switched to replaceState due to an earlier issue with PreserveOnRefresh which meant forwarding did not update the URL. Now neither mechanism works in this scenario.

launch the app with:

mvn jetty:run

Navigate to:

http://localhost:8080/parent/?blah=blah

You will find that the URL changes to:

http://localhost:8080/parent/child?blah=blah

but the desired outcome is that the URL changes to:

http://localhost:8080/parent/child?corrected=true

In ChildRoute.java you can delete the @PreserveOnRefresh and then repeat the test and it will work as desired.

Also, when @PreserveOnRefresh is enabled, you will see that you get two beforeEnter callbacks in ChildRoute but without it enabled, you get 1 as expected.

### Expected behavior

replaceState should update the URL in all instances

### Minimal reproducible example

[preserveonrefresh-issue2.zip](https://github.com/user-attachments/files/16691794/preserveonrefresh-issue2.zip)

### Versions

- Vaadin / Flow version: 24.4.x (tested on 24.4.9)

tepi:
The issue was assigned to a developer and is currently being investigated

Issue #19794, opened 2024-08-16
URL is not changed when using `forwardTo` to the view just navigated from


MatthewVaadin:
### Description of the bug

Forwarding to the same view that the user has just navigated from doesn't change the URL.
For example, suppose we have two views, `FirstView.java` (path `/first`) and `SecondView.java` (path `/second`). The second view calls `forwardTo(FirstView.class)` in `beforeEnter`.  If the user is currently on `/first` and they navigate to `/second`, they are correctly forwarded back to the first view, but the URL is still shown as `/second`.

### Expected behavior

The URL should be shown as the view that the user has been forwarded to (`/first` in the case above).

### Minimal reproducible example

See https://github.com/MatthewVaadin/forward-bug

- Navigate to the "First" view
- Navigate to the "Second" view
- Note that the view is "First", but the URL path is `/second`

### Versions

- Vaadin / Flow version: 24.4.8
- Java version: 17
- OS version: Windows 11
- Browser version (if applicable): Chrome 127.0.6533.120

tepi:
Confirmed to fail on 24.5-SNAPSHOT with both vaadin router and react router.

PR #19959, opened 2024-09-16
fix: forward with optional callback

tepi:
Fixes https://github.com/vaadin/flow/issues/19794
Fixes https://github.com/vaadin/flow/issues/19822

Also fixes https://github.com/vaadin/flow/issues/19813 in a more coherent way. Relevant test is `com.vaadin.flow.uitest.ui.PreserveOnRefreshForwardingIT`

### DIFF
--- a/flow-tests/test-react-router/src/test/java/com/vaadin/flow/ForwardTargetIT.java
+++ b/flow-tests/test-react-router/src/test/java/com/vaadin/flow/ForwardTargetIT.java
@@ -2,6 +2,7 @@ package com.vaadin.flow;
 
 import org.junit.Assert;
 import org.junit.Test;
+import org.openqa.selenium.TimeoutException;
 
 import com.vaadin.flow.component.html.testbench.SpanElement;
 import com.vaadin.flow.testutil.ChromeBrowserTest;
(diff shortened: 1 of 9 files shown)


flow-tests/test-react-router/src/test/java/com/vaadin/flow/ForwardTargetIT.java
```diff
@@ -0,0 +1,43 @@
...
+public class ForwardTargetIT extends ChromeBrowserTest {
+
+    public static final String FORWARD_TARGET_VIEW = "/view/com.vaadin.flow.ForwardTargetView";
+
+    // Test for https://github.com/vaadin/flow/issues/19794
+    @Test
+    public void testUrlIsCorrectAfterForward() {
+        getDriver().get(getTestURL(getRootURL(), FORWARD_TARGET_VIEW, null));
+
+        waitUntil(arg -> driver.getCurrentUrl().endsWith(FORWARD_TARGET_VIEW));
```
caalador, line 18: The waitUntil is a horrible  in that it just fails with a `TimeoutException` it would be nice if (especially the second one) used a `try {...} catch (TimeoutException te) { Assert.fail("Url wasn't updated to expected url " + FORWARD_TARGET_VIEW); }` as at the moment it will never actually come to the final assert if it is failing.

We should ofc fix the TestBench waitUntil to take an custom error message that would be thrown on timeout to make this easier for the developer to add expectation outside of just TimeoutError.

caalador (approved):

PR #19965, opened 2024-09-17
feat: add copilot ui-tests generator as an experimental-feature

manolo:
related to: https://github.com/vaadin/platform/issues/6799


yuriy-fix (changes requested):
We probably don’t need to add a feature flag entry for this one. The user can simply avoid clicking the button if they don't want to use it, and everything will still work as expected, even with the plugin enabled and visible.

However, within the plugin itself, we can add a note mentioning that it’s still experimental.

manolo:
Ok, let keep this as a draft for now, we can close later

Issue #19951, opened 2024-09-13
Toggling react.enable flag to false breaks application


tepi:
### Description of the bug

Just did a quick test with simple Flow+SpringBoot application downloaded from start.vaadin.com.

First I ran the application as-is, it runs and confirmed react-router is used.

Then I set `vaadin.react.enable=false` in `application.properties` and restarted the application. This leads into exception in vite:

```
ERROR(TypeScript)  Cannot find module 'react-router-dom' or its corresponding type declarations.
FILE  /Users/tepi/Downloads/skeleton-starter-flow-spring-24 2/src/main/frontend/generated/vaadin.ts:42:24

   40 | applyTheme(document);
   41 |
 > 42 | import { Outlet } from 'react-router-dom';
      |                        ^^^^^^^^^^^^^^^^^^
   43 | (window as any).Vaadin ??= {};
   44 | (window as any).Vaadin.copilot ??= {};
   45 | (window as any).Vaadin.copilot._ref ??= {};
```

After this I ran `mvn vaadin:clean-frontend` and restarted the app. App runs without errors and I confirmed vaadin-router is being used.

### Expected behavior

After toggling `react.enable` to `false`, Flow should automatically figure out what to do to make the application run without frontend build failures.

### Minimal reproducible example

See description for instructions

### Versions

- Vaadin version: 24.4.11

mshabarov:
We may try to extract a subset of config parameters that potentially affect frontend compilation (where the `react.enable` is among them) and for this group of parameters Flow always deletes `frontend/generated`. On the other hand, files in `jar-resources` are not necessarily need an update after configuration change, but rather after adding/removing JAR dependencies to the project.

tltv:
Copilot adds that import to `vaadin.ts` when Flow runs `TaskGenerateBootstrap`. `TaskGenerateBootstrap` runs before `TaskRunNpmInstall`. Copilot checks if `react-router-dom` module exists in the `node_modules` and if it does, then it adds this import. It's not checking if react is enabled via `options.isReactEnabled()`. I think it should. @Artur- could you check if this can be fixed in `CopilotLoader`?

In more detail, once you build vaadin app with react enabled, it naturally installs `react-router-dom` module. If you run app again after disabling react, it's not installed anymore in the end as expected. Second run and following runs will work again because now Copilot is not adding imports automatically due to `react-router-dom` doesn't exist any more.

[...]

Issue #19911, opened 2024-09-09
Browser back navigation not working on 24.4.11 (on react-router)


FabianEhr:
### Description of the bug

After upgrading to Vaadin 24.4.3 and disabling the vaadin-react feature within the properties the browser back navigation isn't working. The following error can be aquiered from the developer console.

Errorlog
`indexhtml-C98SXJYq.js:62437 You are trying to use a blocker on a POP navigation to a location that was not created by @remix-run/router. This will fail silently in production. This can happen if you are navigating outside the router via `window.history.pushState`/`window.location.hash` instead of using router navigation APIs.  This can also happen if you are using createHashRouter and the user manually changes the URL.`

### Expected behavior

The browser back navigation should work.

### Minimal reproducible example

Currently we got no time to provide a example project.
We demonstrated the bug to Sascha Merkel and Sebastian Kühnau today.

### Versions

- Vaadin / Flow version: 24.3.13 -> 24.4.3 &  also 24.4.11
- Java version: 17 
- OS version: Windows 11 and Ubuntu 24
- Browser version (if applicable): Chrome Version 128.0.6613.120
- Application Server (if applicable): Wildfly 30.0.0
- IDE (if applicable): IntelliJ IDEA 2024.1.4 (Ultimate Edition) Build #IU-241.18034.62

knoobie:
Did you try the latest version?

mcollovati:
Looking at the error mentioning `@remix-run/router`, it seems react-router has not been disabled.
Please try to clean the project with `mvn clean clean:frnontend` and make sure `vaadin-router` is used.

Also, as Knoobie pointed out, Vaadin 24.4.3 is a bit old, the latest version is 24.4.11; several fixes have been released since 24.4.3.

tepi:
There are currently some timing/async related issues when using react router with the latest Vaadin versions. We are working on the issues but solution is not there yet. For now, if you can, change to using vaadin router, that should resolve the issues.

FabianEhr:
We upgraded the version to 24.4.11 as recommended. The problem still exists on that version.
Also we checked that the react-router is disabled and performed several clean-installs which had no effect on the result.

We try to assemble a demo project which has show cases the error.

FabianEhr:
Issue can be closed.
The project had a secondary property which enabled the react-routing again.

tepi:
Reopening this for us since it's still an issue on react-router

simasch:
Even if I disable React with

[...]

PR #19966, opened 2024-09-17
fix: Also take into account the application.properties value

caalador:
When checking for react enable
also look if we have a application.properties
with the vaadin.react.enable flag.

This is especially for production build when only
application.properties has react disabled.

Touches #19951 and #19911

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Properties;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.regex.Matcher;
@@ -1266,7 +1267,8 @@ public class FrontendUtils {
      *            path to the frontend folder in a project.
      * @return {@code false} if vaadin-router is used, {@code true} otherwise.
      */
-    public static boolean isReactRouterRequired(File frontendDirectory) {
+    public static boolean isReactRouterRequired(File frontendDirectory,
+            File buildDirectory) {
         Objects.requireNonNull(frontendDirectory);
         boolean result = true;
         File indexTs = new File(frontendDirectory, FrontendUtils.INDEX_TS);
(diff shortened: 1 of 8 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -1282,13 +1284,38 @@ public static boolean isReactRouterRequired(File frontendDirectory) {
                         e);
             }
         }
+        if (result) {
+            // If we are enabling react we should for spring projects check the
+            // application.properties file flag also.
+            File resource = new File(buildDirectory,
+                    "classes/application.properties");
```
knoobie, line 1291: application.yml/yaml are not supported?

caalador: This is more of a test at the moment and as flow doesn't have any yaml handling libraries it would be a build by hand parsing.
Also not certain if the linking to spring specific property files is a good idea as it might lead to expectations in overall configuration.

tltv: This would also leave out all profile specific properties files `application-{profile}.properties` like `application-prod.properties`, `application-dev.properties` etc.

knoobie: yeah that's also true.. I'm wondering if there could be some kind of way to get the available spring environment in the build so that spring would handle all their "property magic"

caalador: The main problem with possibly trying to use spring tools is that the data is needed during plugin build time also and not just runtime.

knoobie: Understandable.. couldn't find a quick solution as well. I'm wondering if the react flag should have never been in the application.properties and should only be promoted as maven / gradle setting.

caalador: The reason why it is also in the application.properties is that if you run the application class from for instance idea it doesn't read the maven/gradle configuration.

caalador: One possibility could be to add to starter projects into pom.xml `<properties><reactEnable>true</reactEnable></properties>` and have in the default application.properties `vaadin.react.enable=@reactEnable@`.
This would sync the pom.xml and application.properties values for the flag.

The problem here is ofc that it could be removed by mistake and wouldn't help old projects.

knoobie: Doesn't this require additional configuration? Like filtering of resources to work?

caalador: Actually it works out of the box for spring-boot: https://docs.spring.io/spring-boot/docs/2.0.0.M7/reference/html/howto-properties-and-configuration.html#howto-automatic-expansion-maven

`If you use the spring-boot-starter-parent, you can then refer to your Maven ‘project properties’ with @..@ placeholders`

tltv: `vaadin.react.enable=@reactEnable@` looks good idea, but polluting starters with a property that most don't need is not so nice.
If the point is to avoid issues with `vaadin.react.enable` being added only to `application.properties` then better documentation could be enough. Also this `@reactEnable@` should be mentioned somewhere.

Issue #12640, opened 2021-12-27
Unexpected message id from the client (sync ids different)


thomaskiesl:
### Description of the bug

I receive sometimes the same error message on my production system. But I do not have any idea when this happen:

`2021-12-27 07:27:43,174  INFO | https-jsse-nio-18443-exec-10 | com.vaadin.flow.server.communication.ServerRpcHandler   | Ignoring old duplicate message from the client. Expected: 1748, got: 1747`

```
2021-12-27 07:28:01,354  WARN | https-jsse-nio-18443-exec-5 | de.app.ApplicationVaadinServiceInitListener         | Unexpected error
java.lang.UnsupportedOperationException: Unexpected message id from the client. Expected sync id: 1751, got 1752. more details logged on DEBUG level.
        at com.vaadin.flow.server.communication.ServerRpcHandler.handleRpc(ServerRpcHandler.java:310)
        at com.vaadin.flow.server.communication.UidlRequestHandler.synchronizedHandleRequest(UidlRequestHandler.java:115)
        at com.vaadin.flow.server.SynchronizedRequestHandler.handleRequest(SynchronizedRequestHandler.java:40)
        at com.vaadin.flow.server.VaadinService.handleRequest(VaadinService.java:1565)
        at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:302)
        at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:109)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:750)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:227)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162)
        at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:189)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162)
        at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:327)
        at org.springframework.security.web.access.intercept.FilterSecurityInterceptor.invoke(FilterSecurityInterceptor.java:115)
        at org.springframework.security.web.access.intercept.FilterSecurityInterceptor.doFilter(FilterSecurityInterceptor.java:81)
[...]

PR #19983, opened 2024-09-18
fix: delay session expiration handling to prevent canceling ongoing navigation

mcollovati:
Attempts to fix the synchronization issue related to the usage of the Login reported in #12640.
The Login component sends the UIDL request for the `login` event to the server and concurrently submits the form.
If processing the form submission performs a session ID change and a request redirect, the UIDL requests might fail with a session expiration response. The Flow client then can cancel the first redirect because it reloads the page due to the session expiration. Lastly, the beacon request hits again a valid session, but a resynchronization is triggered because the previous UIDL request was rejected.

This change delays a bit the session expiration handling on Flow client, to allow a potential redirect to complete without being cancelled. However, the client application is immediately set in TERMINATED state.


mshabarov (approved):
Cannot say if this is the best solution, but cannot suggest anything else either.
I've made a quick smoke test with CRM example, where I logged in and out and also used short session expiration. Couldn't see any anomalies.

250ms delay sounds fine in this context, when a user does a log-in or session is expired, some delay is expected in such situations. 0-300ms are usually not very visible for user.

mshabarov:
Preferably to apply [this fix ](https://github.com/vaadin/flow-components/pull/6669) first.

flow-tests/test-frontend/vite-embedded-webcomponent-resync/src/test/java/com/vaadin/viteapp/BasicComponentIT.java
```diff
@@ -65,17 +65,18 @@ public void session_resynced_webcomponent_is_active() throws Exception {
...
-        waitForWebComponent("login-form");
 
         // init request to resynchronize expired session and recreate components
         clickButton();
 
+        waitForWebComponent("login-form");
         try {
             // it seems WebDriver needs also sync to new session
             setUsername("");
         } catch (StaleElementReferenceException ex) {
```
caalador, line 76: Just as a side note, this check could be better using the actual expected of selenium so instead of try catch use
```
TestBenchElement input = $("login-form").first().$("input").first();
waitUntil(ExpectedConditions.stalenessOf(input));
```

mcollovati: Do you mean replacing the `setUsername()` part or the `waitForWebComponent`?

caalador: Actually looking at it now I don't know exactly what it is we are waiting for.
I expected that we wait for the components to reconnect making the old reference stale, which would mean we have reloaded and can try the inputs.

mcollovati: `ExpectedConditions.stalenessOf()` seems to work well. Thanks for the suggestion @caalador.

caalador (approved):

PR #19984, opened 2024-09-19
fix: close overlay drawer on click

### LINKED ISSUE #19823: Vaadin 24.4.9 RouterLink in Drawer
### Description of the bug

Previous similar bug https://github.com/vaadin/flow/issues/18736

With Vaadin 24.4.9, when I click on the RouterLink which points to the page where I currently am, the Drawer on the mobile view does not close, unlike in version 24.4.0.beta5 :

drawer

### Expected behavior

The drawer on the mobile view should be closed after the click, even if it is pointed to the page where the user currently is

n/a

### Versions

- Vaadin / Flow version:  24.4.9
- Java version: 17

caalador:
Close the overlay drawer
even when clicking on
same link and no actual
navigation happens.

Fixes #19823


tepi (approved):
Tested to work fine once again

PR #19985, opened 2024-09-19
fix: queued navigate with React Router

### LINKED ISSUE #19839: Browser back navigation does not work when using replaceState with React enabled
### Description of the bug

If the current page used replaceState to modify the page history in an afterNavigation observer (for example, to add a query parameter), page history gets messed up. This causes back navigation using the browser back button to go back to 2 pages ago, instead of one page ago.

This worked in v23 but is broken in v24. A workaround is to disable React.

### Expected behavior

Back navigation should only go back one page.

[react-router-back-nav.zip](https://github.com/user-attachments/files/16768381/react-router-back-nav.zip)

Run using `mvn jetty:run` and open localhost:8080.

### Versions

- Vaadin / Flow version: 24.4.10

platosha:
Fixes #19839


caalador:
Seems that when we are not logged in and click the `Navigate to admin view in 1 second` button in the tests we do navigate to `/admin` then `/my/login/page` which seems fine but then we get one more navigate to `/my/login/page` which does a `proceed` on the blocker even though we have a server roundtrip going which then calls `proceed` again after the last call already let the blocker proceed and is already in the `unblocked` state

caalador:
Actually it seems it is `/admin` goes to the server and then `/my/login/page` releases the blocker before the /admin returns. Blocking the last my/login call still seems to create the blocker state fault.

caalador:
Just as a test dopping any blocker requests when a server roundtrip is waiting does fix the error, but I could expect this to have some drawbacks for other requests in the set requiring a roundtrip when multiple changes are made
```
diff --git forkSrcPrefix/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx forkDstPrefix/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
index b073570de0e7926bcdb2b37ec34abc815e8d1297..c03527c624525b9b362e96e8570510344f197bbb 100644
--- forkSrcPrefix/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ forkDstPrefix/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -257,6 +257,7 @@ function Flow() {
     const {pathname, search, hash} = useLocation();
     const navigated = useRef<boolean>(false);
     const fromAnchor = useRef<boolean>(false);
+    const serverRoundtrip = useRef<boolean>(false);
     const containerRef = useRef<RouterContainer | undefined>(undefined);
     const queuedNavigate = useQueuedNavigate();
 
@@ -306,11 +307,15 @@ function Flow() {
     const vaadinNavigateEventHandler = useCallback((event: CustomEvent<{state: unknown, url: string, replace?: boolean, callback: boolean}>) => {
         const path = '/' + event.detail.url;
         navigated.current = !event.detail.callback;
+        if(event.detail.replace && path === window.location.pathname) {
+            return;
+        }
         queuedNavigate(path, { state: event.detail.state, replace: event.detail.replace});
     }, [navigate]);
 
     const redirect = useCallback((path: string) => {
         return (() => {
+            serverRoundtrip.current = false;
             navigate(path, {replace: true});
         });
     }, [navigate]);
@@ -339,12 +344,16 @@ function Flow() {
 
     useEffect(() => {
         if (blocker.state === 'blocked') {
+            if(serverRoundtrip.current) {
+                return;
+            }
             // Do not skip server round-trip if navigation originates from a click on a link
             if (navigated.current && !fromAnchor.current) {
                 blocker.proceed();
                 return;
             }
             fromAnchor.current = false;
+            serverRoundtrip.current = true;
             const {pathname, search} = blocker.location;
             const routes = ((window as any)?.Vaadin?.routesConfig || []) as AgnosticRouteObject[];
             let matched = matchRoutes(Array.from(routes), pathname);
@@ -357,10 +366,12 @@ function Flow() {
                         prevent() {
                             blocker.reset();
                             navigated.current = false;
+                            serverRoundtrip.current = false;
                         },
                         redirect,
                         continue() {
                             blocker.proceed();
+                            serverRoundtrip.current = false;
                         }
                     }, router);
                 navigated.current = true;
@@ -376,13 +387,16 @@ function Flow() {
                             containerRef.current.serverConnected = (cancel) => {
                                 if (cancel) {
                                     blocker.reset();
+                                    serverRoundtrip.current = false;
                                 } else {
                                     blocker.proceed();
+                                    serverRoundtrip.current = false;
                                 }
                             }
                         } else {
                             // permitted navigation: proceed with the blocker
                             blocker.proceed();
+                            serverRoundtrip.current = false;
                         }
                     });
             }
```

caalador (dismissed):

mcollovati (approved):

PR #20007, opened 2024-09-20
chore: More detailed logging for ignored handling

caalador:
Make log more detailed for
ignored invocation handle.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
@@ -724,6 +724,14 @@ public class ComponentUtil {
         return router;
     }
 
+    /**
+     * Walk up from given component until a Component with a Route annotation is
+     * found or empty if no Route is present in parents.
+     *
+     * @param component
+     *            Component to find current route component for
+     * @return Optional containing Route component if found
+     */
     public static Optional<Component> getRouteComponent(Component component) {
         if (component.getClass().isAnnotationPresent(Route.class)) {
             return Optional.of(component);
--- a/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
@@ -45,6 +45,7 @@ import com.vaadin.flow.internal.StateNode;
 import com.vaadin.flow.internal.StateTree;
 import com.vaadin.flow.internal.nodefeature.ElementData;
 import com.vaadin.flow.internal.nodefeature.VirtualChildrenList;
+import com.vaadin.flow.router.Route;
 import com.vaadin.flow.router.Router;
 import com.vaadin.flow.server.Attributes;
 import com.vaadin.flow.server.VaadinService;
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
```diff
@@ -723,4 +724,14 @@ public static Router getRouter(HasElement component) {
         return router;
     }
 
+    public static Optional<Component> getRouteComponent(Component component) {
```
tltv, line 727: JavaDoc missing from a public API.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -59,21 +64,48 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
+                    targetInfo.append(" Route: ").append("'")
+                            .append(routeComponent.get().getClass()
+                                    .getAnnotation(Route.class).value())
+                            .append("'");
+                }
+            }
+            getLogger().info(
+                    "Ignored RPC for invocation handler '{}' from "
+                            + "the client side for an {} node id='{}' {}",
+                    getClass().getName(), reason, node.getId(), targetInfo);
```
tltv, line 100: Both if-else blocks ends with mostly duplicate lines of code so it can be moved out side the if-else blocks in the end of the method and `targetInfo` could be blank by default.

caalador: done

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
caalador: added

mcollovati: Maybe not for this PR, but we have a `findAncestor(Class)` method in `Component` class.
It could be useful to have also a `findAncestor(Predicate<Component>)` for similar cases

tltv (approved):

flow-server/src/main/java/com/vaadin/flow/component/ComponentUtil.java
caalador: Perhaps if there comes up a need for this.

flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -59,21 +64,43 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
...
+            if (component.isPresent()) {
+                targetInfo.append(" Component: ").append("'")
+                        .append(component.get().getClass().getName())
+                        .append("'");
+                Optional<Component> routeComponent = ComponentUtil
+                        .getRouteComponent(component.get());
+                if (routeComponent.isPresent()) {
+                    targetInfo.append(" Route: ").append("'")
+                            .append(routeComponent.get().getClass()
+                                    .getAnnotation(Route.class).value())
```
stefanuebe, line 93: I assume this will not work with dynamically registered routes using the RouteConfig?

caalador: No that would need checking each parent against the registry which sounds like too much for a default info log message. Perhaps if running a trace log ...

mcollovati (approved):

PR #20014, opened 2024-09-20
chore: HMR for vaadin-react.tsx

Artur-:


Artur-:
As far as I can tell, the `window.Vaadin.routesConfig` value is only used in once place in the router, and there it is never cached

platosha (approved):

PR #20018, opened 2024-09-21
chore: Node 22.9.0

Artur-:
Node 22 will become LTS well before the 24.6 release


mshabarov:
If no objections, I will merge this.

Artur-:
> Also, I'd recommend to update npm to latest in one go (to 10.8.3), but not a blocker.

It is using whatever the node version bundles, right?

mshabarov:
Hmm, that's correct and the final static version is used only in tests, and only a major and minor. Interesting. Not that important then.

PR #20019, opened 2024-09-23
test: increase sleep time to allow GC to perform cleanup

mcollovati:
On CI systems the test randomly fails because the GC seems not to clean up ComponentLocator memory within the 5 milliseconds. However, running the test in isolation works fine. This change increases the delay between the GC requests to try to make the test more reliable.

### DIFF
--- a/flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
@@ -155,12 +155,11 @@ public class ComponentTrackerTest {
 
     private boolean isCleared(Map<?, ?> map) throws InterruptedException {
         for (int i = 0; i < 5; i++) {
-            System.out.println(i);
             System.gc();
             if (map.isEmpty()) {
                 return true;
             }
-            Thread.sleep(10);
+            Thread.sleep(100);
         }
         return false;
     }
--- a/flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
@@ -155,11 +155,12 @@ public class ComponentTrackerTest {
 
     private boolean isCleared(Map<?, ?> map) throws InterruptedException {
         for (int i = 0; i < 5; i++) {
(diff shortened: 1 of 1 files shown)


flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
```diff
@@ -155,11 +155,12 @@ public void memoryIsReleased() throws Exception {
 
     private boolean isCleared(Map<?, ?> map) throws InterruptedException {
         for (int i = 0; i < 5; i++) {
+            System.out.println(i);
```
caalador, line 158: System out is perhaps not needed here

flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
```diff
@@ -155,11 +155,12 @@ public void memoryIsReleased() throws Exception {
...
             System.gc();
-            if (map.size() == 0) {
+            if (map.isEmpty()) {
                 return true;
             }
-            Thread.sleep(1);
+            Thread.sleep(10);
```
caalador, line 163: Sleep is in milliseconds and not seconds. I feel 100 ms or even 1000 ms would be fine for the test.

flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
mcollovati: Ops, leftover. I'll remove it, thanks.

flow-server/src/test/java/com/vaadin/flow/ComponentTrackerTest.java
mcollovati: I suppose it was intended to have a very little delay, to make sure memory is released as soon as possible.
1 second seems a too much to me, I would go at most with 100ms.

mcollovati: Now I realized the commit message was wrong, mentioning 5 seconds instead of milliseconds.
Sorry for the confusion.

caalador (approved):

PR #20027, opened 2024-09-23
feat: Report application URL when application is started

Artur-:
This is useful in a couple of cases:
1. Browser launcher is not enabled
2. Browser launcher does not work
3. You closed the browser tab and a new one is not automatically reopened when you restart

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerManagerImpl.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerManagerImpl.java
@@ -68,6 +68,9 @@ public class DevModeHandlerManagerImpl implements DevModeHandlerManager {
     private BrowserLauncher browserLauncher;
     final private Set<Closeable> watchers = new HashSet<>();
 
+    private String applicationUrl;
+    private boolean fullyStarted = false;
+
     @Override
     public Class<?>[] getHandlesTypes() {
         return DevModeStartupListener.class.getAnnotation(HandlesTypes.class)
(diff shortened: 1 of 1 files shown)


mcollovati:
> Browser launcher is not enabled
If I set `vaadin.launch-browser=false` or do not add the entry at all in `application.properties`, the URL is not printed

Artur-:
True, Spring never reports the URL to the dev server if the launch setting is off. Fixed now

vaadin-spring/src/main/java/com/vaadin/flow/spring/DevModeBrowserLauncher.java
```diff
@@ -82,8 +84,9 @@ private void launchBrowserInDevelopmentMode(ApplicationContext appContext) {
         DevModeHandlerManager devModeHandlerManager = lookup
                 .lookup(DevModeHandlerManager.class);
         if (devModeHandlerManager != null) {
-            devModeHandlerManager
-                    .launchBrowserInDevelopmentMode(getUrl(webAppContext));
+            String url = getUrl(webAppContext);
+            devModeHandlerManager.setApplicationUrl(url);
+            devModeHandlerManager.launchBrowserInDevelopmentMode(url);
```
mcollovati, line 89: Shouldn't this be controlled by the `launch` parameter?

Artur-: Yes... shouldn't be this hard to code. Next time I will ask some AI to do it

mcollovati (approved):

PR #20054, opened 2024-09-25
chore: Add possibility to register shutdown commands for development mode

abdullahtellioglu:
## Description

Sync file watcher should be closed on context shut down so adding watchers to dev tools logical since it handles that 

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
@@ -51,7 +51,7 @@ public interface DevModeHandlerManager {
      * @param classes
      *            classes to check for npm- and js modules
      * @param context
-     *            servlet context we are running in
+     *            Vaadin Context we are running in
      *
      * @throws VaadinInitializerException
      *             if dev mode can't be initialized
@@ -100,8 +100,7 @@ public interface DevModeHandlerManager {
     void setApplicationUrl(String applicationUrl);
 
     /**
-     * Registers a command that will be run then the servlet context is shut
-     * down.
+     * Registers a command that will run when DevModeHandler is shut down
      *
      * @param command
      *            the command to run
--- a/flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.internal;
 
+import java.io.Closeable;
 import java.util.Optional;
 import java.util.Set;
 
@@ -96,6 +97,23 @@ public interface DevModeHandlerManager {
      *            the application url
      */
     void setApplicationUrl(String applicationUrl);
+
+    /**
+     * Adds watcher to watcher list.
+     *
+     * @param watcher
+     *            Closable interface that will be closed on shut down
+     */
+    void addWatcher(Closeable watcher);
+
+    /**
+     * Removes watcher from the list.
+     *
+     * @param watcher
+     *            watcher
+     * @return true if removed, false otherwise.
+     */
+    boolean removeWatcher(Closeable watcher);
 
     /**
      * Gets the {@link DevModeHandler}.
--- a/flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
@@ -20,6 +20,7 @@ import java.util.Optional;
 import java.util.Set;
 
 import com.vaadin.flow.di.Lookup;
+import com.vaadin.flow.server.Command;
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
```diff
@@ -97,6 +98,23 @@ void initDevModeHandler(Set<Class<?>> classes, VaadinContext context)
      */
     void setApplicationUrl(String applicationUrl);
 
+    /**
+     * Adds watcher to watcher list.
```
Artur-, line 102: This javadoc should answer

"Which watcher list?" 
"What happens if I add a watcher to the watcher list?"
"Why would I want to add a watcher to the watcher list?"

Is it even relevant that these are watchers or just resources that will be closed on shutdown? In other words, should it rather be
```
/**
* Registers a command that will be run then the servlet context is shut down. 
*
* @param command the command to run
*/
registerShutdownCommand(Command command)
```

flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
```diff
@@ -97,6 +98,23 @@ void initDevModeHandler(Set<Class<?>> classes, VaadinContext context)
...
+     *
+     * @param watcher
+     *            Closable interface that will be closed on shut down
+     */
+    void addWatcher(Closeable watcher);
+
+    /**
```
Artur-, line 109: If the above is `reigsterShutdownCommand` then this method becomes unnecessary as you would do any checks inside the command instead of removing the command

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerManagerImpl.java
```diff
@@ -196,6 +196,16 @@ private void setDevModeStarted(VaadinContext context) {
                 new DevModeHandlerAlreadyStartedAttribute());
     }
 
+    @Override
+    public void addWatcher(Closeable closeable) {
+        this.watchers.add(closeable);
```
Artur-, line 201: Probably we should update the internal calls to also use this method instead of using `this.watchers` directly

Artur- (changes requested):

flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
abdullahtellioglu: `Command` is not closable,  should it be `Closable`?

Artur-: If it is `Command` then it is not something dev mode handler manager closes at shutdown but it is commands it runs at shutdown. The command can close something if it wants to

abdullahtellioglu: Updated 👍. I created an internal method to register watchers, and external shutdown commands using `Command` interface as you mentioned.

flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
abdullahtellioglu: Updated 👍

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DevModeHandlerManagerImpl.java
abdullahtellioglu: `registerWatcherShutdownCommand` is created for registering internal Closeables.

flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
```diff
@@ -97,6 +99,15 @@ void initDevModeHandler(Set<Class<?>> classes, VaadinContext context)
      */
     void setApplicationUrl(String applicationUrl);
 
+    /**
+     * Registers a command that will be run then the servlet context is shut
+     * down.
```
mcollovati, line 104: nit: I would remove the `servlet context` part. It could just say the commands run when the `DevModeHandler` is stopped.

mcollovati: BTW, also `initDevModeHandler` method Javadoc mentions servlet context. Probably a leftover, since it takes a generic `VaadinContext`.

mcollovati (approved):
Only a comment about Javadoc, but not a blocker.

flow-server/src/main/java/com/vaadin/flow/internal/DevModeHandlerManager.java
abdullahtellioglu: Updated javadoc 👍

PR #20075, opened 2024-09-27
fix: Store layout packages

### LINKED ISSUE #20066: @Layout is lost on view change
### Description of the bug

With
```java
@Layout
public class MainLayout extends AppLayout {
```

and
```java
@PageTitle("Hello World")
@Route("")
@Menu(order = 0, icon = "line-awesome/svg/globe-solid.svg")
public class HelloWorldView extends HorizontalLayout {
```

the hello world view is rendered inside the main layout.

However, if you make any change to the view class then after Spring Boot dev tools has restarted, the view is no longer rendered inside the layout

### Expected behavior

It is inside the layout

npm init vaadin@latest test-layout
mvn
Change to use @Layout + @Route("")
make any change to the hello world view

### Versions

Hilla: 24.5.0.beta3
Flow: 24.5.0.beta2
Vaadin: 24.5.0.beta3
Copilot: 24.5.0.beta2
Frontend Hotswap: Disabled, using pre-built bundle
OS: aarch64 Mac OS X 15.0
Java: JetBrains s.r.o. 21.0.3
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36
Java Hotswap: false
IDE Plugin: false

caalador:
Store layout packages
in development mode reload
cache as layouts may not be
in same packages as routes.

Fixes #20066

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/ReloadCache.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/ReloadCache.java
@@ -29,6 +29,7 @@ class ReloadCache implements Serializable {
     static Set<String> skippedResources = new HashSet<>();
     static Set<String> dynamicWhiteList;
     static Set<String> routePackages;
+    static Set<String> layoutPackages;
     static Set<String> jarClassNames = new HashSet<>();
     static Set<Class<?>> jarClasses = new HashSet<>();
 }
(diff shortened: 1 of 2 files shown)


mshabarov (approved):
Works as expected with this patch.

To note: to reproduce this issue one should place route class to a separate folder/package, if the route and layout are in the same package, this problem does not show up.

Two questions to think about: 
- how can this be covered by tests
- how we can prevent similar issues in the future, when we are adding new routing annotations?

mshabarov:
Let discuss these question on Monday and meanwhile let me make another beta release.

PR #20076, opened 2024-09-27
fix: Layout should be collected as entyrpoint

### LINKED ISSUE #20074: @Layout components are not included in the production bundle
### Description of the bug

When you use `@Layout` then all components only used in the layout are missing from the production bundle

The code here https://github.com/vaadin/flow/blob/4844adb66f1196d797a9a8e9bd0c62f5ffbef55f/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java#L440 use `@Route` and other classes as entry points but does not seem to consider `@Layout`

### Expected behavior

The layout components are included

Any app using `@Layout` + AppLayout + SideNav look like this in production
image

### Versions

24.5 beta 3

caalador:
Layout annotated classes should be
collected as entrypoints for
production builds.

Fixes #20074

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendDependencies.java
@@ -51,6 +51,7 @@ import com.vaadin.flow.component.page.AppShellConfigurator;
 import com.vaadin.flow.internal.ReflectTools;
 import com.vaadin.flow.router.DefaultRoutePathProvider;
 import com.vaadin.flow.router.HasErrorParameter;
+import com.vaadin.flow.router.Layout;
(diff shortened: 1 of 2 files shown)


mshabarov (approved):
Works as expected after this patch and also I don't see messages like:
```
2024-09-27T15:11:15.928+03:00 ERROR 10241 --- [nio-8080-exec-8] c.v.flow.component.internal.UIInternals  : The component class com.vaadin.flow.component.applayout.DrawerToggle includes '@vaadin/app-layout/src/vaadin-drawer-toggle.js' but this file was not included when creating the production bundle. The component will not work properly. Check that you have a reference to the component and that you are not using it only through reflection. If needed add a @Uses(DrawerToggle.class) where it is used.
```
anymore.
LGTM.

PR #20099, opened 2024-10-01
fix: Layouts are accessible if not specifically denied

### LINKED ISSUE #20097: Access control does not work properly when using @Layout with Hilla views
### Description of the bug

If you have a Hilla view that is rendered inside a `@Layout` Java layout, then access to the Hilla view is always denied

### Expected behavior

Access control is based on the Hilla view settings

start.vaadin.com -> add a hilla view with AC "logged in" -> download -> select flow main layout

### Versions

Hilla: 24.5.0.beta4
Flow: 24.5.0.beta3
Vaadin: 24.5.0.beta4
Copilot: 24.5.0.beta3
Frontend Hotswap: Enabled, using Vite
OS: aarch64 Mac OS X 15.0
Java: JetBrains s.r.o. 21.0.3
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36
Java Hotswap: false
IDE Plugin: ☑ 1.0.0

caalador:
Allow request for Layout class
if not specifically denied.

Fixes #20097

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/AccessAnnotationChecker.java
@@ -28,6 +28,7 @@ import java.security.Principal;
 import java.util.Objects;
 import java.util.function.Function;
 
+import com.vaadin.flow.router.Layout;
 import com.vaadin.flow.server.VaadinServletRequest;
 
 /**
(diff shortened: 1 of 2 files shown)


mcollovati (approved):
The fix looks good to me.

I wonder if we should update `@Layout` javadoc to make it clear that annotated components act like `@AnononymousAllowed`.

In addition, is it reasonable to allow access to '@Layout` only if there is no any security annotation?
I mean, would it make sense to have a configuration like the following?
```java
@Layout
@RolesAllowed("ADMIN")
```

PR #20108, opened 2024-10-02
fix: layout access issue

caalador:
Add custom reason for when the 
navigate target is a Layout, so the
component needing the access 
annotations is clear.


caalador:
Closing as it is updated in #20120

PR #20119, opened 2024-10-02
chore: put auto layout behind feature flag

caalador:


mshabarov:
Auto layout feature is shipped normally with 24.5, so this isn't needed.

Issue #20063, opened 2024-09-26
Generate MenuItems api for application menu generation


caalador:
### Describe your motivation

There exists the MenuRegistry that can be used for collecting menu item and route data, but the implementation is aimed for internal consumption and is not as user friendly as it could be.

### Describe the solution you'd like

There should be a `MenuConfiguration` class that returns `MenuOptions` objects as an ordered collection for use within applications.

### Describe alternatives you've considered

Update and remake the current MenuRegistry api.

mshabarov:
**Acceptance Criteria (draft)**
- [ ] `MenuRegistry` is moved under `com.vaadin.flow.internal.menu` and/or documented as "for internal use only".
- [ ] If moved to internal package, `Hilla` and `Start` are modified accordingly.
- [ ] `MenuConfiguration` new class is now public class that has needed methods for dynamic menus, similar to `RouteConfiguration` class that is public API.
- [ ] `MenuConfiguration` has methods `getAvailableMenuOptions` (or similar name) + `getAvailableClientMenuOptions` that returns new `MenuOptions` class (or as an alternative, the existing `MenuData` class? )
- [ ]  `MenuOptions` can give a collection or stream of `MenuOption`
- [ ] `MenuOption` class has path, title, order, exclude and icon fields (what else is needed?)

mshabarov:
Should include fixes for:
- [ ] https://github.com/vaadin/flow/issues/20114
- [ ] https://github.com/vaadin/flow/issues/20109

knoobie:
> MenuOption class has path, title, order, exclude and icon fields (what else is needed?)

Should this include a reference to the class from which the menu comes from? Custom Menu Annotations are probably a bit hard to support.. but this would at least allow to gather all other annotation / informations from the Menu's class.

mshabarov:
Sounds like a useful data to have there, agreed.

PR #20138, opened 2024-10-03
feat: add MenuConfiguration

tltv:
New public API for building application menu: adds `MenuConfiguration`, `MenuOptions` and `MenuOption` where `MenuConfiguration` is the main entry point to access menu data to build main menu.

Fixes: #20063


tltv:
@Artur- do you have any feedback regarding this new public Menu API called `MenuConfiguration`. You updated Start app with `MenuRegisty`, so this new API would be a replacement for `MenuRegistry`. E.g. instead of calling `MenuRegistry.collectMenuItemsList()` you would call `MenuConfiguration.getMenuOptions().get()` which returns `List<MenuOption>`.

Artur-:
Is there a reason I should call `MenuConfiguration.getMenuOptions().get()` instead of `MenuConfiguration.getMenuOptions()` to get the menu options?

tltv:
> Is there a reason I should call `MenuConfiguration.getMenuOptions().get()` instead of `MenuConfiguration.getMenuOptions()` to get the menu options?

If you don't see reason, we might not need to do it this way. What do you think if `MenuConfiguration.getMenuOptions()` would just return `List<MenuOption>` directly? And especially what do you think of `MenuOptions` class in general, does it look like useful for developers or could we just go with a `List` instead?

I'm not so sure how useful `MenuOptions` would be. Also pinging @caalador for more input.

caalador:
From the writing of the ticket the target was to get `List<MenuOptions>` which would be in the implemented case the same as `List<MenuOption>` (Options was noting that it contains multiple parameters instead of only being an Option)

tltv:
> From the writing of the ticket the target was to get `List<MenuOptions>` which would be in the implemented case the same as `List<MenuOption>` (Options was noting that it contains multiple parameters instead of only being an Option)

And MenuOption name was picked because MenuItem was already reserved by components. Let's wait if @mshabarov had some idea with splitting in MenuOptions and MenuOption in ticket requirements. Maybe we can remove MenuOptions. Naming properly is still a bit challenging here.

mshabarov:
I was thinking that `MenuOptions` class could have `List<MenuOption>` and some other fields that are common to menu. However, I don't know what would this be and, if we need to expose it, we can add a new method in `MenuConfiguration`.
Additionally, the code that renders menu should be as simple as
```
MenuConfiguration.getMenuOptions().forEach(option -> new SideNavItem(option.getTitle(), option.getRoute()));
```
Thus, I agree that we could simplify it to have just a collection of `MenuOption`.

mshabarov:
Possible alternatives are `MenuEntry`, `MenuAction`, `MenuChoice`, `MenuNode`,`MenuCommand`, `MenuLinkData`.

How about `MenuEntry` ?

tltv:
> How about `MenuEntry` ?

`MenuEntry` is also vacant. +1 from me.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public class MenuConfiguration implements Serializable {
+
+    /**
+     * Get the {@link MenuOptions} instance containing all menu options for the
+     * application.
+     *
+     * @return the {@link MenuOptions} instance
+     */
+    public static MenuOptions getMenuOptions() {
```
mshabarov, line 33: this probably should expose the same method, but with `Locale`

mshabarov, line 33: I don't see `MenuRegistry` javadocs update, these are needed to highlight it's internal use.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuOption.java
```diff
@@ -0,0 +1,46 @@
...
+ *            whether to exclude the menu option
+ * @param icon
+ *            Icon to use in the menu or null for no icon. Value can go inside a
+ *            {@code <vaadin-icon>} element's {@code icon} attribute which
+ *            accepts icon group and name like 'vaadin:file'. Or it can go to a
+ *            {@code <vaadin-icon>} element's {@code src} attribute which takes
+ *            path to the icon. E.g. 'line-awesome/svg/lock-open-solid.svg'.
+ * @param menuClass
+ *            the source class with {@link com.vaadin.flow.router.Menu}
+ *            annotation or null if not available
```
mshabarov, line 42: Worth to mention that for menu entries pointing to Hilla/TypeScript views, this will be `null`

flow-server/src/main/java/com/vaadin/flow/router/MenuData.java
```diff
@@ -19,12 +19,14 @@
...
 
+import com.vaadin.flow.component.Component;
+
 /**
  * Data class for menu item information.
  * <p>
  * Only for read as data is immutable.
  */
-public record MenuData(String title, Double order, boolean exclude, String icon) implements Serializable {
+public record MenuData(String title, Double order, boolean exclude, String icon, Class<? extends Component> menuClass) implements Serializable {
```
mshabarov, line 29: This should be deprecated and marked for internal use. We can't delete it as it's already shipped with 24.4, but we either shouldn't encourage it to be used in public API.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+
+package com.vaadin.flow.server.menu;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Locale;
+
+/**
+ * Menu configuration helper class to retrieve available menu entries for
+ * application main menu.
```
mshabarov, line 25: Add "since 24.5" note

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/router/MenuData.java
tltv: Good point. Someone might use it already with 24.4 so need to keep and deprecate old. But it's also part of the public RouteData API which is not internal API and it's in 24.4. Marking it internal now is not possible afaik.

mshabarov: Hmm, okay, let's keep it as is then.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.vaadin.flow.internal.menu.MenuRegistry;
+
+/**
+ * Menu configuration helper class to retrieve available menu entries for
+ * application main menu.
+ *
+ * @since 24.5
+ */
+public class MenuConfiguration {
```
mshabarov, line 31: nit: this can be final class

mshabarov (approved):

mshabarov (approved):

tltv:
Drafted PR that updates imports in Hilla: https://github.com/vaadin/hilla/pull/2794
After this one is merged, that one should be merged too.

PR #20139, opened 2024-10-03
fix: reload route configuration upon layout changes

### LINKED ISSUE #20111: Hotswap does not take into account adding or removing @Layout
### Description of the bug

Adding or removing the `@Layout` annotation on a class implementing `RouterLayout` has no effect when using HotSwapAgent. You need to restart the server to see the change.

### Expected behavior

Expected that a server restart is not needed

24.5 pre project from Start

### Versions

- Vaadin / Flow version: 24.5.0.beta4

mcollovati:
## Description

When a [At]Layout annotated class is modified, the changes are not propagated to the route registry after hotswap happens. This change updates Route registry layouts configuration and re-registers routes potentially impacted by the change to apply the new settings. It also checks the route target chain for active UIs in order to trigger a page refresh if the layout changes should be applied.

Fixes #20111

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
@@ -35,6 +35,7 @@ import com.vaadin.flow.internal.AnnotationReader;
 import com.vaadin.flow.internal.ReflectTools;
 import com.vaadin.flow.router.BeforeEnterListener;
 import com.vaadin.flow.router.HasErrorParameter;
+import com.vaadin.flow.router.Layout;
 import com.vaadin.flow.router.Menu;
 import com.vaadin.flow.router.MenuData;
 import com.vaadin.flow.router.NotFoundException;
@@ -55,7 +56,6 @@ import com.vaadin.flow.server.auth.MenuAccessControl;
 import com.vaadin.flow.server.auth.NavigationAccessControl;
 import com.vaadin.flow.server.auth.NavigationContext;
 import com.vaadin.flow.server.auth.ViewAccessChecker;
-import com.vaadin.flow.router.Layout;
 import com.vaadin.flow.internal.menu.MenuRegistry;
 import com.vaadin.flow.shared.Registration;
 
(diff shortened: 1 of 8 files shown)


flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractRouteRegistry.java
```diff
@@ -610,6 +610,19 @@ public void setLayout(Class<? extends RouterLayout> layout) {
...
+    void updateLayout(Class<? extends RouterLayout> layout) {
+        if (layout == null) {
+            return;
+        }
+        synchronized (layouts) {
+            layouts.entrySet()
+                    .removeIf(entry -> layout.equals(entry.getValue()));
+            if (layout.isAnnotationPresent(Layout.class)) {
+                layouts.put(layout.getAnnotation(Layout.class).value(), layout);
+            }
```
tepi, line 622: If the `layout` is already in `layouts` map and still has the annotation present, could/should that be a no-op? And only do the removing if that condition is not met?

mcollovati: The annotation value might have been changed. We can do an additional check, but I'm not sure it is worth it

tepi: Ok, in that case it does not make sense

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
```diff
@@ -125,7 +126,7 @@ public static List<Class<? extends RouterLayout>> getParentLayouts(
 
         if (hasRouteAndPathMatches && !route.get().layout().equals(UI.class)) {
             list.addAll(collectRouteParentLayouts(route.get().layout()));
-        } else if (route.get().autoLayout() && hasRouteAndPathMatches
+        } else if (hasRouteAndPathMatches && route.get().autoLayout()
```
caalador, line 129: #20120 will make this change not needed as the layouts are not stored on load, but used dynamically as is for client layouts also.

mcollovati: Great, thanks. I'll revert it.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskGenerateReactFiles.java
```diff
@@ -164,8 +166,10 @@ private void doExecute() throws ExecutionFailedException {
             writeFile(flowTsx, getFileContent(FLOW_TSX));
             writeFile(vaadinReactTsx,
                     getVaadinReactTsContent(routesTsx.exists()));
-            writeFile(new File(frontendGeneratedFolder, LAYOUTS_JSON),
-                    layoutsContent());
+            // writeFile(new File(frontendGeneratedFolder, LAYOUTS_JSON),
+            // layoutsContent());
```
tepi, line 170: nit: commented out code

mcollovati: removed commented code

tepi (approved):

Issue #20118, opened 2024-10-02
Hotswap of translations does not work using Hotswap Agent plugin


sveinnetnordic:
### Description of the bug

Using hotswapagent 2.0.1 with watchResources=src/main/resources/vaadin-i18n. Seems to work, but UI/Vaadin does not load/refresh translations. Need a full restart.

HOTSWAP AGENT: 11:10:14.384 INFO (org.hotswap.agent.plugin.spring.reload.SpringBeanReload) - try to add changed property 'file:/D:/vaadin/src/main/resources/vaadin-i18n/translations_no.properties' into org.springframework.beans.factory.support.DefaultListableBeanFactory@17273273

### Expected behavior

Should show new translation text on hotswap

### Minimal reproducible example

H1 heading1 = new H1(translate("h1.element-1.text"));

h1.element-1.text=Hello

### Versions

Hilla: 24.5.0.beta4
Flow: 24.5.0.beta3
Vaadin: 24.5.0.beta4
Copilot: 24.5.0.beta3
Frontend Hotswap: Enabled, using Vite
OS: amd64 Windows 11 10.0
Java: JetBrains s.r.o. 21.0.3
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36
Java Hotswap: false
IDE Plugin: ☑ 1.9.9

https://github.com/HotswapProjects/HotswapAgent/releases/tag/RELEASE-2.0.1

Artur-:
`ResourceBundle.getBundle` used by `DefaultI18NProvider` at least caches the bundle lookup and returns the old version. This caching is probably unnecessary in development mode

sveinnetnordic:
FYI: Tried to patch 24.5.0.beta5 with TranslationFileRequestHandler. Did not work for Flow. Maybe patching will not work...
`HOTSWAP AGENT: 08:24:17.649 INFO (org.hotswap.agent.plugin.spring.reload.SpringBeanReload) - try to add changed property 'file:/D:/NNDev/studd2/studd-vaadin/target/classes/vaadin-i18n/translations_no.properties' into org.springframework.beans.factory.support.DefaultListableBeanFactory@67efd2c2`

Artur-:
The latest version of #20142 is supposed to fix it for Flow

mcollovati:
[...]

PR #20142, opened 2024-10-04
feat: HMR for translation files

Artur-:
Clears resource bundle cache on translation resource redeployment
Fires a HMR event to the browser when translations are reloaded
    
Limitation: Only supports DefaultI18NHandler where the paths are known. If you have a custom I18N handler, you might need a custom HMR supporting HotswapListener.
    
For #20118
Requires https://github.com/vaadin/hilla/pull/2795 to fully fix the issue

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/hotswap/Hotswapper.java
+++ b/flow-server/src/main/java/com/vaadin/flow/hotswap/Hotswapper.java
@@ -185,11 +185,16 @@ public class Hotswapper implements ServiceDestroyListener, SessionInitListener,
             liveReload.sendHmrEvent("translations-update", Json.createObject());
 
             // Trigger any potential Flow translation updates
-            List<UI> uis = new ArrayList<>();
             EnumMap<UIRefreshStrategy, List<UI>> refreshActions = new EnumMap<>(
                     UIRefreshStrategy.class);
-            forEachActiveUI(ui -> uis.add(ui));
-            refreshActions.put(UIRefreshStrategy.REFRESH, uis);
+            forEachActiveUI(ui -> {
+                UIRefreshStrategy strategy = ui.getPushConfiguration()
+                        .getPushMode().isEnabled()
+                                ? UIRefreshStrategy.PUSH_REFRESH_CHAIN
+                                : UIRefreshStrategy.REFRESH;
+                refreshActions.computeIfAbsent(strategy, k -> new ArrayList<>())
+                        .add(ui);
+            });
             triggerClientUpdate(refreshActions, false);
         }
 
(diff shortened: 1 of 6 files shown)


flow-server/src/main/java/com/vaadin/flow/hotswap/Hotswapper.java
```diff
@@ -165,13 +166,15 @@ public void onHotswap(URI[] createdResources, URI[] modifiedResources,
...
-        // entry point for future implementations, like reloading I18n provider
         if (LOGGER.isTraceEnabled()) {
             LOGGER.trace(
                     "Created resources: {}, modified resources: {}, deletedResources: {}.",
                     createdResources, modifiedResources, deletedResources);
         }
+
+        // Clear resource bundle cache so that translations (and other
+        // resources) are reloaded
+        ResourceBundle.clearCache();
```
mcollovati, line 177: This method is called for all kinds of resources. Should we set a condition to limit the cache eviction only to Vaadin resources?

flow-server/src/main/java/com/vaadin/flow/hotswap/Hotswapper.java
```diff
@@ -180,6 +179,11 @@ public void onHotswap(URI[] createdResources, URI[] modifiedResources,
             // Clear resource bundle cache so that translations (and other
             // resources) are reloaded
             ResourceBundle.clearCache();
+            // Trigger UI refresh
+            EnumMap<UIRefreshStrategy, List<UI>> refreshStrategy = new EnumMap<>(
+                    UIRefreshStrategy.class);
+            refreshStrategy.put(UIRefreshStrategy.REFRESH, List.of());
```
Artur-, line 185: Shouldn't there be some UIs here?

mcollovati: Since the strategy is REFRESH, UIs are not required, since the refresh is broadcasted by live reload.
But if you prefer to try with PUSH for every UI the strategy should be `PUSH_REFRESH_CHAIN`

mcollovati: Actually, `UIRefreshStrategy.REFRESH` with or without UIs behave the same.

mcollovati (approved):
Translation file changes are correctly handled by the Flow Hotswapper.
For Flow views, the page is refreshed (no page reload) using PUSH (if enabled) or browser live reload websocket channel.
I couldn't test with Hilla, but the browser live reload correctly forwards the hmr event to the client.

Artur-:
Hilla needs a slightly different version of https://github.com/vaadin/hilla/pull/2795 and can be tested as part of that one

PR #20164, opened 2024-10-07
chore: Use local config file for formatting

Artur-:


mcollovati (dismissed):
Looks good to me.
nit: we could rename `.mvn/hello.txt` to `.mvn/.gitkeep`, that seems a commonly used filename to "preserve" directories in git repos

Issue #20158, opened 2024-10-07
Navigating Hilla view throws getContent is null exception on flow layout hybrid application


abdullahtellioglu:
### Description of the bug

I downloaded a project from start.vaadin.com and created one Hilla one Flow view and selected Flow as the layout. Navigating to hilla view throws the exception below in a hybrid flow layout application.
```java
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because the return value of "com.example.application.views.MainLayout.getContent()" is null
```

After reloading the page everything works fine. 

https://github.com/user-attachments/assets/8f523df6-36c1-46f8-a94a-4eca6ade7073

As a workaround `if(getContent() == null)` works.

The exception stack trace 
```java
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because the return value of "com.example.application.views.MainLayout.getContent()" is null
	at com.example.application.views.MainLayout.getCurrentPageTitle(MainLayout.java:85) ~[classes/:na]
	at com.example.application.views.MainLayout.afterNavigation(MainLayout.java:81) ~[classes/:na]
	at com.vaadin.flow.component.applayout.AppLayout.showRouterLayoutContent(AppLayout.java:309) ~[vaadin-app-layout-flow-24.5.0.beta5.jar:na]
	at com.vaadin.flow.component.internal.UIInternals.showRouteTarget(UIInternals.java:856) ~[flow-server-24.5.0.beta4.jar:24.5.0.beta4]
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.handle(AbstractNavigationStateRenderer.java:256) ~[flow-server-24.5.0.beta4.jar:24.5.0.beta4]
```

### Expected behavior

It should not throw the exception. 

### Minimal reproducible example

Downloaded application from Start has 3 classes:

```java

package com.example.application.views;

import com.vaadin.flow.component.applayout.AppLayout;
import com.vaadin.flow.component.applayout.DrawerToggle;
import com.vaadin.flow.component.html.Footer;
import com.vaadin.flow.component.html.H1;
import com.vaadin.flow.component.html.Header;
import com.vaadin.flow.component.html.Span;
import com.vaadin.flow.component.icon.SvgIcon;
import com.vaadin.flow.component.orderedlayout.Scroller;
import com.vaadin.flow.component.sidenav.SideNav;
import com.vaadin.flow.component.sidenav.SideNavItem;
import com.vaadin.flow.router.Layout;
[...]

PR #20165, opened 2024-10-07
feat: Add a helper method for menu's page title

mshabarov:
## Description

Adds a new static helper method that gives a page title for views shown using menu.

Fixes https://github.com/vaadin/flow/issues/20158

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
@@ -19,7 +19,9 @@ package com.vaadin.flow.server.menu;
 import java.io.Serializable;
 import java.util.List;
 import java.util.Locale;
+import java.util.Optional;
 
+import com.vaadin.flow.component.UI;
 import com.vaadin.flow.internal.menu.MenuRegistry;
 
 /**
@@ -53,6 +55,34 @@ public final class MenuConfiguration {
     public static List<MenuEntry> getMenuEntries(Locale locale) {
         return MenuRegistry.collectMenuItemsList(locale).stream()
                 .map(MenuConfiguration::createMenuEntry).toList();
+    }
+
+    /**
+     * Get the optional page title of the currently shown view that is rendered
+     * in the main layout and menu is used for navigation. Returns an empty page
+     * title, when navigating in other way rather than menu.
+     *
+     * @return page title, if navigation with menu, empty optional if navigating
+     *         in other way.
+     */
+    public static Optional<String> getPageTitle() {
+        UI ui = UI.getCurrent();
+        if (ui != null) {
+            String path = ui.getInternals().getActiveViewLocation().getPath();
+            if (path.startsWith("/")) {
+                path = path.substring(1);
+            }
+            String finalPath = path;
+            return MenuRegistry.collectMenuItemsList().stream()
+                    .filter(menuItem -> {
+                        String route = menuItem.route();
+                        if (route.startsWith("/")) {
+                            route = route.substring(1);
+                        }
+                        return route.equals(finalPath);
+                    }).map(AvailableViewInfo::title).findFirst();
+        }
+        return Optional.empty();
     }
 
     private static MenuEntry createMenuEntry(AvailableViewInfo viewInfo) {


flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
```diff
@@ -55,6 +57,34 @@ public static List<MenuEntry> getMenuEntries(Locale locale) {
...
+     *
+     * @return page title, if navigation with menu, empty optional if navigating
+     *         in other way.
+     */
+    public static Optional<String> getPageTitle() {
+        UI ui = UI.getCurrent();
+        if (ui != null) {
+            String path = ui.getInternals().getActiveViewLocation().getPath();
+            if (path.startsWith("/")) {
+                path = path.substring(1);
```
tltv, line 73: Usage of `PathUtil.trimPath` could cleanup a bit.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
```diff
@@ -55,6 +57,34 @@ public static List<MenuEntry> getMenuEntries(Locale locale) {
                 .map(MenuConfiguration::createMenuEntry).toList();
     }
 
+    /**
+     * Get the optional page title of the currently shown view that is rendered
```
tltv, line 61: This is not very clear when it can and can't be used. 

At least it would work when called inside After Navigation Event handler. But not if navigation happens to path with route parameter (any kind: required, optional, wildcard). Works only with paths targeting a route without any custom route parameters, like with automatic menu links.

mshabarov: This does work for optional and wildcard route parameters, because these are trimmed in dynamic menu implementation and the links basically point to routes without route parameters.
However, it doesn't work for mandatory route parameters and query parameters, but these are not usually used in the menu.
So, this method is very limited to be used only for dynamic menus in auto layouts, that's true.
I'll update the javadoc.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
```diff
@@ -58,29 +64,87 @@ public static List<MenuEntry> getMenuEntries(Locale locale) {
...
+     * Flow main layouts to render a page header.
+     * <p>
+     * Attempts to retrieve header from the following sources:
+     * <ul>
+     * <li>from {@code ViewConfig.title} of the client-side views;</li>
+     * <li>from {@link PageTitle} value of the server-side route</li>
+     * </ul>
+     * <p>
+     * For server-side routes it falls back to route's Java class name, if a
+     * non-null {@code content} is given. For client-side views it falls back to
```
tltv, line 108: Remove mention about the `content` as it's not in this method.

mshabarov: thanks, removed.

tltv (approved):
LGTM

Only thing, which was not mentioned in any ticket yet, so consider this as a improvement, is to also check `HasDynamicTitle` before `PageTitle`.

It's done like this for document title, and could work here too: 
```java
ui.getInternals()
                .getActiveRouterTargetsChain().stream()
                .filter(HasDynamicTitle.class::isInstance)
                .map(tc -> ((HasDynamicTitle) tc).getPageTitle())
                .filter(Objects::nonNull).findFirst()
```

mshabarov:
Excellent finding! I've added it as a new method in utils and reused here.

flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
```diff
@@ -88,6 +88,19 @@ public static List<MenuEntry> getMenuEntries(Locale locale) {
      */
     public static Optional<String> getPageHeader(Component content) {
         if (isServerSideContent(content)) {
+            UI ui = UI.getCurrent();
+            if (ui != null) {
```
tltv, line 93: `HasDynamicTitle` would be now checked only for server route (or when content is null). Client routes with flow layout could also use `HasDynamicTitle` in the MainLayout and that would be missed now. Same code could be added in getPageHeaderFromMenuItems method too.

mshabarov: I didn't consider this case, thanks. Let me add it.

mshabarov: done

tltv (approved):

PR #20169, opened 2024-10-07
chore: Change line length to a more modern 120

Artur-:


knoobie:
Welcome in 2024! 🥳

caalador:
Will this lead to updating the line length for other alive versions also so that cherry picking works (has a better change to work) automatically?

Artur-:
It might work by happenstance. Otherwise #20164 and this + #20160 + #20185 should be backported to all relevant versions

mshabarov:
Moved this to a draft as the discussion is ongoing about should we switch to longer lines or not.

Artur-:
> Moved this to a draft as the discussion is ongoing about should we switch to longer lines or not.

Where is this discussion ongoing and what on earth are the arguments against it?

mshabarov:
We discussed this change internally and the picture seems to be so that many people are neutral, few people voted up for 120, but the bigger group of people shared a strong wish to keep the old 80 chars formatting with a thorough argumentation. IMHO it’s fair to keep the existing formatting based on these results.

Issue #20107, opened 2024-10-02
Title in the @Menu annotation has no impact on collectMenuItemsList


Legioth:
### Description of the bug

Seems like `item.title()` from `MenuRegistry.collectMenuItemsList()` for a Flow  view always uses the view's `@PageTitle` value even if the menu title is overridden in the `@Menu` annotation.

### Expected behavior

Expected that the menu uses the title from `@Menu` if one defined and otherwise falls back to the `@PageTitle` value.

### Minimal reproducible example

Any 24.5 prerelase starter from start.vaadin.com

### Versions

- Vaadin / Flow version: 24.5.0.beta4

Legioth:
The same is also the case for a Hilla view with `export const config: ViewConfig = { menu: {title: "Something"} }` which is also ignored.

caalador:
For hilla the structure gives a clearer indication that it is the Menu Title not Page Title. e.g. title for the menu item.

Should the menu title be used as page title and how do all the page title choices override each other?

Legioth:
Seems like I looked at `AvailableViewInfo.title` which is different from `MenuData.title`. Seems like `MenuData.title` is populated as expected for Flow views but not for Hilla views where sometimes the whole object is missing (https://github.com/vaadin/flow/issues/20109) and sometimes it has `null` in `MenuData.title` rather than falling back to the route title like it seems to do for Flow views.

Legioth:
So with Flow views, I would expect that `viewInfo.title` uses `@PageTitle` if defined and otherwise falls back to a title derived from the class name. `viewInfo.menu.title` should use the value from `@Title` if defined and otherwise fall back to `viewInfo.title`.

With Hilla views, I would expect that `viewInfo.title` uses the `title` property from the `ViewConfig` object if defined and otherwise falls back to a title derived from the function name. `viewInfo.menu.title` uses the title from `ViewConfig.menu` if defined and otherwise falls back to `viewInfo.title`.

I assume this would be consistent with how both Flow and Hilla views behave in a Hilla layout.

mshabarov:
@Legioth your last comment probably has a typo, should it be 
```
`viewInfo.menu.title` should use the value from @Menu if defined
```
instead of 
```
`viewInfo.menu.title` should use the value from `@Title` if defined
```
?

mshabarov:
[...]

PR #20184, opened 2024-10-08
fix: update document title for client route

tltv:
Updates document title for client view with flow layout based on `ViewConfig.title` and when not overridden by server's `HasDynamicTitle`.

RelatedTo: #20107

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
@@ -71,6 +71,7 @@ import com.vaadin.flow.server.HttpStatusCode;
 import com.vaadin.flow.server.RouteRegistry;
 import com.vaadin.flow.server.VaadinSession;
 import com.vaadin.flow.internal.menu.MenuRegistry;
+import com.vaadin.flow.server.menu.AvailableViewInfo;
 
 /**
  * Base class for navigation handlers that target a navigation state.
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
```diff
@@ -964,7 +965,11 @@ private static void updatePageTitle(NavigationEvent navigationEvent,
                 .filter(HasDynamicTitle.class::isInstance)
                 .map(tc -> ((HasDynamicTitle) tc).getPageTitle())
                 .filter(Objects::nonNull).findFirst()
-                .orElseGet(lookForTitleInTarget);
+                .orElseGet(() -> Optional
+                        .ofNullable(
+                                MenuRegistry.getClientRoutes(true).get(route))
```
mshabarov, line 970: Are you sure that the `route` has the same format as what `MenuRegistry` returns regarding slash in the beginning and end, e.g. `/foo` vs `foo` ? 
Maybe it's safer to trim slashes ?

tltv: I trust that it's safe here as it's coming from `handle` method where it's used in the same way.

caalador: Note that `handle` does 
```
        String route = event.getLocation().getPath().isEmpty()
                ? event.getLocation().getPath()
                : event.getLocation().getPath().startsWith("/")
                        ? event.getLocation().getPath()
                        : "/" + event.getLocation().getPath();
```
before checking the route.

mshabarov (approved):
Works as expected, LGTM

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
caalador: Also that is exactly the route string that is used.

Issue #20209, opened 2024-10-10
Add usage stats for dynamic menu entries feature in Flow


mshabarov:
### Describe your motivation

Add usage statistics for `MenuConfiguration::getMenuEntries` feature, aka dynamic menu in Flow.

### Additional context

Should target Vaadin 24.5+.

PR #20212, opened 2024-10-10
fix: add usage stats for dynamic menu entries

tltv:
Fixes: #20209

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/menu/MenuConfiguration.java
@@ -24,6 +24,7 @@ import java.util.Optional;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.component.UI;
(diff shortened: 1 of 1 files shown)


mcollovati (commented):
The other PR related to statistics sets entry name in `Constants` class, and it also has a `has-xxxx` format.
Is it a convention to have `flow/xxx` instead of `has-xxxx` for Flow stats?

tltv:
> The other PR related to statistics sets entry name in `Constants` class, and it also has a `has-xxxx` format. Is it a convention to have `flow/xxx` instead of `has-xxxx` for Flow stats?

No, I created this first without any guidelines with naming. Other PR has already existing naming pattern to follow. However, we have other similar patterns with `flow/` in Constants. Point is to be clear it's Flow feature. I considered first to use `flow/MenuConfiguration`, but as this should mark menu entry usage only, not whole class, then `flow/dynamic-menu-entries` feels better.

mcollovati (approved):
LGTM.
I wonder if we should in the future do a little refactor and group all statistic constants in a dedicated class.

PR #20235, opened 2024-10-11
fix: correct route path for MenuEntry

tltv:
MenuEntry should not have route parameters included in the path for client routes. It should omit route path in same way as with Hilla automatic menu. Menu should not include nested routes for excluded menu item or for route with required parameter.


caalador (commented):
Does this change break Hilla?
I feel the filtering should be in MenuConfiguration and not in MenuRegistry as at the moment registry returns what Hilla is expecting.

tltv:
> Does this change break Hilla?
> I feel the filtering should be in MenuConfiguration and not in MenuRegistry as at the moment registry returns what Hilla is expecting.

It doesn't because Hilla uses `collectClientMenuItems(...)`. This change affects only `collectMenuItems()` and `collectMenuItemsList()`.

caalador (approved):

PR #20245, opened 2024-10-14
fix: Add util method for detecting Hilla auto layout

mshabarov:
## Description

Adds a new util method that detects layout entries in `file-routes.json`.

Related-to https://github.com/vaadin/hilla/issues/2825

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
@@ -35,6 +35,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.DeserializationFeature;
@@ -82,10 +83,8 @@ public class MenuRegistry {
     public static Map<String, AvailableViewInfo> collectMenuItems() {
         Map<String, AvailableViewInfo> menuRoutes = MenuRegistry
                 .getMenuItems(true);
-        menuRoutes.entrySet()
-                .removeIf(entry -> Optional.ofNullable(entry.getValue())
-                        .map(AvailableViewInfo::menu).map(MenuData::isExclude)
-                        .orElse(false));
+        filterMenuItems(menuRoutes);
+
         return menuRoutes;
     }
 
@@ -115,7 +114,8 @@ public class MenuRegistry {
         return collectMenuItems().entrySet().stream().map(entry -> {
             AvailableViewInfo value = entry.getValue();
             return new AvailableViewInfo(value.title(), value.rolesAllowed(),
-                    value.loginRequired(), entry.getKey(), value.lazy(),
+                    value.loginRequired(),
+                    getMenuLink(entry.getValue(), entry.getKey()), value.lazy(),
                     value.register(), value.menu(), value.children(),
                     value.routeParameters(), value.flowLayout());
         }).sorted(getMenuOrderComparator(
@@ -437,6 +437,17 @@ public class MenuRegistry {
         }
     }
 
+    private static boolean hasRequiredParameter(AvailableViewInfo viewInfo) {
+        final Map<String, RouteParamType> routeParameters = viewInfo
+                .routeParameters();
+        if (routeParameters != null && !routeParameters.isEmpty()
+                && routeParameters.values().stream().anyMatch(
+                        paramType -> paramType == RouteParamType.REQUIRED)) {
+            return true;
+        }
+        return false;
+    }
+
     /**
      * Check view against authentication state.
      * <p>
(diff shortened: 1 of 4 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
```diff
@@ -303,39 +306,77 @@ public static Map<String, AvailableViewInfo> collectClientMenuItems(
...
+            if (!hasHillaAutoLayout.get() && isMainLayout(clientViewConfig)) {
+                hasHillaAutoLayout.compareAndSet(false, true);
+            }
+        });
+        return hasHillaAutoLayout.get();
+    }
+
+    private static boolean isMainLayout(AvailableViewInfo viewInfo) {
+        return (viewInfo.route() == null || viewInfo.route().isBlank())
+                && viewInfo.children() != null;
```
caalador, line 348: This one seems wrong as now we check through all items when the actual layout for the check should be the initial value only. 
So:
```
[
  {
    "route": "",
    "params": {},
    "title": "Layout",
    "children": [
      {
        "route": "",
        "params": {},
        "title": "Public"
      },
    ]
  }
]
```
Has a main layout, but with the check the following also has a main layout (even though it is another route.):
```

  [
    {
      "route": "hilla",
      "flowLayout": false,
      "params": {},
      "children": [
        {
          "route": "",
          "flowLayout": false,
          "params": {},
          "title": "Layout"
        }
      ]
    },
    {
      "route": "",
      "flowLayout": false,
      "params": {},
      "title": "Layout",
      "children": [
        {
          "route": "components",
          "menu": {
            "title": "React Components"
          },
          "flowLayout": false,
          "params": {},
          "title": "Components"
        }
      ]
    }
  ]
```

For a main layout the check should also verify that there is only one root item for it to be a main layout in the old sense.

Also as a question should only having a layout e.g.
```
            [
                {
                     "route": "",
                     "flowLayout": false,
                     "params": {},
                     "title": "Layout",
                     "children": [
                     ]
                   }
             ]
```
be seen as having a main layout when it would be a layout for only Flow views? (This is returning true with these changes as long as you ignore the `Assert.assertTrue(children.length() > 0);` check in the added test)

When hilla is generating the `file-routes.json` in `createViewConfigJson` we know if it is a layout or not. Should we perhaps just add there a parameter `"layout": true,` that we could check instead of guessing based on amount of children or missing child array?

The big passing, but false should not be generatable, but demonstrates an issue with the check.

flow-server/src/test/java/com/vaadin/flow/server/menu/MenuRegistryTest.java
```diff
@@ -331,6 +334,38 @@ public void getMenuItemsList_assertOrder() {
...
     }
 
+    @Test
+    public void hasHillaAutoLayout_fileRoutesHasLayout_true()
+            throws IOException {
+        JsonArray fileRoutes = JsonUtil.parse(testClientRouteFile);
+        JsonObject layout = fileRoutes.getObject(0);
+        JsonArray children = layout.getArray("children");
+        Assert.assertNotNull(children);
+        Assert.assertTrue(children.length() > 0);
```
caalador, line 344: This assertion is actually wrong as a layout is a layout even with no children.

flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
mshabarov: > This one seems wrong as now we check through all items when the actual layout for the check should be the initial value only.

Yes, better to interrupt once we find the first layout entry.

> Has a main layout, but with the check the following also has a main layout (even though it is another route.):

I didn't get this. How a route can have children ?

> When hilla is generating the file-routes.json in createViewConfigJson we know if it is a layout or not. Should we perhaps just add there a parameter "layout": true, that we could check instead of guessing based on amount of children or missing child array?

That would be better solution indeed. I'm going to try adding this param in Hilla, then this PR can be updated.

flow-server/src/test/java/com/vaadin/flow/server/menu/MenuRegistryTest.java
mshabarov: That sounds weird, but taking into account the suggestion above about new parameter, this test will be modified.

caalador: So basically the following is one `@layout.tsx` without any sub routes which would then work as the layout for any flow view. the generated routes would be `"", children [ serverdside ]`, but that is not visible in the json file.
```
[
  {
    "route": "",
    "flowLayout": false,
    "params": {},
    "title": "Layout",
    "children": [
    ]
  }
]
```

flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
caalador: It's not a mainLayout if there are other items on the root level.
I forgot to edit it to not say route, but layout on the same level.

mshabarov: Ok, got it. I think Flow should collect server routes even if any layout found, no matter is it top level main layout or not.

caalador: But this is used for the client so it basically opens up server routes to the client by sending the data. Flow always collects server routes, the question is should they be sent to the client or not and the consensus has been that only if there is a root layout or if the dev decides to.

mshabarov: I see, thanks. If we follow this logic, then it simplifies the task here: `hasMainMenu` should check that root entries have only one entry and this entry is a layout + it should have "" mapping.
OMG, that's what this codes mean 🤣 
```java
        return !clientItems.isEmpty() && clientItems.size() == 1
                && clientItems.values().iterator().next().route().equals("");
```

caalador (approved):

mshabarov:
It's `MenuRegistryTest`'s new code that impacts `should_not_initialize_UI_and_add_initialUidl_when_invalid_route` somehow.
I'm looking into it.

mshabarov:
Or it's a file routes cache that is not cleaned.

mshabarov:
Added a cache cleanup for the failed test.
I think in real projects or end-to-end tests this problem shouldn't happen, because file routes are not going to be changed during app running in production.

However, unit tests may suffer.

caalador (approved):

PR #20251, opened 2024-10-15
fix: force hotdeploy with hilla

tepi:


flow-server/src/main/java/com/vaadin/flow/server/AbstractConfiguration.java
```diff
@@ -53,9 +53,12 @@ public interface AbstractConfiguration extends Serializable {
     default boolean frontendHotdeploy() {
         if (isProductionMode()) {
             return false;
+        } else if (FrontendUtils.isHillaUsed(getFrontendFolder())) {
+            return true;
+        } else {
+            return getBooleanProperty(InitParameters.FRONTEND_HOTDEPLOY,
+                    FrontendUtils.isHillaUsed(getFrontendFolder()));
```
mcollovati, line 60: I think the fallback can be false here, since `FrontendUtils.isHillaUsed(getFrontendFolder())` will always evaluate to false.

flow-server/src/main/java/com/vaadin/flow/server/PropertyDeploymentConfiguration.java
```diff
@@ -152,8 +152,12 @@ public boolean isProductionMode() {
     @Override
     public boolean frontendHotdeploy() {
         if (isOwnProperty(FRONTEND_HOTDEPLOY)) {
-            return getBooleanProperty(FRONTEND_HOTDEPLOY,
-                    FrontendUtils.isHillaUsed(getFrontendFolder()));
+            if (FrontendUtils.isHillaUsed(getFrontendFolder())) {
+                return true;
+            } else {
+                return getBooleanProperty(FRONTEND_HOTDEPLOY,
+                        FrontendUtils.isHillaUsed(getFrontendFolder()));
```
mcollovati, line 159: I think the fallback can be false here, since `FrontendUtils.isHillaUsed(getFrontendFolder())` will always evaluate to false.

mcollovati (changes requested):

mcollovati (approved):

PR #20255, opened 2024-10-15
Fix flow issue #6349

### LINKED ISSUE #6349: Calling VaadinService.access or accessSynchronously leaks UI ThreadLocals
@tmattsso commented on [Wed Aug 28 2019](https://github.com/vaadin/framework/issues/11699)

When the method is called, CurrentInstance will store a copy of all current instances, then override two objects; VaadinService and VaadinSession. If there is a UI object in the map, it is not cleared (current request and response might be affected too). This means that clients can call UI.getCurrent inside the access block and get a random UI instance.

Either the UI mapping should explicitly be set to null, or the map be cleared, before running client code. The latter would be better, because it will also clear (temporarily) any other mappings that a user might have done for another session.

archiecobbs:
## Description

Commands enqueued by `VaadinSession.access()` in general have nothing to do with each other. The only thing they have in common is they share the same `VaadinSession` (and, by implication, `VaadinService`).

Therefore, if command №1 invoked `UI.setCurrent()` and command №2 invokes `UI.getCurrent()`, command №2 should read `null`, not the random `UI` from command №1 that it has nothing to do with.

Fixes #6349
 
## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

- [X] New and existing tests are passing locally with my change
  - Except `JarContentsManagerTest.copyFilesFromJar_doNotUpdateFileIfContentIsTheSame` which has nothing to do with this change.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
@@ -2147,11 +2147,12 @@ public abstract class VaadinService implements Serializable {
         // Dump all current instances, not only the ones dumped by setCurrent
         Map<Class<?>, CurrentInstance> oldInstances = CurrentInstance
                 .getInstances();
-        CurrentInstance.setCurrent(session);
         try {
             while ((pendingAccess = session.getPendingAccessQueue()
                     .poll()) != null) {
                 if (!pendingAccess.isCancelled()) {
+                    CurrentInstance.clearAll();
+                    CurrentInstance.setCurrent(session);
                     pendingAccess.run();
 
                     try {
--- a/flow-server/src/test/java/com/vaadin/flow/server/VaadinSessionTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/server/VaadinSessionTest.java
@@ -29,6 +29,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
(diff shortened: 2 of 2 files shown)


tepi:
Just to be sure: Is Flow erroneously setting random UI instances to the CurrentInstance value of UI, or is this happening only when the developer is calling `UI.setCurrent()` themselves within a session access command?

archiecobbs:
> Just to be sure: Is Flow erroneously setting random UI instances to the CurrentInstance value of UI, or is this happening only when the developer is calling `UI.setCurrent()` themselves within a session access command?

HI @tepi ,

The overall problem is that whatever `UI` (or `VaadinRequest` or `VaadinResponse`) happens to be leftover from a previous operation is not removed before the next task in the session's pending access queue is executed.

Note that a "previous operation" could be any of these:
* The handling of a normal request/response cycle
* A previous task in the queue (if the affected task is not the first task in the queue)
* Any other operation which requires a session lock/unlock cycle

As a result, when an enqueued session task invokes `UI.getCurrent()`, `VaadinRequest.getCurrent()`, or `VaadinResponse.getCurrent()`, it can get a "random" value when instead `null` should have been returned.

So it's **not** required that the first session task include an explicit call to `UI.setCurrent()`, it's only required that `UI.setCurrent()` is called as a side-effect of the previous operation (for example, handling a normal incoming request).

Here is an example of a sequence of events demonstrating one way this could happen:

* Vaadin HTTP request comes in and is handled by Thread A
* Thread A locks the session and sets the current `UI`, `VaadinRequest`, and `VaadinResponse`
* Background thread B wakes up and invokes `VaadinSession.access(task)`
* Because the session is already locked by thread A, `task` is added to the session's pending access queue
* Thread B finishes handling the request and unlocks the session
* On session unlock, the pending access queue is processed
* The `task` executes, but:
  * The previous `UI`, `VaadinRequest`, and `VaadinResponse` are still in context
  * Yet those three objects have _nothing to do_ with `task`

mshabarov (approved):
The change looks good to me and safe. Excellent that you added the test, it indeed fails when this patch isn't in place.

Thanks for your contribution 👍

PR #20260, opened 2024-10-16
fix: prevent possible error on embedded component disconnect

web-padawan:
## Description

This fixes the following error which sometimes occurs when switching between docs pages with Java examples:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'disconnected')
    at ContextMenuPresentationWc.disconnectedCallback (:8000/vaadin/VAADIN/generated/flow/web-components/context-menu-presentation-wc.js:131:18)
    at removeChild (chunk-UXDD7MME.js:8456:26)
```

I'm not sure how to test this but it makes sense to check for object being defined before calling its methods anyway.

## Type of change

- Bugfix


mshabarov (approved):

mshabarov:
@web-padawan thanks for this fix, we can probably skip the test. Which versions do need this? I guess 24.4 and 24.5 as docs are running using them?

PR #20239, opened 2024-10-14
chore: remove redundant names from package json resources

platosha:

PR #20274, opened 2024-10-17
chore: remove redundant names from package json resources (#20239)(CP: 23.5)

mshabarov:


platosha (changes requested):
Seems that we have redundant test fixtures here for testing Vite Vaadin bundles plugin, accidentally added at `flow-tests/test-frontend/vite-basics`. Those could be safely removed, I think.

platosha (approved):

PR #20275, opened 2024-10-17
chore: remove redundant names from package json resources (#20239)(CP: 2.11)

mshabarov:


mshabarov:
Semantic versioning failing (don't understand why),  but gives `0.0.1`.

platosha (changes requested):
Seems that we have redundant test fixtures here for testing Vite Vaadin bundles plugin, accidentally added at `flow-tests/test-frontend/vite-basics`. Those could be safely removed, I think.

platosha (approved):

PR #20283, opened 2024-10-18
fix: resume client to server communication after web socket reconnection

### LINKED ISSUE #20213: Connection to backend lost after network interruption when using websockets
### Description of the bug

I used the vaadin gradle starter skeleton to have a minimal reproducing example and changed AppShell.java to add a push annotation

```
/**
 * Use the @PWA annotation make the application installable on phones, tablets
 * and some desktop browsers.
 */
@PWA(name = "Project Base for Vaadin", shortName = "Project Base")
@Theme("my-theme")
@Push(transport = Transport.WEBSOCKET, value = PushMode.AUTOMATIC)
public class AppShell implements AppShellConfigurator {
}
```

After starting using `jettyRun`. I open the app in a browser in a virtual machine (I used VirtualBox and libvirt).

I type in something and verify I get a result. I change the textbox content.

Then I virtually pull the cable.

In libvirt this can be done as follows, start `virsh`. Find out machine name (domain `<dom>`) using `list`, find out interface (<iface>) using `domiflist <dom>`. 'Pull cable' using `domif-setlink <dom> <iface> down`.

It is not sufficient to use the developer tools. That does not interrupt the websocket connection.

Press 'Say hello' again to send something and wait some time (best reproduced when browser print something like
'Websocket closed, reason: Connection was closed abnormally (that is, with no close frame being sent). - wasClean: false')

Then reconnect the network (`domif-setlink <dom> <iface> up`).

And observe that nothing is send any more.

### Expected behavior

The connection is reestablished and interaction is possible again.

I do not know if the last update should still be pushed or if resynchronization is better.

I downloaded https://github.com/vaadin/base-starter-gradle and applied the change above. I copied the code when vaadin version was 2.4.10

### Versions

- Vaadin / Flow version: 24.4.10 (probably also most current)
- Java version: 21
- OS version: Client Windows 10, Server Linux Ubuntu Jammy
- Browser version: reproduced with Tatu Lund on Microsoft Edge Version 129.0.2792.79, Firefox 131
- Application Server: jetty as of starter

mcollovati:
## Description

When a websocket PUSH connection is closed and re-established because of a network failure, the RequestResponseTracker.hasActiveRequest is not reset, prenvint the Flow client to send additional messages to the server.
This change will reset the flag on reconnection. It also will track unsent PUSH message over websocket, to retry the delivery once the connection is re-established, preventing client resynchronization. In addition, it sets a default value of 12 for the Atmospehere maxWebsocketErrorRetries setting, to ensure that the Flow client will attempt to reconnect with web socket transport several times, instead of immediately downgrade to long-polling after first failed connection.

Fixes #20213

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


flow-client/src/main/java/com/vaadin/client/communication/MessageSender.java
```diff
@@ -260,7 +280,14 @@ public void resynchronize() {
...
     public void setClientToServerMessageId(int nextExpectedId, boolean force) {
         if (nextExpectedId == clientToServerMessageId) {
-            // No op as everything matches they way it should
+            // Everything matches they way it should
+            // Remove potential pending PUSH message if it has already been seen
+            // by the server.
+            if (pushPendingMessage != null
+                    && (int) pushPendingMessage.getNumber(
+                            ApplicationConstants.CLIENT_TO_SERVER_ID) < nextExpectedId) {
+                pushPendingMessage = null;
```
Artur-, line 289: What happens if the message reaches the server and then the channel is disconnected? It looks like a duplicate of the last message will be sent - what happens after that?

mcollovati: If the message is the same as the last processed one (both client message id and hash must match) then it is ignored on the server side; the response should then have a next expected message id greater than one of the cached message, and `pushPendingMessage` should be nullified.

mcollovati: However, if client id is too old, then the "Unexpected message id from the client ..." error will be raised.

mshabarov (approved):
The patch looks reasonable to me, but I didn't verify it with application.

mshabarov:
Successfully tested with the Ubuntu in Virtual Box by connecting/disconnectiong the network adapter:


https://github.com/user-attachments/assets/b0e4c3fe-9203-41c4-9ae1-41e45a4971a9

I used
```java
@Push(transport = Transport.WEBSOCKET)
```

mshabarov:
@mcollovati I think we can move to ready for review and merge.

PR #20292, opened 2024-10-21
chore: Track all relevant locations for creating and attaching

Artur-:


mshabarov (dismissed):

mshabarov:
Doesn't look like a breaking change, so we can pick it to 24.5.

Artur-:
Let's see that it covers the needs of Copilot before merging

mshabarov (approved):

Issue #20285, opened 2024-10-21
Make WebPushMessage output customisable


caalador:
### Describe your motivation

Make the content of `WebPushMessage` more customizable to enable giving options outside of only title and body.

```
{
  "//": "Visual Options",
  "body": "<String>",
  "icon": "<URL String>",
  "image": "<URL String>",
  "badge": "<URL String>",
  "dir": "<String of 'auto' | 'ltr' | 'rtl'>",
  "timestamp": "<Long>"

  "//": "Both visual & behavioral options",
  "actions": "<Array of Strings>",
  "data": "<Anything>",

  "//": "Behavioral Options",
  "tag": "<String>",
  "requireInteraction": "<boolean>",
  "renotify": "<Boolean>",
  "vibrate": "<Array of Integers>",
  "sound": "<URL String>",
  "silent": "<Boolean>",
}
```

### Describe the solution you'd like

Add a `customParameters` field to the record that is used for the resulting json.

### Describe alternatives you've considered

Make `WebPushMessage` a class so it can be extended.

PR #20304, opened 2024-10-22
feat: Extend WebPushMessage with the custom settings

mshabarov:
## Description

Adds an API to set custom options as described in https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification#parameters

Fixes https://github.com/vaadin/flow/issues/20285

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


mshabarov:
To note: I'm working on figuring out how can we provide an API for adding action listener, else than asking to extend the service worker (sw.js).

Also this needs some testing on mobile.

flow-webpush/src/main/java/com/vaadin/flow/server/webpush/WebPushMessage.java
```diff
@@ -17,24 +17,47 @@
...
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    /**
+     * Creates a new Web Push notification message with title and various options being fetched from a given arbitrary Java Object.
+     *
+     * @param title notification title
+     * @param options any Serializable Java Object representing custom settings that you want to apply to the notification
+     * @see <a href=https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification#parameters</a>
+     */
+    public WebPushMessage(String title, Serializable options) {
```
tepi, line 40: Should we consider providing a proper java class / record for the options instead of just Serializable?

mshabarov: The reason why I didn't add a class for options because otherwise we should extend or modify it each time when the browser API changed. IMO it's better to decouple the Java API from the browser API that way.

mshabarov:
Tested with CRM tutorial in this branch https://github.com/vaadin/flow-crm-tutorial/tree/web-push-options.
Screenshot 2024-11-05 at 15 47 58
this image doesn't show it, but the actions should also be available as well as notification image, if you expand.

When "Open Dashboard" is clicked, the CRM Dashboard should be opened in a new window or focus an existing browser tab.

flow-webpush/src/main/java/com/vaadin/flow/server/webpush/WebPushMessage.java
```diff
@@ -48,9 +73,21 @@ public String toString() {
...
+    }
+
+    private static ObjectNode getBodyOption(String body) {
+        if (body != null) {
+            ObjectNode objectNode = objectMapper.createObjectNode();
+            objectNode.put("body", body);
+            return objectNode;
+        } else {
+            return objectMapper.createObjectNode();
+        }
```
caalador, line 91:
```suggestion
        ObjectNode objectNode = objectMapper.createObjectNode();
        if (body != null) {
            objectNode.put("body", body);
        }
        return objectNode;
```
Could simplify to not have if else and only one create as the difference is if we put content or not.

caalador (approved):

flow-webpush/src/main/java/com/vaadin/flow/server/webpush/WebPushMessage.java
mshabarov: Updated, thanks

caalador (approved):

PR #20306, opened 2024-10-23
chore: don't log missing setter message for records

mcollovati:
Java records are immutable, so they do not provide setters for their properties. However, Flow Binder fully supports writing records by creating new instances on demand.
This change prevents Binder from logging a useless 'missing setter' log message for records.

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -1147,7 +1147,7 @@ public class Binder<BEAN> implements Serializable {
             ValueProvider<BEAN, ?> getter = definition.getGetter();
             Setter<BEAN, ?> setter = readOnly ? null
                     : definition.getSetter().orElse(null);
-            if (!readOnly && setter == null) {
+            if (!readOnly && setter == null && !binder.isRecord) {
                 getLogger().info(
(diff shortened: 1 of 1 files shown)


mshabarov (approved):

flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
```diff
@@ -1147,7 +1147,7 @@ private Binding<BEAN, TARGET> bind(String propertyName,
             ValueProvider<BEAN, ?> getter = definition.getGetter();
             Setter<BEAN, ?> setter = readOnly ? null
                     : definition.getSetter().orElse(null);
-            if (!readOnly && setter == null) {
+            if (!readOnly && setter == null && !binder.isRecord) {
```
knoobie, line 1150: Should there be a test that records with bean annotation work? :)

mshabarov: Bean Validation you mean ?

knoobie: Yeah :)

mshabarov: That would be great to have indeed. But could be a separate PR as well.

Issue #20290, opened 2024-10-21
vaadin.frontend.hotdeploy=false not possible if  Hilla views is used


sveinnetnordic:
### Description of the bug

Vite will always start if some .tsx in the magic views folder. When not developing React Vite will slow down the Flow development

### Expected behavior

https://vaadin.com/docs/latest/upgrading#frontend-bundles-hot-deploy-modes

" If it detects Hilla views in a project, it uses development server — **unless** you set the configuration parameter, vaadin.frontend.hotdeploy=false"

### Minimal reproducible example

NA

### Versions

Hilla: 24.5.0
Flow: 24.5.0
Vaadin: 24.5.0

mshabarov:
As a potential workaround before we fix it, one can move React views aways from frontend/views folder, then Vaadin won't force the dev server.
But ofc you'll loose the File-based router feature then and have to configure the routes manually.

tepi:
Reopening this since dev bundle is not working properly on 24.5, Hilla views are not properly detected. Needs investigation and fix.

tepi:
https://github.com/vaadin/flow/pull/20324 fixes finding the Hilla views in dev-bundle mode

sveinnetnordic:
Works for me :-)

PR #20324, opened 2024-10-23
fix: find file-routes.json in dev bundle

tepi:
Fixes https://github.com/vaadin/flow/issues/20290

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
@@ -18,14 +18,10 @@ package com.vaadin.flow.internal.menu;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.MalformedURLException;
-import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.text.Collator;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -36,7 +32,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -479,7 +474,7 @@ public class MenuRegistry {
                     configuration.getBuildFolder(), FILE_ROUTES_JSON_NAME);
         } catch (IOException e) {
             LoggerFactory.getLogger(MenuRegistry.class).warn(
-                    "Failed to find {} under frontend/generated",
+                    "Failed to find {} in frontend/generated or dev-bundle folder",
                     FILE_ROUTES_JSON_NAME, e);
             throw new RuntimeException(e);
         }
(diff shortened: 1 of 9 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendUtils.java
```diff
@@ -106,6 +106,11 @@ public class FrontendUtils {
      */
     public static final String GENERATED = "generated/";
 
+    /**
+     * Path of the folder to custom dev-bundle.
+     */
+    public static final String DEV_BUNDLE_FOLDER = "target/dev-bundle/";
```
mcollovati, line 112: I guess this will most likely not work for gradle projects.

mcollovati: `StaticFileServer` seems to use `DevBundleUtils.findBundleFile(...)` to get resources from the application bundle

```java
            assetInDevBundleUrl = DevBundleUtils.findBundleFile(
                    deploymentConfiguration.getBuildFolder(), assetInDevBundle);
```

tepi: Good catch, thanks! I'll see if this can be used here as well.

flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
```diff
@@ -471,8 +474,10 @@ public static URL getViewsJsonAsResource(
             if (fileRoutes.toFile().exists()) {
                 return fileRoutes.toUri().toURL();
             }
-            return null;
-        } catch (MalformedURLException e) {
+            return DevBundleUtils.findBundleFile(
+                    configuration.getProjectFolder(),
+                    configuration.getBuildFolder(), FILE_ROUTES_JSON_NAME);
+        } catch (IOException e) {
```
tltv, line 480: Exception handler's log message mention only frontend/generated folder but now it could be from the bundle file too.

flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
```diff
@@ -471,8 +474,10 @@ public static URL getViewsJsonAsResource(
             if (fileRoutes.toFile().exists()) {
                 return fileRoutes.toUri().toURL();
             }
-            return null;
-        } catch (MalformedURLException e) {
+            return DevBundleUtils.findBundleFile(
```
tltv, line 477: What writes file-routes.json in the dev-bundle. Is Hilla doing that?

tepi: I believe so, see https://github.com/vaadin/hilla/blob/d3e74e99ff4ef5294982195348829e3424735335/packages/ts/file-router/src/vite-plugin.ts#L82. I don't think Flow writes that file, only reads it.

flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
tepi: Improved the log message

flow-server/src/main/java/com/vaadin/flow/internal/menu/MenuRegistry.java
tltv: Ah there it is. As long as _outDir there is the path being copied by Flow in dev_bundle, and finally zipped in bundles folder, then it should be good.

tltv (approved):

PR #20346, opened 2024-10-28
fix: remove tokenFile after production build

### LINKED ISSUE #20294: Production build breaks future development
### Description of the bug

After building a production build of my 24.5 application, development mode is broken. 

Initially, I was missing the dev mode gizmo and no changes were reloaded

After running `mvn clean`, I got an empty screen with the following error in the browser console 

```
copilot-Cn85U7pa.js:5073 Uncaught TypeError: Cannot read properties of undefined (reading 'push')
    at copilot-Cn85U7pa.js:5073:20
```

Finally, after deleting `src/main/bundles/prod.bundle`, did the app start working again as expected. 

### Expected behavior

Creating a production build should not break development mode. I should continue to be able to run `Application.java` and have access to the dev mode gizmo and its features. I should not need to run `mvn clean` or delete any files to get my app back to a working state.

App from start.vaadin.com with Flow and Hilla views. 

### Versions

- Vaadin / Flow version: 24.5
- Java version: 21

caalador:
Clean the token file after
production build/execution
to not make ide app execution
run app in production mode.

Closes #20294

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -778,6 +778,7 @@ public class BuildFrontendUtil {
 
             FileUtils.write(tokenFile, JsonUtil.stringify(buildInfo, 2) + "\n",
                     StandardCharsets.UTF_8.name());
+            tokenFile.deleteOnExit();
         } catch (IOException e) {
             adapter.logWarn("Unable to read token file", e);
         }


tltv (commented):
Removal of the token file after build-frontend would also mean one can't run the app in production mode directly by running Spring application class in IDE.
I don't see anything in documentation telling that it should be possible so therefore it should be OK. Running with maven/gradle or running the package directly still works as usual.

A test would be good however.

caalador:
The issue with the test is just that it should check that the file is missing after the JVM closes.
`BuildFrontendUtilTest` in flow-plugin-base contains lots of tests that the token file exists and that it updates the file contents correctly.

tltv (approved):

Issue #20296, opened 2024-10-22
Assertion error when adding and removing classname


samulivaadin:
### Description of the bug

Adding and removing classnames in a Flow component like so

addClassName("a-style");
removeClassName("a-style");

will invoke assertion error when asserts are enabled in JVM (-ea param):
 java.lang.AssertionError: null
	at com.vaadin.flow.internal.nodefeature.NodeList.generateChangesFromEmpty(NodeList.java:392) ~[flow-server-23.5.4.jar:23.5.4]
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183) ~[na:na]
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179) ~[na:na]
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:992) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[na:na]
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150) ~[na:na]
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[na:na]
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596) ~[na:na]

### Expected behavior

No assertion error thrown

### Minimal reproducible example

[flow-classname-assertion.zip](https://github.com/user-attachments/files/17470821/flow-classname-assertion.zip)

### Versions

- Vaadin / Flow version:23.5.8
- Java version:11
- OS version:
- Browser version (if applicable):
- Application Server (if applicable):
- IDE (if applicable):

tltv:
This [same issue](https://github.com/vaadin/flow/issues/18099) was already fixed in 24.5 in June. 

Cherry picked fix now also to 23.5, 24.3 and 24.4.

PR #19489, opened 2024-05-31
Fix assertion error after removing a class name

onuridrisoglu:
## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes #18099

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.

PR #20349, opened 2024-10-28
Fix assertion error after removing a class name (#19489) (CP: 24.4)

tltv:
RelatedTo: #20296


tltv:
Removing as a duplicate. Automatic cherry-pick was just a bit slow to react.

PR #20360, opened 2024-10-28
fix: prevent Optional capture in serializable predicate

mcollovati:
Optional is not serializable, so capturing it into a SerializablePredicate will prevent serialization to succeed.


tepi (approved):

mcollovati:
@tepi @mshabarov  do you think we should also address #20361 in this PR?

mshabarov:
@mcollovati yes, let's address here all these reported issues with Optional, if possible.

mcollovati:
#20361 has been closed because the real culprit of the serialization issue was in `AbstractListDataView`.
No changes, the PR can be re-approved and merged

PR #20364, opened 2024-10-29
feat: Fire an event when all hotswap operations have completed

Artur-:


mcollovati (dismissed):
Looks good.
Does it make sense to have a generic `HotswapEvent` (and maybe an `HotswapSessionEvent`) and refactor `VaadinHotswapper` adding methods to have it as a parameter, deprecating the existing ones? (default implementation would call the existing ones until they will be removed)

mcollovati (approved):

Artur-:
The main reason to have an event is to be able to extend or modify it without breaking existing code. A common super class does not seem super important but maybe good design if you introduce event classes for the other methods, which also could be a good idea. For a separate PR :)

PR #20390, opened 2024-11-01
fix: Theme asset collection and ComponentTracker path check

caalador:
Fix collection of theme assets
in windows. Fix componentTracker
patch check in windows.


tltv (approved):
Verified with `test-themes` integration test and assets works again in Windows with this change.

Issue #19948, opened 2024-09-13
Implement a possibility to optimize development environment for Vaadin/Hilla dev without components


johannest:
### Describe your motivation

Use case: an application development where no Vaadin components are used, but only Vaadin/Hilla TypeScript generation and middleware are utilized. 

For production builds unused Vaadin/React component will be removed more or less automatically. But, to be able to speed up the development environment it would be good to have a possibility to exclude all Vaadin components from the package.json.

### Describe the solution you'd like

A setting or a documented dependency exclusion for excluding Vaadin components from the package.json.

Legioth:
To clarify, this means that neither `@vaadin/react-components` nor any WebComponent like `@vaadin/button` should be included in the application's `package.json` while still including framework parts like `@vaadin/hilla-frontend` and `react-router-dom`.

mshabarov:
**Acceptance Criteria**
- [ ] Vaadin provides an opt out option that during build time excludes Vaadin web-components packages from `package.json`, so these packages are not installed which makes build and development faster
- [ ] In conjunction with `react.enable=false` and excluding `flow-react` dependency, this opt-out option should result in excluding Vaadin React and Web components from project, e.g. neither `@vaadin/react-components` nor any `@vaadin/button` installed
- [ ] Even after these exclusions, `package.json` includes npm packages that are needed for Hilla and Flow frameworks to set up routing, code generation, theming and other core features, e.g. `@vaadin/hilla-frontend`, `react-router-dom`, `@vaadin/router` and etc.

Proposed solution and implementation details: 
- The implementation can be placed into `TaskUpdatePackages::updatePackageJsonDependencies` that has a filter for React already.
- Platform needs a meta data for web-components, so that Flow would know what exactly to exclude.

mshabarov:
Closed, as all three related PR are in.

PR #20392, opened 2024-11-01
feat: opt-out web components from package.json

tltv:
Adds new property `npm.excludeWebComponents` (or `npmExcludeWebComponents` in Maven configurations). By default, it's `false` and everything works as before. `true` will exclude all web component dependencies from `package.json` for development mode (Vite/dev bundle) and production bundle build. Excluded dependencies are all Vaadin core components (e.g. button, grid, login, etc.) and commercial components (e.g. charts, rich-text-editor, etc.), but not lumo/material themes.

RelatedTo: #19948

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -166,7 +166,8 @@ public class BuildFrontendUtil {
                 .withHomeNodeExecRequired(adapter.requireHomeNodeExec())
                 .setJavaResourceFolder(adapter.javaResourceFolder())
                 .withProductionMode(false).withReact(adapter.isReactEnabled())
-                .withIncludeWebComponentNpmPackages(adapter.isIncludeWebComponentNpmPackages());
+                .withIncludeWebComponentNpmPackages(
+                        adapter.isIncludeWebComponentNpmPackages());
 
         // Copy jar artifact contents in TaskCopyFrontendFiles
         options.copyResources(adapter.getJarFiles());
@@ -341,7 +342,8 @@ public class BuildFrontendUtil {
                     .withCiBuild(adapter.ciBuild())
                     .withForceProductionBuild(adapter.forceProductionBuild())
                     .withReact(adapter.isReactEnabled())
-                    .withIncludeWebComponentNpmPackages(adapter.isIncludeWebComponentNpmPackages());
+                    .withIncludeWebComponentNpmPackages(
+                            adapter.isIncludeWebComponentNpmPackages());
             new NodeTasks(options).execute();
         } catch (ExecutionFailedException exception) {
             throw exception;
@@ -408,7 +410,8 @@ public class BuildFrontendUtil {
                     .skipDevBundleBuild(adapter.skipDevBundleBuild())
                     .withCompressBundle(adapter.compressBundle())
                     .withReact(adapter.isReactEnabled())
-                    .withIncludeWebComponentNpmPackages(adapter.isIncludeWebComponentNpmPackages());
+                    .withIncludeWebComponentNpmPackages(
+                            adapter.isIncludeWebComponentNpmPackages());
             new NodeTasks(options).execute();
         } catch (ExecutionFailedException exception) {
             throw exception;
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -762,6 +762,7 @@ public class BuildFrontendUtil {
             buildInfo.remove(Constants.CONNECT_OPEN_API_FILE_TOKEN);
             buildInfo.remove(Constants.PROJECT_FRONTEND_GENERATED_DIR_TOKEN);
             buildInfo.remove(InitParameters.BUILD_FOLDER);
+            buildInfo.remove(InitParameters.NPM_EXCLUDE_WEB_COMPONENTS);
             // Premium features flag is always true, because Vaadin CI server
             // uses Enterprise sub, thus it's always true.
             // Thus, resets the premium feature flag and DAU flag before asking
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -166,8 +166,8 @@ public class BuildFrontendUtil {
                 .withHomeNodeExecRequired(adapter.requireHomeNodeExec())
                 .setJavaResourceFolder(adapter.javaResourceFolder())
                 .withProductionMode(false).withReact(adapter.isReactEnabled())
-                .withIncludeWebComponentNpmPackages(
-                        adapter.isIncludeWebComponentNpmPackages());
+                .withNpmExcludeWebComponents(
+                        adapter.isNpmExcludeWebComponents());
 
         // Copy jar artifact contents in TaskCopyFrontendFiles
         options.copyResources(adapter.getJarFiles());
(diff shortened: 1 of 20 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/ExclusionFilter.java
```diff
@@ -92,11 +110,27 @@ private List<String> getExclusions() throws IOException {
...
-                    reactEnabled);
-            return convert.getExclusions();
+            if (!includeWebComponentNpmPackages) {
+                // By default get excluded dependencies for react mode, plus
+                // include all react mode dependencies to get a full list of web
+                // component dependencies for both lit and react mode.
+                VersionsJsonConverter convert = new VersionsJsonConverter(
+                        Json.parse(IOUtils.toString(content,
+                                StandardCharsets.UTF_8)),
+                        true, true);
```
caalador, line 120: Should this be `true, false);` as this is in the do not include webcomponent packages.

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
```diff
@@ -74,16 +78,22 @@ class VersionsJsonConverter {
...
+
     private static Logger getLogger() {
         return LoggerFactory.getLogger(VersionsJsonConverter.class);
     }
 
-    VersionsJsonConverter(JsonObject platformVersions,
-            boolean collectReactComponents) {
-        this.reactEnabled = collectReactComponents;
+    VersionsJsonConverter(JsonObject platformVersions, boolean reactEnabled,
+            boolean includeWebComponents) {
```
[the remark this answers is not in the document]
tltv: Yes. `NodeUpdater` calls it with false if `includeWebComponentNpmPackages` is false.

flow-server/src/main/java/com/vaadin/flow/server/frontend/ExclusionFilter.java
tltv: It's calling it with false to read exclusion list from `react-components` which is in `react` mode. It's a trick, but this way there's no need to maintain exclusion list in other place in version.json or no need for extra flags.

tltv: I'll update the comment, it's misleading a bit.

flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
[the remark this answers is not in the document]
caalador: Right. `new VersionsJsonConverter` was just cut out on the comparison due to the amount of new lines due to formatting.

flow-server/src/main/java/com/vaadin/flow/server/frontend/ExclusionFilter.java
caalador: Having react as `true` seems fine for simplicity, but why is the `includeWebComponents` also given as `true` here?

tltv: With false it wouldn't even reach far enough to read the exclusion list. Exclusions are read only from included nodes.

caalador: I feel this will generate issues later when changes are needed. Can ewe somehow change this so that we wouldn't have conflicting values.

tltv: Logic could be moved inside VersionsJsonConverter to populate exclusion list always if includeWebComponents=false.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
```diff
@@ -470,6 +470,62 @@ public void testGetPlatformPinnedDependencies_reactNotAvailable_noReactComponent
...
+                    .getPlatformPinnedDependencies();
+
+            // @vaadin/button doesn't have 'mode' set, so it should be included
+            Assert.assertTrue(pinnedVersions.hasKey("@vaadin/button"));
+            Assert.assertFalse(
+                    pinnedVersions.hasKey("@vaadin/react-components"));
+            Assert.assertFalse(
+                    pinnedVersions.hasKey("@vaadin/react-components-pro"));
+        } finally {
+            options.withIncludeWebComponentNpmPackages(true);
```
caalador, line 500: There should be no need to reset the options as the options is generated `Before` and each test will get the same default initialized Options.

flow-server/src/test/java/com/vaadin/flow/server/frontend/NodeUpdaterTest.java
```diff
@@ -470,6 +470,62 @@ public void testGetPlatformPinnedDependencies_reactNotAvailable_noReactComponent
...
+            // @vaadin/button doesn't have 'mode' set, so it should be included
+            Assert.assertTrue(pinnedVersions.hasKey("@vaadin/button"));
+            Assert.assertFalse(
+                    pinnedVersions.hasKey("@vaadin/react-components"));
+            Assert.assertFalse(
+                    pinnedVersions.hasKey("@vaadin/react-components-pro"));
+        } finally {
+            options.withReact(true);
+            options.withIncludeWebComponentNpmPackages(true);
+        }
```
caalador, line 524: no need to reset the options

flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
```diff
@@ -285,4 +285,9 @@ public class InitParameters implements Serializable {
      */
     public static final String APPLICATION_IDENTIFIER = "applicationIdentifier";
 
+    /**
+     * Configuration name for including npm packages for web components.
+     */
+    public static final String INCLUDE_WEB_COMPONENT_NPM_PACKAGES = "include.web.component.npm.packages";
```
caalador, line 291: `includeWebComponents` could be enough

tltv: Camel-case yes. I would still keep the npm in the name just to be clear it's only about npm packages. Also as suggested in daily, changing this to `exclude` could be better as it's only supposed to be used when excluding anyway.
So maybe `npm.excludeWebComponents` would work.

flow-server/src/main/java/com/vaadin/flow/server/frontend/ExclusionFilter.java
tltv: Noticed a mistake in the last change. If we exclude both react+lit moded dependencies, then we would also exclude vaadin-lumo-styles and vaadin-material-styles which are not in exclude list of react-components and should not be excluded.

caalador:
Just running a project with `mvn -Dnpm.excludeWebComponents=true` or having 
```
<configuration>
  <npmExcludeWebComponents>true</npmExcludeWebComponents>
</configuration>
```
I still get in my `package.json` the lines:
```
    "@vaadin/react-components": "24.6.0-alpha7",
    "@vaadin/react-components-pro": "24.6.0-alpha7",
```

Am I missing something or is the value not propagated correctly?

tltv:
> Am I missing something or is the value not propagated correctly?

Try again. Now it should be written in build-info from maven configs and delegated forward.
It has worked for me with application.properties and just now also with maven prop. Need try with `-Dnpm.excludeWebComponents=true` too

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
```diff
@@ -263,6 +266,10 @@ public static File propagateBuildInfo(PluginAdapterBase adapter) {
         }
 
         buildInfo.put(REACT_ENABLE, adapter.isReactEnabled());
+        if (adapter.isNpmExcludeWebComponents()) {
+            buildInfo.put(NPM_EXCLUDE_WEB_COMPONENTS,
```
caalador, line 270: As this is a build/dev time data it should probably be cleaned in `updateBuildFile`
Else seems ok and works through plugin con and system property

caalador (approved):

PR #20394, opened 2024-11-04
fix: support serialization of Vaadin scoped beans

### LINKED ISSUE #19967: Serialization of scoped beans fails due to the missing UI instance
### Description of the bug

This issue is tightly connected to kubernetes kit issue: https://github.com/vaadin/kubernetes-kit/issues/140

While discussing with @mcollovati he discovered that as a part of fixing the issue above, there is a need for a fix also in the Flow:

"We need to set the UI threadlocal. Unfortunately we cannot do it in the UI class, because the reference to the bean is held by its parent class Component that is serialized before UI.writeObject() gets called. But we can "fix" this in Flow by adding the thread local logic in Component class. Once this is done, the serialization completes correctly (also needs setting threadlocal in VaadinSession writeObject)"

### Expected behavior

Serialization and the de-serialization works also when the application contains scoped beans (@SessionScope, @UIScope, @RouteScope).

See https://github.com/vaadin/kubernetes-kit/issues/140

The Flow fix could be something along these lines.

**com.vaadin.flow.component.Component. Add a Thread Local logic:**
```
@Serial
    private void readObject(ObjectInputStream stream)
            throws IOException, ClassNotFoundException {
        Optional<UI> uiOptional = getUI();
        if (uiOptional.isPresent()) {
            UI ui = uiOptional.get();
            Map<Class<?>, CurrentInstance> old = CurrentInstance.setCurrent(ui);
            try {
                stream.defaultReadObject();
            } finally {
                CurrentInstance.restoreInstances(old);
            }
        }
    }

    @Serial
    private void writeObject(java.io.ObjectOutputStream stream)
            throws IOException {
        Optional<UI> uiOptional = getUI();
        if (uiOptional.isPresent()) {
            UI ui = uiOptional.get();
            Map<Class<?>, CurrentInstance> old = CurrentInstance.setCurrent(ui);
            try {
                CurrentInstance.setCurrent(ui);
                stream.defaultWriteObject();
            } finally {
                CurrentInstance.restoreInstances(old);
            }
        }
    }
```

**VaadinSession improve the existing writeObject with a Thread Local handling:**
```
private void writeObject(java.io.ObjectOutputStream stream)
            throws IOException {
        boolean serializeUIs = true;

        // If service is null it has just been deserialized and should be
        // serialized in
        // the same way again
        if (getService() != null) {
            ApplicationConfiguration appConfiguration = ApplicationConfiguration
                    .get(getService().getContext());
            if (!appConfiguration.isProductionMode() && !appConfiguration
                    .isDevModeSessionSerializationEnabled()) {
                serializeUIs = false;
            }
        }

        Map<Class<?>, CurrentInstance> old = CurrentInstance.setCurrent(this);
            try {
                CurrentInstance.setCurrent(this);
                stream.defaultWriteObject();
            if (serializeUIs) {
                stream.writeObject(uIs);
                stream.writeObject(resourceRegistry);
            } else {
                stream.writeObject(new HashMap<>());
                stream.writeObject(new StreamResourceRegistry(this));
            }
        } finally {
            CurrentInstance.restoreInstances(old);
        }
```

### Versions

- Vaadin / Flow version: 24.4.11
- Java version: 17
- OS version: Mac (M2), Kubernetes Kit 2.2.2

mcollovati:
## Description

Makes sure that VaadinSession and UI thread locals are available during both serialization and deserialization, to allow other libraries to perform inspection and injection of Vaadin scoped beans.

Also refactors VaadinRouteScope to be independent from VaadinService when fetching RouteScopeOwner annotation for the bean, and replaces VaadinSession.unlock() calls with direct access to the lock instance to prevent unwanted push during bean lookup.

Fixes #19967
Part of vaadin/kubernetes-kit#140

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/Component.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/Component.java
@@ -15,10 +15,15 @@
  */
 package com.vaadin.flow.component;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serial;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Stream;
 import java.util.stream.Stream.Builder;
@@ -32,6 +37,7 @@ import com.vaadin.flow.dom.PropertyChangeListener;
 import com.vaadin.flow.dom.ShadowRoot;
 import com.vaadin.flow.i18n.I18NProvider;
 import com.vaadin.flow.internal.AnnotationReader;
+import com.vaadin.flow.internal.CurrentInstance;
 import com.vaadin.flow.internal.LocaleUtil;
 import com.vaadin.flow.internal.nodefeature.ElementData;
 import com.vaadin.flow.server.Attributes;
@@ -818,6 +824,40 @@ public abstract class Component
      */
     public void removeFromParent() {
         getElement().removeFromParent();
+    }
+
+    @Serial
+    private void writeObject(ObjectOutputStream out) throws IOException {
+        if (this instanceof UI ui) {
+            Map<Class<?>, CurrentInstance> instances = CurrentInstance
+                    .setCurrent(ui);
+            try {
+                out.defaultWriteObject();
+            } finally {
+                CurrentInstance.restoreInstances(instances);
+            }
+        } else {
+            out.defaultWriteObject();
+        }
+    }
+
+    @Serial
+    private void readObject(ObjectInputStream in)
+            throws IOException, ClassNotFoundException {
+        if (this instanceof UI ui) {
+            Map<Class<?>, CurrentInstance> instances = CurrentInstance
+                    .getInstances();
+            // Cannot use CurrentInstance.setCurrent(this) because it will try
+            // to get VaadinSession from UI.internals that is not yet available
+            CurrentInstance.set(UI.class, ui);
+            try {
+                in.defaultReadObject();
+            } finally {
+                CurrentInstance.restoreInstances(instances);
+            }
+        } else {
+            in.defaultReadObject();
+        }
     }
 
 }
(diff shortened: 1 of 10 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
```diff
@@ -369,10 +355,52 @@ private NavigationListener getNavigationListener() {
 
     }
 
+    private record ObjectWithOwner(Object object, RouteScopeOwner owner) {
```
mshabarov, line 358: Does this need `implements Serializable`?

mshabarov (commented):
Do we need to double check CDI and Quarkus add-ons for the similar changes?

mcollovati:
> Do we need to double check CDI and Quarkus add-ons for the similar changes?

Currently, custom logic for handling scoped beans during deserialization is implemented only in Kubernetes Kit and only for Spring.

flow-server/src/main/java/com/vaadin/flow/component/Component.java
```diff
@@ -820,4 +826,38 @@ public void removeFromParent() {
...
+                out.defaultWriteObject();
+            } finally {
+                CurrentInstance.restoreInstances(instances);
+            }
+        } else {
+            out.defaultWriteObject();
+        }
+    }
+
+    @Serial
```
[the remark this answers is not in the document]
mcollovati: Yeah, makes sense, even though it is not required

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
mcollovati: No. The object is not stored anywhere. It is only used as a carrier of information for other logics.

mshabarov (approved):

PR #20397, opened 2024-11-04
feat: Add support for customizable projectFileExtensions

### LINKED ISSUE #19527: Include `.scss` and `.sass` in `stats.json` hashes

caalador:
Fixes #19527

### DIFF
--- a/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
+++ b/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
@@ -283,11 +283,11 @@ public abstract class VaadinFlowPluginExtension @Inject constructor(private val 
     public abstract val applicationIdentifier: Property<String>
 
     /**
-     * The list of file extensions that are considered project files.
+     * The list of extra file extensions that are considered project files.
      * Hashes are calculated for these files as part of detecting if a new prod
      * bundle should be generated.
      */
-    public abstract val projectFileExtensions: ListProperty<String>
+    public abstract val extraProjectFileExtensions: ListProperty<String>
 
     public fun filterClasspath(@DelegatesTo(value = ClasspathFilter::class, strategy = Closure.DELEGATE_FIRST) block: Closure<*>) {
         block.delegate = classpathFilter
@@ -449,6 +449,9 @@ public class PluginEffectiveConfiguration(
     public val projectFileExtensions: ListProperty<String> = extension.projectFileExtensions
             .convention(listOf(".js", ".js.map", ".ts", ".ts.map", ".tsx", ".tsx.map", ".css", ".css.map"))
 
+    public val extraProjectFileExtensions: ListProperty<String> = extension.extraProjectFileExtensions
+            .convention(emptyList())
+
     /**
      * Finds the value of a boolean property. It searches in gradle and system properties.
      *
@@ -510,7 +513,7 @@ public class PluginEffectiveConfiguration(
             "frontendHotdeploy=${frontendHotdeploy.get()}," +
             "reactEnable=${reactEnable.get()}," +
             "cleanFrontendFiles=${cleanFrontendFiles.get()}" +
-            "projectFileExtensions=${projectFileExtensions.get()}" +
+            "extraProjectFileExtensions=${extraProjectFileExtensions.get()}" +
             ")"
     public companion object {
         public fun get(project: Project): PluginEffectiveConfiguration =
--- a/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
+++ b/flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
@@ -282,6 +282,13 @@ public abstract class VaadinFlowPluginExtension @Inject constructor(private val 
 
     public abstract val applicationIdentifier: Property<String>
 
+    /**
+     * The list of file extensions that are considered project files.
+     * Hashes are calculated for these files as part of detecting if a new prod
+     * bundle should be generated.
+     */
+    public abstract val projectFileExtensions: ListProperty<String>
+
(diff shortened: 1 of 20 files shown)


flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
```diff
@@ -500,6 +511,7 @@ public class PluginEffectiveConfiguration(
             "frontendHotdeploy=${frontendHotdeploy.get()}," +
             "reactEnable=${reactEnable.get()}," +
             "cleanFrontendFiles=${cleanFrontendFiles.get()}" +
```
tltv, line 513: Missing comma

flow-server/src/main/java/com/vaadin/flow/server/InitParameters.java
```diff
@@ -285,4 +285,5 @@ public class InitParameters implements Serializable {
      */
     public static final String APPLICATION_IDENTIFIER = "applicationIdentifier";
 
+    public static final String FRONTEND_EXTRA_EXTENSIONS = "frontendExtraFileExtensions";
```
tltv, line 288: What's the correct naming pattern here, is it like this or "frontend.extra.file.extensions"? We probably should keep them more consistent. I added just new property in other PR with dot format based on majority of properties in InitParameters class.. But if camel case is more correct today, then the other PR should be changed.

tltv (changes requested):
Should also gradle plugin's `PrepareFrontendInputProperties.kt` be updated for the new property?
Also `GenerateNpmBOMMojo` maven goal could need to use this new property.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
```diff
@@ -118,6 +122,8 @@ private void createGeneratedConfig() throws IOException {
                 .getResource(FrontendUtils.VITE_GENERATED_CONFIG);
         String template = IOUtils.toString(resource, StandardCharsets.UTF_8);
 
+        System.out.println(
+                "=== '" + options.getFrontendExtraFileExtensions() + "'");
```
mcollovati, line 126:
```suggestion
        System.out.println(
                "=== '" + options.getFrontendExtraFileExtensions() + "'");
```

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
```diff
@@ -282,6 +282,13 @@ public abstract class VaadinFlowPluginExtension @Inject constructor(private val
 
     public abstract val applicationIdentifier: Property<String>
 
+    /**
+     * The list of extra file extensions that are considered project files.
+     * Hashes are calculated for these files as part of detecting if a new prod
+     * bundle should be generated.
```
mcollovati, line 288: Is it only for prod bundle or also for dev bundle?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
```diff
@@ -127,14 +133,29 @@ private void createGeneratedConfig() throws IOException {
...
     }
 
+    private String getFrontendExtraFileExtensions() {
+        Optional<List<String>> frontendExtraFileExtensions = Optional
+                .ofNullable(options.getFrontendExtraFileExtensions());
+        if (frontendExtraFileExtensions.isPresent()
+                && frontendExtraFileExtensions.get().size() > 0) {
+            return frontendExtraFileExtensions.get().stream()
+                    .map(ext -> ext.replace("'", "\\'"))
+                    .map(ext -> ext.startsWith(".") ? ext : "." + ext)
```
mcollovati, line 153: Does it make sense to also trim values?
For example, I can imagine configurations like 
`<frontendExtraFileExtensions>.ico, .png</frontendExtraFileExtensions>`

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
caalador: I forgot to push the removal =P

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateVite.java
caalador: added trim

flow-plugins/flow-gradle-plugin/src/main/kotlin/com/vaadin/gradle/VaadinFlowPluginExtension.kt
caalador: both. Changed to only say new bundle

caalador:
Added input property for the gradle plugin.

tltv (approved):
LGTM. After conflicts are resolved.

tltv (approved):

Issue #20365, opened 2024-10-29
Synchronized property unexpectedly reverts to last synchronized state


vursen:
### Description of the bug

Flow reverts the property’s value to the last synchronized state if the property changes again while the previous update is scheduled or is still being processed on the server. This behavior has blocked progress on [DatePicker fallback parser feature](https://github.com/vaadin/flow-components/pull/6743).

https://github.com/user-attachments/assets/7ab2dd8b-6a47-48b5-8fee-8bbd5d0aa859

### Expected behavior

I'd expect Flow to keep the new value and wait for the next synchronization event to send it to the server.

### Minimal reproducible example

```ts
// frontend/my-field/my-field.ts

import { html, LitElement } from 'lit';
import { customElement, query } from 'lit/decorators.js';

@customElement("my-field")
class MyField extends LitElement {
  @query('input')
  inputElement!: HTMLInputElement;

  get inputElementValue() {
    return this.inputElement.value;
  }

  set inputElementValue(value) {
    console.warn('set inputElementValue', value);
    this.inputElement.value = value;
  }

  onChange() {
    this.dispatchEvent(new CustomEvent('change', { bubbles: true, composed: true }));
    this.inputElementValue = 'foobar'; // <------ Change the property right after the synchronization event
  }

  render() {
    return html`<input @change="${this.onChange}" />`
  }
}
```

```java
@Tag("my-field")
@JsModule("my-field/my-field.ts")
public class MyField extends Component {
    @Synchronize(property = "inputElementValue", value = { "change" })
    public String getInputElementValue() {
        return getElement().getProperty("inputElementValue", "");
    }
}
```

```java
@Route("my-field")
public class MyFieldView extends Div {
    public MyFieldView() {
        MyField myField = new MyField();
        add(myField);
    }
}
```

### Versions

- Vaadin / Flow version: 24.5 and possibly earlier
- Java version: 17
- OS version: Mac OS

vursen:
Actually, there is a more real example. I can reproduce the bug using a TextField with a slow internet connection:

```java
TextField textField = new TextField();
add(textField);
```

1. Enable network throttling in browser devtools
2. Type "test" and press <kbd>Enter</kbd>
3. After a round-trip begins, press <kbd>Backspace</kbd>
4. Result: The backspace is discarded when the round-trip completes

https://github.com/user-attachments/assets/b31940fc-89e0-417b-a4f4-f2b18555c6b3

tepi:
[...]

PR #20431, opened 2024-11-07
fix: flush logic

tepi:
Changes client side value binding logic so that if the user modifies the value during server round-trip, the value earlier sent to the server no longer overwrites user's changes once the round-trip finishes. Instead, user's changes are preserved.

Fixes https://github.com/vaadin/flow/issues/20365


tepi:
Ready for review

tltv (approved):

PR #20446, opened 2024-11-12
fix: multiple fast navigate calls

### LINKED ISSUE #20404: Javascript error: Invalid blocker state transition
### Description of the bug

When I'm navigating in my Vaadin application I have some javascript errors.
> Uncaught (in promise) Error: Invalid blocker state transition: unblocked -> proceeding

image

image

### Expected behavior

A javascript error should not appear in a normal case of usage.

Create a Vaadin application with two views in a menu, clicks multiple times on the menu.

I think the error appears when you click on the menu when the application is already navigating so if the view is really fast to load you need to click fast.

### Versions

- Vaadin / Flow version: 24.5.2

caalador:
Fix issue where a slow connection
and fast `navigate` calls throws
exception due to faulty blocker
state change.

Fixes #20404


mshabarov (approved):
The patch looks good.
Smoke tested it, couldn't spot anything weird.
However, I couldn't reproduce the original issue.

Another pair of eyes would be beneficial.

caalador:
Easiest way to replicate original issue is to have a hilla layout and 2 flow layouts, setting the network throttling to slow 4g , then navigating to a flow view from a flow view through the layout menu and clicking 2 times on the menu item.

mshabarov (changes requested):
We talked with @platosha on Friday, this patch slightly changes how navigation works: with a slow network, when you first time click on the menu link and immediately click on the second one, the application will ignore the second navigation and basically all other navigations until the first one completes.
Maybe it's better to queue these navigations and apply the latest in the end. Flow navigation works like this IIRC.
@platosha promised to see how we can do it.
@caalador do you think it's better to queue like I described?

caalador:
We could push the navigate to `queuedNavigate` instead of canceling the navigation. I think we have enough information to do that.

The issue is that when the programmer uses react-router `navigate()` directly when on a Flow view it bypasses all the features for navgation we have except for the `blocker` and this will start multiple blockers at the same time setting the blocker state multiple times for the same instance as the latter one overwrites the previous one without checking if it is blocking already or not.

mshabarov (commented):
Thanks for the tip, the exception has eventually appeared for me and after the latest changes I don't see it anymore and the multiple navigations are queued and handled as I expected - the last click wins in the end.

I think we can try adding a test that simulates the same steps, I wonder if these codes works:
```java
public class UseBlockerIT extends ChromeDeviceTest {

    @Test
    public void test() {
        ChromeDriver driver = (ChromeDriver) getDriver();
        DevTools devTools = driver.getDevTools();
        devTools.send(Network.enable(Optional.empty(), Optional.empty(), Optional.empty()));
        devTools.send(Network.emulateNetworkConditions(
                false, // offline
                100,   // latency (ms)
                750000, // download through (bits per second)
                250000, // upload throughput (bits per second)
                Optional.of(ConnectionType.CELLULAR4G), // connection type
                Optional.empty(),
                Optional.empty(),
                Optional.empty()
        ));
        // open a page
        // click twice
        // check for errors
        // check the latest navigation
    }
}
```
If Flow repo is not a suitable place for it, could we add it to platform test module?

caalador:
Couldn't get the test to fail for multiple fasts clicks when clicking programmatically.
Setting `network_conditions` before open made the test really slow indicating that it probably does work, but it doesn't help with the test. Adding either a timeout 0 or animation frame makes the second click throw due to undefined so it is too late.

flow-tests/test-react-router/pom.xml
```diff
@@ -38,6 +38,19 @@
...
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-react</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.seleniumhq.selenium</groupId>
+            <artifactId>selenium-chrome-driver</artifactId>
+            <version>4.26.0</version>
```
mshabarov, line 50: I'm worrying about maintenance for this: future Chrome updates can be incompatible with the ChromeDriver, so we have to bump this version all the time.
If the test always passes, maybe we can disable it, so it will show case the failing scenario, but won't run in CI.

mshabarov (approved):

flow-tests/test-react-router/pom.xml
caalador: Removed test and added  doc on the view class.

mshabarov (dismissed):

mshabarov (approved):

PR #20470, opened 2024-11-14
feat: add Vite variable for determining service worker build context

krissvaa:
Hilla needs to dynamically import modules/libraries for the service workers, so there is some property to determine if the build is for SW or base.  

Fixes https://github.com/vaadin/hilla/issues/2867


mshabarov:
Closed as stale, the referenced issue seems fixed.

Issue #19938, opened 2024-09-11
Potential deadlock with Jetty


mperktold:
### Description of the bug

We have several reports of our application not being able to shutdown. Apparently, some VaadinSessions stay alive and cannot be destroyed. Here the stack dumps of two such cases:
[StackTraces1.txt](https://github.com/user-attachments/files/16964574/StackTraces1.txt)
[StackTraces2.txt](https://github.com/user-attachments/files/16964572/StackTraces2.txt)

I found some common patterns in these dumps:

One of the threads blocks on a Jetty semaphore while reading the request content of an UIDL request. Note that this thread holds the lock on the VaadinSession while blocking.
```
java.base@21.0.4/jdk.internal.misc.Unsafe.park(Native Method)
java.base@21.0.4/java.util.concurrent.locks.LockSupport.park(Unknown Source)
java.base@21.0.4/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionNode.block(Unknown Source)
java.base@21.0.4/java.util.concurrent.ForkJoinPool.unmanagedBlock(Unknown Source)
java.base@21.0.4/java.util.concurrent.ForkJoinPool.managedBlock(Unknown Source)
java.base@21.0.4/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(Unknown Source)
app//org.eclipse.jetty.ee10.servlet.AsyncContentProducer$LockedSemaphore.acquire(AsyncContentProducer.java:393)
app//org.eclipse.jetty.ee10.servlet.BlockingContentProducer.nextChunk(BlockingContentProducer.java:119)
app//org.eclipse.jetty.ee10.servlet.HttpInput.read(HttpInput.java:245)
app//org.eclipse.jetty.ee10.servlet.HttpInput.read(HttpInput.java:226)
java.base@21.0.4/sun.nio.cs.StreamDecoder.readBytes(Unknown Source)
java.base@21.0.4/sun.nio.cs.StreamDecoder.implRead(Unknown Source)
java.base@21.0.4/sun.nio.cs.StreamDecoder.lockedRead(Unknown Source)
java.base@21.0.4/sun.nio.cs.StreamDecoder.read(Unknown Source)
java.base@21.0.4/java.io.InputStreamReader.read(Unknown Source)
java.base@21.0.4/java.io.BufferedReader.read1(Unknown Source)
java.base@21.0.4/java.io.BufferedReader.implRead(Unknown Source)
[...]

PR #20475, opened 2024-11-14
Fix: move blocking calls outside session lock (#19938)

tepi:


flow-server/src/main/java/com/vaadin/flow/server/SynchronizedRequestHandler.java
```diff
@@ -58,13 +84,60 @@ public boolean handleRequest(VaadinSession session, VaadinRequest request,
...
      */
     public abstract boolean synchronizedHandleRequest(VaadinSession session,
             VaadinRequest request, VaadinResponse response) throws IOException;
 
+    /**
+     * Returns {@literal true} if
+     * {@link #synchronizedHandleRequest(VaadinSession, VaadinRequest, VaadinResponse, String)}
+     * should be called. Returns {@literal false} if
+     * {@link #synchronizedHandleRequest(VaadinSession, VaadinRequest, VaadinResponse)}
+     * should be called.
```
mcollovati, line 99: I would make "Gets if request body should be read before calling synchronizedHandleRequest." as initial documentation, to make immediately clear the effect of the flag.

```suggestion
     * Gets if request body should be read before calling synchronizedHandleRequest.
     *<p></p>
     * Returns {@literal true} if
     * {@link #synchronizedHandleRequest(VaadinSession, VaadinRequest, VaadinResponse, String)}
     * should be called. Returns {@literal false} if
     * {@link #synchronizedHandleRequest(VaadinSession, VaadinRequest, VaadinResponse)}
     * should be called.
```

flow-server/src/main/java/com/vaadin/flow/server/SynchronizedRequestHandler.java
```diff
@@ -28,18 +32,41 @@
  */
 public abstract class SynchronizedRequestHandler implements RequestHandler {
 
+    public static final int MAX_BUFFER_SIZE = 64 * 1024;
+
+    @FunctionalInterface
+    public interface ResponseWriter extends Serializable {
```
mcollovati, line 38: This public API needs Javadoc

flow-server/src/main/java/com/vaadin/flow/server/SynchronizedRequestHandler.java
```diff
@@ -58,13 +84,60 @@ public boolean handleRequest(VaadinSession session, VaadinRequest request,
...
+     */
+    public boolean isReadRequestBodyFirstEnabled() {
+        return false;
+    }
+
+    /**
+     * Identical to
+     * {@link #synchronizedHandleRequest(VaadinSession, VaadinRequest, VaadinResponse)}
+     * except the {@link VaadinSession} is locked before this is called and the
+     * response requestBody has been read before locking the session and is
```
mcollovati, line 112: Could the following be simpler to undertand?
`expect the request body is read before locking the VaadinSession and calling this method.`

flow-server/src/main/java/com/vaadin/flow/server/SynchronizedRequestHandler.java
```diff
@@ -58,13 +84,60 @@ public boolean handleRequest(VaadinSession session, VaadinRequest request,
...
+     * provided as a separate parameter. Implementations should also take care
+     * to unlock the session before writing to the response object.
```
mcollovati, line 114: > Implementations should also take care to unlock the session before writing to the response object.

Could be removed since lock is handled internally.

flow-server/src/main/java/com/vaadin/flow/server/SynchronizedRequestHandler.java
```diff
@@ -58,13 +84,60 @@ public boolean handleRequest(VaadinSession session, VaadinRequest request,
...
+     *            The request to handle
+     * @param response
+     *            The response object to which a response can be written.
+     * @param requestBody
+     *            Request body pre-read from the request object
+     * @return if this handler has handled the request and no further request
+     *         handlers should be called an {@link Optional} of
+     *         {@link ResponseWriter} is returned and it must be run after
+     *         releasing the lock on {@link VaadinSession}. Otherwise an empty
+     *         {@link Optional} is returned.
```
mcollovati, line 128: This sound a bit unclear to me. Would it make sense something like

`@return a ReponseWriter wrapped into an Optional, if this handler will write the response and no further request handlers should be called, otherwise an empty Optional.`

And maybe add to the Javadoc that the ResponseWriter will be executed after VaadinSession is unlocked.

flow-server/src/main/java/com/vaadin/flow/server/communication/UidlRequestHandler.java
```diff
@@ -102,48 +103,64 @@ protected ServerRpcHandler createRpcHandler() {
...
     public boolean synchronizedHandleRequest(VaadinSession session,
             VaadinRequest request, VaadinResponse response) throws IOException {
+        String requestBody = SynchronizedRequestHandler
+                .getRequestBody(request.getReader());
+        Optional<ResponseWriter> responseWriter = synchronizedHandleRequest(
+                session, request, response, requestBody);
+        if (responseWriter.isPresent()) {
+            responseWriter.get().writeResponse();
+        }
+        return responseWriter.isPresent();
```
mcollovati, line 113: Why is needed to override this `synchronizedHandleRequest(...)` method? Isn't this logic already applied in `SynchronizedRequestHandler` if `isReadRequestBodyFirstEnabled()` is true?

flow-server/src/main/java/com/vaadin/flow/server/SynchronizedRequestHandler.java
```diff
@@ -28,18 +32,41 @@
...
+                String requestBody = reader == null ? null
+                        : getRequestBody(reader);
+                session.lock();
+                Optional<ResponseWriter> responseWriter = synchronizedHandleRequest(
+                        session, request, response, requestBody);
+                session.unlock();
+                if (responseWriter.isPresent()) {
+                    responseWriter.get().writeResponse();
+                }
+                return responseWriter.isPresent();
```
mcollovati, line 61: It looks like we bind the response write stuff with the `isReadRequestBodyFirstEnabled`.
But I wonder if handlers that do no need the request body could benefit from the session unlock before response write feature.
Actually, it should be the default behavior. But unfortunately, this cannot be done without changing the API or providing a `VaadinResponse` wrapper that postpones the actual writing

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/communication/UidlRequestHandler.java
tepi: Well that comes from `RequestHandler` interface so I must implement something. Would it be better to just throw `UnsupportedOperationException` since this should never be called for this handler?

flow-server/src/main/java/com/vaadin/flow/server/SynchronizedRequestHandler.java
tepi: I think the deadlock issues are all in `UidlRequestHandler`. If some other handlers start to show this issue, we can use the same API to fix them (even if they don't need the request body). Just override `isReadRequestBodyFirstEnabled` to return true and then return a `ResponseWriter`. The request body might be empty but that should not matter.

tepi: Or maybe `isReadRequestBodyFirstEnabled` should be called `readAndWriteOutsideSessionLock` or something more general?

mcollovati: I would make the assumption that reading the body beforehand is directly related to writing the response without lock. So, if we do not want two separate flags (unnecessary complexity?)  I'd prefer the `readAndWriteOutsideSessionLock` name that makes the behavior clear.

mcollovati: And additional though about writing after the lock is released. Is there any chance that unlocking the session causes PUSH to send a message that will reach the client before the current response is committed (e.g. when transport is web socket)?
Probably this situation is handled by the Flow client, putting the unexpected message on hold until the previous one is received, but maybe better double check.

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/SynchronizedRequestHandler.java
tepi: Changed the name. I'll need to check the PUSH case still. Initial guess is it is handled properly since I think some IT should have exploded if it is not.

tepi: I ran some manual tests for the push case and it seems to work fine. Whatever is queued for push is sent within the same message as the final response that is written.

mcollovati (approved):

Issue #20417, opened 2024-11-06
Corrupt filename in Upload component when server's file encoding is not UTF-8


mas4ivv:
### Description of the bug

When using the Upload component with a filename that has non-ASCII characters, the server receives a corrupt filename if it's file encoding is not UTF-8.
The filename is part of the upload request and seems to be UTF-8 encoded:
![grafik](https://github.com/user-attachments/assets/48732508-5d58-4561-86ee-b5d6dd42e7c3)

The filename is then read via Apache Commons' ServletFileUpload, created by StreamReceiverHandler.getItemIterator(VaadinRequest). As there is no encoding set on creation, it falls back to the server's default.

Example running the example below from Windows command line:
![grafik](https://github.com/user-attachments/assets/dce3425e-7caf-4892-ac00-703a5096c84c)

### Expected behavior

It seems to me that the client is always sending the filename UTF-8 encoded. If this is correct, the ServletFileUpload created in StreamReceiverHandler should be configured to use UTF-8 by calling setHeaderEncoding("UTF-8").

### Minimal reproducible example

```
import java.io.ByteArrayOutputStream;

import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.component.textfield.TextField;
import com.vaadin.flow.component.upload.Upload;
import com.vaadin.flow.router.Route;

@Route("upload")
public class UploadPage extends VerticalLayout {

    public UploadPage() {
        Upload upload = new Upload();
        TextField filename = new TextField();
        upload.setReceiver((name, type) -> {
            filename.setValue(name);
            return new ByteArrayOutputStream();
        });
        this.add(upload, filename);
    }
}
```

### Versions

- Vaadin / Flow version: 23.4.1
- Java version: 11
- OS version:
- Browser version (if applicable):
- Application Server (if applicable):
- IDE (if applicable):

tepi:
@mas4ivv I was unfortunately unable to reproduce this on macOS (uses UTF-8 by default, but even changing the server JVM encoding to soemthing else did not break the special characters). Are you by any chance on Windows?

tepi:
Alright, in that case it should be reproducable on mac as well since I think most Linux default to utf-8 anyway. I'll retry.

tltv:
@mas4ivv Could you please clarify these questions to help us reproducing the issue and find good fix for it. Also if possible, attaching full example application would also answer some of these:
- which browser you use? or is it maybe same with all you try?
- which jetty version you use?
[...]

PR #20480, opened 2024-11-15
fix: fix ServletFileUpload header encoding

tltv:
This change will set ServletFileUpload's header encoding always to UTF-8 in StreamReceiverHandler when request character encoding is null. This ensures that system's default character encoding is not applied when parsing filename of the uploaded file, unless request's character encoding is set otherwise. Only for setups without multipart config for servlet.

Fixes: #20417


tepi (approved):

flow-tests/pom.xml
mcollovati, line 1: Since the PR is marked to be picked on higher version, would it make sense to apply this pom change in a separated PR?

PR #20484, opened 2024-11-15
feat!: WebPush Subscription Wrapper and feature flag removal

mshabarov:
Removes third-party `nl.martijndwars.webpush.Subscription` from Flow public API.
Removes the feature flag.


caalador (approved):
Looks good. Update needed to docs.

PR #20487, opened 2024-11-18
test: bump selenium devtools for CI system

mcollovati:


mcollovati:
CI currently failing with the following error

> Caused by: org.openqa.selenium.devtools.DevToolsException: You are using a no-op implementation of the CDP. The most likely reason for this is that Selenium was unable to find an implementation of the CDP protocol that matches your browser. Please be sure to include an implementation on the classpath, possibly by adding a new (maven) dependency of `org.seleniumhq.selenium:selenium-devtools-vNN:4.26.0` where `NN` matches the major version of the browser you're using.

> org.openqa.selenium.devtools.CdpVersionFinder findNearestMatch
WARNING: Unable to find an exact match for CDP version 127, returning the closest version; found: 123; Please update to a Selenium version that supports CDP version 127

caalador (approved):

PR #20522, opened 2024-11-20
fix: vaadin/router added with react-router

### LINKED ISSUE #20496: Vaadin Router is installed while it's not used by default
### Description of the bug

`@vaadin/router` is added to `package.json`, installed and added to a bundle in a Vaadin 24.6 application that doesn't specify a router, nor that opts-out from React.
React Router should be used by default and Vaadin Router shouldn't be added.

See e.g. this starter https://github.com/vaadin/skeleton-starter-hilla-react/blob/b54e9e0f9cdb14d47409e45b78e71373b6cac91c/package.json#L20.

### Expected behavior

`@vaadin/router` is not added to a project.

https://github.com/vaadin/skeleton-starter-hilla-react/blob/b54e9e0f9cdb14d47409e45b78e71373b6cac91c/package.json#L20

### Versions

- Vaadin / Flow version: 24.6.0.alpha3.

caalador:
Exclude vaadin/router when runing
in react mode.
Without exclusion vaadin/router
gets added from the vaadin-core.json
package in platform.

Fixes #20496

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/VersionsJsonConverter.java
@@ -70,6 +70,7 @@ class VersionsJsonConverter {
      * Mode value for dependency for all modes.
      */
     public static final String MODE_ALL = "all"; // same as empty string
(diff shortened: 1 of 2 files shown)


mshabarov (changes requested):
Looks good, but needs a test.

PR #20523, opened 2024-11-20
fix: exclude project maven dependencies from isolated class loader

mcollovati:
If the user project depends directly or transitively from maven artifacts mojos can fail at runtime because of Maven API loaded from both isolated class loader and maven.api realm.
This change prevents maven artifacts to be added to the isolated class loader.


flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/Reflector.java
```diff
@@ -223,6 +223,10 @@ private static URLClassLoader createIsolatedClassLoader(
 
         Map<String, Artifact> projectDependencies = new HashMap<>(project
                 .getArtifacts().stream()
+                // Exclude all maven artifacts to prevent class loading clash
+                // with maven.api class realm
+                .filter(artifact -> !"org.apache.maven"
```
mshabarov, line 228: Just a though: is it enough to filter out Maven core deps, or should we also filter out
```
org.codehaus.plexus.*
org.sonatype.*
```

caalador (approved):

flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/Reflector.java
mcollovati: Good point. I think we don't have any explicit reference to `org.sonatype.*` but definitely the mojo has a field referencing `org.codehaus.plexus.build.BuildContext`.
I'll double-check.

Issue #20527, opened 2024-11-21
React Router Future Flag warning


mshabarov:
### Description of the bug

 There's a number of warnings triggered by React Router when using the latest Vaadin Flow snapshot/pre-release (24.6):
⚠️ React Router Future Flag Warning: React Router will begin wrapping state updates in `React.startTransition` in v7. You can use the `v7_startTransition` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_starttransition. indexhtml-CdhAQHz9.js:51:6661
⚠️ React Router Future Flag Warning: Relative route resolution within Splat routes is changing in v7. You can use the `v7_relativeSplatPath` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_relativesplatpath. indexhtml-CdhAQHz9.js:51:6661
⚠️ React Router Future Flag Warning: The persistence behavior of fetchers is changing in v7. You can use the `v7_fetcherPersist` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_fetcherpersist. indexhtml-CdhAQHz9.js:51:6661
⚠️ React Router Future Flag Warning: Casing of `formMethod` fields is being normalized to uppercase in v7. You can use the `v7_normalizeFormMethod` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_normalizeformmethod. indexhtml-CdhAQHz9.js:51:6661
⚠️ React Router Future Flag Warning: `RouterProvider` hydration behavior is changing in v7. You can use the `v7_partialHydration` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_partialhydration. indexhtml-CdhAQHz9.js:51:6661
⚠️ React Router Future Flag Warning: The revalidation behavior after 4xx/5xx `action` responses is changing in v7. You can use the `v7_skipActionErrorRevalidation` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_skipactionerrorrevalidation.

Options:
- to opt-in these features, fix the behaviour that fails after that
- to opt-out from these features for now, this though will come back later.

### Expected behavior

No warnings

### Minimal reproducible example

Any Vaadin 24.6 and 24.5 application that uses React Router (by default).

### Versions

- Vaadin / Flow version: 24.6 pre and 24.5.

mshabarov:
[...]

PR #20537, opened 2024-11-22
chore: React Router v7

Artur-:


mshabarov:
We can try opt-in the feature flag immediately, i.e. https://github.com/vaadin/flow/issues/20527, what do you think? Flow team can take over and try how validation goes.

mshabarov:
And I propose to target this jump to Vaadin 24.7.

mshabarov:
> > We can try opt-in the feature flag immediately
> 
> Aren't all the flags gone in v7 and automatically on?

I'd expect that, yes. So we can remove opt-in codes I think.

PR #20540, opened 2024-11-25
fix: remove modal component on route refresh

### LINKED ISSUE #20473: Dialog is not updated when doing hotswap
### Description of the bug

Given

```
@Route(value = "overlay", layout = MainLayout.class)
@AnonymousAllowed
public class OverlayView extends VerticalLayout {

    @Override
        super.onAttach(attachEvent);

        Dialog dialog = new Dialog("The Dialog");
        dialog.setCloseOnOutsideClick(true);
        dialog.setCloseOnEsc(true);

        dialog.add(new Div("Text nr 1"));

        dialog.open();
    }

}
```

if you are running using HotswapAgent and edit the div text to "Text nr 2", then a dialog with "Text nr 1" remains on screen and a dialog with "Text nr 2" is not added at all. If you manually refresh the page, the dialog with "Text nr 2" is shown

### Expected behavior

The old dialog is closed and the new one opened

as above

### Versions

- Vaadin / Flow version: 24.5.3

mcollovati:
## Description

Modal components attached to the UI were not removed or replaced during self-navigation triggered by a route refresh.
This change updates navigation handler to ensure modal components are removed and adds a new navigation trigger for route refresh to differentiate programmatic navigations (e.g., forward actions).
It also modifies Hotswapper to require a full chain refresh when modal components are present.

Fixes #20473

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/hotswap/Hotswapper.java
+++ b/flow-server/src/main/java/com/vaadin/flow/hotswap/Hotswapper.java
@@ -349,13 +349,14 @@ public class Hotswapper implements ServiceDestroyListener, SessionInitListener,
                                 .isAssignableFrom(chainItem.getClass())))
                 .distinct().toList();
 
+        UIRefreshStrategy refreshStrategy;
         // A full chain refresh should be triggered if there are modal
         // components, since they could be attached to UI or parent layouts
-        boolean hasModalComponents = ui.hasModalComponent();
-        UIRefreshStrategy refreshStrategy;
-        if (!targetChainChangedItems.isEmpty()) {
-            refreshStrategy = targetChainChangedItems.stream().allMatch(
-                    chainItem -> chainItem == route && !hasModalComponents)
+        if (ui.hasModalComponent()) {
+            refreshStrategy = UIRefreshStrategy.PUSH_REFRESH_CHAIN;
+        } else if (!targetChainChangedItems.isEmpty()) {
+            refreshStrategy = targetChainChangedItems.stream()
+                    .allMatch(chainItem -> chainItem == route)
                             ? UIRefreshStrategy.PUSH_REFRESH_ROUTE
                             : UIRefreshStrategy.PUSH_REFRESH_CHAIN;
         } else {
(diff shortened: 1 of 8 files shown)


flow-server/src/main/java/com/vaadin/flow/hotswap/Hotswapper.java
```diff
@@ -349,10 +349,13 @@ private UIRefreshStrategy computeRefreshStrategy(UI ui,
...
 
+        // A full chain refresh should be triggered if there are modal
+        // components, since they could be attached to UI or parent layouts
+        boolean hasModalComponents = ui.hasModalComponent();
         UIRefreshStrategy refreshStrategy;
         if (!targetChainChangedItems.isEmpty()) {
-            refreshStrategy = targetChainChangedItems.stream()
-                    .allMatch(chainItem -> chainItem == route)
+            refreshStrategy = targetChainChangedItems.stream().allMatch(
+                    chainItem -> chainItem == route && !hasModalComponents)
```
tepi, line 358: Is this checking `hasModalComponents` for each `chainItem`? If I read this correctly the value of it will not change

mcollovati: `allMatch` should stop evaluating the predicate when the condition is true. But effectively, there's no need to iterate at all if there are modal components.
I'll fix it, thanks

mcollovati: done

tepi (approved):

netsrotr:
We have tested Vaadin 24.5.7 and this fix cause issues on our side with our hotswap -it close opened modal dialogs randomly and influence the browser refresh behavior (e.g. via key F5 in chrome).
It would really help if you would provide any documentation how to setup a working hotswap environment (favor eclipse IDE, but IntelliJ would also fit), where your fix really work. For now we stoped using hotswap because of that, any older version 24.5.x before was better than that, sorry.

mcollovati:
@netsrotr do you mind creating a new ticket, providing a code example (or even better, a sample project) showing how you are using dialogs in the application? This would help us investigate further and see if we missed some use cases.

PR #20547, opened 2024-11-25
feat: resend payload of reoccuring client request

### LINKED ISSUE #20506: Server to resend a response message that client waits for
When server receives the same message id and hash as latest handled server should resend the previous payload to the client.

- Store the latest handled payload until next id is handled
- Only resend payload if client id & hash is the latest handled message

This should help to avoid unnecessary resynchronisations and loggings like:
```
com.vaadin.flow.server.communication.ServerRpcHandler   | Ignoring old duplicate message from the client. Expected: 1748, got: 1747.
```

caalador:
If the client request contains
the previous id and exactly the
same message content respond with
the same payload as previously.

Closes #20506

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -215,6 +215,8 @@ public class UIInternals implements Serializable {
 
     private byte[] lastProcessedMessageHash = null;
 
+    private String lastRequestResponse = "for(;;);[{}]";
+
     private String contextRootRelativePath;
 
     private String appId;
@@ -303,6 +305,25 @@ public class UIInternals implements Serializable {
             byte[] lastProcessedMessageHash) {
         this.lastProcessedClientToServerId = lastProcessedClientToServerId;
         this.lastProcessedMessageHash = lastProcessedMessageHash;
+    }
+
+    /**
+     * Sets the response created for the last UIDL request.
+     *
+     * @param lastRequestResponse
+     *            The request that was sent for the last UIDL request.
+     */
+    public void setLastRequestResponse(String lastRequestResponse) {
+        this.lastRequestResponse = lastRequestResponse;
+    }
+
+    /**
+     * Returns the response created for the last UIDL request.
+     *
+     * @return The request that was sent for the last UIDL request.
+     */
+    public String getLastRequestResponse() {
+        return lastRequestResponse;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
```diff
@@ -215,6 +215,8 @@ public List<Object> getParameters() {
 
     private byte[] lastProcessedMessageHash = null;
 
+    private String lastRequestResponse = "for(;;);[{}]";
```
tepi, line 218: Is the initial value needed for this field?

flow-server/src/main/java/com/vaadin/flow/server/communication/ServerRpcHandler.java
```diff
@@ -236,6 +236,19 @@ public ResynchronizationRequiredException() {
         }
     }
 
+    /**
+     * Exception thrown then the client side re-sent the same request.
```
tepi, line 240: -> "`Exception thrown when the client side re-sends the same request`"

flow-server/src/main/java/com/vaadin/flow/server/communication/ServerRpcHandler.java
```diff
@@ -236,6 +236,19 @@ public ResynchronizationRequiredException() {
...
+     */
+    public static class ResendPayloadException extends RuntimeException {
```
tepi, line 242: `PayloadResentException` might be better name, or maybe `ClientResentPayloadException` to emphasize that it is the client doing the resending

mcollovati:
Tested with a custom buggy UIDLRequestHandler that seldom drops a response, returning a 410 response code.
The change proved to work very well. When a response for an UIDL message does not reach the client, then the connection status indicator appears for a moment, but then when the request is re-sent, the client receives the correct response and all proceeds correctly.

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
caalador: I would expect that not really. The thought was more that if something goes wrong the payload can be handled by the client.

flow-server/src/main/java/com/vaadin/flow/server/communication/ServerRpcHandler.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
caalador: removed

flow-server/src/main/java/com/vaadin/flow/server/communication/ServerRpcHandler.java
caalador: changed

PR #20551, opened 2024-11-26
fix: compute unique key for artifact

### LINKED ISSUE #20555: project failed with Duplicated key on 24.5.6 (flow 24.5.7)
### Describe the bug

After upgrade to Vaadin 24.5.6, project builds failed with 
```
Execution configure of goal com.vaadin:vaadin-maven-plugin:24.5.6:configure failed: Duplicate key org.yaml:snakeyaml (attempted merging values org.yaml:snakeyaml:jar:2.2:compile and org.yaml:snakeyaml:jar:android:1.23:compile)
```

```
Execution default of goal com.vaadin:flow-maven-plugin:24.5.7:prepare-frontend failed: Duplicate key javax.validation:validation-api (attempted merging values javax.validation:validation-api:jar:1.0.0.GA:compile and javax.validation:validation-api:jar:sources:1.0.0.GA:compile)
```

### Expected-behavior

project builds without failures. 

### Reproduction

take either the `ce-demo` or `mpr-demo`
upgrade the flow.version to 24.5.7 and vaadin.version to 24.5.6
run `mvn install -DskipTests -Pproduction`

### System Info

Vaadin version 24.5.6 and flow 24.5.7

mcollovati:
Computes a unique key for maven artifacts when collecting dependencies for the Flow maven plugin isolated classloader.

fixes #20555

### DIFF
--- a/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/Reflector.java
+++ b/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/Reflector.java
@@ -223,7 +223,10 @@ public final class Reflector {
         }
 
         Function<Artifact, String> keyMapper = artifact -> artifact.getGroupId()
-                + ":" + artifact.getArtifactId();
+                + ":" + artifact.getArtifactId() + ":" + artifact.getType()
+                + ((artifact.getClassifier() != null)
+                        ? ":" + artifact.getClassifier()
+                        : "");
 
         Map<String, Artifact> projectDependencies = new HashMap<>(project
                 .getArtifacts().stream()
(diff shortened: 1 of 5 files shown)


mshabarov (approved):
This patch should fix the reported issue, LGTM.

Just a question: should we also add `version` and `scope` to the artifact's key mapper?
I don't know the whole context of this feature, maybe the version is handled already in some other place.

mcollovati:
> Just a question: should we also add version and scope to the artifact's key mapper?

Key mapper is used to identify artifacts that are defined as dependencies for both project and Flow maven plugin. If there are duplicates, we choose the project dependency.
So, version and scope should not be considered; otherwise we will not be able to detect duplicates.
And in the same project, it is an error to define the same dependency twice with a different version or a different scope.

Also, Maven `DefaultArtifact.getId()` does not consider scope. It is composed by concatenating `getDependencyConflictId()` with `getBaseVersion()`, and `getDependencyConflictId()` is  basically `groupId:artifactId:type[:classifier]]`

Issue #20577, opened 2024-11-29
DetachListeners and onDetach hooks are not called when an exception is thrown during detach of contained components


mperktold:
### Description of the bug

This is related to #20293, but focuses on the `DetachEvent` rather then the `SessionDestroyEvent`.

When a component is removed from the view, the whole component subtree will be notified bottom-up about this `DetachEvent`. If one of the detach hooks or listeners of "lower" components throws an exception, the components higher up the tree will never get notified.

This can lead to resource leaks, because `onDetach` is a good place to release resources (which are typically aquired in `onAttach`). 

Currently, the only way to make sure a custom `onDetach` hook is always called is to check every implementation of `onDetach` of components further down the tree and make sure they don't throw, possibly using `try-catch`. And there might be third-party components which you don't control directly.

### Expected behavior

Ideally, there should be `try-finally`-like semantics for `DetachEvent`, meaning that the corresponding hooks and listeners should always be called, even if an exception is thrown from detach hooks further down the tree.

As I said in #20293, it shouldn't be the responsibility of these other components to make sure another component higher up the tree can clean up its resources. Instead, this should be the responsibility of that component itself, and Vaadin should provide the mechanisms to make that possible.

I am not sure what exactly should happen when an exception is thrown, though. I guess a good start is to wrap every invocation of a detach hook or listener in a `try-catch` block. In the catch block, one option would be to simply pass all exceptions to `VaadinSession.getErrorHandler()`. Another option would be to remember the first exception and to add all following exceptions as suppressed to it, so the first exception can be rethrown once the whole detach process has completed.

### Minimal reproducible example

```java
public class DetachTest extends VerticalLayout {

    public DetachTest() {
        var badDiv = new Div();
        badDiv.addDetachListener(e -> {
            throw new IllegalStateException("Test");
        });
        add(badDiv);
    }

    @Override
    protected void onDetach(DetachEvent detachEvent) {
        super.onDetach(detachEvent);
        System.out.println("super important cleanup logic");
        Notification.show("detached");
    }
}
```

1. Add this view to your project, with at least one other view
2. Open this view in your browser
3. Navigate to another view

`onDetach` will not print anything to the console, nor will it show the notification.

### Versions

- Vaadin / Flow version: 24.5.5
- Java version: Temurin 21.0.5+11
- OS version: Windows 11

Legioth:
Agreed. These types of listeners should be resilient so that application code can rely on those being called when appropriate even if other parts of the application misbehaves.

[...]

PR #20622, opened 2024-12-05
fix: service destroy listener exceptions

tepi:
Related to https://github.com/vaadin/flow/issues/20577


caalador (commented):
Looks ok, but there should perhaps be a test with multiple listeners that check that they get called even if one or more throw an exception.

tepi:
> Looks ok, but there should perhaps be a test with multiple listeners that check that they get called even if one or more throw an exception.

Good point, I'll add a test. Maybe a similar one for `SessionDestroyListener` as well since the logic is there already but looks like no test for that either.

caalador (approved):

PR #20656, opened 2024-12-09
fix: catch exceptions from detach calls

tepi:
Wraps detach listener calls and onDetach calls into try-catch blocks so all
possible exceptions are handled by session ErrorHandler, and they will not
interrupt running the remaining detach handlers.

Fixes https://github.com/vaadin/flow/issues/20577


caalador (commented):
ComponentTest.testAttachDetachListeners_parentAttachDetach_childListenersTriggered
could probably be used to create a test that collects the correct amount of events even if one of the listeners throws.

mcollovati (commented):
Looks good. The only concern is that this could be some sort of behavioral breaking change, since it doesn't break the application as before, but it "silently" continues the execution (although errors should be logged, unless a custom error handler suppresses them).

It would probably be good to highlight the change in the release notes.

caalador (approved):

PR #20669, opened 2024-12-11
chore!: Change supported Node from 18 to 20

Artur-:
Node 20 is the active LTS version for Node.

Node 18 will be end of life soon and new libraries like React 19 support Node 20+


mshabarov (approved):

mshabarov:
@PhilippWendler thanks for the comment!

About npm 9.0 I'm not sure. Node 20.0 was shipped with npm 9.6, 9.0 may work, but I'd stick with 9.6.
Here is the update for compatibility page https://github.com/vaadin/docs/pull/4257.

PR #20687, opened 2024-12-12
fix: ensure requestEnd clears Vaadin thread locals

mcollovati:
Makes sure that Vaadin thread locals are cleared even if something fails durung requestEnd execution. It also wraps Vaadin interceptors execution in a try/catch block to ensure all of them are invoked and that potential failures does not affect the continuation of requestEnd method.


mcollovati:
Automatic cherry-pick to 23.5 and 2.11 will most likely and require manual pick.

caalador (approved):

PR #20694, opened 2024-12-13
chore: Add a new maven module for Full-Stack Signals common lib

mshabarov:
Adds a new empty maven module for Full-Stack Signals common lib for Flow and Hilla

### DIFF
--- a/flow-signals/pom.xml
+++ b/flow-signals/pom.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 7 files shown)


flow-signals/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+    <properties>
+        <junit5.version>5.11.3</junit5.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-server</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
taefi, line 26: Maybe I'm under a wrong impression, but I thought it should not depend on anything Flow related, or am I missing a point?

taefi (changes requested):
Should this module depend on a flow module?

flow-signals/pom.xml
mshabarov: I was thinking that it can depend on flow-server as it has many base interfaces and signals may need them, but what does depend could be placed into flow-server directly, whereas this module should be flow core agnostic.

taefi (approved):
LGTM!

Issue #8163, opened 2020-04-27
Format flow code and investigate how to validate formatting of future code


joheriks:
There are Maven goals `formatter:format` and `formatter:validate` in place using the Flow conventions (https://github.com/vaadin/flow/blob/master/eclipse/VaadinJavaConventions.xml), but the code is not up to date with these. This causes unwanted formatting changes in PRs when the IDE to formats a source files (alternatively the developer has to take care to format to only modified code).

### To do:
- [ ] All existing flow code has been formatted according to VaadinJavaConventions, and manually checked that any potentially destroyed chunks are protected from auto-formatting.
- [ ] Investigate ways to prevent unformatted code from entering the repository and/or cluttering PRs (for instance validating in a Git pre-commit hook or on PR).

denis-anisimov:
As we agreed: before making this as a validation rule we should check whether the automatic formatting in Eclipse still causes formatting changes after the code format is applied.

In fact I totally disagree to include a validation rule for formatting.
It totally destroys any development: I know this from my experience that most of time is spent on formatting instead of real coding.

Formatting may not be a validation rule.
Either we are able to reduce formatting changes via making sure that all formatters we are using in IDEs matches each other or this should work automatically without any involvement if formatting is absolutely necessary: validation should have a task to update the PR applying formatting automatically.

I may do only one thing: 
-write a code
- care about formatting.
I personally can't do both.

pleku:
I'd like to point out again that the time is not only spent on writing the code, but also in review. And that time is also important to be spent efficiently and not having to care about looking on unrelated formatting changes.

I think the purpose is to get to a phase where nobody has to care about formatting related changes when either writing code or when doing review. To achieve this, I think what we need is
- all existing code is formatted according to the formatting
- instructions how to configure Eclipse & IDEA on how to apply exactly the same formatting automatically
- a way to format code outside the two IDEs (like with a maven plugin)

The validation should be fine for everyone as long as they use Eclipse or IDEA with the correct settings. It doesn't change anything. It only changes things for those people who have not setup their IDEs and try to push unformatted changes.

denis-anisimov:
In that way I agree.

So if we are able to :
- provide instructions how to configure Eclipse and IDEA so that formatting is done automatically (and those instructions really works).
- a way to do formatting without any IDE

then I'm OK with this approach.

My point is : to be able to require formatting we have to define the rules exactly and make it easy to use.
If we are not able to do this then we may not require this.

About the current situation :
[...]

PR #20695, opened 2024-12-13
feat: Maven plugin improvements

### LINKED ISSUE #19596: Speed up vaadin:prepare-frontend and vaadin:build-frontend goals in the Maven plugin
### Describe your motivation

I noticed that ``vaadin:prepare-frontend`` and ``vaadin:build-frontend`` are relatively slow - a rerunning production build takes around 15s in one of our projects that I recently updated to 24.4 - even when they have to do nearly nothing, as everything was already generated before.

I had a look at the log messages and the code an could narrow down the following places where performance can be optimized:
| Problem | Possible fix |
| --- | --- |
| In 24.4 a new goal ``configure`` is invoked when running ``prepare-frontend`` or ``build-frontend``, this goal is only required when you use hilla (for example we do not) | Make the goal optional or at least provide a parameter to control if it should be invoked |
| It's always checked if Node/NPM versions are correct - this takes around 2-3s on my machine | Add a parameter which makes it possible to disable this |
| In ``prepare-frontend`` or ``build-frontend`` all classpathElements are scanned for classes (see ``getClassFinder``). This also includes libraries which have nothing to do with Vaadin at all, e.g. Hibernate, Spring Boot, etc | Provide an option to ignore these libraries/jar from scanning |
| The scan is done multiple times (counted at least 3 for ``build-frontend``) | Cache the scan for the mojo |
| Inherited lookup is nearly always performed: When e.g. looking up classes that use  ``@NpmPackage`` also all classes that extend from classes which have the annotation are looked up (→ ``ReflectionsClassFinder``) | Do not search for extending classes in all lookups |

For now we forked the plugin inside our repository and applied some fixes manually.

For example:
```java
// Workaround to ignore Node/NPM checks
System.setProperty(FrontendUtils.PARAM_IGNORE_VERSION_CHECKS, "true");
```

```java
// Workaround to invoke ClassFinder only once per goal
protected ClassFinder cached;

@Override
public ClassFinder getClassFinder()
{
	if(this.cached == null)
	{
		this.cached = ... // createClassFinder
	}
	return this.cached;
}
```

```java
// Filtering only relevant classpathElements
final List<String> classpathElements = FlowModeAbstractMojo.getClasspathElements(project);
final List<String> filtered = classpathElements.stream()
	.filter(s -> {
		if(!s.endsWith(".jar") // The current workdir is not a JAR
			|| s.contains("target") // If a jar is inside a target folder, it's likely a sibling project (multi-module)
			|| s.contains("vaadin") // Vaadin projects are always needed
		)
		{
			return true;
		}
		... // Additional user configure conditions go here
		return false;
	})
	.toList();
return BuildFrontendUtil.getClassFinder(filtered);
```

The above changes improve the build speed (in our case) from ~15s down to 3s when rerunning the production build.

### LINKED ISSUE #20359: Consider removing ``com.google.gwt:gwt-elemental`` from ``flow-maven-plugin``
### Describe your motivation

The ``flow-maven-plugin`` contains  ``com.google.gwt:gwt-elemental`` here:
https://github.com/vaadin/flow/blob/f1c9673ae2b9f94678144f4918145001c153baf8/flow-plugins/flow-maven-plugin/pom.xml#L50-L59

However it doesn't seem to be used anywhere and can probably be removed.

Remove the dependency if it's never used.

AB-xdev:
aka a (bit early) Christmas present from XDEV to Vaadin 🎄

## Description

This PR reworks the Vaadin Maven Plugin to improve performance.
The changes primarily target the ``prepare-`` and ``build-frontend`` goals (all other goals are irrelevant for us) and are based on some experimental changes we have done in internal projects.

Fixes #19596
Fixes #20359

Overview of the changes:
* A core problem of the plugin has always been that it scans/loads all classes of the entire project. This might be ok for a small Vaadin starter but if you have a big project with a few hundred dependencies this takes forever.<br/>If you combine this situation with frequent (re)builds due to e.g. not a good working IDEs, this results in further problems like increased coffee consumption in between builds.<br/>The improvement here is to only scan Vaadin dependencies by default.<br/>The user can further customize and include/exclude certain dependencies if they want.
* A lot of additional optimizations have been introduced, that enable the user to disable additional checks which might not be required.
* Made dependency onto ``org.codehaus.plexus.build.BuildContext`` optional (not everyone uses Eclipse ;) )
* Most things in the plugin can now be replaced/overridden if required

<details><summary>Example configuration</summary>

```xml
<configuration>
	<!-- We don't require React -->
	<reactEnable>false</reactEnable>
	<!-- Only needed when updating -->
	<frontendIgnoreVersionChecks>true</frontendIgnoreVersionChecks>
	<!-- We don't use Hilla -->
	<hillaAvailable>false</hillaAvailable>
	<!-- We know how to use a BOM -->
	<checkPluginFlowCompatibility>false</checkPluginFlowCompatibility>
	<!-- We also don't use that -->
	<supportDAU>false</supportDAU>
	<fastReflectorIsolation>
		<includeFromTargetDirectory>false</includeFromTargetDirectory>
		<includes>
			<additional>
				<selector>
					<groupId>...</groupId>
					<artifactId>web...*</artifactId>
				</selector>
				<!-- Required because of interface usage -->
				<selector>
					<groupId>...</groupId>
					<artifactId>logger-...</artifactId>
					<scan>false</scan>
				</selector>
			</additional>
		</includes>
	</fastReflectorIsolation>
	<!-- Build-Frontend specific -->
	<!-- This project contains no things that require a license during build -->
	<performLicenseCheck>false</performLicenseCheck>
	<!-- We know how to use a BOM -->
	<checkRuntimeDependency>false</checkRuntimeDependency>
</configuration>
```

</details>

In a example project (with ~40k LoC and ~150 dependencies) with above configuration we now have the following times:
| | ``prepare-frontend`` | ``build-frontend`` |
| --- | --- | --- |
| Vaadin Default | ~1600ms | ~5100ms |
| with improvements (this PR) | ~900ms (~75% faster) | ~2000ms (~250% faster) |

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have added tests to ensure my change is effective and works as intended.<br/>
_Fixed existing tests so that they work. Feel free to add more tests if you like._
- [x] New and existing tests are passing locally with my change.<br/>
_There seem to be some tests that don't work locally, e.g.:
``[main] ERROR com.vaadin.flow.server.frontend.FrontendToolsLocator - Failed to execute the command '[C:\Program Files\maven\apache-maven-3.9.8\bin\mvn, -v]'
java.io.IOException: Cannot run program "C:\Program Files\maven\apache-maven-3.9.8\bin\mvn": CreateProcess error=193, %1 ist keine zulässige Win32-Anwendung`` or ``DefaultInstantiatorI18NTest.translationFileOnClasspath_instantiateDefaultI18N:122 expected:<[Default lang]> but was:<[deutsch]>``_

PS:
* If you see Matti, please give him our best regards
* Would be great if the Formatter & Co is checked into the IntelliJ IDE settings (yes that's possible - have a look at [our repos](https://github.com/xdev-software/java-template/tree/master/.idea)) so you/I don't have to set it up manually or correct it constantly ;)


mstahv:
Awesome @AB-xdev 👏 Does this also fix performance regressions in the development mode 🤔

AB-xdev:
> Awesome @AB-xdev 👏 Does this also fix performance regressions in the development mode 🤔

@mstahv We only changed the Maven Plugin and nothing noteworthy of the ``flow-server`` or similar.
If you use the plugin during "development mode" to e.g. re-generate the bundle then maybe yes (I'm not quite sure what you guys count to "development mode" and what not).

mcollovati:
@AB-xdev thank you so much for the contribution. The PR contains many changes, so it will take some time to review it.

A couple of questions and notes:

* what's the reason to make all fields protected?
* why remove the `logTroubleshootingHints` method?
* the original solution included only a subset of provided dependencies; however, with this change, the filter is gone. I need to check the reasons behind the filter and see if they are still valid.
* all public classes and methods required Javadoc

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/Reflector.java
mcollovati, line 1: I wonder if we need this interface. I do not expect custom implementations to be provided.
Do you have any good use cases that would need a custom implementation?

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/ReflectorController.java
mcollovati, line 1: As for `Reflector`, I wonder if an interface is needed. Wouldn't the default implementation be enough?

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/ReflectorController.java
```diff
@@ -0,0 +1,24 @@
...
+    String getReflectorClassIdentifier();
+
+    /**
+     * Tries to reuse/adapt the given reflector.
+     *
+     * @throws RuntimeException Might be thrown if reuse fails.
+     */
+    Reflector adaptFrom(Object reflector);
+
+    Reflector of(MavenProject project, MojoExecution mojoExecution);
```
mcollovati, line 23: Javadoc is missing. Also, could we have a better method name?
I would approve `of` if the return type would be `ReflectorController`, but the method instead returns a `Reflector`. Something like `newReflector`, `createReflector` or `buildReflector`.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/ReflectorController.java
```diff
@@ -0,0 +1,24 @@
+package com.vaadin.flow.plugin.maven;
+
+import org.apache.maven.plugin.MojoExecution;
+import org.apache.maven.project.MavenProject;
+
+
+/**
+ * "Builder" for {@link Reflector}
+ */
+public interface ReflectorController {
```
mcollovati, line 10: nit: I wonder if we could come up with a better name. None of the methods make me think of this class as a **controller**. It looks like to me some sort of **provider**. Its responsibility seems to be to provide a new or adapted Reflector instance.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/ReflectorIsolatedClassLoader.java
mcollovati, line 1: Can you explain the reason for introducing this abstract class?

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/Fragment.java
mcollovati, line 1: This class seems unused, but being public, we cannot remove it. We have to deprecate it so that removal can happen in the next major.

mcollovati (changes requested):
I did a first review round and added some comments.
I still need to test the changes locally because I currently see several test failures.
I'll continue the review as soon as I get the validation pass locally.

mcollovati:
I #20776, rebased on main, to test the changes.
After a first run, it looks like that the change is potentially breaking existing projects, most likely because it filters out too many dependencies by default.

> Error:  Failed to execute goal com.vaadin:flow-maven-plugin:999.99-SNAPSHOT:prepare-frontend (default) on project test-prod-bundle: Execution default of goal com.vaadin:flow-maven-plugin:999.99-SNAPSHOT:prepare-frontend failed: A required class was missing while executing com.vaadin:flow-maven-plugin:999.99-SNAPSHOT:prepare-frontend: org/codehaus/plexus/build/BuildContext


I'll continue the investigation, pushing changes on the linked PR.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/DefaultReflectorController.java
```diff
@@ -0,0 +1,411 @@
...
+                        artifact);
+        if (includeSelector.isPresent()) {
+            final FastReflectorIsolationConfig.ArtifactSelector selector = includeSelector.get();
+            logArtifactInclusionOrExclusion(artifact, true, "in includes [" + selector + "]");
+            return new ArtifactWrapper(artifact, selector.isScan());
+        }
+
+        // If a jar is inside a target folder, it's likely a sibling project
+        if (fastReflectorIsolationConfig.isIncludeFromTargetDirectory()
+                && "target".equals(artifact.getFile().getParentFile().getName())) {
```
mcollovati, line 246: Output directory can be configured in maven build, so it would be better to avoid hard-coding "target" and get the value from the Maven project.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/PrepareFrontendMojo.java
```diff
@@ -24,6 +25,10 @@
 import org.apache.maven.plugins.annotations.ResolutionScope;
 
 import com.vaadin.flow.plugin.base.BuildFrontendUtil;
+import org.codehaus.plexus.build.BuildContext;
```
mcollovati, line 28: This seems to be a wrong import. In `FlowModeAbstractMojo` it was `org.codehaus.plexus.build.BuildContext`

mcollovati:
Also note that Hilla plugin recently removed the need for executing `configure` goal (vaadin/hilla#2616). That change should also improve the overall performance of `vaadin-maven-plugin`.

AB-xdev:
Wow, that's a lot of comments. I will try to answer them all.

> what's the reason to make all fields protected?

So it's possible that one can override them without getting having to use reflection or copying code ^^
This also includes methods, classes, etc.

> why remove the logTroubleshootingHints method?

1. The problems are reported [directly by ``DefaultReflectorController``](https://github.com/vaadin/flow/blob/43183dac03f5468d5c0a6eafc20558f7ca2dfc62/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/DefaultReflectorController.java#L167-L184) this way there is no need to store them as a variable
2. The other messages ("The build process encountered an error: .." and "To diagnose the issue, please re-run Maven with the -X ...") are nearly the same as Maven itself reports when the build fails. There is no need to duplicate them.

> what's the benefit of the supportDAU flag? Is it only to skip the application identifier hashing?

Correct

> the original solution included only a subset of provided dependencies; however, with this change, the filter is gone. I need to check the reasons behind the filter and see if they are still valid.

Yes, the original only contained  ``portlet-api`` and ``javax.servlet-api``.
I removed it because 
* the code looked [very ancient](https://github.com/vaadin/flow/commit/015d6664d5278d406dacec3eb772bd43956cf4f5#diff-04a23c32a5945d5aeec9577577b8184b016651195605ebaa756bd758749ad379R40) 
   * There seems to have been also [a ``VaadinPortlet`` in the past](https://github.com/vaadin/flow/pull/6769).
* there was no documentation why it's still there 
* Vaadin nowadays requires ``jakarta`` (which was not present in the regex)

> all public classes and methods required Javadoc

I didn't know that.

_WIP_

flow-server/src/main/java/com/vaadin/flow/internal/ReflectTools.java
```diff
@@ -68,31 +68,6 @@ public class ReflectTools implements Serializable {
...
-     *            The name of the method
-     * @param parameterTypes
-     *            The parameter types for the method.
-     * @return A reference to the method
-     * @throws ExceptionInInitializerError
-     *             Wraps any exception in an {@link ExceptionInInitializerError}
-     *             so this method can be called from a static initializer.
-     */
-    public static Method findMethod(Class<?> cls, String methodName,
-            Class<?>... parameterTypes) throws ExceptionInInitializerError {
```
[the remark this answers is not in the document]
AB-xdev: > Why is this public method removed?

It was only used in the plugin and is also marked as ``internal`` (in package)

> If it is not used anywhere, we can consider deprecating it, but it cannot be removed.

We can change it back if you like

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
```diff
@@ -240,46 +252,49 @@ public boolean compressBundle() {
...
+            getLog().info("Ignoring runtime dependency check");
+            return true;
+        }
+
         Objects.requireNonNull(groupId, "groupId cannot be null");
         Objects.requireNonNull(artifactId, "artifactId cannot be null");
-        if (missingDependencyMessage == null) {
-            missingDependencyMessage = text -> {
-            };
-        }
```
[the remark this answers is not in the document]
AB-xdev: I used an ``Optional`` further down in the ``if``s as I found it more readable and it makes more sense there (the previously created consumer may have never been used in the ``if``s below).

As an alternative one can also do a null check, like:
```java
if(...) {
  if (missingDependencyMessage != null) { // Potential replacement for Optional
    missingDependencyMessage.accept(...);
  }
}
```

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/DefaultReflector.java
```diff
@@ -0,0 +1,143 @@
...
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.Set;
+
+
+public class DefaultReflector implements Reflector {
```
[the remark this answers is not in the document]
AB-xdev: > Can we make the class final?

What's the advantage of this?
Currently one can override the class if required. If it's ``final`` that's no longer possible.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/DefaultReflector.java
```diff
@@ -0,0 +1,143 @@
...
+    protected final ReflectorIsolatedClassLoader isolatedClassLoader;
+    protected Object classFinder;
+
+    public DefaultReflector(final ReflectorIsolatedClassLoader isolatedClassLoader) {
+        this.isolatedClassLoader = Objects.requireNonNull(isolatedClassLoader);
+    }
+
+    public DefaultReflector(final Object copyFromOtherClassLoader) {
```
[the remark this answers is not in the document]
AB-xdev: > I would make this constructor package-protected, as it looks like to me, it should be used only internally.

Would also keep it public so one can override/instantiate it if required.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/DefaultReflector.java
```diff
@@ -0,0 +1,143 @@
...
+    protected synchronized void initClassFinder() throws ReflectiveOperationException {
+        if (classFinder == null) {
+            final Class<?> classFinderImplClass = getIsolatedClassLoader().loadClass(
+                    ReflectionsClassFinder.class.getName());
+            classFinder = classFinderImplClass
+                    .getConstructor(ClassLoader.class, URL[].class)
+                    .newInstance(
+                            isolatedClassLoader,
+                            isolatedClassLoader.urlsToScan());
+        }
```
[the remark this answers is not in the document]
AB-xdev: +1

Overlooked that, it should probably use the method too like on all other occurrence.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildFrontendMojo.java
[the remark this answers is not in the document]
mcollovati: To be honest, I would prefer the null check in this case, since there is no pipeline built on top of the `Optional`.

flow-server/src/main/java/com/vaadin/flow/internal/ReflectTools.java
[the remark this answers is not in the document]
mcollovati: Even if it is `internal`, usually we try not to remove public API in minor or releases.
So, yes, please bring the method back and mark it as `@Deprecated(forRemoval)`. Also, the Javadoc `@deprecation` should state that there will be no replacement.

mcollovati:
An additional note about the changes in visibility: these changes are significantly increasing the API surface that we would be promising to keep stable between versions. Before committing to something like that, we would want to better understand what potential benefits you expect to get from that.


About the provided scoped dependencies:

> Yes, the original only contained portlet-api and javax.servlet-api.
> I removed it because
>
> the code looked [very ancient](https://github.com/vaadin/flow/commit/015d6664d5278d406dacec3eb772bd43956cf4f5#diff-04a23c32a5945d5aeec9577577b8184b016651195605ebaa756bd758749ad379R40)
> There seems to have been also https://github.com/vaadin/flow/pull/6769.
there was no documentation why it's still there
> Vaadin nowadays requires jakarta (which was not present in the regex)

My concern is that this modification might add more dependencies to the class loader, potentially leading to behavior changes compared to the original code and introducing a breaking change. However, it could make sense to consider also provided dependency. Needs to be investigated.

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/DefaultReflector.java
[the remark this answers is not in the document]
AB-xdev: Fixed

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/ReflectorController.java
AB-xdev: Done

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/Fragment.java
AB-xdev: Restored and marked as deprecated

flow-server/src/main/java/com/vaadin/flow/internal/ReflectTools.java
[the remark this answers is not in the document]
AB-xdev: Restored and marked as deprecated

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/DefaultReflectorController.java
AB-xdev: Done

flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/DefaultReflectorController.java
```diff
@@ -0,0 +1,411 @@
...
+        if (selector.endsWith("*") && selector.startsWith("*")) {
+            return target.contains(selector.substring(1, selector.length() - 1));
+        } else if (selector.endsWith("*")) {
+            return target.startsWith(selector.substring(0, selector.length() - 1));
+        } else if (selector.startsWith("*")) {
+            return target.endsWith(selector.substring(1));
+        } else {
+            return selector.equals(target);
+        }
+    }
```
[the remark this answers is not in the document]
AB-xdev: I would keep this here since the Config class is just a data holder which contains no logic.

AB-xdev:
Thank you for the additional feedback.
I think I responded now to all/most topics.

> An additional note about the changes in visibility: these changes are significantly increasing the API surface that we would be promising to keep stable between versions. Before committing to something like that, we would want to better understand what potential benefits you expect to get from that.

Generally speaking I want to be able to overwrite/customize your code without constantly having to copy it when an update happens because some method is private or something like that.

Since I had this problem now [for the like 50th time, I decided to write a short blog post about it](https://ab-xdev.github.io/blog/blog/2025/01/13/why-java-libraries-should-be-customizable-extensible.html).

> these changes are significantly increasing the API surface that we would be promising to keep stable between versions

I was not aware that you have to keep the plugin "stable" (you know since the average Vaadin user will likely never extend it).

---

> My concern is that this modification might add more dependencies to the class loader, potentially leading to behavior changes compared to the original code and introducing a breaking change. However, it could make sense to consider also provided dependency. Needs to be investigated.

It should make sense to include provided dependencies as the class-loading will otherwise very likely fail (e.g. if a class extends from another one that is inside a provided dependency).
In case of emergency we can also provide an option to exclude / include provided dependencies :)

---

> Currently, the changes in the PR will break projects using ``vaadin-maven-plugin`` because of the removal of static methods used by ``hilla-maven-plugin``.

My opinion here is that the hilla Plugin can implement these itself since only hilla needs them.

>Even re-adding the missing methods, Hilla projects seems not to work out of the box: they require manual configuration of the plugin to include dependencies that are excluded by this change (e.g. swagger, jackson, classgraph).

And now my extensibility effort from above pays off: Hilla can simply use a derivative of ``DefaultReflectorProvider`` where the default excludes/includes are slightly different and work for them.

> Given that existing projects should not require any configuration update if this one gets merged, I would make the dependency filtering an opt-in and not active by default.

The new system is currently designed in a way that it's compatible to the old system and only some - very special - projects might break.
I think if the changes are also ported over into Hilla there should be nothing speaking against this being enabled out of the box.
Additionally a troubleshooting guide can also be provided, which instructs users to disable the ``fastReflector`` on errors.

Anyway I would recommended to bundle these changes only with the next minor/major release and not inside a patch release.

mshabarov:
> Would be great if the Formatter & Co is checked into the IntelliJ IDE settings

Yes, I agree with this concern. We have this rather old ticket, but still relevant nowadays https://github.com/vaadin/flow/issues/8163. In short, would be nice to include an automatic formatting step into all pull requests automation.

mshabarov:
@AB-xdev this pull request is quite large and combines several distinct changes, making it challenging to review effectively. To streamline the process and ensure clarity, I propose splitting it into separate pull requests as follows:
1. Remove `gwt-elemental`
This is an isolated change and can be handled in its own pull request for simplicity and a quicker review.
2. Add `frontendIgnoreVersionChecks`
This can also be split into a standalone pull request, as it is independent and has minimal dependencies.
3. Add `fastReflectorIsolation`
This is the most complex part of the pull request and should be further divided into smaller, logical units:
    - Core implementation of `fastReflectorIsolation`.
    - Related configuration or refactoring changes.

**Access Modifiers (Private → Protected)**
The PR makes significant relaxations to access modifiers, which we need to avoid to prevent an overly open API. Here’s why:
The Flow Maven Plugin is extended by the Hilla Maven Plugin, and the Vaadin Maven Plugin is the public, recommended method for building Vaadin applications.
Opening access to extend every single plugin parameter introduces potential risks for plugin compatibility and increases maintenance complexity (e.g., for hybrid Vaadin applications with custom Flow plugins).
Instead of relaxing access, we should:
- Share additional configuration via plugin parameters (e.g., for version check ignoring or granular customisation for reflector/class scanning).
- Avoid extensions of Java Mojo classes for the reasons stated above - we don’t want to courage doing this because of potential problems of compatibility and maintainability.

**License-Related Parameters**
Some parameters proposed in this PR, like `ignoreLicenseChecker` and `ignoreDAU`, are not acceptable, even if they improve performance. We cannot allow Vaadin users to decide when a license is required or not. Instead, we should focus on optimizing the license-checking process to address performance concerns.

**checkRuntimeDependency**
This is a critical part of the license-checking process. This is actually done only for builds with activated Daily Active Users functionality (building with a subscription key).  Allowing it to be disabled could result in broken production artifacts, so we must retain this flag as mandatory.

**hillaAvailable**
We need to investigate whether it is safe and performant to expose this flag.

**checkPluginFlowCompatibility**
This flag disables Flow compatibility checks between the project and the plugin. Since the warnings generated by this check are vital (as the cause may result in runtime issues), we should not allow disabling this check.

**@NpmPackage Inherited Lookup**
Finally, one additional item that deserves its own investigation is the behavior mentioned in point 5:
“Inherited lookup is nearly always performed: When looking up classes with @NpmPackage, all classes extending those with the annotation are also looked up (via ReflectionsClassFinder).”
This should be explored as a separate effort (by Flow team) to determine the impact and feasibility of any changes.

**Proposed Next Steps**
To proceed, I suggest:
- Creating three separate pull requests to extract the above parts from this pull request. (Vaadin Flow team will do it)
- Investigating the @NpmPackage inherited lookup issue independently. (Vaadin Flow team).
- Investigating the license-checking process to address performance concerns. (Vaadin Flow team)
- Investigating the option of adding `hillaAvailable` (Vaadin Flow team).

mshabarov:
Closed in favour of https://github.com/vaadin/flow/pull/21125.

PR #20704, opened 2024-12-13
test: set Jetty stop timeout (2.11)

mcollovati:
Set the Jetty stop timeout to a value less than the Jetty plugin's stop wait time. This change prevents the ShutdownManager from being blocked while waiting for Jetty to stop, avoiding an inconsistent state when subsequent Maven modules are executed.


mcollovati:
Fixes PR validation

caalador (approved):

Issue #11026, opened 2021-05-18
Exception in push connection when logging out


Hayo-Baan:
### Description of the bug / feature
Since switching to the default Push transport (#10931) we always get an “Exception in push connection” when logging out.

Note: the exception is not thrown when simply closing the browser window…

### Minimal reproducible example
Our application is a pretty basic view/form application. Login/authentication is handled by a remote service using REST calls (`KioskAuthenticationProvider`). Logout is via a “logout” anchor in the top bar of the layout.

Our security configuration  is as follows:
```
@RequiredArgsConstructor
@EnableWebSecurity
@Configuration
public class SecurityConfiguration extends WebSecurityConfigurerAdapter {

    private static final String LOGIN_PROCESSING_URL = "/login";
    private static final String LOGIN_FAILURE_URL = "/login?error";
    private static final String LOGIN_URL = "/login";
    private static final String LOGOUT_SUCCESS_URL = "/login";

    private final LoginService loginService;

    /**
     * Tests if the request is an internal framework request. The test consists of
     * checking if the request parameter is present and if its value is consistent
     * with any of the request types know.
     *
     * @param request {@link HttpServletRequest}
     * @return true if the request is an internal framework request. False otherwise.
     */
    public static boolean isFrameworkInternalRequest(HttpServletRequest request) {
        final String parameterValue = request.getParameter(ApplicationConstants.REQUEST_TYPE_PARAMETER);
        return parameterValue != null &&
               Stream.of(HandlerHelper.RequestType.values()).anyMatch(r -> r.getIdentifier().equals(parameterValue));
    }

    @Bean
    public AuthenticationProvider authenticationProvider() {
        return new KioskAuthenticationProvider(loginService);
    }

    @Override
    protected void configure(AuthenticationManagerBuilder auth) {
        auth.authenticationProvider(authenticationProvider());
    }

    /**
     * Require login to access internal pages and configure login form.
     */
    @Override
    protected void configure(HttpSecurity http) throws Exception {
        // Not using Spring CSRF here to be able to use plain HTML for the login page
        http.csrf()
            .disable()

            // Register our CustomRequestCache, that saves unauthorized access attempts, so
            // the user is redirected after login.
            .requestCache()
            .requestCache(new SimpleRequestCache())

            // Restrict access to our application.
            .and()
            .authorizeRequests()

            // Allow health checks
            .antMatchers("/actuator/health")
            .permitAll()

            // Allow all Vaadin internal requests.
[...]

PR #20709, opened 2024-12-13
fix: fix role checking when using websocket push (#20679) (CP: 24.6)

mcollovati:
When using PUSH with websocket transport, the atmosphere wrapped request can be a no-op implementation whose isUserInRole method alwasy returns false, causing, for example, wrong access checking during navigation. This change falls back to Spring Securty for role checking when PUSH transport is websocket.
It also fixes some tests in order to propagate the Spring Security context when starting Thread that perform UI operations.

References psi#123
Part of #11026

### DIFF
--- a/flow-tests/vaadin-spring-tests/test-spring-security-flow/src/test/java/com/vaadin/flow/spring/flowsecurity/AppViewIT.java
+++ b/flow-tests/vaadin-spring-tests/test-spring-security-flow/src/test/java/com/vaadin/flow/spring/flowsecurity/AppViewIT.java
@@ -12,6 +12,7 @@ import java.util.stream.Collectors;
 import org.apache.commons.io.IOUtils;
 import org.junit.Assert;
 import org.junit.Test;
+import org.openqa.selenium.By;
 import org.openqa.selenium.WebDriver;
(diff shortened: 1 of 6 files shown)


flow-tests/vaadin-spring-tests/test-spring-security-flow/src/test/java/com/vaadin/flow/spring/flowsecurity/AppViewIT.java
```diff
@@ -381,4 +387,37 @@ private List<MenuItem> getMenuItems() {
...
+                intervalId = window.setInterval(function() {
+                  var inProgress = upload.files.filter(function(file) { return file.uploading;}).length >0;
+                  if (!inProgress) {
+                    window.clearInterval(intervalId);
+                    callback();
+                  }
+                }, 500);
+                """;
+        getCommandExecutor().getDriver().executeAsyncScript(script, element);
+    }
```
caalador, line 409: Is this test change needed for the cherry pick or should it be its own pick for the upload test fix from #20642?

caalador (approved):

PR #20729, opened 2024-12-16
fix: re-build production bundle if index.html changes

### LINKED ISSUE #20629: Application Shell Template does not work: index.html does not get used in production
### Description of the bug

The index.html file does not get used in production mode. I added a `<script>` to the head and in dev mode it works fine. When using `mvn package -Pproduction` the index.html does not contain any of my changes. In fact the default template is used.

### Expected behavior

The /frontend/index.html is used in production mode.

1. Download a new default starter with an empty Flow page from start.vaadin.com
2. Change the index.html, remove all comments & add script
3. Execute `mvn package -Pproduction`
4. Execute the jar
5. Incpect the website in browser. No script is there and all comments are back (like `<!-- This outlet div is where the views are rendered -->`)

### Versions

- Vaadin / Flow version: 24.5.7
- Java version: 17 & 21
- OS version: Win 10
- Browser version (if applicable): Chrome, Firefox
- Application Server (if applicable): starter default
- IDE (if applicable): VS Code

mcollovati:
Stores index.html hash in stats.json and forces production bundle to be re-built if file contents have changed.
Changes to index.html do not trigger a dev bundle re-generation since in dev mode the file is served directly from the frontend folder.

Fixes #20629


mcollovati:
If this change gets merged, we need to check and potentially update the index.html file in starters to match the default contents, to prevent a custom production bundle to be generated if not needed.

For example, `skeleton-starter-flow-spring` contains a version of `index.html` whose only difference with default is a removal of a comment.

https://github.com/vaadin/skeleton-starter-flow-spring/blob/v24/src/main/frontend/index.html

caalador (approved):

mcollovati:
@caalador I was thinking if we should remove indentation and comments before hashing, to prevent irrelevant changes causing a bundle rebuild. What do you think?

caalador:
> @caalador I was thinking if we should remove indentation and comments before hashing, to prevent irrelevant changes causing a bundle rebuild. What do you think?

removing comments might be a bit too much for running the check and hash creation. trimming might be fine, but it will make it more complex than now when we only read and hash.
Perhaps keep it in mind and add it if someone has a problem big enough that they all the time meet the issue with whitespace or comment updates.

PR #20743, opened 2024-12-18
feat: Add interfaces for CRUD services

Artur-:
These were previously in Hilla but are equally useful in Flow applications

PR #20745, opened 2024-12-18
chore: Register JpaFilterConverter in a way that can be overridden

Artur-:
Fixes https://github.com/vaadin/hilla/issues/2569


vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
```diff
@@ -142,4 +144,10 @@ public ServerEndpointExporter websocketEndpointDeployer() {
         return new VaadinWebsocketEndpointExporter();
     }
 
+    @Bean
+    @ConditionalOnMissingBean
+    public JpaFilterConverter jpaFilterConverter(EntityManager entityManager) {
```
knoobie, line 149: I would not want this Bean as flow user. Additionally it might be problematic for projects without JPA because of the EntityManager (so conditional on missing bean is not enough)

knoobie: And wait.. is this the code that made "Hilla utterly broken with multiple Data sources"? I'm strict against this in flow!

Artur-: Sounds like this is something personal. Can we discuss the merits of the code instead and what it is useful for

Artur-: As you can see from the description, it is supposed to fix https://github.com/vaadin/hilla/issues/2569, not introduce it to all Flow users

knoobie: It's added to "SpringBootAutoConfiguration" which effects flow.

Artur-: Indeed, that is the point of this PR. To add the features to Flow and not only restrict them to Hilla

knoobie: Your last two comments are kinda conflicting? 

But nevertheless back to code - this JPAFilter probably has to be added in a second AutoConfiguration class to ensure it's only loaded with JPA and Spring Data in use.. or does it have to be a bean at all? It looks just like some helper methods that need access to the EntityManager.. with the "move" to flow and re-creating of all classes.. it's binary incompatible anyway so perfect timing for breaking changes. 

Just a quick note on the other code: 
- doing lowercasr on a database column by default might be a performance problem with big tables without proper indexes 
- the string concatenation for like looks un-sanitized

Artur-: Let's see first where #20743 ends up. The original version in Hilla contained quite a bit of magic and this PR now contains way less of it

Artur-: > Your last two comments are kinda conflicting?

The point is to introduce _the features_ to all Flow users but not to introduce _the problem_ to all Flow users

PR #20770, opened 2024-12-20
fix(flow-react): avoid Flow-React portal outlet removal conflicts

platosha:
Some routing cases in hybrid Flow layout + React view applications could produce DOM tree conflicts from Flow server-side changes and React client-side portal removal happening simultaneously. This could throw DOM `NotFoundError` in the browser. This change introduces a dedicated DOM element for React portal outlet, which allows to avoid the error.

Fixes vaadin/hilla#3002


caalador (commented):
Still doesn't happen when manually running, but if I have the server open and run the `hillaViewInFlowLayout` test from `vaadin-platform-react-hybrid-test` it always fails with
```
Unexpected Application Error!
Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at removeChildFromContainer (http://localhost:8080/VAADIN/@fs/C:/Users/.../platform/vaadin-platform-react-hybrid-test/node_modules/.vite/deps/chunk-UJKGNB4G.js?v=3c974d7f:8509:23)
    ```

File: flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx [hunk 1]
```diff
@@ -14,136 +14,138 @@
  * the License.
  */
 /// <reference lib="es2018" />
+import { nanoid } from 'nanoid';
+import type { ReactAdapterElement } from 'Frontend/generated/flow/ReactAdapter.js';
```
caalador, line 18: Flow.tsx can't have a direct link to ReactAdapter.js as it is in another module that may not be available in the application.

caalador (commented):
Could we do a PR with only formatting and then rebase this on it so that the changes would be clearer to see.
Also what is the formatter used as I get different results than those here. I don't think flow has any defined javascript format and uses the idea default instead.

caalador:
For me locally the tests still fail in the hybrid module when executed `[ERROR] Tests run: 12, Failures: 2, Errors: 2, Skipped: 0` the issue seems to still be the same. Also happens when only running one of the tests.
Locally clicking I still can't reproduce the issue.

tepi:
> Also what is the formatter used as I get different results than those here. I don't think flow has any defined javascript format and uses the idea default instead.

I think prettier is used, we have some rules for that in [Flow](https://github.com/vaadin/flow/blob/main/.prettierrc.js)

caalador:
Now I'm getting the exception even just clicking around. Most often repeated by going to `/hilla` and then to `/flow` from where I click to `/flow/hello-hilla`, but also just starting from flow and going to hilla often throws.

caalador (approved):

PR #20771, opened 2024-12-20
fix: ReactRouterOutlet interferes with layout

### LINKED ISSUE #20767: react-router-outlet interferes with layouting / sizing of components
### Description of the bug

When you put a Hilla view inside a Flow main layout, you get a `< react-router-outlet>` element between the main layout component and the Hilla view. This affects how the view is sized, e.g. 100% view width does not work

### Expected behavior

react-router-outlet should have `display: contents` so it is does affect layouting

https://github.com/Artur-/flow-crud-demo/commit/9e05db7f8bfffa77818b03a5d9589dafdf110be6

### Versions

- Vaadin / Flow version: 24.7

Artur-:
Fixes #20767

### DIFF
--- a/flow-react/src/main/resources/META-INF/resources/frontend/ReactRouterOutletElement.tsx
+++ b/flow-react/src/main/resources/META-INF/resources/frontend/ReactRouterOutletElement.tsx
@@ -3,6 +3,11 @@ import { ReactAdapterElement } from "Frontend/generated/flow/ReactAdapter.js";
 import React from "react";
 
 class ReactRouterOutletElement extends ReactAdapterElement {
+  connectedCallback() {
+    super.connectedCallback();
+    this.style.display = 'contents';
+  }
+
   protected render(): React.ReactElement | null {
     return <Outlet />;
   }


caalador (commented):
@platosha does the change in #20770 with the new element fix this same issue as the created element sets display to contents?
The react-router-outlet is left without the style though so this might be needed.
I couldn't reproduce the issue with the grid not showing which might be because I don't have hilla 24.7.crudflow-SNAPSHOt available

Artur-:
The issue can be seen in https://github.com/Artur-/flow-crud-demo/tree/hilla, potentially requires removing some grid data binding not yet in Hilla snapshots. The `crudflow` snapshot is in the pre-releases repository so that one should work

platosha:
> @platosha does the change in #20770 with the new element fix this same issue as the created element sets display to contents? The react-router-outlet is left without the style though so this might be needed. I couldn't reproduce the issue with the grid not showing which might be because I don't have hilla 24.7.crudflow-SNAPSHOt available

Separate issue. #20770 does not change the style of the ReactRouterOutlet wrapper, so this is still needed.

mshabarov (approved):
Without `display:contents` the `Grid` component is not rendered at all, else rendered fine.

The only thing that I'm not entirely sure about is that `VerticalLayout` sets `width:100%` by default and this prevents `react-router-outlet` to show up always, because the side menu vertical layout takes all the width on the page. However, if I set it's width to `auto`, it shows up nicely.

That's probably fine as the example project uses divs and vertical layouts instead of `AppLayout` that has other width default apparently.

PR #20778, opened 2024-12-23
chore: Remove unused declarations in flow-server and flow-data

mshabarov:
Related-to https://github.com/vaadin/flow/issues/16668


flow-server/src/main/java/com/vaadin/flow/server/startup/ServletDeployer.java
```diff
@@ -180,8 +180,6 @@ private void logServletCreation(VaadinServletCreation servletCreation,
             logger.warn(servletCreationMessage);
         } else {
             logger.info(servletCreationMessage);
-            ServletRegistration vaadinServlet = findVaadinServlet(
-                    servletContext);
```
mcollovati, line 184: If this call gets removed, we can also remove the `servletContext` method parameter.

mcollovati (changes requested):

mcollovati (approved):

PR #20818, opened 2025-01-09
fix: remove broken node archive

caalador:
If archive extraction throws
remove arcive to next time re-download.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/NodeInstaller.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/NodeInstaller.java
@@ -265,6 +265,8 @@ public class NodeInstaller {
                             + "Verify connection and proxy settings or follow the https://nodejs.org/en/download/ guide to install Node.js globally.",
                     e);
         } catch (ArchiveExtractionException e) {
+            // AArchive is broken. Remove to download again.
+            data.getArchive().delete();
             throw new InstallationException(
                     "Could not extract the Node archive", e);
         }


flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/NodeInstaller.java
```diff
@@ -265,6 +265,8 @@ private void installNode(InstallData data) throws InstallationException {
                             + "Verify connection and proxy settings or follow the https://nodejs.org/en/download/ guide to install Node.js globally.",
                     e);
         } catch (ArchiveExtractionException e) {
+            // AArchive is broken. Remove to download again.
+            data.getArchive().delete();
```
mcollovati, line 269: Should we try to delete also when `DownloadException` is thrown?

flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/NodeInstaller.java
```diff
@@ -265,6 +265,8 @@ private void installNode(InstallData data) throws InstallationException {
                             + "Verify connection and proxy settings or follow the https://nodejs.org/en/download/ guide to install Node.js globally.",
                     e);
         } catch (ArchiveExtractionException e) {
+            // AArchive is broken. Remove to download again.
```
mcollovati, line 268:
```suggestion
            // Archive is broken. Remove to download again.
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/NodeInstaller.java
caalador: added for failed downloads also.

flow-server/src/main/java/com/vaadin/flow/server/frontend/installer/NodeInstaller.java
caalador: Comment removed due to change in removal.

mcollovati (approved):

PR #20829, opened 2025-01-10
chore: Run npm install with --legacy-peer-deps for React 19

Artur-:
Needed until there is a new Lit release

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
@@ -32,6 +32,7 @@ import java.util.stream.Collectors;
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunNpmInstall.java
```diff
@@ -267,6 +268,9 @@ private void runNpmInstall() throws ExecutionFailedException {
         }
 
         npmInstallCommand.add("--ignore-scripts");
+        if (options.getFeatureFlags().isEnabled(FeatureFlags.REACT19)) {
+            npmInstallCommand.add(" --legacy-peer-deps");
```
caalador, line 272: Just for clarity this could have a comment that should/can be removed after Lit x.y.z has been released and taken into use.

Artur-: There is a new lit release so maybe we don’t need this pr at all

PR #20836, opened 2025-01-13
fix: set viewport-fit: cover to enable CSS env() variables in PWAs

vursen:
## Description

The `env()` variables aren't working in PWAs unless `viewport-fit: cover` is added to the `<meta name="viewport">`, see https://developer.mozilla.org/en-US/docs/Web/CSS/env#usage.

Fixes https://github.com/vaadin/web-components/issues/8449

## Type of change

- [x] Bugfix

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/page/Viewport.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/page/Viewport.java
@@ -25,7 +25,7 @@ import java.lang.annotation.Target;
 /**
  * Defines a viewport tag that will be added to the HTML of the host page of a
  * UI class. If no viewport tag has been defined, a default of
- * <code>width=device-width, initial-scale=1.0</code> is used.
+ * <code>width=device-width, initial-scale=1.0, viewport-fit=cover</code> is used.
  *
  * @author Vaadin Ltd
  * @since 1.0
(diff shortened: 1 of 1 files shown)


flow-tests/test-ccdm/src/test/java/com/vaadin/flow/ccdmtest/IndexHtmlRequestHandlerIT.java
```diff
@@ -53,7 +53,7 @@ public void should_add_appShellAnnotations() {
                 By.cssSelector("meta[name=viewport]"));
         Assert.assertNotNull(metaViewPort);
         Assert.assertEquals(
-                "width=device-width, height=device-height, initial-scale=1.0",
+                "width=device-width, height=device-height, initial-scale=1.0, viewport-fit: cover",
```
vursen, line 56:
```suggestion
                "width=device-width, height=device-height, initial-scale=1.0, viewport-fit=cover",
```

vursen:
I wonder if it's potentially a breaking change?

mshabarov:
> I wonder if it's potentially a breaking change?

I see a benefit of setting this option by default, because this gives a flexibility and customisability to developers to setup the viewport best alignment on modern side-to-side device screens, e.g. with `env(safe-area-inset-*)`. 

Though this should be tested on various devices before released. And if we spot any significant changes in displaying, we may want to postpone this change to Vaadin 25.

Do you @vursen or someone from Design System team mind to test this on various devices or shall Flow team do?

vursen:
> Do you @vursen or someone from Design System team mind to test this on various devices or shall Flow team do?

Based on our internal testing across various devices (iOS, Android), this change only has an effect in PWAs, where it enables the env variables so that they start behaving the same as in regular iOS Safari. I guess it should be safe.

PR #20846, opened 2025-01-15
fix: Navigate with full url

### LINKED ISSUE #19580: Strange behavior with History -> replaceState(state,location) when react is enabled
### Description of the bug

Since update to Vaadin 24.4.1 there is a strange behavoir when using replaceState() in History class. If I use the method to update the displayed URL (eg. history.replaceState(null,"http://localhost:8080/?test=test") , it results in opening the invalid url http://localhost:8080/http://localhost:8080/?test=test. If I set react.enabled=false, then it works as expected.

### Expected behavior

Browser shows expected URL after history.replaceState call.

Use any Vaadin 24.4.1 project (eg. skeleton-starter-flow-spring.24). 

Call method for updating url parameter, eg:

 ```
       Button button = new Button("Test - Add URL Parameter", e -> {
            updateUrlRequestParameter("test",textField.getValue());
        });
```

with:

```
public static void updateUrlRequestParameter(String key, String value) {
        Page page = UI.getCurrent().getPage();
        page.fetchCurrentURL(url -> {
            try {
                UriComponentsBuilder builder = UriComponentsBuilder.fromUri(url.toURI());
                UriComponents components = builder.build();
                MultiValueMap<String, String> parameters = components.getQueryParams();
                MultiValueMap<String, String> newParams = new LinkedMultiValueMap<>();
                newParams.putAll(parameters);
                if(newParams.get(key) != null) {
                    newParams.remove(key);
                }
                if(value != null) {
                    newParams.put(key, List.of(value));
                }
                String newUrl = builder.replaceQueryParams(newParams).build().toUriString();
                page.getHistory().replaceState(null, newUrl);
            } catch(Exception exception) {
                //handle exception
            }
        });
    }
```

### Versions

- Vaadin / Flow version: 24.4.1
- Java version: 17
- OS version: Windows 11

caalador:
Fix navigating with full url
for instance when adding a
query parameter at the end.

Fixes #19580

### DIFF
--- a/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
+++ b/flow-server/src/main/resources/com/vaadin/flow/server/frontend/Flow.tsx
@@ -382,7 +382,11 @@ function Flow() {
         (event: CustomEvent<{ state: unknown; url: string; replace?: boolean; callback: boolean }>) => {
             // @ts-ignore
             window.Vaadin.Flow.navigation = true;
-            const path = '/' + event.detail.url;
+            // clean base uri away if for instance redirected to http://localhost/path/user?id=10
+            // else the whole http... will be appended to the url see #19580
+            const path = event.detail.url.startsWith(document.baseURI)
+                ? '/' + event.detail.url.slice(document.baseURI.length)
+                : '/' + event.detail.url;
             fromAnchor.current = false;
             queuedNavigate(path, event.detail.callback, { state: event.detail.state, replace: event.detail.replace });
         },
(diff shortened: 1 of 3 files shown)


mcollovati (dismissed):
Tested locally and it works fine.

I just wonder if we should raise an error on the client side if the given URL is absolute and does not match `document.baseURI`. 
With Vaadin router, an error is raised, so it would probably be good to have a similar behavior also with React.

image

caalador:
I guess we should, but I wonder should it be  part of this fix or should we have an enhancement ticket for it.

mcollovati (approved):

Issue #20589, opened 2024-12-02
Error message from duplicate routes could be more helpful


Legioth:
It's easy to end up with multiple routes with the same path by mistake, especially when playing around with the platform and copypasting examples from docs, so this is a fairly common scenario. The problem is that the resulting stack trace does not make it easy to understand what the problem is.

Could the framework print a clear and unambiguous error in the output when this is encountered, instead of relying on a cryptic stacktrace?

PR #20847, opened 2025-01-15
fix: make route config errors more prominent

tepi:
When route configuration errors happen, make sure to output the actual error message as the first output at `ERROR` level. When this is detected at startup, we can't prevent printing out the stack trace too since the container does that when initialization throws. Also checked all the messages for different route misconfiguration cases, and they seem to be fine.

Fixes https://github.com/vaadin/flow/issues/20589

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
@@ -791,9 +791,9 @@ final class RouteSegment implements Serializable {
 
         String messageFormat;
         if (isParameter()) {
-            messageFormat = "Navigation targets (@Route and @RouteAlias values) must be unique, found navigation targets '%s' and '%s' with parameter having the same route.";
+            messageFormat = "Navigation target paths (considering @Route, @RouteAlias and @RoutePrefix values) must be unique, found navigation targets '%s' and '%s' with parameter having the same route.";
         } else {
-            messageFormat = "Navigation targets (@Route and @RouteAlias values) must be unique, found navigation targets '%s' and '%s' with the same route.";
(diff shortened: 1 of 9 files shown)


caalador (approved):
Is fine in devMode when adding/renaming a route that conflicts with server running, but gets hidden on clean devMode start as the routes.tsx is generated later and the exception gets thrown.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
```diff
@@ -791,9 +791,9 @@ private RuntimeException ambigousTarget(Class<? extends Component> target) {
 
         String messageFormat;
         if (isParameter()) {
-            messageFormat = "Navigation targets must have unique routes, found navigation targets '%s' and '%s' with parameter have the same route.";
+            messageFormat = "Navigation targets (@Route and @RouteAlias values) must be unique, found navigation targets '%s' and '%s' with parameter having the same route.";
```
caalador, line 794: I wouldn't say this as there is also `RoutePrefix` which is used with route and routeAlias and that is taken into account when collecting routes.
So it is more `Navigation target paths` that need to be unique.

flow-server/src/main/java/com/vaadin/flow/router/internal/RouteSegment.java
```diff
@@ -791,9 +791,9 @@ private RuntimeException ambigousTarget(Class<? extends Component> target) {
 
         String messageFormat;
         if (isParameter()) {
-            messageFormat = "Navigation targets must have unique routes, found navigation targets '%s' and '%s' with parameter have the same route.";
+            messageFormat = "Navigation target paths (considering @Route, @RouteAlias and @RoutePrefix values) must be unique, found navigation targets '%s' and '%s' with parameter having the same route.";
```
caalador, line 794: As 9 tests failed after updating the message perhaps it would make sense to have these as constants that are also used in the tests.

caalador (approved):

PR #20867, opened 2025-01-16
chore: Add convenience list and count without Filter

Artur-:
Using a default implementation does not put any extra burden on the interface implementor


mshabarov (approved):
Why didn't I suggest it last time 🤔 
LGTM.

Artur-:
Haha, last time I thought about rejecting `count(null)` but this works much better

PR #20871, opened 2025-01-16
fix: Add classpath elements to Surefire plugin

Ansku:
## Description

Maven doesn't use `.classpath` and as such isn't aware that it should use `flow-server` as a source directory. That leads to a whole lot of Project Errors in IntelliJ IDEA if `flow-client` module isn't ignored. This configuration change gets rid of those errors.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/flow-client/pom.xml
+++ b/flow-client/pom.xml
@@ -192,6 +192,9 @@
                     <excludes>
                         <exclude>**/Gwt*.java</exclude>
                     </excludes>
+                    <additionalClasspathElements>
+                        <additionalClasspathElement>${basedir}/../flow-server/src/main/java</additionalClasspathElement>
+                    </additionalClasspathElements>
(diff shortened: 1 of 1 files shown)


tepi:
I can't reproduce the original issue without this change (IDEA 2024.3 & latest macos). On the other hand this change does not seem to break anything for me either.

mcollovati:
I couldn't reproduce the problem either.
Honestly, I wonder how adding a source folder to the classpath elements, where compiled classes are supposed to be added, can fix project errors.
However, since the flow-client setup for 2.11 is a bit tricky and the change does not seem to break anything as @tepi said, if it solves the issues for @Ansku  I agree to merge.

mcollovati (approved):

PR #20877, opened 2025-01-18
Revert "fix: Store the exception and not an array with the exception for uncaught promise exceptions"

Artur-:
Reverts vaadin/flow#20840


Artur-:
This causes other issues

caalador (approved):

PR #20879, opened 2025-01-19
Revert "chore: Add convenience list and count without Filter"

Artur-:
Reverts vaadin/flow#20867

This causes problems with Hilla that does not support method overloading, as TS/JS does not support that

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/data/CountService.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/data/CountService.java
@@ -16,6 +16,6 @@ public interface CountService {
      *            the filter, or {@code null} to use no filter
      * @return the number of items in the service that match the filter
      */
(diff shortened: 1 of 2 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/data/CountService.java
```diff
@@ -9,22 +9,13 @@
...
-
     /**
      * Counts the number of items that match the given filter.
      *
      * @param filter
      *            the filter, or {@code null} to use no filter
      * @return the number of items in the service that match the filter
      */
-    long count(@Nullable Filter filter);
+    public long count(@Nullable Filter filter);
```
mshabarov, line 19:
```suggestion
   long count(@Nullable Filter filter);
```

mshabarov (approved):

mshabarov (approved):

Issue #20896, opened 2025-01-23
Support Bun 1.2


tepi:
[Bun 1.2](https://bun.sh/blog/bun-v1.2) was released yesterday and at least the lockfile name and format have been changed. Flow needs a fix to support the new version.

PR #20900, opened 2025-01-23
fix: support bun 1.2 lock file

tepi:
Fixes https://github.com/vaadin/flow/issues/20896


mcollovati (approved):
Tested locally with bun 1.1.45 and 1.2.0, and it works correctly with both.

PR #20909, opened 2025-01-24
fix: build service worker without writing it to FS

vursen:
## Description

https://github.com/vaadin/flow/pull/20894 reintroduced https://github.com/vaadin/flow/issues/13123 where the service worker build would cause a server restart in dev mode. This PR fixes this issue again.

## Type of change

- [x] Bugfix

### DIFF
--- a/flow-server/src/main/resources/vite.generated.ts
+++ b/flow-server/src/main/resources/vite.generated.ts
@@ -106,7 +106,7 @@ function injectManifestToSWPlugin(): rollup.Plugin {
         const { manifestEntries } = await getManifest({
           globDirectory: buildOutputFolder,
           globPatterns: ['**/*'],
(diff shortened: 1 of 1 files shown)


flow-server/src/main/resources/vite.generated.ts
```diff
@@ -155,7 +157,12 @@ function buildSWPlugin(opts: { devMode: boolean }): PluginOption {
     },
     async buildStart() {
       if (devMode) {
-        await build(buildConfig);
+        buildOutput = await build(buildConfig) as rollup.RollupOutput;
+      }
+    },
+    async load(id) {
+      if (id.endsWith('sw.js')) {
+        return buildOutput.output[0].code;
```
vursen, line 165: **note:** The output code already includes a sourcemap link, so there is no need to return it separately in `transform(code, id)` as it was done previously.

mcollovati (approved):

PR #20929, opened 2025-01-29
feat: mark full size components with data attribute

sissbruecker:
## Description

Adds data attributes to components that have been configured to use the full size in a layout, by either calling `setWidthFull`, `setHeightFull` or `setSizeFull`. Vaadin components such as `HorizontalLayout` and `VerticalLayout` will include additional styles targeting children with these data attributes, to effectively make these children take up the remaining space in a layout, rather than the full size.

For now, the respective styles will only be added by the Vaadin components if the `com.vaadin.experimental.layoutImprovements` feature flag is enabled (see https://github.com/vaadin/web-components/pull/8552).

Part of https://github.com/vaadin/flow-components/issues/6998

## Type of change

- Feature

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
@@ -341,12 +341,12 @@ public interface HasSize extends HasElement {
      * with a fixed size may shrink to allow the full-size component to take up
      * as much space as possible, or the full-size component may cause the
      * layout to overflow. To improve this, you can enable the
-     * {@code com.vaadin.experimental.layoutImprovements} feature flag to
-     * effectively make full-size components take up the <b>remaining</b> space
-     * in the layout, rather than explicitly using 100% size of the layout. This
-     * applies additional CSS styles that allow the component to shrink below
-     * 100% if there are other components with fixed or relative sizes in the
-     * layout.
+     * {@code com.vaadin.experimental.layoutComponentImprovements} feature flag
+     * to effectively make full-size components take up the <b>remaining</b>
+     * space in the layout, rather than explicitly using 100% size of the
+     * layout. This applies additional CSS styles that allow the component to
+     * shrink below 100% if there are other components with fixed or relative
+     * sizes in the layout.
      */
     default void setSizeFull() {
         setWidth("100%");
@@ -367,12 +367,12 @@ public interface HasSize extends HasElement {
      * with a fixed size may shrink to allow the full-size component to take up
      * as much space as possible, or the full-size component may cause the
      * layout to overflow. To improve this, you can enable the
-     * {@code com.vaadin.experimental.layoutImprovements} feature flag to
-     * effectively make full-size components take up the <b>remaining</b> space
-     * in the layout, rather than explicitly using 100% size of the layout. This
-     * applies additional CSS styles that allow the component to shrink below
-     * 100% if there are other components with fixed or relative sizes in the
-     * layout.
+     * {@code com.vaadin.experimental.layoutComponentImprovements} feature flag
+     * to effectively make full-size components take up the <b>remaining</b>
+     * space in the layout, rather than explicitly using 100% size of the
+     * layout. This applies additional CSS styles that allow the component to
+     * shrink below 100% if there are other components with fixed or relative
+     * sizes in the layout.
      */
     default void setWidthFull() {
         setWidth("100%");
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
```diff
@@ -332,30 +334,73 @@ default Optional<Unit> getHeightUnit() {
...
+     * in the layout, rather than explicitly using 100% size of the layout. This
+     * applies additional CSS styles that allow the component to shrink below
+     * 100% if there are other components with fixed or relative sizes in the
+     * layout.
      */
     default void setSizeFull() {
         setWidth("100%");
         setHeight("100%");
+        getElement().setAttribute("data-v-width-full", true);
+        getElement().setAttribute("data-v-height-full", true);
```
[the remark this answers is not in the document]
sissbruecker: [`data-*`](https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Solve_HTML_problems/Use_data_attributes) attributes are basically the standard method to extend HTML elements with custom attributes. But those always must start with `data-*`. Using any other attribute name that is not defined in the HTML spec runs the risk of triggering errors in HTML validators.

I feel we could actually drop the `v` from the name, it seems very unlikely that Flow users will make use of data attributes at all, which is more of an extension mechanism used by client-side code. So there's basically no risk of a conflict.

sissbruecker:
@mshabarov I've dropped the `v-` from the attribute name.

flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
```diff
@@ -47,6 +47,7 @@ public interface HasSize extends HasElement {
      */
     default void setWidth(String width) {
         getElement().getStyle().setWidth(width);
+        getElement().removeAttribute("data-width-full");
```
mshabarov, line 50: This would leave a small gap which is when you call `getElement().getStyle().setWidth()` instead (add-ons may use this), you'll not get this attribute removed.

I was thinking of adding `removeAttribute` to the `getElement().getStyle().setWidth()`, but there is no direct access to the `Element` API. However, we can check if `this` is an `Element`, case and call this method.

What do you think, does it worth it or we can make an assumption that component's method would be sufficient for most of the usages?

flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
```diff
@@ -194,6 +195,7 @@ default Optional<Unit> getWidthUnit() {
      */
     default void setHeight(String height) {
         getElement().getStyle().setHeight(height);
+        getElement().removeAttribute("data-height-full");
```
mshabarov, line 198: Could we place these constants to `Constants` ?

flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
```diff
@@ -332,30 +334,73 @@ default Optional<Unit> getHeightUnit() {
...
+     * in the layout, rather than explicitly using 100% size of the layout. This
+     * applies additional CSS styles that allow the component to shrink below
+     * 100% if there are other components with fixed or relative sizes in the
+     * layout.
      */
     default void setSizeFull() {
         setWidth("100%");
         setHeight("100%");
+        getElement().setAttribute("data-width-full", true);
+        getElement().setAttribute("data-height-full", true);
```
mshabarov, line 355: Would it better to just call 
```
setWidthFull();
setHeightFull();
```
where these attributes are already added?

sissbruecker: Makes sense, done.

flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
sissbruecker: I think you have a better idea what kind of constants that class should contain. I think it wasn't too bad with just a few duplicates and covering those with unit tests. But I've extracted the attributes into constants for now.

flow-server/src/main/java/com/vaadin/flow/component/HasSize.java
sissbruecker: I did take a look at `Style` and the probably relevant `BasicElementStyle` subclass, but am not sure how get a reference to the element from there.

In general I would assume that:
- Users would usually use `HasSize` consistently
- Probably not switch between full size and custom size that often

For now I would probably go with what we have. If this turns out to be a problem we have another option of changing the CSS selector in the web components to check for both, the data attribute and `style="width: 100%"` / `style="height: 100%"`.

mshabarov (approved):

Issue #20741, opened 2024-12-18
Use Jackson for public API and implementation instead of Elemental


mshabarov:
### Describe your motivation

Jackson library (`com.fasterxml.jackson`) is more popular than Elemental (`elemental.json`), thus Vaadin should better use it in public API.

### Describe the solution you'd like

0. Analyse and list the usage of Elemental in Flow: how many publicly using cases do we have and how many are used internally. I expect the main usage in `Element` class, but there may be more, e.g. in util classes.
1. Add an adapter code that converts Jackson to/from Elemental, if it's problematic to replace it in implementation right away, else go to step 2.
2. Change Flow internal implementation to use Jackson
3. Add same public methods found in (0), but that use Jackson for arguments and return values
4. Deprecate existing public method with Elemental
5. Remove Elemental variants in next major version.

knoobie:
Just a note: I would suggest to implement some low level tests that Jaxkson does what you think it does.. I'm saying this because of their kinda "fast" releases the last years - which can sometimes break things.. so it might be worth to improve coverage in that area - especially with different Jackson versions in client projects and an upcoming new Jackson major (soon TM)

mshabarov:
**Acceptance Criteria**:
- [ ] As a general requirement, we have to avoid any breaking changes including behaviour changes in Vaadin 24.8, e.g. that can break Vaadin standard components/custom components/add-ons; all breaking changes should be done in Vaadin 25.
- [ ] For methods with `Elemental` classes in arguments a new method is created that has the same signature except it takes a `Jackson` object. The old method is deprecated. E.g. `WebComponent::fireEvent`.
- [ ] When it’s not possible to create a duplicate method, e.g. in functional interfaces like `DataGenerator` - to add a note in javadoc for Vaadin 24.8 to the existing method and change the method's signature in Vaadin 25.
- [ ] For consumers/functional interfaces, if they are used in the implementation of methods from point (1), we need to do the conversion from `Elemental` to `Jackson` and vice versa in Vaadin 24.8 and just swap `Elemental`/`Jackson` in Vaadin 25.
- [ ] For API shared with components or even having a small chance to be extended in projects (e.g. `DataCommunicator` or custom `DataCommunicator` extension) - to update javadoc in Vaadin 24.8 and to change method signature in Vaadin 25.
[...]

PR #20948, opened 2025-02-05
feat: Add jackson utils and codec

caalador:
Create jackson versions
of the JsonUtils and JsonCodec.

part of #20741

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/JacksonCodec.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/JacksonCodec.java
@@ -47,7 +47,7 @@ import com.vaadin.flow.internal.nodefeature.ReturnChannelRegistration;
  * For internal use only. May be renamed or removed in a future release.
  *
  * @author Vaadin Ltd
- * @since 1.0
+ * @since 24.7
  */
 public class JacksonCodec {
     /**
@@ -217,7 +217,8 @@ public class JacksonCodec {
     /**
      * Helper for decoding any "primitive" value that is directly supported in
      * JSON. Supported values types are {@link String}, {@link Number},
-     * {@link Boolean}, {@link JsonNode}. <code>null</code> is also supported.
+     * {@link Boolean}, {@link JsonNode}.
+     * {@link com.fasterxml.jackson.databind.node.NullNode} is also supported.
      *
      * @param json
      *            the JSON value to decode
--- a/flow-server/src/main/java/com/vaadin/flow/internal/JacksonCodec.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/JacksonCodec.java
@@ -1,0 +1,283 @@
+/*
+ * Copyright 2000-2025 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.internal;
+
+import java.io.Serializable;
+import java.util.stream.Stream;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeType;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.dom.Node;
+import com.vaadin.flow.internal.nodefeature.ReturnChannelRegistration;
+
+/**
+ * Utility for encoding objects to and from JSON.
+ * <p>
+ * Supported types are
+ * <ul>
+ * <li>{@link String}
+ * <li>{@link Boolean} and <code>boolean</code>
+ * <li>{@link Integer} and <code>int</code>
+ * <li>{@link Double} and <code>double</code> (<code>NaN</code> and infinity not
+ * supported)
+ * <li>{@link JsonNode} and all its sub types
+ * <li>{@link Element} (encoded as a reference to the element)
+ * <li>{@link Component} (encoded as a reference to the root element)
+ * </ul>
+ *
+ * <p>
+ * For internal use only. May be renamed or removed in a future release.
+ *
+ * @author Vaadin Ltd
+ * @since 1.0
+ */
+public class JacksonCodec {
+    /**
+     * Type id for a complex type array containing an {@link Element}.
+     */
+    public static final int NODE_TYPE = 0;
+
+    /**
+     * Type id for a complex type array containing a {@link ArrayNode}.
+     */
+    public static final int ARRAY_TYPE = 1;
+
+    /**
+     * Type id for a complex type array identifying a
+     * {@link ReturnChannelRegistration} reference.
+     */
+    public static final int RETURN_CHANNEL_TYPE = 2;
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -0,0 +1,454 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+
+/**
+ * Helpers for using <code>jackson</code>.
+ * <p>
+ * For internal use only. May be renamed or removed in a future release.
+ *
+ * @author Vaadin Ltd
+ * @since 1.0
```
mcollovati, line 52:
```suggestion
 * @since 24.7
```

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -0,0 +1,454 @@
...
+     */
+    public static ArrayNode createArrayNode() {
+        return objectMapper.createArrayNode();
+    }
+
+    /**
+     * Collects a stream of JSON values to a JSON array.
+     *
+     * @author Vaadin Ltd
+     * @since 1.0
```
mcollovati, line 92:
```suggestion
     * @since 24.7
```

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -0,0 +1,454 @@
...
+        assert b.getNodeType() == JsonNodeType.BOOLEAN;
+        return a.asBoolean() == b.asBoolean();
+    }
+
+    public static boolean numbersEqual(JsonNode a, JsonNode b) {
+        assert a.getNodeType() == JsonNodeType.NUMBER;
+        assert b.getNodeType() == JsonNodeType.NUMBER;
+        return Double.doubleToRawLongBits(a.doubleValue()) == Double
+                .doubleToRawLongBits(b.doubleValue());
+    }
```
mcollovati, line 192: nit: need Javadoc, unless we can reduce the visibility

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -0,0 +1,454 @@
...
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    public static List<String> getKeys(JsonNode a) {
+        List<String> keys = new ArrayList<>();
+        for (Iterator<Map.Entry<String, JsonNode>> it = a.fields(); it
```
mcollovati, line 221: nit: can maybe use `JsonNode.fieldNames()` to simplify the method.

```
a.fieldNames().forEachRemaining(keys::add);
```

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -0,0 +1,454 @@
...
+     * @param tClass
+     *            class of converted object instance
+     * @return converted object instance
+     * @param <T>
+     *            type of result instance
+     */
+    public static <T> T readToObject(ObjectNode jsonObject, Class<T> tClass) {
+        Objects.requireNonNull(jsonObject, CANNOT_CONVERT_NULL_TO_OBJECT);
+        try {
+            return objectMapper.readValue(jsonObject.toString(), tClass);
```
mcollovati, line 394: Can we use `ObjectMapper.treeToValue` to avoid the Object to String conversion?

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -0,0 +1,454 @@
...
+     *
+     * @param jsonValue
+     *            JSON value to convert, not {@code null}
+     * @param tClass
+     *            class of converted object instance
+     * @return converted object instance
+     * @param <T>
+     *            type of result instance
+     */
+    public static <T> T readValue(ObjectNode jsonValue, Class<T> tClass) {
```
mcollovati, line 412: It looks like the same as `readToObject`. Can we remove it?
Otherwise, could this method just call `readToObject`?

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -0,0 +1,454 @@
...
+     *            type reference of converted object instance
+     * @return converted object instance
+     * @param <T>
+     *            type of result instance
+     */
+    public static <T> T readValue(ObjectNode jsonValue,
+            TypeReference<T> typeReference) {
+        Objects.requireNonNull(jsonValue, CANNOT_CONVERT_NULL_TO_OBJECT);
+        try {
+            return objectMapper.readValue(jsonValue.toString(), typeReference);
```
mcollovati, line 437: Same as before, what about using `treeToValue` instead of `readValue` plus object to JSON string conversion?

flow-server/src/main/java/com/vaadin/flow/internal/JacksonCodec.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+        assert !canEncodeWithoutTypeInfo(type);
+        throw new IllegalArgumentException(
+                "Can't encode " + value.getClass() + " to json");
+    }
+
+    /**
+     * Helper for decoding any "primitive" value that is directly supported in
+     * JSON. Supported values types are {@link String}, {@link Number},
+     * {@link Boolean}, {@link JsonNode}. <code>null</code> is also supported.
```
mcollovati, line 220: Does it mean that **null JSON representation** is supported, or that null can be given as a parameter?
If it is the first case, we should maybe make it explicit; otherwise, if it is the second case, we should remove the assertion.

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/internal/JacksonCodec.java
```diff
@@ -0,0 +1,283 @@
...
+ * <li>{@link JsonNode} and all its sub types
+ * <li>{@link Element} (encoded as a reference to the element)
+ * <li>{@link Component} (encoded as a reference to the root element)
+ * </ul>
+ *
+ * <p>
+ * For internal use only. May be renamed or removed in a future release.
+ *
+ * @author Vaadin Ltd
+ * @since 1.0
```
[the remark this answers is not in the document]
caalador: done

flow-server/src/main/java/com/vaadin/flow/internal/JacksonCodec.java
caalador: Null node not literal null.

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
caalador: added jdoc as they may be useful somewhere

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
caalador: done

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
caalador: It's mostly there as JsonUtils has it. Having it call readToObject.

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
caalador: done

mcollovati (approved):

PR #20980, opened 2025-02-11
feat(crud): support configured max page size

cromoteca:
When using `ListRepositoryService` and its subclasses, one can pass an arbitrary page size, potentially querying many lines. This is also true in Hilla, where the client side decides.

This PR uses the existing `spring.data.rest.pageable.max-page-size` property, if defined, and silently caps the maximum page size.

Closes https://github.com/vaadin/hilla/issues/3203.


mshabarov:
We are considering moving these services back to Hilla, see https://vaadin.slack.com/archives/C6RAXJATF/p1737967424448129 for discussion on this topic.

PR #20985, opened 2025-02-12
feat: RpcHandlers to use Jackson

caalador:
Make rpc handlers use jackson
instead of elemental

part of #20741

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
@@ -211,9 +211,7 @@ public final class WebComponent<C extends Component> implements Serializable {
 
         if (value == null) {
             componentHost.executeJs(UPDATE_PROPERTY_NULL, propertyName);
-        }
-
-        if (value instanceof Integer) {
+        } else if (value instanceof Integer) {
             componentHost.executeJs(UPDATE_PROPERTY, propertyName,
                     (Integer) value);
         } else if (value instanceof Double) {
--- a/flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
@@ -19,6 +19,9 @@ package com.vaadin.flow.component.webcomponent;
 import java.io.Serializable;
 import java.util.Objects;
 
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.node.ValueNode;
+
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.server.webcomponent.PropertyConfigurationImpl;
@@ -228,6 +231,22 @@ public final class WebComponent<C extends Component> implements Serializable {
             // that expected behavior.
             componentHost.executeJs(String.format(UPDATE_PROPERTY_FORMAT,
                     ((JsonValue) value).toJson()), propertyName);
+        } else if (value instanceof ObjectNode) {
+            // this gets around executeJavaScript limitation.
+            // Since properties can take JsonValues, this was needed to allow
+            // that expected behavior.
+            componentHost.executeJs(
+                    String.format(UPDATE_PROPERTY_FORMAT, value), propertyName);
+        } else if (value instanceof ValueNode) {
+            // this gets around executeJavaScript limitation.
+            // Since properties can take JsonValues, this was needed to allow
+            // that expected behavior.
+            componentHost.executeJs(
+                    String.format(UPDATE_PROPERTY_FORMAT, value), propertyName);
+        } else {
+            throw new IllegalArgumentException(
+                    String.format("Unknown property type '%s'",
+                            value.getClass().getSimpleName()));
         }
     }
 }
--- a/flow-server/src/main/java/com/vaadin/flow/dom/DomEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/dom/DomEvent.java
@@ -23,6 +23,7 @@ import java.util.function.Consumer;
 
 import com.fasterxml.jackson.databind.JsonNode;
 
+import com.vaadin.flow.internal.JacksonUtils;
 import com.vaadin.flow.internal.NodeOwner;
 import com.vaadin.flow.internal.StateNode;
 import com.vaadin.flow.internal.StateTree;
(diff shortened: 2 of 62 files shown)


flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
```diff
@@ -228,6 +231,22 @@ private void setProperty(String propertyName, Object value) {
...
+        } else if (value instanceof ValueNode) {
+            // this gets around executeJavaScript limitation.
+            // Since properties can take JsonValues, this was needed to allow
+            // that expected behavior.
+            componentHost.executeJs(
+                    String.format(UPDATE_PROPERTY_FORMAT, value), propertyName);
+        } else {
+            throw new IllegalArgumentException(
+                    String.format("Unknown property type '%s'",
+                            value.getClass().getSimpleName()));
```
mcollovati, line 249: This could be a breaking change, since before no exception was thrown.
However, I suppose this would never happen.

Anyway, IDEA suggested that `value.getClass().getSimpleName()` may produce NPE

flow-server/src/main/java/com/vaadin/flow/dom/DomEvent.java
```diff
@@ -59,7 +61,7 @@ public class DomEvent extends EventObject {
      * @see Element#addEventListener(String, DomEventListener)
      * @see DomEventListener
      */
-    public DomEvent(Element source, String eventType, JsonObject eventData) {
+    public DomEvent(Element source, String eventType, JsonNode eventData) {
```
mcollovati, line 64: I wonder if we should preserve the original constructor and deprecate it, for this public API.
There is the possibility that testing frameworks like UI Unit Test and Karibu are using this constructor.

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -682,6 +689,26 @@ public Element setPropertyJson(String name, JsonValue value) {
         return this;
     }
 
+    // Distinct name so setProperty("foo", null) is not ambiguous
```
mcollovati, line 692: Javadoc missing

flow-server/src/main/java/com/vaadin/flow/dom/Element.java
```diff
@@ -682,6 +689,26 @@ public Element setPropertyJson(String name, JsonValue value) {
...
+        if (value == null) {
+            throw new IllegalArgumentException(USE_SET_PROPERTY_WITH_JSON_NULL);
+        }
+
+        setRawProperty(name, value);
+        return this;
+    }
+
+    // Distinct name so setProperty("foo", null) is not ambiguous
+    public Element setPropertyJson(String name, ValueNode value) {
```
mcollovati, line 703: Javadoc missing

mcollovati (changes requested):

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -106,6 +111,44 @@ public static JsonNode createNode(Object value) {
...
+    }
+
+    /**
+     * Map JsonObject to ObjectNode.
+     *
+     * @param jsonObject
+     *            JsonObject to change
+     * @return ObjectNode of elemental json object
+     */
+    public static ObjectNode mapElemental(JsonObject jsonObject) {
```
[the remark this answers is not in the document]
caalador: We can't completely remove elemental as the client side needs it. Also any shared class will also always use elemental instead of jackson. So basically this will probably always be needed in some place where the client and server intercept.

flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -106,6 +111,44 @@ public static JsonNode createNode(Object value) {
...
+    }
+
+    /**
+     * Map JsonValue to ObjectNode.
+     *
+     * @param jsonValue
+     *            JsonValue to change
+     * @return ObjectNode of elemental json value
+     */
+    public static JsonNode mapElemental(JsonValue jsonValue) {
```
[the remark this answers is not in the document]
caalador: Same as for other.

flow-server/src/main/java/com/vaadin/flow/server/communication/UidlWriter.java
```diff
@@ -317,7 +319,13 @@ private static ReturnChannelRegistration createReturnValueChannel(
...
                     registrations.forEach(ReturnChannelRegistration::remove);
 
-                    action.accept(arguments.get(0));
+                    if (arguments.get(0) instanceof NullNode) {
+                        action.accept(Json.createNull());
+                    } else if (arguments.get(0) instanceof ValueNode) {
+                        action.accept(Json.create(arguments.get(0).asText()));
+                    } else {
+                        action.accept(Json.parse(arguments.get(0).toString()));
+                    }
```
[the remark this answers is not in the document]
caalador: done

flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
caalador: The problem before was that it died silently not giving any information on why things didn't render or work.
Also we have a value == null if in the beginning so I can't see how value.getClass() could be null.

flow-server/src/main/java/com/vaadin/flow/dom/DomEvent.java
caalador: added

flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
mcollovati: But `value == null` is not in the same `if/else-if/else` block, so the NPE could potentially happen

mcollovati: > The problem before was that it died silently not giving any information on why things didn't render or work.

Yeah, right. This is indeed a fix.

caalador: The value doesn't change anywhere, but I can add the "missing" else to connect the if

mcollovati: I'm probably missing something here: the value does not change, but we do not exit the method if `value` is null, so in this case we will always fall in the `else` block and have an NPE.
Is there something I did not see?

caalador: Right. That is true. What a weird method it is. Added Else to make it not check anything if null value.

mcollovati (approved):

PR #21000, opened 2025-02-14
Refactor MutableTreeRevision

Legioth:
* Rename TreeManipulator to ResultBuilder and use it only for multi-step commands and commands that affect multiple nodes
* Explicitly return a result when handling a command instead of indirectly producing one
* Transactions produce multiple results so they have custom top-level handling instead of following the general pattern
* Handle errors caused by concurrent modifications by returning a Reject result instead of passing the error through an instance field

### DIFF
--- a/signals/src/main/java/com/vaadin/signals/impl/MutableTreeRevision.java
+++ b/signals/src/main/java/com/vaadin/signals/impl/MutableTreeRevision.java
@@ -25,12 +25,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
-import java.util.function.Consumer;
-import java.util.function.UnaryOperator;
+import java.util.function.Function;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.DoubleNode;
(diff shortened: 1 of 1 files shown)


signals/src/main/java/com/vaadin/signals/impl/MutableTreeRevision.java
```diff
@@ -70,235 +67,215 @@
...
+        private Reject updateMapChildren(Id resolvedParentId,
+                Function<Map<String, Id>, CommandResult.Reject> mapUpdater) {
+            Data node = data(resolvedParentId);
+
             LinkedHashMap<String, Id> map = new LinkedHashMap<>(
                     node.mapChildren());
-            mapUpdater.accept(map);
+            if (mapUpdater.apply(map) instanceof Reject error) {
+                return error;
+            }
```
taefi, line 213: Compiler complains about Java 17 compatibility of this expression, considering it to be an unconditional pattern matching `instanceof` which is supported in Java 21+:
```
/vaadin/flow/signals/src/main/java/com/vaadin/signals/impl/MutableTreeRevision.java:211:39
java: unconditional patterns in instanceof are not supported in -source 17
  (use -source 21 or higher to enable unconditional patterns in instanceof)
```
Same applies everywhere in this file where `(... instanceof Reject error)` is used.
Since the updated function is returning `CommandResult.Reject`, a null check is enough without `instanceof`.

Artur-:
Are you working on this or should we close it?

Legioth:
I'm just waiting for someone to review it...

signals/src/main/java/com/vaadin/signals/impl/MutableTreeRevision.java
taefi: Is this refactoring going to be back ported to V24 branches, or could the original code with the cleaner `instanceof` syntax be back?

Artur-: There is no need to backport

Issue #20983, opened 2025-02-12
BrowserCallable-aware package handling in Flow


mshabarov:
### Describe your motivation

Follow-up for https://github.com/vaadin/flow/issues/20289.

When Hilla's browser callable aka endpoints are present in a project, the Hilla TS endpoints generator packages have to be added.
Flow should know whether to add them during build time based on presence `BrowserCallable` / `Endpoint` annotated classes in a project.

### Describe the solution you'd like

These annotations are Hilla owned, so there should be a new API to pass this flag from Hilla.
It could be a method like `hasBrowserCallables()` in `EndpointGeneratorTaskFactory` interface that calls `BrowserCallableFinder` internally or Hilla can make `BrowserCallableFinder` discoverable through lookup for Flow.

We have to change the ordering of node tasks in Flow from
```
TaskUpdatePackages.class,
TaskRunNpmInstall.class,
TaskGenerateOpenAPI.class,
TaskGenerateEndpoint.class,
```

to 

```
TaskGenerateOpenAPI.class,
TaskUpdatePackages.class,
TaskRunNpmInstall.class,
TaskGenerateEndpoint.class,
```

(discover endpoints -> add generators if needed -> npm install -> generate endpoints if needed)

### Describe alternatives you've considered

Revert the https://github.com/vaadin/flow/issues/20289 and document the workaround.

### Additional context

Target version - 24.7

tepi:
Reverted due to the changes causing [test failures in Hilla](https://github.com/vaadin/hilla/pull/3255). For now, the workaround of adding a dummy view should be used, and the issue will be revisited in the future.

PR #21009, opened 2025-02-19
feat: improved endpoint detection

tepi:
Fixes https://github.com/vaadin/flow/issues/20983

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -73,9 +73,9 @@ public class NodeTasks implements FallibleCommand {
             TaskGenerateWebComponentBootstrap.class,
             TaskGenerateFeatureFlags.class,
             TaskInstallFrontendBuildPlugins.class,
+            TaskGenerateOpenAPI.class,
             TaskUpdatePackages.class,
             TaskRunNpmInstall.class,
-            TaskGenerateOpenAPI.class,
             TaskGenerateEndpoint.class,
             TaskCopyFrontendFiles.class,
             TaskCopyLocalFrontendFiles.class,
@@ -317,6 +317,9 @@ public class NodeTasks implements FallibleCommand {
 
     private void addEndpointServicesTasks(Options options) {
         if (!EndpointRequestUtil.isHillaAvailable(options.getClassFinder())) {
+            return;
+        }
+        if (!EndpointRequestUtil.areHillaEndpointsUsed(options)) {
             return;
         }
         Lookup lookup = options.getLookup();
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
@@ -319,22 +319,22 @@ public class NodeTasks implements FallibleCommand {
         if (!EndpointRequestUtil.isHillaAvailable(options.getClassFinder())) {
             return;
         }
-        if (!EndpointRequestUtil.areHillaEndpointsUsed(options)) {
-            return;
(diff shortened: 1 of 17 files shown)


tepi:
Needs to be tested with a Hilla build from [this PR branch](https://github.com/vaadin/hilla/pull/3255)

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -319,6 +319,9 @@ private void addEndpointServicesTasks(Options options) {
         if (!EndpointRequestUtil.isHillaAvailable(options.getClassFinder())) {
             return;
         }
+        if (!EndpointRequestUtil.areHillaEndpointsUsed(options)) {
```
mshabarov, line 322: What if views are there but no endpoints yet? 
I'd do it in this way:
1. If Hilla not available - return
2. if endpoints are used or views are used - continue, else - return.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
```diff
@@ -620,7 +621,8 @@ private ObjectNode generateVersionsFromPackageJson(JsonNode packageJson) {
     private void putHillaComponentsDependencies(
             Map<String, String> dependencies, String packageJsonKey) {
         if (FrontendUtils.isHillaUsed(options.getFrontendDirectory(),
-                options.getClassFinder())) {
+                options.getClassFinder())
+                || EndpointRequestUtil.areHillaEndpointsUsed(options)) {
```
mshabarov, line 625: Would it make sense to merge `areHillaEndpointsUsed` into `isHillaUsed` with OR for views and endpoints? This would clean up the codes a bit.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -559,6 +559,14 @@ public void failFastContextInitialized(ServletContextEvent event)
             collectHandleTypes(devModeHandlerManager.getHandlesTypes(),
                     annotations, superTypes);
 
+            EndpointGeneratorTaskFactory endpointGeneratorTaskFactory = lookup
+                    .lookup(EndpointGeneratorTaskFactory.class);
+
+            if (endpointGeneratorTaskFactory != null) {
+                annotations.addAll(endpointGeneratorTaskFactory
+                        .getBrowserCallableAnnotations());
+            }
```
mshabarov, line 568: This doesn't seem to be covered by tests, so needs a unit test.

mshabarov (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeUpdater.java
mshabarov: Let's not merge them, because `isHillaUsed` is bound to choosing hotdeploy and bundle modes. We don't want to force hotdeploy mode if only endpoints are present, bundle mode should be available for this case as well.

flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
tepi: Changed the logic

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
tepi: Added test

mshabarov:
Works as expected for me.
Need to figure out why tests are failing.

mshabarov (approved):

PR #21030, opened 2025-02-21
chore: Update example feature flag description

mshabarov:


mcollovati:
Actually, the example feature is almost used to test the functionalities related to feature flags.
If we remove it, we have to introduce dirty reflection hacks in the test code, making the tests less reliable.

I propose to preserve the feature, and update its description to state that it is there for testing purpose.

```
    public static final Feature EXAMPLE = new Feature(
            "Example feature. Internally used for testing purposes. Does not have any effect on production applications.",
            "exampleFeatureFlag", "https://github.com/vaadin/flow/pull/12004",
            false,
            "com.vaadin.flow.server.frontend.NodeTestComponents$ExampleExperimentalComponent");
```

mcollovati (dismissed):

mcollovati (approved):

PR #21039, opened 2025-02-24
chore: Upgrade target to es2022

Artur-:
The combination of TS target es2022 and Vite build target including `safari15` keeps the output compatible with Safari 15. It also enables using all features added in es2021 and es2022.


flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig-osgi.json
```diff
@@ -3,7 +3,7 @@
     "sourceMap": true,
     "inlineSources": true,
     "module": "esNext",
-    "target": "es2020",
+    "target": "es2022",
```
mshabarov, line 6: The change for "osgi" version is not needed as this is one of the old version that is used just for changes detection.

flow-server/src/main/resources/com/vaadin/flow/server/frontend/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "jsx": "react-jsx",
     "inlineSources": true,
     "module": "esNext",
-    "target": "es2020",
+    "target": "es2022",
```
mshabarov, line 13: If I recall correctly, once the template is changed, we need to create a new template with latest content, store the current file as old file and add some version suffix in `TaskGenerateTsConfig::tsconfigVersions`.

mshabarov (changes requested):

mshabarov (approved):

PR #21041, opened 2025-02-25
fix: fix json decode for null value

caalador:


flow-server/src/main/java/com/vaadin/flow/component/ComponentEventBus.java
```diff
@@ -381,8 +382,12 @@ private List<Object> createEventDataObjects(DomEvent domEvent,
                 Object value;
                 if (JsonValue.class.isAssignableFrom(type)) {
                     // TODO: Remove after History uses Jackson.
-                    value = JsonCodec.decodeAs(
-                            domEvent.getEventData().get(expression), type);
+                    if (jsonValue == null) {
```
mcollovati, line 385: I don't understand how `jsonValue` can be null here.
A couple of lines above, there's another null check for `jsonValue`, and if it is null, it gets assigned to `JacksonUtils.nullNode();`, so if I'm not missing something, this condition should never happen.

caalador: Also it's not even null in any moment as mapElemental for a null value will return a NullNode.

mcollovati (approved):

mcollovati:
@caalador can we close this PR?

PR #21064, opened 2025-02-28
feat: Internals to jackson

caalador:
Change internal functions to
use jackson.

part of #20741

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
@@ -19,13 +19,16 @@ package com.vaadin.flow.component.webcomponent;
 import java.io.Serializable;
 import java.util.Objects;
 
+import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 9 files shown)


flow-server/src/main/java/com/vaadin/flow/component/webcomponent/WebComponent.java
```diff
@@ -222,6 +265,12 @@ private void setProperty(String propertyName, Object value) {
         } else if (value instanceof Boolean) {
             componentHost.executeJs(UPDATE_PROPERTY, propertyName,
                     (Boolean) value);
+        } else if (value instanceof ValueNode) {
+            // this gets around executeJavaScript limitation.
+            // Since properties can take JsonValues, this was needed to allow
```
mcollovati, line 270: Should `JsonValues` be replaced with `ValueNode`? Or maybe just `JSON values`?

caalador: reworded both comments to JSON values

mcollovati (approved):

PR #21080, opened 2025-03-05
fix: constant pool to jackson

caalador:
Use jackson in constant pool
UidlWriter to handle response
in jackson.

Part of #20741

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
@@ -105,11 +105,12 @@ public final class JacksonUtils {
     }
 
     /**
-     * Map JsonObject to ObjectNode.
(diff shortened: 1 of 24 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/JacksonUtils.java
```diff
@@ -111,6 +111,21 @@ public static ValueNode nullNode() {
      *            JsonObject to change
      * @return ObjectNode of elemental json object or null for null jsonObject
      */
+    public static ArrayNode mapElemental(JsonArray jsonObject) {
+        try {
```
mcollovati, line 115: null check needed?

caalador: Added.

mcollovati (approved):

PR #21083, opened 2025-03-05
fix: fix equality check in NavigationAccessControl

### LINKED ISSUE #21082: NavigationAcessControl compares strings with equality operator
## The `NavigationAccessControl` class uses the equality operator to compare the URLs of the login view.

The Control Center Starter uses an application property to set the login view of Control Center [to the Keycloak login view]. When upgrading from one version of Control Center to a newer version, this application property is read again by the configuration class in the newly deployed version, but fails with an exception because of this check, **even though the property has the same value as before**.

https://github.com/vaadin/flow/blob/main/flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessControl.java#L218-L226

### Expected behavior

Upgrading a Control Center instance should not fail when the login URL has not changed between versions.

## In a simple Vaadin app with security enabled

1. Have a property class with two different properties.
2. Set the value of both properties in `application.properties` or `application.yml` to the same value.
3. In a configuration class (extending `VaadinWebSecurity`), load the properties and set the login URL of the Vaadin app using one and then the other with the `setOAuth2LoginPage` method. It should not fail on the first one, because it has not been set before. It should not fail on the second one, because it has the same value as the first one. 
Image

Image

Image

### Versions

- Vaadin / Flow version: 24.6
- Java version: 21
- OS version: macOS 15.3.1

mcollovati:
Fixes #21082

Co-authored-by: David Sosa (@sosa-vaadin)

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessControl.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/NavigationAccessControl.java
@@ -216,7 +216,7 @@ public class NavigationAccessControl implements BeforeEnterListener {
      *            the frontend view to use as login view
      */
     public void setLoginView(String loginUrl) {
-        if (loginUrl == this.loginUrl) {
+        if (Objects.equals(loginUrl, this.loginUrl)) {
             // Probably hot reload
             return;
         }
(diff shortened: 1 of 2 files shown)


sosa-vaadin:
Test case:
```java
    @Test
    public void setLoginViewStringShouldNotThrowWithSameString() {
        accessControl.setLoginView("/foo");
        accessControl.setLoginView("/foo");
        accessControl.setLoginView(new String("/foo"));
        Assert.assertEquals("/foo", accessControl.getLoginUrl());
    }
```

sosa-vaadin (approved):

tepi (approved):

Issue #20992, opened 2025-02-13
frontendHotploy isn't working for a Vaadin multi-module gradle project


nico-fri:
### Description of the bug

I have a Vaadin 24 multi-module gradle project where I want to use the frontendHotdeploy so that I don't have to restart the server every time I make a css or js change. Unfortunately the hotdeploy does not work in this constellation. 

In the Vaadin 24 gradle starter template, the frontend folder is under src/main/frontend. In a multi-module project it must be under src/main/resources/META-INF/resources/frontend. Otherwise, an error message will appear at startup stating that the files cannot be found. If the frontendHotploy flag is now set to true in build.gradle, it no longer works

### Expected behavior

The hotreload of css and js works so that the server does not have to be constantly restarted

### Minimal reproducible example

Two Vaadin 24 gradle starter projects where one is added as a dependency via build.gradle.

**Project 1:**
- settings.gradle: rootProject.name = 'project-components'

**Project 2:**
- settings.gradle: 
rootProject.name = 'project-main'
include(":project-components")
project(":project-components").projectDir = new File("../project-components")

- build.gradle
vaadin {
    frontendHotdeploy = true;
}
dependencies {
    implementation(project(":project-components"))
}

### Versions

- Vaadin / Flow version: 24
- Java version: 17
- OS version: MacOS 15.3
- Browser version (if applicable): Safari
- Application Server (if applicable): Wildfly V35
- IDE (if applicable): Intellij

knoobie:
Did you try this property? At least in Maven it works.

```xml
  <properties>
    <vaadin.frontend.hotdeploy.dependencies>../moduleB</vaadin.frontend.hotdeploy.dependencies>
  </properties>
```

nico-fri:
The property frontendHotdeployDependencies is not available in a gradle project

> Did you try this property? At least in Maven it works.
> 
>   <properties>
>     <vaadin.frontend.hotdeploy.dependencies>../moduleB</vaadin.frontend.hotdeploy.dependencies>
>   </properties>

mshabarov:
Additional observations:

> I just tried the system property -Dvaadin.frontend.hotdeploy.dependencies=“../project-components” and the `frontendHotdeploy` worked for project-components, but not for the project-main. I tried to include the project-main additionally via `:` or `;`. However, this did not work =”../project-components;../project-main”

We have to test and be sure that this property works consistently.

tepi:
[...]

PR #21099, opened 2025-03-07
fix: watch active project resources

tepi:
Watch resources from active project if active project defined in hotdeployDependencyFolders.

If hotdeployDependencyFolders is not set, add active project to watched set of directories.

Fixes https://github.com/vaadin/flow/issues/20992

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/ExternalDependencyWatcher.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/ExternalDependencyWatcher.java
@@ -57,6 +57,9 @@ public class ExternalDependencyWatcher implements Closeable {
                 }
             }
         } else {
+            // Always watch src/main/resources/META-INF from active project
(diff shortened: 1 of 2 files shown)


mshabarov:
A general comment, could we have a test for it? Hopefully this shouldn't be hard as changing the file in meta inf is no harder than in any other directory, so can be even a part of existing test.

tepi:
Good point. I'll check the existing tests for frontend hotdeploy and see how this can be squeezed in.

mshabarov (approved):
Works like a charm, tested with the [maven multi-module example](https://github.com/vaadin/multi-module-example/) and:
1. Having `vaadin-app/src/main/resources/META-INF/resources/frontend/foo.css` and `backend/src/main/resources/META-INF/resources/frontend/bar.css` files imported as
```
@CssImport("./foo.css")
@CssImport("./bar.css")
public class HelloWorldView
``` 
2. System property `-Dvaadin.frontend.hotdeploy.dependencies="./, ../backend"`.

HMR works for both of these files.

PR #21103, opened 2025-03-07
fix: prevent duplicated messages in client queue

### LINKED ISSUE #21095: Connection recovery does not work when a request is terminated while it's being processed
### Description of the bug

The client does not recover properly if the HTTP connection is terminated while a request is processed on the server. The client manages to apply the changes from the original request when connectivity is restored but it also shows an Internal error notification and keeps retrying every five seconds. Each retry leads to a "Unexpected message id from the client. Expected client id: 4, got 3" error in the server console.

Image

### Expected behavior

Expected that there's no Internal error and no retries.

1. Create application at https://start.vaadin.com/ with just a Hello World view for Flow. Select Vaadin 24.7 before downloading since Vaadin 24.6 is the default at the time of writing.
2. Edit the click handler in `HelloWorldView` to have a 5-second delay (so that you will have time to kill the connection before the response is sent)
3. Run in production mode with `mvn -Pproduction package; java -jar target/my-app-1.0-SNAPSHOT.jar`
4. Open a local SSH connection that forwards port 8081 to 8080. `ssh -L 8081:localhost:8080 localhost`
5. Open http://localhost:8081/ in the browser
6. Click the button and then immediately terminate the SSH connection using `^d ^c` (only  `^d` would make it wait until all forwarded connections are closed)
7. Observe that the black offline status bar shows up immediately in the browser
8. Open the SSH connection again
9. Observe that the black offline bar changes to a blue online bar after a couple of seconds.
10. Observe that the expected notification is shown in the lower left corner
11. Observe that there's also an Internal error notification in the upper right corner
12. Observe that there's a new "Unexpected message id from the client. Expected client id: 4, got 3." message in the server console every 5 seconds until you reload in the browser or close the tab

### Versions

- Vaadin / Flow version: Vaadin 24.7.0.beta1
- Java version: Temurin-21.0.4+7
- OS version: macOS 15.3.1
- Browser version (if applicable): Chrome 133. Also tested with the latest versions of Firefox and Safari.
- Application Server (if applicable): Tomcat 10.1.36 through Spring Boot
- IDE (if applicable): None, running in production mode from the command line

mcollovati:
Messages sent by the Flow client are queued until the server acknowledges their reception. However, if there's a network fault, the Flow client immediately attempts to resend the message, resulting in duplicates in the queue.

This change ensures that the same message is not added to the queue multiple times.

Fixes #21095

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/communication/MessageSender.java
+++ b/flow-client/src/main/java/com/vaadin/client/communication/MessageSender.java
@@ -20,7 +20,6 @@ import java.util.List;
 
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.user.client.Timer;
-
 import com.vaadin.client.ConnectionIndicator;
 import com.vaadin.client.Console;
 import com.vaadin.client.Registry;
@@ -208,8 +207,10 @@ public class MessageSender {
      */
     public void send(final JsonObject payload) {
         if (hasQueuedMessages()) {
-            if (messageQueue.stream().noneMatch(
-                    message -> message.toJson().equals(payload.toJson()))) {
+            // The sever sync id is set in the private sendPayload method.
+            // If it is already present on the payload, it means the message has
+            // been already sent and enqueued.
+            if (!payload.hasKey(ApplicationConstants.SERVER_SYNC_ID)) {
                 messageQueue.add(payload);
             }
             return;
--- a/flow-client/src/main/java/com/vaadin/client/communication/MessageSender.java
+++ b/flow-client/src/main/java/com/vaadin/client/communication/MessageSender.java
@@ -208,7 +208,10 @@ public class MessageSender {
      */
     public void send(final JsonObject payload) {
         if (hasQueuedMessages()) {
-            messageQueue.add(payload);
+            if (messageQueue.stream().noneMatch(
+                    message -> message.toJson().equals(payload.toJson()))) {
+                messageQueue.add(payload);
+            }
             return;
         }
         messageQueue.add(payload);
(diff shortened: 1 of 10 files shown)


flow-client/src/main/java/com/vaadin/client/communication/MessageSender.java
```diff
@@ -208,7 +208,10 @@ private JsonObject preparePayload(final JsonArray reqInvocations,
      */
     public void send(final JsonObject payload) {
         if (hasQueuedMessages()) {
-            messageQueue.add(payload);
+            if (messageQueue.stream().noneMatch(
```
knoobie, line 211: nit: with this being a crucial hot path would it make sense to use the "old school" for loop to ensure no unnecessary stream / overhead is created?

mcollovati: Thanks for pointing out. 
This is converted into Javascript by GWT. I don't know if using a for loop will change performance at runtime.
I will double check on Monday

mcollovati: For the reviewers: I wonder if a more performant check would be to just ignore messages that already have a server sync id. For new messages it should be assigned only in the sendPayload method, so if the key is present it means the message is already queued

caalador: It would probably be enough to see if there is a syncId already as that would mean that it is already sent.

mcollovati: Thanks. Updated the check

caalador (approved):

Issue #21016, opened 2025-02-20
vaadin-maven-plugin scans twice


jorgheymans:
### Description of the bug

A fresh app from start.vaadin.com gives this output:

```
mvn clean install -Pproduction
....
[INFO] >>> vaadin:24.6.5:prepare-frontend (default) > :configure @ my-app >>>
[INFO] 
[INFO] --- vaadin:24.6.5:configure (configure) @ my-app ---
[INFO] 
[INFO] <<< vaadin:24.6.5:prepare-frontend (default) < :configure @ my-app <<<
[INFO] 
[INFO] 
[INFO] --- vaadin:24.6.5:prepare-frontend (default) @ my-app ---
[INFO] Reflections took 1605 ms to scan 200 urls, producing 11676 keys and 60256 values
[INFO] The globally installed Node.js version 14.18.0 is older than the required minimum version 18.12.0. Using Node.js from /home/user/.vaadin.
[INFO] Visited 115 classes. Took 150 ms.
[INFO] Copying frontend resources from jar files ...
[INFO] Visited 178 resources. Took 143 ms.
[INFO] 
[INFO] >>> vaadin:24.6.5:build-frontend (default) > :configure @ my-app >>>
[INFO] 
[INFO] --- vaadin:24.6.5:configure (configure) @ my-app ---
[INFO] 
[INFO] <<< vaadin:24.6.5:build-frontend (default) < :configure @ my-app <<<
[INFO] 
[INFO] 
[INFO] --- vaadin:24.6.5:build-frontend (default) @ my-app ---
**[INFO] Scanning classes to find frontend configurations and dependencies...**
**[INFO] Visited 6509 classes. Took 3874 ms.**
[INFO] Checking if a production mode bundle build is needed
[INFO] A production mode bundle build is not needed
[INFO] Copying frontend resources from jar files ...
[INFO] Visited 178 resources. Took 76 ms.
**[INFO] Scanning classes to find frontend configurations and dependencies...**
**[INFO] Visited 6509 classes. Took 2704 ms.**
[INFO] Build frontend completed in 7080 ms.

```

-->
`**[INFO] Visited 6509 classes. Took 3874 ms.**`
`**[INFO] Visited 6509 classes. Took 2704 ms.**`

This was addressed in https://github.com/vaadin/flow/issues/19874 , but it seems to have come back?

### Expected behavior

Class scanning occurs only once.

### Minimal reproducible example

[...]

PR #21104, opened 2025-03-07
feat: use single scanner for plugin class scan

tepi:
Fixes https://github.com/vaadin/flow/issues/21016


flow-server/src/main/java/com/vaadin/flow/server/frontend/NodeTasks.java
```diff
@@ -96,20 +96,30 @@ public class NodeTasks implements FallibleCommand {
 
     private Path lockFile;
 
+    public NodeTasks(Options options) {
+        this(options,
```
caalador, line 100: This scanner builder doesn't check the router required.
`boolean reactEnabled = options.isReactEnabled() && FrontendUtils.isReactRouterRequired(options.getFrontendDirectory());`

PR #21108, opened 2025-03-11
feat: add defaultAutoResponsiveFormLayout feature flag

vursen:
## Description

The PR introduces a new feature flag, `defaultAutoResponsiveFormLayout`, which enables auto-responsive mode for all form layouts by default.

Part of https://github.com/vaadin/flow-components/issues/7164

## Type of change

- [x] Feature

### DIFF
--- a/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
+++ b/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
@@ -109,6 +109,12 @@ public class FeatureFlags implements Serializable {
             "https://github.com/vaadin/flow-components/issues/6998", true,
             null);
 
+    public static final Feature DEFAULT_AUTO_RESPONSIVE_FORM_LAYOUT = new Feature(
+        "Form Layout auto-responsive mode enabled by default",
+        "defaultAutoResponsiveFormLayout",
+        "https://github.com/vaadin/platform/issues/7172", true,
+        null);
+
     private List<Feature> features = new ArrayList<>();
 
     File propertiesFolder = null;
@@ -140,6 +146,7 @@ public class FeatureFlags implements Serializable {
         features.add(new Feature(REACT19));
         features.add(new Feature(ACCESSIBLE_DISABLED_BUTTONS));
         features.add(new Feature(LAYOUT_COMPONENT_IMPROVEMENTS));
+        features.add(new Feature(DEFAULT_AUTO_RESPONSIVE_FORM_LAYOUT));
         loadProperties();
     }
 
--- a/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
+++ b/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
@@ -110,10 +110,9 @@ public class FeatureFlags implements Serializable {
             null);
 
     public static final Feature DEFAULT_AUTO_RESPONSIVE_FORM_LAYOUT = new Feature(
-        "Form Layout auto-responsive mode enabled by default",
-        "defaultAutoResponsiveFormLayout",
-        "https://github.com/vaadin/platform/issues/7172", true,
-        null);
+            "Form Layout auto-responsive mode enabled by default",
+            "defaultAutoResponsiveFormLayout",
+            "https://github.com/vaadin/platform/issues/7172", true, null);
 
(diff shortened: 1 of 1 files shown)


flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java
```diff
@@ -109,6 +109,11 @@ public class FeatureFlags implements Serializable {
             "https://github.com/vaadin/flow-components/issues/6998", true,
             null);
 
+    public static final Feature DEFAULT_AUTO_RESPONSIVE_FORM_LAYOUT = new Feature(
+            "Form Layout auto-responsive mode enabled by default",
```
mshabarov, line 113: It's not clear from the flag name, what's this flag for - to enable this mode or to disable?

If it's for enabling this mode, I'd then remove "default" from the names.

If it's about disabling this mode, which is new, I'd give another way to disable it, e.g. through annotation, overriding a method or config parameter.

rolfsmeds: The flag makes all FormLayouts use autoResponsive mode by default, which is how it will behave in V25. Without the flag, you need to explicitly switch to autoResponsive mode on each FormLayout instance.

What would you name it?

mshabarov: Thanks for the clarifications, then I'd name it just as "AUTO_RESPONSIVE_FORM_LAYOUT" and "Form Layout auto-responsive mode", "AutoResponsiveFormLayout" (excluding all "default" prefixes).

However, do you want this flag to be ON by default in V25, right? 

Would it be better to swap the meaning, like "Disable Form Layout auto-responsive mode", so that if you want to go back to old mode from V24, you toggle the flag?

rolfsmeds: The plan is to replace the flag with the opposite in V25 (something like `DEFAULT_RESPONSIVE_STEPS_FORM_LAYOUT` if we go by the currently proposed naming), rather than have this one be on by default.

vursen: To me, without additional context, `autoResponsiveFormLayout` gives the impression that this flag _must_ be enabled to use the auto-responsive feature in general, which isn't true because the feature is available anyway. The flag only determines whether it should be enabled _by default_. Personally, I would name it `enableAutoResponsiveFormLayoutByDefault`, but it is a bit long :)

rolfsmeds: `FormLayoutAutoResponsiveByDefault` would be fine IMO. 

@web-padawan suggested using the term "global" instead of "default", but I still feel that "default" makes sense as part of the name as the flag changes the _default_ mode in FormLayout.

mshabarov: We discussed this with @vursen on a call.
The "default" word caused question from me, because I thought that we have to add this feature flag ON for all Vaadin 24.8 projects. This is false. We want to add it only to the new 24.8 projects, which makes sense. But in my recollection this hasn't been a practice ever with feature flags. Usually, we add a feature flag for two cases:
1. When we want to claim that the feature and API may change
2. When we don't want to change the behaviour and make it default.

This feature flag basically pursues 2., but with addition that we want to add it in start.vaadin.com.
All in all, I'm ready to approve, the naming looks good to me now after explanations, but still we need to think what is the best way:
1. Add FF now in 24.8 and add it to new projects only in start.vaadin.com.
2. Not add it now but only add in 25, so users may revert to older mode.
3. Do 2. but in 24.8.

rolfsmeds: I'd say we don't necessarily have to have the flag _on by default_ in SVC-generated projects in 24.8, but it would be good to _have_ a flag so that user can already in 24.8 switch to the V25-behavior and give feedback about potential issues with it. And of course we don't have to decide on the SVC question here and now.

So I would say let's just implement the flag in 24.8 and not do anything else for now.

mshabarov: Yes, I agree, let's apply it.

mshabarov (approved):

PR #21116, opened 2025-03-13
chore: Explicitly set resources encoding in Maven resources plugin

edler-san:
To avoid maven warnings and be compatible with the upcoming 4.x release.


pom.xml
```diff
@@ -671,6 +671,9 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-resources-plugin</artifactId>
                     <version>3.3.1</version>
+                    <configuration>
+                        <propertiesEncoding>UTF-8</propertiesEncoding>
```
mcollovati, line 675: nit: does it make sense to use the `project.build.sourceEncoding` property?

edler-san: That is even the recommended way: https://maven.apache.org/plugins/maven-resources-plugin/examples/encoding.html

edler-san: Addendum:
I wanted to include this because without it set I get the following info when building flow under Windows/Java 17:
```
[INFO] --- resources:3.3.1:resources (default-resources) @ flow-server ---
[INFO] Copying 1 resource from src\main\resources to target\classes
[INFO] Copying 75 resources from src\main\resources to target\classes
[INFO] The encoding used to copy filtered properties files have not been set. This means that the same encoding will be used to copy filtered properties files as when copying other filtered resources. This might not be what you want! Run your build with --debug to see which files might be affected. Read more at https://maven.apache.org/plugins/maven-resources-plugin/examples/filtering-properties-files.html
```
Going through the link it seems `ISO-8859-1` might even be better here.

mcollovati: The linked page also says 

> Starting with Java 9 the preferred encoding is UTF-8 for property resource bundles.

So, I would go with UTF-8.
Would you like to use the property, or do you prefer to directly set the value?

edler-san: The property is already set in the pom. So to avoid the nagging, I'd go with the value (also, I don't expect this to change in the foreseeable future).

mcollovati: Ok. Let's merge it as is

mcollovati (approved):

PR #21119, opened 2025-03-13
fix: add allowInert support for synchronized properties

tepi:
Fixes https://github.com/vaadin/flow/issues/21085


flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/AbstractRpcInvocationHandler.java
```diff
@@ -61,6 +62,22 @@ public Optional<Runnable> handle(UI ui, JsonObject invocationJson) {
             return Optional.empty();
         }
 
+        // Allow handling of RPC request if any listener for the event type or
+        // the synchronized property have enabled allowInert.
+        if (node.isInert() && node.hasFeature(ElementListenerMap.class)) {
```
caalador, line 67: This should probably not be put before the isInactive check. Now we approve handling for inactive nodes that are also inert.

I think we should have the property check only in `MapSyncRpcHandler.allowInert` override.
Event check is already in ElementListenerMap as `EventRpcHanlder` always allows inert and the actual acceptance check is in the listener map.

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/internal/nodefeature/ElementListenerMap.java
```diff
@@ -512,4 +512,21 @@ public DisabledUpdateMode getPropertySynchronizationMode(
                 .reduce(DisabledUpdateMode::mostPermissive).orElse(null);
     }
 
+    public boolean hasAllowInertForType(String eventType) {
```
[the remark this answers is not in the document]
tepi: Removed this method

caalador (approved):
If possible a test case would be that checks that a opened synchornize change does actually get handled e.g. before this change a closed on client doesn't propagate to server, but after this is should be.

Possibly simplest would be to use native details and for instance the modal dialog mock from com.vaadin.flow.uitest.ui.ModalDialogView

flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/AllowInertSynchronizedPropertyView.java
```diff
@@ -0,0 +1,98 @@
...
+        button.addClickListener(listener);
+        button.getStyle().set("border", "1px solid black");
+        button.setWidth("100px");
+        return button;
+    }
+
+    public class Dialog extends Div {
+
+        public Dialog() {
+            final NativeButton toggleNativeDetailsButton = new NativeButton(
```
caalador, line 77: Nit: the button name states it toggles when it actually only reads the state.

tepi: Fixed name

PR #21193, opened 2025-03-28
fix: align npm package order between full and annotation scanners

platosha:
Using `HashSet` for the `@NpmPackage` annotation values results in unreliable order. The issue is visible in [PR validation ITs in Hilla](https://github.com/vaadin/hilla/actions/runs/14127153546/job/39578829883?pr=3374#step:7:2189), `build-frontend` has unexpected `alpha5` version picked over before `alpha4`, whereas `prepare-frontend` is using `alpha4` chosen by  FQN sorted class order.


taefi (approved):

tepi:
Comment from @mcollovati on Slack:

> If the issue is a different ordering between `prepare-frontend` and `build-frontend`, I'm not 100% confident the fix will solve the problem in all cases.
Which version is the first depends also on the order the classes are visited by `FrontendAnnotatedClassVisitor`
```
        for (Class<?> component : getFinder()
                .getAnnotatedClasses(NpmPackage.class.getName())) {
            npmPackageVisitor.visitClass(component.getName());
        }
``` 
Are we sure that the `Set<Class<?>> etAnnotatedClasses(...)` method has a stable order? Or should we do here some kind of sorting?
Also to be noted that, looking at the build logs, it looks like `prepare-frontend` and `build-frontend` are executed in different maven phases, one before compile and the other after so two different `ClassFinder` instance are in use
But most likely my concerns have been already verified.

platosha:
- `ClassFinder#getAnnotatedClasses(String)` delegates to `ClassFinder.getAnnotatedClasses(Class<? extends Annotation>)` from default implementation
- `ClassFinder#getAnnotatedClasses(Class<? extends Annotation>)` has 4 implementations:
  1.  `CachedClassFinder` does not interfere with the order, it's only a cache for another class finder
  2. `DefaultClassFinder` has stable order derived from `classes` LinkedHashSet, which is populated from FQN-sorted classes, and also using LinkedHashSet for result:
  ```java
          @Override
        public Set<Class<?>> getAnnotatedClasses(
                Class<? extends Annotation> annotation) {
            return classes.stream().filter(
                    cl -> cl.getAnnotationsByType(annotation).length > 0)
                    .collect(Collectors.toCollection(LinkedHashSet::new));
        }
  ```
  3. `DevModeInitializer` is delegating to superclass `DefaultClassFinder`
  4. `ReflectionsClassFinder` is also using LinkedHashSet and FQN-sorting:
  ```java
      @Override
    public Set<Class<?>> getAnnotatedClasses(
            Class<? extends Annotation> clazz) {
        Set<Class<?>> classes = new LinkedHashSet<>();
        classes.addAll(reflections.getTypesAnnotatedWith(clazz, true));
        classes.addAll(getAnnotatedByRepeatedAnnotation(clazz));
        return sortedByClassName(classes);

    }
    ```
 
This confirms what I saw in the debugger: the classes are name-sorted elsewhere, including `data` field in the visitor.

mshabarov (approved):

PR #21217, opened 2025-04-03
fix: Compliant telemetry

AB-xdev:
## Description

PoC showcasing how a better way of handling telemetry can be provided.

1. There is now a GLOBAL opt out in the form of the ``VAADIN_TELEMETRY_OPT_OUT`` environment variable
2. When executed for the first time it discloses the telemetry collection to the user (avoids potential violation of [GDPR Art 13](https://gdpr-info.eu/art-13-gdpr/)).

Overall the whole thing is quite similar to how [Dotnet](https://learn.microsoft.com/en-us/dotnet/core/tools/telemetry) handles telemetry.

Please note that there is currently a placeholder for a site with more details about telemetry in place, as this can't be provided by the PR.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [ ] New and existing tests are passing locally with my change.<br/> - I can't verify that because I don't have all tools required for running tests locally installed and they are not in place in the repo.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/Constants.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/Constants.java
@@ -149,10 +149,13 @@ public final class Constants implements Serializable {
      */
     public static final boolean DEFAULT_REQUIRE_HOME_NODE_EXECUTABLE = false;
 
+    public static final String TELEMETRY_OPT_OUT_ENV_VARIABLE = "VAADIN_TELEMETRY_OPT_OUT";
+
     /**
      * The default value for whether usage statistics is enabled.
      */
-    public static final boolean DEFAULT_DEVMODE_STATS = true;
+    public static final boolean DEFAULT_DEVMODE_STATS =
+            System.getenv(Constants.TELEMETRY_OPT_OUT_ENV_VARIABLE) == null;
 
     /**
      * Internal parameter which prevent validation for annotations which are
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/DevModeUsageStatistics.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/DevModeUsageStatistics.java
@@ -108,7 +108,7 @@ public class DevModeUsageStatistics {
                 Files.createDirectories(statisticDirPath);
                 Files.writeString(firstSeenPath, Instant.now().toString());
             } catch (IOException ioe) {
-                getLogger().warn("Failed to create statistics first seen file", ioe);
+                getLogger().warn("Failed to create telemetry notice first seen file", ioe);
             }
         }
 
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/DevModeUsageStatistics.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/stats/DevModeUsageStatistics.java
@@ -17,8 +17,14 @@
 package com.vaadin.base.devserver.stats;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.Instant;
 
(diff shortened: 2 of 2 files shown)


jojule (changes requested):
Intent here is driven by GDPR compliance. The telemetry collected does not include any Personal Data and thus GDPR does not apply for anonymous users.

Users who have a license key have explicitly approved usage terms and Personal Data usage terms there.

Requesting to drop additional log info here, but supporting addig a global telemetry opt-out environment variable.

mshabarov:
I agree that the Vaadin statistics collector is not a case for GDPR. It doesn't add any extra to what license checking process already uses (machine id, user keys).

Having the env variable for disable it makes sense to me, however I'd keep one single name for it (or at least similar), i.e. reuse existing `vaadin.devmode.usageStatistics.enabled` in upper case. 

Having one time logging looks acceptable to me with the current mild wording.

We have a short mention in the [online docs](https://vaadin.com/docs/latest/flow/configuration/properties#properties) about the existing property, would be good, as a pre-requisite for this, to add a dedicated short section in "development" article about statistics.

Finally, keep in mind that there is more stats collection in the browser controlled by https://github.com/vaadin/vaadin-usage-statistics, it needs a separate opt-out.

AB-xdev:
Sorry for the late response, the last months were quite busy.

> Intent here is driven by GDPR compliance

This is not limited to only GDPR.
For example AFAIK my company's contracts with customers usually contain a no-spy clause.
Having some badly documented, intransparent and not globally deactivatable telemetry traffic is kind of problematic.

> The telemetry collected does not include any Personal Data and thus the GDPR does not apply.
> ...
> Requesting to drop additional log info here

Although I'm not 100% sure about the legal aspects here, note that nearly every other major company or product that is collecting telemetry data - even without "personal data" - performs a disclosure on the first run. Examples:
* https://learn.microsoft.com/en-us/dotnet/core/tools/telemetry
* https://quarkus.io/usage/
* https://code.visualstudio.com/docs/configure/telemetry
  > To ensure GDPR compliance, we made several updates to VS Code, these include:
  > * Making it easier to opt out of telemetry collection by placing a notification in product for all existing and new users.
  > * Reviewing and classifying the telemetry that we send (documented in [our OSS codebase](https://github.com/microsoft/vscode/pull/34997)).
  > * Ensuring that we have valid data retention policies in place for any data we do collect, for example crash dumps.
* https://www.jetbrains.com/help/idea/settings-usage-statistics.html#data_protection_laws_compliance
  > Given our adherence to these principles, the General Data Protection Regulation (GDPR) does not apply to feature usage statistics. However, in alignment with the [ePrivacy directive](https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A02002L0058-20091219), we ask for the user’s consent for data collection, even when the information collected is anonymized.

If you still want to remove the first-run telemetry disclosure feel free to do this on the branch, I granted maintainers write access.


> Having the env variable for disable it makes sense to me, however I'd keep one single name for it (or at least similar), i.e. reuse existing `vaadin.devmode.usageStatistics.enabled` in upper case.

You can also change that if you want however I would prefer a simple environment variable that is not that long and won't change because someone e.g. renames devmode in the future.


> Finally, keep in mind that there is more stats collection in the browser controlled by https://github.com/vaadin/vaadin-usage-statistics, it needs a separate opt-out.

+1
Feel free to provide a additional opt-out functionality for that as the PR is currently scoped to Vaadin Flow Java code only.

mshabarov:
Hi @AB-xdev , thanks for your patience and sorry for a long response: we eventually passed through the vacation season and completed 24.9 preparations. I raised a thread internally about this topic and seems like, taking your and your customers motivations, we can apply this. 
Before merging, I'd make an addition to https://vaadin.com/docs/latest/flow/configuration/properties to mention the parameter and statistics in general, then we can put a link into PR. 
Also I'll double check how to align the existing parameter name with the new one. 
I'll take over the PR and finalise it. Let me know if you have any further thoughts.

tltv (approved):

PR #21304, opened 2025-04-16
Exclude trees without changes from the result collector

Legioth:
If a tree was only read but not written inside a transaction, then the transaction result was never published. This was because the result collector would wait for a result from the unchanged tree but that tree wasn't involved in the actual two-phase commit procedure and thus never published any result. This is fixed by creating the result collector based on the trees that participate in the two-phase commit.

### DIFF
--- a/signals/src/main/java/com/vaadin/signals/impl/StagedTransaction.java
+++ b/signals/src/main/java/com/vaadin/signals/impl/StagedTransaction.java
@@ -135,15 +135,14 @@ public class StagedTransaction extends Transaction {
             return;
         }
 
-        ResultCollector collector = new ResultCollector(openTrees.keySet(),
-                resultHandler);
-
         if (outer instanceof StagedTransaction outerTx) {
+            ResultCollector collector = new ResultCollector(openTrees.keySet(),
+                    resultHandler);
             for (SignalTree tree : openTrees.keySet()) {
                 outerTx.include(tree, createChange(tree, collector), true);
             }
         } else {
-            commitTwoPhase(collector);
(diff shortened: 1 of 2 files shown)


signals/src/main/java/com/vaadin/signals/impl/StagedTransaction.java
```diff
@@ -156,7 +155,7 @@ protected void commit(Consumer<ResultOrError<Void>> resultHandler) {
         }
     }
 
-    private void commitTwoPhase(ResultCollector collector) {
+    private void commitTwoPhase(Consumer<ResultOrError<Void>> resultHandler) {
```
mshabarov, line 158: Not sure which is mode safe and better, but maybe we can keep `ResultCollector` as an argument and do `ResultCollector collector = new ResultCollector(trees, resultHandler);` along with `commitTwoPhase(collector);` ?

mshabarov: `commitTwoPhase(ResultCollector collector)` has better readability than `commitTwoPhase(Consumer<ResultOrError<Void>> resultHandler)`, but this is very minor.

mshabarov (approved):

signals/src/main/java/com/vaadin/signals/impl/StagedTransaction.java
Legioth: I don't see how that would work without additional changes since `trees` is now computed inside `commitTwoPhase`.

PR #21305, opened 2025-04-16
fix: prevent duplicated history entry with vaadin-router

### LINKED ISSUE #19494: [Placeholder] NavigationTriggerIT::testNavigationTriggers fails
### Description of the bug

NavigationTriggerIT.testNavigationTriggers:35->assertMessageCount:83 expected:<1> but was:<2>:

Navigation with RouterLink triggers an extra `HISTORY` event, which is sent to server and makes navigation lifecycle hooks, like `setParameter` to be called one more time. 

Note: This only happens if page URL has trailing slash, e.g. `localhost:8080/view/foo/`. Extra history event isn't trigered if initial URL is `localhost:8080/view/foo`

### Expected behavior

.

.

### Versions

- Vaadin / Flow version:
- Java version:
- OS version:
- IDE (if applicable):

### LINKED ISSUE #21243: Disabling react causes 2x "navigation" with RouterLinks
### Description of the bug

When react is disabled by [setting ``reactEnabled`` to ``false``](https://vaadin.com/docs/latest/flow/configuration/properties) all RouterLinks "navigate" twice:

https://github.com/user-attachments/assets/21e6df71-f206-4bdf-b7f1-1b1f85415fd3

NOTE: We didn't notice the problem before updating to 24.7

### Expected behavior

Navigation works like with react being enable.

https://github.com/AB-xdev/routerlink-problems-24.7

### Versions

- Vaadin / Flow version: 24.7.1
- Browser version: Happens in all browsers (tested with Chrome/Firefox)

mcollovati:
When using legacy vaadin-router, browser history for client-side navigation is updated by the router itself when navigation completes. However, in some situations the history gets updated also by the server side, causing two history entries for the same navigation and consequently a wrong behavior if pressing the browser back button.

This change:
- Prevents updating history on the server side, unless the server is handling a postponed navigation for which vaadin-router does not update the history
- Adds a test module for vaadin-router with the most common navigation tests
- Re-enables several tests that are now passing, likely because previous changes fixed the related issues

Fixes #21243
Fixes #19494
Fixes vaadin/hilla#335


flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
```diff
@@ -443,25 +443,27 @@ protected List<Class<? extends RouterLayout>> getTargetParentLayouts(
     }
 
     private void pushHistoryStateIfNeeded(NavigationEvent event, UI ui) {
+        boolean reactEnabled = ui.getInternals().getSession().getService()
+                .getDeploymentConfiguration().isReactEnabled();
```
caalador, line 447: This feels like we should add a shorthand `UiInternals.getDeploymentConfiguration()` that would take care of the session->service->configuration path so it would be easier to use in different places where we need it.

Just a note for making it better later.

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
```diff
@@ -443,25 +443,27 @@ protected List<Class<? extends RouterLayout>> getTargetParentLayouts(
...
+        Location currentLocation = event.getUI().getInternals()
```
caalador, line 448:
```suggestion
        Location currentLocation = ui.getInternals()
```
I hope the given UI is the same as the event UI.

caalador (approved):

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
mcollovati: Good point! I'll make an enhancement ticket

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
mcollovati: done

flow-server/src/main/java/com/vaadin/flow/router/internal/AbstractNavigationStateRenderer.java
mcollovati: #21317

PR #21314, opened 2025-04-18
fix: prevent resynchronization loop

### LINKED ISSUE #21266: Push messages during resync lead to resync loop
### Description of the bug

Related to #14147

In our app, we have a "resync" button which deliberately triggers a resync for testing. On some views, the app always goes into a resync loop, where every resync cycle triggers another.

Here is an excerpt of the console logs:

Image

Note that for each resync, there is an additional `Ignoring message from the server as a resync request is ongoing.` And indeed, some `onAttach` hook in this view starts a background process which eventually pushes a message to the client. When I disable that, the app recovers after the first resync without going into a loop.

My strong suspicion is that the client ignores the push message during resync, and when resync completes, it waits for a message of a missing id, because that would be the exact message that was ignored previously. In the above example, that would be message 17, and indeed that was a push message. I've attached the full console and network logs: [resyncloop.zip](https://github.com/user-attachments/files/19669426/resyncloop.zip)

I understand the idea behind ignoring push messages during resync, but that inevitably leads to this problem. I'm not sure whether it would be better to defer it until after resync completes, or to just adjust the syncId while ignoring the message itself, but something along those lines should be done here.

In my case, the push message comes from the `onAttach` hook that was triggered by resync and somehow managed to reach the client faster than the resync response. Therefore, it would be "more correct" to defer it and then execute it, because it is part of the resync itself, not something else that happened before. Still, this probably doesn't apply to every case.

### Expected behavior

The application should recover from resync without going into a loop, regardless of whether push messages are also sent or not.

I tried to reproduce it with the following approach, but only managed to do so once:

```java
public class ResyncLoop extends VerticalLayout {

    private final ExecutorService executor = Executors.newCachedThreadPool();
    private final Faker faker = new Faker();
    private final Grid<Person> grid;

    public ResyncLoop() {
        UI ui = UI.getCurrent();
        add(new Button("resync", clickEvent -> ui.getInternals().incrementServerId()));
        grid = new Grid<>();
        grid.setItems(fakePeople());
        grid.setSizeFull();
        grid.addColumnReorderListener(e -> executor.execute(() -> {
            ui.accessSynchronously(() -> ui.getElement().setAttribute("bogus-attr", UUID.randomUUID().toString()));
        }));
        addAndExpand(grid);
    }

    @Override
        super.onAttach(attachEvent);
        grid.removeAllColumns();
        var columns = List.of(
            grid.addColumn(Person::firstName).setHeader("Firstname"),
            grid.addColumn(Person::lastName).setHeader("Lastname")
        );
        grid.setColumnOrder(columns);
        grid.getDataProvider().refreshAll();
    }

    private List<Person> fakePeople() {
        return Stream.generate(this::fakePerson)
            .limit(100)
            .toList();
    }

    private Person fakePerson() {
        return new Person(faker.name().firstName(), faker.name().lastName());
    }

    record Person(String firstName, String lastName) {}
}
```

It uses [JavaFaker](https://github.com/DiUS/java-faker) to generate a Grid of reasonable size.
The `onAttach` hook sets the columns order, which fires a listener, which in turn starts a background process that sets some attribute in the DOM.

This is also very similar in principle to what happens in our production code, where I can reproduce every time. I just can't figure out how to reliably make the push message arrive during resync.

It's probably also possible to simplify this reproducer, e.g. by dropping JavaFaker and/or the Grid. But this is very close to our production environment, and it's the only thing that showed the problem at least once.

### Versions

- Vaadin / Flow version: 24.7.1
- Java version: OpenJDK 64-Bit Server VM Temurin-21.0.6+7 (build 21.0.6+7-LTS, mixed mode, sharing)
- OS version: Windows 11

mcollovati:
Changes pushed to the client during components' attach phase of a resynchronization request can be ignored, triggering a continuous resync cycle.

This change adds messages received during resync to the pending messages queue, so that messsages older than the resync request are discarded, whereas messages pushed after server resync are processed when client resync completes. In addition, it prevents asynchronous message sending while the server prepares the UI tree for resynchronization.

Fixes #21266


caalador (approved):

mperktold:
Wow, I didn't expect this to be fixed so fast, great work! 👍

Looking forward to the next version containing this fix!

PR #21320, opened 2025-04-22
chore!: Upgrade default Node.js to 22 and supported to 20 (23.6)

mshabarov:


flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java
```diff
@@ -60,11 +60,11 @@ public class FrontendTools {
...
      */
-    public static final String DEFAULT_NODE_VERSION = "v18.20.7";
+    public static final String DEFAULT_NODE_VERSION = "v22.14.0";
     /**
      * This is the version shipped with the default Node version.
      */
-    public static final String DEFAULT_NPM_VERSION = "10.8.2";
+    public static final String DEFAULT_NPM_VERSION = "10.9.2";
 
     public static final String DEFAULT_PNPM_VERSION = "8.6.11";
```
mcollovati, line 69: In the 2.12 PR, DEFAULT_PNPM_VERSION is 10.9.2.
Is the change missing here or has `DEFAULT_NPM_VERSION` updated accidentally instead of the pnpm verion?

mshabarov: > In the 2.12 PR, DEFAULT_PNPM_VERSION is 10.9.2

This was accidental indeed. Changed to pnpm 8 in the 2.12 PR.

We may consider changing pnpm to more recent separately.

mshabarov: I did change the node to 22 in GH workflow for both build and it-tests, I don't remember why we used 16 for build, but it makes sense to use latest in both.

https://github.com/vaadin/flow/pull/21326

mcollovati:
`FrontendStubs$ToolStubBuilder.DEFAULT_NPM_VERSION` needs to be updated

mcollovati (approved):

Issue #21255, opened 2025-04-09
Add DownloadHandler to components API


mshabarov:
Add `DownloadHandler` callback to the constructors and separate methods of the components:
- [ ] `Anchor` 
- [ ] `Image`, 
- [ ] `IFrame`
- [ ] `Avatar`
- [ ] `Icon`

PR #21338, opened 2025-04-24
feat: add download handler to html components

caalador:
Part of #21255

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -113,14 +113,13 @@ public class Anchor extends HtmlContainer
     }
 
     /**
-     * Creates an anchor component with the given text content and download
-     * handler resource.
+     * Creates an anchor component with the given text content and a callback that handles data download from the server to the client when clicking an anchor.
      *
      * @see #setHref(DownloadHandler)
      * @see #setText(String)
      *
      * @param downloadHandler
-     *            the download handler resource, not null
+     *            the callback that handles data download, not null
      * @param text
      *            the text content to set
      */
@@ -194,11 +193,10 @@ public class Anchor extends HtmlContainer
     }
 
     /**
-     * Sets the URL that this anchor links to with the URL of the given
-     * {@link DownloadHandler}.
+     * Sets the URL that this anchor links to and that is bound to a given {@link DownloadHandler} callback on the server for handling data download from the server to the client when clicking an anchor.
      *
      * @param downloadHandler
-     *            the download handler resource, not null
+     *            the callback that handles data download, not null
      */
     public void setHref(DownloadHandler downloadHandler) {
         this.href = new StreamResourceRegistry.ElementStreamResource(
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -116,7 +116,7 @@ public class Anchor extends HtmlContainer
      * Creates an anchor component with the given text content and download
      * handler resource.
      *
-     * @see #setHref(AbstractStreamResource)
+     * @see #setHref(DownloadHandler)
      * @see #setText(String)
      *
      * @param downloadHandler
@@ -125,8 +125,7 @@ public class Anchor extends HtmlContainer
      *            the text content to set
      */
     public Anchor(DownloadHandler downloadHandler, String text) {
-        setHref(new StreamResourceRegistry.ElementStreamResource(
-                downloadHandler, this.getElement()));
+        setHref(downloadHandler);
         setText(text);
     }
 
@@ -190,6 +189,20 @@ public class Anchor extends HtmlContainer
      */
     public void setHref(AbstractStreamResource href) {
         this.href = href;
+        setRouterIgnore(true);
+        assignHrefAttribute();
+    }
+
+    /**
+     * Sets the URL that this anchor links to with the URL of the given
+     * {@link DownloadHandler}.
+     *
+     * @param downloadHandler
+     *            the download handler resource, not null
+     */
+    public void setHref(DownloadHandler downloadHandler) {
+        this.href = new StreamResourceRegistry.ElementStreamResource(
+                downloadHandler, this.getElement());
         setRouterIgnore(true);
         assignHrefAttribute();
     }
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -113,7 +113,9 @@ public class Anchor extends HtmlContainer
     }
 
     /**
-     * Creates an anchor component with the given text content and a callback that handles data download from the server to the client when clicking an anchor.
+     * Creates an anchor component with the given text content and a callback
+     * that handles data download from the server to the client when clicking an
+     * anchor.
      *
      * @see #setHref(DownloadHandler)
      * @see #setText(String)
@@ -193,7 +195,9 @@ public class Anchor extends HtmlContainer
     }
 
     /**
-     * Sets the URL that this anchor links to and that is bound to a given {@link DownloadHandler} callback on the server for handling data download from the server to the client when clicking an anchor.
+     * Sets the URL that this anchor links to and that is bound to a given
+     * {@link DownloadHandler} callback on the server for handling data download
+     * from the server to the client when clicking an anchor.
      *
      * @param downloadHandler
      *            the callback that handles data download, not null
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -27,6 +27,8 @@ import com.vaadin.flow.component.PropertyDescriptor;
 import com.vaadin.flow.component.PropertyDescriptors;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.server.AbstractStreamResource;
+import com.vaadin.flow.server.DownloadHandler;
+import com.vaadin.flow.server.ElementRequestHandler;
 import com.vaadin.flow.server.StreamResource;
 import com.vaadin.flow.server.StreamResourceRegistry;
 
@@ -107,6 +109,24 @@ public class Anchor extends HtmlContainer
      */
     public Anchor(AbstractStreamResource href, String text) {
         setHref(href);
+        setText(text);
+    }
+
+    /**
+     * Creates an anchor component with the given text content and download
+     * handler resource.
+     *
+     * @see #setHref(AbstractStreamResource)
+     * @see #setText(String)
+     *
+     * @param downloadHandler
+     *            the download handler resource, not null
+     * @param text
+     *            the text content to set
+     */
+    public Anchor(DownloadHandler downloadHandler, String text) {
+        setHref(new StreamResourceRegistry.ElementStreamResource(
+                downloadHandler, this.getElement()));
         setText(text);
     }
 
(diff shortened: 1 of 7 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -110,6 +112,24 @@ public Anchor(AbstractStreamResource href, String text) {
...
+     * @see #setHref(AbstractStreamResource)
+     * @see #setText(String)
+     *
+     * @param downloadHandler
+     *            the download handler resource, not null
+     * @param text
+     *            the text content to set
+     */
+    public Anchor(DownloadHandler downloadHandler, String text) {
+        setHref(new StreamResourceRegistry.ElementStreamResource(
```
mshabarov, line 128: How about adding a new method like `setDownloadHandler` or similar and deprecate `setHref` ? 

We have to deprecate public API using `StreamResource` (that deserves a separate PR as it's not only these components).

mshabarov: Also could we maybe add a helper to `ElementRequestHandler` like
```
    static AbstractStreamResource asStreamResource(ElementRequestHandler handler, Element owner) {
        return new StreamResourceRegistry.ElementStreamResource(handler, owner);
    }
```
or maybe make the `ElementStreamResource` an outer public class.
I don't know which is better. From one point of view I don't want to expose any stream resource mentions in the new API and from another point of view we need a simple maintainable way of setting `DownloadHandler`, `UploadHandler` and even `ElementRequestHandler` to components.

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
```diff
@@ -131,6 +133,47 @@ public HtmlObject(AbstractStreamResource data, String type,
...
+     *
+     * @see #setData(AbstractStreamResource)
+     * @see #setType(String)
+     *
+     * @param data
+     *            the resource value, not null
+     * @param type
+     *            a type attribute value
+     */
+    public HtmlObject(DownloadHandler data, String type) {
```
mshabarov, line 148:
```suggestion
    public HtmlObject(DownloadHandler dataHandler, String type) {
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
```diff
@@ -131,6 +133,47 @@ public HtmlObject(AbstractStreamResource data, String type,
         add(params);
     }
 
+    /**
+     * Creates a new <code>&lt;object&gt;</code> component with given data
+     * resource and type value.
+     *
+     * @see #setData(AbstractStreamResource)
```
mshabarov, line 140: Same here, how about `setDataHandler` new method?

flow-html-components/src/main/java/com/vaadin/flow/component/html/IFrame.java
```diff
@@ -159,6 +161,21 @@ public void setSrc(AbstractStreamResource src) {
...
+    /**
+     * Sets the source of the iframe with a source URL with the URL of the given
+     * {@link StreamResource}.
+     *
+     * @see #setSrc(String)
+     *
+     * @param downloadHandler
+     *            the download handler resource, not null
+     */
+    public void setSrc(DownloadHandler downloadHandler) {
```
mshabarov, line 173: This need a new CTOR with the handler.

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
mshabarov: This is what Flow does for web-component.
But giving a `DownloadHandler` to `Anchor` you are telling _how_ the data should be loading, so it's not just a href or a url.
When saying `anchor.setHref(DownloadHandler.fromFile(new File("text.txt")))` you may make people confusing that they give actually a callback where the method expects a static url.
To be honest this was confusing me even before this API, with the `StreamResource`.

mshabarov: From the other side if you call `anchor.setHref(() -> { ... })` and `anchor.setHref("my/url/to/resource")`, the later should override the handler version and these methods should look the same.
Let's use your implementation.

mshabarov (approved):

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -110,6 +112,24 @@ public Anchor(AbstractStreamResource href, String text) {
...
+     *
+     * @see #setHref(AbstractStreamResource)
+     * @see #setText(String)
+     *
+     * @param downloadHandler
+     *            the download handler resource, not null
+     * @param text
+     *            the text content to set
+     */
+    public Anchor(DownloadHandler downloadHandler, String text) {
```
mshabarov, line 127: Needs `setHref(DownloadHandler handler)`

flow-html-components/src/main/java/com/vaadin/flow/component/html/Image.java
```diff
@@ -85,6 +87,26 @@ public Image(AbstractStreamResource src, String alt) {
...
+     * @param downloadHandler
+     *            the download handler resource, not null
+     * @param alt
+     *            the alternate text
+     *
+     * @see #setSrc(AbstractStreamResource)
+     * @see #setAlt(String)
+     */
+    public Image(DownloadHandler downloadHandler, String alt) {
+        setSrc(new StreamResourceRegistry.ElementStreamResource(downloadHandler,
```
mshabarov, line 105: Needs `setSrc(DownloadHandler handler)`

mshabarov (changes requested):
Oh, actuall

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
caalador: added

flow-html-components/src/main/java/com/vaadin/flow/component/html/Image.java
caalador: done

flow-html-components/src/main/java/com/vaadin/flow/component/html/IFrame.java
caalador: added.

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -110,6 +112,23 @@ public Anchor(AbstractStreamResource href, String text) {
         setText(text);
     }
 
+    /**
+     * Creates an anchor component with the given text content and download
+     * handler resource.
```
mshabarov, line 117:
```suggestion
     * Creates an anchor component with the given text content and a callback that handles data download from the server to the client when clicking an anchor.
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -110,6 +112,23 @@ public Anchor(AbstractStreamResource href, String text) {
...
+     *
+     * @see #setHref(DownloadHandler)
+     * @see #setText(String)
+     *
+     * @param downloadHandler
+     *            the download handler resource, not null
```
mshabarov, line 123:
```suggestion
     *            the callback that handles data download, not null
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
```diff
@@ -131,6 +133,46 @@ public HtmlObject(AbstractStreamResource data, String type,
         add(params);
     }
 
+    /**
+     * Creates a new <code>&lt;object&gt;</code> component with given data
+     * resource and type value.
```
mshabarov, line 138:
```suggestion
     * Creates a new <code>&lt;object&gt;</code> component with given {@link DownloadHandler} callback for providing an object data and type value.
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
```diff
@@ -131,6 +133,46 @@ public HtmlObject(AbstractStreamResource data, String type,
...
+     *
+     * @see #setData(AbstractStreamResource)
+     * @see #setType(String)
+     *
+     * @param data
+     *            the resource value, not null
```
mshabarov, line 144:
```suggestion
     *            the callback for providing resource data, not null
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
```diff
@@ -199,6 +241,18 @@ public void setData(AbstractStreamResource data) {
         getElement().setAttribute("data", data);
     }
 
+    /**
+     * Sets the {@link StreamResource} URL as "data" attribute value .
```
mshabarov, line 245:
```suggestion
     * Sets the URL for {@link DownloadHandler} callback as "data" attribute value .
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/IFrame.java
```diff
@@ -134,6 +136,16 @@ public IFrame(String src) {
         setSrc(src);
     }
 
+    /**
+     * Creates a new iframe with download handler resource.
```
mshabarov, line 140:
```suggestion
     * Creates a new iframe with download handler callback that provides a resource from server.
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/IFrame.java
```diff
@@ -134,6 +136,16 @@ public IFrame(String src) {
...
+     *
+     * @param downloadHandler
+     *            the download handler resource, not null
```
mshabarov, line 143:
```suggestion
     *            the download handler callback that provides a resource from server, not null
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/IFrame.java
```diff
@@ -159,6 +171,21 @@ public void setSrc(AbstractStreamResource src) {
         getElement().setAttribute("src", src);
     }
 
+    /**
+     * Sets the source of the iframe with a source URL with the URL of the given
+     * {@link StreamResource}.
```
mshabarov, line 176:
```suggestion
     * {@link DownloadHandler} callback.
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/Image.java
```diff
@@ -85,6 +87,25 @@ public Image(AbstractStreamResource src, String alt) {
         setAlt(alt);
     }
 
+    /**
+     * Creates an image with the given stream resource and an alternative text.
```
mshabarov, line 91:
```suggestion
     * Creates an image with the given download handler callback for providing an image data and an alternative text.
```

flow-html-components/src/main/java/com/vaadin/flow/component/html/Image.java
```diff
@@ -85,6 +87,25 @@ public Image(AbstractStreamResource src, String alt) {
...
+     * <p>
+     * The alternative text given to constructor is always set even if it is the
+     * default empty string which is not retained with {@link #setAlt(String)}.
+     *
+     * @param downloadHandler
+     *            the download handler resource, not null
```
mshabarov, line 97:
```suggestion
     *            the download handler callback that provides an image data, not null
```

mshabarov (changes requested):
LGTM, but few suggestions to make javadoc more clear (hopefully) regarding download handler: mentioning a stream resource in the new API is not a good idea IMO, also better to describe the download handler as a "callback that handlers data download from server to client" rather than a "resource" or an "URL". Technically it turns into URL on the client that is bound to this handler, yes, but the handler is not a resource nor an URL.

mshabarov (approved):

mshabarov (approved):

PR #21342, opened 2025-04-24
fix: Do not starve the common ForkJoinPool with blocking listener tasks

Artur-:
Calling watchAsync without any specific executor uses the common ForkJoinPool. This is bad as all the watchers will remain in a blocking state until there is a change and will thus always use up a large part of the ForkJoinPool. If other tasks try to use the common ForkJoinPool, it might run out of concurrent threads and will no longer start any tasks at all

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/FileWatcher.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/FileWatcher.java
@@ -34,7 +34,8 @@ public class FileWatcher {
 
     private DirectoryWatcher watcher;
 
-    private static ExecutorService executorService = Executors.newCachedThreadPool();
+    private static ExecutorService executorService = Executors
+            .newCachedThreadPool();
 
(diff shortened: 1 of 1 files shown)


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/FileWatcher.java
```diff
@@ -32,6 +34,8 @@ public class FileWatcher {
 
     private DirectoryWatcher watcher;
 
+    private Executor executor = Executors.newCachedThreadPool();
```
mcollovati, line 37: Does it make sense to shut down the executor when `stop` is called? Or can `start` be invoked again after `stop`?

Artur-: I changed the executor to static so it is shared among all the `FileWatcher` instances

mcollovati (approved):

PR #21357, opened 2025-04-26
Add support for VaadinWebSecurity customization via bean injection

### LINKED ISSUE #21356: Support modular security customization of VaadinWebSecurity
### Describe your motivation

Currently, `VaadinWebSecurity` is the single entry point for customizing the `HttpSecurity` configuration in Vaadin applications. However, in multi-module projects and library development, it is often necessary to modularize security customizations without forcing users to subclass or directly modify `VaadinWebSecurity`.  
Having a standard extension mechanism would allow libraries and application modules to provide security configurations independently, while still preserving the flexibility for users to override or extend their own application security setup.

Introduce a `VaadinWebSecurityCustomizer` interface.  Spring beans implementing this interface would be automatically detected and applied by `VaadinWebSecurity` during the `configure(HttpSecurity http)` phase. The customizers should be invoked in a well-defined order, e.g., respecting `@Order` annotations or implementing `Ordered`.

This mechanism would enable security customizations to be contributed from multiple modules and libraries in a non-breaking and additive way, while still allowing users to subclass `VaadinWebSecurity` if they need full control.

Example:

~~~java
public interface VaadinWebSecurityCustomizer {
    void customize(HttpSecurity http) throws Exception;
}
~~~

Then inside `VaadinWebSecurity`:

~~~java
@Autowired(required = false)
private List<VaadinWebSecurityCustomizer> customizers = Collections.emptyList();

// In the configure method
for (VaadinWebSecurityCustomizer customizer : customizers) {
    customizer.customize(http);
}
~~~

- **Direct subclassing of `VaadinWebSecurity`**: This approach is limiting when security logic is spread across modules. It can lead to tight coupling and does not scale well for larger applications or reusable libraries.
- **Using application-specific configuration classes**: While additional `@Configuration` classes could configure `HttpSecurity`, they often need to replicate the Vaadin-specific configuration steps to avoid breaking Vaadin internals, increasing the risk of misconfiguration.
- **Custom `SecurityFilterChain` beans**: Defining separate `SecurityFilterChain` beans bypasses the Vaadin-specific conventions and could lead to unexpected behavior if Vaadin's needs are not fully respected.

### Additional context

- This pattern would align Vaadin Flow security customization with other Spring extensibility patterns (like `WebMvcConfigurer`, `SecurityConfigurerAdapter`).
- This change would be fully backward compatible: users extending `VaadinWebSecurity` would continue to work as before.
- It would also encourage best practices for modular development and library support in the Vaadin ecosystem.

heruan:
Introduced `VaadinWebSecurityCustomizer` interface to allow customizations of `HttpSecurity` for Vaadin-related security configurations. Implementations of this interface are automatically detected and applied to the security filter chain. Added tests to ensure the customizers are properly applied.

Closes #21356

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
@@ -24,6 +24,7 @@ import javax.crypto.SecretKey;
 
 import java.io.IOException;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Consumer;
@@ -125,6 +126,9 @@ public abstract class VaadinWebSecurity {
 
     private NavigationAccessControl accessControl;
 
+    @Autowired(required = false)
+    private List<VaadinWebSecurityCustomizer> webSecurityCustomizers;
+
     @PostConstruct
     void afterPropertiesSet() {
         accessControl = accessControlProvider.getIfAvailable();
@@ -145,6 +149,10 @@ public abstract class VaadinWebSecurity {
      */
     @Bean(name = "VaadinSecurityFilterChainBean")
     public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+        if (webSecurityCustomizers != null) {
+            webSecurityCustomizers
+                    .forEach(customizer -> customizer.customize(http));
+        }
         configure(http);
         http.logout(cfg -> {
             cfg.invalidateHttpSession(true);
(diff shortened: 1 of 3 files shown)


heruan:
This still does not allow invoking methods of `VaadinWebSecurity`, such as `setOAuth2LoginPage`. The customizer could take `VaadinWebSecurity` as an additional argument, but still `setOAuth2LoginPage` is protected and wouldn't be accessible.

heruan:
> This still does not allow invoking methods of `VaadinWebSecurity`, such as `setOAuth2LoginPage`. The customizer could take `VaadinWebSecurity` as an additional argument, but still `setOAuth2LoginPage` is protected and wouldn't be accessible.

The latest commit change the visibility of some useful methods of `VaadinWebSecurity` to allow the customizer to set the login page. For example:

~~~java
@AutoConfiguration
class MyAutoConfiguration {

    @Bean
    VaadinWebSecurityCustomizer configure() {
        return (conf, http) -> {
            conf.setOAuth2LoginPage(http, "/oauth2/authorization/keycloak");
            http.oidcLogout(this::configureOidcLogout);
        };
    }
}
~~~

heruan:
Closing this after discussion in #21356 and coming up with a new one.

PR #21365, opened 2025-04-29
fix: prevent file watcher threads from slowing down JVM shutdown

mcollovati:
FileWatcher was using an executor that creates non-daemon threads to watch for file changes. This caused a delay in JVM shutdown, as it would wait at least 60 seconds for all threads in the executor pool to be evicted.

This change provides a custom thread factory that creates daemon threads for the executor to prevent JVM shutdown delays.


Artur- (dismissed):

Artur- (approved):

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/InternalThreadFactory.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * A factory for creating daemon threads with custom naming conventions, used to
+ * generate threads with a predefined naming pattern.
+ * <p>
+ * </p>
+ * This class implements the {@link ThreadFactory} interface to provide a
+ * mechanism for instantiating threads that are configured with specific
+ * attributes such as name prefix, thread priority, and daemon status.
+ */
+class InternalThreadFactory implements ThreadFactory {
```
Artur-, line 31: If this was public, then Copilot could reuse it instead of creating its own

mcollovati: Should the class be moved to flow-server, or is it ok for Copilot to pick it from vaadin-dev-server?

Artur-: dev-server should be fine, it's a dev dependency

Artur- (dismissed):

Artur- (approved):

PR #21373, opened 2025-04-30
feat: Introduce VaadinSecurityConfigurer for modular security configuration

### LINKED ISSUE #21356: Support modular security customization of VaadinWebSecurity

heruan:
Closes #21356 by extracting security configuration logic into a `VaadinWebSecurityConfigurer` adhering to Spring's `SecurityConfigurer` pattern. Attempts to keep backwards compatibility to existing extensions of `VaadinWebSecurity`, needs extensive testing to ensure this.

- [x] Prototype the ideas discussed in #21356
- [x] Write tests once the prototype is completed


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -105,34 +80,19 @@
...
-    @PostConstruct
-    void afterPropertiesSet() {
-        accessControl = accessControlProvider.getIfAvailable();
-        authenticationContext.setRolePrefixHolder(vaadinRolePrefixHolder);
-    }
-
     private final AuthenticationContext authenticationContext = new AuthenticationContext();
 
+    private final VaadinWebSecurityConfigurer configurer = new VaadinWebSecurityConfigurer();
+
```
knoobie, line 95: Just questioning: why is this created manually here instead of using the Bean that was exposed?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurer.java
```diff
@@ -0,0 +1,589 @@
...
+     * <p>
+     * This matcher combines multiple specific matchers, including those for
+     * framework internal requests, anonymous endpoints, allowed Hilla views,
+     * anonymous routes, custom web icons, and default security configurations.
+     *
+     * @return a {@link RequestMatcher} that matches requests to be allowed
+     *         without authentication
+     */
+    public RequestMatcher vaadinPermitAllMatcher() {
+        return RequestMatchers.anyOf(
```
knoobie, line 350: nit: I would personally like to see this method only add Hilla related auth permit all calls if Hilla is on the classpath to reduce the rules spring has to check on each request (my main concern why O haven't used VaadinWebSecurity in the past)

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
heruan: It is still created here for backwards compatibility, in case subclasses are overriding `getAuthenticationContext` with a custom instance (similar for `getNavigationAccessControl`). Most probably not needed such override would likely fail elsewhere, I'd be happy to remove both once we confirm it won't break existing subclasses.

heruan: We ended up with reverting all changes in `VaadinWebSecurity`, see https://github.com/vaadin/flow/issues/21356#issuecomment-2847129263

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurer.java
knoobie: Performance and security - I don't like to open my security (especially permitAll calls) for something that we don't use. Especially when I don't have studied the inner workings of Vaadin / Hilla's RequestMatcher and what is all considered "internally". I remember this nifty security bug introduced in V15 (or later) where the router link navigation skipped spring security chain because the link was handled internally which was a big problem.. the same could be said for Performance, every request matcher has to be evaluated creating unnecessary CPU circles

mcollovati: If I get it correctly, the point is just omit Hilla related matchers if Hilla dependencies have been excluded from the classpath.
On Flow there should already be some similar check somewhere (perhaps in the menu registry stuff)

knoobie: Marco is right 🙂 there should be already some snippets in flow that can be used to check if Hilla is available. That should be enough, no need to create additional APIs for user to trigger imho

heruan: Then something like this should do the trick:
~~~java
    public RequestMatcher defaultPermitMatcher() {
        var urlMapping = getRequestUtil().getUrlMapping();
        var baseMatcher = RequestMatchers.anyOf(
                getRequestUtil()::isFrameworkInternalRequest,
                getRequestUtil()::isAnonymousRoute,
                getRequestUtil()::isCustomWebIcon,
                getDefaultWebSecurityIgnoreMatcher(urlMapping),
                getDefaultHttpSecurityPermitMatcher(urlMapping));
        if (EndpointRequestUtil.isHillaAvailable()) {
            return RequestMatchers.anyOf(baseMatcher,
                    getRequestUtil()::isAllowedHillaView,
                    getRequestUtil()::isAnonymousEndpoint);
        }
        return baseMatcher;
    }
~~~

knoobie: Just speaking out thoughts (nothing to be done in this pr): with this kind of design, it might be possible to split Flow and Hilla related security config into separate classes or even repositories so that the Hilla team could maintain their own Implementation.. 

e.g. VaadinFlowSecurity, VaadinHillaSecurity and VaadinSecurity (containing both) allowing people to explicitly choose what they want / need and allowing the dedicated teams to focus on their side. Even tho I got a feeling that's not what the Managements wants for the future where both frameworks should coexist in applications more naturally 😅

mcollovati:
To be sure everything is working properly, we need to run all `flow-tests/vaadin-spring-tests/test-spring-security-XXXX` with the new configurer, in addition to the current one using `VaadinWebSecurity`.
I'm still not sure what's the best way of doing it without duplicating the test modules.
A dirty solution could be having an additional pom file that starts the application with a specific Spring profile.

vaadin-spring/src/test/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurerTest.java
```diff
@@ -0,0 +1,215 @@
...
+        http.with(configurer, c -> {
+            c.enableNavigationAccessControl(enableNavigationAccessControl);
+        }).build();
+
+        var nac = http.getSharedObject(NavigationAccessControl.class);
+        assertThat(nac.isEnabled()).isEqualTo(enableNavigationAccessControl);
+    }
+
+    @Route
+    static class TestLoginView extends Component {
```
knoobie, line 213: @mcollovati hijacking your comment into a "thread" :) 

> To be sure everything is working properly, we need to run all flow-tests/vaadin-spring-tests/test-spring-security-XXXX with the new configurer, in addition to the current one using VaadinWebSecurity.
I'm still not sure what's the best way of doing it without duplicating the test modules.
A dirty solution could be having an additional pom file that starts the application with a specific Spring profile.

It depends how "long" you wanna have the old way available.. if you deprecate it.. and don't change anything in VaadinWebSecurity.. you could even say: there are multiple branches checking that the old stuff works.. lets only test the new VaadinSecurityConfigurer in 24.8 onwards ;)

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -0,0 +1,556 @@
...
+        getSharedObject(LogoutSuccessHandler.class).or(() -> {
+            if (logoutSuccessUrl != null) {
+                return createSimpleUrlLogoutSuccessHandler(logoutSuccessUrl);
+            } else if (postLogoutRedirectUri != null) {
+                return createOidcLogoutSuccessHandler(postLogoutRedirectUri);
+            }
+            return Optional.empty();
+        }).ifPresent(configurer::logoutSuccessHandler);
+        logoutHandlers.forEach(configurer::addLogoutHandler);
+        logoutHandlers.clear();
```
mcollovati, line 450: I wonder if it would be better to move these lines inside the `!alreadyInitializedOnce` block, instead of clearing the handlers’ list. My concern is that clearing handlers may cause trouble when debugging issues.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -0,0 +1,556 @@
...
+                            configurer.getLogoutSuccessHandler(),
+                            configurer.getLogoutHandlers());
+                    return filter;
+                }
+            };
+            configurer.withObjectPostProcessor(postProcessor);
+        }
+    }
+
+    Optional<LogoutSuccessHandler> createSimpleUrlLogoutSuccessHandler(
```
mcollovati, line 469: Why is this method package protected? I cannot see any usage in tests or other classes.
If we can make it private, then I would also use `LogoutSuccessHandler` as return type, since the method never returns an empty Optional.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -0,0 +1,556 @@
...
+            String logoutSuccessUrl) {
+        var handler = new VaadinSimpleUrlLogoutSuccessHandler();
+        handler.setRedirectStrategy(new UidlRedirectStrategy());
+        handler.setDefaultTargetUrl(logoutSuccessUrl);
+        return Optional.of(handler);
+    }
+
+    Optional<LogoutSuccessHandler> createOidcLogoutSuccessHandler(
```
mcollovati, line 477: Same here: can this method be private? Or is it used elsewhere in the code?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -0,0 +1,556 @@
...
+    }
+
+    private AccessDeniedHandler createAccessDeniedHandler() {
+        var exceptionHandlers = new LinkedHashMap<Class<? extends AccessDeniedException>, AccessDeniedHandler>();
+        exceptionHandlers.put(CsrfException.class, (req, res, exc) -> res
+                .setStatus(HttpStatus.UNAUTHORIZED.value()));
+        var requestHandlers = new LinkedHashMap<RequestMatcher, AccessDeniedHandler>();
+        requestHandlers.put(getRequestUtil()::isEndpointRequest,
+                new DelegatingAccessDeniedHandler(exceptionHandlers,
+                        new AccessDeniedHandlerImpl()));
```
mcollovati, line 524: nit: can perhaps be set only if Hilla is available.

mcollovati (changes requested):
A first round of comments

vaadin-spring/src/test/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurerTest.java
mcollovati: @knoobie it could make sense if `VaadinWebSecurity` will be dropped in Vaadin 25.
However, there's still the possibility that we could change something security-related (not `VaadinWebSecurity`) in `vaadin-spring` or in the Hilla implementation of the bridge interfaces (`EndpointRequestUtil`, `FileRouterRequestUtil`) and the changes are not backported, so we could not rely on older branches to ensure there are no regressions.

knoobie: You are right, it would have been too easy :)

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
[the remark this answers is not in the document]
mcollovati: Maybe we can add a static `vaadin()` method that returns the configurer instance, the same as suggested in the Spring Security documentation.
I wonder if we can then potentially expand by adding `flow()` `hilla()` factory methods that return a pre-configured configurer instance (slightly related to Knookie's comment).

mcollovati:
`VaadinWebSecurity` has this annotation: `@Import(VaadinAwareSecurityContextHolderStrategyConfiguration.class)`.
The import looks like fundamental to make the configuration work. 
Can we do it automatically with the new configurer? Or is the developer forced to put the annotation on its configuration class?

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
```diff
@@ -197,4 +198,9 @@ public VaadinRolePrefixHolder vaadinRolePrefixHolder(
                 .map(GrantedAuthorityDefaults::getRolePrefix).orElse(null));
     }
 
+    @Bean
+    @ConditionalOnMissingBean
```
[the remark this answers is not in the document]
heruan: I didn't put JavaDoc because I added the method as package-private, following Spring conventions on auto-configuration classes.

I can add JavaDoc, do I also need to make the method public? It's usually discouraged by Spring since auto-configuration classes shouldn't be subclassed, but rather use conditions to provide beans.

This method for example has a condition to allow overriding, that is needed when you extend `VaadinWebSecurity` since it provides its own bean. I left the bean there since it's a named bean, and I didn't want to bring it here with the name since only `VaadinWebSecurity` uses named beans and it's not necessary if you don't extend it.

mcollovati:
Just for the record: according to the Spring [documentation](https://docs.spring.io/spring-security/reference/servlet/configuration/java.html#jc-custom-dsls), the configurer could be added automatically by registering it in `META-INF/spring.factories`, and then potentially disabled by the developer.

My opinion is that we should not apply this automatic configuration, but what do you think?

knoobie:
> My opinion is that we should not apply this automatic configuration, but what do you think?

Same here - "cool idea", but lets not do it :) This could create really interesting problems, especially for people that don't use VaadinWebSecurity at the moment and have hardcoded Vaadin's integration (like we do)

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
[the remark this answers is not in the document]
heruan: Is that suggested to be a singleton instance? Since it has state, I'm not sure how that would work: what if I want two different security configuration for different filter chains? Maybe we can provide a factory method to return a new instance, if having a public constructor is really discouraged for some reason.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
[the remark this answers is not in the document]
mcollovati: > I didn't put JavaDoc because I added the method as package-private, following Spring conventions on auto-configuration classes.

Ok, makes sense.

> I can add JavaDoc, do I also need to make the method public? It's usually discouraged by Spring since auto-configuration classes shouldn't be subclassed, but rather use conditions to provide beans.
> 
> This method for example has a condition to allow overriding, that is needed when you extend VaadinWebSecurity since it provides its own bean. I left the bean there since it's a named bean, and I didn't want to bring it here with the name since only VaadinWebSecurity uses named beans and it's not necessary if you don't extend it.

Got it. Let's keep it as proposed.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -0,0 +1,556 @@
...
+                configurer.loginPage(oauth2LoginPage).permitAll();
+                configurer.successHandler(getAuthenticationSuccessHandler());
+            });
+        }
+        http.csrf(this::customizeCsrf);
+        http.logout(this::customizeLogout);
+        http.requestCache(this::customizeRequestCache);
+        http.exceptionHandling(this::customizeExceptionHandling);
+        if (!alreadyInitializedOnce) {
+            http.authorizeHttpRequests(this::customizeAuthorizeHttpRequests);
```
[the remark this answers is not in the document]
heruan: No, the `alreadyInitializedOnce` flag is used also in other customizer methods.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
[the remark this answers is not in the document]
mcollovati: Nope, not stateless. Just a factory method as in Spring docs.

```java
	public static VaadinSecurityConfigurer vaadin() {
		return new VaadinSecurityConfigurer();
	}
	@Bean
	public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
		http
			.with(VaadinSecurityConfigurer.vaadin(), (dsl) -> dsl
				.flag(true)
			)
			// ...
		return http.build();
	}
```

The `flow()` and `hilla()` methods could potentially pre-configure something on the `VaadinSecurityConfigurer` new instance. But we can consider this in another PR.

heruan:
> `VaadinWebSecurity` has this annotation: `@Import(VaadinAwareSecurityContextHolderStrategyConfiguration.class)`. The import looks like fundamental to make the configuration work. Can we do it automatically with the new configurer? Or is the developer forced to put the annotation on its configuration class?

Is there a reason why the bean is defined in that class? Feels like it should be in `SpringSecurityAutoConfiguration` 🤔

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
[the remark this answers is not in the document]
heruan: > Please review all Javadoc to report relevant information present in the related `VaadinWebSecurity` methods.

Done.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -0,0 +1,556 @@
...
+        }
+        // Configuring the authorized requests here allows other configurers to
+        // customize the authorized requests during their own initialization.
+        // Also, it ensures that the anyRequest authorize-rule is configured as
+        // late as possible, since it must be the last authorize-rule to be set.
+        http.authorizeHttpRequests(registry -> {
+            if (anyRequestAuthorizeRule != null) {
+                anyRequestAuthorizeRule.accept(registry.anyRequest());
+            }
+        });
```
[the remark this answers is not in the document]
heruan: Yes, done.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -0,0 +1,556 @@
...
+        return this;
+    }
+
+    /**
+     * Configures the access rule for any request not matching other configured
+     * rules.
+     * <p>
+     * The default rule is to require authentication, which is the equivalent of
+     * passing {@link AuthorizedUrl#authenticated()} to this method.
+     *
```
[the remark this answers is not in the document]
heruan: Good point, done.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
mcollovati: I see. 
It looks to me a bit strange to mess up with the internal state :thinking: 
How are Spring out-of-the-box configurers handling multiple calls to the `init` method?

heruan: Good point, I changed it to filter our existing handlers before adding the new ones.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
mcollovati: Let's make it private. I would not use Optional in this case, but I'll leave it up to you.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
mcollovati: Let's make it private

heruan:
> It is probably not in `SpringSecurityAutoConfiguration` because it is required only if `VaadinWebSecurity` is used. Moving it now could potentially cause unwanted side effects. Needs to be investigated.

I don't see how it depends from `VaadinWebSecurity` being used, but I might not have the full picture here. Ideal would be to add it conditional in `SpringSecurityAutoConfiguration` with default naming, and leave the import in `VaadinWebSecurity` for backwards compatibility.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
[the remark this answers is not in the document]
heruan: > Nope, not stateless. Just a factory method as in [Spring docs](https://docs.spring.io/spring-security/reference/servlet/configuration/java.html#jc-custom-dsls).

Added the static `vaadin()` factory method and made the constructor private.

mcollovati:
@heruan a recap of what's missing:

* fix stack overflow caused by `customizeRequestCache`
* handling `VaadinAwareSecurityContextHolderStrategyConfiguration` without `@Import`
* methods for stateless authentication configuration (we already have `VaadinStatelessSecurityConfigurer` but it would be good to have helper methods in `VaadinSecurityConfigurer` like the `setStatelessAuthentication` in `VaadinWebSecurity`

heruan:
> * fix stack overflow caused by `customizeRequestCache`

How can I replicate this? Tests seems to pass and on my test project I don't get the exception.

> * handling `VaadinAwareSecurityContextHolderStrategyConfiguration` without `@Import`

I'll add it to `SpringSecurityAutoConfiguration` as `@ConditionalOnMissingBean`, so it does not conflict when extending `VaadinWebSecurity`.

> * methods for stateless authentication configuration (we already have `VaadinStatelessSecurityConfigurer` but it would be good to have helper methods in `VaadinSecurityConfigurer` like the `setStatelessAuthentication` in `VaadinWebSecurity`

Not sure about this, since `VaadinStatelessSecurityConfigurer` is already a configurer so it should be applied on its own with `HttpSecurity::with`, e.g.

~~~java
http.with(new VaadinStatelessSecurityConfigurer(), configurer -> {
    configurer.withSecretKey(...).expiresIn(...);
});
~~~

mcollovati:
> How can I replicate this? Tests seem to pass and on my test project I don't get the exception.

I just run the flow spring security test modules and immediately spot the issue.

> Not sure about this, since VaadinStatelessSecurityConfigurer is already a configurer so it should be applied on its own with HttpSecurity::with, e.g.

Sure. But I wonder if it would be good to have a helper in `VaadinSecurityConfigurer` to have a single "entry point" to simplify the code and to be similar to what we provided with `VaadinWebSecurity`. Internally it will delegate to `VaadinStatelessSecurityConfigurer`
```java
http.with(vaadin(), configurer -> {
    configurer.statelessAuthentication(secretKey, issuer)
});
```

mcollovati:
Another question: `VaadinWebSecurity` has a protected `applyUrlMapping` method that can be useful when defining custom authorization rules for paths that should be prepended with the Vaadin servlet mapping.
Here's an example from Hilla codebase

```java
        http.authorizeHttpRequests(auth -> {
            auth.requestMatchers(new AntPathRequestMatcher("/public/**"))
                .permitAll()
                .requestMatchers(
                    new AntPathRequestMatcher(applyUrlMapping("/")))
                .permitAll().requestMatchers(
                    new AntPathRequestMatcher(applyUrlMapping("/form")))
                .permitAll()
                .authorizeHttpRequests().requestMatchers(
                    new AntPathRequestMatcher(applyUrlMapping("/proxied-service")))
                .permitAll();
```

I wonder if something like `applyUrlMapping` should be somehow provided by the configurer or not.
Something like:
```java
http.with(vaadin(), cfg -> cfg.authorizeServletRequest("/", auht -> auth.permitAll()))

// auht -> auth.permitAll() is a Consumer<AuthorizedUrl>
```

@knoobie also asking your point of view on this, since you have experience in configuring Spring Security for Vaadin apps without `VaadinWebSecurity`

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
```diff
@@ -197,4 +200,15 @@ public VaadinRolePrefixHolder vaadinRolePrefixHolder(
...
+    AuthenticationContext authenticationContext() {
+        return new AuthenticationContext();
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    public SecurityContextHolderStrategy vaadinAwareSecurityContextHolderStrategy() {
```
mcollovati, line 211: can it be package private?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -0,0 +1,723 @@
...
+        this.enableRequestCacheConfiguration = enableRequestCacheConfiguration;
+        return this;
+    }
+
+    /**
+     * Enables or disables automatic configuration of exception handling
+     * (enabled by default).
+     * <p>
+     * This configurer will automatically configure exception handling during
+     * traversal of the security filter chain.
```
mcollovati, line 356: I don't really understand what "during traversal of the security filter chain" means.
Could it just say "... configure exception handling to work properly with Flow and Hilla"?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -0,0 +1,723 @@
...
+                    getRequestUtil()::isFrameworkInternalRequest);
+        }
+        if (formLoginPage != null) {
+            configurer.ignoringRequestMatchers(
+                    new AntPathRequestMatcher(formLoginPage));
+        }
+    }
+
+    private void customizeLogout(LogoutConfigurer<HttpSecurity> configurer) {
+        configurer.invalidateHttpSession(true);
```
mcollovati, line 597: Should we make `invalidateHttpSession` configurable? Otherwise it cannot be overridden without disable the whole logout settings.

mcollovati (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
heruan: I assumed it was necessary for Vaadin to work, since it's hardcoded also in `VaadinWebSecurity`. It would make sense to make this configurable if it's not strictly necessary for Vaadin to work.

mcollovati: Yes, let's make it configurable

mcollovati: BTW, I think it would be good to document which configurations are applied and which shared objects are used and set.

heruan: I just noticed the flag is true by default, so we don't need to set it. Better just remove it.

heruan: > BTW, I think it would be good to document which configurations are applied and which shared objects are used and set. 

I'll add it to the JavaDoc.

heruan: Done both.

vaadin-spring/src/test/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurerTest.java
mcollovati, line 1: Could we add a couple of tests ensuring the `enableXYZ` methods work as expected?

mcollovati (approved):
Tested the new configurer with Hilla and MPR and found that it worked well.
If something is still missing, we can create additional PRs to fill the gaps.

PR #21398, opened 2025-05-07
fix: Preserve handler type when chaining listener methods

mshabarov:
This allows to type
```java
DownloadHandler.forFile(new File("image.png"))
       .whenComplete(success -> {});
```

and get the proper type after you ask "Introduce local variable" in IDE:

```java
FileDownloadHandler fileDownloadHandler =
                DownloadHandler.forFile(new File("image.png"))
                        .whenComplete(success -> {});
```

Without it, the IDE create a variable of type `TransferProgressAwareHandler` that you cannot pass to component.

Related-to https://github.com/vaadin/flow/pull/21343

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/TransferProgressAwareHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/TransferProgressAwareHandler.java
@@ -41,8 +41,12 @@ import elemental.json.JsonValue;
  * @param <T>
  *            type of transfer event, e.g.
  *            {@link com.vaadin.flow.server.DownloadHandler}
+ * @param <R>
+ *            type of the subclass implementing this abstract class, needed for
+ *            revealing a proper type when you chain the methods
  */
-public abstract class TransferProgressAwareHandler<T> implements Serializable {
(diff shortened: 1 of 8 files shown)


flow-server/src/main/java/com/vaadin/flow/server/streams/TransferProgressAwareHandler.java
```diff
@@ -103,8 +107,7 @@ public Registration addTransferProgressListener(
      * @param <R>
```
caalador, line 107: <R> param doesn't exist anymore breaking the javadoc build with an error

caalador: Also the javadoc could be fixed for the `{@link com.vaadin.flow.component.UI#access}` part that should be `{@link com.vaadin.flow.component.UI#access(Command)}` for all places it is used

mshabarov: Done

caalador (approved):

PR #21399, opened 2025-05-07
chore: use github variable for concurrency group

ZheSun88:
seems the environment variable is not working for concurrency group..

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -11,6 +11,8 @@ concurrency:
   group: ${{ github.head_ref }} || ${{ github.ref_name }}
   cancel-in-progress: true
 env:
+  HEAD_REF: ${{ github.head_ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/validation.yml
```diff
@@ -8,11 +8,9 @@ on:
 permissions:
   contents: read
 concurrency:
-  group: $HEAD_REF || $REF_NAME
+  group: ${{ github.head_ref }} || ${{ github.ref_name }}
   cancel-in-progress: true
 env:
-  HEAD_REF: ${{ github.head_ref }}
-  REF_NAME: ${{ github.ref_name }}
```
mcollovati, line 15: Env variables are used in the first step of `build` job. We should not remove them.

mcollovati: Never mind

PR #21402, opened 2025-05-08
feat: add subscription mechanism to receive all processed events

### LINKED ISSUE #21370: [Signals] Provide a way of subscribing to changes / events
### Describe your motivation

Hilla emits signal instances from the endpoints, which behind the scene is translated to a Flux connection that could be subscribed to, and get notified asynchronously about the changes applied to the server-side signal instance. 

Signal instance should have public API to enable Hilla's Full-stack Signals library subscribing to the details of submitted commands and their results. 

Some background:
The client-side implementation sends updates in form of commands via `POST` requests that optimistically assumed as successful as long as it reaches the server, since it is not awaiting for a response. Actual updates are received though a Flux (websocket subscription) asynchronously. Each received update is either the result of a command that is sent by an instance of the client-side signal itself, or the commands sent by other clients. As the result, a client-side signal instance should be able to distinguish between the command sent itself, and whether it was applied successfully or got rejected. This enables a client-side signal to ignore, accept, or retry a command based on the received update. Previous implementation of the Full-stack Signals in Hilla was using the same unique ID of the client command also for the asynchronous responses (sent through websocket).

This feature request is about a public API that enables subscribing to the signal changes in a way that the received updates are identifiable.

taefi:
## Description

This pull request introduces a new subscription mechanism for signals, enhances utility methods, and adds comprehensive test coverage to ensure the correctness of new functionality. The changes are primarily focused on enabling subscribers to receive notifications about the results of processed commands and improving the maintainability of the codebase.

### New Subscription Mechanism for Signals:
* Added a `subscribeToPublished` method in `SignalTree` that allows subscribers to listen for the results of processed commands. Subscribers can be removed using the returned callback.

### Utility Enhancements:
* Added a new utility class `SignalUtils` with methods to retrieve the underlying signal tree and validate commands. This class is marked as internal API.

Fixes #21370

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


signals/src/main/java/com/vaadin/signals/impl/AsynchronousSignalTree.java
```diff
@@ -77,6 +77,9 @@ public void confirm(List<SignalCommand> commands) {
             notifyObservers(oldSubmitted, submitted);
 
             unconfirmedCommands.notifyResultHandlers(results, commands);
+
+            commands.forEach(command -> notifySubscribers(command,
```
Legioth, line 81: Nit: I have tried to use old-fashioned `for` loops instead of `forEach` to avoid making stacktraces deeper than necessary.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -334,4 +339,38 @@ public void commitSingleCommand(SignalCommand command) {
...
+        subscribers.add(subscriber);
+        return () -> subscribers.remove(subscriber);
+    }
+
+    /**
+     * Notifies all subscribers about the result of a processed command.
+     *
+     * @param command
+     *            the command that was processed, not <code>null</code>
+     * @param result
```
Legioth, line 369: Why is this needed?

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -334,4 +339,38 @@ public void commitSingleCommand(SignalCommand command) {
...
+     *            the result of the command, not <code>null</code>
+     */
+    protected void notifySubscribers(SignalCommand command,
```
Legioth, line 372: Could have "published" in the name here to make clear which subscribers are notified.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
taefi: Which one, the command or the result?

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
taefi: Do you mean something like `notifyPublishedSubscribers`?

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
taefi: I recall a discussion about not having the `accepted` or such in the command, however, there should be a way of associating commands to their results somehow.

signals/src/main/java/com/vaadin/signals/impl/AsynchronousSignalTree.java
taefi: Hmm... I noticed other places such as [here](https://github.com/vaadin/flow/blob/main/signals/src/main/java/com/vaadin/signals/impl/SignalTree.java#L234) where the functional style is used, should I also refactor those?

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
Legioth: > there should be a way of associating commands to their results somehow.

Why is that needed?

As discussed elsewhere, the client will anyways be able to resolve the same result so it seems redundant to extract it on the server and send it over the network.

> I'm also OK to include the `commandId` in the `CommandResult` type.

Why is that needed?

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
Legioth: Something like that, even though one might the wonder why the subscriber has been published...

signals/src/main/java/com/vaadin/signals/impl/AsynchronousSignalTree.java
Legioth: That particular example iterates over a map rather than a list. Changing those would uglify the code more so I haven't touched them.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -334,4 +339,38 @@ public void commitSingleCommand(SignalCommand command) {
...
+     * @param subscriber
+     *            the callback to run when a command is confirmed, not
+     *            <code>null</code>
+     * @return a callback that can be used to remove the subscriber before it's
+     *         triggered, not <code>null</code>
+     */
+    public Runnable subscribeToPublished(
+            BiConsumer<SignalCommand, CommandResult> subscriber) {
+        assert subscriber != null;
+        subscribers.add(subscriber);
```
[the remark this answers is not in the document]
taefi: True, and TBH, I (maybe unnecessarily) hesitated adding the `runWithLock` in this method because I had an uncertainty about the `notifySubscribers` (the other added method). 

The `subscribeToPublished` is called when a client-signal is subscribing to a server-side signal. The block of code handling that request (for creating the Flux instance and returning it) is also doing its business after acquiring its own lock. As both are reentrant locks, there's no problem for a thread to go through these methods. The way I can imagine the possibilities around the order of acquiring locks for two or more threads, I cannot see a situation that ends up in a deadlock, since that Hilla class is solely for handling the client subscription and I don't see situation where an application thread first calls `subscribeToPublished` and then ends up calling that method in Hilla while holding the lock. So it is safe to add `runWithLock` here. 

However, the `notifySubscribers` (the other added method) should also do its job by acquiring the lock, and that's where it can get confusing, since the callback being called as the subscriber is basically another method in the same class in Hilla that should acquire the lock when going through the list of client-subscriptions and sending them the updates. While this is in progress, if another client wants to subscribe, there's a chance of deadlock.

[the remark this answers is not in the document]
taefi: Probably the issue could be alleviated with the `notifySubscribers` to only create a local copy of the subscribers while holding the lock, and then calling them one-by-one only after the lock is released. But, I'm open to better suggestions.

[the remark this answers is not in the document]
taefi: Or using a `CopyOnWriteArrayList` instead of the HashSet + Lock could be solution for `notifySubscribers`.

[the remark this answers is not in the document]
Legioth: The notify method must be run with a lock because ordering of events is critical in this case. At the same time, that method is called from places that are already protected by the tree lock so we would only need to add an assert to the notify method.

Separate concurrency handling for this specific field using e.g. CopyOnWriteArrayList doesn't seem advicable since that would still leave the door open about which listeners will actually be notified in case they are added concurrently with an event being fired.

I wouldn't worry about any interference between multiple locks in this case since we're not running any application code while subscribing.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
taefi: Not sure I completely get the design you have in mind. Maybe I misunderstood the other discussion for not having `accepted/rejected` property in the response. How a command and its processing result should be associated together? 
A SignalCommand coming from the client has an Id, and can be sent back to the client (who sent it) so that the client can distinguish it (whether this is a command sent itself, or it is coming as the result of another client's changes). If it is from other clients, the client-signal will apply it locally, but if it is its own command, not having a property that shows whether it's accepted/rejected, how can it know what happened on the server, and what to do with it?

Legioth: > how can it know what happened on the server

The client has exactly the same information that the server had at that time which means that the client can re-create the same result. We need to implement this for the client anyways to handle more complex latency compensation cases properly.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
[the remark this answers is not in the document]
taefi: Hmm... the fact that `notifySubscribers` is only being called from already protected methods by the tree's lock makes the explicit acquiring of the tree's reentrant lock irrelevant, so the assertion is enough. But, this makes it more complex. 

With the above mentioned scenario, it still seems like a classical example of a situation when a deadlock can happen:
`Hilla class`:
```java
public class A {
   private final ReentrantLock lock = new ReentrantLock();
   private final Set<Sinks.Many<ObjectNode>> subscribers = new HashSet<>();
   private final SignalTree tree; // initialized in constructor
   
   // this method is called when a client tries to subscribe to a signal
   public void handleClientSubscription() {
      lock.lock();
      try {
         // create sink and add it to the list of subscribers (client-subscription) - needs locking
         tree.subscribeToPublished(this::notifySubscribers);
      } finally {
         lock.unlock();
      }
    }
    private void notifySubscribers(SignalCommand command, CommandResult result) {
      lock.lock();
      try {
        subscribers.removeIf(sink -> {
          ObjectNode processedEvent = null; // TODO: serialize to JSON
            boolean failure = sink.tryEmitNext(processedEvent).isFailure();
            if (failure) {
              getLogger().debug("Failed push");
            }
            return failure;
         });
      } finally {
         lock.unlock();
      }
    }
}
```
and
```java
public class SignalTree {
    private final ReentrantLock lock = new ReentrantLock();
    private final Set<BiConsumer<SignalCommand, CommandResult>> subscribers = new HashSet<>();
    public Runnable subscribeToPublished(BiConsumer<SignalCommand, CommandResult> subscriber) {
      runWithLock(() -> {
        assert subscriber != null;
         subscribers.add(subscriber);
        return () -> subscribers.remove(subscriber);
       });
    }
    protected void notifySubscribers(SignalCommand command, CommandResult result) {
      assert hasLock(); // running the Hilla's callback while the tree's lock is acquired by current thread
      subscribers.forEach(subscriber -> subscriber.accept(command, result));
    }
}
```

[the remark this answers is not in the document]
Legioth: I would spontaneously remove synchronization from `A` if it doesn't do anything else. I would instead make it use a `ConcurrentHashMap<SignalTree, Set<Sinks.Many<ObjectNode>>>` for the subscribers and then rely on the tree lock for the rest.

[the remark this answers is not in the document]
taefi: Anyway, I guess the problem should be handled from Hilla side. Will add the locking for these two methods.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
taefi: Renamed to `notifyPublishedCommandSubscribers`. As an internal API, the new name's length should not an issue IMO.

taefi: Resolving this. Feel free to reopen.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
taefi: removed the command result.

signals/src/main/java/com/vaadin/signals/SignalUtils.java
```diff
@@ -0,0 +1,47 @@
...
+public class SignalUtils {
+
+    /**
+     * Returns the underlying <code>SignalTree</code> instance of the given
+     * signal.
+     *
+     * @param signal
+     *            the signal to get the tree of, not <code>null</code>
+     * @return the signal tree instance, not <code>null</code>
+     * @throws NullPointerException
```
Legioth, line 21: Seems quite pointless to document this unchecked exception

signals/src/main/java/com/vaadin/signals/Signal.java
```diff
@@ -219,7 +219,21 @@ protected Predicate<SignalCommand> mergeValidators(
      */
     protected abstract Object usageChangeValue(Data data);
 
-    private boolean isValid(SignalCommand command) {
+    /**
```
Legioth, line 222: Seems pointless to have this description duplicated. There's already at least one subtle difference between the two copies.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -334,4 +338,38 @@ public void commitSingleCommand(SignalCommand command) {
...
+     *            the callback to run when a command is confirmed, not
+     *            <code>null</code>
+     * @return a callback that can be used to remove the subscriber before it's
+     *         triggered, not <code>null</code>
+     */
+    public Runnable subscribeToPublished(Consumer<SignalCommand> subscriber) {
+        return getWithLock(() -> {
+            assert subscriber != null;
+            subscribers.add(subscriber);
+            return () -> subscribers.remove(subscriber);
```
Legioth, line 360: Lock needed for removal as well

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -334,4 +338,38 @@ public void commitSingleCommand(SignalCommand command) {
...
+     *
+     * @param subscriber
+     *            the callback to run when a command is confirmed, not
+     *            <code>null</code>
+     * @return a callback that can be used to remove the subscriber before it's
+     *         triggered, not <code>null</code>
+     */
+    public Runnable subscribeToPublished(Consumer<SignalCommand> subscriber) {
+        return getWithLock(() -> {
+            assert subscriber != null;
```
Legioth, line 358: nit: could check this already before locking

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -334,4 +338,38 @@ public void commitSingleCommand(SignalCommand command) {
...
         return type;
     }
+
+    /**
+     * Subscribes to the published result of processed commands. The subscriber
+     * callback is executed after commands are processed and results are
+     * published (either accepted or rejected). Contrary to the observers that
+     * are attached to a specific node by calling {@link #observeNextChange},
+     * the <code>subscriber</code> remains active indefinitely until it is
+     * removed by executing the returned callback or the tree is destroyed.
```
Legioth, line 348: This gives the impression that there exists and operation like `tree.destroy()`. Could just as well leave out that part since it's quite implicit that nothing will happen if the tree is no longer used.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -334,4 +338,38 @@ public void commitSingleCommand(SignalCommand command) {
...
+    public Runnable subscribeToPublished(Consumer<SignalCommand> subscriber) {
+        return getWithLock(() -> {
+            assert subscriber != null;
+            subscribers.add(subscriber);
+            return () -> subscribers.remove(subscriber);
+        });
+    }
+
+    /**
+     * Notifies all subscribers about the result of a processed command. This
```
Legioth, line 365: Doesn't notify about the result but only about the fact that is has been processed.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -334,4 +338,38 @@ public void commitSingleCommand(SignalCommand command) {
     public Type type() {
         return type;
     }
+
+    /**
+     * Subscribes to the published result of processed commands. The subscriber
```
Legioth, line 343: "published result" makes it sound like there may also be "unpublished results". Also, the actual result is not relevant here but only the fact that the command was processed. More relevant is that we guarantee that callbacks will be invoked in the order that the commands are processed.

signals/src/main/java/com/vaadin/signals/impl/SynchronousSignalTree.java
```diff
@@ -72,12 +72,21 @@ public void publishChanges() {
...
+                for (SignalCommand command : changes.getCommands()) {
+                    notifyPublishedCommandSubscribers(command);
+                }
             }
 
             @Override
             public void markAsAborted() {
-                changes.notifyResultHandlers(CommandResult.rejectAll(results,
-                        "Transaction aborted"));
+                var rejected = CommandResult.rejectAll(results,
```
Legioth, line 83: No longer serves any purpose to extract this local variable.

signals/src/test/java/com/vaadin/signals/SignalUtilsTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.mockito.ArgumentMatchers.any;
+
+public class SignalUtilsTest {
+
+    @Test
+    void treeOf_returnsSignalsUnderlyingTree() {
+        SignalTree tree = Mockito.mock(SignalTree.class);
+        Signal<?> signal = Mockito.mock(Signal.class);
+        Mockito.when(signal.tree()).thenReturn(tree);
+        assertEquals(tree, SignalUtils.treeOf(signal));
```
Legioth, line 20: nit: `assertSame`

signals/src/test/java/com/vaadin/signals/SignalUtilsTest.java
```diff
@@ -0,0 +1,34 @@
...
+        Mockito.verify(signal, Mockito.times(1)).tree();
+
+    }
+
+    @Test
+    void isValid_callsSignalsIsValid() {
+        Signal<?> signal = Mockito.mock(Signal.class);
+        SignalCommand command = new SignalCommand.SetCommand(Id.random(),
```
Legioth, line 28: Nit: There's a helper in `TestUtil` for creating this type of command.

signals/src/test/java/com/vaadin/signals/impl/SynchronousSignalTreeTest.java
```diff
@@ -350,4 +350,102 @@ void observe_observeInCallback_registeredAgain() {
...
+        var transactionCommand = new SignalCommand.TransactionCommand(commandId,
+                List.of(conditionCommand, setCommand));
+
+        tree.commitSingleCommand(transactionCommand);
+
+        assertEquals(1, count.get());
+        assertEquals(commandId, resultContainer.get().commandId());
+
+        // keeps the order of underlying commands:
+        assertEquals(conditionId,
```
Legioth, line 410: Seems redundant in this context since it basically just verifies that `TransactionCommand` is immutable.

Legioth (changes requested):

Legioth (changes requested):

signals/src/main/java/com/vaadin/signals/SignalUtils.java
```diff
@@ -0,0 +1,45 @@
...
+     * might make changes to the data.
+     *
+     * @param command
+     *            the command to check, not <code>null</code>
+     * @param signal
+     *            the signal to check the command against, not<code>null</code>
+     * @return <code>true</code> if the command is valid, <code>false</code>
+     *         otherwise
+     */
+    public static boolean isValid(SignalCommand command, Signal<?> signal) {
```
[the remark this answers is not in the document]
taefi: I'm reading the method signature as "Is {it a} valid command {for the} signal?", and that's why I defined the order as is.
However, I have no strong tendency toward keeping the current order, in case there's a global convention in place.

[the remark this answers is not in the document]
Legioth: There's no convention yet but I suspect we will eventually need more static methods in this class and they would all take a signal instance as one parameter. These are all basically instance methods that are static methods for technical reasons, and in that case it's quite conventional that the `this` that isn't an actual `this` is the first parameter.

signals/src/main/java/com/vaadin/signals/impl/SynchronousSignalTree.java
```diff
@@ -72,12 +72,20 @@ public void publishChanges() {
...
+                    notifyPublishedCommandSubscribers(command);
+                }
             }
 
             @Override
             public void markAsAborted() {
                 changes.notifyResultHandlers(CommandResult.rejectAll(results,
                         "Transaction aborted"));
+
+                for (SignalCommand command : changes.getCommands()) {
```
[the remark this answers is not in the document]
taefi: > Commands for an aborted transaction will not be "processed" in the sense that that they will not be involved in the official event ordering.

Well, it depends on what we interpret the meaning of processed. To me, the term "processed" means a command that is delivered to the server, and either managed to change the value or got rejected. The subscriber is just interested in an indicator of signal(tree) received and considered this command (regardless of the result).

Or maybe, I'm not understanding the question here.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -334,4 +338,38 @@ public void commitSingleCommand(SignalCommand command) {
...
+     * {@link #observeNextChange}, the <code>subscriber</code> remains active
+     * indefinitely until it is removed by executing the returned callback.
+     *
+     * @param subscriber
+     *            the callback to run when a command is confirmed, not
+     *            <code>null</code>
+     * @return a callback that can be used to remove the subscriber, not
+     *         <code>null</code>
+     */
+    public Runnable subscribeToPublished(Consumer<SignalCommand> subscriber) {
```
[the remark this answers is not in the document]
taefi: If we're changing this, probably it makes sense to also rename the `notifyPublishedCommandSubscribers` to `notifyProcessedCommandSubscribers`.

signals/src/main/java/com/vaadin/signals/SignalUtils.java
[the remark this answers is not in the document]
taefi: Done.

signals/src/main/java/com/vaadin/signals/impl/AsynchronousSignalTree.java
```diff
@@ -77,6 +77,10 @@ public void confirm(List<SignalCommand> commands) {
             notifyObservers(oldSubmitted, submitted);
 
             unconfirmedCommands.notifyResultHandlers(results, commands);
+
+            for (SignalCommand command : commands) {
```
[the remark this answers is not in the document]
taefi: Done.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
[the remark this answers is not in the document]
taefi: Done.

signals/src/main/java/com/vaadin/signals/impl/SignalTree.java
```diff
@@ -90,6 +92,8 @@ public enum Type {
 
     private final Type type;
 
+    private final Set<Consumer<SignalCommand>> subscribers = new HashSet<>();
```
knoobie, line 95: nit: I like my subscriber ordered to ensure consistency

signals/src/main/java/com/vaadin/signals/impl/SynchronousSignalTree.java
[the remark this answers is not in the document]
taefi: > I don't see any reason for ever showing those aborted commands to anyone else

Just to make sure we are on the same page: by "anyone else" did you mean:
1. that only the sender of an aborted command should get it back? 
    * This looks ideal, and should technically be possible to create a mapping in Hilla's server-side handler to prevent sending the failed ones to others.
2.  Or, no one should get it back? 
    * I don't agree. From a Hilla client's POV (an external observer), it only cares about making sure that my command "arrived" to the server, server look at it, and whatever was the decision (acceptance or rejection) its effect is already reflected in the local _server's confirmed state_. So it should be able to either apply or reject it on its own _server's confirmed state_ based on the guaranteed order of received commands and the previous _state_. 
    
>  we again need to introduce some external way for the subscriber to know that those commands were actually not accepted by the tree since it's not apparent from the previous state of the tree 

Maybe, I'm missing all the reasons for an aborted transaction, and there are commands that should not be even delivered back to the client that issued it. If the sender of an aborted/failed command doesn't receive it back, it remains in the _"locally applied, not reflected in server's state"_ forever. Ofc, there could be mechanisms to decide a command wasn't accepted, however, those could be partial.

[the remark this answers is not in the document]
Legioth: No wait. I forgot the that non-transactional path still goes through the same logic when `canCommit()` is false. So that complicates things significantly. Maybe we have to go back to passing the status to the subscriber after all.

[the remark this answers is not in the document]
taefi: > Maybe we have to go back to passing the status to the subscriber after all.

I find this useful when we want to distinguish between an accepted command that should be sent to all Hilla client subscribers, versus the failed commands that should only be pushed back to the issuer client.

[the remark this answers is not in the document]
taefi: At the same time, it is an internal API. Let's keep it simple and revisit this if/when there was a need.

Legioth (approved):

PR #21405, opened 2025-05-08
feat: extract generated service worker content into sw.generated.ts

platosha:


platosha:
Probably makes little sense, as a lot in the file is intended for customization. Let me close for now.

PR #21408, opened 2025-05-09
refactor!: Renamings in new handlers API

mshabarov:
This refactors the API:
- moves all new classes into new package `com.vaadin.flow.server.streams`
- renames `fileSize` to `contentLength`
- adds missing `getOwningElement()` to `UploadEvent` and add the same method to `DownloadEvent` for unification.


caalador (approved):
Have a PR ready for flow-components to update the import before merging this PR as the components will break from this.

PR #21424, opened 2025-05-13
feat: Make an Executor available through VaadinService

### LINKED ISSUE #21366: Make an Executor available through VaadinService
### Describe your motivation

Occasionally, there's a need to dispatch some task asynchronously. The specific case on my mind is for the new signals library where we want to invoke effect callbacks asynchronously so that the user posting a message to a chat room don't have to wait until all subscribers have been notified since that might take a while if there are e.g. 1000 of them.

Add `public Executor getExecutor()` to `VaadinService`.

The application developer can implement a service init listener that sets a custom executor through the init event (so there should not be any setter in `VaadinService` to allow setting an executor after initialization). In that case, the developer is responsible for shutting down the executor.

If no executor is provided by the application developer, then a default is chosen based on what's available in the environment:
* When running in a Spring environment and a `TaskExecutor` bean is available, then it is used.
* When running in a CDI environment and a `ManagedExecutorService` bean is available, then it is used.
* If no appropriate bean is available, then Vaadin creates a single thread executor with a thread factory that sets an appropriate thread name. The executor is configured to be shut down when the Vaadin service is destroyed. If this executor is used, then a warning should be logged the first time the executor is used. The purpose of the warning is to inform the developer that the application uses a default implementation that might not be tuned for production use.

We could consider further simplifying configuration for Spring and CDI environments by defining a custom bean name, e.g. `vaadinExecutor` that will be checked for and used in case the application defines a bean with that name.

mcollovati:
Provides access to an Executor instance from VaadinService to submit asynchronous tasks. The default single-threaded executor can be replaced by a custom instance by registering it with a VaadinServiceInitListener. The spring add-on tries to detect an existing TaskExecutor bean, otherwise it falls back to the default executor.
To provide a specific TaskExecutor bean, different from the application default, the bean definition can be named VaadinTaskExecutor or be annotated with `@VaadinTaskExecutor`.

Closes #21366


vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java
```diff
@@ -90,6 +95,59 @@ protected Optional<Instantiator> loadInstantiators()
...
+        }
+
+        String message = String.format("Multiple TaskExecutor beans found: %s. "
+                + "Please resolve this conflict by either: "
+                + "(1) Providing a single TaskExecutor bean, or "
+                + "(2) Marking the bean to use with Vaadin by: "
+                + "naming it '%s' (e.g. @Bean(\"%s\")), or "
+                + "applying the @%s qualifier annotation to the bean definition. "
+                + "Note: To prevent a custom executor from replacing the default one, "
+                + "consider setting '@Bean(defaultCandidate=false)' on your bean definition.",
```
mshabarov, line 145: Not sure I understand, isn't it so that the custom executor should replace the default, if detected? Could you explain please?

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -503,6 +528,95 @@ public Instantiator getInstantiator() {
...
+                corePoolSize, Integer.MAX_VALUE, keepAliveTimeSec,
+                TimeUnit.SECONDS, new LinkedBlockingQueue<>(),
+                new VaadinThreadFactory());
+        // Enables dynamic growing and shrinking of the pool.
+        threadPoolExecutor.allowCoreThreadTimeOut(true);
+        return threadPoolExecutor;
+    }
+
+    /**
+     * Gets the executor instance used for managing concurrent tasks.
```
mshabarov, line 597: This could say that it is used by Vaadin and can be used by application.

mshabarov (commented):
LGTM, just a small suggestion about javadoc and defaultCandidate.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java
mcollovati: Definitely, the message is not clear enough.

The meaning is that if you define an `Executor` bean, it will overwrite the default task executor provided by Spring Boot. If you just want to define an executor specific for Vaadin, you can mark it as `@Bean(defaultCandidate=false)` to prevent it from overriding the Spring Boot one.

mshabarov: Super clear now, thanks. Could you change the message?

mcollovati: Could the following message be a better choice?
```
                + "Note: To prevent an Executor bean from replacing the default one "
                + "provided by Spring, consider setting '@Bean(defaultCandidate=false)' on its definition.",
```

mshabarov: Yes, sounds good

mcollovati: done

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
mcollovati: done

flow-server/src/main/java/com/vaadin/flow/server/ServiceInitEvent.java
```diff
@@ -107,6 +110,23 @@ public void addVaadinRequestInterceptor(
         addedVaadinRequestInterceptors.add(vaadinRequestInterceptor);
     }
 
+    /**
+     * Sets the {@link Executor} to be used by the service.
```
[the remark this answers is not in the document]
mcollovati: done

mshabarov (approved):

PR #21436, opened 2025-05-16
Add VaadinUser for accessing common user attributes.

peholmst:
## Description

The Walking Skeleton comes with a parent layout that is intended to be production ready. It also contains a user menu that shows the name of the current user, an avatar, a link to the profile page, and a logout item. It is currently a mock, but we'd like it to be fully implemented.

Because security can be implemented in many different ways, we don't want the parent layout to worry about it. `AuthenticationContext` is a good start, but it only gives the principal name. When using OIDC, such as via Control Center, this is an UUID. This PR aims to extend the API of `AuthenticationContext` with additional user information, such as email, full name, profile image, profile page, time zone, and locale.

This is done by introducing a new `VaadinUser` interface. 

If the principal is an OIDC principal, there is a built-in implementation of this interface that extracts the necessary claims from the token. 

If the principal implements the interface, it is returned directly. You can achieve this by e.g. creating your own `UserDetailsService` and have it return objects that also implement `VaadinUser`.

There is also a fallback implementation for principals that don't fit these criteria.

With this API in place, the Walking Skeleton can contain a working user menu that works both with Control Center, and with a custom security configuration.

This PR is marked as a draft because it does not contain any tests yet, and the JavaDocs could be improved. Before proceeding, I want to discuss whether this is the right way to go.

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
@@ -97,6 +97,18 @@ public class AuthenticationContext {
     }
 
     /**
+     * Gets an {@link Optional} with an instance of the current user if it has
+     * been authenticated, or empty if the user is not authenticated.
+     *
+     * @return an {@link Optional} with the current authenticated user, or empty
+     *         if none available
+     */
+    public Optional<VaadinUser> getAuthenticatedUser() {
+        return getAuthentication()
+                .map(VaadinUserFactory::createVaadinUser);
+    }
+
+    /**
      * Gets an {@link Optional} containing the authenticated principal name, or
      * an empty optional if the user is not authenticated.
      *
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/OidcVaadinUser.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/OidcVaadinUser.java
@@ -1,0 +1,90 @@
+package com.vaadin.flow.spring.security;
+
+import org.jspecify.annotations.Nullable;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.core.oidc.StandardClaimAccessor;
+
+import java.time.ZoneId;
+import java.util.Locale;
+import java.util.Optional;
+
+/**
+ * Implementation of {@link VaadinUser} that assumes the authenticated principal implements
+ * {@link StandardClaimAccessor}.
+ */
+final class OidcVaadinUser implements VaadinUser {
+
+    private final String userId;
+    private final String email;
+    private final String fullName;
+    private final String pictureUrl;
+    private final String profileUrl;
+    private final ZoneId timeZone;
+    private final Locale locale;
+
+    /**
+     * Creates a new {@code OidcVaadinUser}.
+     *
+     * @param authentication the authentication object to use for extracting the user information.
+     * @throws ClassCastException if the authentication principal is not an instance of {@link StandardClaimAccessor}.
+     */
+    OidcVaadinUser(Authentication authentication) {
+        var claimAccessor = (StandardClaimAccessor) authentication.getPrincipal();
+        var subject = claimAccessor.getSubject();
+
+        userId = subject != null ? subject : authentication.getName();
+        email = claimAccessor.getEmail();
+        fullName = claimAccessor.getFullName();
+        pictureUrl = claimAccessor.getPicture();
+        profileUrl = claimAccessor.getProfile();
+        timeZone = parseZoneId(claimAccessor.getZoneInfo());
+        locale = parseLocale(claimAccessor.getLocale());
+    }
+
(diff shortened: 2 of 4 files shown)


peholmst:
@mshabarov @Legioth @Artur- @heruan Please comment.

knoobie:
Imho: I'm against this - it feels like this should not belong to Vaadin (Flow). Once a VaadinUser is introduced people are going to think that's the only way to interact with Vaadin, especially when newcomers discover Vaadin or LLM learn from your examples. Vaadin should use the industry standards and not create their own flavoring of classes for customer domains (read authentication/authorization)

Artur-:
> Vaadin should use the industry standards

It would be great to use standards but where can these industry standards be found, because I don't think Spring Security includes or will include this type of interface as it's not really needed to implement Spring Security itself?

So when is this interface needed? It seems to be needed if you want to make the UI work the same way regardless of how auth/auth is implemented, which sounds like a reasonable goal, e.g. whatever UI component displays user info does not have to care about if we are using Keycloak, in memory users, own database etc. 

If Vaadin does not include anything for that, won't it be included in one copy paste way or another in each and every project? 

> Once a VaadinUser is introduced people are going to think that's the only way to interact with Vaadin,

I think the naming is quite a lot off here because it is not a "Vaadin user" in any way. It is an interface to provide user details, like ChatGPT suggests

```
✅ Recommended Name: ExtendedUserDetails
Why? This clearly communicates that it's a superset of UserDetails, adding more profile-related fields.

Conventionally aligned with Spring Security’s naming.

Already a familiar pattern for developers.
```

knoobie:
> If Vaadin does not include anything for that, won't it be included in one copy paste way or another in each and every project?

Correct, that's exactly what's happening in 90% of the projects I've seen. People have custom `XYZUser` objects that either implement  Authentication, UserDetails, OIDCUser or other interfaces from Spring to please Spring Security. Each of those with application specify logic, fields and such. So don't get me wrong - I totally understand the desire for it. Including one example in skeleton would also be a must have in my opinion because every business application is going to need it, but forcing it into flow - just to please the walking sleketon feels off.

Hypothetically: Once this interface is within flow it would make sense to also use it in other projects or components - like Avatar for example could get a constructor with it.. enforcing me to implement this interface to get a good DX for Vaadin components. Which can be quite hard with all the fields it exposes / requires and Optional default implementation that are definitly going to clash with developer's code.

knoobie:
Example: I just copy pasted your code into my simplest project where we use Spring's `DefaultOidcUser` as User class - trying to implement `VaadinUser` on this already blew up because of different method signatures for full name and the other fields

peholmst:
It may very well be that this is not the right way to go. I could also put this directly into the skeleton, allowing users to change anything they want. But this is not as easy as it sounds. Let me provide some context for clarity (because I don't yet know how to solve this problem myself).

Right now, the skeleton has no security code built-in at all. You have to add it yourself, for instance by following our guides. At the moment, we have guides that show you how to set up security with an in-memory user service. I'd also like to write guides for using OIDC, your own user database, and the new Vaadin Control Center (which is using Keycloak and OIDC).

So which one is going to go into the skeleton, and which options do you have to implement yourself? 

Can we assume and recommend that all Vaadin applications should use OIDC for authentication? If so, we need to improve the developer experience a lot and make it super-easy to either mock or fire up a lightweight identity provider. I don't think requiring users to install Kubernetes and Control Center, or set up Keycloak themselves just to get a login screen in their app is acceptable.

peholmst:
This discussion may better suited for https://github.com/vaadin/walking-skeleton/pull/41 than this PR, but since it's already started:

It would be great if you could download a skeleton, build it, deploy it to Control Center, enable identity management and then you'd have authentication, a working user menu, and logout. It has a "wow" effect. However, we need to be careful about what we put into the skeleton and why. The code in the skeleton should fall into one of three categories:

- Production ready code. You can use it as-is in your real applications, or extend as you see fit.
- Placeholder code. You have to implement some parts yourself.
- Example code. This code demonstrates certain concepts and you should delete it once you understand it.

Into which category should the security code go?

knoobie:
I probably don't know the full picture / problem, but my first suggestion would be to provide this common Interface within your walking skeleton and provide different implementation for this interface which you can guide users through - the interface can be populated by default and the implementation based on the route / way the user goes, e.g. he choose "Keycloak" and is redirected to the branch with the oidc class.. or he choose "simple user" and is redirected to branch with the normal user

peholmst:
Anyway, let's not discuss the skeleton here. If adding a `VaadinUser` (I know the name has to be better) like this to Flow is not a good idea, then it's not a good idea. We can drop it, and I'll try to find another solution to the root problem.

heruan:
> @mshabarov @Legioth @Artur- @heruan Please comment.

Commented here: https://github.com/vaadin/walking-skeleton/pull/41#discussion_r2093353031

peholmst:
This is not the right way to go. Closing.

PR #21438, opened 2025-05-16
fix!: Add setter helpers to DownloadEvent and new method for inline

mshabarov:
Refactors the `DownloadEvent` to be a normal class.
Removed the getters for file name, type and size and add setters instead that are basically helper methods for setting various attributes to a response.
Adds inline method so that you can call `DownloadHandler.forFile(new File("path/to/file")).inline()` and the download will be inlined onto page.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/ClassDownloadHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/ClassDownloadHandler.java
@@ -90,6 +90,9 @@ public class ClassDownloadHandler
         try (OutputStream outputStream = downloadEvent.getOutputStream();
                 InputStream inputStream = clazz
                         .getResourceAsStream(resourceName)) {
+            downloadEvent.setContentType(getContentType(getUrlPostfix(),
+                    downloadEvent.getResponse()));
+            downloadEvent.setFileName(getUrlPostfix());
             TransferProgressListener.transfer(inputStream, outputStream,
                     getTransferContext(downloadEvent), getListeners());
         } catch (IOException ioe) {
@@ -99,9 +102,6 @@ public class ClassDownloadHandler
             notifyError(downloadEvent, ioe);
             throw new UncheckedIOException(ioe);
         }
-
-        downloadEvent.getResponse()
-                .setContentType(downloadEvent.getContentType());
     }
 
     @Override
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/ClassDownloadHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/ClassDownloadHandler.java
@@ -100,7 +100,7 @@ public class ClassDownloadHandler
             String resourceName = getUrlPostfix();
             downloadEvent.setContentType(
                     getContentType(resourceName, downloadEvent.getResponse()));
-            if (isAttachment()) {
+            if (!isInline()) {
                 downloadEvent.setFileName(resourceName);
             }
             TransferProgressListener.transfer(inputStream, outputStream,
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/ClassDownloadHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/ClassDownloadHandler.java
@@ -44,6 +44,10 @@ public class ClassDownloadHandler
     /**
      * Create a class resource download handler with the resource name as the
      * url postfix (file name).
+     * <p>
+     * The downloaded file name and download URL postfix will be set to
+     * <code>resourceName</code>. If you want to use a different file name, use
+     * {@link #ClassDownloadHandler(Class, String, String)} instead.
      *
      * @param clazz
      *            class to use for getting resource
@@ -57,6 +61,9 @@ public class ClassDownloadHandler
     /**
      * Create a class resource download handler with the given file name as the
      * url postfix.
+     * <p>
+     * The downloaded file name and download URL postfix will be set to
+     * <code>fileName</code>.
      *
      * @param clazz
      *            class to use for getting resource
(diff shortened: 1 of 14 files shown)


flow-server/src/main/java/com/vaadin/flow/server/streams/ClassDownloadHandler.java
```diff
@@ -90,6 +90,9 @@ public void handleDownloadRequest(DownloadEvent downloadEvent) {
         try (OutputStream outputStream = downloadEvent.getOutputStream();
                 InputStream inputStream = clazz
                         .getResourceAsStream(resourceName)) {
+            downloadEvent.setContentType(getContentType(getUrlPostfix(),
+                    downloadEvent.getResponse()));
+            downloadEvent.setFileName(getUrlPostfix());
```
mshabarov, line 95: I'm not confident that it is a good default to set the file name as a resource name, might be something to hide instead.

mshabarov: The same comment about exposing resource name in the getPostfixUrl.

flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
```diff
@@ -111,21 +129,49 @@ public VaadinSession getSession() {
...
      *
-     * @return file name
+     * @param fileName
+     *            the name to be assigned to the file
      */
-    public String getFileName() {
-        return fileName == null ? "" : fileName;
+    public void setFileName(String fileName) {
+        response.setHeader("Content-Disposition",
+                "attachment;filename=" + fileName);
```
knoobie, line 143: I was thinking about adding overloaded methods to support inline vs attachment as well as encoding but skipped since it's just a convenient helper for people. 

I would still suggest to change the filename to use "" around it so that spaces in the filename are correctly supported. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition for more details 

It's up to you if you also wanna support filename* by default since Vaadin is a globally used framework

mshabarov: Good comment! I've added the "" wrapping. filename* would make sense to do, but I'd prefer it in a separate PR.

flow-server/src/main/java/com/vaadin/flow/server/streams/ServletResourceDownloadHandler.java
```diff
@@ -66,22 +74,26 @@ public ServletResourceDownloadHandler(String path, String name) {
...
+                        .setContentType(getContentType(resourceName, response));
+                if (isAttachment()) {
+                    downloadEvent.setFileName(resourceName);
+                }
                 TransferProgressListener.transfer(inputStream, outputStream,
                         getTransferContext(downloadEvent), getListeners());
             } catch (IOException ioe) {
                 // Set status before output is closed (see #8740)
-                downloadEvent.getResponse().setStatus(
+                response.setStatus(
```
caalador, line 92: Should be `downloadEvent.setStatus(int)` but can be in other PR just wouldn't change this here in that case.

mshabarov: Let's add this helper separately.

flow-server/src/main/java/com/vaadin/flow/server/streams/InputStreamDownloadHandler.java
```diff
@@ -67,6 +70,12 @@ public InputStreamDownloadHandler(
 
     @Override
     public void handleDownloadRequest(DownloadEvent downloadEvent) {
+        String resourceName = getUrlPostfix();
```
caalador, line 73: resourceName should be taken from the DownloadResponse as it is dynamically determined when handler.appy(downloadEvent) is called.

flow-server/src/main/java/com/vaadin/flow/server/streams/InputStreamDownloadHandler.java
```diff
@@ -67,6 +70,12 @@ public InputStreamDownloadHandler(
...
+        downloadEvent.setContentType(
+                getContentType(resourceName, downloadEvent.getResponse()));
+        if (isAttachment()) {
+            downloadEvent.setFileName(resourceName);
+        }
```
caalador: These should be set after the hasError check.

flow-server/src/main/java/com/vaadin/flow/server/streams/InputStreamDownloadHandler.java
```diff
@@ -67,6 +70,12 @@ public InputStreamDownloadHandler(
 
     @Override
     public void handleDownloadRequest(DownloadEvent downloadEvent) {
+        String resourceName = getUrlPostfix();
+        downloadEvent.setContentType(
+                getContentType(resourceName, downloadEvent.getResponse()));
+        if (!isInline()) {
+            downloadEvent.setFileName(resourceName);
+        }
```
caalador, line 78: These should be set after the hasError check

caalador (changes requested):

flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
```diff
@@ -111,21 +129,50 @@ public VaadinSession getSession() {
...
-     * @return file name
+     * @param fileName
+     *            the name to be assigned to the file
      */
-    public String getFileName() {
-        return fileName == null ? "" : fileName;
+    public void setFileName(String fileName) {
+        fileName = fileName != null ? fileName : "";
+        response.setHeader("Content-Disposition",
+                "attachment; filename=\"" + fileName + "\"");
```
caalador, line 144: This should be
```
if (fileName.isEmpty()) {
} else {
  response.setHeader("Content-Disposition",
                "attachment; filename=\"" + fileName + "\"");
}
```

mshabarov: Done

caalador (approved):

PR #21456, opened 2025-05-19
fix: Inline download in data html elements

mshabarov:
Automatically sets pre-defined download handlers to inline in data html components.

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
@@ -26,6 +26,7 @@ import com.vaadin.flow.component.PropertyDescriptor;
 import com.vaadin.flow.component.PropertyDescriptors;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.server.AbstractStreamResource;
+import com.vaadin.flow.server.streams.AbstractDownloadHandler;
 import com.vaadin.flow.server.streams.DownloadHandler;
 import com.vaadin.flow.server.StreamResource;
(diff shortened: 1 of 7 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
```diff
@@ -294,12 +292,17 @@ public void setData(AbstractStreamResource data) {
...
      *
      * @param data
-     *            a "data" attribute value,, not {@code null}
+     *            a "data" attribute value, not {@code null}
      */
     public void setData(DownloadHandler data) {
+        if (data instanceof AbstractDownloadHandler<?> handler) {
+            // change disposition to inline in pre-defined handlers,
+            // where it is 'attachment' by default
+            handler.inline();
```
Legioth, line 304: Might be good to explicitly document this behavior both for the setter and for the constructor parameter. (And the same obviously also applies for the other affected components)

mshabarov: added.

caalador: Perhaps `AbstractDownloadHandler implementations` would be better as you could extend that yourself and then wonder about the inline being set even though it is a custom and not pre-defined handler.

mshabarov: Added AbstractDownloadHandler as an addition, would be better to mention the helpers anyway IMO.
Also supplemented the inline() contract to mention that the implementations should ensure the the value is inline if the method is called.

caalador (approved):

PR #21459, opened 2025-05-20
fix: Remove faulty method

caalador:
remove sendUploadResponse as ResponseHandled
is used and should be the one to
override for custom responses.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/UploadEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/UploadEvent.java
@@ -146,7 +146,7 @@ public class UploadEvent {
     /**
      * Get the set file name for current upload.
      * <p>
-     * The file name is always {@code unknown} for filexhr.
+     * The file name will only be available for multipart uploads.
      *
      * @return file name
      */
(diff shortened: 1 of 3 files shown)


flow-server/src/main/java/com/vaadin/flow/server/streams/UploadEvent.java
```diff
@@ -170,7 +144,9 @@ public VaadinSession getSession() {
     }
 
     /**
-     * Get the set file name.
+     * Get the set file name for current upload.
+     * <p>
+     * The file name is always {@code unknown} for filexhr.
```
Legioth, line 149: I'm not sure anyone knows what "filexhr" means

caalador: woudl `XMLHttpRequestUpload` be better?

Legioth: Easier to understand but not 100% accurate since you can also just do a regular `fetch` and manually populate the body.

How about turning it the other way: name is available only for multipart uploads

caalador: done

mshabarov (approved):

PR #21467, opened 2025-05-20
chore: Update byte-buddy version

TatuLund:


mcollovati:
Can this update somehow clash with Spring Boot 2.7 that is currently using 1.12.23?

TatuLund:
> Can this update somehow clash with Spring Boot 2.7 that is currently using 1.12.23?

If so, we need to add doc that instructs to exclude it. I checked that Spring Boot 2.7 is claimed to be Java 21 compatible, Java 21 support was added in 1.12.22. So if you plan to use say Java 24 or 25, it needs to be updated.

mcollovati:
Did you perhaps already test this change against the vaadin-spring 15.2?
If not, we can try, just to be sure there are no errors.
Otherwise we could update to 1.12.23 to be consistent with Spring Boot 2.7

TatuLund:
> Did you perhaps already test this change against the vaadin-spring 15.2?

Not yet.

> If not, we can try, just to be sure there are no errors.

Please do.

mcollovati:
It looks like there are no issues running a Vaadin 14 + Spring Boot 2.7 application with the latest byte-buddy version.

However, in a regular Spring Boot project, the byte-buddy dependency gets managed by Spring, so it ends-up in version 1.12.23.

So, we can go with the latest or just align with Spring and use 1.12.23.

mcollovati (approved):

PR #21470, opened 2025-05-21
doc: Add note about need to use setKey in Grid for getting sortOrders in Query object

TatuLund:


flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
```diff
@@ -170,6 +170,9 @@ public int getPageSize() {
      * <strong>Note: </strong> Sort orders and in-memory sorting are mutually
      * exclusive. If the {@link DataProvider} handles one, it should ignore the
      * other.
+     * <p>
+     * <strong>Note: </strong> Sort orders are populated only with the
+     * properties in Grid that have key set.
```
mshabarov, line 175: I have no idea on what should I put my attention as a Vaadin user.
What are the properties? What is the key set? Why this generic API tells about Grid?
Could we explain it a bit in a human readable way explaining what's the problem and how should users solve it?
Would be also good to put links (or at least via <code></code>) to reference related classes/methods/fields.

TatuLund: I tried to keep it short in JavaDoc.

Basically you need to do something like this

`        nameCol = treeGrid.addHierarchyColumn(ManagedPerson::getFullName)
                .setHeader("Full name")
                .setSortable(true).setKey("fullName");`
                
Or 

`        grid.setColumns("firstName", "lastName", "age", "phoneNumber",
                "maritalStatus");
         grid.setSortableColumns("firstName", "lastName", "age");`

Then Grid has key set for the column, and sort orders will be populated. In the later case `setColumns` will internally call `setKey`.

Linking to Grid in JavaDoc does not work as it is in different repo.

TatuLund: I reformulated the sentence, perhaps it is now specific enough.

mshabarov: Now it's clearer, thanks! I'd propose a small change in wording:
```
Sort orders are populated in Vaadin Grid component only for the columns that have key set using {@code Column#setKey} or {@code Grid#setColumns} methods.
```

mshabarov (approved):

PR #21478, opened 2025-05-21
feat: create transfer util

caalador:
Create a TransferUtil for handling data transfer
for upload and download.
For upload handle xhr vs multipart part vs multipart iterator.
This makes the UploadHanlder interface
easier to handle.


flow-server/src/main/java/com/vaadin/flow/server/streams/UploadUtil.java
```diff
@@ -0,0 +1,165 @@
...
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.server.VaadinRequest;
+import com.vaadin.flow.server.VaadinResponse;
+import com.vaadin.flow.server.VaadinSession;
+
+/**
+ * Utility class for upload requests. Specifically handles checking if request
+ * is multipart of filexhr.
+ */
+public final class UploadUtil {
```
mshabarov, line 46: How about UploadAndDownloadUtil so we could place here methods that can potentially be used by develoeprs, e.g. transfer method

caalador: Renamed to TransferUtil. Moved transfer there.

flow-server/src/main/java/com/vaadin/flow/server/streams/TransferUtil.java
```diff
@@ -0,0 +1,240 @@
...
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.server.VaadinRequest;
+import com.vaadin.flow.server.VaadinResponse;
+import com.vaadin.flow.server.VaadinSession;
+
+/**
+ * Utility class with methods for handling transfer of upload and download
+ * requests.
+ *
+ * For internal use only. May be renamed or removed in a future release.
```
mshabarov, line 51: transfer method may be used as a helper for custom handlers.
Could we place this disclaimer only to method that we know we are not going to recommend to use?

flow-server/src/main/java/com/vaadin/flow/server/streams/TransferUtil.java
```diff
@@ -0,0 +1,240 @@
...
+     * @param request
+     *            The VaadinRequest for this upload
+     * @param response
+     *            The VaadinResponse for this upload
+     * @param session
+     *            Current VaadinSession
+     * @param owner
+     *            The element that owns the request handler
+     */
+    public static void handleUpload(UploadHandler handler,
```
mshabarov, line 139:
```suggestion
   static void handleUpload(UploadHandler handler,
```

mshabarov: Could be even a package-private?

mshabarov (approved):

mshabarov:
Fixes https://github.com/vaadin/flow/issues/21486

PR #21488, opened 2025-05-22
fix: remove name/fileName from inputStreamDownload

caalador:
Remove the possibility to give
fileName/name to inputStreamDownloadHandler
as the atual downloaded name is
decided on transfer and will in many cases
not match or be confusing to match in 2 places.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadHandler.java
@@ -242,25 +242,6 @@ public interface DownloadHandler extends ElementRequestHandler {
 
     /**
      * Generate a function for downloading from a generated InputStream with the
-     * given download name.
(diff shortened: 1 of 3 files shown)


mshabarov (approved):

Legioth:
Not sure this is the right approach since there's now not a good way of showing a nice URL to the user before the download starts. We should instead allow the developer to choose whether they provide a name up-front and then it's their responsibility to decide whether they keep the names in sync or not.

caalador:
I wouldn't say it is an especially nice URL in any case, but the name missmatch generated a lot of confusion for inputStream case where the name is set on download.

mshabarov:
Is this solution too ugly for this case when you want different names for file and postfix?
```java
        new InputStreamDownloadHandler(event -> {}) {
            @Override
            public String getUrlPostfix() {
                return "custom-url-postfix";
            }
        }
```

PR #21490, opened 2025-05-22
feat: initialize signals environment

### LINKED ISSUE #21474: Feature Flag for Signals
Add a feature flag to the list and add a check for it in the `Signal` class constructor, it should throw an exception with warning that this is the preview feature similar to "CARD_COMPONENT" flag.
The feature flag name can be "Flow Full-stack Signals"
The reference can be https://github.com/vaadin/platform/issues/7373.

### LINKED ISSUE #21475: Add Signal environment initialisation
Add the Signals environment initialisation code into `VaadinService` that also sets up the Signals dispatcher to run signal effects immediately if the change is for current UI and run asynchronously otherwise.
```java
             // get default or custom executor
            // executor = ...
            // create a new ObjectMapper with timing module until we have a singleton instance with Hilla
            SignalEnvironment.tryInitialize(objectMapper, executor);

            // Setup specific for Flow
            Executor dispatcher = SignalEnvironment.defaultDispatcher();
            SignalEnvironment.addDispatcherOverride(() -> {
                UI owner = UI.getCurrent();
                if (owner == null) {
                    return null;
                }

                return task -> {
                    if (UI.getCurrent() == owner) {
                        task.run();
                    } else {
                        try {
                            dispatcher.execute(() -> owner.access(task::run));
                        } catch (Exception e) {
                            // a task is submitted when executor is shut down, ignore
                        }
                    }
                };
            };
```

This initialisation can also be skipped if the Signlas feature flag is OFF, but depends on whether we can read the feature flag at that point.

mcollovati:
Configures Signals environment when VaadinService is initialized, providing Vaadin defaults Executor and ObjectMapper.
Signals functionality is behind a feature flag; if the flag is not set, an exception is thrown at runtime if Signals are used.

Closes #21474
Closes #21475

### DIFF
--- a/flow-server/pom.xml
+++ b/flow-server/pom.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>com.vaadin</groupId>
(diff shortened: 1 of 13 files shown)


mshabarov (dismissed):
Looks good to me and works as expected as well.

About tests: some use static mocks for feature flags, some set the system property. I even think that we can keep them as you suggested, not much concerned about it.

mshabarov (approved):

mshabarov:
We are approaching beta milestone, thus I think we should merge this and make minor changes to flow-bom and tests later, if needed. The feature works as expected.

PR #21491, opened 2025-05-22
chore: Update byte-buddy version

mcollovati:


mcollovati:
Byte-buddy has been updated in 2.12, so we need to update it also in 23.6

mshabarov (approved):

PR #21498, opened 2025-05-22
feat: Add onAllComplete listener

mshabarov:
Adds onAllComplete listener to `TransferProgressListener`.


flow-server/src/main/java/com/vaadin/flow/server/streams/AbstractDownloadHandler.java
```diff
@@ -47,6 +53,18 @@ protected String getContentType(String fileName, VaadinResponse response) {
                 .orElse("application/octet-stream");
     }
 
+    @Override
+    public void handleRequest(VaadinRequest request, VaadinResponse response,
+            VaadinSession session, Element owner) throws IOException {
+        DownloadHandler.super.handleRequest(request, response, session, owner);
+        if (this instanceof FileDownloadHandler downloadHandler) {
```
mshabarov, line 60: This is ugly but I found no better way so far.

Legioth:
Seems to me like this would belong in the `Upload` component instead?

mshabarov:
Actually yes. I was wrong about `upload.addAllFinishedListener` - thought that it is based on old stream implementation, but looks like it is based on client-server upload intercommunication/events and can be just used as is with the new API. Excellent! But I need to test it just in case.

mshabarov:
Yes, this codes shows all the notifications as expected:
```java
        Upload upload = new Upload(UploadHandler.inMemory((meta, bytes) -> {
            Notification.show("File name: " + meta.fileName());
        }));
        upload.addAllFinishedListener(event -> {
            Notification.show("All Upload finished");
        });
        add(upload);
```

PR #21503, opened 2025-05-23
feat: Make current user available without an AuthenticationContext instance

Artur-:
Makes it possible to write an application method like

```
public static Optional<User> getCurrentUser() {
    return AuthenticationContext.getAuthenticatedUser(AppUserInfo.class).map(AppUserInfo::user);
}
```

instead of re-implementing everything from AuthenticationContext

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/AuthenticationContext.java
@@ -111,7 +111,8 @@ public class AuthenticationContext {
      *             if the current user instance does not match the given
      *             {@code userType}.
      */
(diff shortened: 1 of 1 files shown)


mcollovati (commented):
This could be a breaking change in the rare case someone extended `AuthenticationContext` overriding the `getAuthenticatedUser()` method (`AuthenticationContext` bean definition has `@ConditionalOnMissingBean`).

Could we instead introduce a new static method with a different name, e.g. `currentAuthenticatedUser(...)`?

Artur-:
Now adds a `getCurrentAuthenticatedUser` method instead

mcollovati (dismissed):
Looks good.
Only needs formatting fix

mcollovati (approved):

Issue #21487, opened 2025-05-21
UI is not updated by transfer progress logic in Firefox


mshabarov:
### Description of the bug

I want to update my UI while download is ongoing:
```java
DownloadHandler.fromInputStream(event -> new DownloadResponse(
                        getSlowInputStream(), "slow-download.bin", "application/octet-stream", SLOW_CONTENT_SIZE))
                .whenStart(() -> {
                    Notification.show("Download started", 3000, Notification.Position.BOTTOM_START);
                    progressBar.setVisible(true);
                })
                .onProgress((transferred, total) -> progressBar.setValue((double) transferred / SLOW_CONTENT_SIZE))
                .whenComplete(success -> {
                    progressBar.setVisible(false);
                    if (success) {
                        Notification.show("Download completed", 3000, Notification.Position.BOTTOM_START).addThemeVariants(
                                NotificationVariant.LUMO_SUCCESS);
                    } else {
                        Notification.show("Download failed", 3000, Notification.Position.BOTTOM_START)
                                .addThemeVariants(NotificationVariant.LUMO_ERROR);
                    }
                });
```

This works as expected in Chrome and Safari, but not in Firefox.
Server shows errors related to Push/Atmosphere/WebSockets:
```
2025-05-21T17:11:29.190+03:00 ERROR 12901 --- [phere-Shared-37] c.v.f.s.c.PushAtmosphereHandler          : Exception in push connection

java.io.IOException: Connection remotely closed for 310fcb66-7cfc-4de3-afa2-2a2d8aab6c7c
        at org.atmosphere.websocket.WebSocket.write(WebSocket.java:237) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
        at org.atmosphere.websocket.WebSocket.write(WebSocket.java:227) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
        at org.atmosphere.websocket.WebSocket.write(WebSocket.java:48) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
        at org.atmosphere.cpr.AtmosphereResponseImpl$Stream.write(AtmosphereResponseImpl.java:955) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
[...]

PR #21512, opened 2025-05-23
fix: set download attribute for Anchor DownloadHandlers

caalador:
Part of #21487

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -27,6 +27,7 @@ import com.vaadin.flow.component.PropertyDescriptor;
 import com.vaadin.flow.component.PropertyDescriptors;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.server.AbstractStreamResource;
+import com.vaadin.flow.server.streams.AbstractDownloadHandler;
 import com.vaadin.flow.server.streams.DownloadHandler;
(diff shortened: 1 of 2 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -202,6 +203,9 @@ public void setHref(AbstractStreamResource href) {
      * Sets the URL that this anchor links to and that is bound to a given
      * {@link DownloadHandler} callback on the server for handling data download
      * from the server to the client when clicking an anchor.
+     * <p>
+     * Sets the 'download' attribute for link when given a non-inline handler
+     * implementing AbstractDownloadHandler.
```
knoobie, line 208: I'm wondering if this should also be added to the constructor using this setter? Yes, the constructor also says "see setHref" but it might be good for people to see when they instantiate the class

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -211,6 +215,10 @@ public void setHref(DownloadHandler downloadHandler) {
                 downloadHandler, this.getElement());
         setRouterIgnore(true);
         assignHrefAttribute();
+        if (downloadHandler instanceof AbstractDownloadHandler<?> abstractDownloadHandler
+                && !abstractDownloadHandler.isInline()) {
+            getElement().setAttribute("download", true);
```
mshabarov, line 220: Needs a unit test

PR #21522, opened 2025-05-26
fix: do not generate list for empty string

### LINKED ISSUE #21509: Build Parameter projectFileExtensions is evaluated differently
### Description of the bug

I noticed and sadly couldn't reliable reproduce in simple projects that our `vite.generated.ts` file is constantly updated because of the "newly" released property `projectFileExtensions`. (We don't use the property)

If we run `mvn clean install` the `.` gets removed from the generated file resulting in a git diff because we have commited this file to keep track on any changes. 

If we run / execute our application the `.` gets added again. (Start is done via spring-boot:run)

Image

### Expected behavior

The property `projectFileExtensions` should only be added if it is != null AND is not empty. Empty default list should not result in a difference in the vite.generated.ts file.

.

### Versions

- Vaadin / Flow version: 24.6?++

caalador:
Do not generate a 1 item
list for the default empty
string in dev mode init
for file extensions.

Fixes #21509

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java
@@ -370,12 +370,12 @@ public class DevModeInitializer implements Serializable {
         }
     }
 
-    private static List<String> getFrontendExtraFileExtensions(
+    static List<String> getFrontendExtraFileExtensions(
             ApplicationConfiguration config) {
-        List<String> stringProperty = Arrays.asList(config
+        List<String> stringProperty = Arrays.stream(config
                 .getStringProperty(InitParameters.FRONTEND_EXTRA_EXTENSIONS, "")
-                .split(","));
-        return stringProperty;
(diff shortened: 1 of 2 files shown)


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/startup/DevModeInitializer.java
```diff
@@ -370,12 +370,12 @@ public static DevModeHandler initDevModeHandler(Set<Class<?>> classes,
...
 
-    private static List<String> getFrontendExtraFileExtensions(
+    static List<String> getFrontendExtraFileExtensions(
             ApplicationConfiguration config) {
-        List<String> stringProperty = Arrays.asList(config
+        List<String> stringProperty = Arrays.stream(config
                 .getStringProperty(InitParameters.FRONTEND_EXTRA_EXTENSIONS, "")
-                .split(","));
-        return stringProperty;
+                .split(",")).filter(input -> !input.isEmpty()).toList();
```
mcollovati, line 377: nit: what about using `isBlank()` instead of `isEmpty()`?

mcollovati (dismissed):

PR #21523, opened 2025-05-26
feat!: Use React 19

Artur-:
Removes the React 19 feature flag and always uses React 19


mshabarov (dismissed):

mshabarov:
⚠️ LGTM, but let's not merge before Vaadin Platform 25.0.0-alpha4 (to not complicate the release process with one more breaking change in the list)

PR #21524, opened 2025-05-26
feat: auto-reload service worker in dev mode

platosha:


taefi (changes requested):
Generally, LGTM! Just asked for a couple of small changes.

One thing to consider is the copilot. I personally not sure whether the copilot counts on the previous behavior (no auto-reload happening for service worker), and this could break it. Maybe a second review from @Artur- could be beneficial?

flow-tests/test-frontend/vite-pwa-custom-sw/src/test/java/com/vaadin/viteapp/MainIT.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private void reloadPage() {
+        executeScript("window.location.reload()");
+        waitUntil(webDriver -> ((JavascriptExecutor) driver)
+                .executeScript("return document.readyState")
+                .equals("complete"));
+    }
+
+    private void openPage(String url) {
```
platosha, line 88:
```suggestion
    private void openAndReload(String url) {
```

PR #21538, opened 2025-05-27
fix: UI available from event in upoad handling.

caalador:
Do not set  UI current from uri UI id, but store uiid to 
possible handler component data and get UI from 
session if component.getUI doesn't return UI.


Legioth:
If the session is unlocked before calling `handleRequest`, then why isn't `UI.current` also cleared before?

caalador:
To have the correct ui available during upload handling.
It's been this way since 7.1.0.beta1 for FileUploadHandling.

mshabarov:
The only danger that I see to not populate the UI into thread local is possible `UI.getCurrent()` calls in the `Upload` component. The application code can use `event.getUI`. The current apps shouldn't break because they use old API. I'd clean it up and see if this breaks component or apps and then would investigate why.

PR #21558, opened 2025-05-29
fix: refine request caching rules

### LINKED ISSUE #21394: VaadinWebSecurity: allow to configure restrictions for saved requests URL
### Describe your motivation

When using `VaadinWebSecurity` to secure a Spring-based application, a custom `RequestCache` implementation is configured to prevent storing URL for Vaadin internal requests. The implementation then delegates the logic to `HttpSessionRequestCache`.

The request cache might store a protected resource URL.  This URL could be an image or JavaScript file.  After login, it redirects to this URL instead of the main view.

Usually, those resources should be publicly accessible, but some might still be protected and accessible only after authentication for security reasons.

By default, Spring Security sets a set of URL restrictions to `HttpSessionRequestCache` that are unfortunately not applied by `VaadinWebSecurity`. However, other URLs, such as `.well-known/*` needs manual configuration.

* `VaadinWebSecurity` should delegate to the `HttpSessionRequestCache` instance provided by Spring Security instead of creating a new one.
* `VaadinWebSecurity` should allow defining a `RequestMatcher` to restrict the URLs eligible for request cache.

Manually configure `RequestCache` in the application.

### Additional context

N/A

### LINKED ISSUE #21496: User is redirected to /.well-known/appspecific/com.chrome.devtools.json after login when Chrome dev tools are open
### Description of the bug

If you have devtools open in Chrome, Chrome sends a request to http://localhost:8088/.well-known/appspecific/com.chrome.devtools.json even though it is not visible in the network tab. This request gets stored as the "most recent request" with Spring Security and the user is redirected to this non-existing URL on login

### Expected behavior

The url http://localhost:8088/.well-known/appspecific/com.chrome.devtools.json should automatically be ignored by the default `VaadinSecurityConfiguration`

Create a new app with authentication
Log in with dev tools open

### Versions

Flow 24.8.0.alpha13

mcollovati:
Updates VaadinDefaultRequestCache to skip caching requests based on default and custom rules, including matching media types, headers, and request paths such as `/.well-known/...`.
Refactors request handling utility logic for reusability.

Fixes #21496
Closes #21394

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
@@ -15,18 +15,30 @@
  */
 package com.vaadin.flow.spring.security;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
-
-import com.vaadin.flow.server.HandlerHelper;
-
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.MediaType;
 import org.springframework.security.web.savedrequest.HttpSessionRequestCache;
 import org.springframework.security.web.savedrequest.RequestCache;
 import org.springframework.security.web.savedrequest.SavedRequest;
+import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
+import org.springframework.security.web.util.matcher.MediaTypeRequestMatcher;
+import org.springframework.security.web.util.matcher.OrRequestMatcher;
+import org.springframework.security.web.util.matcher.RequestHeaderRequestMatcher;
+import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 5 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
```diff
@@ -125,4 +171,53 @@ private boolean isServiceWorkerInitiated(HttpServletRequest request) {
...
+        this.ignoreRequestMatcher = requestMatcher;
+    }
+
+    /*
+     * Rules adapted from Spring Security's RequestCacheConfigurer
+     */
+    private static RequestMatcher createDefaultIgnoreRules() {
+        RequestMatcher favIcon = new AntPathRequestMatcher("/**/favicon.*");
+        RequestMatcher wellKnown = new AntPathRequestMatcher(
+                "/**/.well-known/**");
```
knoobie, line 194: Just fyi: `**/something/**` won't work with the replacement API for the deprecated AntPathRequestMatcher

mcollovati: Thanks for the info.
Now that the main branch is switched to version 25, we will make a PR to replace all the deprecated API.
For this specific case, we will again align to `RequestCacheConfigurer` and use the appropriate patterns.

mcollovati: I created #21562 as a reminder

caalador (approved):

PR #21566, opened 2025-05-30
test: Delete outdated test

Artur-:


mcollovati (changes requested):
Need to also remove `flow-tests/test-dev-mode/src/main/resources/META-INF/services/com.vaadin.base.devserver.DevToolsMessageHandler`

Issue #21587, opened 2025-06-02
DownloadHandler disconnects from UI on Firefox


HerbertsVaadin:
### Description

Seems like there's a connection issue on Firefox, causing delay of notifications (at some point they still appear) and temporary UI disconnect to occur (causing "Online" text to appear in the loader.  
Occurs consistently after a couple of downloads. 
I have not had any issues with it on Chrome. 

Logged error:
```
2025-06-02T13:10:50.800+03:00 ERROR 53611 --- [phere-Shared-43] c.v.f.s.c.PushAtmosphereHandler          : Exception in push connection

java.io.IOException: Connection remotely closed for 7d9d4774-99c6-4995-bd0e-53d1e30d8a14
	at org.atmosphere.websocket.WebSocket.write(WebSocket.java:237) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
	at org.atmosphere.websocket.WebSocket.write(WebSocket.java:227) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
	at org.atmosphere.websocket.WebSocket.write(WebSocket.java:48) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
	at org.atmosphere.cpr.AtmosphereResponseImpl$Stream.write(AtmosphereResponseImpl.java:955) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
	at org.atmosphere.handler.AbstractReflectorAtmosphereHandler.onStateChange(AbstractReflectorAtmosphereHandler.java:154) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
	at com.vaadin.flow.server.communication.PushAtmosphereHandler.onStateChange(PushAtmosphereHandler.java:54) ~[flow-server-24.8.0.beta2.jar:24.8.0.beta2]
	at org.atmosphere.cpr.DefaultBroadcaster.invokeOnStateChange(DefaultBroadcaster.java:1036) ~[atmosphere-runtime-3.0.5.slf4jvaadin1.jar:3.0.5.slf4jvaadin1]
[...]

PR #21589, opened 2025-06-02
fix: Add attribute always when file name is set

mshabarov:
Related-to https://github.com/vaadin/flow/issues/21587

Always adds the attribute to &lt;a&gt; when `event.setFileName()` is called.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
@@ -132,8 +132,8 @@ public class DownloadEvent {
     /**
      * Sets the name of the file to be downloaded. This method utilizes the HTTP
      * Content-Disposition header to specify the name of the file to be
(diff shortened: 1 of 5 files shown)


flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
```diff
@@ -147,9 +148,11 @@ public void setFileName(String fileName) {
         }
         if (fileName.isEmpty()) {
             response.setHeader("Content-Disposition", "attachment");
+            getOwningElement().setAttribute("download", true);
```
mshabarov: Still it would be better to only add this to <a>, the tag can be examined perhaps.

Legioth:
`event.setFileName()` is called only after the download request has been set and it's then too late to make changes to how the browser would have sent that request.

mshabarov:
Closed in favour of https://github.com/vaadin/flow/pull/21624

PR #21610, opened 2025-06-03
chore!: Declare IOException in download/upload callbacks

mshabarov:
This allows to write the code in the callbacks without requiring `catch(IOException e)` each time.


flow-server/src/main/java/com/vaadin/flow/server/streams/FileUploadCallback.java
```diff
@@ -0,0 +1,49 @@
...
+     *
+     * @param metadata
+     *            the upload metadata containing relevant information about the
+     *            upload
+     * @param file
+     *            the file to which the data will be written
+     * @throws IOException
+     *             if an I/O error occurs in the callback
+     */
+    void apply(UploadMetadata metadata, File file) throws IOException;
```
caalador: Would perhaps prefer `complete`, `completed` or `uploadComplete` as there is no need to name it `apply` which doesn't really tell what this is for. I would opt for a more descriptive method name.
Same for the other callbacks that use `apply`

mshabarov: Renamed

PR #21611, opened 2025-06-03
fix: Deduce content type from file name

mshabarov:


flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadResponse.java
```diff
@@ -52,16 +54,26 @@ public class DownloadResponse implements Serializable {
      *            Content-Disposition header to 'attachment' if the value is not
      *            <code>null</code>, otherwise the header is not set
      * @param contentType
-     *            content type
+     *            content type or a value obtained from
+     *            {@link VaadinService#getMimeType(String)} if the given
+     *            parameter is <code>null</code> and <code>fileName</code> is
+     *            not <code>null</code>
```
caalador: This feels like the explanation should not be for the param, but instead in the main text.
```suggestion
     *            content type or to be determined for {@code null}
```

flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadResponse.java
```diff
@@ -52,16 +54,26 @@ public class DownloadResponse implements Serializable {
...
-        this.contentType = contentType;
         this.contentLength = contentLength;
+        if (contentType == null && fileName != null) {
+            this.contentType = Optional
+                    .ofNullable(
+                            VaadinService.getCurrent().getMimeType(fileName))
+                    .orElse("application/octet-stream");
+        } else {
+            this.contentType = contentType;
+        }
```
caalador: Shouldn't we for contentType== null && fileName == null always set `application/octet-stream`
Also InputStreamDownloadHandler (which is the only user of this) doesn't even take into account the set content type for the response and does the `getContentType(String, VaadinResponse)` which is the same as the `contentType== null && fileName != null` case here.
This should be updated so we would there check if content type set or else use the default.

caalador (approved):

PR #21612, opened 2025-06-03
chore!: Remove isMultipart method

mshabarov:
We don't want the multipart way of upload to leak into handlers API.
We may want to remove multipart in the future, so lets not stick with it.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/UploadEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/UploadEvent.java
@@ -212,13 +212,4 @@ public class UploadEvent {
             session.unlock();
         }
     }
-
-    /**
-     * Check if this file is part of a multipart upload request.
-     *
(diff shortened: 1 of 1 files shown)


caalador:
The idea behind the isMultipart was that you know there will be multiple calls and not just the one.
So for instance if you want to collect all files before pushing to the database in one go instead of for each file.

mshabarov:
Good example. We can keep this method, but maybe rename it a bit to decouple from "multipart request" naming and describe where it can be useful? Also make sense to add it into docs.

Legioth:
`isMultipartRequest()` on its own doesn't seem to be enough for that kind of use case. You would also need some way of knowing which file is the last one. This means that you would do finalization through `addAllFinishedListener` but then you can anyways do the finalization there regardless of whether there are multiple files in the same request or not.

caalador:
Also `responseHandled` will be called only after all multipart requests are done.

caalador (approved):

PR #21615, opened 2025-06-03
fix: prevent route collision check before dev server is started

### LINKED ISSUE #21592: Application does not startup because of route path collision after file is removed
### Description of the bug

I had `@index.tsx` and `MyView.java` with `my-view` route path. Then I decided to make MyView.java as the home page. Then I shut down the server, removed `@index.tsx` and replaced the path `my-view` to ``. I realized that application throws `InvalidRouteConfigurationException` on startup even though there are no conflicting paths. 

I investigated and realized that `file-routes.ts` is not up-to-date and has `''` configuration with `@index.tsx`.

After deleting the generated folder, the application started up fine.

### Expected behavior

The application should verify the corresponding route file when a conflict arises and should start as expected 

 - Create a full-stack app from the start.vaadin.com
 - Run the application once, so the generated folder gets created
 - Stop the application
 - Create a View with "" route.
 - Remove the `@index.tsx`
 - Start the application
You will get InvalidRouteConfigurationException 

### Versions

Hilla: 24.8.0.beta2
Flow: 24.8.0.beta2
Vaadin: 24.8.0.beta2
Spring Boot: 3.5.0
Spring: 6.2.7
Spring Security: 
Spring Data JPA: 
Copilot: 24.8.0.beta2
Frontend Hotswap: Enabled ⋅ Vite
OS: aarch64 Mac OS X 14.7.1
Java: JetBrains s.r.o. 21.0.5
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36
Java Hotswap: Java Hotswap is enabled
IDE Plugin: 1.0-SNAPSHOT IntelliJ

mcollovati:
Fixes #21592

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
@@ -346,8 +346,12 @@ public abstract class VaadinService implements Serializable {
             UsageStatistics.markAsUsed("flow/bun", null);
         }
 
-        RouteUtil.checkForClientRouteCollisions(this,
-                getRouteRegistry().getRegisteredRoutes());
+        if (getDeploymentConfiguration().isProductionMode()) {
+            // Postpone the check until dev-server is fully initialized and
+            // client routes are computed.
+            RouteUtil.checkForClientRouteCollisions(this,
+                    getRouteRegistry().getRegisteredRoutes());
+        }
 
         initialized = true;
     }
--- a/flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
+++ b/flow-server/src/main/java/com/vaadin/flow/router/internal/RouteUtil.java
@@ -656,8 +656,12 @@ public class RouteUtil {
         if (!collisions.isEmpty()) {
             String msg = String.format(
                     "Invalid route configuration. The following Hilla "
-                            + "route(s) conflict with configured Flow routes: %s",
-                    String.join(", ", collisions));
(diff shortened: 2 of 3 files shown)


caalador (approved):

flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
```diff
@@ -346,8 +346,12 @@ public void init() throws ServiceException {
             UsageStatistics.markAsUsed("flow/bun", null);
         }
 
-        RouteUtil.checkForClientRouteCollisions(this,
-                getRouteRegistry().getRegisteredRoutes());
+        if (getDeploymentConfiguration().isProductionMode()) {
+            // Postpone the check until dev-server is fully initialized and
+            // client routes are computed.
```
mshabarov, line 351: This comment is no longer valid as this block runs only in production

mcollovati: The comment is indeed not explanatory. The meaning is that in development mode the collision check is not executed at startup but postponed until the dev server is started.
By removing the comment, it is not clear if the check will be done or not, and it could be tempting to remove the `if (production)` condition.

caalador: It says exactly what is done, but the surrounding code confuses the issue.

PR #21624, opened 2025-06-04
feat: add method for `download`

caalador:
Add methods for checking if
anchor is a download and
setting the download status.

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -220,11 +220,40 @@ public class Anchor extends HtmlContainer
         assignHrefAttribute();
         if (downloadHandler instanceof AbstractDownloadHandler<?> abstractDownloadHandler) {
             if (!abstractDownloadHandler.isInline()) {
-                getElement().setAttribute("download", true);
+                setDownload(true);
             } else {
-                getElement().removeAttribute("download");
+                setDownload(false);
             }
         }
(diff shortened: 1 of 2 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -220,10 +220,39 @@ public void setHref(DownloadHandler downloadHandler) {
         assignHrefAttribute();
         if (downloadHandler instanceof AbstractDownloadHandler<?> abstractDownloadHandler
                 && !abstractDownloadHandler.isInline()) {
+            setDownload(true);
```
mshabarov: How about adding the flag/mode into a new CTOR and `setHref` overloads so that you can customise with `setHref(downloadHandler.inline(), true)` or `setHref(downloadHandler.inline(), DownloadMode.INLINE)`. And simultaneously make DownloadMode.ATTACHMENT be a default for `Anchor`?

caalador: That sounds like it makes the API more confusing. also we have #21623 which handles inline clearing for download also.

mshabarov: Yes, this imaginary overload indeed looks confusing.
Don't know what is the best default - download true for all or, as now, only for helpers.

mshabarov (dismissed):

mshabarov (approved):

PR #21633, opened 2025-06-04
chore: skip flow-tests when publishing (main)

ZheSun88:
this refactor is needed as central-publishing-maven-plugin does not recognize the `maven deploy` property when doing maven central publish


pom.xml
```diff
@@ -684,6 +683,18 @@
         </pluginManagement>
     </build>
     <profiles>
+        <profile>
+            <id>defalut</id>
```
knoobie: typo?

ZheSun88: good catch. fixed. thanks.

PR #21636, opened 2025-06-04
fix(hilla-i18n): remove custom fallback logic from handler

cromoteca:
Remove custom fallback logic in the translation file request handler to use default Java logic and align Hilla to Flow.

In addition, this PR adds some methods to `I18NProvider` to get bulk translations, with default implementations to not break existing code.

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/HtmlObject.java
@@ -328,9 +328,7 @@ public class HtmlObject extends HtmlContainer implements
             // where it is 'attachment' by default
             handler.inline();
         }
-        getElement().setAttribute("data",
-                new StreamResourceRegistry.ElementStreamResource(data,
-                        this.getElement()));
+        getElement().setAttribute("data", data);
     }
 
     /**
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/IFrame.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/IFrame.java
@@ -201,9 +201,7 @@ public class IFrame extends HtmlComponent implements HasAriaLabel {
             // where it is 'attachment' by default
             handler.inline();
         }
-        getElement().setAttribute("src",
-                new StreamResourceRegistry.ElementStreamResource(
-                        downloadHandler, this.getElement()));
+        getElement().setAttribute("src", downloadHandler);
     }
 
     /**
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Image.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Image.java
@@ -165,9 +165,7 @@ public class Image extends HtmlContainer
             // where it is 'attachment' by default
             handler.inline();
         }
-        getElement().setAttribute("src",
-                new StreamResourceRegistry.ElementStreamResource(
-                        downloadHandler, this.getElement()));
+        getElement().setAttribute("src", downloadHandler);
     }
 
     /**
--- a/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
+++ b/flow-plugins/flow-dev-bundle-plugin/src/main/java/com/vaadin/flow/plugin/maven/BuildDevBundleMojo.java
@@ -495,6 +495,11 @@ public class BuildDevBundleMojo extends AbstractMojo
 
     @Override
     public File webpackOutputDirectory() {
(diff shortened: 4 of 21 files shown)


flow-server/src/main/java/com/vaadin/flow/i18n/I18NProvider.java
```diff
@@ -70,6 +73,36 @@ default String getTranslation(Object key, Locale locale, Object... params) {
...
+     *            the keys to be translated
+     * @param locale
+     *            locale to use
+     * @return a map of translations
+     */
+    default Map<String, String> getTranslations(Collection<String> keys,
+            Locale locale) {
+        return keys.stream().distinct().collect(Collectors
+                .toMap(Function.identity(), k -> getTranslation(k, locale)));
+    }
```
taefi, line 104: Is something dictating the `keys` to be a `Collection`? Its combination with the `distinct()` suggests that it is a `Set`.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -113,20 +111,20 @@ protected boolean canHandleRequest(VaadinRequest request) {
...
     }
 
     private void handleNotFound(VaadinResponse response) {
         response.setStatus(HttpStatusCode.NOT_FOUND.getCode());
     }
 
-    private void handleCustomI18NProvider(VaadinSession session,
+    private void handleMissingI18NProvider(VaadinSession session,
             VaadinResponse response) throws IOException {
         String errorMessage = "Loading translations is not supported when using a custom i18n provider.";
```
platosha: Note: the message should be changed.

```suggestion
        String errorMessage = "Missing I18nProvider implementation, loading translations is not supported.";
```

flow-server/src/main/java/com/vaadin/flow/i18n/I18NProvider.java
cromoteca: It's mostly about who discards duplicates, but I agree with you that imposing a `Set` makes it clearer.
My intention was to avoid asking users for additional conversions, as anyway the fact that it returns a `Map` implies that there won't be duplicate keys.

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
```diff
@@ -41,8 +41,6 @@ public class DefaultI18NProvider implements I18NProvider {
     // Get bundles named `translations` from `vaadin-i18n` folder.
     public static final String BUNDLE_PREFIX = BUNDLE_FOLDER + "."
             + BUNDLE_FILENAME;
-    // Resource path for the file that contains the list of keys for each chunk.
-    public static final String CHUNK_RESOURCE = BUNDLE_FOLDER + "/i18n.json";
```
mshabarov, line 45: This is probably better to keep and deprecate as this is a public field in a public class potentially used by developers or our own products.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
```diff
@@ -72,37 +72,35 @@ public class TranslationFileRequestHandler extends SynchronizedRequestHandler {
...
 
-    private final DefaultI18NProvider i18NProvider;
+    private final I18NProvider i18NProvider;
 
-    private final boolean hasFallbackBundle;
+    private final ClassLoader classLoader;
 
     private Map<String, String[]> chunkData;
 
-    public TranslationFileRequestHandler(I18NProvider i18NProvider) {
```
mshabarov, line 83: Could also be deprecated, I know it's marked as internal, but nevertheless :)

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
platosha: On the other hand, this was introduced in 24.8 alpha, never was in a stable release so far.

flow-server/src/main/java/com/vaadin/flow/i18n/TranslationFileRequestHandler.java
platosha: Also was a feature-flagged feature.

platosha (approved):

flow-server/src/main/java/com/vaadin/flow/i18n/DefaultI18NProvider.java
mshabarov: Ah, then no worries.

mshabarov (approved):

PR #21649, opened 2025-06-05
feat: Add link type control to Anchor

caalador:
Add LinkMode to Anchor for
selecting if the anchor should
be download or inline.

### DIFF
--- a/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
+++ b/flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
@@ -137,6 +137,34 @@ public class Anchor extends HtmlContainer
     }
 
     /**
+     * Creates an anchor component with the given text content and a callback
+     * that handles data download from the server to the client when clicking an
+     * anchor.
+     * <p>
+     * Sets the 'download' attribute for link when given a non-inline handler
+     * implementing AbstractDownloadHandler.
+     * <p>
+     * LinkMode determines if the attribute `download` should be set or not.
+     * {@link LinkMode#DOWNLOAD} will set the download attribute, where as
+     * {@link LinkMode#INLINE} will remove it.
+     *
+     * @see #setHref(DownloadHandler, LinkMode)
+     * @see #setText(String)
+     *
+     * @param downloadHandler
+     *            the callback that handles data download, not null
+     * @param text
+     *            the text content to set
+     * @param linkMode
+     *            set the correct attribute for anchor according to given mode
+     */
+    public Anchor(DownloadHandler downloadHandler, String text,
+            LinkMode linkMode) {
+        setHref(downloadHandler, linkMode);
+        setText(text);
+    }
+
+    /**
      * Creates an anchor component with the given href and components as
      * children of this component.
      *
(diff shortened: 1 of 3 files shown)


flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -136,6 +136,34 @@ public Anchor(DownloadHandler downloadHandler, String text) {
...
+     *
+     * @param downloadHandler
+     *            the callback that handles data download, not null
+     * @param text
+     *            the text content to set
+     * @param linkMode
+     *            set the correct attribute for anchor according to given mode
+     */
+    public Anchor(DownloadHandler downloadHandler, String text,
+            LinkMode linkMode) {
```
knoobie: Some general questions / suggestions:

- I would swap the constructor arguments so that LinkMode is before text 
- What happens with null as LinkMode? 
- Could the other constructor without a LinkMode call this one to reduce duplicated code? 
- Could the setHref method without LinkMode delegate to the one with a LinkMode? (duplicated code) 
- LinkMode doesn't ring a bell for me.. some ideas: Disposition(Type), AttachmentType, DownloadMode

caalador: AttachmentType could be. DownloadMode doesn't sit well with the anchor overall. Lets see the consensus on naming.

Else done. null handled as inline as that is the default and you need to specifically call `setHref(handler, null)` to get a null value there

flow-html-components/src/main/java/com/vaadin/flow/component/html/Anchor.java
```diff
@@ -132,7 +132,36 @@ public Anchor(AbstractStreamResource href, String text) {
...
+     * @param attachmentType
+     *            set the correct attribute for anchor according to given mode
+     * @param text
+     *            the text content to set
+     */
+    public Anchor(DownloadHandler downloadHandler,
+            AttachmentType attachmentType, String text) {
+        setHref(downloadHandler,
+                attachmentType == null ? getLinkMode(downloadHandler)
+                        : attachmentType);
```
mshabarov: I'd move this resolution of `attachmentType` to the constructor and make download be a default. Now, when this resolution is in the `setHref`, it's imposible to know whether it's an initial state or not. I'd prefer the logic that always sets the attribute initially for all handlers, and preserves the existing state when you re-set the handler later:

1. When you create an anchor with only lambda handler parameter, then it's download.
2. When you create an anchor with only abstract handler parameter, then it checks inline()
3. When you re-set the handler via `setHref` then it preserves the attribute, unless you pass the download handler with `inline()` that should remove the attribute.
4. In all cases when the AttachmentType is explicitly set, it just follows it.

mshabarov: So 
```
return isDownload() ? AttachmentType.DOWNLOAD : AttachmentType.INLINE;
```
should become
```
return AttachmentType.DOWNLOAD;
```
when the `Anchor(DownloadHandler downloadHandler, String text)` is resolved.

caalador: This is the constructor.
In no case can we know if this is an initial state or not an it shouldn't matter.

2 is what happens now
3 and 4 keeps the current state as long as you do not set a pre-defined handler

caalador: You mean 
```
        setDownload(attachmentType != null
                && attachmentType.equals(AttachmentType.DOWNLOAD));
```

Should become
```
        setDownload(attachmentType == null
                || attachmentType.equals(AttachmentType.DOWNLOAD));
```
not the resolution that keeps the current state

mshabarov: Yes, exactly. The resolution that keeps the current state is good. The only change needed is around your lines, yes:
```
setDownload(attachmentType == null
                || attachmentType.equals(AttachmentType.DOWNLOAD));
```

PR #21664, opened 2025-06-06
Get rid of transitive dependency with little value

### LINKED ISSUE #18719: Consider removing ``com.helger:ph-css``

Backstory: I stumble upon this while searching for ``javax`` usages (which is brought in - among other things - by ``jsr305`` inside ``ph-commons`` inside ``ph-css`` inside ``flow-server`` ...)

Currently ``com.helger:ph-css`` is required by ``com.vaadin:flow-server``:
grafik

This dependency (with transitive dependencies) adds around ~2MB to the final jar, but only seems to be required for a single place:
grafik

Also note the likely invalid imports inside 
https://github.com/vaadin/flow/blob/8dad559c9f9f0bdaad756018a891c9224dfec472/flow-server/src/main/java/com/vaadin/flow/server/Platform.java#L23  and
https://github.com/vaadin/flow/blob/8dad559c9f9f0bdaad756018a891c9224dfec472/flow-server/src/test/java/com/vaadin/flow/server/frontend/FrontendVersionTest.java#L6

Now about the single usage inside [StyleAttributeHandler](https://github.com/vaadin/flow/blob/8dad559c9f9f0bdaad756018a891c9224dfec472/flow-server/src/main/java/com/vaadin/flow/dom/impl/StyleAttributeHandler.java):
According to it's description this class emulates the style attribute.
However I checked my complete app and could not find a single piece of code that called ``setAttribute`` (breakpoint never triggered).

I think it would be much better if ``com.helger:ph-css`` would be optional or not present:
* StyleAttributeHandler should simply not allow setting the ``style`` attribute manually (without using the ``Styles`` class) if ``ph-css`` is not present
* if ``ph-css`` Dependency is present the same logic as currently can be applied
* This would also stop people from accidentally importing stuff from ``com.helger``-packages

### Versions:
    - Vaadin / Flow version: 24.3.4

mstahv:
## Description

Gets rid of largish expendency that brings very little value.

Fixes #18719

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/dom/impl/StyleAttributeHandler.java
+++ b/flow-server/src/main/java/com/vaadin/flow/dom/impl/StyleAttributeHandler.java
@@ -18,13 +18,6 @@ package com.vaadin.flow.dom.impl;
 import java.util.LinkedHashMap;
 import java.util.stream.Collectors;
 
-import com.helger.css.ECSSVersion;
-import com.helger.css.decl.CSSDeclaration;
-import com.helger.css.decl.CSSDeclarationList;
-import com.helger.css.reader.CSSReaderDeclarationList;
-import com.helger.css.reader.errorhandler.CollectingCSSParseErrorHandler;
-import com.helger.css.writer.CSSWriterSettings;
-
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.dom.Style;
 import com.vaadin.flow.dom.StyleUtil;
@@ -38,6 +31,7 @@ import com.vaadin.flow.dom.StyleUtil;
  * @since 1.0
  */
 public class StyleAttributeHandler extends CustomAttribute {
+
     private static final String ERROR_PARSING_STYLE = "Error parsing style '%s': %s";
 
     @Override
@@ -75,30 +69,26 @@ public class StyleAttributeHandler extends CustomAttribute {
      */
     public static LinkedHashMap<String, String> parseStyles(
             String styleString) {
-        CollectingCSSParseErrorHandler errorCollector = new CollectingCSSParseErrorHandler();
-        CSSDeclarationList parsed = CSSReaderDeclarationList.readFromString(
-                styleString, ECSSVersion.LATEST, errorCollector);
-        if (errorCollector.hasParseErrors()) {
-            throw new IllegalArgumentException(String
-                    .format(ERROR_PARSING_STYLE, styleString, errorCollector
-                            .getAllParseErrors().get(0).getErrorMessage()));
+        try {
+            LinkedHashMap<String, String> parsedStyles = new LinkedHashMap<>();
+            String[] rules = styleString.split(";");
+            for (String rule : rules) {
+                String name = rule.substring(0, rule.indexOf(":")).trim();
+                String value = rule.substring(rule.indexOf(":") + 1);
+                if (name.isEmpty() || value.isEmpty()) {
+                    throw new IllegalArgumentException(
+                            "Style rule must contain name and value");
+                }
+                // TODO figure out if any better sanitatization is
+                // needed/feasible
+                // Probably not: users are on the "pro API" already and modern
+                // browsers doing enough sanitatision later in the browser
+                parsedStyles.put(name, value);
+            }
+            return parsedStyles;
+        } catch (Exception ex) {
+            throw new IllegalArgumentException(ex);
         }
-        if (parsed == null) {
-            // Did not find any styles
-            throw new IllegalArgumentException(String.format(
-                    ERROR_PARSING_STYLE, styleString, "No styles found"));
-        }
-
-        LinkedHashMap<String, String> parsedStyles = new LinkedHashMap<>();
-        for (CSSDeclaration declaration : parsed.getAllDeclarations()) {
-            String key = declaration.getProperty();
-            String value = declaration.getExpression()
-                    .getAsCSSString(new CSSWriterSettings(ECSSVersion.LATEST)
-                            .setOptimizedOutput(true), 0);
-            parsedStyles.put(StyleUtil.styleAttributeToProperty(key), value);
-        }
-
-        return parsedStyles;
     }
 
     @Override
--- a/flow-server/pom.xml
+++ b/flow-server/pom.xml
@@ -126,13 +126,6 @@
             <version>${jsoup.version}</version>
         </dependency>
 
-        <!-- ph-css for parsing style attribute in Element API -->
-        <dependency>
-            <groupId>com.helger</groupId>
-            <artifactId>ph-css</artifactId>
-            <version>7.0.4</version>
-        </dependency>
-
         <!-- Byte code analysis -->
         <dependency>
             <groupId>org.ow2.asm</groupId>


flow-server/src/main/java/com/vaadin/flow/dom/impl/StyleAttributeHandler.java
```diff
@@ -75,30 +69,26 @@ public void setAttribute(Element element, String attributeValue) {
...
-        if (errorCollector.hasParseErrors()) {
-            throw new IllegalArgumentException(String
-                    .format(ERROR_PARSING_STYLE, styleString, errorCollector
-                            .getAllParseErrors().get(0).getErrorMessage()));
-        }
-        if (parsed == null) {
-            // Did not find any styles
-            throw new IllegalArgumentException(String.format(
-                    ERROR_PARSING_STYLE, styleString, "No styles found"));
+        try {
```
knoobie, line 77: I'm highly in favor of this change.. only thing I'm a bit scared is how well this feature is tested from a unit test point of view.. 

First thing that comes to mind are special cases like this which are probably currently not tested in flow but heavily used by application developer to change Vaadin's variables on the fly `.getElement().getStyle().set("--lumo-primary-text-color", "hsl(12, 12%, 12%)")`

mstahv: I didn't look much into the tests, but was happy the they didn't accept my first draft 😎 Quickly checking there was now one commented out line that works with the new implementation and I guess one test method was not really executed (missing @Test annotation). In the same change, added trivial checks for basic css variables.

mstahv: But I wonder why ITs don't now compile 🤔

mstahv: And I get some TS compilation error when trying to to build those locally 😬

Legioth:
This simplistic "parser" breaks for anything that has a `;` or `:` within quotes. The first case that comes to mind might be something like `content: "TL;DR";` but that's mostly applicable for `::before` and `::after` pseudo elements so not directly applicable when applied directly to an element. Might also be some edge cases with e.g. font names or resource URLs. 

Probably not widely used by I'm sure there's _some_ application out there that will be affected. That's why I wonder if there's something we could do to help those application developers understand what goes on. Introducing this only in a major version and documenting it as a potentially breaking change is a good start but can we reasonably do more? I guess there's no logical place to update JavaDocs since this logic is triggered through the generic `setAttribute` method. I guess we cannot just detect such cases to log an warning since we could just as well handle it correctly if could detect the cases. Is there anything else that could be done or would we just have to accept the risk?

knoobie:
We could reduce the problems by trying to improve Matti's "parser" - I think it should be reasonable doable to improve the current form to get your examples to pass (good call with the font / url example.. I've seen some for dynamic images) while still staying away from the external dependency for such a minor task. 

I'm personally a big fan of "throw it in my face while developing, but keep quite once in production" - like have a big warning log over multiple lines once something unparseable is detected it in development mode and a single "Failed to parse XYZ. Enable development for more information at Component.class::setStyle in line 25 at your package com.bla.blub" 

Having a call out in the upgrade guide might also help, especially with the transitive dependencies this on added.. seen here in the test

mstahv:
This version of the "parser" shouldn't care about that having additional ":" characters, even in quotes, as only the first occurance is seeked (and at least I don't know a css variable with : in the name.

; could maybe be in a URL, but I wonder if anwhere else 🤔 (Checking for parenthesis should be rather easy).

I would be suprised if it would be hard to make this good enough even for minors. And if we don't make it a feature (ditching the dependency), reverting to ph-csss is very easy if the fix looks scary.

flow-server/src/main/java/com/vaadin/flow/dom/impl/StyleAttributeHandler.java
```diff
@@ -65,6 +59,10 @@ public void setAttribute(Element element, String attributeValue) {
         parseStyles(attributeValue).forEach(style::set);
     }
 
+    private static final char SEMICOLON = ";".charAt(0);
+    private static final char PARENTHESIS_OPEN = "(".charAt(0);
+    private static final char PARENTHESIS_CLOSED = "(".charAt(0);
```
caalador:
```suggestion
    private static final char PARENTHESIS_CLOSED = ')';
```
This doesn't seem right at all and any test testing an open close should probably fail. 
`parseStyles` method requires tests for its parsing functionality.

caalador:
```suggestion
    private static final char SEMICOLON = ';';
    private static final char PARENTHESIS_OPEN = '(';
    private static final char PARENTHESIS_CLOSED = ')';
```
We don't need to generate a string here.

flow-server/src/main/java/com/vaadin/flow/dom/impl/StyleAttributeHandler.java
```diff
@@ -75,30 +73,62 @@ public void setAttribute(Element element, String attributeValue) {
...
+                    if (valueTerminated) {
+                        addRule(nameBuffer, valueBuffer, parsedStyles);
+                        nameBuffer = new StringBuilder();
+                        valueBuffer = new StringBuilder();
+                        nameRead = false;
+                    } else {
+                        valueBuffer.append(c);
+                    }
+                } else {
+                    if (c == ":".charAt(0)) {
```
caalador:
```suggestion
                    if (c == ':') {
```
This could also have the `COLON` constant as the other parts.

caalador (changes requested):
Parser should get tests for the expected return values for different inputs.

flow-server/src/main/java/com/vaadin/flow/dom/impl/StyleAttributeHandler.java
mstahv: Please elaborate. You think the "parser" should detect non-closed functions? I changed this to this kind of implementation mostly to support semicolons in the URL values. As this is so little used "feature", I'm not convinved  it helps anybody, but that would be easy to add.

mstahv:
> Parser should get tests for the expected return values for different inputs.

@caalador I wonder if you noticed the tests in ElementJacksonTest ? There is bunch that (indirectly) test it.

flow-server/src/main/java/com/vaadin/flow/dom/impl/StyleAttributeHandler.java
caalador: Well a parenthesis is `(` to `)` but at the moment the check checks from `(` to `(` so for instance `name:(hello);fame:(oh no)` would not see any valueTerminated for hello leading to a value pair name -> (hello);fame:(oh no).
Thus the method itself would need some validating tests with input->output

caalador:
Non of the `ElementJacksonTest`s test with a value wrapped in `(xyz);` or `(z;x);`

flow-server/src/main/java/com/vaadin/flow/dom/impl/StyleAttributeHandler.java
mstahv: Eh 🤦‍♂️ Looks like complex tests don't include setting multiple css rules at all 🤣 

The tests even worked with the dyslexia issue (once completed a test to fail) as the rest of the rule was considered as value. Thus addedd couple of cases where it is tested via the Style object that rules are propertly split.

An easy enhancement/validation could be to throw an error if a css function is not closed, but I'm not sure if it pays of (the expectation is that the feature is not really used and will be used even less in the future).

flow-tests/pom.xml
```diff
@@ -72,6 +72,11 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
```
caalador, line 76: findbugs has been eol since 2016-2017:ish and the active one at the moment would be [com.github.spotbugs/spotbugs-annotations](https://github.com/spotbugs/spotbugs)

mstahv: That transitively pulls in the same dependency in anyways for the ones used in some of the test modules, so I'd keep the original until some of the actual spotbugs annotations are used:

```
[INFO] +- com.github.spotbugs:spotbugs-annotations:jar:4.9.3:compile
[INFO] |  \- com.google.code.findbugs:jsr305:jar:3.0.2:compile
```

PR #21665, opened 2025-06-06
refactor: Obtain SecurityContext from the SecurityContextHolderStrategy bean

### LINKED ISSUE #21401: Refactor Spring utilities to avoid static access to SecurityContextHolder
### Describe your motivation

A couple of Vaadin spring utility classes are using the static `SecurityContextHolder.getContext()` method to get the Spring security context.
That method relies on a `SecurityContextHolderStrategy` that is statically set based on system properties or calling static methods on the `SecurityContextHolder` class.

Vaadin provides a custom strategy (`VaadinAwareSecurityContextHolderStrategy`) built on top of  `VaadinSession` with a fallback to thread local.
However, since Vaadin utilities are using ``SecurityContextHolder.getContext()`, it is required to eagerly set the instance as global strategy using the workaround present in `VaadinAwareSecurityContextHolderStrategyConfiguration`

```java
   @Bean(name = "VaadinSecurityContextHolderStrategy")
    @ConditionalOnMissingBean
    public VaadinAwareSecurityContextHolderStrategy securityContextHolderStrategy() {
        VaadinAwareSecurityContextHolderStrategy vaadinAwareSecurityContextHolderStrategy = new VaadinAwareSecurityContextHolderStrategy();
        // Use a security context holder that can find the context from Vaadin
        // specific classes
        SecurityContextHolder.setContextHolderStrategy(
                vaadinAwareSecurityContextHolderStrategy);
        return vaadinAwareSecurityContextHolderStrategy;
    }
```

This is not optimal, especially with the new `VaadinSecurityConfigurer` because it requires the developer to import the `VaadinAwareSecurityContextHolderStrategyConfiguration` into the application security configuration class.(`VaadinWebSecurity` already does it).
In addition, Spring recommends wiring the strategy into other components instead of accessing the static method.

Make sure that Vaadin spring utilities use an injected `SecurityContextHolderStrategy` bean instead of invoking static access methods in `SecurityContextHolder`.

The current alternative is the workaround in `VaadinAwareSecurityContextHolderStrategyConfiguration`

### Additional context

https://github.com/spring-projects/spring-security/issues/10973
https://github.com/spring-projects/spring-security/issues/11886
https://github.com/spring-projects/spring-security/issues/12561

heruan:
This fixes #21401 by providing a `SecurityContextHolderStrategy` bean as part of Spring Security auto-configuration and replaces static invocations of `SecurityContextHolder.getContext()` by using the strategy bean instead.

- [x] Provide `SecurityContextHolderStrategy` in `SpringSecurityAutoConfiguration`
- [x] Remove conflicting `VaadinAwareSecurityContextHolderStrategyConfiguration`
- [x] Set the strategy on filters during `VaadinSecurityConfigurer` build lifecycle
- [x] Set the strategy statically when using `VaadinWebSecurity` for backwards compatibility
- [x] Inject the strategy bean in `AuthenticationContext` and `SpringAccessPathChecker`
- [x] Deprecate constructors that obtain the strategy statically
- [x] Avoid static access in `AuthenticationUtil` methods
- [ ] Update tests

### Breaking changes
- `VaadinAwareSecurityContextHolderStrategyConfiguration` has been removed — mild since it was purely for internal use
- `SpringSecurityAutoConfiguration::accessPatchChecker` signature has changed to include the strategy parameter — mild since this class shouldn't be extended (better have package-private bean methods)
- Applications that have set a custom strategy statically _after_ `VaadinAwareSecurityContextHolderStrategyConfiguration` might expect that custom strategy to be used by Flow, instead of the bean — those apps should now provide the custom strategy as a bean (if they expect Flow to use it)

**DRAFT** Tests setting the strategy statically must be updated (some already are)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/SpringAccessPathChecker.java
```diff
@@ -62,7 +64,10 @@ public class SpringAccessPathChecker implements AccessPathChecker {
      *
      * @param evaluator
      *            evaluator to check path permissions.
+     * @deprecated Use
+     *             {@link #SpringAccessPathChecker(WebInvocationPrivilegeEvaluator, String)}
```
mcollovati, line 115:
```suggestion
     *             {@link #SpringAccessPathChecker(SecurityContextHolderStrategy, WebInvocationPrivilegeEvaluator)}
```

mcollovati: It would be good to add a note mentioning usage of `SecurityContextHolder#getContextHolderStrategy()` in the deprecated constructors to explain the deprecation (similar to `AuthenticationContext`)

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinAwareSecurityContextHolderStrategyConfiguration.java
mcollovati, line 1: I wonder if we should just remove the `@Configuration` annotation and deprecate the class for 24.8

mcollovati (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinAwareSecurityContextHolderStrategyConfiguration.java
heruan: Yes, that would be better. I initially removed the class to see what failed without it, but we can keep it and deprecate before complete removal.

mcollovati:
One missing part: restore and deprecate `VaadinAwareSecurityContextHolderStrategyConfiguration`
This PR can be revised after `VaadinWebSecurity` removal gets merged.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringSecurityAutoConfiguration.java
```diff
@@ -206,7 +210,24 @@ public VaadinRolePrefixHolder vaadinRolePrefixHolder(
...
+
+    // Ensure AuthenticationUtil is wired to whichever
+    // SecurityContextHolderStrategy bean is in use,
+    // even if a custom one is provided by the application.
+    @Bean
+    SmartInitializingSingleton securityContextHolderStrategyInitializer(
+            SecurityContextHolderStrategy securityContextHolderStrategy) {
+        return () -> AuthenticationUtil.setSecurityContextSupplier(
+                securityContextHolderStrategy::getContext);
+    }
```
heruan, line 225: This defeats the purpose of having the strategy as a bean. This PR makes sense only by removing all static methods for security.

heruan: The strategy bean enables to have a different strategies for different scopes (in the same thread). Setting a static accessor opens to the risk of using the wrong strategy if the method is called on a scope where the expected strategy is different.

To keep backwards compatibility, apps that want to use static access could still use `SecurityContextHolder.getContext()` by explicitly setting the strategy statically.

mcollovati: I don't understand your point. `AuthenticationUtil` takes a `Supplier<SecurityContext>` so any time this is used it should call `securityContextHolderStrategy.getContext()`, where `securityContextHolderStrategy` is the defined bean.
What am I missing?

heruan: The point is that the bean instance could be different during runtime, while here we statically set that specific `securityContextHolderStrategy` bean instance. This is equivalent to
~~~java
SecurityContextHolder.setContextHolderStrategy(securityContextHolderStrategy)
~~~
and then obtain the context statically with `SecurityContextHolder.getContext()`.

heruan: Or is `SmartInitializingSingleton` evaluated every time based on the current context?

mcollovati: > SecurityContextHolder.setContextHolderStrategy(securityContextHolderStrategy)

And how is this different from ` new AuthenticationContext(securityContextHolderStrategy)` or `new SpringAccessPathChecker(securityContextHolderStrategy, evaluator, vaadinProperties.getUrlMapping())`?

heruan: Those other beans are created explicitly with that instance: if the developer wants to provide their own bean, they explicitly pass the strategy they need.

The problem with static methods is that you don't have much control of what they return and using them in different contexts/scopes might result in unexpected instances.

If we want to keep support for static access to the security context, then better stick with

~~~java
SecurityContextHolder.setContextHolderStrategy(securityContextHolderStrategy)
~~~

called somewhere, but I think that should be a choice on the app side.

mcollovati: I see, but I'm not sure if we can completely get rid of static methods in AuthenticationUtil.
What about a temporary workaround, like:

* add a constructor to `AuthenticationUtil` that gets `SecurityContextHolderStrategy` and sets it to the static field
* expose `AuthenticationUtil` as a bean with `@ConditionalOnMissingBean`
* if the developer needs, it can override the bean provide the required strategy

heruan: I wouldn't introduce a temporary workaround at this point. Let's split this change in multiple steps, maybe removing the separate configuration for the strategy bean and set it statically allowing the user to provide their own.

heruan:
Closing this in favor of an incremental approach and a proper deprecation cycle of the static accessor to the holder strategy, with #22745 as a starting point.

PR #21683, opened 2025-06-11
Support running the project from target/test-classes

peholmst:
It is possible in Spring Boot to create a `TestApplication` class, that extends the configuration of the main `Application` class with mocks or Testcontainers. This is very useful for local development as it does not require a separate development profile: 
* Run the main app with `mvn spring-boot:run` with real external services.
* Run the test app with `mvn spring-boot:test-run` with Testcontainer services (or mocks you configure).

For this to work, Vaadin must support running both from `target/classes` (main app) and `target/test-classes`(test app). This PR adds support for that.

PR https://github.com/vaadin/walking-skeleton/pull/71 adds a workaround to the walking skeleton, but I think it would be better to get this into Flow directly.

### DIFF
--- a/flow-client/pom.xml
+++ b/flow-client/pom.xml
@@ -126,7 +126,7 @@
         <dependency>
             <groupId>org.eclipse.platform</groupId>
             <artifactId>org.eclipse.equinox.preferences</artifactId>
-            <version>3.11.300</version>
+            <version>3.11.400</version>
             <scope>test</scope>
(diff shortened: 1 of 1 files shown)


peholmst:
Converted back to draft as this change breaks several tests.

mshabarov:
Testcontainers are not used for optimizing production build configs (it uses optional today), thus this PR is outdated and can be closed.

PR #21686, opened 2025-06-14
fix: Use handlers registered using ResourceHandlerRegistry

Artur-:
For instance Swagger uses ResourceHandlerRegistry to register a handler for /swagger-ui/index.html and this is now called without excluding any url from Vaadin

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
@@ -25,6 +25,8 @@ import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
@@ -36,6 +38,7 @@ import org.springframework.web.context.WebApplicationContext;
 import org.springframework.web.context.support.ServletContextResource;
 import org.springframework.web.servlet.DispatcherServlet;
 import org.springframework.web.servlet.HandlerExecutionChain;
+import org.springframework.web.servlet.HandlerMapping;
 import org.springframework.web.servlet.handler.SimpleUrlHandlerMapping;
 import org.springframework.web.servlet.mvc.Controller;
 import org.springframework.web.servlet.mvc.ServletForwardingController;
@@ -101,11 +104,11 @@ public class VaadinServletConfiguration {
         private List<String> excludeUrls;
         private AntPathMatcher matcher;
         private UrlPathHelper urlPathHelper = new UrlPathHelper();
-        private SimpleUrlHandlerMapping resourceHandlerMapping;
+        private HandlerMapping resourceHandlerMapping;
 
         public RootExcludeHandler(List<String> excludeUrls,
                 Controller vaadinForwardingController,
-                SimpleUrlHandlerMapping resourceHandlerMapping) {
+                HandlerMapping resourceHandlerMapping) {
             this.excludeUrls = excludeUrls;
             this.resourceHandlerMapping = resourceHandlerMapping;
             matcher = new AntPathMatcher();
@@ -176,23 +179,10 @@ public class VaadinServletConfiguration {
      */
     @Bean
     public RootExcludeHandler vaadinRootMapping(Environment environment,
-            WebApplicationContext webApplicationContext) {
-        SimpleUrlHandlerMapping resourceHandlerMapping = null;
-        try {
-            resourceHandlerMapping = webApplicationContext.getBean(
-                    "resourceHandlerMapping", SimpleUrlHandlerMapping.class);
-        } catch (Exception e) {
-            getLogger().debug(
-                    "Unable to find resourceHandlerMapping bean, "
-                            + "will not handle all static resources correctly",
-                    e);
-        }
+            WebApplicationContext webApplicationContext,
+            @Autowired(required = false) @Qualifier("resourceHandlerMapping") HandlerMapping resourceHandlerMapping) {
         return new RootExcludeHandler(getExcludedUrls(environment),
-                vaadinForwardingController(), resourceHandlerMapping);
-    }
-
-    private Logger getLogger() {
-        return LoggerFactory.getLogger(getClass());
+                vaadinForwardingController(), null);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
```diff
@@ -144,9 +176,24 @@ protected Logger getLogger() {
...
+        SimpleUrlHandlerMapping resourceHandlerMapping = null;
+        try {
+            resourceHandlerMapping = webApplicationContext.getBean(
+                    "resourceHandlerMapping", SimpleUrlHandlerMapping.class);
+        } catch (Exception e) {
+            getLogger().debug(
+                    "Unable to find resourceHandlerMapping bean, "
+                            + "will not handle all static resources correctly",
+                    e);
+        }
```
mcollovati: Couldn't we just get `HandlerMapping` injected as a method parameter? Something like
`@Autowired(required = false) @Qualifier("resourceHandlerMapping") HandlerMapping resourceHandlerMapping`
The logic in `RootExcludeHandler` seems to work also with that interface, without the need for a `SimpleUrlHandlerMapping`.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
```diff
@@ -127,9 +134,34 @@ protected Object getHandlerInternal(HttpServletRequest request)
                     }
                 }
             }
+            if (resourceHandlerMapping != null) {
+                // Check if the request is for a static resource
+                Object handler = resourceHandlerMapping.getHandler(request);
```
mcollovati: `getHandler` returns a `HandlerExecutionChain` so perhaps code can be shortened a bit

```java
HandlerExecutionChain chain = resourceHandlerMapping.getHandler(request);
if (chain != null && chain
        .getHandler() instanceof ResourceHttpRequestHandler resourceHttpRequestHandler) {
    if (!mapsToRoot(resourceHttpRequestHandler)) {
        // We cannot use the context resource mapped to / as
        // it would overwrite all
        // routes. Anything mapped to only a certain part of
        // the app is okay to use.
        return chain;
    }
}
```

mcollovati:
Should we remove `vaadin.exclude-urls=/swagger-ui/**` from all test modules configuration?

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
Artur-: My first attempt was with `SimplerUrlHandlerMapping resourceHandlerMapping` as a parameter but that only caused a circular reference exception as it apparently tried to inject `RootExcludeHandler` which is also a `SimpleUrlHandlerMapping` and ignores the name of a parameter. When adding `@Qualifier("resourceHandlerMapping")` it failed because of no bean with the qualifier being present - which sounds strange as the equivalent programmatic injection works

Artur-:
> Should we remove `vaadin.exclude-urls=/swagger-ui/**` from all test modules configuration?

Ah, sneaky. Some places use `exclude-urls` and some places `excludeUrls`. Removed the rest

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
Artur-: Removed the extra check

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
mcollovati: Weird thing. So the programmatic approach seems safer. Let's go this way

mcollovati: But I would try to use `HandlerMapping` instead of `SimplerUrlHandlerMapping`, if it works.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
```diff
@@ -127,9 +134,33 @@ protected Object getHandlerInternal(HttpServletRequest request)
                     }
                 }
             }
+            if (resourceHandlerMapping != null) {
+                // Check if the request is for a static resource
+                HandlerExecutionChain handler = resourceHandlerMapping
+                        .getHandler(request);
```
mcollovati, line 143: This call could potentially return null; maybe better add a check for that.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
Artur-: Retested: 
```
@Qualifier("resourceHandlerMapping") SimpleUrlHandlerMapping resourceHandlerMapping
```
->
```
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.web.servlet.handler.SimpleUrlHandlerMapping' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Qualifier("resourceHandlerMapping")}
```

but
```
@Qualifier("resourceHandlerMapping") HandlerMapping resourceHandlerMapping
```
->
```
Using resourceHandlerMapping: org.springframework.web.servlet.handler.SimpleUrlHandlerMapping@590013c7
```

Let's change to that then

Artur-: Updated

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
Artur-: Added check

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
mcollovati: So it looks like that with the qualifier you need to use the exactly same type as in the bean definition. Interesting!

mcollovati:
API-DIFF label is 1.0.0 because the PR removes the public updates the constructor adding the `HandlerMapping` parameter. 
Should this change be backported to 24.x? If so, we can consider re-adding and deprecating the original constructor in the back-port PRs.

Artur-:
I would put it in 24.x as it's almost a bugfix

mcollovati:
Ok. Let's merge it as is, then I'll make the back-ports with the deprecated constructor

Artur-:
Backport pr in #21712 but should it be for 24.8 or 24.9?

mcollovati:
> Backport pr in #21712 but should it be for 24.8 or 24.9?

I think 24.8 is fine as the change shouldn't break anything

PR #21687, opened 2025-06-16
chore: replace Spring deprecated API usage

### LINKED ISSUE #21562: Replace deprecated APIs in vaadin-spring
Since Spring Boot 3.5, some APIs used in `vaadin-spring` are deprecated and will be removed in the future.
One example is `AntPathRequestMatcher` (to be removed in Spring Security 7) used by `VaadinWebSecurity` and `VaadinSecurityConfigurer`, that should be replaced by the new `PathPatternRequestMatcher`.

All deprecated classes and methods used in `vaadin-spring` should be replaced in Flow 25.

mcollovati:
Closes #21562

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
@@ -1,16 +1,16 @@
 package com.vaadin.flow.spring.security;
 
-import jakarta.servlet.http.HttpServletRequest;
-
 import java.util.Optional;
 import java.util.stream.Stream;
 
+import jakarta.servlet.http.HttpServletRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.web.servlet.ServletRegistrationBean;
 import org.springframework.http.HttpMethod;
+import org.springframework.security.web.servlet.util.matcher.PathPatternRequestMatcher;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.stereotype.Component;
@@ -181,7 +181,10 @@ public class RequestUtil {
      *            and patterns
      * @return an array or {@link RequestMatcher} instances for the given
      *         patterns.
+     * @deprecated AntPathRequestMatcher is deprecated and will be removed, use
+     *             {@link #pathMatchers(String...)} instead.
      */
+    @Deprecated(since = "25.0", forRemoval = true)
     public static RequestMatcher[] antMatchers(String... patterns) {
         return Stream.of(patterns).map(AntPathRequestMatcher::new)
                 .toArray(RequestMatcher[]::new);
@@ -195,10 +198,46 @@ public class RequestUtil {
      *            and patterns
      * @return an array or {@link RequestMatcher} instances for the given
      *         patterns.
+     * @deprecated AntPathRequestMatcher is deprecated and will be removed, use
+     *             {@link #routePathMatchers(String...)} instead.
      */
+    @Deprecated(since = "25.0", forRemoval = true)
     public static RequestMatcher[] routeMatchers(String... patterns) {
         return Stream.of(patterns)
                 .map(p -> AntPathRequestMatcher.antMatcher(HttpMethod.GET, p))
+                .toArray(RequestMatcher[]::new);
+    }
+
+    /**
+     * Utility to create {@link RequestMatcher}s from path patterns.
+     *
+     * @param patterns
+     *            and patterns
+     * @return an array or {@link RequestMatcher} instances for the given
+     *         patterns.
+     * @see PathPatternRequestMatcher#matcher(HttpServletRequest)
+     */
+    public static RequestMatcher[] pathMatchers(String... patterns) {
+        var builder = PathPatternRequestMatcher.withDefaults();
+        return Stream.of(patterns).map(builder::matcher)
+                .toArray(RequestMatcher[]::new);
+    }
+
+    /**
+     * Utility to create {@link RequestMatcher}s for a Vaadin routes, using path
+     * patterns and HTTP get method.
+     *
+     * @param patterns
+     *            and patterns
+     * @return an array or {@link RequestMatcher} instances for the given
+     *         patterns.
+     * @see PathPatternRequestMatcher#matcher(HttpServletRequest)
+     */
+    public static RequestMatcher[] routePathMatchers(String... patterns) {
+        PathPatternRequestMatcher.Builder matcherBuilder = PathPatternRequestMatcher
+                .withDefaults();
+        return Stream.of(patterns)
+                .map(p -> matcherBuilder.matcher(HttpMethod.GET, p))
                 .toArray(RequestMatcher[]::new);
     }
 
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
@@ -15,13 +15,7 @@
  */
 package com.vaadin.flow.spring.security;
 
-import jakarta.annotation.PostConstruct;
-import jakarta.servlet.ServletException;
-import jakarta.servlet.http.HttpServletRequest;
-import jakarta.servlet.http.HttpServletResponse;
-
 import javax.crypto.SecretKey;
-
 import java.io.IOException;
 import java.util.LinkedHashMap;
 import java.util.Objects;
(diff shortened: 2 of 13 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -181,7 +181,10 @@ public boolean isCustomWebIcon(HttpServletRequest request) {
      *            and patterns
      * @return an array or {@link RequestMatcher} instances for the given
      *         patterns.
+     * @deprecated AntPathRequestMatcher is deprecated and will be removed, use
+     *             {@link #pathMatchers(String...)} instead.
      */
+    @Deprecated(since = "25.0", forRemoval = true)
```
knoobie: Question: Do you think this assumption holds true? Shouldn't this be since 24.8? and removed in 25.0? I got a feeling that the class might be unavailable in Spring (Security) 7 :( 

Suggestion: Are those replacement APIs needed? in the end.. their are just "convinent" methods for the developer.. which you guys have to maintain.. and hope Spring does not deprecate / remove them again :) time to ditch them?

mcollovati: > Question: Do you think this assumption holds true? Shouldn't this be since 24.8? and removed in 25.0? I got a feeling that the class might be unavailable in Spring (Security) 7 :(

You're right, it should be 24.8 even if this change will most likely be shipped in a patch release and not in 24.8.0

About the replacement method: the main reason for introducing them is that they are public API and we cannot drop them in a minor or patch.

mcollovati: I think I have misunderstood your comment. Do you  suggest to completely get rid of the helper methods?

knoobie: Yeah, I would completely remove those methods in 25 :) People can just use Spring's API directly / do not add new methods

mcollovati: IIRC they were added to get rid of some annoying issues with Spring automatically decide between ANT and MVC matcher when the pattern was given as a simple string.

I agree it would make sense to remove them. @mshabarov any concerns about removing the helpers? I suppose they were mainly used internally.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -195,13 +198,49 @@ public static RequestMatcher[] antMatchers(String... patterns) {
...
         return Stream.of(patterns)
                 .map(p -> AntPathRequestMatcher.antMatcher(HttpMethod.GET, p))
                 .toArray(RequestMatcher[]::new);
     }
 
+    /**
+     * Utility to create {@link RequestMatcher}s from path patterns.
+     *
+     * @param patterns
+     *            and patterns
```
mshabarov: This was copy-pasted, but "and patterns" looks incomplete, could we give more details?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -195,13 +198,49 @@ public static RequestMatcher[] antMatchers(String... patterns) {
...
+     */
+    public static RequestMatcher[] pathMatchers(String... patterns) {
+        var builder = PathPatternRequestMatcher.withDefaults();
+        return Stream.of(patterns).map(builder::matcher)
+                .toArray(RequestMatcher[]::new);
+    }
+
+    /**
+     * Utility to create {@link RequestMatcher}s for a Vaadin routes, using path
+     * patterns and HTTP get method.
```
mshabarov: I'd add a note describing the difference from `pathMatchers` method, because otherwise these methods look very similar.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -300,8 +298,9 @@ public static RequestMatcher getDefaultHttpSecurityPermitMatcher(
                 .map(path -> RequestUtil.applyUrlMapping(urlMapping, path))
                 .forEach(paths::add);
 
-        return new OrRequestMatcher(paths.build()
-                .map(AntPathRequestMatcher::new).collect(Collectors.toList()));
+        return new OrRequestMatcher(
+                paths.build().map(path -> RequestUtil.pathMatchers(path)[0])
```
mshabarov: Should this assert the array to have at least one element, that could simplify debugging in the future.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -195,13 +198,49 @@ public static RequestMatcher[] antMatchers(String... patterns) {
...
+     * Utility to create {@link RequestMatcher}s from path patterns.
+     *
+     * @param patterns
+     *            and patterns
+     * @return an array or {@link RequestMatcher} instances for the given
+     *         patterns.
+     * @see PathPatternRequestMatcher#matcher(HttpServletRequest)
+     */
+    public static RequestMatcher[] pathMatchers(String... patterns) {
+        var builder = PathPatternRequestMatcher.withDefaults();
```
mshabarov: This uses var and the other method uses explicit type, could they be unified?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -359,12 +362,43 @@ public RequestMatcher[] antMatchers(String... patterns) {
...
     public RequestMatcher[] routeMatchers(String... patterns) {
         return RequestUtil.routeMatchers(Stream.of(patterns)
                 .map(this::applyUrlMapping).toArray(String[]::new));
     }
 
+    /**
+     * Utility to create {@link RequestMatcher}s from path patterns.
+     *
+     * @param patterns
+     *            ant patterns
```
mshabarov: "ant" doesn't seem correct as ant matcher is deprecated. Could have a more detailed javadoc?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/VaadinStatelessSecurityConfigurer.java
```diff
@@ -130,29 +126,37 @@ public void setSharedObjects(HttpSecurity http) {
      * @param customizer
      *            the {@link Customizer} to provide more options for the
      *            {@link VaadinStatelessSecurityConfigurer}
+     * @deprecated use
+     *             {@code http.with(new VaadinStatelessSecurityConfigurer(), customizer)}
+     *             instead.
      */
+    @Deprecated(since = "24.8", forRemoval = true)
```
mshabarov, line 133: Could you please explain why is this deprecated and JWT code moved to init?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
mshabarov: No concerns about removing them from `RequestUtil`. I'd suggest to keep the similar helper methods in `VaadinWebSecurity` as these are documented and used in security config implementations. Could we also suggest a code example in the javadoc of the deprecated util methods of `RequestUtil` so that developers at least could copy-paste this examples or use them as a starting point?

mshabarov (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
mshabarov: Not valid anymore if we gonna skip adding the new methods.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/VaadinStatelessSecurityConfigurer.java
mcollovati: Because it was a wrong implementation. With the configurer you should be able to simply do `http.with(new VaadinStatelessSecurityConfigurer(), customizer)`. There is no good reason for this `apply` method.

mshabarov (dismissed):

mshabarov:
Though it needs a fix to address compilation errors, see e.g. Sonar build.

mcollovati:
> Though it needs a fix to address compilation errors, see e.g. Sonar build.

Yes, working on it

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
mcollovati: Update the method to get rid of RequestUtil and added the `assert`

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
mcollovati: it should have been "ANT patterns" :smile: 
Should be fixed now

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
mcollovati: method removed

PR #21712, opened 2025-06-17
fix: Use handlers registered using ResourceHandlerRegistry (#21686) (CP: 24.8)

Artur-:
For instance Swagger uses ResourceHandlerRegistry to register a handler for /swagger-ui/index.html and this is now called without excluding any url from Vaadin

### DIFF
--- a/flow-plugins/flow-plugin-base/pom.xml
+++ b/flow-plugins/flow-plugin-base/pom.xml
@@ -6,6 +6,7 @@
         <version>24.8-SNAPSHOT</version>
     </parent>
     <artifactId>flow-plugin-base</artifactId>
+    <name>Flow Plugin Base</name>
 
     <properties>
         <bnd.skip>false</bnd.skip>
--- a/flow-server/pom.xml
+++ b/flow-server/pom.xml
@@ -90,8 +90,8 @@
 
         <dependency>
             <groupId>org.apache.commons</groupId>
-            <artifactId>commons-fileupload2-jakarta</artifactId>
-            <version>2.0.0-M1</version>
+            <artifactId>commons-fileupload2-jakarta-servlet6</artifactId>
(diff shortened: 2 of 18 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
```diff
@@ -144,9 +186,11 @@ protected Logger getLogger() {
      *         servlet
      */
     @Bean
-    public RootExcludeHandler vaadinRootMapping(Environment environment) {
+    public RootExcludeHandler vaadinRootMapping(Environment environment,
+            WebApplicationContext webApplicationContext,
```
mcollovati: WebApplicationContext is not needed. I overlooked it in the main PR. I'll fix it soon
```suggestion
```

mcollovati:
We can probably make the diff label 0.0.1 by adding a deprecated `public RootExcludeHandler vaadinRootMapping(Environment environment)` method, but maybe it is not worth it.

Artur-:
Isn't the correct "fix" to remove all "public" modifiers from `@Bean` declarator methods?

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletConfiguration.java
Artur-: Included the fix here

mcollovati:
> Isn't the correct "fix" to remove all "public" modifiers from `@Bean` declarator methods?

Yes, that would be the best option.
But it could be considered a breaking change. Perhaps we can refactor all bean definitions in 25 an make all of them package private.

mcollovati:
We agreed on adding the deprecated method. I'll update the PR and then merge it.

PR #21725, opened 2025-06-18
fix: Update Query getPageSize to check for offset

### LINKED ISSUE #21481: ComboBox with AbstractBackEndDataProvider - Query pageSize always null
### Description

I have noticed that, when using a `ComboBox` with an `AbstractBackEndDataProvider`, the `Query` object in `fetchFromBackEnd` never has `pageSize` set, which causes it to always default to `limit`.

This was an issue in my current Project, where my `AbstractBackEndDataProvider` implementation queries a `JpaRepository` with a `Pageable` Object. In order to instantiate this object I need the pageNumber and pageSize, which I can't properly compute without explicitly having a pageSize.

The issue would namely occur when the DataProvider would fetch the last page. Say we have 27 entries and a pageSize set to 20:
- The first page would be fetched without a an issue: in Query, `limit` it set to 20 (and therefore `getPage()=20` also). 
- The problem occurs once we scroll down; a second query is executed, this time with `limit` is 7 (and therefore `getPageSize()` returns 7), and so if I try to instantiate the corresponding PageRequest, I am forced to use `pageNumber = offset / pageSize = 20 / 7 = 2` which is wrong, as we are querying for the second page, and not the third. 

This would cause some values in the dropdown to be displayed twice, and some to not be displayed. See image belowm, here we would expect to see "String 1" to "String 27". This corresponds to the snippet in the "Minimal reproducible example" section:

Image

The workaround I built in our code consisted in extending my implementation of the `AbstractBackEndDataProvider` in order to remember the `pageSize` set in the `ComboBox`. Maybe I am not aware of all details, but it seems to me it would be better to have `DataComunicator#doFetchFromDataProvider` setting the pageSize to the instantiated `Query` object.

Image

### Expected outcome

I would expect the input `Query` parameter of `AbstractBackEndDataProvider#fetchFromBackEnd` to have the pageSize set to the set `pageSize` value in the ComboBox, and not the `limit` of the current query.

```
class SomeDataProvider extends AbstractBackEndDataProvider<String, String> {

    private final List<String> listOfStrings = new ArrayList<>();

    public SomeDataProvider() {
        for (int i = 1; i <= 27; i++) {
            listOfStrings.add("String " + i);
        }
    }

    /**
     * This simulates what a JpaQuery would do.
     */
    private Stream<String> fetchPage(int pageNumber, int pageSize) {
        int calculateOffset = pageNumber * pageSize;
        return listOfStrings.subList(calculateOffset, Math.min(calculateOffset + pageSize, listOfStrings.size())).stream();
    }

    @Override
    protected Stream<String> fetchFromBackEnd(Query<String, String> query) {
        System.out.println("pageSize : " + query.getPageSize());
        System.out.println("limit : " + query.getLimit());
        int pageNumber = query.getOffset() / query.getPageSize();
        return fetchPage(pageNumber, query.getPageSize());
    }

    @Override
    protected int sizeInBackEnd(Query<String, String> query) {
        return listOfStrings.size();
    }
}

ComboBox<String> comboBox = new ComboBox<>();
comboBox.setPageSize(20);
comboBox.setDataProvider(new SomeDataProvider(), SerializableFunction.identity());
```

### Steps to reproduce

1. Add the ComboBox described above to a view
2. Open the ComboBox dropdown and scroll to the bottom
3. The last few elements cannot be seen. Additionally, in the logs see how the following values are logged:
pageSize : 20
limit : 20
pageSize : 7
limit : 7

### Environment

Vaadin version(s): 24.6.7 and 24.7.3
OS: Windows 10
Spring-Boot: 3.4.4

### Browsers

Firefox

caalador:
Make getPageSize check that it
matches the offset.
Update javadocs for Query to
explain that getPage updates
the pageSize to match offset.

fixes #21481

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
@@ -153,15 +153,12 @@ public class Query<T, F> implements Serializable {
     }
 
     /**
-     * Returns the page size that should be returned. The amount of items can be
-     * smaller if there is no more items available in the backend.
+     * Returns the current page size. The amount of items can be smaller if
+     * there is no more items available in the backend.
      * <p>
-     * Vaadin asks data from the backend in paged manner.
-     * <p>
-     * This is an alias for {@link #getLimit()} if the page size has not been
-     * set or page offset is evenly divisible with page size . Else the page
-     * size will be increased to evenly divide offset so the items skip for page
-     * will go to the correct item.
+     * The page size will be increased if it is not evenly divisible with the
+     * page offset. This will make it so that the page will go to the correct
+     * item.
      *
      * @return the page size used for data access
      */
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
@@ -121,6 +121,9 @@ public class Query<T, F> implements Serializable {
      * Vaadin asks data from the backend in paged manner. This shorthand
      * calculates the page index for backends using paged data access, such as
      * Spring Data repositories.
+     * <p>
+     * If page offset is not evenly divisible with pageSize raise pageSize until
+     * it is. Updates the pageSize value if it has been raised.
      *
      * @return the zero-based page index
      */
@@ -149,9 +152,12 @@ public class Query<T, F> implements Serializable {
      * <p>
      * Vaadin asks data from the backend in paged manner.
      * <p>
-     * This is an alias for {@link #getLimit()} if the page offset can be evenly
-     * divided by the limit. Else the page size will be increased to evenly
-     * divide offset so the items skip for page will go to the correct item.
+     * This is an alias for {@link #getLimit()} if the page size has not been
+     * set.
+     * <p>
+     * {@link #getPage()} will increase page size to evenly divide offset so the
+     * items skip for page will go to the correct item, in case the offset can
+     * not be evenly divided by the limit.
      *
      * @return the page size used for data access
      */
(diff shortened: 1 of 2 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
```diff
@@ -121,6 +121,9 @@ public int getLimit() {
      * Vaadin asks data from the backend in paged manner. This shorthand
      * calculates the page index for backends using paged data access, such as
      * Spring Data repositories.
+     * <p>
+     * If page offset is not evenly divisible with pageSize raise pageSize until
+     * it is. Updates the pageSize value if it has been raised.
```
mcollovati: nit: could use `page size` for consistency with `page offset` and the Javadoc of `getPageSize()`. 
```suggestion
     * If page offset is not evenly divisible with page size raise page size until
     * it is. Updates the page size value if it has been raised.
```

mcollovati: Another minor: the first line `Returns a zero-based page index to be retrieved` could maybe be changed into `Computes the zero-based page index to be retrieved` to make it clearer that this is not a fixed value.

caalador: updated

flow-data/src/main/java/com/vaadin/flow/data/provider/Query.java
```diff
@@ -149,17 +158,18 @@ private void setPageSize(Integer pageSize) {
      * <p>
      * Vaadin asks data from the backend in paged manner.
      * <p>
-     * This is an alias for {@link #getLimit()} if the page offset can be evenly
-     * divided by the limit. Else the page size will be increased to evenly
-     * divide offset so the items skip for page will go to the correct item.
+     * This is an alias for {@link #getLimit()} if the page size has not been
+     * set or page offset is not evenly divisible with page size . Else the page
```
mcollovati:
```suggestion
     * set and page offset is evenly divisible with page size . Else the page
```

PR #21729, opened 2025-06-19
Implicit support for proper namespaces for svg and mathml elements

### LINKED ISSUE #2842: When using the Element API I want to be able to create SVG elements

mstahv:
## Description

Html5 doesn't require namespace declarations for svg and math elements, but as elements are created with JS in Flow, they are incorrectly handled with Flow, in case one tries to create those elements with Element API. This change adds a tiny bit of intelligence to the front end so that svg and math elements implicitly get the right namespace and other elements default to inheriting it from the parent.

Fixes #2842 to some extent. Some trivial cases work, but for proper support we should also somehow get rid of lowercasing property names.

Creating a draft PR to see how IT tests pass on this early prototype. SvgElement and usage drafted in https://github.com/viritin/flow-viritin/commit/f360f556c61ef4c66eff8f73ff6ec2a459bc326e

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
+++ b/flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
@@ -172,6 +172,27 @@ public class SimpleElementBindingStrategy implements BindingStrategy<Element> {
 
         assert tag != null : "New child must have a tag";
 
+        // HTML5 supports svg & mathml without namespace delarations, with these
+        // checks Flow does as well
+        if (tag.equals("svg")) {
+            return Browser.getDocument()
+                    .createElementNS("http://www.w3.org/2000/svg", "svg");
+        } else if (tag.equals("math")) {
+            return Browser.getDocument().createElementNS(
+                    "https://www.w3.org/1998/Math/MathML/", "math");
+        }
+
+        // Else, get the namespace from parent element (svg & mathml elements
+        // inherits their custom ns)
+        if (node.getParent() != null) {
+            String namespaceURI = node.getParent().getDomNode()
+                    .getNamespaceURI();
+            if (namespaceURI != null) {
+                return Browser.getDocument().createElementNS(namespaceURI, tag);
+            }
+
+        }
+        // fallback (some root element!?), use default
         return Browser.getDocument().createElement(tag);
     }
 
--- a/flow-client/src/test/java/com/vaadin/client/DomApiAbstractionUsageTest.java
+++ b/flow-client/src/test/java/com/vaadin/client/DomApiAbstractionUsageTest.java
@@ -31,6 +31,7 @@ import elemental.dom.Element;
 import elemental.dom.Node;
 import elemental.dom.Text;
 import elemental.html.AnchorElement;
+import org.junit.Ignore;
 
 public class DomApiAbstractionUsageTest {
     private static final Set<String> ignoredClasses = Stream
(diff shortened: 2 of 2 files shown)


mstahv:
Added IT, ready from my side. I wonder what logic gave this +1.0.0 tag? I think it ought to be fine for next minor as well 🤷‍♂️

flow-client/src/main/java/com/vaadin/client/flow/binding/SimpleElementBindingStrategy.java
```diff
@@ -172,6 +172,27 @@ public Element create(StateNode node) {
 
         assert tag != null : "New child must have a tag";
 
+        // HTML5 supports svg & mathml without namespace delarations, with these
+        // checks Flow does as well
+        if (tag.equals("svg")) {
```
caalador, line 177: I get the feeling that instead of having a if/else for different tags we should add the `NodeProperties.NAMESPACE = "namespace";` to be used in `ElementData` and have here on the client only 
```
if (hasNamespace(node)) {
  return Browser.getDocument().createElementNS(getNamespace(node), tag);
}
```
This would make the client much more dynamic for having other namespace needs also without the need to change client code.

The namespace would then, now in the beginning for these 2 tags, be set in `BasicElementStateProvider.createStateNode(String tag)` (or `Element.createStateNode(String tag)`) to not complicate it too much and have time to make design for how to add a namespace for nodes.

The child elements could check for the parent namespace on attach on the server and add the data there to support different needs in these cases also.

mstahv: That would be then pretty much [this abandoned project](https://github.com/vaadin/flow/tree/element-namespace) + some more logic to the server side (toString, implicit logic, inheriting of ns). This approach would for each and every default ns node require transmitting the ns url.

I made this start instead, as I don't know of any other namespaces that are relevant or supported (and probably also mathml is irrelevant) and I don't really know why server would need to know anything about the ns (except for the "lowercasing-non-feature", which I think we should get rid of anyways and which can be overridden if needed). Same end result with less code (and strain for the server + client-server comms). I have zero doubts that these 2 checks in the client code makes it too complex (but I do have concerns it already is 😬).

The end goal would add a lot of flexibility for custom component develoment and it has now been lingering in wishlisht for pretty much 8 years. I don't care how it is achived, but I'm afraid if we still spend too much time figuring out the "perfectest" way to do it, it will never happen.

caalador: I don't see why that never went past a prototype in this case.
Just goin for the feature results as:
https://github.com/vaadin/flow/compare/feat/enable-element-namespace

caalador:
Namespaces added with #21805

mstahv:
Is this change planned to be released in 24.9? Didn't find this from [24.9.0-alpha3](https://github.com/vaadin/platform/releases/tag/24.9.0-alpha3).

mcollovati:
According to the PR label `1.0.0` it should not, but at a first look I don't see any blocker to pick it into 24.9.
I'll double-check with the team.

PR #21730, opened 2025-06-19
feat!: No default theme loaded

caalador:
Do not load Lumo as the
default theme if not defined.


flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
```diff
@@ -404,12 +399,19 @@ private ThemeData verifyTheme() {
...
-                                    .getName(),
-                            getAnnotationValueAsString(theme, "variant"),
-                            getAnnotationValueAsString(theme, VALUE)))
-                    .collect(Collectors.toSet());
+                    .map(theme -> {
+                        String themeClassName = "";
+                        Class<?> themeClass = (Class<?>) getAnnotationValue(
+                                theme, "themeClass");
+                        if (themeClass.equals(AbstractTheme.class)) {
+                            themeClassName = getLumoTheme().getName();
```
mcollovati: If I get it correctly `getLumoTheme()` can potentially return null. Should this case be handled by throwing an exception with a detailed message? Similar to what we do in `FrontendDependencies.java` ?
image

flow-server/src/test/java/com/vaadin/flow/server/frontend/scanner/ScannerThemeTest.java
```diff
@@ -92,30 +90,16 @@ public void should_throw_when_ThemeAndNoTheme() {
...
-            throws Exception {
-
-        DefaultClassFinder finder = spy(new DefaultClassFinder(
-                Collections.singleton(RootViewWithoutThemeAnnotation.class)));
-
-        // we'll do a partial mock here since we want to keep the other
-        // behavior of the DefaultClassFinder. Theme4 is used as a fake Lumo
-        // since it has @JsModule annotation which makes it easy to verify
-        // that the Theme was actually visited and modules collected
+    public void should_visitDefaultTheme_when_noThemeAnnotationIsGiven() {
```
mcollovati: It would be good to rename the test method since there is no default theme to visit anymore.

mcollovati (changes requested):
Tested locally and it behaves like applying the `@NoTheme` annotation.
Some minor changes requested

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FullDependenciesScanner.java
caalador: Changed to not use getLumoTheme().getName and instead use the LUMO constant.
Missing the theme should be checked already through the FrontendDependencies earlier.

flow-server/src/test/java/com/vaadin/flow/server/frontend/scanner/ScannerThemeTest.java
caalador: renamed tests.

mcollovati (approved):

mshabarov:
Let's wait until all components are ported to CSS files, there are [4 PRs](https://docs.google.com/spreadsheets/d/15_JoZ5v1pCtfzamlkD1e74KHjoTX5OZ0jwt5Hka3LSg/edit?gid=0#gid=0) currently on review.

vursen:
Depends on 

- https://github.com/vaadin/flow-components/pull/7668

web-padawan:
The PR for switching to Lumo CSS has been merged and the new Flow components snapshot is available.

PR #21738, opened 2025-06-23
Disallow update operation in readonly transactions

Legioth:
For some unknown reason, the `update` operation was wrapped in `runWithoutTransaction` which as a side effect meant that updates could be triggered from inside a readonly transaction. I don't see any reason for why `update` would be different from all the other operations, and there's also no previous test that would show why it's necessary.

### DIFF
--- a/signals/src/main/java/com/vaadin/signals/ValueSignal.java
+++ b/signals/src/main/java/com/vaadin/signals/ValueSignal.java
@@ -188,14 +188,13 @@ public class ValueSignal<T> extends Signal<T> {
          * Cannot easily optimize this to directly submit a transaction command
          * since we need the previous value from the set command result
          */
-        SignalOperation<T> setOperation = Transaction.runWithoutTransaction(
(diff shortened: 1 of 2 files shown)


signals/src/main/java/com/vaadin/signals/ValueSignal.java
```diff
@@ -188,14 +188,13 @@ private void tryUpdate(UnaryOperator<T> updater,
          * Cannot easily optimize this to directly submit a transaction command
          * since we need the previous value from the set command result
          */
```
mshabarov, line 190: This comment is probably not needed as now the transaction command is fired right away.

Legioth: The intent is to describe why we're using `Signal.runInTransaction` rather than `new TransactionCommand(id, commands)`.

mshabarov (approved):

Issue #21724, opened 2025-06-18
Vaadin 24.x Build fails: Rollup failed to resolve import "@vaadin/button/src/vaadin-button-base.js"


realulim:
### Description of the bug

This error has appeared for the first time yesterday on our CI server during build, while it did not occur on my development machine.

After cleaning caches and trying various Vaadin upgrades (currently on 24.6.5 and tried up to 24.7.0), the error now also appears on my development machine. So it has something to do with the environment, apparently cleaning up has removed some crucial file. Here is the build command and log output:

```
./gradlew -Dvaadin.force.production.build=true clean build

Execution failed for task ':my-app:vaadinBuildFrontend'.
> Vite process exited with non-zero exit code.
  Stderr: 'vite v6.0.11 building for production...
  transforming...
  ✓ 155 modules transformed.
  vite v6.0.11 building for production...
  transforming...
  ✓ 79 modules transformed.
  rendering chunks...
  computing gzip size...
  ../../../build/resources/main/META-INF/VAADIN/webapp/sw.js  19.33 kB │ gzip: 6.48 kB
  ✓ built in 665ms
  x Build failed in 1.64s
  error during build:
  [vite]: Rollup failed to resolve import "@vaadin/button/src/vaadin-button-base.js" from "/xxx/xxx/work/git/xxx/my-app/node_modules/@vaadin/app-layout/src/vaadin-drawer-toggle.js".
  This is most likely unintended because it can break your application at runtime.
  If you do want to externalize this module explicitly add it to
  `build.rollupOptions.external`
      at viteWarn (file:///xxx/xxx/work/git/xxx/my-app/node_modules/.pnpm/vite@6.0.11_terser@5.43.0/node_modules/vite/dist/node/chunks/dep-M1IYMR16.js:51220:17)
      at Object.onwarn (file:///xxx/xxx/work/git/xxx/my-app/node_modules/.vite-temp/vite.config.ts.timestamp-1750236091742-6cf765c6cef19.mjs:1209:11)
      at onwarn (file:///xxx/xxx/work/git/xxx/my-app/node_modules/.pnpm/@vitejs+plugin-react@4.3.4_vite@6.0.11_terser@5.43.0_/node_modules/@vitejs/plugin-react/dist/index.mjs:280:40)
[...]

PR #21747, opened 2025-06-23
Revert "refactor: Use overrides also for pnpm (#17667)"

Artur-:
There are two issues in the old PR:
1. pnpm does not actually use "overrides", it uses "overrides" inside "pnpm"
2. All vaadin dependencies are not listed in the overrides section. Pnpm can install newer versions of e.g. vaadin-button for dependencies that do not define a strict range. That will lead to conflicts and problems like #21724

This reverts commit 7dc87e4d0f0f125eae4a9773f86f7217c8caa996.


Artur-:
This does not solve the issue in #21724 because it only writes versions for the dependencies in package.json. It would need to write the versions for all dependencies included in the platform

mshabarov:
This is probably not needed after https://github.com/vaadin/flow/pull/21760.

PR #21786, opened 2025-06-26
fix: do not store build info in bundle

caalador:
Do not store the flow-build-info
inside the production bundle
zip file.

### DIFF
--- a/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
+++ b/flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
@@ -507,20 +507,25 @@ public class BuildFrontendUtil {
         FrontendTools tools = new FrontendTools(settings);
         tools.validateNodeAndNpmVersion();
         BuildFrontendUtil.runVite(adapter, tools);
+        String tokenContent = "";
+        File tokenFile = getTokenFile(adapter);
         try {
-            File tokenFile = getTokenFile(adapter);
-            Path tempToken = Paths.get(
-                    Files.createTempDirectory("build_temp").toString(),
-                    "flow-build-info.json");
-            Files.move(tokenFile.toPath(), tempToken);
+            tokenContent = Files.readString(tokenFile.toPath());
+            tokenFile.delete();
+        } catch (IOException ex) {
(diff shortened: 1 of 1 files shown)


flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
```diff
@@ -504,8 +507,21 @@ public static void runFrontendBuild(PluginAdapterBase adapter)
         FrontendTools tools = new FrontendTools(settings);
         tools.validateNodeAndNpmVersion();
         BuildFrontendUtil.runVite(adapter, tools);
-        ProdBundleUtils.compressBundle(adapter.projectBaseDirectory().toFile(),
-                adapter.servletResourceOutputDirectory());
+        try {
+            File tokenFile = getTokenFile(adapter);
+            Path tempToken = Paths.get(
+                    Files.createTempDirectory("build_temp").toString(),
+                    "flow-build-info.json");
```
mcollovati: nit: could maybe be simplified as following
```suggestion
            Path tempToken = Files.createTempDirectory("build_temp")
                    .resolve("flow-build-info.json");
```

flow-plugins/flow-plugin-base/src/main/java/com/vaadin/flow/plugin/base/BuildFrontendUtil.java
```diff
@@ -504,8 +507,21 @@ public static void runFrontendBuild(PluginAdapterBase adapter)
...
+            Files.move(tokenFile.toPath(), tempToken);
+
+            ProdBundleUtils.compressBundle(
+                    adapter.projectBaseDirectory().toFile(),
+                    adapter.servletResourceOutputDirectory());
+
+            Files.move(tempToken, tokenFile.toPath());
```
mcollovati: Should we move this into a `finally` block to "ensure" the token file is restored even if compression fails?
It would probably also need a check for `tempToken` file to exist, as also `createTempDirectory` and the first `Files.move()` may fail.

caalador: I did consider also just reading the file to a string, deleting the file and then rewriting the string contents. Might be better than moving the file around.

mcollovati (approved):

PR #21789, opened 2025-06-26
fix: remove default value for deprecated property

mcollovati:
Having the default value leads to a wrong warning message "Both 'frontendOutputDirectory' and 'webpackOutputDirectory' are set"


caalador (commented):
Should GradlePluginAdapter.kt also not set the webpackOutputDirectory?

mcollovati:
> Should GradlePluginAdapter.kt also not set the webpackOutputDirectory?

It should already handle that, by first looking for `frontendOutputDirectory` and then falling back the deprecated `webpackOutputDirectory`. However, there is no log message to indicate this problem.

mcollovati:
Or do you mean setting `frontendOutputDirectory` should nullify `webpackOutputDirectory`?
I'm not so familiar with Gradle properties, so I don't know how that could be achieved

PR #21802, opened 2025-06-27
feat!: Move component theme behind feature flag

### LINKED ISSUE #21608: Shadow DOM injection through components folder behind feature flag in V25
### Describe your motivation
The V24 theme folder's `components` subfolder is used to inject stylesheets into the shadow DOM of custom elements whose element names match that of the stylesheet.

In [V25, the theme folder in its entirety is deprecated](https://github.com/vaadin/platform/issues/7453) (but still supported), and shadow DOM styling is considered a legacy feature.

While the theme folder, and the `components`-folder-based injection mechanism will still be supported for backward compatibility, it should be disabled by default.

- A feature flag for enabling the `components` folder's shadow DOM injection.
- Perferrably, a warning shown if the `theme/sometheme/components/vaadin-something.css` structure is found in the project and the flag is not enabled.

caalador:
Add feature flag for component
theme files in `themes/themename/components`

Closes #21608

### DIFF
--- a/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
+++ b/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
@@ -193,11 +193,7 @@ function getThemeProperties(themeFolder, options) {
     };
   }
   let themeJson = JSON.parse(themePropertyFileAsString);
-  if ("autoInjectComponents" in themeJson) {
-    themeJson.autoInjectComponents = componentFeature && themeJson.autoInjectComponents;
-  } else {
-    themeJson.autoInjectComponents = componentFeature;
-  }
+  themeJson.autoInjectComponents = componentFeature;
   return themeJson;
 }
 
--- a/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
+++ b/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
@@ -193,9 +193,9 @@ function getThemeProperties(themeFolder, options) {
     };
   }
   let themeJson = JSON.parse(themePropertyFileAsString);
-  if("autoInjectComponents" in themeJson) {
+  if ("autoInjectComponents" in themeJson) {
     themeJson.autoInjectComponents = componentFeature && themeJson.autoInjectComponents;
-  }else {
+  } else {
     themeJson.autoInjectComponents = componentFeature;
   }
   return themeJson;
--- a/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
+++ b/flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
@@ -151,7 +151,7 @@ function handleThemes(themeName, themesFolder, options, logger) {
   if (existsSync(themeFolder)) {
     logger.debug('Found theme ', themeName, ' in folder ', themeFolder);
 
-    const themeProperties = getThemeProperties(themeFolder);
+    const themeProperties = getThemeProperties(themeFolder, options);
 
     // If theme has parent handle parent theme immediately.
     if (themeProperties.parent) {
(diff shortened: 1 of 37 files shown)


flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -174,16 +174,27 @@ function handleThemes(themeName, themesFolder, options, logger) {
...
   const themePropertyFileAsString = readFileSync(themePropertyFile);
   if (themePropertyFileAsString.length === 0) {
-    return {};
+    return {
+      "autoInjectComponents": componentFeature
+    };
   }
-  return JSON.parse(themePropertyFileAsString);
+  let themeJson = JSON.parse(themePropertyFileAsString);
+  themeJson.autoInjectComponents = componentFeature && themeJson.autoInjectComponents;
```
mshabarov: Why do we need to maintain both of these - `autoInjectComponents` parameter and `COMPONENT_STYLE_INJECTION` flag ? 

I'd suggest to just swap the default for `autoInjectComponents` to false.

mshabarov: Ah, that's probably not enough as `theme.json` is gonna be deprecated, thus the parameter too.

mshabarov: But this raised another question - why one have to turn ON both parameter and feature flag if wants components styles injection?
I'd remove the old parameter and leave only the feature flag then.

caalador: yes that's wrong. No value doesn't make it true.
The point is just that just by setting the featureFlag it shouldn't inject component styles if the theme.json flag is false.

mshabarov: I see no reason to maintain both of them. 
V25 will make the breaking change by swapping the injection from ON to OFF. Just use the feature flag to change the default.

caalador:
Added check and warning for components folder.

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
```diff
@@ -174,16 +174,31 @@ function handleThemes(themeName, themesFolder, options, logger) {
...
   if (themePropertyFileAsString.length === 0) {
-    return {};
+    return {
+      "autoInjectComponents": componentFeature
+    };
+  }
+  let themeJson = JSON.parse(themePropertyFileAsString);
+  if("autoInjectComponents" in themeJson) {
+    themeJson.autoInjectComponents = componentFeature && themeJson.autoInjectComponents;
+  }else {
```
mshabarov:
```suggestion
  if ("autoInjectComponents" in themeJson) {
    themeJson.autoInjectComponents = componentFeature && themeJson.autoInjectComponents;
  } else {
```

mshabarov (approved):

flow-server/src/main/resources/plugins/application-theme-plugin/theme-handle.js
mshabarov: The component styles inject mechanism works thanks to `autoInjectComponents` being passed through the theme plugin and injection codes, I guess it's hard to completely remove it without removing the whole injection code. Looks good to me now when we don't require `autoInjectComponents` to be set for changing the default.

caalador: We can just remove the themeJson autoInjectComponents from use as that is not required.
The other uses are just internal that are used from the theme.json if set.

mshabarov (approved):

vursen:
Depends on 

- https://github.com/vaadin/flow-components/pull/7668

PR #21805, opened 2025-06-30
feat: element namespace support

caalador:
Enable giving namespace for element.
for `svg` and `math` elements the default 
namespace is set automatically.
Namespace is inherited from the parent
if not specifically set for the child element.


flow-client/src/test/java/com/vaadin/client/DomApiAbstractionUsageTest.java
```diff
@@ -31,6 +31,7 @@
 import elemental.dom.Node;
 import elemental.dom.Text;
 import elemental.html.AnchorElement;
+import org.junit.Ignore;
```
mcollovati: nit: unused
```suggestion
```

PR #21813, opened 2025-07-01
fix: sw load should not be system dependent

### LINKED ISSUE #21743: sw.js not found, when server is running on Windows with hotdeployable frontend
### Description of the bug

When in dev mode with `vaadin.frontend.hotdeploy=true`, the dev mode server does not serve the service worker scripts `sw.js`, when the server is running on Windows. 

The reason for this is, that the id / sw source path check does not take into account, that path separators in windows are backslashes `\` instead of normal ones `/`. 

This makes the following check in the `vite-plugin-service-worker` always resulting in `false`.

```javascript
if (id === swSourcePath) {
  return buildOutput.output[0].code;
}
```

### Expected behavior

The check should handle path separators correctly, e.g. by replacing it with normal slashes:

```javascript
if (id === swSourcePath.replace(/\\/g, "/")) {
  return buildOutput.output[0].code;
}
```

- Create a blank vaadin project with a simple view.
- Add a minimal `@PWA` annotation.
- Enable frontend hotdeployment
- Start the app and open it in the browser
- Check the dev tools error log. It should log an error about a bad response code for the service worker script. You also cannot open the sw.js file via the browser url 

### Versions

- Vaadin / Flow version: 24.8
- Java version: 17
- OS version: Windows

caalador:
Check the load request as
relative instead of ===
to not have a problem between
systems that use \ instead of /

Fixes #21743

### DIFF
--- a/flow-server/src/main/resources/plugins/vite-plugin-service-worker/index.ts
+++ b/flow-server/src/main/resources/plugins/vite-plugin-service-worker/index.ts
@@ -1,5 +1,5 @@
 /// <reference types="node" />
-import { resolve } from 'node:path';
+import { resolve, relative } from 'node:path';
 import type { RollupOutput } from 'rollup';
 import { build, InlineConfig, Plugin } from 'vite';
 import { getManifest, ManifestTransform } from 'workbox-build';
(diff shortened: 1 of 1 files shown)


mshabarov (approved):

mshabarov:
Vaadin Flow 24.7 has a bit different implementation of `load(id)` in the `vite.generated.ts` (before refactoring).

PR #21828, opened 2025-07-02
fix: Support callpod user agent format

TatuLund:


flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
```diff
@@ -356,6 +356,12 @@ private void parseAndroidVersion(String userAgent) {
             return;
         }
 
+        // DuckDuckGo bot;
+        if (userAgent.contains("duckduckgo")) {
+            log("OS version could not be parsed from DuckDuckGo userAgent: "+userAgent);
```
knoobie: Off topic: I don't really understand the comment why SLF4J isn't used here.. using system out is a really big problem with e.g. structured logging... 

Off topic 2: why is this even logged as "error" 

Off topic 3: why is this even parsed when 90% of application do not care about user agents anymore? Wouldn't it be better so only parse it if needed? Like using is accessing any of the methods of this class?

TatuLund: > I don't really understand the comment why SLF4J

I think it is due legacy reasons ... This code is compiled to JavaScript by GWT and executed in browser.

> Like using is accessing any of the methods of this class?

This code is executed client side, but yes alternative could be to refactor this major way so that user agent is parsed in server side when used.

knoobie: Interesting! Didn't know this :) I've only seen this fascinating usage on the server side https://github.com/vaadin/flow/blob/ede2a2bd47fb5929375ee1f4477e526604d32935/flow-server/src/main/java/com/vaadin/flow/server/WebBrowser.java#L70

Legioth: Good suggestion to stop parsing those outdated strings. Let's see if we could get https://github.com/vaadin/flow/issues/21830 done for Vaadin 25.

flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
```diff
@@ -366,6 +374,18 @@ private void parseAndroidVersion(String userAgent) {
             return;
         }
 
+        if (userAgent.contains("callpod keeper for android")) {
+            int startIndex = userAgent.indexOf("; android ");
+            String osVersionString = safeSubstring(userAgent,
+                    startIndex + "; android ".length(), userAgent.length());
```
mcollovati: Looking around, I found examples of this user agent with only the major number, and the proposed change does not support it. I'd suggest determining the end of the token before parsing the parts, so it could handle versions defined as N, N.N and N.N.N.

```suggestion
            String token = "; android ";
            int startIndex = userAgent.indexOf(token) + token.length();
            int endIndex = userAgent.indexOf(";", startIndex);
            String osVersionString = safeSubstring(userAgent,
                    startIndex, endIndex);
```

flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
```diff
@@ -356,6 +356,14 @@ private void parseAndroidVersion(String userAgent) {
             return;
         }
 
+        // DuckDuckGo bot;
+        if (userAgent.contains("duckduckgo/")) {
```
mcollovati: It looks like there are variants for DDG that can be parsed, e.g. `Mozilla/5.0 (Linux; Android 14) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/130.0.6723.106 Mobile DuckDuckGo/5 Safari/537.36`.

Maybe just exclude the ones containing `com.duckduckgo.mobile.android`.

In addition:
* I would not print any message if this UA could not be parsed
* It would be good to add a tests also for this case

Issue #17624, opened 2023-09-14
Restrict VaadinWebSecurity to urlMapping


andykno:
The default security configuration [permits any requests for authenticated users](https://github.com/vaadin/flow/blob/main/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java#L210). Is this really necessary? Note that it does not only impact the scope of the Vaadin UI, but also all other parts of the application (e.g. REST-APIs like Spring Actuator).

The expected behavior is that the application is secure by default (principle of least privilege). The actual behavior is that authenticated users can access anything that is not secured by explicitly defined matchers.

mshabarov:
Currently we cannot say, would this change be a breaking change for existing projects using `VaadinWebSecurity`.
Basically, we can what is done here in this patch https://github.com/vaadin/flow/pull/17631.
This test pull request fails and we need to investigate the reason.

Two concerns so far:
- would `ViewAccessChecker` work as expected if we do this change
- would index page be blocked by Spring Security or should projects set up some extra rules to make it work as expected.

So far, this looks like a possible breaking change.
Needs further investigation.

tltv:
Based on the failing tests, only thing broken after removing `urlRegistry.anyRequest().authenticated()` is the [contract written in the `VaadinWebSecurity` JavaDoc](https://github.com/vaadin/flow/blob/cb63006b56777d46c881d41bc416d8d7189cd47d/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java#L78C10-L78C10). Failing tests follows this contract and expects authenticated routes and resources to be accessible, but they aren't without that configuration telling so.

So this would be a breaking change. This could be introduced as a optional new feature in `VaadinWebSecurity` to exclude this rule easier when needed.

mshabarov:
Based on this analysis I propose:
1. Open a GitHub discussion about what would be the best default. Our opinions went in different directions during internal discussion, so we'd like to hear the community opinion.
2. Based on the result of 1. we might plan changing default in next major version - Vaadin 25.
3. We can introduce a new method or parameter in Vaadin 24.3, that can help change the current default behaviour.

mshabarov:
Please vote here https://github.com/orgs/vaadin/discussions/4528.

PR #21711, opened 2025-06-17
chore: replace commons-fileupload2-jakarta with commons-fileupload2-jakarta-servlet6

ZheSun88:
vulnerability report https://github.com/advisories/GHSA-vv7r-c36w-3prj
since version 2.0.0-M2, the commons-fileupload2-jakarta  has been separated to commons-fileupload2-jakarta-servlet5  and commons-fileupload2-jakarta-servlet6

PR #21831, opened 2025-07-02
feat!: Restrict access by default with Spring Security

mshabarov:
Denies access to urls that have no request matcher in Spring Security config.

Fixes https://github.com/vaadin/flow/issues/17624

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
@@ -1,10 +1,7 @@
 package com.vaadin.flow.spring.security;
 
-import java.util.Optional;
 import java.util.stream.Stream;
 
-import jakarta.annotation.security.PermitAll;
-import jakarta.annotation.security.RolesAllowed;
 import jakarta.servlet.http.HttpServletRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -151,15 +148,16 @@ public class RequestUtil {
     }
 
     /**
-     * Checks whether the request targets a Flow route.
+     * Checks whether the request targets a Flow route secured with navigation
+     * access control.
      *
      * @param request
      *            the servlet request
-     * @return {@code true} if the request is targeting a Flow route ,
-     *         {@code false} otherwise
+     * @return {@code true} if the request is targeting a Flow route secured
+     *         with navigation access control, {@code false} otherwise
      */
-    public boolean isFlowRoute(HttpServletRequest request) {
-        return isFlowRouteInternal(request);
+    public boolean isSecuredFlowRoute(HttpServletRequest request) {
+        return isSecuredFlowRouteInternal(request);
     }
 
     /**
@@ -245,13 +243,16 @@ public class RequestUtil {
                 .toArray(RequestMatcher[]::new);
     }
 
-    private boolean isFlowRouteInternal(HttpServletRequest request) {
+    private boolean isSecuredFlowRouteInternal(HttpServletRequest request) {
         NavigationAccessControl navigationAccessControl = accessControl
                 .getObject();
         if (!navigationAccessControl.isEnabled()) {
             return false;
         }
+        return isFlowRouteInternal(request);
+    }
 
+    private boolean isFlowRouteInternal(HttpServletRequest request) {
         String path = getRequestRoutePath(request);
         if (path == null)
             return false;
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
@@ -735,7 +735,8 @@ public final class VaadinSecurityConfigurer
     private void customizeAuthorizeHttpRequests(
             AuthorizeHttpRequestsConfigurer<HttpSecurity>.AuthorizationManagerRequestMatcherRegistry registry) {
         registry.requestMatchers(defaultPermitMatcher()).permitAll()
-                .requestMatchers(getRequestUtil()::isFlowRoute).authenticated();
+                .requestMatchers(getRequestUtil()::isSecuredFlowRoute)
+                .authenticated();
     }
 
     private ApplicationContext getApplicationContext() {
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
@@ -224,7 +224,7 @@ public abstract class VaadinWebSecurity {
             urlRegistry.requestMatchers(requestUtil::isCustomWebIcon)
                     .permitAll();
             // private routes require authentication
-            urlRegistry.requestMatchers(requestUtil::isFlowRoute)
+            urlRegistry.requestMatchers(requestUtil::isSecuredFlowRoute)
                     .authenticated();
 
             // all other requests are denied
--- a/vaadin-spring/src/test/java/com/vaadin/flow/spring/security/RequestUtilTest.java
+++ b/vaadin-spring/src/test/java/com/vaadin/flow/spring/security/RequestUtilTest.java
@@ -420,11 +420,11 @@ public class RequestUtilTest {
 
         MockHttpServletRequest request = createRequest(null);
         request.setServletPath("/");
-        Assert.assertTrue(requestUtil.isFlowRoute(request));
+        Assert.assertTrue(requestUtil.isSecuredFlowRoute(request));
 
         request = createRequest("other");
         request.setServletPath("/");
-        Assert.assertTrue(requestUtil.isFlowRoute(request));
+        Assert.assertTrue(requestUtil.isSecuredFlowRoute(request));
     }
 
     @Test
@@ -437,11 +437,11 @@ public class RequestUtilTest {
 
         MockHttpServletRequest request = createRequest(null);
         request.setServletPath("/bar");
-        Assert.assertFalse(requestUtil.isFlowRoute(request));
+        Assert.assertFalse(requestUtil.isSecuredFlowRoute(request));
 
         request = createRequest("other");
         request.setServletPath("/bar");
-        Assert.assertFalse(requestUtil.isFlowRoute(request));
+        Assert.assertFalse(requestUtil.isSecuredFlowRoute(request));
     }
 
     @Test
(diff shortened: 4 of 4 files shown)


flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/resources/static/restricted/secret.txt
```diff
@@ -0,0 +1 @@
+Top secret restricted content
```
knoobie, line 1: nit: shouldn't this be labelled feat! instead of chore! so that it's pick-uped by the release notes? It's really frustrating to look at git diff what has really happened between releases.. especially if something so important happens 😬

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/SecurityConfig.java
```diff
@@ -93,14 +93,23 @@ SecurityFilterChain vaadinSecurityFilterChain(HttpSecurity http)
...
                 .requestMatchers(
                         RequestUtil.antMatchers("/public/**", "/error"))
-                .permitAll());
+                .permitAll()
+                .requestMatchers(new AntPathRequestMatcher("/all-logged-in/**"))
+                .authenticated()
+                .requestMatchers(new AntPathRequestMatcher("/private"))
+                .authenticated()
+                .requestMatchers(new AntPathRequestMatcher("/admin"))
+                .hasAnyRole(ROLE_ADMIN));
```
mshabarov: I don't really understand why these additions are needed after removing `authenticated()` from Vaadin web security. This `SecurityConfig` is not based on Vaadin class so how changes there affect this configuration...Main question how it worked before

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/resources/static/restricted/secret.txt
mshabarov: Makes sense! I was just thinking that this change is not "valuable" enough to be called feature, but this is subjective.

knoobie: You could also call it fix! or include chore! in the release notes templates :) in the end I'm open for all options.. just no information is bad imho

Let's use this as example https://github.com/vaadin/flow/pull/21711 I had to bisect the last releases to check if this commit was released in there because it wasn't mentioned 😬

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/SecurityConfig.java
tltv: After removing `authenticated()` request matcher, flow's access control is not called unless navigation happens via Flow API (e.g. from menu). Flow should apply `authenticated()` request matcher for flow routes annotated with `PermitAll` or `RolesAllowed` automatically. This needs an update in VaadinSecurityConfigurer and VaadinWebSecurity.
`AnonymousAllowed` has a request matcher already. `DenyAll` should work already since access is denied by default.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -734,7 +734,9 @@ private AccessDeniedHandler createAccessDeniedHandler() {
 
     private void customizeAuthorizeHttpRequests(
             AuthorizeHttpRequestsConfigurer<HttpSecurity>.AuthorizationManagerRequestMatcherRegistry registry) {
-        registry.requestMatchers(defaultPermitMatcher()).permitAll();
+        registry.requestMatchers(defaultPermitMatcher()).permitAll()
+                .requestMatchers(getRequestUtil()::isAuthenticatedRoute)
```
knoobie: I'm wondering if this should only be done if `enableNavigationAccessControl` returns true. I don't wanna use your access control annotations in my application - wouldn't this require me to add them? (see your test where you had to add @PermitAll

tltv: This wouldn't require you to add them. but if you do add them, then they are noticed automatically by this request matcher.
You can also disable all default configuration of authorized requests (including this whole customizeAuthorizeHttpRequests method) with `enableAuthorizedRequestsConfiguration` set to false.

tltv: Oh, and as `isAuthenticatedRoute` is not actually calling navigation access control, we don't need to check `enableNavigationAccessControl` here.

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/SecurityConfig.java
tltv: I've updated PR to do this now, but it might still need to also check parent layout annotations too.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -306,6 +345,24 @@ private boolean isAnonymousRouteInternal(HttpServletRequest request) {
...
+        if (maybePath.isEmpty()) {
+            return null;
+        }
+        String path = maybePath.get();
+        if (path.startsWith("/")) {
+            // Requested path includes a beginning "/" but route mapping is done
+            // without one
+            path = path.substring(1);
+        }
+        return path;
```
mcollovati: nit: can perhaps be simplified like this

```java
        return HandlerHelper
                .getPathIfInsideServlet(vaadinMapping, requestedPath)
                // Requested path includes a beginning "/" but route mapping is
                // done without one
                .map(path -> path.startsWith("/") ? path.substring(1) : path)
                .orElse(null);
```

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -232,22 +248,45 @@ public static RequestMatcher[] routeMatchers(String... patterns) {
...
+    private boolean isAuthenticatedRouteInternal(HttpServletRequest request) {
+        String path = getRequestRoutePath(request);
+        if (path == null)
+            return false;
+
+        SpringServlet servlet = springServletRegistration.getServlet();
+        VaadinService service = servlet.getService();
+        if (service == null) {
+            // The service has not yet been initialized. We cannot know if this
+            // is an anonymous route, so better say it is not.
```
mcollovati:
```suggestion
            // is an authenticated route, so better say it is not.
```

flow-tests/vaadin-spring-tests/test-spring-security-flow/src/main/java/com/vaadin/flow/spring/flowsecurity/SecurityConfig.java
tltv: Added check for parent layouts and auto-layout.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -232,22 +253,79 @@ public static RequestMatcher[] routeMatchers(String... patterns) {
...
+        if (routeTarget == null) {
+            return false;
+        }
+        Class<? extends com.vaadin.flow.component.Component> targetView = routeTarget
+                .getTarget();
+        if (targetView == null) {
+            return false;
+        }
+
+        if (isAuthenticationNeeded(targetView)) {
```
mcollovati: I think this method should also use NavigationAccessControl (if enabled) to determine if the route requires authentication. I'd say that at the very end it should return true if the NAC decision is `DENY` or `REJECT`.
This would take into account out-of-the-box `AnnotatedViewAccessChecker` (if used, since checkers can be disabled) and custom access checker as well.

tltv: We can probably simplify this by removing this annotation check and return true here if it's flow view. Then all flow views will go through access control in the end. Just need to test this once more.

tltv: And let's rename the method.

mcollovati: Makes sense, given that anonymous routes are already allowed before.
However, I think that the check should return true only if NAC is enabled, otherwise we cannot be sure if the view should be protected or not.

tltv: Updated.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -148,6 +150,18 @@ public boolean isAnonymousRoute(HttpServletRequest request) {
...
 
+    /**
+     * Checks whether the request targets a Flow route.
+     *
+     * @param request
+     *            the servlet request
+     * @return {@code true} if the request is targeting a Flow route ,
+     *         {@code false} otherwise
+     */
+    public boolean isFlowRoute(HttpServletRequest request) {
```
mcollovati: Since we have the `NAC.enabled` check, could this maybe be renamed to `isSecuredFlowRoute` or something similar?

mcollovati (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -426,8 +426,8 @@ public VaadinSecurityConfigurer enableAuthorizedRequestsConfiguration(
      * Configures the access rule for any request not matching other configured
      * rules.
      * <p>
-     * The default rule is to require authentication, which is the equivalent of
-     * passing {@link AuthorizedUrl#authenticated()} to this method.
+     * The default rule is to restrict access, which is the equivalent of
+     * passing {@link AuthorizedUrl#denyAll()} to this method.
```
vursen, line 430: In existing applications, some routes may _unintentionally_ be missing an explicit `authenticated()` rule. These now silently fail with a 403. I ran into this with images, for example. It could be helpful to log a debug message when an authenticated user is trying to access an _existing_ route that has become "denyAll".

Issue #21410, opened 2025-05-09
Deprecate VaadinWebSecurity in Vaadin 24 and remove in Vaadin 25


mcollovati:
#21373 introduced `VaadinSecurityConfigurer` for a better integration and modularization of Spring Security in Vaadin applications.

`VaadinSecurityConfigurer` will be the new recommended way to configure Sprin Security, so `VaadinWebSecurity` should be deprecated and then removed in a future major release.

Migration steps from `VaadinWebSecurity` to `VaadinSecurityConfigurer` should be documented (vaadin/docs#4298)

mshabarov:
We have to do https://github.com/vaadin/docs/issues/4298 first.

mshabarov:
Done.

PR #21838, opened 2025-07-03
feat: Deprecate VaadinWebSecurity

mshabarov:
Related-to https://github.com/vaadin/flow/issues/21410


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -99,7 +99,29 @@
...
  * </pre>
+ *
+ * @deprecated Use {@link VaadinSecurityConfigurer} instead. It follows the
+ *             Spring's SecurityConfigurer pattern and we recommend use it to
+ *             configure Spring Security with Vaadin:
+ *
+ *             <pre>
+ * <code>&#64;Configuration
+ * &#64;EnableWebSecurity
+ * &#64;Import(VaadinAwareSecurityContextHolderStrategyConfiguration.class)
```
knoobie, line 110: I would suggest to remove the 24.8 tag and try to finish https://github.com/vaadin/flow/pull/21665 for 24.9 allowing people at 24.9 to migrate to a clean solution without this workaround, more time for documentation updates and testing before people at 24.8 are "forced" in a 0.0.x release to get active because the old class is deprecated for removal

mshabarov: Deprecate only in 24.9 is probably quite a late notice as 25 beta and 24.9 are shipped approx. in the same time.

We could also just deprecate `VaadinWebSecurity` in V25, so more time to upgrade. It feels to me that the upgrade may take time especially if non-trivial config is used (LDAP, stateless, SSO).

So I'd say the choice is between deprecating in 24.8 and remove in 25 or deprecate in 24.9/25 and not remove until 26.

knoobie: I'm personally in favor of removal in 25 (but I'm also not using it..so take it with a grain of salt) - with no date for 26 on the horizon.. it feels like a big maintenance burden.. 

Speaking out load: I'm thinking about removal in 25.1 or 25.2 - even tho it's against semver to allow people to migrate to 25/Boot 4 etc without sooo much impact on their security config.. but this also comes with the problem that Spring Security 7 removes a lot of classes that are indirectly used by People in their old config.. so in the end they have to change it anyway.. 

TLDR: I'm not sure myself :)

mcollovati (dismissed):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
mshabarov: After having few weeks of vacay, with a fresh sight, it looks now to me that the better option is what you suggested initially - deprecate in 24.9 and remove in 25.0. Few months back it did make sense to include 24.8 here, but now we are approaching 24.9 beta already, and moreover we did a similar deprecation strategy for other places in Flow. So voting now for only 24.9 label.

mcollovati (dismissed):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -99,7 +99,29 @@
  * }
  * </code>
  * </pre>
+ *
+ * @deprecated Use {@link VaadinSecurityConfigurer} instead. It follows the
+ *             Spring's SecurityConfigurer pattern and we recommend use it to
+ *             configure Spring Security with Vaadin:
```
vursen, line 105: The deprecation notice should state the version in which this class will be removed. Otherwise, it's not immediately clear.

Issue #21811, opened 2025-06-30
Don't fail eagerly for conflicting task executors


Legioth:
If an application has multiple task executor beans and it's not clearly defined which one should be used as the `VaadinService` executor, then applciation startup currently fails. This is parcitularly problematic at this time when that Vaadin executor is used only for the signals preview feature that most users might not use and certainly not while upgrading from a previous version.

Instead, we should consider changing the initialization logic so that startup will always succeed but any problem will instead be reported only when trying to use any feature that needs an executor. In that way, users can upgrade right away and deal with any problems only as they start using the new features that become available from upgrading.

mshabarov:
As a workaround for Spring apps, to bypass the executor check that fails, one can add `VaadinTaskExecutor` annotation to either of these executors used in a project.

mcollovati:
We should also remove from the exception the suggestion to set `@Bean(defaultCandidate=false)`, since it seems not to work as expected.

PR #21839, opened 2025-07-04
fix: Fail on multiple executors only when the executor is accessed

mshabarov:
Does not fail eagerly if a Spring app has multiple `TaskExecutor` bean (without `VaadinTaskExecutor` qualifier).
Instead it fails when an app tries to access `VaadinService.getExecutor()` directly or via using a feature that uses this executor (currently the signals preview).

Fixes https://github.com/vaadin/flow/issues/21811

### DIFF
--- a/flow-tests/vaadin-spring-tests/pom.xml
+++ b/flow-tests/vaadin-spring-tests/pom.xml
@@ -171,7 +171,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-enforcer-plugin</artifactId>
-                    <version>3.5.0</version>
+                    <version>3.6.0</version>
                     <dependencies>
                         <dependency>
                             <groupId>de.skuzzle.enforcer</groupId>
--- a/flow-tests/vaadin-spring-tests/test-spring/pom.xml
+++ b/flow-tests/vaadin-spring-tests/test-spring/pom.xml
@@ -151,7 +151,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-enforcer-plugin</artifactId>
-                <version>3.5.0</version>
+                <version>3.6.0</version>
                 <executions>
                     <execution>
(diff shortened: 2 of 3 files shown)


Legioth:
From what I see, this approach bypasses the possibility of bypassing conflicting Spring beans through `serviceInitEvent.setExecutor` since the getter will run the logic from `SpringVaadinServletService` even if the value from that class wouldn't be used.

As an alternative approach, we could maybe create a placeholder executor instance where `execute` always throws the exception instead of making `getExecutor()` throw?

mshabarov:
> From what I see, this approach bypasses the possibility of bypassing conflicting Spring beans through serviceInitEvent.setExecutor since the getter will run the logic from SpringVaadinServletService even if the value from that class wouldn't be used.

Did not get this. 
If you set an executor via `serviceInitEvent.setExecutor`, then the `SpringVaadinServletService.createDefaultExecutor` should not run at all meaning that the multiple beans detection is skipped.

Then the throw block in `SpringVaadinServletService.getExecutor` will be also skipped as the `multipleExecutorsCandidates` is null.

> As an alternative approach, we could maybe create a placeholder executor instance where execute always throws the exception instead of making getExecutor() throw?

That can be an option, but in that case the stack trace may be less readable and also you spot this when a signal will try to execute something at runtime. I'd prefer an exception during initialisation.

Legioth:
> Did not get this.
If you set an executor via `serviceInitEvent.setExecutor`, then the `SpringVaadinServletService.createDefaultExecutor` should not run at all meaning that the multiple beans detection is skipped.

If that's the case, then I do indeed prefer this approach excactly because of the clearer stacktraces. I was for some reason assuming the logic in `SpringVaadinServletService` would always run.

mshabarov:
My statement was based on 
```java
this.executor = event.getExecutor()
    .orElseGet(this::createDefaultExecutor);
```
in `VaadinService.init` that is the only place where the default executor is accessed.

So Vaadin either uses the one from `serviceInitEvent` or makes this multi executor detection (for Spring).

PR #21852, opened 2025-07-08
test: use npm for pwa tests (2.12)

mcollovati:
Fixes Bender jobs failures caused by pnpm 7.33.5 dependencies resolution in `ŧest-pwa` module


mcollovati:
> Could also set the property `<vaadin.pnpm.enable>false</vaadin.pnpm.enable>` so that the configuration wouldn't need to be added as flow-tests pluginManagement set it as the property.

I think this does not work in 2.12, since the `flow-tests` pom file has a hard-coded value and does not use a property.

PR #21859, opened 2025-07-10
feat!: Deny access for parent layout with no annotation

### LINKED ISSUE #21832: Deny access if Flow layout class has no annotation
### Describe your motivation

Vaadin 24 has an inconsistency in how annotation based access works for layout classes when no annotation is defined:

Access denied:
```java
@Layout
public class MainLayout extends AppLayout { }

@Route
@PermitAll
public class MainView extends Div { }
```

Access granted for authenticated users:
```java
public class MainLayout extends AppLayout { } 

@Route(value = "", layout = MainLayout.class)
@PermitAll
public class MainView extends Div { }
```

For both configurations, if no access control annotation is given to `MainLayout` class, the access to routes within this layout should be denied.

### Additional context

This is a breaking change, so targets Vaadin 25.

caalador:
For a parent layout class with
no access control annotation,
the access to routes within this
layout are denied.

Fixes #21832

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/AnnotatedViewAccessChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/AnnotatedViewAccessChecker.java
@@ -16,6 +16,9 @@
 
 package com.vaadin.flow.server.auth;
 
+import java.util.Collections;
+import java.util.List;
+
 import jakarta.annotation.security.DenyAll;
 import jakarta.annotation.security.PermitAll;
 import jakarta.annotation.security.RolesAllowed;
@@ -26,6 +29,8 @@ import org.slf4j.LoggerFactory;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.router.BeforeEnterListener;
 import com.vaadin.flow.router.Layout;
+import com.vaadin.flow.router.RouteBaseData;
+import com.vaadin.flow.router.RouterLayout;
 import com.vaadin.flow.router.internal.RouteUtil;
 import com.vaadin.flow.server.RouteRegistry;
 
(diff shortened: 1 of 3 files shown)


flow-tests/vaadin-spring-tests/test-spring-security-flow-methodsecurity/src/main/java/com/vaadin/flow/spring/flowsecurity/views/MainView.java
```diff
@@ -20,9 +22,12 @@
 import com.vaadin.flow.router.PageTitle;
 import com.vaadin.flow.router.RouterLink;
 import com.vaadin.flow.server.auth.AccessAnnotationChecker;
+import com.vaadin.flow.server.auth.AnonymousAllowed;
 import com.vaadin.flow.spring.flowsecurity.SecurityUtils;
 import com.vaadin.flow.spring.flowsecurity.data.UserInfo;
 
+@PermitAll
+@AnonymousAllowed
```
mcollovati: Why are both annotations needed here? Shouldn't `@AnonymousAllowed` be enough to allow access to the layout?

caalador: Seems like it. There were just so many same setup tests that I confused the module that was failing =)

flow-server/src/main/java/com/vaadin/flow/server/auth/AnnotatedViewAccessChecker.java
```diff
@@ -89,6 +94,29 @@ public AccessCheckResult check(NavigationContext context) {
...
+                    .filter(routeData -> routeData.getNavigationTarget()
+                            .equals(targetView))
+                    .map(RouteBaseData::getParentLayouts).findFirst()
+                    .orElse(Collections.emptyList());
+            if (!parents.isEmpty()) {
+                for (Class<? extends RouterLayout> parent : parents) {
+                    boolean hasAccess = accessAnnotationChecker.hasAccess(
+                            parent, context.getPrincipal(), context::hasRole);
+                    if (!hasAccess) {
+                        LOGGER.debug(
```
vursen, line 110: Could it be a warning? Because otherwise, it's hard to spot this problem.

caalador: 👍 https://github.com/vaadin/flow/pull/22172

Issue #21890, opened 2025-07-15
Expose certain DataCommunicator APIs to allow overriding


vursen:
Make [DataCommunicator#requestFlush](https://github.com/vaadin/flow/blob/main/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java#L1127) protected to allow overriding it in the new HierarchicalDataCommunicator.

vursen:
Following the suggestion by @tltv, I implemented HierarchicalDataCommunicator's own flush mechanism instead of changing the visibility of DataCommunicator methods.

PR #21903, opened 2025-07-17
refactor: make requestFlush method protected

vursen:
## Description

The PR changes the visibility of `DataCommunicator#requestFlush` to protected to allow access and overriding in HierarchicalDataCommunicator.

Fixes https://github.com/vaadin/flow/issues/21890

## Type of change

- [x] Refactor

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
@@ -1120,11 +1120,11 @@ public class DataCommunicator<T> implements Serializable {
         }
     }
 
-    private void requestFlush() {
+    protected void requestFlush() {
         requestFlush(false);
     }
 
-    private void requestFlush(boolean forced) {
+    protected void requestFlush(boolean forced) {
         if (!shouldRequestFlush(forced)) {
             return;
         }


flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java
```diff
@@ -1120,11 +1120,11 @@ private void handleDetach() {
         }
     }
 
-    private void requestFlush() {
+    protected void requestFlush() {
```
tltv, line 1123: I'm not sure if this is good idea since `requestFlush` is called from constructor, overriding it in sub class could introduce issues due to the object may not be fully initialized yet. 

What if instead
1. change sub type to call super constructor always with  `fetchEnabled=false` parameter which makes `requestFlush()` to do nothing at all.
2. copy `requestFlush(boolean)` as a new private `internalRequestFlush(boolean)` method
3. change `requestFlush(boolean)` to protected and change it to just call `internalRequestFlush(boolean)`
4. repeat steps 2 and 3 for `requestFlush()` with similar naming pattern
5. change constructor to call `internalRequestFlush()`

`HierarchicalDataCommunicator` constructor seems to flush second time via `setDataProvider` call in the end of the constructor, which should make step 1 safe.

vursen: So, to confirm, you are suggesting that we disable DataCommunicator's requestFlush by setting `fetchEnabled` to `false`, and then implement a custom flush mechanism in HierarchicalDataCommunicator? Sounds like a good idea. 👍

tltv: Yes. However, `fetchEnabled` could be re-enabled after DataCommunicator instance is constructed. If that is needed.

vursen:
Following the suggestion by @tltv, I implemented HierarchicalDataCommunicator's own flush mechanism instead of changing the visibility of DataCommunicator methods. Closing this PR as unnecessary.

PR #21914, opened 2025-07-18
fix: validate all detected commercial components for Hilla apps

mcollovati:
During frontend build only npm packages imported by Java classes are validate by the license checker. However, for apps with Hilla views the check is skipped for packages imported only by frontend views. This change forces validation for all detected commercial components if the application is using Hilla views.


tltv (approved):

mshabarov:
No one can tell for sure if this is a good to go approach. Let's close the PR and keep the branch just in case.

PR #21928, opened 2025-07-23
chore: deprecate setConfiguration in VaadinSession

tepi:
Preparing for https://github.com/vaadin/flow/issues/18318


mcollovati:
It looks like a potential breaking change, since it is not only deprecating the setter, but also always ignoring the given value in `VaadinSession.getConfiguration()`.
Maybe all changes other that the deprecation should be reverted, if we want to backport this PR to 24.8.

tepi:
> It looks like a potential breaking change, since it is not only deprecating the setter, but also always ignoring the given value in `VaadinSession.getConfiguration()`. Maybe all changes other that the deprecation should be reverted, if we want to backport this PR to 24.8.

Agreed. Changed to deprecate only with no functional changes, moving those to other PR

mcollovati (approved):

PR #21938, opened 2025-07-23
chore!: remove setConfiguration from VaadinSession

tepi:
Fixes https://github.com/vaadin/flow/issues/18318

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/VaadinService.java
@@ -1203,6 +1203,13 @@ public abstract class VaadinService implements Serializable {
         // Initial WebBrowser data comes from the request
         session.setBrowser(new WebBrowser(request));
 
+        SessionLockCheckStrategy sessionLockCheckStrategy = getDeploymentConfiguration()
+                .isProductionMode()
+                        ? getDeploymentConfiguration()
+                                .getSessionLockCheckStrategy()
+                        : SessionLockCheckStrategy.THROW;
+        assert sessionLockCheckStrategy != null;
+
         // Initial locale comes from the request
         if (getInstantiator().getI18NProvider() != null) {
             setLocale(request, session);


flow-server/src/test/java/com/vaadin/flow/server/VaadinSessionTest.java
```diff
@@ -572,6 +574,14 @@ public static class TestVaadinSession extends VaadinSession {
 
         public TestVaadinSession() {
             super(new MockVaadinServletService());
+            try {
+                Field serviceField = getClass().getSuperclass()
+                        .getDeclaredField("service");
+                serviceField.setAccessible(true);
+                serviceField.set(this, null);
```
mcollovati: Out of curiosity: why do we need to nullify the service field?

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
```diff
@@ -350,38 +345,14 @@ private void refreshLock() {
...
-        if (configuration == null) {
-            throw new IllegalArgumentException("Can not set to null");
-        }
-        checkSetConfiguration();
-        this.configuration = configuration;
-
-        sessionLockCheckStrategy = configuration.isProductionMode()
-                ? configuration.getSessionLockCheckStrategy()
-                : SessionLockCheckStrategy.THROW;
-        assert sessionLockCheckStrategy != null;
```
mcollovati, line 369: Should we move this assertion elsewhere? For example in `VaadinService.createAndRegisterSession()` where we previously called `session.setConfiguration(getDeploymentConfiguration());`?

flow-server/src/test/java/com/vaadin/flow/server/VaadinSessionTest.java
tepi: This is for a valueUnbound test but it's handled elswehere already, probably a refactoring glitch, removed it

flow-server/src/main/java/com/vaadin/flow/server/VaadinSession.java
tepi: Good catch! Re-added the assertion.

PR #21939, opened 2025-07-23
fix(spring): CSRF login response for TypeScript forms

platosha:
Fixes vaadin/hilla#3697


davidbellem:
Just out of curiosity: If I see it correctly, this pull request does not add any additional tests to ensure that this behavior won't break again. Does it make sense to have an integration or other test that makes sure stateless security works with CSRF?

platosha:
> Just out of curiosity: If I see it correctly, this pull request does not add any additional tests to ensure that this behavior won't break again. Does it make sense to have an integration or other test that makes sure stateless security works with CSRF?

Correct, however, the setup is in the Hilla repository. The issue was not caught previously due to a difference in security config between docs and test setup, see https://github.com/vaadin/hilla/pull/3807 addressing this.

Issue #21873, opened 2025-07-15
Deprecate legacy HierarchicalDataCommunicator APIs


vursen:
Deprecate the following APIs in Vaadin 24.9 and remove in Vaadin 25:

- [ ] `DataProvider#refreshItem(T item, boolean refreshChildren)`
- [ ] `HierarchicalDataCommunicator#setParentRequestedRange`
- [ ] `HierarchicalDataCommunicator#confirmUpdate(int id, String parentKey)`
- [ ] `HierarchicalDataCommunicator#collapse(T item, boolean syncClient)`
- [ ] `HierarchicalDataCommunicator#expand(T item, boolean syncClient)`
- [ ] `HierarchicalDataCommunicator#createHierarchyMapper`
- [ ] `HierarchicalDataCommunicator#doUnregister`
- [ ] `HierarchicalDataCommunicator#getPassivatedKeys`
- [ ] `HierarchicalDataCommunicator#getHierarchyMapper`
- [ ] `HierarchicalDataCommunicator#getParentOfItem`
- [ ] `HierarchicalDataCommunicator#getIndex`
- [ ] `HierarchicalDataCommunicator#getParentIndex`
- [ ] `HierarchicalArrayUpdater`
- [ ] `HierarchicalCommunicationController`
- [ ] `HierarchyMapper`

PR #21953, opened 2025-07-25
refactor: undeprecate refreshItem, improve its JavaDoc

vursen:
## Description

It was decided to keep the `refreshItem(T item, boolean refreshChildren)` method available for non-flat data providers in the new version of  `HierarchicalDataCommunicator`. It doesn't fix scroll issues in every case, but it is known to help avoid them in some situations – for example when the refreshed item doesn't contain _expanded_ children. Since moving to flat data providers can be not straightforward, keeping this method around is the safer option, at least until Vaadin 26.

The PR also improves the JavaDoc of this method to warn about possible content shifts when it's called on items with more than one level of nesting.

Part of #21873 

## Type of change

- [x] Refactor

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/DataProvider.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/DataProvider.java
@@ -109,39 +109,20 @@ public interface DataProvider<T, F> extends Serializable {
     void refreshItem(T item);
 
     /**
-     * Refreshes the given item and all of the children of the item as well.
-     *
-     * @see #refreshItem(Object)
-     *
-     *      By default it just does a standard refreshItem, in a hierarchical
-     *      DataProvider it is supposed to refresh all of the children as well
-     *      in case 'refreshChildren' is true.
+     * Refreshes the given item and its children if refreshChildren is true and
+     * a hierarchical data provider is used. Otherwise, it behaves like regular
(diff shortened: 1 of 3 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/DataProvider.java
```diff
@@ -109,39 +109,20 @@ public interface DataProvider<T, F> extends Serializable {
...
-     *      DataProvider it is supposed to refresh all of the children as well
-     *      in case 'refreshChildren' is true.
+     * Refreshes the given item and its children if refreshChildren is true and
+     * a hierarchical data provider is used. Otherwise, it behaves like regular
+     * {@link #refreshItem(Object)}.
+     * <p>
+     * It's important to note that this method resets the item's hierarchy which
+     * can cause a content shift if the item contains expanded children: their
+     * descendants aren't guaranteed to be re-fetched eagerly, which may affect
+     * the overall size of the rendered hierarchy, leading to content shifts.
```
mcollovati, line 119: Should we anyway suggest `refreshAll()` if issues in the application are noticed? Or it does not make sense?

vursen: No, that doesn't make sense. Instead, we can suggest flat data providers. I'm going to add this when they are available.

PR #21960, opened 2025-07-28
feat: add commercial banner for embedded Flow components

mcollovati:


mcollovati:
> LGTM. One minor thing, there's still `watermarked` in a package name. It should probably say `bannered` now.

Good point. I'll update the package names.

PR #21966, opened 2025-07-29
refactor!: centralize hierarchy management in HierarchicalDataCommunicator

### LINKED ISSUE #21876: Replace existing HierarchicalDataCommunicator with new implementation
Replace the existing HierarchicalDataCommunicator with the new version developed in https://github.com/vaadin/flow-components/pull/7676

Depends on
- https://github.com/vaadin/flow-components/issues/7684

vursen:
## Description

WIP

Fixes #21876 

## Type of change

- [x] Refactor


flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/hierarchicaldatacommunicator/Cache.java
```diff
@@ -0,0 +1,160 @@
...
+ * <p>
+ * WARNING: This class is intended for internal use only and may change at any
+ * time without notice. It is not part of the public API and should not be used
+ * directly in your applications.
+ *
+ * @param <T>
+ *            the type of items in the cache
+ * @private
+ */
+public class Cache<T> implements Serializable {
```
knoobie: Is it required to have all Cache related internal classes public? or can those be package private? or are this mandatory to allow subclassing of the new Communicator?

vursen: Good question. They can be package-private, but in that case, we can't place them in a subpackage like `hierarchicaldatacommunicator`, so they would have to reside in the `hierarchy` namespace which already contains a variety of different classes while `Cache` is only relevant to `HierarchicalDataCommunicator`. What do you think?

knoobie: I think this should be decided by the flow team, it comes down to DX of Vaadin Users vs. DX of Flow Developer :) As Vaadin User I would prefer Cache and other class names not available in my recommended list of class names by IDEA

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorDataTest.java
```diff
@@ -1,383 +0,0 @@
...
-                Mockito.mock(CompositeDataGenerator.class), arrayUpdater,
-                json -> {
-                }, element.getNode(),
-                () -> (ValueProvider<Item, String>) item -> String
-                        .valueOf(item.id));
-        communicator.setDataProvider(dataProvider, null);
-    }
-
-    @Test
-    public void sameKeyDifferentInstance_latestInstanceUsed() {
```
vursen, line 162: **note:** Replaced by `HierarchicalDataCommunicatorDataRefreshTest#refreshItems_itemRefreshedInKeyMapper`

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorDataTest.java
```diff
@@ -1,383 +0,0 @@
...
-
-        dataProvider.refreshItem(FOLDER, true);
-
-        fakeClientCommunication();
-
-        Assert.assertSame(updatedLeaf, communicator.getKeyMapper().get(key));
-    }
-
-    @Test
-    public void uniqueKeyProviderIsSet_keysGeneratedByProvider() {
```
vursen, line 187: **note:** Replaced by `HierarchicalDataCommunicatorKeyGenerationTest#setUniqueKeyProvider_keysGeneratedByProvider`

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorDataTest.java
```diff
@@ -1,383 +0,0 @@
...
-        fakeClientCommunication();
-
-        Arrays.asList("ROOT", "FOLDER", "LEAF")
-                .forEach(key -> Assert.assertNotNull("Expected key '" + key
-                        + "' to be generated when unique key provider used",
-                        communicator.getKeyMapper().get(key)));
-    }
-
-    @Test
-    public void uniqueKeyProviderIsNotSet_keysGeneratedByKeyMapper() {
```
vursen, line 207: **note:** Replaced by 

1. `HierarchicalDataCommunicatorKeyGenerationTest#changeViewportRangeBackAndForth_generatedKeysMatchItems`
2. `HierarchicalDataCommunicatorKeyGenerationTest#changeViewportRangeBackAndForth_reset_generatedKeysMatchItems`

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorDataTest.java
```diff
@@ -1,383 +0,0 @@
...
-
-        communicator.confirmUpdate(committedUpdateId);
-        fakeClientCommunication();
-        assertKeyItemPairIsPresentInKeyMapper(initialKey, itemToTest);
-
-        communicator.reset();
-    }
-
-    @Test
-    public void expandItem_tooMuchItemsRequested_maxItemsAllowedRequested() {
```
vursen, line 312: **note:** Replaced by `HierarchicalDataCommunicatorViewportTest#setTooLargeViewportRange_onlyMaxAllowedRangeSent`

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorTest.java
```diff
@@ -1,338 +0,0 @@
...
-    public void folderRemove() {
-        testItemRemove(FOLDER, false);
-    }
-
-    @Test
-    public void leafRemove() {
-        testItemRemove(LEAF, false);
-    }
-
-    private void testItemRemove(String item, boolean refreshAll) {
```
vursen, line 179: **note:** Replaced by `HierarchicalDataCommunicatorDataRefreshTest`

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorTest.java
```diff
@@ -1,338 +0,0 @@
...
-                .runWhenAttached(attachCaptor.capture());
-
-        attachCaptor.getAllValues().forEach(consumer -> consumer.accept(ui));
-
-        Mockito.verify(stateTree, Mockito.times(number))
-                .beforeClientResponse(Mockito.any(), Mockito.any());
-    }
-
-    @Test
-    public void replaceAll() {
```
vursen, line 205: **note:** Replaced by `HierarchicalDataCommunicatorDataRefreshTest`

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalCommunicatorTest.java
```diff
@@ -1,338 +0,0 @@
...
-     * another root node, the key is held in KeyMapper after two flush events.
-     *
-     * Related: <a href="https://github.com/vaadin/flow/issues/14351">Original
-     * issue</a> Related:
-     * <a href="https://github.com/vaadin/flow/pull/17774">Fix</a> Related:
-     * <a href="https://github.com/vaadin/flow-components/pull/5545">Integration
-     * Test</a>
-     */
-    @Test
-    public void moveNodeFromRootToChildAndFlushTwice_keyShouldBeInKeyMapper() {
```
vursen, line 229: **note:** Replaced by `HierarchicalDataCommunicatorDataRefreshTest`

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -85,128 +103,123 @@ public class HierarchicalDataCommunicator<T> extends DataCommunicator<T> {
...
-                                size -> arrayUpdater
-                                        .startUpdate(getDataProviderSize()),
-                                (pkey, range) -> mapper.fetchChildItems(
-                                        getKeyMapper().get(pkey), range)));
-        controller.setHasUniqueKeyProviderSupplier(
-                uniqueKeyProviderSupplier.get() != null);
-        Range range = computeRequestedRange(start, length);
-        controller.setRequestRange(range.getStart(), range.length());
-        requestFlush(controller);
+    public void refresh(T item, boolean refreshChildren) {
```
vursen, line 195: **note:** This is a new public method that I created for consistency with the corresponding  `DataProvider#refreshItem` method

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -85,128 +103,123 @@ public class HierarchicalDataCommunicator<T> extends DataCommunicator<T> {
...
+        if (refreshChildren && subCache != null) {
+            subCache.clear();
+            subCache.setSize(getDataProviderChildCount(item));
+        }
+
+        requestFlush();
+    }
+
+    @Override
+    public Stream<T> fetchFromProvider(int offset, int limit) {
```
vursen, line 224: **idea:** This method and `getDataProviderSize` are inherited from the previous implementation. A more flexible solution might be to replace them with a `buildQuery()` method (already exists in `DataCommunicator`) that returns a `HierarchicalQuery` backed with all sort and filter parameters, which users can then pass directly to the data providers.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -400,107 +366,239 @@ public boolean hasChildren(T item) {
      * @return {@code true} if item is expanded; {@code false} if not
      */
     public boolean isExpanded(T item) {
-        return mapper.isExpanded(item);
+        if (item == null) {
+            // Root nodes are always visible.
+            return true;
+        }
```
vursen: This is inheritance from HierarchyMapper, but it's unclear whether it's still necessary.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -400,107 +366,239 @@ public boolean hasChildren(T item) {
...
      *
-     * @return {@code true} if there is any expanded items.
+     * @param path
+     *            the hierarchical path to the item, where each element
+     *            represents the index within its respective level
+     * @return the flat index of the target item after resolving all ancestors
      */
-    public boolean hasExpandedItems() {
-        return mapper.hasExpandedItems();
+    protected int resolveIndexPath(int... path) {
```
vursen, line 430: **note:** Used for TreeGrid's scrollToIndex

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -400,107 +366,239 @@ public boolean hasChildren(T item) {
...
-                /* Short-circuit root item passivation */
-                if (parent == null) {
-                    return !isExpanded(item);
-                }
-                while (parent != null) {
-                    if (!isItemActive(parent) || !isExpanded(parent)) {
-                        return true;
-                    }
-                    parent = mapper.getParentOfItem(parent);
+    protected List<T> preloadRange(int start, int length) {
```
vursen: **note:** Also used for scrollToIndex in TreeGrid, tests will come separately.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/Cache.java
```diff
@@ -0,0 +1,159 @@
...
+ * and references to their child caches if they are expanded.
+ * <p>
+ * WARNING: This class is intended for internal use only and may change at any
+ * time without notice. It is not part of the public API and should not be used
+ * directly in your applications.
+ *
+ * @param <T>
+ *            the type of items in the cache
+ */
+class Cache<T> implements Serializable {
```
tltv, line 44: Javadoc for methods and constructor are missing. E.g. it's not exactly clear what's the difference of `getSize()` and `getFlatSize()`. Having Javadoc would help to understand quicker.

vursen: Thanks – added for Cache, and will add for RootCache as well.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/RootCache.java
```diff
@@ -0,0 +1,155 @@
...
+     * @param flatIndex
+     *            the flat index to get the context for
+     * @return an {@link FlatIndexContext} record, or {@code null} if not found
+     */
+    public FlatIndexContext<T> getFlatIndexContext(int flatIndex) {
+        return getFlatIndexContext(this, flatIndex);
+    }
+
+    private FlatIndexContext<T> getFlatIndexContext(Cache<T> cache,
+            int flatIndex) {
```
tltv: Rename `flatIndex` parameter to `localFlatIndex` for clarity.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -400,107 +366,239 @@ public boolean hasChildren(T item) {
...
+
+        while (result.size() < Math.abs(length)) {
+            var context = rootCache.getFlatIndexContext(start);
+            if (context == null) {
+                break;
+            }
+            var cache = context.cache();
+            var index = context.index();
+
+            if (!cache.hasItem(index)) {
```
tltv, line 496: This if-block could be extracted into private method `preloadCache(Cache, ...)` to shorten method and for clarity.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -400,107 +366,239 @@ public boolean hasChildren(T item) {
...
+                        : Range.between(index + 1 - remainingLength, index + 1);
+
+                range = range.restrictTo(Range.withLength(0, cache.getSize()));
+
+                var items = fetchDataProviderChildren(cache.getParentItem(),
+                        range).toList();
+                cache.setItems(range.getStart(), items);
+            }
+
+            var item = cache.getItem(index);
```
tltv: Should we have a null check here since having null item here could cause interesting issues with isExpanded returning always true?

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -400,107 +366,239 @@ public boolean hasChildren(T item) {
...
+
+            if (isExpanded(item) && !cache.hasCache(index)
+                    && (direction > 0 || result.size() > 0)) {
```
tltv: This could be refactored a bit for better readability. Here's couple suggestions:

1. `(direction > 0 || result.size() > 0)` could be set in a local variable with a name like `allowSubCacheCreation`.

2. Or Cache could have `ensureCache` method to create cache only if it's allowed, but that would require some extra parameters e.g. `Cache#ensureCache(int index, SerializableSupplier<Integer> sizeSupplier, boolean allowCreation)`.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/RootCache.java
vursen: I'm not sure if `localFlatIndex` is the best name here since flat indexes are somewhat opposite to local indexes. Just to double-check, did you perhaps mean renaming `index` to `localIndex` in `Cache`?

vursen: Ah, I got what you mean 👍  sorry for the confusion. Thanks, will do.

tltv: Maybe I've missed something but I meant attribute in this private method. Looks like it's always local index when looked from caller perspective. public method takes in flatIndex, but it's used for the root cache there anyway.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
vursen: Thanks, done.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/RootCache.java
vursen: Done.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
vursen: How about splitting this method into separate ones for each direction? It might slightly increase the amount of code but it would make the control flow clearer. I've pushed the update, please let me know what you think.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
vursen: Good question. What would you expect in the array when the item isn't found? null?

vursen: That said, I guess there is a difference between preloadRange and resolveIndexPath in that the latter can be invoked with indexes coming from the client side, which means it may receive outdated or invalid paths, whereas `preloadRange` is intended to be used internally relying on the internal cache. So, it seems that this code can break only if the data provider returns incorrect data. In that case, would it be more appropriate to throw an exception?

tltv: It could return null since `HierarchicalDataProvider#fetchChildren` could return nulls. Maybe `isExpanded` mehod shouldn't return true when item is null but callers should decide what to do with null item themselves instead.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
tltv: It makes code a bit more readable. But is preloadFlatRangeBackward used now at all anywhere?

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
tltv: Yes, you are right. `fetchFromProvider` is not even used now by `HierarchicalDataCommunicator`. Maybe it's used by TreeGrid or something else since it's public. Overriding buildQuery method and using that primarily would make sense.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
vursen: Yes, it's used in [`TreeGrid#setViewportRangeByIndexPath`](https://github.com/vaadin/flow-components/blob/9b0f5ebc4392d44ad9d057bca444e1841cdc5f0d/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java#L1074-L1085) which is required for `scrollToIndex(int... indexes)`, see https://github.com/vaadin/flow-components/pull/7676#issue-3222937751 for an explanation.

I'm planning to add respective tests in Flow as well.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
vursen: Just for the record, TreeGrid doesn't use `fetchFromProvider` either.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/RootCache.java
vursen: Refactored the implementation a bit more by moving all flat-view-related methods to RootCache. This should improve code organization and avoid "flat vs local index" confusion when working with Cache.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
tltv: As commented [in here](https://github.com/vaadin/flow/pull/21966#discussion_r2257329023) already, this could be removed now. Default `DataProvider#getId(Object)` implementation will throw checked NPE for null.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
vursen: Added.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
vursen: After looking into this more deeply, I'm inclined to believe that null can't be a valid value in any case, so I suggest we always throw an exception when a null item is received from the data provider:

```
throw new IllegalStateException("Data provider returned a null item. Null values are not supported");
```

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
vursen: Done.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
tltv: `HierarchicalDataProvider#fetch` and `HierarchicalDataProvider#fetchChildren` can be also updated to tell that null values in the returned Stream are not supported.

vursen: Would it be enough to add this check to `HierarchicalDataProvider#fetch` only? I'm not sure how to achieve this for `fetchChildren` since it's a method that should be implemented by the user.

tltv: I think it would be enough to just mention it in Javadoc of both those methods.

vursen: Thanks, done.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/Cache.java
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * A cache for hierarchical data. Each instance of {@link Cache} represents a
+ * level in the hierarchy. It is used to store the size of the level, the items,
+ * and references to their child caches if they are expanded.
+ * <p>
+ * WARNING: This class is intended for internal use only and may change at any
+ * time without notice. It is not part of the public API and should not be used
+ * directly in your applications.
+ *
+ * @param <T>
```
mshabarov:
```suggestion
 * directly in your applications.
 *
 * @since 25.0
 *
 * @param <T>
```

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/RootCache.java
```diff
@@ -0,0 +1,215 @@
...
+
+/**
+ * An extension of {@link Cache} that represents the root level in the
+ * hierarchy.
+ * <p>
+ * WARNING: This class is intended for internal use only and may change at any
+ * time without notice. It is not part of the public API and should not be used
+ * directly in your applications.
+ *
+ * @param <T>
```
mshabarov:
```suggestion
 * directly in your applications.
 *
 * @since 25.0
 *
 * @param <T>
```

mshabarov (approved):

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -85,128 +105,123 @@ public class HierarchicalDataCommunicator<T> extends DataCommunicator<T> {
...
+     * update to re-render this item.
+     */
     @Override
-    public Stream<T> fetchFromProvider(int offset, int limit) {
-        // Instead of adding logic to this class, delegate request to the
-        // separate object handling hierarchies.
-        return mapper.fetchRootItems(Range.withLength(offset, limit));
+    public void refresh(T item) {
+        refresh(item, false);
     }
```
mshabarov, line 174: Could this be deprecated in 24.9 with the words that it's an internal API and will be replaced in Vaadin 25?

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -400,107 +368,270 @@ public boolean hasChildren(T item) {
...
-    public void setInMemorySorting(SerializableComparator<T> comparator) {
-        if (mapper != null) {
-            mapper.setInMemorySorting(comparator);
-        }
-        super.setInMemorySorting(comparator);
+    public int getDataProviderSize() {
+        return getDataProviderChildCount(null);
     }
 
-    protected <F> void setFilter(F filter) {
```
mshabarov: Same here, better to deprecate in 24.9 with possibly a suggestion what is the replacement. DataProvider add-ons may use these methods I assume.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -400,107 +368,270 @@ public boolean hasChildren(T item) {
...
+     * cache, starting from the root level, and returns the flat index of the
+     * target item.
      *
-     * @return {@code true} if there is any expanded items.
+     * @param path
+     *            the hierarchical path to the item, where each element
+     *            represents the index within its respective level
+     * @return the flat index of the target item after resolving all ancestors
      */
-    public boolean hasExpandedItems() {
```
mshabarov, line 449: Deprecate in 24.9.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
mshabarov: This wasn't actually removed, my bad.

PR #21969, opened 2025-07-30
feat: add base license checker init listener

mcollovati:


flow-server/src/main/java/com/vaadin/flow/server/startup/BaseLicenseCheckerServiceInitListener.java
```diff
@@ -0,0 +1,152 @@
...
+
+            String script = """
+                    window.Vaadin = window.Vaadin || {};
+                    window.Vaadin.devTools = window.Vaadin.devTools || {};
+                    window.Vaadin.devTools.createdCvdlElements =
+                    window.Vaadin.devTools.createdCvdlElements || []; const product =
+                    {}; product.constructor['cvdlName'] = '%1$s';
+                    product.constructor['version'] = '%2$s'; product.tagName =
+                    '--%1$s';
+                    window.Vaadin.devTools.createdCvdlElements.push(product);
```
tltv: Lots of boilerplate code before this line. It would be also repeated for each product adding lots of code. Is it because script is inserted before dev tools is initialized?

If script is added in the end of the head instead, after dev tools, could all boilerplate be removed?
Couple reasons why it seems so to me:
1. `vaadin-dev-tools.ts` imports `License.ts` which overrides `createdCvdlElements` function to check license on `push` call (in addition to doing it in dev tools `connectedCallback()`).
2. script is added only when dev tools is enabled anyway

mcollovati: Good point. However, I am not sure there is an absolute guarantee that a script added after another is also executed after.

mcollovati: Another option could be to register a single instance of `CheckProductLicense` listener.
The first add-on will register it, the others will just add their own product name and version.
This way there will be a single script that adds all products.

tltv: Probably since `async` or `defer` is not involved here, it's executing in order they appear.

Single listener adding all in one script sounds promising.

mcollovati: I'll give the single listener a try.

mcollovati: updated the code to use a single listener

tltv (approved):
LGTM. 
However, one potential issue with usage of VaadinServiceInitListener is that it could be excluded accidentally or in purpose from Flow's class scanner.

mshabarov (approved):

Issue #21675, opened 2025-06-09
Remove "construct-style-sheets-polyfill" in Vaadin 25


TatuLund:
This is supported by Safari 16.4 and Firefox 101 and no longer needed.

https://github.com/vaadin/flow/blob/main/flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/default/package.json#L7

PR #21977, opened 2025-08-01
fix!: remove construct-style-sheets-polyfill

tltv:
construct-style-sheets-polyfill dependency is no longer needed since supported browsers for Vaadin 25 don't need it.

Fixes: #21675


mshabarov:
Should we mark this as a breaking change for old browser versions?

tltv:
> Should we mark this as a breaking change for old browser versions?

Probably yes, it's a braking change since polyfill is still required with Safari 15...16.3.

PR #21981, opened 2025-08-04
feat: try to download license key when log in from splash screen

mcollovati:
When pressing the login button on the splash screen, it redirects the user to the vaadin.com license validation endpoint and polls the license server to download a license key. If the license key is available the application browser page gets reloaded. Otherwise an error message is shown.


vaadin-dev-server/src/main/frontend/pre-trial-splash-screen.ts
```diff
@@ -172,12 +176,13 @@ class PreTrial extends HTMLElement {
...
         this.openNewWindow('https://vaadin.com/pricing');
       } else {
-        (window as any).Vaadin.devTools.startPreTrial();
+        startPreTrial();
       }
     });
     const loginButton = this.#shadowRoot.querySelector('button.login-button')!;
     loginButton.addEventListener('click', () => {
-      this.openNewWindow('https://vaadin.com/my/account');
+      //this.openNewWindow('https://vaadin.com/my/account');
```
mshabarov: Can this be removed?

Also this is replaced by license download, but do we still need to call `openNewWindow` is a license cannot be downloaded by some reason?

mcollovati: Yeah, the comment should be removed.

If the license key is missing, we do not need to open a new window from the splash screen; License Checker will open a browser tab using the vaadin-open library.

mshabarov: Got it, thanks!
Let's remove the comment then.

mcollovati:
```suggestion
```

PR #22027, opened 2025-08-13
feat: Assets using NpmPackage

### LINKED ISSUE #21806: Load resources from npm packages for use in Flow
### Describe your motivation

In Vaadin 25 we are going to deprecated `Theme` annotation and features in `theme.json`, including `importCss` instruction that then needs a replacement.

A new Flow API for loading frontend assets (like stylesheets, images, etc) from npm packages (e.g. through additional parameters for the `NpmPackage` annotation) that allows you to specify a path that you can then use with `Image` in Java. 

### Additional context

Vaadin 25.0

caalador:
Fixes #21806

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/dependency/NpmPackage.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/dependency/NpmPackage.java
@@ -75,6 +75,15 @@ public @interface NpmPackage {
     boolean dev() default false;
 
     /**
+     * Assets to copy from this npm package for static use. Format is
+     * "copyFromPathInModule:copyTargetFolderUnderFrontend", e.g.
+     * "dist/line-awesome/css/**:line-awesome/dist/line-awesome/css"
+     *
+     * @return array of assets to copy from npm package
+     */
+    String[] assets() default {};
+
+    /**
      * Internal annotation to enable use of multiple {@link NpmPackage}
      * annotations.
      */
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyNpmAssetsFiles.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyNpmAssetsFiles.java
@@ -178,7 +178,7 @@ public class TaskCopyNpmAssetsFiles
         log().debug("getting files using pattern {}", pattern);
 
         final PathMatcher matcher = FileSystems.getDefault()
-                .getPathMatcher("glob:**" + pattern);
+                .getPathMatcher("glob:" + pattern);
 
         filePaths.addAll(getFileNames(basePath, matcher));
 
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyNpmAssetsFiles.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyNpmAssetsFiles.java
@@ -1,0 +1,218 @@
+/*
+ * Copyright 2000-2025 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.server.frontend;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.FileSystems;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.security.InvalidParameterException;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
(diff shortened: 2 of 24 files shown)


flow-server/src/main/java/com/vaadin/flow/component/dependency/NpmPackage.java
```diff
@@ -74,6 +74,15 @@
...
     boolean dev() default false;
 
+    /**
+     * Assets to copy from this npm package for static use. Format is
+     * "copyFromPathInModule:copyTargetFolderUnderFrontend", e.g.
+     * "dist/line-awesome/css/**:line-awesome/dist/line-awesome/css"
+     *
+     * @return array of assets to copy from npm package
+     */
+    String[] assets() default {};
```
knoobie, line 84: @mshabarov Could this change be so small to allow targeting 24.9 as well? This would allow people to already migrate some parts of their theme.json (assets) to this new mechanism before 25.0 hits like a truck?

mshabarov: Sounds like a good idea. This feature stays apart and doesn't depend on the new V25 theming system, so should be possible to backport it. Let us try to backport and if no big blockers, let's expect it to be in 24.9 as well.

knoobie: This might also allow you to deprecate "assets" from the theme.json and e.g. printing a warning when building or starting an application still containing those to allow people to easily find this change and adapt.

mshabarov: Whole `theme.json` feature will be deprecated but still supported in Vaadin 25, so that's not necessary to do the deprecation in 24.9 IMO.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyNpmAssetsFiles.java
```diff
@@ -0,0 +1,209 @@
...
+                        .getPath();
+            }
+
+            staticOutput = new File(webappResources, VAADIN_STATIC_FILES_PATH);
+        }
+    }
+
+    @Override
+    public void execute() {
+        if (!options.copyAssets() && !options.isDevBundleBuild()) {
```
tltv: shouldn't this return always if `copyAssets()` alone is false?

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyNpmAssetsFiles.java
```diff
@@ -0,0 +1,209 @@
...
+                        } catch (IOException e) {
+                            throw new UncheckedIOException(String.format(
+                                    "Failed to copy project frontend resources from '%s' to '%s'",
+                                    file, destFile), e);
+                        }
+                    }
+                });
+            });
+        });
+    }
```
tltv: nit: Looks like a good candidate to test a bit AI refactoring here. Extracted into smaller methods. it's formatted. Looks correct to me. Feel free to write better version:
```suggestion
    private void copyNpmAssets(Map<String, List<String>> npmAssets) {
        npmAssets.forEach((npmModule, npmAssetList) -> npmAssetList
                .forEach(npmAsset -> processNpmAsset(npmModule, npmAsset)));
    }

    private void processNpmAsset(String npmModule, String npmAsset) {
        if (npmAsset.isBlank()) {
            return;
        }
        String[] split = validateAndSplitNpmAsset(npmModule, npmAsset);
        File npmModuleDir = new File(options.getNodeModulesFolder(), npmModule);
        List<Path> paths = collectFiles(npmModuleDir.toPath(),
                split[0].strip());
        log().debug("Paths amount {}", paths.size());
        paths.stream().map(Path::toFile)
                .forEach(file -> copyFileToTarget(file, split[1].strip()));
    }

    private String[] validateAndSplitNpmAsset(String npmModule,
            String npmAsset) {
        String[] split = npmAsset.split(":");
        if (split.length != 2) {
            throw new InvalidParameterException("Invalid npm asset: " + npmAsset
                    + " for npm module: " + npmModule);
        }
        log().debug("Rule {} to {}", split[0], split[1]);
        return split;
    }

    private void copyFileToTarget(File file, String targetPath) {
        File targetFolder = new File(staticOutput, targetPath);
        File destFile = new File(targetFolder, file.getName());
        if (!destFile.exists()
                || destFile.lastModified() < file.lastModified()) {
            log().debug("Copying npm file {} to {}", file.getAbsolutePath(),
                    destFile.getAbsolutePath());
            try {
                FileUtils.copyFile(file, destFile);
            } catch (IOException e) {
                throw new UncheckedIOException(String.format(
                        "Failed to copy project frontend resources from '%s' to '%s'",
                        file, destFile), e);
            }
        }
    }
```

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
```diff
@@ -333,7 +333,7 @@ private static class DataAnnotationVisitor
 
         @Override
         public AnnotationVisitor visitArray(String name) {
-            if (isRepeatableContainer) {
+            if (isRepeatableContainer && !"assets".equals(name)) {
```
tltv: Use `FrontendClassVisitor.ASSETS`

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyNpmAssetsFiles.java
caalador: copyAssets skips the copy except if we have a devBundleBuild in which case we should copy the assets.
So for a dev bundle build the copyAssets flag doesn't block copying as we need to copy as there will be no later execution after prepare frontend which running a dev build actually will have.

flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyNpmAssetsFiles.java
caalador: simplified.

flow-server/src/main/java/com/vaadin/flow/server/frontend/scanner/FrontendAnnotatedClassVisitor.java
caalador: done

PR #22028, opened 2025-08-13
fix: Fixed high CPU usage due to pending JS commands (#22024)

### LINKED ISSUE #22024: UIInternals field pendingJsInvocations should be a Set for performance reasons
### Description of the bug

Hello,

The field UIInternals.**pendingJsInvocations** generates very large CPU usage and hangs our application, therefore it should be a Set instead of an ArrayList, for performance reasons. 

Our application contains an interactive area which has many fields. Some of these fields are hidden/readonly. On each user interaction, fields are updated. Because some fields are not visible, the field UIInternals.**pendingJsInvocations** keeps growing. We can grow this field to size 50.000 very quickly. 

When the application closes the interactive state, the UI is cleaned up:
-  Vaadin framework removes child nodes and all **PendingJavaScriptInvocation** objects.
- **PendingJavaScriptInvocationDetachListener.execute()** is called for each component
- If each of our components has ~400 pending invocations, and we have 200 components, and there are 50.000 pending invocations total, the application hangs due to a very large loop: 50000 x 400 x 200 = ~4 billion iterations

Proposed fix:
The field UIInternals.**pendingJsInvocations** should be a Set instead of an ArrayList, and removal of PendingJavaScriptInvocation objects should be done using **HashSet.remove()** instead of Collection.removeIf()

Image

### Expected behavior

The application should not hang due to pending Javascript invocations that are no longer needed

-

### Versions

- Vaadin / Flow version: 24.3
- Java version: 21
- OS version: Windows 10
- Browser version (if applicable): Chrome/Firefox
- IDE (if applicable): Eclipse 2024

daniel-pss-prosoft:
## Description

Field "pendingJsInvocations" should be a Set instead of a List in order to avoid O(n^2) operations. This greatly improves performance in situations where many fields are updated constantly and the JS command queue becomes very large.

Fixes #22024

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/internal/UIInternals.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -184,7 +185,7 @@ public class UIInternals implements Serializable {
      */
     private long lastHeartbeatTimestamp = System.currentTimeMillis();
 
-    private List<PendingJavaScriptInvocation> pendingJsInvocations = new ArrayList<>();
+    private Set<PendingJavaScriptInvocation> pendingJsInvocations = new LinkedHashSet<>();
 
     private final HashMap<StateNode, PendingJavaScriptInvocationDetachListener> pendingJsInvocationDetachListeners = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


daniel-pss-prosoft:
I signed the CLA 20 minutes ago, but it isn't marked as signed... not sure if this is a problem.

image

Legioth:
It seems like the email address used in the commit is not associated with your GitHub user account. See https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user for more info.

daniel-pss-prosoft:
Bad commit username/email, i'll make another request

PR #22037, opened 2025-08-14
fix(spring): allow authenticated Hilla endpoint requests in Spring Security by default

platosha:


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurity.java
```diff
@@ -245,6 +245,9 @@ protected void configure(HttpSecurity http) throws Exception {
             // matcher for custom PWA icons and favicon
             urlRegistry.requestMatchers(requestUtil::isCustomWebIcon)
                     .permitAll();
+            // authenticated endpoints
+            urlRegistry.requestMatchers(requestUtil::isEndpointRequest)
```
knoobie: This should not be here, it sounds like something to be within `defaultPermitMatcher` where Hilla rules are inserted

platosha: It's not suitable for `defaultPermitMatcher` as this rule deals with authenticated Hilla endpoints. Also, mind that it's the deprecated file in the old style, not using `defaultPermitMatcher`.

knoobie: Then `if (EndpointRequestUtil.isHillaAvailable()) {` should be added around it.

mshabarov: `isEndpointRequest` internally checks the presence of `endpointRequestUtil` that is empty if no Hilla is available, thus it returns immediately.

knoobie: But it's still another matcher that is registered and spring is always executing it

platosha: Well, yes, would be more optimal to avoid adding the matcher in the first place. Will do.

platosha: Done, thanks for the suggestion.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -744,8 +744,14 @@ private AccessDeniedHandler createAccessDeniedHandler() {
     private void customizeAuthorizeHttpRequests(
             AuthorizeHttpRequestsConfigurer<HttpSecurity>.AuthorizationManagerRequestMatcherRegistry registry) {
         registry.requestMatchers(defaultPermitMatcher()).permitAll()
+                .requestMatchers(getRequestUtil()::isEndpointRequest)
+                .authenticated()
```
knoobie: Note: both lines can be removed now :)

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java
```diff
@@ -746,6 +746,10 @@ private void customizeAuthorizeHttpRequests(
         registry.requestMatchers(defaultPermitMatcher()).permitAll()
                 .requestMatchers(getRequestUtil()::isSecuredFlowRoute)
                 .authenticated();
+        if (EndpointRequestUtil.isHillaAvailable()) {
```
cromoteca, line 749: It seems to ignore `@AnonymousAllowed` endpoints.

platosha: Aren't they covered in `defaultPermitMatcher` already? See https://github.com/vaadin/flow/blob/0400e9de47400ff3fd8017feb120987c1bcd12d8/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinSecurityConfigurer.java#L484-L490

PR #22039, opened 2025-08-14
fix: filename encoding of download

### LINKED ISSUE #22007: DownloadHandler incorrectly handles files with non-ASCII characters in filename
### Description of the bug

Hi, 
Recently I switched to Vaadin 24.8.3 and migrated all "download" places to new DownloadHandler way. It's much easier and convenient, but whenever I try to download file with non-ascii characters in name (in my case with accents - `Re nedoručený časopis dne 4.2.2025 - zák. č. 12312312 Ing. Zděnek Řehoř.msg`), the file is downloaded but the filename looks like: `cd0e9a7f-c1b2-4958-9cbf-66b6181d8a37`. If I strip the accents in DownloadHandler calling, everything is working correctly.
Before this change everything worked correctly.

This is Vaadin forum issue related to this: [https://vaadin.com/forum/t/downloadhandler-nonascii-filename/178386](https://vaadin.com/forum/t/downloadhandler-nonascii-filename/178386)

Thx
JS

### Expected behavior

Download file with correct filename as was uploaded to the application. 

```
var fileName = "Re nedoručený časopis dne 4.2.2025 - zák. č. 12312312 Ing. Zděnek Řehoř.msg";
var fileSize = 127L;
var contentType = "application/vnd.ms-outlook";
var anchor = new Anchor(DownloadHandler.fromInputStream(e -> new DownloadResponse(loadInputStream(), fileName, contentType, fileSize)), fileName);
```

### Versions

- Vaadin / Flow version: 24.8.3
- Java version: 21
- OS version: MacOS 15.5
- Browser version (if applicable): Safari/Chrome
- Application Server (if applicable): Tomcat
- IDE (if applicable): Intellij

tltv:
Applies utf-8 encoding by default in Content-Disposition header for DownloadEvent.

Fixes: #22007

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
@@ -150,9 +150,20 @@ public class DownloadEvent {
         if (fileName.isEmpty()) {
             response.setHeader("Content-Disposition", "attachment");
         } else {
-            response.setHeader("Content-Disposition",
-                    "attachment; filename*=UTF-8''"
-                            + EncodeUtil.rfc5987Encode(fileName));
+            StringBuilder value = new StringBuilder();
+            value.append("attachment; ");
+            if (EncodeUtil.isPureUSASCII(fileName)) {
+                value.append("filename=\"").append(fileName).append("\"");
+            } else {
+                value
+                        // fallback legacy support
+                        .append("filename=\"")
+                        .append(EncodeUtil.rfc2047Encode(fileName))
+                        // used primarily
+                        .append("\"; filename*=UTF-8''")
+                        .append(EncodeUtil.rfc5987Encode(fileName));
+            }
+            response.setHeader("Content-Disposition", value.toString());
         }
         this.fileName = fileName;
     }
--- a/flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
@@ -27,6 +27,7 @@ import org.slf4j.LoggerFactory;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.internal.EncodeUtil;
 import com.vaadin.flow.server.VaadinRequest;
 import com.vaadin.flow.server.VaadinResponse;
 import com.vaadin.flow.server.VaadinSession;
@@ -150,7 +151,8 @@ public class DownloadEvent {
             response.setHeader("Content-Disposition", "attachment");
         } else {
             response.setHeader("Content-Disposition",
-                    "attachment; filename=\"" + fileName + "\"");
+                    "attachment; filename*=UTF-8''"
+                            + EncodeUtil.rfc5987Encode(fileName));
         }
         this.fileName = fileName;
     }
--- a/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DownloadHandlerView.java
+++ b/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DownloadHandlerView.java
@@ -54,9 +54,18 @@ public class DownloadHandlerView extends Div {
         File jsonFile = new File(getClass().getClassLoader()
                 .getResource("download.json").getFile());
 
+        File jsonFileUnicodeName = new File(getClass().getClassLoader()
+                .getResource("download.json").getFile());
+
         Anchor fileDownload = new Anchor("", "File DownloadHandler shorthand");
         fileDownload.setHref(DownloadHandler.forFile(jsonFile).inline());
         fileDownload.setId("download-handler-file");
+
+        Anchor fileDownloadUnicodeName = new Anchor("",
+                "File (unicode name) DownloadHandler shorthand");
+        fileDownloadUnicodeName.setHref(DownloadHandler
+                .forFile(jsonFileUnicodeName, "download-Řřüñîçødë.json"));
+        fileDownloadUnicodeName.setId("download-handler-file-unicode");
 
         Anchor classDownload = new Anchor("",
                 "Class resource DownloadHandler shorthand");
@@ -117,9 +126,10 @@ public class DownloadHandlerView extends Div {
         inputStreamCallbackError
                 .setId("download-handler-input-stream-callback-error");
 
-        add(handlerDownload, fileDownload, classDownload, servletDownload,
-                inputStreamDownload, inputStreamErrorDownload,
-                inputStreamExceptionDownload, inputStreamCallbackError);
+        add(handlerDownload, fileDownload, fileDownloadUnicodeName,
+                classDownload, servletDownload, inputStreamDownload,
+                inputStreamErrorDownload, inputStreamExceptionDownload,
+                inputStreamCallbackError);
 
         NativeButton reattach = new NativeButton("Remove and add back",
                 event -> {
--- a/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DownloadHandlerView.java
+++ b/flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DownloadHandlerView.java
@@ -54,18 +54,22 @@ public class DownloadHandlerView extends Div {
         File jsonFile = new File(getClass().getClassLoader()
                 .getResource("download.json").getFile());
 
-        File jsonFileUnicodeName = new File(getClass().getClassLoader()
-                .getResource("download.json").getFile());
-
         Anchor fileDownload = new Anchor("", "File DownloadHandler shorthand");
         fileDownload.setHref(DownloadHandler.forFile(jsonFile).inline());
         fileDownload.setId("download-handler-file");
 
         Anchor fileDownloadUnicodeName = new Anchor("",
                 "File (unicode name) DownloadHandler shorthand");
-        fileDownloadUnicodeName.setHref(DownloadHandler
-                .forFile(jsonFileUnicodeName, "download-Řřüñîçødë.json"));
+        fileDownloadUnicodeName.setHref(DownloadHandler.forFile(jsonFile,
+                "download-Řřüñîçødë 1中文.json"));
         fileDownloadUnicodeName.setId("download-handler-file-unicode");
+
+        Anchor fileDownloadUnicodeNameWithQuote = new Anchor("",
+                "File (unicode name with quote) DownloadHandler shorthand");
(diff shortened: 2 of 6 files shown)


flow-tests/test-root-context/src/main/java/com/vaadin/flow/uitest/ui/DownloadHandlerView.java
```diff
@@ -54,10 +54,19 @@ public String getUrlPostfix() {
...
         Anchor fileDownload = new Anchor("", "File DownloadHandler shorthand");
         fileDownload.setHref(DownloadHandler.forFile(jsonFile).inline());
         fileDownload.setId("download-handler-file");
 
+        Anchor fileDownloadUnicodeName = new Anchor("",
+                "File (unicode name) DownloadHandler shorthand");
+        fileDownloadUnicodeName.setHref(DownloadHandler
+                .forFile(jsonFileUnicodeName, "download-Řřüñîçødë.json"));
+        fileDownloadUnicodeName.setId("download-handler-file-unicode");
+
```
knoobie, line 66: I would suggest to also include a test with a filename containing a space character.. while we Developer hate those.. end-user love to use spaces 😅

knoobie: Another problem / test-case might be https://github.com/spring-projects/spring-framework/commit/f0e7b42704e6b33958f242d91bd690d6ef7ada9c

tltv: Added tests for  ` "1中文`. Download already works with them.
 
Interesting finding is that in tests running with jetty 12, quote `"` in url causes jetty to throw `MalformedURLException` which is handled gracefully and logged as a warning by Flow when it checks download request for static resource before passing it to download handler.

mstahv:
Maybe revisit the implementation EncodeUtil. Spring's implementation is slightly different:
```
Mismatch for 0:  Vaadin: UTF-8''%0 Spring: UTF-8''%00
Mismatch for 1:  Vaadin: UTF-8''%1 Spring: UTF-8''%01
Mismatch for 2:  Vaadin: UTF-8''%2 Spring: UTF-8''%02
Mismatch for 3:  Vaadin: UTF-8''%3 Spring: UTF-8''%03
Mismatch for 4:  Vaadin: UTF-8''%4 Spring: UTF-8''%04
Mismatch for 5:  Vaadin: UTF-8''%5 Spring: UTF-8''%05
Mismatch for 6:  Vaadin: UTF-8''%6 Spring: UTF-8''%06
Mismatch for 7:  Vaadin: UTF-8''%7 Spring: UTF-8''%07
Mismatch for 8: Vaadin: UTF-8''%8 Spring: UTF-8''%08
Mismatch for 9: 	 Vaadin: UTF-8''%9 Spring: UTF-8''%09
Mismatch for 10: 
 Vaadin: UTF-8''%A Spring: UTF-8''%0A
Mismatch for 11:  Vaadin: UTF-8''%B Spring: UTF-8''%0B
Mismatch for 12:  Vaadin: UTF-8''%C Spring: UTF-8''%0C
 Vaadin: UTF-8''%D Spring: UTF-8''%0D
Mismatch for 14:  Vaadin: UTF-8''%E Spring: UTF-8''%0E
Mismatch for 15:  Vaadin: UTF-8''%F Spring: UTF-8''%0F
Mismatch for 33: ! Vaadin: UTF-8''%21 Spring: UTF-8''!
Mismatch for 35: # Vaadin: UTF-8''%23 Spring: UTF-8''#
Mismatch for 36: $ Vaadin: UTF-8''%24 Spring: UTF-8''$
Mismatch for 38: & Vaadin: UTF-8''%26 Spring: UTF-8''&
Mismatch for 43: + Vaadin: UTF-8''%2B Spring: UTF-8''+
Mismatch for 45: - Vaadin: UTF-8''%2D Spring: UTF-8''-
Mismatch for 94: ^ Vaadin: UTF-8''%5E Spring: UTF-8''^
Mismatch for 95: _ Vaadin: UTF-8''%5F Spring: UTF-8''_
Mismatch for 96: ` Vaadin: UTF-8''%60 Spring: UTF-8''`
Mismatch for 124: | Vaadin: UTF-8''%7C Spring: UTF-8''|
Mismatch for 126: ~ Vaadin: UTF-8''%7E Spring: UTF-8''~

```

tltv:
> Maybe revisit the implementation EncodeUtil. Spring's implementation is slightly different:
> 
Corrected as RFC 5987 allows !#$&+-.^_`|~ without encoding and put inline with Spring's implementation.

mshabarov (dismissed):

mshabarov (dismissed):

flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
```diff
@@ -150,7 +151,8 @@ public void setFileName(String fileName) {
             response.setHeader("Content-Disposition", "attachment");
         } else {
             response.setHeader("Content-Disposition",
-                    "attachment; filename=\"" + fileName + "\"");
+                    "attachment; filename*=UTF-8''"
+                            + EncodeUtil.rfc5987Encode(fileName));
```
knoobie: nit: According to Appendix D "Advice on Generating Content-Disposition Header Fields" in https://httpwg.org/specs/rfc6266.html#header.field.definition

It might be worth to think about:

> Include a "filename" parameter when US-ASCII ([[US-ASCII]](https://httpwg.org/specs/rfc6266.html#US-ASCII)) is sufficiently expressive.
> Include a "filename*" parameter where the desired filename cannot be expressed faithfully using the "filename" form. Note that legacy user agents will not process this, and will fall back to using the "filename" parameter's content.
> When a "filename*" parameter is sent, to also generate a "filename" parameter as a fallback for user agents that do not support the "filename*" form, if possible. This can be done by substituting characters with US-ASCII sequences (e.g., Unicode character point U+00E4 (LATIN SMALL LETTER A WITH DIARESIS) by "ae"). Note that this may not be possible in some locales.

```java
var encodedFileName = ncodeUtil.rfc5987Encode(fileName);
if(encodedFileName != fileName) {
// use filename* + (optionally unsure myself if you wanna support legacy) filename
} else
// use filename
```

tltv: Good point, I thought nobody brings it up. I leaved it out intentionally because the Legacy user agent most likely means IE 9 and predecessors, which are not supported by Vaadin anymore. But could there really be some user agent that Vaadin does support but doesn't understand "filename*"? If so, then filename has to be added too.

knoobie: I'm also torn by the inclusion of both fields at the same time. I'm more thinking about the usage of filename or filename* depending on the file name's characters to ease consumption for the clients like recommend above.

mshabarov: The impact is probably low- legacy agent + non-ASCII characters, but from the other side I don't want to come back to this topic again if someone requests a support for this case. Let's add the handling for both if it's not that hard?

tltv: Need to test that `filename` doesn't break with special characters.

knoobie: I would suggest to get inspiration from https://github.com/spring-projects/spring-framework/blob/f0e7b42704e6b33958f242d91bd690d6ef7ada9c/spring-web/src/main/java/org/springframework/http/ContentDisposition.java#L254 - it's battle proven :) no need to re-invent the wheel

tltv: I agree since we don't have any existing code to encode RFC 2047. We can't add dependency naturally as this should work also without Spring.

tltv: Added support for RFC 2047 encoded `filename` attribute for better legacy support and pure US-ASCII detection to use filename attribute alone without encoding with US-ASCII text.

flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
```diff
@@ -150,9 +150,20 @@ public void setFileName(String fileName) {
...
+            } else {
+                value
+                        // fallback legacy support
+                        .append("filename=\"")
+                        .append(EncodeUtil.rfc2047Encode(fileName))
+                        // used primarily
+                        .append("\"; filename*=UTF-8''")
+                        .append(EncodeUtil.rfc5987Encode(fileName));
+            }
+            response.setHeader("Content-Disposition", value.toString());
```
knoobie, line 166: Would it make sense for you to move this whole attachment value part into a package protected static method?

This would allow testing and e.g. re-useability for the inline part ;) (topic for another day)

e.g. 

```java

@NonNull
static String generateContentDisposition(@NonNull String type, @NonNull String fileName) {
            StringBuilder value = new StringBuilder();
            value.append(type).append("; ");
            if (EncodeUtil.isPureUSASCII(fileName)) {
                value.append("filename=\"").append(fileName).append("\"");
            } else {
                value
                        // fallback legacy support
                        .append("filename=\"")
                        .append(EncodeUtil.rfc2047Encode(fileName))
                        // used primarily
                        .append("\"; filename*=UTF-8''")
                        .append(EncodeUtil.rfc5987Encode(fileName));
            }
            return value.toString();
  }


response.setHeader("Content-Disposition", generateContentDisposition("attachment", fileName);

```

tltv: I don't see problems moving it later when it's needed for `inline` or something else. Maybe even new utility class would make sense. Something similar that Spring has.

caalador (approved):

lipisak:
Hello,
I guess the change did't get to release. I have bumped Vaadin to 24.8.7 which includes vaadin-server 24.8.6 but the DownloadEvent is still the same (doesn't contain change mentioned above). Was merge to builded branch successful?

Thank you

JS

caalador:
It seems it is not in flow release 24.8.6 even though for some reason the change notes say it would be.
The Change was merged 6 minutes after the build was created so it will come with Flow 24.8.7

flow-server/src/main/java/com/vaadin/flow/server/streams/DownloadEvent.java
knoobie: @tltv Just a note, cause I suggested the "battle proven" impl - it changed.. you might have to check your impl as well. https://github.com/spring-projects/spring-framework/commit/68338aa81891eb43afff7d6da9ad9b23c4db747e

PR #22019, opened 2025-08-12
refactor!: Centralize hierarchy management server-side

vursen:
## Description

> [!NOTE]
> This PR is a reopen of #21966 after its unintentional merge and revert.

> [!WARNING]
> DO NOT MERGE until the [related flow-components PR](https://github.com/vaadin/flow-components/pull/7676) is approved.

The PR refactors `HierarchicalDataCommunicator` so that hierarchy management (expanded items, hierarchical cache) is centralized on the server side instead of being shared between client and server. This makes the client a simple consumer of flattened data lists, which significantly reduces the client-side complexity and opens the door for introducing [flat data providers](https://github.com/vaadin/flow/pull/22040) to support fetching flattened data directly from the data source. 

### Key changes

1. `HierarchyMapper` and `HierarchicalCommunicationController` have been replaced with the new concept, `Cache`. This new class provides a system for storing data in a hierarchical structure while enabling access in a flattened format for client-side consumption. 
2. `setRequestedRange` and `setParentRequestedRange` have been replaced with a single `setViewportRange` which spans all hierarchy levels.

### Caveats

1. `setViewportRange` does not currently remove items from Cache and KeyMapper when they move out of the viewport. However, the previous version also only removed items from KeyMapper while keeping them in HierarchyMapper, so item instances actually continued to remain in memory. I feel this could be optimized separately if needed.
2. `setViewportRange`  currently always sends the full viewport to the client even when some items are already present on the client-side. I plan to address this separately, see [#21989](https://github.com/vaadin/flow/issues/21989)
4. `refresh(T item)` currently also sends the full viewport on every item refresh. I plan to address this separately, see [#21989](https://github.com/vaadin/flow/issues/21989)

### Test coverage

1. The PR aims at providing test coverage for all core functionality and public APIs. Individual tests for internal APIs such as `Cache` will be added separately.
2. The refactoring successfully passes all TreeGrid ITs, see [#7676](https://github.com/vaadin/flow-components/pull/7676)

Depends on 
- #21965 

Fixes #21876, #21877

## Type of change

- [x] Refactor

Issue #21878, opened 2025-07-15
Introduce isFlatHierarchy and getDepth to HierarchicalDataProvider


vursen:
1. Introduce `isFlatHierarchy` and `getDepth` methods to HierarchicalDataProvider:

```java
/** Override to enable flat hierarchy mode for a data provider */
default void isFlatHierarchy() {
  return false;
}

/** Provide implementation when isFlatHierarchy is set to true */
default int getDepth(T item) {
    throw new UnsupportedOperationException();
}
```

2. Update the new HierarchicalDataCommunicator to switch to flat hierarchy mode when `isFlatHierarchy` returns `true`.

PR #22040, opened 2025-08-14
feat: Introduce flattened hierarchy format

### LINKED ISSUE #21880: Introduce getExpandedItemIds to HierarchicalQuery
Update `HierarchicalQuery` to accept expanded item ids from the new HierarchicalDataCommunicator:
 
1. `public HierarchicalQuery(F filter, Set<Object> expandedItemIds, T parent)`
2. `public HierarchicalQuery(int offset, int limit, List<QuerySortOrder> sortOrders, Comparator<T> inMemorySorting, F filter, Set<Object> expandedItemIds, T parent)`
3. `Set<Object> getExpandedItemIds()`

vursen:
## Description

The PR introduces a new `getHierarchyFormat()` method to HierarchicalDataProvider. It can be configured to return either `HierarchyFormat#NESTED` (default) or `HierarchyFormat#FLATTENED` (new). The selected format defines in what way the data provider returns hierarchical data and how HierarchicalDataCommunicator should fetch and render it.

Image

### Nested Hierarchy Format (default)

The nested hierarchy format refers to a data provider implementation in which each expanded parent's children are fetched via a separate request, on demand:

```
└── Item 0               | Fetched in 1st request
    └── Item 0-0         | Fetched in 2nd request
        └── Item 0-0-0   | Fetched in 3rd request
└── Item 1               | Fetched in 1st request
```

Every request to the data provider returns a paginated list that contains only the direct children of the `HierarchicalQuery#getParent()`. The component decides when to request deeper levels and how much to load based on the current viewport and the expansion state of items, storing this lazy loaded hierarchy state in memory.

Example:

```java
class MyDataProvider
        implements HierarchicalDataProvider<String, Void> {
    private HashMap<String, List<String>> data = new HashMap<>() {
        {
            put(null, List.of("Item 0", "Item 1"));
            put("Item 0", List.of("Item 0-0"));
            put("Item 0-0", List.of("Item 0-0-0"));
        }
    };

    @Override
    public HierarchyFormat getHierarchyFormat() {
        return HierarchyFormat.NESTED;
    }

    @Override
    public Stream<String> fetchChildren(
            HierarchicalQuery<String, Void> query) {
        return data.get(query.getParent()).stream()
                .skip(query.getOffset()).limit(query.getLimit());
    }

    @Override
    public int getChildCount(HierarchicalQuery<String, Void> query) {
        return data.get(query.getParent()).size();
    }
}
```

Pros:
- Simple and fast data queries by avoiding hierarchy construction – each request fetches only the direct children, which are then cached hierarchically.

Cons:
- The full size and structure of the tree remains unknown without recursively fetching the entire hierarchy, which is impractical due to potentially a lot of consecutive requests and heavy memory usage. As a result, the scroll position cannot be restored automatically after using `HierarchicalDataProvider#refreshAll()` which resets the cached hierarchy state.
- The scroll container size updates dynamically while the user scrolls, which can cause them to jump over and miss some levels when scrolling quickly.

### Flattened Hierarchy Format (new)

The flattened hierarchy format refers to a data provider implementation that returns the entire subtree of the `HierarchicalQuery#getParent()` in a single, flattened, paginated list:

```
└── Item 0
└── Item 0-0
└── Item 0-0-0
└── Item 1
```

The list contains all expanded descendants, arranged in depth-first order: starting from the parent, then its children and their
descendants, before the next sibling. The component delegates hierarchy construction entirely to the data provider by supplying a `HierarchicalQuery` where:
- `HierarchicalQuery#getOffset()` points to the start of the requested range across all flattened descendants
- `HierarchicalQuery#getLimit()` specifies the number of items to return
- `HierarchicalQuery#getExpandedItemIds()` contains the set of expanded item IDs

The flattened format also requires the data provider to implement `HierarchicalDataProvider#getDepth(T)`, which the component uses to make items appear as a hierarchy by applying visual indentation.

Example:

```java
class MyDataProvider
        implements HierarchicalDataProvider<String, Void> {
    private HashMap<String, List<String>> data = new HashMap<>() {
        {
            put(null, List.of("Item 0", "Item 1"));
            put("Item 0", List.of("Item 0-0"));
            put("Item 0-0", List.of("Item 0-0-0"));
        }
    };

    @Override
    public HierarchyFormat getHierarchyFormat() {
        return HierarchyFormat.FLATTENED;
    }

    @Override
    public Stream<String> fetchChildren(
            HierarchicalQuery<String, Void> query) {
        return flatten(query.getParent(), query.getExpandedItemIds())
                .skip(query.getOffset()).limit(query.getLimit());
    }

    @Override
    public int getChildCount(HierarchicalQuery<String, Void> query) {
        return (int) flatten(query.getParent(),
                query.getExpandedItemIds()).count();
    }

    @Override
    public int getDepth(String item) {
        return item.split("-").length - 1;
    }

    private Stream<String> flatten(String parent,
            Set<Object> expandedItemIds) {
        return data.getOrDefault(parent, List.of()).stream().flatMap(
                child -> expandedItemIds.contains(getId(child))
                        ? Stream.concat(Stream.of(child),
                                flatten(child, expandedItemIds))
                        : Stream.of(child));
    }
}
```

Pros:

- The component can fetch the total size of the tree upfront and set a fixed scroll container size that does not change while scrolling, resulting in more predictable behavior.
- `refreshAll()` avoids unexpected scroll jumps by refetching the total tree size and retaining the size of expanded items whose nested structure hasn't changed.
- The developer has full control over how data is queried, making it possible to leverage more advanced optimizations and storage strategies at the database level.

Cons:
- Increased complexity and potentially heavier data queries due to the need for hierarchy reconstruction, which may require, for example, the use of recursive CTEs (Common Table Expressions) to retrieve all descendants of an item in a single SQL query.

Depends on 
- https://github.com/vaadin/flow/pull/22019

Fixes #21880, #21878 

## Type of change

- [x] Feature

### DIFF
--- a/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataProvider.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataProvider.java
@@ -53,19 +53,19 @@ public interface HierarchicalDataProvider<T, F> extends DataProvider<T, F> {
          * └── Item 1               | Fetched in 1st request
          * </pre>
          *
-         * Each request to the data provider returns a paginated list containing
-         * only the direct children of the {@link HierarchicalQuery#getParent()
-         * requested parent}. The component decides when to request deeper
-         * levels and how much to load based on the current viewport and the
-         * expansion state of items, storing the loaded hierarchy state in
-         * memory.
+         * Every request to the data provider returns a paginated list that
+         * contains only the direct children of the
+         * {@link HierarchicalQuery#getParent() requested parent}. The component
+         * decides when to request deeper levels and how much to load based on
+         * the current viewport and the expansion state of items, storing the
+         * loaded hierarchy state in memory.
          * <p>
          * Example:
          *
          * <pre>
          * class MyDataProvider
-         *         implements HierarchicalDataProvider<String, Void> {
-         *     private HashMap<String, List<String>> data = new HashMap<>() {
+         *         implements HierarchicalDataProvider&lt;String, Void&gt; {
+         *     private HashMap&lt;String, List&lt;String&gt;&gt; data = new HashMap&lt;&gt;() {
          *         {
          *             put(null, List.of("Item 0", "Item 1"));
          *             put("Item 0", List.of("Item 0-0"));
@@ -79,14 +79,14 @@ public interface HierarchicalDataProvider<T, F> extends DataProvider<T, F> {
          *     }
          *
          *     &#64;Override
-         *     public Stream<String> fetchChildren(
-         *             HierarchicalQuery<String, Void> query) {
+         *     public Stream&lt;String&gt; fetchChildren(
+         *             HierarchicalQuery&lt;String, Void&gt; query) {
          *         return data.get(query.getParent()).stream()
          *                 .skip(query.getOffset()).limit(query.getLimit());
          *     }
          *
          *     &#64;Override
-         *     public int getChildCount(HierarchicalQuery<String, Void> query) {
+         *     public int getChildCount(HierarchicalQuery&lt;String, Void&gt; query) {
          *         return data.get(query.getParent()).size();
          *     }
          * }
@@ -149,8 +149,8 @@ public interface HierarchicalDataProvider<T, F> extends DataProvider<T, F> {
          *
          * <pre>
          * class MyDataProvider
-         *         implements HierarchicalDataProvider<String, Void> {
-         *     private HashMap<String, List<String>> data = new HashMap<>() {
+         *         implements HierarchicalDataProvider&lt;String, Void&gt; {
+         *     private HashMap&lt;String, List&lt;String&gt;&gt; data = new HashMap&lt;>() {
          *         {
          *             put(null, List.of("Item 0", "Item 1"));
          *             put("Item 0", List.of("Item 0-0"));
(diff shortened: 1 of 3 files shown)


flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataProvider.java
```diff
@@ -38,16 +40,210 @@
...
+         * to potentially a lot of consecutive requests and heavy memory usage.
+         * As a result, the scroll position cannot be restored automatically
+         * after using {@link HierarchicalDataProvider#refreshAll()} which
+         * resets the cached hierarchy state.
+         * <li>The scroll container size updates dynamically while the user
+         * scrolls, which can cause them to jump over and miss some levels when
+         * scrolling quickly.
+         * </ul>
+         */
+        NESTED,
```
vursen, line 115: The alternative could be `RECURSIVE`

tltv: `RECURSIVE` describes the hierarchy format a bit better since data providers are supposed to fetch tree branch recursively.
And just `FLAT` for the other one.

tltv: I changed my mind about this change after realizing it's too much focused on edge case of fetching the whole tree at once.

tltv (dismissed):

PR #22044, opened 2025-08-15
fix: Support callpod user agent format (#21828)(CP: 23.6)

mshabarov:


caalador:
It would seem that some other changes to the user agent parser have not been cherrypicked to 23.6

mcollovati:
Fixed unit tests.
it-test-9 failures are fixed by #22203 
it-test-5 are the Spring Security flaky tests (need multiple re-runs)

caalador (approved):

PR #22052, opened 2025-08-18
fix: Theme classes that are not set as theme do not add css

### LINKED ISSUE #21605: Lumo class should be deprecated and just load stylesheets in V25
### Describe your motivation

Currently, applying a theme (in practice Lumo or Material) [rewrites components' JSModule paths to the theme-specific module](https://github.com/vaadin/flow/blob/17e89db79670c12c0a8c9f7ed433985641d51be2/flow-server/src/main/java/com/vaadin/flow/theme/AbstractTheme.java#L95).

[As of V25](https://github.com/vaadin/platform/issues/7453), Material is removed, and the Lumo styles for components are no longer loaded as part of the component dependencies but as regular stylesheets loaded separately from the components themselves: `@CssImport("@vaadin/vaadin-lumo-styles/lumo.css")` or, in css, `@import("@vaadin/vaadin-lumo-styles/lumo.css")`.

The `@Theme` annotation is deprecated but retained for backward compatibility, but only as a shorthand for loading the Lumo stylesheet, and, if specified, the application theme. The `Lumo` class itself should be similarly deprecated but retained, as it only serves a legacy purpose.

- `@Theme(themeClass=Lumo.class)` should just do a `@CssImport("@vaadin/vaadin-lumo-styles/lumo.css")`
- `@Theme("mytheme")` should do the import and then load the specified folder's `styles.css`
- `@Theme(themeClass=Lumo.class, variant=Lumo.DARK)` should do the import and apply the `theme="dark"` to the page (as it does today).

Perhaps this could be implemented simply by a `@CssImport` in the Lumo theme class, instead of [the current `lumo-styles` npm package import](https://github.com/vaadin/flow-components/blob/2e680d7325f3e9e183d42a08a60834151ea8919c/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java#L32)?

Additionally, the Lumo theme class should no longer [load various JSModules](https://github.com/vaadin/flow-components/blob/2e680d7325f3e9e183d42a08a60834151ea8919c/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java#L33-L38) as these will no longer exist.

Finally, the Lumo class should be deprecated in V25.

### LINKED ISSUE #22014: Lumo theme is loaded in dev mode although no @Theme annotation is present (v25)
### Description of the bug

Just having `vaadin-lumo-theme-flow` as a dependency, but with no `@Theme` annotation in the project, loads the Lumo theme in dev mode but not in production mode.

As such when doing an upgrade to v25 you might just run dev mode and see that the theme is still working fine, however when you make a production deployment it turns out that it actually doesn't work.

### Expected behavior

My understanding is that in v25 the Lumo theme should only be applied when there is a `@Theme` annotation present in the application. I'm not 100% sure though, as I don't know what has been discussed previously on how Flow should handle themes in v25.

Can be reproduced with the IT modules in flow-components.

- Clone `git@github.com:vaadin/flow-components.git`
- `cd vaadin-button-flow-parent/vaadin-button-flow-integration-tests/`

Dev mode:
- `mvn jetty:run`
- Open http://localhost:8080/vaadin-button
- Observe that Lumo theme is loaded, although the app has no `@Theme` annotation

Prod mode:
- Update the `run-jetty` profile in `vaadin-button-flow-parent/vaadin-button-flow-integration-tests/pom.xml` like so:
```xml
        <profile>
            <id>run-jetty</id>
            <activation>
                <property>
                    <name>!skipJetty</name>
                </property>
            </activation>
            <build>
                <plugins>
                    <plugin>
                        <groupId>com.vaadin</groupId>
                        <artifactId>flow-maven-plugin</artifactId>
                        <executions>
                            <execution>
                                <goals>
                                    <goal>build-frontend</goal>
                                </goals>
                            </execution>
                        </executions>
                    </plugin>
                    <plugin>
                        <groupId>org.eclipse.jetty.ee10</groupId>
                        <artifactId>jetty-ee10-maven-plugin</artifactId>
                        <configuration>
                            <systemProperties>
                                <vaadin.productionMode>true</vaadin.productionMode>
                            </systemProperties>
                        </configuration>
                    </plugin>
                </plugins>
            </build>
        </profile>
```
- `mvn jetty:run`
- Observe that Lumo theme is not loaded

### Versions

- Vaadin / Flow version: latest v25 snapshot

caalador:
If `themeClass` defines CssImport
the import is not
taken into account if
the themeClass is not active.

Fixes #21605
Fixes #22014

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -2,7 +2,7 @@
 name: Feature request
 about: Request for a new feature to Flow
 title: ''
-labels: enhancement
+labels: ''
 assignees: ''
 
 ---
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java
@@ -3039,13 +3039,14 @@ public class Binder<BEAN> implements Serializable {
      * status.
      * <p>
      * If all field level validators pass, and {@link #setBean(Object)} has been
-     * used to bind to a bean, bean level validators are run for that bean. Bean
-     * level validators are ignored if there is no bound bean or if any field
-     * level validator fails.
+     * used to bind to a bean, bean level validators are run for that bean.
+     * <p>
+     * <strong>Note:</strong>Bean level validators are ignored if there is no
+     * bound bean or if any field level validator fails.
      * <p>
      * <strong>Note:</strong> This method will attempt to temporarily apply all
-     * current changes to the bean and run full bean validation for it. The
-     * changes are reverted after bean validation.
+     * current changes to the set bean and run full bean validation for it. The
+     * changes are reverted after bean validation is run.
      *
      * @return validation status for the binder
      */
@@ -3065,16 +3066,11 @@ public class Binder<BEAN> implements Serializable {
      *
      */
     protected BinderValidationStatus<BEAN> validate(boolean fireEvent) {
-        if (getBean() == null && !validators.isEmpty()) {
-            throw new IllegalStateException("Cannot validate binder: "
-                    + "bean level validators have been configured "
-                    + "but no bean is currently set");
-        }
         List<BindingValidationStatus<?>> bindingStatuses = validateBindings();
 
         BinderValidationStatus<BEAN> validationStatus;
-        if (validators.isEmpty() || bindingStatuses.stream()
-                .anyMatch(BindingValidationStatus::isError)) {
+        if (validators.isEmpty() || getBean() == null || bindingStatuses
+                .stream().anyMatch(BindingValidationStatus::isError)) {
             validationStatus = new BinderValidationStatus<>(this,
                     bindingStatuses, Collections.emptyList());
         } else {
@@ -3104,9 +3100,12 @@ public class Binder<BEAN> implements Serializable {
      * unlike {@link #validate()} and will not modify the UI. To also update
      * error indicators on fields, use {@code validate().isOk()}.
      * <p>
+     * <strong>Note:</strong>Bean level validators are ignored if there is no
+     * bound bean or if any field level validator fails.
+     * <p>
      * <strong>Note:</strong> This method will attempt to temporarily apply all
-     * current changes to the bean and run full bean validation for it. The
-     * changes are reverted after bean validation.
+     * current changes to the set bean and run full bean validation for it. The
+     * changes are reverted after bean validation is run.
      *
      * @see #validate()
      *
--- a/flow-data/src/main/java/com/vaadin/flow/data/binder/Validator.java
+++ b/flow-data/src/main/java/com/vaadin/flow/data/binder/Validator.java
@@ -79,6 +79,18 @@ public interface Validator<T>
     }
 
     /**
+     * Returns a validator that fails on any value.
+     *
+     * @param <T>
+     *            the value type
+     * @return an always-failing validator
+     */
+    static <T> Validator<T> alwaysFail(String errorMessage) {
+        Objects.requireNonNull(errorMessage);
+        return (value, context) -> ValidationResult.error(errorMessage);
+    }
+
+    /**
      * Builds a validator out of a conditional function and an error message. If
      * the function returns true, the validator returns {@code Result.ok()}; if
      * it returns false or throws an exception,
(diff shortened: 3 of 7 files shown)


flow-server/src/test/java/com/vaadin/flow/server/frontend/scanner/FrontendDependenciesTest.java
```diff
@@ -159,6 +160,42 @@ public void appThemeDefined_getsLumoAsTheme() {
...
+            cssFound = cssFound || dependencies.getCss().get(key).stream()
+                    .anyMatch(css -> css.getValue()
+                            .equals("@vaadin/vaadin-lumo-styles/lumo.css"));
+        }
+
+        Assert.assertTrue(cssFound);
+    }
+
+    @Test
+    public void themNotDefined_ButReferenced_themeCssNotLoaded() {
```
tltv:
```suggestion
    public void themeNotDefined_ButReferenced_themeCssNotLoaded() {
```

flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractNodeUpdateImportsTest.java
```diff
@@ -115,35 +116,38 @@ public void generateImportsFile_fileContainsThemeLinesAndExpectedImportsAndCssIm
...
+        expectedLines.add(
+                "addCssBlock\\(`<style include=\"bar\">\\$\\{\\$css_\\d\\}</style>`\\);");
+        expectedLines.add(
+                "registerStyles\\('', \\$css_\\d, \\{moduleId: 'baz'\\}\\);");
+        expectedLines.add(
+                "registerStyles\\('', \\$css_\\d, \\{include: 'bar', moduleId: 'baz'\\}\\);");
+        expectedLines.add(
+                "registerStyles\\('foo-bar', \\$css_\\d, \\{moduleId: 'flow_css_mod_\\d'\\}\\);");
+        expectedLines.add(
+                "registerStyles\\('foo-bar', \\$css_\\d, \\{include: 'bar', moduleId: 'flow_css_mod_\\d'\\}\\);");
```
tltv, line 150: No way to understand what's going on here anymore.. it wasn't much better before either so refactoring may happen some other time.

flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractUpdateImportsTest.java
```diff
@@ -322,32 +323,35 @@ public void generateLines_resultingLinesContainsThemeLinesAndExpectedImportsAndC
...
+                "registerStyles\\('', \\$css_\\d, \\{moduleId: 'baz'\\}\\);");
         expectedLines.add(
-                "registerStyles('', $css_4, {include: 'bar', moduleId: 'baz'});");
+                "registerStyles\\('', \\$css_\\d, \\{include: 'bar', moduleId: 'baz'\\}\\);");
         expectedLines.add(
-                "registerStyles('foo-bar', $css_5, {moduleId: 'flow_css_mod_5'});");
+                "registerStyles\\('foo-bar', \\$css_\\d, \\{moduleId: 'flow_css_mod_\\d'\\}\\);");
         expectedLines.add(
-                "registerStyles('foo-bar', $css_6, {include: 'bar', moduleId: 'flow_css_mod_6'});");
+                "registerStyles\\('foo-bar', \\$css_\\d, \\{include: 'bar', moduleId: 'flow_css_mod_\\d'\\}\\);");
```
tltv, line 354: Same thing here.

flow-server/src/test/java/com/vaadin/flow/server/frontend/AbstractNodeUpdateImportsTest.java
caalador: It checks the generated imports to see that expected lines are available, but the Theme that is added to the tests gets injected in the beginning for full scanner and end for byte scanner so we can not for both modes check that $cssFromFile_0 import is vaadin-mixed-component as depending on the scanner used for the test it might be the new lumo-css-import.css from the theme class. 
So the only change is basically that we accept the number to be any number. So they are now regex checked lines.

caalador:
The rebuild perhaps needs to be checked if it adds the theme, as we accept having too many things in the bundle, but rebuild for missing things.

PR #22074, opened 2025-08-21
feat: Make signal environment init optional

### LINKED ISSUE #22033: Allow using signals without initializing the signal environment
### Describe your motivation

Multiple problems have been discovered related to the way the signal environment is initialized:
* https://github.com/vaadin/flow/issues/21800
* https://github.com/vaadin/flow/issues/22009

We need two changes to address these problems:
* `SignalEnvironment` creates its own `ObjectMapper` rather than sharing a mapper with `VaadinService`. There might be some case where the application needs to customize this mapper but we can add support for that when we better understand the actual needs.
* `SignalEnvironment` has a default dispatcher that runs the task synchronously. `VaadinService` registers a supplier (similar to, but probably distinct from, the current `dispatcherOverrides`) to use the service dispatcher for any tasks that are triggered when `VaadinService.getCurrent()` is defined. This means that signal value changes that are applied when no service is available will block until all affected effects have run rather than returning eagerly. Applications can avoid this by registering their own supplier that is less selective about when it's used (but only if there are affected effects to notify to avoid logging for signals created but not further used in service beans). We should make applications aware of this by logging a warning the first time the default synchronous dispatcher is used. Once we require Java 21+, we can maybe dispatch on a virtual thread rather than running synchronously and it would then not be necessary to log any warnings.

Legioth:
Also removes the duplicate ___staticMocks variants in VaadinServiceSignalsInitializationTest

Fixes #22033


signals/src/main/java/com/vaadin/signals/SignalEnvironment.java
```diff
@@ -19,164 +19,138 @@
...
+    private static final List<SignalEnvironment> environments = new CopyOnWriteArrayList<>();
 
-    private static final List<Supplier<Executor>> dispatcherOverrides = new CopyOnWriteArrayList<>();
+    private static final Executor IMMEDIATE_EXECUTOR = Runnable::run;
 
-    private SignalEnvironment() {
-        // Only static stuff
-    }
+    private static final Executor DEFAULT_EFFECT_DISPATCHER = task -> {
+        resolve(SignalEnvironment::getFallbackEffectDispatcher,
```
mshabarov, line 38: IMHO `EFFECT_DISPATCHER_FALLBACK` would be a better name, as it tries to find the fallback dispatcher.

signals/src/main/java/com/vaadin/signals/SignalEnvironment.java
```diff
@@ -19,164 +19,138 @@
...
-     * @throws IllegalStateException
-     *             if the environment has not yet been initialized
+     * @return the executor to use for invoking affected effects that don't have
+     *         their own dispatcher, or <code>null</code> to invoke the
+     *         callbacks immediately
      */
-    public static ObjectMapper objectMapper() {
-        return getState().objectMapper;
-    }
+    public abstract Executor getFallbackEffectDispatcher();
```
mshabarov, line 101: `getFallbackEffectDispatcher`, `getEffectDispatcher` and `getResultNotifier` could be `protected` as they are implemented by framework or by custom signal environment in projects, but not supposed to be used by the Signals codes or code that just uses Signals.

mshabarov (approved):
Looks good to me. Tested with walking skeleton by:
- adding a `ValueSignal` to `TaskService`
- adding a UI unit test that triggers task create,
No errors, app and test work as expected.

Left two comments that do not block the merge.

mshabarov:
Merging this so to include into beta. We can discuss/fix later.

PR #22104, opened 2025-08-26
fix: Component styles feature flag not working in Gradle projects

### LINKED ISSUE #22073: Feature flag for component styles doesn't work in Gradle projects
### Description of the bug

When I set the feature flag for auto-injecting the component styles from a theme folder `com.vaadin.experimental.themeComponentStyles=true`, I don't see any difference when me project is built with Gradle.

### Expected behavior

Component styles injection should be activated and used same way as with Maven.

Add `/components/vaadin-text-field.css` into https://github.com/vaadin/base-starter-spring-gradle/tree/v24/src/main/frontend/themes/my-theme and run the app, notice that your styles are not applied to the text field.

### Versions

- Vaadin / Flow version: Vaadin 25.0.0-alpha3

Artur-:
The theme component styles feature flag (com.vaadin.experimental.themeComponentStyles) was not working in Gradle projects due to the theme plugin incorrectly parsing the path to find the vaadin-featureflags.properties file.

The fix adds an explicit javaResourceFolder path to the theme options, allowing the theme plugin to directly locate the feature flags file instead of trying to parse it from the projectStaticOutput path, which has different structures in Maven and Gradle projects.

Fixes #22073

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateSettingsFile.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdateSettingsFile.java
@@ -53,6 +53,7 @@ public class TaskUpdateSettingsFile implements FallibleCommand, Serializable {
     String buildDirectory;
     String themeName;
     PwaConfiguration pwaConfiguration;
+    File javaResourceFolder;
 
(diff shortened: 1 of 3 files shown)


caalador:
The other choice would ofc be to just give the actual feature flag value directly from Java code into the settings file.

PR #22105, opened 2025-08-26
chore: Update license-checker version to 2.0.0-beta4 and enable pre-trial for kits

mshabarov:
Allows to use kits (and other products that use this service init listeners) when 7-days trial has been started and not yet expired.


ZheSun88 (approved):

pom.xml
```diff
@@ -246,7 +246,7 @@
             <dependency>
                 <groupId>com.vaadin</groupId>
                 <artifactId>license-checker</artifactId>
-                <version>2.0.0-alpha3</version>
+                <version>2.0.0-beta3</version>
```
mshabarov:
```suggestion
                <version>2.0.0-beta4</version>
```

PR #22137, opened 2025-09-01
feat: React properties vite plugin

abdullahtellioglu:
## Description

The PR introduces a new Vite plugin that identifies properties of each JSX element during the transformation step, which is used by the Copilot Properties Panel in the browser. 

How it works (build-time):
1) For each `.tsx` module (excluding anything under `/generated/`), the plugin:
   - Loads the TypeScript Program using the project's tsconfig.
   - Parses the source file and collects distinct JSX opening elements.
   - For each element, resolves the props type via a TypeScript TypeChecker (Resolver).
   - Serializes the resolved properties into a `ResponseData` payload.

How it works (runtime, via injected code):
2) Adds a global registry under `window.Vaadin.copilot.ReactProperties`:
   - `properties: Record<string, ResponseData>` – in-memory map of tagName → ResponseData.
   - `registerer(tagName: string, value: ResponseData)` – helper to populate `properties`.
3) For non-intrinsic elements (custom React components), the plugin tries to attach
   the payload directly to the component function/object on `__debugProperties`
   (mimicking a "Fiber-adjacent" location) for quick introspection:

      MyComponent.__debugProperties = { error: false, properties: [...] }

   If that attachment isn’t possible (e.g., `MyComponent` isn't resolvable as a value),
   it falls back to `registerer("MyComponent", payload)`.

4) For intrinsic elements (e.g., 'div', 'button'), the plugin always registers to the
   global registry via `registerer("div", payload)`.

 https://github.com/vaadin/copilot-internal/pull/6576 is the inline version of both client and plugin implementation. 

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

### DIFF
--- a/flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
+++ b/flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
@@ -238,7 +238,7 @@ export default function reactComponentPropertiesPlugin(): Plugin {
                 } catch (e: any) {
                     const responseDataStr = responseDataStringBuilder(true, e.message);
                     injectCode += registererStringBuilder(nodeName, responseDataStr);
-                    this.debug(e);
+                    console.debug(e);
                 }
             });
             const magicString = new MagicString(code);


flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
```diff
@@ -0,0 +1,806 @@
...
+            if (!bareId.endsWith('.tsx')) {
+                return;
+            }
+            if (id.indexOf('/generated/') !== -1) {
+                return;
+            }
+            let program: Program | undefined = undefined;
+            try {
+                const parsed = resolveTsConfigAndGetParsed();
+                program = ts.createProgram({
```
Artur-: Not sure how valid this is as it is from AI:

  1. Performance Problem - TypeScript Program Recreation: The plugin creates a new TypeScript program for EVERY .tsx file transformation (line 150-154 in index.ts).
  This is extremely inefficient and will significantly slow down builds, especially in large projects. The TypeScript program should be created once and reused.

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
```diff
@@ -0,0 +1,806 @@
...
+                errorMessage?: string,
+                properties?: PropertyDescriptor[],
+            ): string => {
+                const responseData: ResponseData = { error, errorMessage, properties };
+                return JSON.stringify(responseData);
+            };
+            const registererStringBuilder = (nodeName: string, value: string) => {
+                return `window.Vaadin.copilot.ReactProperties.registerer("${nodeName}", ${value});\n`;
+            };
+            const fiberInjectionStringBuilder = (nodeName: string, responseDataStr: string) => {
```
Artur-, line 217: 2. Potential Runtime Errors with Variable Injection: The fiberInjectionStringBuilder function (lines 192-200) injects code that assumes the component name exists as
  a runtime variable. This will fail for components imported with different names or aliased imports:
  import { Button as MyButton } from './components';
  // Will generate: if(typeof MyButton === 'object'...) but MyButton won't exist

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
```diff
@@ -0,0 +1,806 @@
...
+            }
+            let program: Program | undefined = undefined;
+            try {
+                const parsed = resolveTsConfigAndGetParsed();
+                program = ts.createProgram({
+                    rootNames: parsed.fileNames,
+                    options: parsed.options,
+                });
+            } catch (e) {
+                console.debug('Failed to parse program file:', e);
```
Artur-: Is there a reason that errors should be handled as debug?

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
```diff
@@ -0,0 +1,806 @@
...
+                } catch (e: any) {
+                    const responseDataStr = responseDataStringBuilder(true, e.message);
+                    injectCode += registererStringBuilder(nodeName, responseDataStr);
+                    console.debug(e);
+                }
+            });
+
+            return {
+                code: code + injectCode,
+                map: null,
```
Artur-: If this breaks all source maps, it's a no-go

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
```diff
@@ -0,0 +1,806 @@
...
+export default function reactComponentPropertiesPlugin(): Plugin {
+    return {
+        name: 'vaadin-react-component-properties-plugin',
+        enforce: 'pre',
+        transform(code, id) {
+            const [bareId] = id.split('?');
+            if (!bareId.endsWith('.tsx')) {
+                return;
+            }
+            if (id.indexOf('/generated/') !== -1) {
```
Artur-, line 168: Are we sure that generated files can be ignored? If you put a TS/JS file in an add-on jar, it will be copied into `generated/jar-resources/...`

Artur-: Also, this would match any `/generated/` in the path, not only the generated folder inside frontend?

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
```diff
@@ -0,0 +1,806 @@
...
+                program = ts.createProgram({
+                    rootNames: parsed.fileNames,
+                    options: parsed.options,
+                });
+            } catch (e) {
+                console.debug('Failed to parse program file:', e);
+                return;
+            }
+            if (!program) {
+                return;
```
Artur-, line 185: What case would this be? Is it an error?

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
```diff
@@ -0,0 +1,806 @@
+import type { Plugin } from 'vite';
```
Artur-: These are some things Claude thinks you should do. Probably not everything makes sense https://github.com/vaadin/flow/commit/ee8051a40e81171e69353fcf32c08a10b9c9589c

Artur-: Also https://github.com/vaadin/flow/commit/d3960865ee5031ff332f4515e3a1ab60771b656a but don't get too hung up on these

Artur-:
When running this with the Copilot change, I only see this. Am I holding it wrong?
image

abdullahtellioglu:
> When running this with the Copilot change, I only see this. Am I holding it wrong?

Are you sure intrinsic element types are imported? Here, vite tests do not load them so I mimic them [here](https://github.com/vaadin/flow/pull/22137/files#diff-46f63204deb7fea55f169f6f3948b3a03e3b5ef8ec3e0d672cfef6c180cf304bR4). 

It should not happen in the PR where plugin is inline with a test project https://github.com/vaadin/copilot-internal/pull/6576

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
abdullahtellioglu: This is not valid as my first approach is to create a program once and use that instance later. 

This approach failed because ts.createProgram caches the file content by the id. After HMR, the content is not updated so new components are not registered.

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
abdullahtellioglu: This is also invalid. Added an explicit IT test for this

Artur-:
> Are you sure intrinsic element types are imported? 

I have no idea what that means. I ran the test project in Copilot.

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
abdullahtellioglu: Added console.error. Program cannot be undefined at that point, as it is created above in a try catch block, but TS assumes otherwise

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
```diff
@@ -0,0 +1,865 @@
...
+        .replace(/([A-Z]+)([A-Z][a-z\d]+)/g, '$1 $2') // E.g. HTMLParser → HTML Parser
+        .replace(/^./, (str) => str.toUpperCase());
+}
+
+/**
+ * Utility method that helps the user to know the type of object while developing.
+ * It does byte-wise operations for the given type and returns a string
+ * @param type
+ */
+function describeType(type: ts.Type): string {
```
Artur-, line 835: Is this more or less the same as `typeToString` in `TS.TypeChecker`?

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
```diff
@@ -0,0 +1,865 @@
...
+        return 'STRING';
+    } else if (type === 'number') {
+        return 'FLOAT';
+    } else if (type === 'boolean') {
+        return 'BOOLEAN';
+    }
+    return undefined;
+}
+
+function camelCaseToHumanReadable(input?: string): string | undefined {
```
Artur-, line 819: In Copilot there is some npm package used for this - should we use the same here?

Artur- (approved):

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
abdullahtellioglu: More or less true, this gives more structural classification than TypeChecker.

flow-server/src/main/resources/plugins/react-component-properties-plugin/index.ts
abdullahtellioglu: I am assuming you are referring to camelCase library in Copilot, which converts a string to camelCase. This method converts camel case to a human-readable format.

Artur-: yes, I was thinking that there was another one for the other direction but probably does not matter

Artur-:
Did check the performance impact of this?

abdullahtellioglu:
> Did check the performance impact of this?

I did not see any performance impact so far but it would be better to test it with a page with many components and pages. I will create a snapshot locally and perform testing in the Hilla application to save timestamps of the transform operation

abdullahtellioglu:
> Did check the performance impact of this?

`Ts.createProgram` consumes most of the time in the transform step (~200-500ms). I did a couple of improvements but it never goes under 100ms, nor have the changes affected the performance a lot. 

Tried the solution that Claude suggested, and it is almost the same result. It is not possible to reuse the same ts.program with different files, as it is immutable and given files are cached.

PR #22142, opened 2025-09-02
fix: fix hotswap with static Signal with effects

tltv:
Fixes: #51582

### DIFF
--- a/signals/src/main/java/com/vaadin/signals/impl/Effect.java
+++ b/signals/src/main/java/com/vaadin/signals/impl/Effect.java
@@ -64,7 +64,12 @@ public class Effect {
             return;
         }
 
-        dependencies = UsageTracker.track(action);
+        try {
+            dependencies = UsageTracker.track(action);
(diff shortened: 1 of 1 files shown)


signals/src/main/java/com/vaadin/signals/impl/Effect.java
```diff
@@ -64,7 +64,12 @@ private void revalidate() {
             return;
         }
 
-        dependencies = UsageTracker.track(action);
+        try {
+            dependencies = UsageTracker.track(action);
```
Legioth, line 68: Did you also consider the `UsageTracker.track` use from `ComputedSignal`.

I haven't tested in practice but I suspect it will also be affected though you might need a slightly different test case.

tltv: Didn't test yet it but it is indeed another place to handle.

tltv:
Fixed in https://github.com/vaadin/flow/pull/22148

PR #22148, opened 2025-09-02
Handle exceptions from effect and compute callbacks

### LINKED ISSUE #21582: Signals do not work after Hotswap if an effect is introduced
### Description of the bug

After I introduce a new `Signal.effect`, Signals stop working due to hotswapping. Sometimes exceptions are logged, as in the video, or sometimes nothing happens.

https://github.com/user-attachments/assets/bd9f4088-00ab-4106-8c6c-cde8532f1e42

### Expected behavior

Signal values should continue updating the UI without forcing a full page reload.

You can try it with any view that reacts to Signal. 

Example - simplified- snippet from the video:
```java
@PageTitle("HelloWorldView")
@AnonymousAllowed
@Route(value = "hello-world-view")
public class HelloWorldView extends Main {

    private static final NumberSignal counter = SignalFactory.IN_MEMORY_EXCLUSIVE.number("counter");
    public HelloWorldView() {
        Button btn = new Button("Submit", click -> counter.incrementBy(1.0));
        for(int i=0;i<10;i++){
            TextField textField = new TextField("Your name " + i);
            Signal.effect(() -> {
                textField.setValue(counter.valueAsInt() + "times clicked");
            });
            add(textField);
        }
        Signal.effect(() -> {
           Notification.show("Another signal effect");
        });

        add(btn);

        Signal.effect(() -> {btn.setText(counter.valueAsInt() + " times clicked" );});

    }
}

```

then add `Signal.effect(() -> { Notification.show("Something"); });` at below and then build the source code so class is hotswapped. Then you will see that signal effects do not work until page reload. 

### Versions

Hilla: 24.8.0.beta2
Flow: 24.8.0.beta2
Vaadin: 24.8.0.beta2
Spring Boot: 3.5.0
Spring: 6.2.7
Spring Security: 
Spring Data JPA: 
Copilot: 24.8.0.beta2
Frontend Hotswap: Enabled ⋅ Vite
OS: aarch64 Mac OS X 14.7.1
Java: JetBrains s.r.o. 21.0.5
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36
Java Hotswap: Java Hotswap is enabled
IDE Plugin: 1.0-SNAPSHOT IntelliJ

Legioth:
Effect exceptions are handled by passing them to `Thread::getUncaughtExceptionHandler`. For an error, the effect is disabled so that it will no longer be run and `getLogger()::error` is used to communicate the state change.

Exceptions from computed signals are cached and re-thrown when reading the signal value.

Fixes #21582

### DIFF
--- a/signals/src/main/java/com/vaadin/signals/impl/Effect.java
+++ b/signals/src/main/java/com/vaadin/signals/impl/Effect.java
@@ -58,7 +58,9 @@ public class Effect {
             try {
                 action.run();
             } catch (Exception e) {
-                getLogger().error("Uncaught exception from effect.", e);
+                Thread thread = Thread.currentThread();
+                thread.getUncaughtExceptionHandler().uncaughtException(thread,
+                        e);
             } catch (Error e) {
                 getLogger().error(
                         "Uncaught error from effect. The effect will no longer be active.",


signals/src/main/java/com/vaadin/signals/impl/Effect.java
```diff
@@ -51,7 +54,18 @@ public class Effect {
      */
     public Effect(Runnable action) {
         assert action != null;
-        this.action = action;
+        this.action = () -> {
+            try {
+                action.run();
+            } catch (Exception e) {
+                getLogger().error("Uncaught exception from effect.", e);
```
knoobie: Calling it "graceful" and logging on error-level feels kinda wrong, especially with your simple example with NumberFormatException.. I would instantly mute this class :/

Legioth: Note that this isn't like the "Connection remotely closed" exceptions that just happen. Instead, this is always an exception that has been thrown through application code and not caught before returning from the effect. It's thus more like throwing exceptions from `service()` in a vanilla Servlet or from the action handler of a Swing button. These are the things that developers or testers should notice so that they can e.g. add an input validator or explicitly catch and handle the exception in way that makes sense for the user.

Would it make sense to ensure that effects that are run in a Vaadin context (i.e. through `ComponentEffect`) would have any uncaught exceptions passed to `VaadinSession::getErrorHandler` whereas other effects would use `Thread::getUncaughtExceptionHandler`?

knoobie: >  Instead, this is always an exception that has been thrown through application code and not caught before returning from the effect.

Sound like the perfect case to use the Error Handler and let the user decide how to process it instead of throwing it in the log file

> VaadinSession::getErrorHandler 

to be honest, I would have expected that this is already the case and therefore did not bring it up here :)

Legioth: Any opinion on `Thread::getUncaughtExceptionHandler` vs `getLogger()::error` in the cases when Vaadin's error handler is not applicable?

knoobie: I would vote for `Thread::getUncaughtExceptionHandler` if no error handler is available. I like to go with the following mantra: Exceptions are only logged if they are also handled. If an exception is not handled (i.e. passed on to the caller), it is not logged.

Otherwise you get multiple warn/error/info logs with the exact same exception / information throughout the stack.

Legioth: Updated to use `getUncaughtExceptionHandler` in the default cause.

I created https://github.com/vaadin/flow/issues/22157 for the `ComponentEffect` case since I've got another change in progress for that part of the code and I want to avoid causing merge conflicts for myself.

signals/src/main/java/com/vaadin/signals/impl/Effect.java
```diff
@@ -51,7 +54,20 @@ public class Effect {
...
     public Effect(Runnable action) {
         assert action != null;
-        this.action = action;
+        this.action = () -> {
+            try {
+                action.run();
+            } catch (Exception e) {
+                Thread thread = Thread.currentThread();
+                thread.getUncaughtExceptionHandler().uncaughtException(thread,
+                        e);
```
mshabarov, line 63: I don't get what this brings up: so you delegate the `e` handling to standard handler and mark the exception as uncaught, nothing else in this catch block. Then `Error` will dispose the effect. Is it so that you are trying to discriminate between `Exception` (that can easily be thrown from the effect due to project's logic) and `Error` that happens basically because of crash or serious issues? If yes, how it's related to hotswap/livereload issues?

Legioth: `Exception` is a "regular" mistake, e.g. a `NullPointerException` or such. After that happens, it makes sense to keep the effect alive since it's quite likely that it will work as expected when run with some other input. Delegating to the exception handler is just a way of giving the application developer control over how the stacktrace is logged - it's not expected that the handler does anything else.

`Error` is a systematic failure, e.g. an incompatibility caused by hotswap, and it's very unlikely that it will work any different the next time the signal value changes. We could delegate logging to the uncaught exception handler here as well but that would make it more complicated to provide the additional message that tells that the effect will no longer be active.

PR #22162, opened 2025-09-04
Use UI::access only for ComponentEffect

### LINKED ISSUE #22153: Don't use UI::access automatically for Signal.effect
### Describe your motivation

With the current implementation, invocations of `Signal.effect` capture `UI.getCurrent()` if present and then uses  `UI::access` to invoke the effect callback unless the change happens on a thread that already has the UI locked. While effects related to the UI should typically be created through `ComponentEffect`, the idea has been that effects not related to any specific component instance should not require providing any instance. One example of this is `Signal.effect(() -> Notification.show(statusUpdateSignal.value()));`.

There are two big problems with the current implementation
1. This is not deterministic with effects that are created in e.g. a static initializer since that particular class might be initialized either with or without `UI.getCurrent()` defined depending on subtle timing differences.
2. An effect not tied to any component instance will outlive the UI instance that it has captured. This is problematic for effects not related to a UI, e.g. something using `System.out.println`, since the effect will suddenly stop working when the associated UI is closed.

Effects created using `Signal.effect` should not use `UI::access`. Furthermore, we should make sure that such effects are consistently run without Vaadin's thread locals defined so that there's no risk of inconsistent behavior depending on how an effect update is triggered.

Effects created using `ComponentEffect` should use the UI that the owner component is attached to rather than relying on `UI.getCurrent()`. Such effects should still retain the functionality that they use `UI:access` only if the change isn't triggered from the same UI. This probably means that we have to introduce some mechanism to preserve `UI.getCurrent()` for this purpose even though it's explicitly cleared for general `Signal.effect` use.

A similar automatic `UI.getCurrent()` capture mechanism is also used for resolving the `SignalOperation::result` future. That behavior is still appropriate and should be retained.

### LINKED ISSUE #22157: Make sure VaadinSession::getErrorHandler is used for ComponentEffect errors
We should give the application developer control over handling of uncaught exceptions from component effects by passing them to the regular error handler.

I expect this to work automatically for the code paths that go through `UI:access` but we need to handle it explicitly for the code paths that run the effect immediately.

Legioth:
Remove the distinction between effect dispatchers and fallback effect dispatchers since only the "fallback" variant is relevant now.

The initial effect invocation now goes through the dispatcher so that `UI.getCurrent()` is never defined when using `Signal.effect`.

Fixes #22153
Fixes #22157

### DIFF
--- a/signals/src/main/java/com/vaadin/signals/SignalEnvironment.java
+++ b/signals/src/main/java/com/vaadin/signals/SignalEnvironment.java
@@ -34,10 +34,9 @@ public abstract class SignalEnvironment {
 
     private static final Executor IMMEDIATE_EXECUTOR = Runnable::run;
 
-    private static final Executor EFFECT_DISPATCHER_FALLBACK = task -> {
-        resolve(SignalEnvironment::getFallbackEffectDispatcher,
-                IMMEDIATE_EXECUTOR).execute(task);
-    };
+    private static final Executor EFFECT_DISPATHCER = command -> resolve(
+            SignalEnvironment::getEffectDispatcher, IMMEDIATE_EXECUTOR)
+            .execute(command);
 
     /**
      * Checks whether this environment is active on the current thread. No other
@@ -65,40 +64,17 @@ public abstract class SignalEnvironment {
     protected abstract Executor getResultNotifier();
 
     /**
-     * Gets an executor to use for running the callback of an effect. This
-     * method is run when an effect is created and the returned executor is used
-     * for running the effect callback after any change has been detected. The
-     * executor can thus be used to make sure the effect callback is invoked in
-     * the same context as where the effect was created. It is recommended that
-     * the executor is asynchronous so that the thread that submitted the change
-     * can proceed without waiting for all affected effects to be dispatched. If
-     * no dispatcher is provided when an effect is created, then the effect
-     * callback will be run according to the
-     * {@link #getFallbackEffectDispatcher()} of the environment that is active
-     * when a change is applied. This executor does not need to preserve
-     * ordering since the effect callback always uses the latest signal values
-     * without concern for in which order values have been changed.
+     * Gets an executor to use for running the callback of an effect. The
+     * executor can be used to make effect callback invocations asynchronous
+     * rather than blocking the thread that applied the change until all
+     * affected effects have been processed. This executor does not need to
+     * preserve ordering since the effect callback always uses the latest signal
+     * values without concern for in which order values have been changed.
      *
-     * @return an executor to use for invoking effect callbacks, or
-     *         <code>null</code> to use the fallback dispatcher
+     * @return the executor to use for invoking affected effects, or
+     *         <code>null</code> to invoke the callbacks immediately
      */
     protected abstract Executor getEffectDispatcher();
-
-    /**
-     * Gets an executor to use for running the callback of an effect that
-     * doesn't have its own dispatcher. This method is run when applying any
-     * change to an effect that has no own dispatcher. The executor can thus be
-     * used to make effect callback invocations asynchronous rather than
-     * blocking the thread that applied the change until all affected effects
-     * have been processed. This executor does not need to preserve ordering
-     * since the effect callback always uses the latest signal values without
-     * concern for in which order values have been changed.
-     *
-     * @return the executor to use for invoking affected effects that don't have
-     *         their own dispatcher, or <code>null</code> to invoke the
-     *         callbacks immediately
-     */
-    protected abstract Executor getFallbackEffectDispatcher();
 
     /**
      * Registers a signal environment to consider when processing signal
@@ -138,19 +114,15 @@ public abstract class SignalEnvironment {
     }
 
     /**
-     * Queries currently active environments for an executor to use for running
-     * the callbacks of an effect that is currently being created. If no
-     * registered environment provides an executor, then this method returns an
-     * executor that will delegate to the environment that is active when a
-     * change is applied and otherwise run the callback immediately.
+     * Gets an executor that queries currently active environments for an
+     * executor to use for evaluating changes for an effect. An immediate
+     * executor is used if no executor is provided by registered environments.
      *
      * @see #getEffectDispatcher()
-     * @see #getFallbackEffectDispatcher()
      *
      * @return the executor to use, not <code>null</code>
      */
-    public static Executor getCurrentEffectDispatcher() {
-        return resolve(SignalEnvironment::getEffectDispatcher,
-                EFFECT_DISPATCHER_FALLBACK);
+    public static Executor getDefaultEffectDispatcher() {
+        return EFFECT_DISPATHCER;
     }
 }
--- a/flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
@@ -15,17 +15,17 @@
  */
 package com.vaadin.flow.component;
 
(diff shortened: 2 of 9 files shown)


signals/src/main/java/com/vaadin/signals/SignalEnvironment.java
```diff
@@ -34,10 +34,9 @@ public abstract class SignalEnvironment {
 
     private static final Executor IMMEDIATE_EXECUTOR = Runnable::run;
 
-    private static final Executor EFFECT_DISPATCHER_FALLBACK = task -> {
-        resolve(SignalEnvironment::getFallbackEffectDispatcher,
-                IMMEDIATE_EXECUTOR).execute(task);
-    };
+    private static final Executor EFFECT_DISPATHCER = command -> resolve(
```
tltv: typo. rename to `EFFECT_DISPATCHER`.

tltv (changes requested):
Effects created in view constructor like following stops being invoked at all or they run sometimes without any pattern. Here the signal is changed on button click, but effect added in view class constructor is not run except initially:
```java
private final NumberSignal counter = SignalFactory.IN_MEMORY_EXCLUSIVE.number("counter");

public MyView {
Button btn = new Button("Submit", click -> counter.incrementBy(1.0));
...
Signal.effect(() -> {
            System.out.println("foo"); // run only on page load
        });
```

Legioth:
There are some side effects from another signal-related PR that was recently merged and there's another adjustment coming on top of that one so I'll put this change on hold until the dust has settled there.

signals/src/main/java/com/vaadin/signals/SignalEnvironment.java
Legioth: Fixed

Legioth:
> `System.out.println("foo");`

This effect doesn't depend on any signal value and will thus be run only once.

If I changed it to `System.out.println("foo " + counter.valueAsInt());`, then it does run again every time the button is clicked.

tltv:
This may not be directly related to changes in this ticket since it's reproducible with hotswap only. I see issue with hotswap and Signal.effect + ComponentEffect.effect when signal is stored in a static field.
E.g. consider following case:

```java
private static final NumberSignal counter = SignalFactory.IN_MEMORY_EXCLUSIVE.number("counter");
public MyView() {
...
Button btn = new Button("Submit", click -> counter.incrementBy(1.0));
...
Signal.effect(() -> {
            System.out.println("foo" + counter.value());
        });
ComponentEffect.effect(this, () -> {
            Notification.show("Signal effect " + counter.value());
        });
```

This works fine initially. But if you 
1. remove first `Signal.effect` and let hotswap finish changing classes, click the button
  => see how UI works as expected but there's an exception from the `ComponentEffect` due to missing UI instance.
2. add the first `Signal.effect` back and wait for hotswap again, click the button
  => sysout is run twice with same value
3. repeat first two steps and see how each iteration just adds another sysout and exception polluting logs

Looks like already disposed `ComponentEffect` effects are still run somehow. Maybe through the usage tracking.

And looks like Signal.effect is never disposed. And since in this case it's not failing, old effects keeps piling up and running successfully as nothing disposes them. Maybe signals need optional cleanup API to cover this kind of case with a static field.

However, when I change both to `ComponentEffect.effect`, there's no issues. So that's also one way to go. I just think that usage of `Signal.effect` API is more natural choice if it's not touching any flow component.

Legioth:
> but there's an exception from the `ComponentEffect` due to missing UI instance

Cannot reproduce. This is supposed to be handled by `catch (UIDetachedException e)`  in `ComponentEffect`. Do you have a stacktrace that shows what path in takes instead?

Also, make sure that you run with locally built snapshots of both `flow-server` and `signals` since changes from both are required for correct behavior.

> And looks like Signal.effect is never disposed

Regular signals require manual cleanup through the `Runnable` returned from `Signal.effect`. It's intentional to not do anything "magic" with them. You should use `ComponentEffect` for cleanup based on component instances, or clean up manually in a detach listener, or move the effect creation to a `static` init block so that it's run only once.

tltv:
After calling Signal.effect return callback on view detach both issues were resolved.

Exception I got looked like this:
```
Exception in thread "task-6" java.lang.IllegalStateException: UI instance is not available. It means that you are calling this method out of a normal workflow where it's always implicitly set. That may happen if you call the method from the custom thread without 'UI::access' or from tests without proper initialization.
	at com.vaadin.flow.component.shared.internal.OverlayAutoAddController.getUI(OverlayAutoAddController.java:144)
	at com.vaadin.flow.component.shared.internal.OverlayAutoAddController.handleOpen(OverlayAutoAddController.java:111)
	at com.vaadin.flow.component.shared.internal.OverlayAutoAddController.lambda$new$cb78cf17$1(OverlayAutoAddController.java:55)
	at com.vaadin.flow.internal.nodefeature.ElementPropertyMap.lambda$fireEvent$2(ElementPropertyMap.java:465)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at com.vaadin.flow.internal.nodefeature.ElementPropertyMap.fireEvent(ElementPropertyMap.java:465)
	at com.vaadin.flow.internal.nodefeature.ElementPropertyMap$PutResult.run(ElementPropertyMap.java:170)
	at com.vaadin.flow.internal.nodefeature.ElementPropertyMap.put(ElementPropertyMap.java:152)
	at com.vaadin.flow.internal.nodefeature.AbstractPropertyMap.setProperty(AbstractPropertyMap.java:63)
	at com.vaadin.flow.internal.nodefeature.ElementPropertyMap.setProperty(ElementPropertyMap.java:99)
	at com.vaadin.flow.dom.impl.BasicElementStateProvider.setProperty(BasicElementStateProvider.java:273)
	at com.vaadin.flow.dom.Element.setRawProperty(Element.java:889)
	at com.vaadin.flow.dom.Element.setProperty(Element.java:668)
	at com.vaadin.flow.component.notification.Notification.setOpened(Notification.java:468)
	at com.vaadin.flow.component.notification.Notification.open(Notification.java:392)
	at com.vaadin.flow.component.notification.Notification.show(Notification.java:281)
	at com.vaadin.flow.component.notification.Notification.show(Notification.java:309)
	at com.vaadin.flow.component.notification.Notification.show(Notification.java:332)
	at io.blitzen.magic.taskmanagement.ui.view.HelloWorldView.lambda$new$0(HelloWorldView.java:51)
	at com.vaadin.signals.impl.Effect.lambda$new$0(Effect.java:75)
	at com.vaadin.signals.impl.Transaction.lambda$asSupplier$0(Transaction.java:276)
	at com.vaadin.signals.impl.Transaction.runInTransaction(Transaction.java:255)
	at com.vaadin.signals.impl.Transaction.runInTransaction(Transaction.java:297)
	at com.vaadin.signals.impl.UsageTracker.track(UsageTracker.java:157)
	at com.vaadin.signals.impl.Effect.revalidate(Effect.java:104)
	at com.vaadin.signals.impl.Effect.invalidate(Effect.java:122)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

Edit: btw, I get this also if I click the button quickly before hotswap is fully done. But that's expected.

Legioth:
That stacktrace looks like using `Notification.show` from inside a regular `Signal.effect` rather than through a `ComponentEffect`. This is explicitly supposed to not work because `UI.getCurrent()` is explicitly defined to not be available inside `Signal.effect`.

PR #22172, opened 2025-09-05
fix: make access denied a warning

caalador:
Make the access denied issue
when paren layout access is
denied a warning log instead
of a debug log.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/auth/AnnotatedViewAccessChecker.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/auth/AnnotatedViewAccessChecker.java
@@ -107,7 +107,7 @@ public class AnnotatedViewAccessChecker implements NavigationAccessChecker {
                     boolean hasAccess = accessAnnotationChecker.hasAccess(
                             parent, context.getPrincipal(), context::hasRole);
                     if (!hasAccess) {
-                        LOGGER.debug(
+                        LOGGER.warn(
                                 "Denied access to view due to parent layout '{}' access rules",
                                 parent.getSimpleName());
                         return context.deny(


flow-server/src/main/java/com/vaadin/flow/server/auth/AnnotatedViewAccessChecker.java
```diff
@@ -107,7 +107,7 @@ public AccessCheckResult check(NavigationContext context) {
                     boolean hasAccess = accessAnnotationChecker.hasAccess(
                             parent, context.getPrincipal(), context::hasRole);
                     if (!hasAccess) {
-                        LOGGER.debug(
+                        LOGGER.warn(
```
vursen, line 111: There is a similar message in the first branch. Should it be converted to a warning as well?

flow-server/src/main/java/com/vaadin/flow/server/auth/AnnotatedViewAccessChecker.java
```diff
@@ -85,7 +85,7 @@ public AccessCheckResult check(NavigationContext context) {
                 boolean hasAccess = accessAnnotationChecker.hasAccess(layout,
                         context.getPrincipal(), context::hasRole);
                 if (!hasAccess) {
-                    LOGGER.debug(
+                    LOGGER.warn(
                             "Denied access to view due to layout '{}' access rules",
                             layout.getSimpleName());
                     return context.deny(
```
vursen: Shouldn't actually these messages be the other way around: "consider adding..." as the log message and "denied access" as the reason argument. With the current approach, I got "consider adding" in the exception reason which was surprising.

caalador: As far as I can see the deny reason should be visible on the error page in devMode. Production mode returns always "" as the reason.

vursen: Yes, but the deny reason is currently "consider adding..." rather than "denied access due to..." which seems illogical.

caalador: Added the reason for deny to the start of the deny reasons

vursen: Did you push it?

vursen: How about we show the same message as the log warning and the deny reason?

caalador: Did push... To the wrong branch =_=

PR #22176, opened 2025-09-05
feat!: Jackson for last elemental use

caalador:
Use Jackson in all places
except flow-server/shared
and flow-client

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/StringToEnumDecoder.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/StringToEnumDecoder.java
@@ -18,10 +18,6 @@ package com.vaadin.flow.server.communication.rpc;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeType;
 
-import elemental.json.JsonString;
-import elemental.json.JsonType;
-import elemental.json.JsonValue;
-
 /**
  * Decodes a {@link JsonValue} with {@link JsonType#STRING} type to {@link Enum}
  * subclass type.
--- a/flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
+++ b/flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
@@ -18,6 +18,7 @@ package com.vaadin.flow.component.page;
 
 import java.io.Serializable;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
@@ -25,11 +26,9 @@ import org.mockito.Mockito;
 
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.function.DeploymentConfiguration;
+import com.vaadin.flow.internal.JacksonUtils;
 import com.vaadin.flow.server.VaadinService;
 import com.vaadin.flow.server.VaadinSession;
-
-import elemental.json.Json;
-import elemental.json.JsonString;
 
 public class HistoryTest {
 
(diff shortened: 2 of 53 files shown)


flow-server/src/main/java/com/vaadin/flow/server/communication/rpc/StringToEnumDecoder.java
```diff
@@ -18,16 +18,13 @@
...
- * Decodes a {@link JsonValue} with {@link JsonType#STRING} type to {@link Enum}
- * subclass type.
+ * Decodes a {@link JsonNode} with {@link JsonNodeType#STRING} type to
+ * {@link Enum} subclass type.
  * <p>
- * This decoder is applicable to any {@link JsonValue} which is
- * {@link JsonString} and any {@link Enum} sublcass
+ * This decoder is applicable to any {@link JsonNode} which is
+ * {@link com.fasterxml.jackson.databind.node.TextNode} and any {@link Enum}
+ * sublcass
```
tltv:
```suggestion
 * subclass
```

flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
```diff
@@ -92,21 +91,23 @@ public void setup() {
 
     @Test
     public void pushState_locationWithQueryParameters_queryParametersRetained() {
-        history.pushState(Json.create("{foo:bar;}"), "context/view?param=4");
+        history.pushState(JacksonUtils.readTree("{\"foo\":\"bar\"}"),
```
tltv: Do you know if there's a way to relax Flow's Jackson ObjectMapper read/write to allow original format with unquoted name and single quoted value?

I found that reader is easy to configure by setting couple flags: JsonReadFeature.ALLOW_UNQUOTED_FIELD_NAMES and JsonReadFeature.ALLOW_SINGLE_QUOTES. 
But no idea about writing.

caalador: I think we should just go with the stricter Json style as next in Jackson 3 even with lifted restrictions:
```
.disable(JsonWriteFeature.QUOTE_PROPERTY_NAMES)
.enable(JsonReadFeature.ALLOW_UNQUOTED_PROPERTY_NAMES)
.enable(JsonReadFeature.ALLOW_UNESCAPED_CONTROL_CHARS)
```
The test will throw for unquoted BAR.
`tools.jackson.core.exc.StreamReadException: Unrecognized token 'bar': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')`

Allowing single quotes might be a good addition.

tltv: I was thinking just in this test context. Having it strict by default is fine. It would be nice if old pattern worked as is without adding `\"` all over and didn't see any way to do it easily just for tests.

caalador: I guess the only way would be to first parse to elemental and from elemental to jackson. But the output would still be with `\"` from jackson

flow-server/src/test/java/com/vaadin/flow/component/webcomponent/WebComponentTest.java
```diff
@@ -174,8 +168,7 @@ public void setProperty_attemptsToWriteSupportedTypes() {
         webComponent.setProperty(jsonNodeConfiguration,
                 (ValueNode) JacksonUtils.createNode(true));
         // JsonValue has a different number of arguments
```
tltv:
```suggestion
```

flow-tests/vaadin-spring-tests/pom.xml
```diff
@@ -16,7 +16,8 @@
 
     <properties>
         <maven.deploy.skip>true</maven.deploy.skip>
-        <component.version>25.0.0-alpha8</component.version>
+        <component.version>25.0-SNAPSHOT</component.version>
+        <nimbus-jose-jwt.version>10.4.2</nimbus-jose-jwt.version>
```
tltv: I wonder is it now nimbus-jose-jwt.version 10.4.2 or 10.5? Remove extra one.

caalador: 10.5 is in main branch. This duble version is from some merge from master that has gone wrong.

flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
tltv: That's one way. In tests its easier to change ObjectMapper with reflection to make the input work. But not so sure how to do the same for the output.

caalador: I don't see why we should change the used ObjectMapper in tests instead of using the actual ObjectMapper used during runtime. 
Also the input is a serializable JsonNode so its up to the user to give it in the correct format. We could change the test to not read from a string, but add the key+value into a Object node through a new method, but that wouldn't really make the test better.

tltv: To minimize changes in the tests for starters. It could be relaxed also in JacksonUtils by default, like you already did by allowing single quoutes. I'm not fan of requiring even bigger changes than requiring `\"`.

tltv (approved):

flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
caalador: Which starters have tests that handle json directly?

tltv:
I would suggest to wait for platform 25.0.0-alpha9 release before merging.

flow-server/src/test/java/com/vaadin/flow/component/page/HistoryTest.java
tltv: Ah, in Vaadin we can have starters for starters :) I meant "to begin with". Our tests looks like the very first place where this difference shows up. I don't know yet if it shows up anywhere else than in flow repository.

PR #22185, opened 2025-09-08
fix: Release no longer visible items from memory

vursen:
## Description

The PR ensures that items are removed from the HierarchicalDataCommunicator's cache and properly disposed of in both KeyMapper and DataGenerator once they move out of the viewport. Without this, such features as ComponentRenderer can cause high memory usage on the client side because the Flow client continues to retain references to DOM elements that are no longer visible.

Regression from https://github.com/vaadin/flow/issues/21876

## Type of change

- [x] Bugfix


flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicatorDataGenerationTest.java
```diff
@@ -0,0 +1,158 @@
...
+        dataCommunicator.expand(
+                Arrays.asList(new Item("Item 0"), new Item("Item 0-0")));
+        dataCommunicator.setViewportRange(0, 4);
+        fakeClientCommunication();
+
+        Mockito.doAnswer((invocation) -> {
+            Item item = invocation.getArgument(0);
+            if (!dataCommunicator.getKeyMapper().has(item)) {
+                throw new AssertionError(
+                        "Item should still be in keyMapper when destroyData is called");
```
vursen, line 151: ComponentRenderer assumes that the item key is still present when destroyData is called. If the key is missing, the component won't be detached from the DOM and memory starts to build up.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -772,8 +782,10 @@ private RootCache<T> ensureRootCache() {
                 void removeItemContext(T item) {
                     super.removeItemContext(item);
 
-                    getKeyMapper().remove(item);
-                    dataGenerator.destroyData(item);
+                    if (getKeyMapper().has(item)) {
+                        dataGenerator.destroyData(item);
+                        getKeyMapper().remove(item);
```
vursen, line 790: The call order is important. ComponentRenderer assumes that the item key is still present when destroyData is called. If the key is missing, the component won't be detached from the DOM and will cause memory to build up on the client side.

flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java
```diff
@@ -685,6 +689,12 @@ private void flush(ExecutionContext context) {
             }
         }
         update.commit(nextUpdateId++);
+
+        // Clear items that are no longer visible from the cache to free memory.
+        // This also removes them from keyMapper and disposes of their data
+        // generated by dataGenerator.
+        rootCache.removeDescendantItemIf((item) -> getKeyMapper().has(item)
```
vursen, line 699: Checking `getKeyMapper().has(item)` makes sure to only remove items that were actually rendered, leaving any other cached items untouched.

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicatorBasicTest.java
```diff
@@ -255,55 +258,6 @@ public void collapseCollectionOfItems_returnsEffectivelyCollapsedItems() {
                         Arrays.asList(new Item("Item 0"), new Item("Item 1"))));
     }
 
-    @Test
-    public void collapseItems_collapsedChildrenRemovedFromKeyMapper() {
```
vursen, line 259: **note:** this test was moved to `HierarchicalDataCommunicatorKeyGenerationTest`.

flow-data/src/test/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicatorBasicTest.java
```diff
@@ -255,55 +258,6 @@ public void collapseCollectionOfItems_returnsEffectivelyCollapsedItems() {
...
-        dataCommunicator.collapse(new Item("Item 0"));
-        fakeClientCommunication();
-        Assert.assertFalse(
-                dataCommunicator.getKeyMapper().has(new Item("Item 0-0")));
-        Assert.assertFalse(
-                dataCommunicator.getKeyMapper().has(new Item("Item 0-1")));
-    }
-
-    @Test
-    public void collapseItems_dataGeneratorDestroyDataCalledForCollapsedChildren() {
```
vursen, line 281: **note:** this test was moved to the new `HierarchicalDataCommunicatorDataGenerationTest`.

tltv (approved):

PR #22191, opened 2025-09-08
fix: Do not run any Maven goals automatically in Eclipse / VS Code

Artur-:
It seems like especially VS Code can start running `prepare-frontend` over and over again in some cases which will interfere with an externally running `mvn spring-boot:run` process that will continuously restart because of an updated `target/classes/META-INF/VAADIN/config/flow-build-info.json`

The only identified case where `prepare-frontend` needs to run is when you use "Debug on Server" in Eclipse and the project is packaged as a war that is copied to a servlet container folder or used from an external servlet container. To handle this case, the Eclipse plugin integrates with "Debug on Server" and defines the project folder in `flow-build.info.json`

### DIFF
--- a/flow-plugins/flow-maven-plugin/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml
+++ b/flow-plugins/flow-maven-plugin/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml
@@ -5,24 +5,12 @@
       <pluginExecutionFilter>
         <goals>
           <goal>validate</goal>
+          <goal>prepare-frontend</goal>
           <goal>build-frontend</goal>
         </goals>
       </pluginExecutionFilter>
       <action>
         <ignore></ignore>
-      </action>
-    </pluginExecution>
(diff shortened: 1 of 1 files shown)


vursen:
This goal seems to be also causing VS Code to generate a bunch of frontend files in the test packages, such as `web-component.html` and similar, which I always have to revert afterwards.

Artur-:
The same files will still be generated in dev mode, as the dev mode class runs the same code

vursen:
> as the dev mode class runs the same code

Is there a way to prevent this somehow?

Artur-:
Why would you want that? Something would likely stop working 

If I rmember correctly an index.html is generated so you can modify it and the same for web-component.html if you have exported web components

vursen:
> If I rmember correctly an index.html is generated so you can modify it and the same for web-component.html if you have exported web components

I mean I'd prefer them not to be created eagerly.

Artur-:
That sounds like a totally different issue though

tepi (approved):

Issue #21874, opened 2025-07-15
Element API causes unnecessary overload


subITCSS:
### Description

When the list of serilized Elements is empty - this code (image 1) in the Element.java causes a huge overload of cpu time.

This is causes in the `com.vaadin.flow.dom.Element`

<img width="1120" height="389" alt="Image" src="https://github.com/user-attachments/assets/6855ad51-0b14-44ba-9469-ff9304cad453" />
and

<img width="800" height="428" alt="Image" src="https://github.com/user-attachments/assets/4fe1e14c-6704-4d9c-b1dc-7c6459452c3b" />

The swap of stream to array to stream conversion handling is additionally a huge overload.

### Expected outcome

When this is just wrapped with an if - the execution could be much faster.

### Minimal reproducible example

Give a empty list to the element api.

### Steps to reproduce

Use the element api

### Environment

Vaadin version(s): 24.7.11
OS:

mukherjeesudebi:
I did a simple test with the below code 
```
public class ExecuteJsPerformance extends Div {

	private static final int SAMPLE_SIZE = 100_000;

	public ExecuteJsPerformance() {
		var button = new Button().getElement();

		runTest("Test with if without parameters", this::getDurationWithIf);
		runTest("Test with if with parameters", this::getDurationWithIf, button);

		runTest("Test without if without parameters", this::getDurationWithoutIf);
		runTest("Test without if with parameters", this::getDurationWithoutIf, button);

		
	}

	private void runTest(String message, Function<Serializable[], Long> consumer, Serializable... parameters) {
		var median = IntStream.range(0, SAMPLE_SIZE).boxed().map(item -> consumer.apply(parameters)).sorted().toList()
				.get(SAMPLE_SIZE / 2);
		System.err.println(message + ":" + median);
	}

	private long getDurationWithoutIf(Serializable... parameters) {
		var stopwatch = StopWatch.create();
		stopwatch.start();
		Stream<Serializable> wrappedParameters = Stream.concat(Stream.of(parameters), Stream.of(this));
		stopwatch.stop();
		return stopwatch.getNanoTime();
	}

	private long getDurationWithIf(Serializable... parameters) {
		var stopwatch = StopWatch.create();
		stopwatch.start();
		Stream<Serializable> wrappedParameters;
		if (parameters.length == 0) {
			wrappedParameters = Stream.of(this);
		} else {
			wrappedParameters = Stream.concat(Stream.of(parameters), Stream.of(this));
		}
[...]

PR #22196, opened 2025-09-08
fix: optimize Element API for small datasets

tltv:
Reduces usage of Stream API in executeJs and callJsFunction methods for improved performance with small datasets. These methods are usually used with one to few objects in the dataset, they can be called a lot depending on a number of active sessions. This change replaces Stream API usage with Array API, Arrays.copyOf, and System.arraycopy.

Fixes: #21874


tltv:
Test class to check performance of Stream (before) vs if+Arrays.copyOf (after):
```java
public class ExecuteJsPerformanceTest extends Element {

    private static final int SAMPLE_SIZE = 10;

    public ExecuteJsPerformanceTest() {
        super("div");
    }

        ExecuteJsPerformanceTest test = new ExecuteJsPerformanceTest();
        test.test();
    }

    public void test() {
        var div = new Element("div");

        runTest("Test after (no params)", this::getDurationAfter);
        runTest("Test after (with params)", this::getDurationAfter, div);

        runTest("Test before (no params)", this::getDurationBefore);
        runTest("Test before (with params)", this::getDurationBefore, div);

    }

    private void runTest(String message,
            Function<Serializable[], Long> consumer,
            Serializable... parameters) {
        var median = IntStream.range(0, SAMPLE_SIZE).boxed()
                .map(item -> consumer.apply(parameters)).sorted().toList()
                .get(SAMPLE_SIZE / 2);
        System.err.println(message + ":" + median);
    }

    private long getDurationBefore(Serializable... parameters) {
        var stopwatch = StopWatch.create();
        stopwatch.start();
        Stream<Serializable> wrappedParameters = Stream
                .concat(Stream.of(parameters), Stream.of(this));
        stopwatch.stop();
        return stopwatch.getNanoTime();
    }

    private long getDurationAfter(Serializable... parameters) {
        var stopwatch = StopWatch.create();
        stopwatch.start();
        Serializable[] wrappedParameters;
        if (parameters.length == 0) {
            wrappedParameters = new Serializable[] { this };
        } else {
            wrappedParameters = Arrays.copyOf(parameters,
                    parameters.length + 1);
            wrappedParameters[parameters.length] = this;
        }
        stopwatch.stop();
        return stopwatch.getNanoTime();
    }

}
```

tltv:
Test result from CI:
Test after (no params):1393
Test after (with params):1262
Test before (no params):3617
Test before (with params):3617

PR #22203, opened 2025-09-09
test: fix PWA test

mcollovati:
Looking at test failures it seems the content type of sw.js produced by webpack is now text/javascript instead of application/javascript. In addition the generated sw.js contains eval instructions, so assertions on expected javascript code need to be escaped.


mcollovati:
Tested on Flow 23.6 Snapshot job, build number 764

caalador (approved):

PR #22209, opened 2025-09-09
Fail tests if there are uncaught exceptions from effects

Legioth:
As part of refactoring signal error handling, exceptions from effects are no longer propagated to the caller. To ensure issues are not hidden in tests, we capture those exceptions and fail the test for any exception that wasn't cleared using assertUncaughtException.

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/AbstractSinglePropertyField.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/AbstractSinglePropertyField.java
@@ -154,10 +154,10 @@ public abstract class AbstractSinglePropertyField<C extends AbstractField<C, T>,
      * @param defaultValue
      *            the default value to use if the property isn't defined
      * @param acceptNullValues
-     *            if <code>true</code>, an exception will be thrown if the model
-     *            value is set to <code>null</code>; if <code>false</code> the
-     *            property will be removed when the model value is set to
-     *            <code>null</code>
+     *            if <code>true</code>, the property will be removed when the
+     *            model value is set to <code>null</code> ; if
+     *            <code>false</code> an exception will be thrown if the model
+     *            value is set to <code>null</code>
      */
     @SuppressWarnings("unchecked")
     public AbstractSinglePropertyField(String propertyName, T defaultValue,
--- a/flow-tests/vaadin-spring-tests/test-spring-common/pom.xml
+++ b/flow-tests/vaadin-spring-tests/test-spring-common/pom.xml
@@ -152,7 +152,7 @@
         <dependency>
             <groupId>org.springdoc</groupId>
             <artifactId>springdoc-openapi-starter-webmvc-ui</artifactId>
-            <version>2.8.12</version>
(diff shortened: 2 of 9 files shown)


Legioth:
Not yet ready. Need to also fail tests for assertion errors that are currently not passed to the uncaught exception handler.

Legioth:
New attempt. Now also passing uncaught errors to the same handler.

signals/src/main/java/com/vaadin/signals/impl/Effect.java
```diff
@@ -62,9 +59,11 @@ public Effect(Runnable action) {
                 thread.getUncaughtExceptionHandler().uncaughtException(thread,
                         e);
             } catch (Error e) {
-                getLogger().error(
-                        "Uncaught error from effect. The effect will no longer be active.",
-                        e);
+                Thread thread = Thread.currentThread();
+                thread.getUncaughtExceptionHandler().uncaughtException(thread,
+                        new Error(
+                                "Uncaught error from effect. The effect will no longer be active.",
```
mshabarov, line 65: Does it make sense to add a similar message to `Exception` handler ? E.g. `new Exception("Uncaught recoverable exception from effect. The effect will be kept alive.")`.

signals/src/test/java/com/vaadin/signals/SignalTestBase.java
```diff
@@ -114,8 +122,37 @@ protected TestExecutor useTestFallbackEffectDispatcher() {
...
+    void setupExceptionHandler() {
+        Thread.currentThread()
+                .setUncaughtExceptionHandler((thread, throwable) -> {
+                    uncaughtExceptions.add(throwable);
+                });
+    }
+
     @AfterEach
     void clear() {
+        Thread.currentThread().setUncaughtExceptionHandler(null);
```
mshabarov, line 162: Would it be better to save `Thread.currentThread().getUncaughtExceptionHandler()` in `BeforeEach` and restore it here? As setting `null` may affect other unit tests.

signals/src/main/java/com/vaadin/signals/impl/Effect.java
Legioth: 1. The extra wrapping makes the stacktrace more difficult to read. I'd avoid putting that burden on users unless really necessary.
2. I assume that it's expected that effects remain active so I don't see any value in explicitly pointing out that.

signals/src/test/java/com/vaadin/signals/SignalTestBase.java
Legioth: That wouldn't do exactly what you expect due to the way getUncaughtExceptionHandler() is implemented (but it might still be good enough).

The only reason why setting back to `null` would have any impact is if the testing framework has defined a custom exception handler and that doesn't seem to be the case at least today.

Legioth: Added a guard in the setup method to verify that no custom error handler was present before running the test.

PR #22215, opened 2025-09-09
Allow writing signals from callbacks

### LINKED ISSUE #22189: Allow changing signals values from inside an effect or computed callback
### Describe your motivation

We currently disallow making changing any signal value from inside the callback of a signal effect or computed signal. The intent of this restriction is to avoid the risk of causing infinite loops. The problem is that there are cases where such changes are very appropriate and they are fine as long as there are no cycles in signal dependencies.

In many cases, using a "[linked signal](https://github.com/vaadin/flow/issues/22184)" is a better approach than writing signals from effects. One example where a linked signal becomes problematic is if multiple target signals are involved and you want to write to only of of them depending on the circumstances.

The implementation in [Preact](https://github.com/preactjs/signals/blob/main/packages/core/src/index.ts) has a couple of checks to help debug any possible infinite loops:
* Each effect and computed signal instance has a `RUNNING` flag. An error is thrown if trying to run the callback of an effect or computed signal when this flag is set.
* There's a global `batchIteration` value that keeps track of the nesting depth of modifications. Trying to set the value of any signal throws an error if the counter is over 100.

Because of the potentially asynchronous nature of resolving effects on the server, we cannot rely on simple instance flags or global counters. Instead, we need to explicitly pass context along whenever an effect invalidation is dispatched. This could take the form of collecting a set of affected effect/computed instances and then throwing if the same instance is already in the set or if the size of the set grows beyond e.g. 100.

Legioth:
Infinite loops are detected by checking that an effect is not active
while it's being invalidated by a regular signal write. This in turn
requires changing the Usage concept so listeners can be added to new
usages immediately when they are created rather than only after running
the whole effect callback. Also introducing a concept of immediate
changes for the listeners to avoid reporting infinite loops for
concurrent changes which are reported immediately when adding the
listener.

The whole concept of a read-only transaction is removed since it's no
longer relevant.

Fixes #22189

### DIFF
--- a/signals/src/main/java/com/vaadin/signals/impl/Effect.java
+++ b/signals/src/main/java/com/vaadin/signals/impl/Effect.java
@@ -32,7 +32,7 @@ import com.vaadin.signals.SignalEnvironment;
  * based the signals read during the most recent invocation.
  */
 public class Effect {
-    private static ThreadLocal<LinkedList<Effect>> activeEffects = ThreadLocal
+    private static final ThreadLocal<LinkedList<Effect>> activeEffects = ThreadLocal
             .withInitial(() -> new LinkedList<>());
 
     private final Executor dispatcher;
@@ -134,11 +134,11 @@ public class Effect {
 
     private void scheduleInvalidate() {
         if (invalidateScheduled.compareAndSet(false, true)) {
-            var inhertedActive = new LinkedList<>(activeEffects.get());
+            var inheritedActive = new LinkedList<>(activeEffects.get());
 
             dispatcher.execute(() -> {
                 var oldActive = activeEffects.get();
-                activeEffects.set(inhertedActive);
+                activeEffects.set(inheritedActive);
 
                 try {
                     invalidate();
@@ -158,10 +158,8 @@ public class Effect {
     }
 
     private void clearRegistrations() {
-        if (registrations != null) {
-            registrations.forEach(Runnable::run);
-            registrations.clear();
-        }
+        registrations.forEach(Runnable::run);
+        registrations.clear();
     }
 
     /**


signals/src/main/java/com/vaadin/signals/impl/Effect.java
```diff
@@ -30,9 +32,11 @@
  * based the signals read during the most recent invocation.
  */
 public class Effect {
+    private static ThreadLocal<LinkedList<Effect>> activeEffects = ThreadLocal
```
tltv:
```suggestion
    private final static ThreadLocal<LinkedList<Effect>> activeEffects = ThreadLocal
            .withInitial(LinkedList::new);
```

signals/src/main/java/com/vaadin/signals/impl/Effect.java
```diff
@@ -94,23 +98,54 @@ public Effect(Runnable action, Executor dispatcher) {
...
+        }
+
+        scheduleInvalidate();
+        return false;
     }
 
     private void scheduleInvalidate() {
         if (invalidateScheduled.compareAndSet(false, true)) {
-            dispatcher.execute(this::invalidate);
+            var inhertedActive = new LinkedList<>(activeEffects.get());
```
tltv: Typo. Rename to `inheritedActive` if that's what it was supposed to be.

signals/src/main/java/com/vaadin/signals/impl/Effect.java
```diff
@@ -123,9 +158,9 @@ private synchronized void invalidate() {
     }
 
     private void clearRegistrations() {
-        if (registration != null) {
-            registration.run();
-            registration = null;
+        if (registrations != null) {
```
tltv:
```suggestion
        registrations.forEach(Runnable::run);
        registrations.clear();
```

signals/src/main/java/com/vaadin/signals/impl/UsageTracker.java
```diff
@@ -132,15 +133,15 @@ public Runnable onNextChange(TransientListener listener) {
         }
     };
 
-    private static final ThreadLocal<Collection<Usage>> currentTracker = new ThreadLocal<>();
+    private static final ThreadLocal<Consumer<Usage>> currentTracker = new ThreadLocal<>();
```
tltv, line 136: Could we replace Consumer<Usage> with a new interface:

```java
@FunctionalInterface
public interface Tracker {
    void track(Usage usage);
}
```

signals/src/main/java/com/vaadin/signals/impl/Effect.java
Legioth: Done

signals/src/main/java/com/vaadin/signals/impl/Effect.java
Legioth: Done

signals/src/main/java/com/vaadin/signals/impl/Effect.java
Legioth: Done

signals/src/main/java/com/vaadin/signals/impl/UsageTracker.java
Legioth: Would that bring enough additional clarity to justify the additional code? I don't think so for this kind of internal API.

tltv: Maybe not alone. Some clarity would be welcome in `Effect.revalidate()`. This was just one small idea. Nothing blocking.

PR #22225, opened 2025-09-11
fix: fix init location computation

### LINKED ISSUE #22211: Error 400 Invalid location: Relative path cannot start with /
### Description of the bug

see https://vaadin.com/forum/t/error-400-invalid-location-relative-path-cannot-start-with/178484

### Expected behavior

the VaadinServlet should work correctly

https://vaadin.com/forum/t/error-400-invalid-location-relative-path-cannot-start-with/178484

### Versions

- Vaadin / Flow version:  24.8.3
- Java version: 17
- OS version: OpenSuse 15.6 
- Application Server (if applicable): Jetty 12.0.23
- IDE (if applicable): IntelliJ

mcollovati:
Flow client computes the initial location to be sent to the server by getting the current location and removing the document base URI part. However, the process performs an URL decode only for the location causing replacement to fail if the web application context path contains URL encodable characters.
This change makes sure that both texts are URL decoded before performing the replacement. It also ensure that the regex based on document baseURI is correctly escaped.

Fixes #22211


mcollovati:
Added `target` labels, but PRs will need manual update to fix the project version in the new POM files.

flow-tests/pom.xml
```diff
@@ -295,6 +295,8 @@
                 <module>test-pwa-disabled-offline</module>
                 <module>test-pwa-disabled-offline/pom-production.xml</module>
                 <module>test-router-custom-context</module>
+                <module>test-router-custom-context/pom-encoded-ctx.xml</module>
```
caalador: The Multiple poms in same module path generates so many issues with IDEA. Could we instead have them as separate module folders and just bring in the `test-routest-custom-context` content as we do for instance with embedding-test-assets
```
        <dependency>
            <groupId>com.vaadin</groupId>
            <artifactId>flow-embedding-test-assets</artifactId>
        </dependency>

        <dependency>
            <groupId>com.vaadin</groupId>
            <artifactId>flow-embedding-test-assets</artifactId>
            <scope>test</scope>
            <classifier>tests</classifier>
            <type>test-jar</type>
        </dependency>
```

mcollovati: Good point. I'll change it to use multiple modules as you suggest.

caalador:
Caused by: org.eclipse.aether.resolution.ArtifactResolutionException: The following artifacts could not be resolved: com.vaadin:test-router-custom-context:war:999.99-SNAPSHOT, com.vaadin:test-router-custom-context:jar:tests:999.99-SNAPSHOT: Could not find artifact com.vaadin:test-router-custom-context:war:999.99-SNAPSHOT in vaadin-prereleases

caalador:
Seems they drop into different collection sets so the war is not available.
Probably need to give them all a `pos` so they are in the same run.

mcollovati:
> Caused by: org.eclipse.aether.resolution.ArtifactResolutionException: The following artifacts could not be resolved: com.vaadin:test-router-custom-context:war:999.99-SNAPSHOT, com.vaadin:test-router-custom-context:jar:tests:999.99-SNAPSHOT: Could not find artifact com.vaadin:test-router-custom-context:war:999.99-SNAPSHOT in vaadin-prereleases

Damn, the test-router-custom-context module is not in the projects list. Need to fix somehow the matrix

Issue #22218, opened 2025-09-10
Correct commercial products list on the 7-days trial screens


mshabarov:
### Description of the bug

To correct the list of commercial products on the 7-days trial screens as follows:

Start trial screen
```
Get full access:
Vaadin Copilot - AI-powered assistant for building UIs
Pro components - Charts, Grid Pro, CRUD and more
Team features: Acceleration Kits
```

Extend trial screen
```
Continue getting full access to:
Vaadin Copilot - AI-powered assistant for building UIs
Pro components - Charts, Grid Pro, CRUD and more
Team features: Acceleration Kits
```

### Versions
Vaadin 24.9.0-rc1

PR #22229, opened 2025-09-11
chore: List same relevant products for both screens

mshabarov:
Fixes https://github.com/vaadin/flow/issues/22218


mshabarov:
New screen:

Screenshot 2025-09-11 at 15 52 30

(Same list for "Start 7-day trial")

mshabarov:
> The code inside <ul>...</ul> seems to be exactly the same in both cases.
Can we move it into a function to avoid duplication?

Good point, done.

PR #22231, opened 2025-09-11
Synchronize initial effect invocation

Legioth:
The initial invocation may add listeners to shared signal instances which means that there's a possibility that invalidation is run from a different thread while the initial invocation is still ongoing. This means that the initial effect invocation also needs to be synchronized in the same way as invalidation.

### DIFF
--- a/signals/src/main/java/com/vaadin/signals/impl/Effect.java
+++ b/signals/src/main/java/com/vaadin/signals/impl/Effect.java
@@ -90,7 +90,11 @@ public class Effect {
         assert dispatcher != null;
         this.dispatcher = dispatcher;
 
-        dispatcher.execute(this::revalidate);
+        dispatcher.execute(() -> {
+            synchronized (this) {
+                revalidate();
+            }
+        });
     }
 
     private void revalidate() {


signals/src/main/java/com/vaadin/signals/impl/Effect.java
```diff
@@ -90,7 +90,11 @@ public Effect(Runnable action, Executor dispatcher) {
         assert dispatcher != null;
         this.dispatcher = dispatcher;
 
-        dispatcher.execute(this::revalidate);
+        dispatcher.execute(() -> {
+            synchronized (this) {
+                revalidate();
```
mshabarov, line 95: What if `revalidate()` itself would be `synchronized` similar to `invalidate()`? They both are run with `execute`, i.e. in a separate thread, so would be logical to align them this way.

Legioth: The "contract" now is that you synchronize _before_ calling `revalidate()` since that's what `invalidate()` needed to do anyways. We can also simplify the code a bit by making the method `synchronized` at the expense of making the semantics slightly less clear.

mshabarov: For the case being considered, yes, this should fix the problem of concurrent invalidation while not being completed with the initialisation in constructor.

To be honest it is hard to analyse what would be the best thread safe solution here as `revalidate()` and `invalidate` are both private (cannot be invoked by outer classes) and they depend on each other.

I only afraid a potential mistake that, if `revalidate` will be invoked in other place in the future without being locked same way, we'll be pulled into the same problem.

PR #22232, opened 2025-09-12
chore: add pre-trial usage statistics

### LINKED ISSUE #22111: Usage statistics for pre-trial
### Describe your motivation

We want to know to many new users start the 7-days pre-trial and how many users see the trial screen overall.

- Record a new usage stats when Vaadin app opens the 7-days pre-trial splash screen.
- Record a new usage stats when user clicks on "Start 7-days trial" button.

We could move the stats to the LicenseChecker, but it doesn't depend on Flow and thus has no access to vaadin-dev-server.

mcollovati:
Fixes #22111

### DIFF
--- a/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
+++ b/vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
@@ -363,6 +363,8 @@ public class DebugWindowConnection implements BrowserLiveReload {
                     BuildType.DEVELOPMENT, null,
                     Capabilities.of(Capability.PRE_TRIAL));
         } catch (PreTrialLicenseValidationException e) {
+            DevModeUsageStatistics
+                    .collectEvent("pre-trial/" + product.getName());
             errorMessage = e.getMessage();
             preTrial = e.getPreTrial();
             command = "license-check-nokey";
@@ -395,10 +397,12 @@ public class DebugWindowConnection implements BrowserLiveReload {
             JsonNode data) {
         try {
             PreTrial preTrial = LicenseChecker.startPreTrial();
+            DevModeUsageStatistics.collectEvent("pre-trial/activated");
             send(resource, "license-pretrial-started", preTrial);
         } catch (PreTrialCreationException.Expired ex) {
             send(resource, "license-pretrial-expired", null);
         } catch (Exception ex) {
+            DevModeUsageStatistics.collectEvent("pre-trial/activation-failed");
             send(resource, "license-pretrial-failed", null);
         }
     }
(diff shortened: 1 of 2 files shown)


vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -395,10 +397,12 @@ private void handlePreTrialStart(AtmosphereResource resource,
             JsonNode data) {
         try {
             PreTrial preTrial = LicenseChecker.startPreTrial();
+            DevModeUsageStatistics.collectEvent("pre-trial/activated");
             send(resource, "license-pretrial-started", preTrial);
         } catch (PreTrialCreationException.Expired ex) {
             send(resource, "license-pretrial-expired", null);
         } catch (Exception ex) {
+            DevModeUsageStatistics.collectEvent("pre-trial/activation-failed");
```
mshabarov, line 405: Not sure what this parameter would tell us. We don't record a reason for failure, so don't see a big value in recording it.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
```diff
@@ -363,6 +363,8 @@ private void handleLicenseCheck(AtmosphereResource resource,
                     BuildType.DEVELOPMENT, null,
                     Capabilities.of(Capability.PRE_TRIAL));
         } catch (PreTrialLicenseValidationException e) {
+            DevModeUsageStatistics
+                    .collectEvent("pre-trial/" + product.getName());
```
mshabarov, line 367: I'd also add a record for commercial banner in `NodeTasks`, if productionMode=true and system parameter / plugin parameter is present.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
mcollovati: It was to get an idea if the trial activation process is working smoothly or not.
For example, if we do not see any ping for trial activation, it could mean that nobody clicked the button, but also that there's a bug on the client side preventing activation. The failed ping could work as an early alert of such bugs.

mshabarov: Ok, makes sense.

vaadin-dev-server/src/main/java/com/vaadin/base/devserver/DebugWindowConnection.java
mshabarov: Not possible with the current design as `UsageStatistics` does not report stats during production builds.

Issue #21733, opened 2025-06-19
Add a ComponentEffect helper for creating child components from a list signal


Legioth:
### Describe your motivation

One very common use case is that you have a list of items in a `ListSignal` and you want to create one child component for each of those items. The only obvious way of doing that right now is to create an effect that removes all child components and creates new ones each time the structure of the list changes. This is relatively inefficient and requires some boilerplate code. An efficient solution would require significantly more code.

### Describe the solution you'd like

```java
ComponentEffect.bindChildren(parent, listSignal, childSignal -> createChildComponent(childSignal));
```

This effect should make sure that the child components of `parent` always match the contents of `listSignal`. If an item is inserted into the list, then a new component instance is created and inserted at the corresponding location in `parent`. If an item is removed, then the corresponding child component is detached. If the order of items changes, then the child component order is also updated.

A simple way of implementing this is to maintain a map from child signals to child components and then iterate the whole list and the child components in parallel to detect and directly address any mismatches. 

### Describe alternatives you've considered

It would be even more efficient to update the child components based on fine-grained change events for the list structure. The problem with this approach is that it would become very complicated when taking latency compensation into account since that might lead to previously applied changes suddenly "disappearing".

### Additional context

Rough proof of concept. I wouldn't trust this implementation without quite comprehensive test coverage.
```java
public static <T> void bindChildren(HasComponents parent,
        ListSignal<T> list,
        Function<ValueSignal<T>, Component> childFactory) {
    Component parentAsComponent = (Component) parent;

    HashMap<ValueSignal<T>, Component> signalToChild = new HashMap<>();

    ComponentEffect.effect(parentAsComponent, () -> {
        List<ValueSignal<T>> items = list.value();

        // Remove all items that are no longer present
        var toRemove = new HashSet<>(signalToChild.keySet());
        toRemove.removeAll(items);
        for (ValueSignal<T> removedItem : toRemove) {
            Component component = signalToChild.remove(removedItem);
            component.removeFromParent();
        }

        // Cache the children to avoid multiple traversals
        LinkedList<Component> remainingChildren = parentAsComponent
                .getChildren()
                .collect(Collectors.toCollection(LinkedList::new));

        for (int i = 0; i < items.size(); i++) {
            ValueSignal<T> item = items.get(i);

            Component expectedChild = signalToChild.computeIfAbsent(item,
                    childFactory);
            if (remainingChildren.isEmpty()
                    || expectedChild.getParent().orElse(null) != parent) {
[...]

PR #22240, opened 2025-09-12
feat: add ComponentEffect.bindChildren

tltv:
Adds helper API ComponentEffect for binding ListSignal to any parent component or element.

Fixes: #21733


flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -228,6 +239,70 @@ public static <C extends Component> Registration format(C owner,
         return format(owner, setter, locale, format, signals);
     }
 
+    /**
+     * Binds {@link ListSignal} to parent component with a component factory.
```
Legioth: Do we expect that users already know what the concept of binding a list signal to a parent component means?

Could be useful to describe what binding will lead to, i.e. having one child component for each entry in the list and automatically adding/removing/reordering children if the structure of the list changes.

Could also be useful to describe the point of receiving `Signal<T>` rather than just a basic `T` to the factory method.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -228,6 +239,70 @@ public static <C extends Component> Registration format(C owner,
...
+     * Parent component mush implement {@link HasComponents} and it must be
```
Legioth: Seems redundant since this is just using more words to say what's already said by the type definition: `PARENT extends Component & HasComponents`

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -228,6 +239,70 @@ public static <C extends Component> Registration format(C owner,
...
+     * subtype of {@link Component}
+     *
+     * @param parent
+     *            Target parent component, must not be <code>null</code>
```
Legioth: Don't capitalize incomplete sentences.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -228,6 +239,70 @@ public static <C extends Component> Registration format(C owner,
...
+     * @param list
+     *            list signal to bind to the parent, must not be
+     *            <code>null</code>
+     * @param childFactory
+     *            factory to create new component, must not be <code>null</code>
+     * @param <T>
+     *            the value type
```
Legioth: Value type of what?

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -228,6 +239,70 @@ public static <C extends Component> Registration format(C owner,
...
+     *            Target parent element, must not be <code>null</code>
+     * @param list
+     *            list signal to bind to the parent, must not be
+     *            <code>null</code>
+     * @param childFactory
+     *            factory to create new element, must not be <code>null</code>
+     * @param <T>
+     *            the value type
+     */
+    public static <T> void bindChildren(Component parentComponent,
```
Legioth: Not sure if it's useful to have this method in the public API of the high-level `ComponentSignal` type. I'd suggest keeping it `private` until we have clarity on the `Element` API where this low-level functionality would actually belong.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -228,6 +239,70 @@ public static <C extends Component> Registration format(C owner,
...
+    public static <T, PARENT extends Component & HasComponents> void bindChildren(
+            PARENT parent, ListSignal<T> list,
+            SerializableFunction<ValueSignal<T>, Component> childFactory) {
+        Objects.requireNonNull(parent, "Parent component cannot be null");
+        Objects.requireNonNull(childFactory,
+                "Child component factory cannot be null");
+        bindChildren(parent, parent.getElement(), list,
+                signalValue -> Optional
+                        .ofNullable(childFactory.apply(signalValue))
+                        .map(Component::getElement).orElse(null));
```
Legioth: This behavior is not documented

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -228,6 +239,70 @@ public static <C extends Component> Registration format(C owner,
...
+                "Parent component cannot be null");
+        Objects.requireNonNull(parent, "Parent element cannot be null");
+        Objects.requireNonNull(list, "ListSignal cannot be null");
+        Objects.requireNonNull(childFactory,
+                "Child element factory cannot be null");
+
+        HashMap<ValueSignal<T>, Element> valueSignalToChild = new HashMap<>();
+
+        ComponentEffect.effect(parentComponent, () -> {
+            new UpdateByListSignal<>(parent, list, childFactory,
```
Legioth: I don't understand why any new instances are needed here since all the state is anyways `final` and could thus just as well be captured in the closure.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -275,4 +350,125 @@ private void close() {
...
+    /**
+     * Abstract base class for updating children of a parent element according
+     * to a signal.
+     *
+     * @param <SIGNAL>
+     *            the type of the signal
+     * @param <T>
+     *            the value type
+     */
+    private static class AbstractUpdateBySignal<SIGNAL, T>
```
Legioth: What's the point of an abstract class with only one implementation and no way for others to define additional subclasses?

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -275,4 +350,125 @@ private void close() {
...
+    /**
+     * Helper class to update children of a parent element according to a list.
+     *
+     * @param <T>
+     *            the value type
+     */
+    private static class UpdateByListSignal<T>
+            extends AbstractUpdateBySignal<ListSignal<T>, T> {
+
+        private final List<ValueSignal<T>> list;
```
Legioth: How about `childSignals` to describe the content rather than the structure?

Legioth (changes requested):
I'm giving up at this point. I might be biased since I created the prototype but it seems to me like the additional structure (e.g. the newly introduced classes) is just making things more confusing by moving related things further apart from each other.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: I'll explain a bit to help understand it better.

It would be a very long closure where all these four parameters (plus more when going deeper) makes it hard to make it reusable without introducing methods with long list of parameters. I avoided that by only requiring minimal number of needed parameters in cost of new instance creation. 

Everything related to synchronizing element with the signal is now in one class (or two, if super type is counted in). If/when needed, this gives better building blocks to add same for another signal type like map signal. Or for new features like possibility to choose insertion point for the signal list in the parent element (like a default `slot` in web component). Everything is private to limit usage in ComponentEffect only, which makes it simple to adjust later, hopefully without full rewrite. Keeping it all in closure would probably make it harder to add new stuff to it and would probably require rewriting a lot more.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: As explained in other comment, this would be a building block to add support for other signal types. It's now private just for ComponentEffect.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: Done

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: Done

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: Added comment

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: Done

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
Legioth: Could it make the structure easier to follow if the parameters would be grouped together in a `record` that is passed around to static methods instead? `doSomething(new Instance(a, b, c, d));` is much more easily digested than `new Instance(a, b, c, d).doSomething();`.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
Legioth: YAGNI.

We will most likely not get the abstraction right when it's used by only one case. That's why I suggest introducing the abstraction once we need it based on what we know then rather than introducing it preemptively based on speculation.

tltv: For a starters, let's then agree on removal of the abstract layer.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
```diff
@@ -339,6 +355,207 @@ public void format_defaultLocale_signalValuesChange_formattedStringUpdated() {
...
+            assertEquals(1, parentComponent.getComponentCount());
+            assertEquals(expectedComponent,
+                    parentComponent.getChildren().findFirst().orElse(null));
+            assertEquals("first", expectedComponent.getValue());
+        });
+    }
+
+    @Test
+    public void bindChildren_listSignalWithItemsWithNotEmptyParent_parentUpdated() {
+        // if parent has children initially, they are preserved after list
```
Legioth: Preserving after new items is probably what's easiest to implement but I don't think that's what users would expect. Would be more expected to throw, remove, or preserve before.

The reason preserving after is surprising is that both `add` and `bindChildren` are a stateful operations which means that you would expect different outcomes if you change the ordering between them. But with the current implementation, you would get the same result no matter if you do `add` or `bindChildren` first.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
```diff
@@ -339,6 +355,207 @@ public void format_defaultLocale_signalValuesChange_formattedStringUpdated() {
...
+
+            assertEquals("Parent component children count is wrong", 2,
+                    parentComponent.getComponentCount());
+            assertEquals("first", ((TestComponent) parentComponent.getChildren()
+                    .toList().get(0)).getValue());
+            assertEquals("added directly", ((TestComponent) parentComponent
+                    .getChildren().toList().get(1)).getValue());
+        });
+    }
+
```
Legioth, line 601: Some more things to test for:
* Verify that an optimal number of `Element` operations are applied for a simple addition, removal or reorder. (I realize we might need some ugly mocking to test this but it defeats the purpose to have an overly complex implementation if we're not also verifying that the additional complexity does what it's supposed to do.)
* Verify that the right end result is also reached when multiple operations are applied at the same time through a transaction. In that situation, the effect should be run only once at the end of the transaction. We don't need to ensure an optimal number of `Element` operations that involve reordering but we should still see that we avoid redundant operations for multiple additions or removals.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: It may not be necessary to call static method as we have the record. I've updated code to use record instead, and using `doSomething(new Instance(a, b, c, d));` pattern too but it could as well just be `Instance(a, b, c, d).doSomething();` since magic happens in the record.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
tltv: Current behavior can be documented in JavaDoc to be clear with end result. There's also possibility to insert element in between (inside the list signal), anywhere. What would one expect to happen in that case?

Exception or just removal is also option, but then some use cases like having custom tooltip, overlay etc. element that just need to be attached somewhere would stop working inside parent element.

I think the question really is do we want to support this kind of "alien" elements in the parent element?

Legioth: Throwing is always a safe default since it's not really a breaking change to change that into doing something else once we learn what's actually needed.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
Legioth: Doesn't really change since the structure is the same but now the `runEffect` method gives an additional abstraction level to understand. I think the core of the confusion is that `UpdateElementByChildSignals` isn't really a "proper" object from an OOP point of view and then it also becomes weird to have instance methods that apply side effects. My idea with the record was to let the record handle only the data but not any behavior.

How about further removing abstractions for something like this?
```java
ComponentEffect.effect(parentComponent, () -> {
  var context = new BindChildrenEffectContext(parent, list.value(), childFactory, valueSignalToChild);
  removeNotPresentChildren(context);
  addAndOrderRemainingChildren(context);
});
```

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
tltv: Changed to throw IllegalStateException. One from bindChildren, if element has children. Another from effect itself validating that there are no unknown elements in the parent.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: I see your point. But don't you think it's more practical to have all code in the record as then there's no need to pass extra parameter to those methods which makes code little bit cleaner/shorter?

After previous change, your snippet would grow already since element children is also needed for validation.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
tltv: Added tests for Element operations (`insertChild(int, Element...)` and `removeFromParent()`).

tltv: Added also test with transaction.

tltv: Tests are now updated and using element attach/detach listeners to count inserts and removals.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: JavaDoc updated. Notice that the other method which takes in element is now private and I dropped Javadoc from there just to not duplicate same thing in private method. Feels unnecessary for private.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: updated

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: updated

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -134,6 +145,94 @@ public static <C extends Component, T> Registration bind(C owner,
...
+     * <p>
+     * Example of usage:
+     *
+     * <pre>
+     * ListSignal&lt;String&gt; taskList = new ListSignal&lt;&gt;(String.class);
+     *
+     * Div div = new Div();
+     *
+     * ComponentEffect.bindChildren(div, taskList, taskValueSignal -> {
+     *     var listItem = new ListItem(taskValueSignal.value());
```
Legioth: If that's a `com.vaadin.flow.component.html.ListItem`, then the parent should be `OrderedList` or `UnorderedList` rather than `Div`.

Legioth: Redundant to pass the initial text to the constructor since that's anyways done by the binding from the next line of code.

It's even a bad practice to do that since the `.value()` call creates a dependency for the outer effect on the value in each child signal and thus causes redundant no-op updates.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -134,6 +145,94 @@ public static <C extends Component, T> Registration bind(C owner,
...
+     *            <code>null</code>
+     * @param childFactory
+     *            factory to create new component, must not be <code>null</code>
+     * @param <T>
+     *            the value type of the {@link ValueSignal}s in the
+     *            {@link ListSignal}
+     * @param <PARENT>
+     *            the type of the parent component
+     * @throws IllegalStateException
+     *             thrown if parent component has children not belonging to the
```
Legioth: If there are any children in the parent when this method is called, then those children do not belong to the signal since no such children have been created yet? From that follows that it could be easier to describe that the exception is thrown if the parent wasn't empty.

Note that exceptions from the validate step inside the effect won't be thrown from this method (assuming the underlying signal-related exception handling works as it should).

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
Legioth: Where?

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
Legioth: It's weird to have complex functionality in record methods. The point of records is to "model the data, the whole data, and nothing but the data". The record as it's used now is mostly modeling behavior and the data is secondary.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -181,4 +280,126 @@ private void close() {
...
+        /**
+         * Update children of the parent element according to the list of child
+         * signals.
+         */
+        private void doUpdate() {
+            // Cache the children to avoid multiple traversals
+            LinkedList<Element> remainingChildren = parentElement.getChildren()
+                    .collect(Collectors.toCollection(LinkedList::new));
+
+            validate(remainingChildren);
```
Legioth: Wouldn't it be more relevant to do validation after creating and attaching new components since unexpected children might also appear as a side effect from those operations?

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
Legioth: One more factor is that it makes it difficult to track the scope of different variables when `valueSignalToChild` is preserved across all invocations of the effect callback while the other variables are reset with every invocation. It might be good to somehow keep them separate from each other to make it easier to keep track of the logic.

This was bad also in the prototype where everything was just mashed together but it becomes even harder to track now when they are collected into the same place which gives the impression that they work in similar ways.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -181,4 +280,126 @@ private void close() {
...
+            updateByChildSignals(remainingChildren);
+        }
+
+        /**
+         * Validate that parent element has no children not belonging to the
+         * list of child signals.
+         */
+        private void validate(LinkedList<Element> children) {
+            if (children.stream().anyMatch(
+                    element -> !valueSignalToChild.containsValue(element))) {
```
Legioth: `containsValue` is O(n) so this loop is O(n^2). Would it work to just count the children and compare that to the size of the map?

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -181,4 +280,126 @@ private void close() {
...
+                throw new IllegalStateException(
+                        "Parent element must not have children not belonging to the signal");
```
Legioth: Would be really useful to have a pointer to one of the unexpected elements in the error message to help pinpoint the problem.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -181,4 +280,126 @@ private void close() {
...
+            }
+        }
+
+        /**
+         * Remove all existing children in valueSignalToChild map that are no
+         * longer present in the list of child signals.
+         */
+        private void removeNotPresentChildren() {
+            var toRemove = new HashSet<>(valueSignalToChild.keySet());
+            childSignals.forEach(toRemove::remove);
```
Legioth: I was first confused about why we're not just looping over all items and checking `childSignals.contains` inside the loop. But then I realized that `childSignals` is a `List` so the lookup would be slow. Could add a comment to clarify this optimization.

Also: could this use `toRemove.removeAll(childSignals)` instead or would that also lead to O(n^2) performance due to list lookups?

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
```diff
@@ -179,15 +189,28 @@ public void effect_throwExceptionWhenRunningAsynchronously_delegatedToErrorHandl
...
+                effectRun.set(true);
                 throw new RuntimeException("Expected exception");
             });
 
-            ErrorEvent event = events.poll(500, TimeUnit.MILLISECONDS);
-            assertNotNull(event);
-
-            Throwable throwable = event.getThrowable();
-            assertEquals(RuntimeException.class, throwable.getClass());
+            int tries = 0;
```
Legioth: nit: this is a non-trivial amount of code change that has nothing to do with the actual topic of this PR. While I think it's fine to do smaller unrelated changes when appropriate, this approaches the size where I think it would be better to make the change as a separate PR.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
```diff
@@ -275,6 +298,384 @@ public void bind_signalValueChanges_componentUpdated() {
...
+                    parentComponent.getChildren().findFirst().orElse(null));
+            assertEquals("first", expectedComponent.getValue());
+
+            assertEquals(1, expectedComponent.attachCounter);
+            assertEquals(0, expectedComponent.detachCounter);
+        });
+    }
+
+    @Test
+    public void bindChildren_listSignalWithItemsWithNotEmptyParent_throw() {
```
Legioth: Seems like a relatively redundant test since I don't see any reasonable reason for why the test for the same condition with an empty list wouldn't also cover this case.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
```diff
@@ -275,6 +298,384 @@ public void bind_signalValueChanges_componentUpdated() {
...
+            TestLayout parentComponent = prepareTestLayout(taskList);
+
+            // move first to last
+            taskList.moveTo(taskList.value().get(0),
+                    ListSignal.ListPosition.last());
+
+            List<TestComponent> children = parentComponent.getChildren()
+                    .map(TestComponent.class::cast).toList();
+
+            assertEquals(1, children.get(0).attachCounter);
```
Legioth: Seems like a bug in the algorithm if multiple components are affected when moving in this direction? All that should be needed is something along the lines of `parent.insertChild(2, firstChild)` to move that child to the last position rather than moving the two other ones before the first one.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: JavaDoc explains child factory. "wrap childFactory to convert Component to Element" comment above this lambda expression tries to clarify this internal logic.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: About the note, yes I know, and it works as it should :)
I thought first to write it shorter like that too, but then added reference to the signal list anyway. Shorter version however feels nicer now, so I'll change it.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: Good point. It could indeed happen as a side effect too. This might require keeping track of non-signal related attach and detach events and throw immediately when detecting one. This could also improve performance since no need to loop, just need to know if attach/detach came from this binding children effect.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: `toRemove.removeAll(childSignals)` may be slow, says IDEA (not AI), and suggest to use forEach like that.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
Legioth: My question was with regards to the additional `null` handling that comes through the use of `Optional`.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: There's no way in Element API to reliably detect attach/detach of unknown child element, which makes my previous comment not working. Since parent element could get new children without any way to listen to them being attached/detached, validation should probably happen in the end of the effect.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
tltv: Yes, bug is that it should use Objects.equals instead of `==` in one remaining place.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: Refactored to static private methods.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
tltv: Moved to https://github.com/vaadin/flow/pull/22327

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
tltv: Removed

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: New validation does so now, but since it's expensive operation to pinpoint culprit element (requires getChildren() call), it's only done in the end of the effect if not already caught by fast size check earlier.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -134,6 +145,115 @@ public static <C extends Component, T> Registration bind(C owner,
...
+     * needed. Note that <code>childFactory</code> is run inside a
+     * {@link Effect}, and therefore {@link ValueSignal#value()} calls makes
+     * effect re-run automatically on signal value change.
+     * <p>
+     * Example of usage:
+     *
+     * <pre>
+     * ListSignal&lt;String&gt; taskList = new ListSignal&lt;&gt;(String.class);
+     *
+     * UnorderedList div = new UnorderedList();
```
Legioth: Weird variable name for something that isn't a `Div`

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
Legioth: Why is there explicit null handling here when the javadoc says that null should not be returned?

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -134,6 +145,115 @@ public static <C extends Component, T> Registration bind(C owner,
...
+     *            factory to create new component, must not be <code>null</code>
+     * @param <T>
+     *            the value type of the {@link ValueSignal}s in the
+     *            {@link ListSignal}
+     * @param <PARENT>
+     *            the type of the parent component
+     * @throws IllegalStateException
+     *             thrown if parent component isn't empty
+     * @throws IllegalArgumentException
+     *             thrown if childFactory returns null
```
Legioth: Don't see how that would happen since the callback is only run inside the effect and exceptions from inside an effect shouldn't escape to the caller.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -134,6 +145,115 @@ public static <C extends Component, T> Registration bind(C owner,
...
+        Objects.requireNonNull(parent, "Parent element cannot be null");
+        Objects.requireNonNull(list, "ListSignal cannot be null");
+        Objects.requireNonNull(childFactory,
+                "Child element factory cannot be null");
+
+        if (parent.getChildCount() > 0) {
+            throw new IllegalStateException(
+                    "Parent element must not have children when binding ListSignal to it");
+        }
+        // Create a child element cache outside the effect.
```
Legioth: nit: This comment only states the obvious. Could be more useful to describe why the cache needs to be outside the effect

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -181,4 +301,158 @@ private void close() {
...
+    private static <T> void validate(BindChildrenEffectContext<T> context) {
+        LinkedList<Element> children = context.parentChildrenToLinkedList();
+        if (children.size() != context.getCachedChildrenSize()) {
+            throw new IllegalStateException(
+                    "Parent element must have children matching the list signal. Unexpected child count: "
+                            + children.size() + ", expected: "
+                            + context.getCachedChildrenSize());
+        }
+        for (int index = 0; index < children.size(); index++) {
+            Element actualElement = children.get(index);
```
Legioth: Unecessary O(n^2) caused by using a linked list

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -181,4 +301,158 @@ private void close() {
...
+    private static <T> void removeNotPresentChildren(
+            BindChildrenEffectContext<T> context,
+            LinkedList<Element> remainingChildren) {
+        var toRemove = new HashSet<>(context.valueSignalToChildCache.keySet());
+        context.childSignalsList.forEach(toRemove::remove);
+        for (ValueSignal<T> removedItem : toRemove) {
+            Element element = context.valueSignalToChildCache
+                    .remove(removedItem);
+            element.removeFromParent();
+            remainingChildren.remove(element);
```
Legioth: O(n^2) by removing them one at a time instead of all in one go. On the other hand, might not matter since the inevitable `element.removeFromParent()` also has the same impact.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: IllegalStateException is thrown later when inserting null element, but makes more sense to throw it from here already.

Legioth: But does it ever actually get thrown when calling the method? I haven't yet gotten to the tests to see if we test for the case but my gut feeling is that the exception will not go to the `bindChildren` caller.

Legioth: This `Optional.ofNullabe().orElse(null)` really just means that a `null` is passed upstream which will lead to an exception in some other place. Why not instead make it throw immediately as a `NullPointerException` when trying to do `component.getElement()`?

tltv: It's never thrown from `bindChildren` call, that was wrongly documented. 
There is a test `bindChildren_withNullFromChildFactory_throws`. 

Returning null from child factory throws exception now (earlier it was `IllegalArgumentException` from instert). Now I made it throw `IllegalStateException` with a better message. Made that before I read your comment, do you think that changing it to `NullPointerException` would make difference?

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -181,4 +306,159 @@ private void close() {
...
         closed = true;
     }
+
+    /**
+     * Validate that parent element has no children not belonging to the list of
+     * child signals.
+     */
+    private static <T> void validate(BindChildrenEffectContext<T> context) {
+        LinkedList<Element> children = context.parentChildrenToLinkedList();
+        if (children.size() != context.getCachedChildrenSize()) {
```
Legioth: Seems pointless to fail fast with a less informative error message when the happy path will anyways go through a slower path that might give a more informative error message if there's any problem.

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -181,4 +306,159 @@ private void close() {
...
+         * Return existing element or generate new by child element factory.
+         *
+         * @throws IllegalStateException
+         *             if child factory adds or removes unexpected child
+         */
+        private Element getElement(ValueSignal<T> item) {
+            return valueSignalToChildCache.computeIfAbsent(item, value -> {
+                int sizeBefore = parentElement.getChildCount();
+                var element = childElementFactory.apply(value);
+                // quick validation to catch illegal state early
```
Legioth: Rather than quick validation with a non-informative error message, I might prefer the more informative error message that could anyways be produced by the slower validation that would anyways be run?

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
```diff
@@ -181,4 +306,159 @@ private void close() {
...
+                if (element != null
```
Legioth: I guess the null check is no longer needed now when we require that the factory returns a non-null value?

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
```diff
@@ -296,6 +304,583 @@ public void bind_signalValueChanges_componentUpdated() {
...
+                    valueSignal -> new TestComponent(valueSignal.value()));
+
+            var expectedComponent = new TestComponent("added directly");
+            // doing wrong
+            parentComponent.add(expectedComponent);
+
+            // causes the effect to run and exception being thrown
+            taskList.insertLast("last");
+
+            ErrorEvent event = events.poll(1000, TimeUnit.MILLISECONDS);
```
Legioth: Is this really necessary? My assumption is that everything should run directly on the main thread as long as session is locked. If that's not the case, then we might have some other bug hiding somewhere...

(repeated n times below)

flow-server/src/main/java/com/vaadin/flow/component/ComponentEffect.java
tltv: Original thinking here was to stop as soon as possible and leave the parent state there. But since parent will be in illegal state anyways (without any kind of "rollback" feature), why not let it go all the way to the final validation. I'll remove this check.

flow-server/src/test/java/com/vaadin/flow/component/ComponentEffectTest.java
tltv: Looks like it's not necessary as indeed session is locked. Fixing.

PR #22264, opened 2025-09-16
fix: copy all frontend resources from JAR files

### LINKED ISSUE #22208: Production build are missing .vue files
### Description of the bug

B"H
Hi, seems that production mode build do not collect into generate folder the .vue files.
Any suggestions how to overcome this wired behave?
`error during build:
Could not resolve "./AccountsCard.vue" from "frontend/generated/jar-resources/policyDashboard/AccountsCard/accountsCardWidget.js"
file: C:/Projects/_TRUNK/PL/shell/frontend/generated/jar-resources/policyDashboard/AccountsCard/accountsCardWidget.js
    at getRollupError (file:///C:/Projects/_TRUNK/PL/shell/node_modules/.pnpm/rollup@4.50.1/node_modules/rollup/dist/es/shared/parseAst.js:401:41)
    at error (file:///C:/Projects/_TRUNK/PL/shell/node_modules/.pnpm/rollup@4.50.1/node_modules/rollup/dist/es/shared/parseAst.js:397:42)
    at ModuleLoader.handleInvalidResolvedId (file:///C:/Projects/_TRUNK/PL/shell/node_modules/.pnpm/rollup@4.50.1/node_modules/rollup/dist/es/shared/node-entry.js:21539:24)
    at file:///C:/Projects/_TRUNK/PL/shell/node_modules/.pnpm/rollup@4.50.1/node_modules/rollup/dist/es/shared/node-entry.js:21499:26`

### Expected behavior

expected behavior will be clean exception for troubleshoot this unexpected behave + warning if some files are not loading from whatever reason.

just add some .vue and load it via one of your .js files. run production build.

### Versions

- Vaadin / Flow version: 24.6.5
- Java version: 21

mcollovati:
Frontend resources in JAR files are filtered to only copy a hard-coded set of file types like JS, TS, CSS, etc.
Since the resources are collected only from specific folders, the filter is most likely unecessary and also makes it impossible to use fronted frameworks that rely on custom extensions (e.g. Vue).
This change removes the file extension filter, allowing all files in META-INF/frontend and META-INF/resources/frontend to be copied in the frontend generated folder.

Fixes #22208


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskCopyFrontendFiles.java
```diff
@@ -32,21 +32,21 @@
 import static com.vaadin.flow.server.Constants.RESOURCES_JAR_DEFAULT;
 
 /**
- * Copies JavaScript and CSS files from JAR files into a given folder.
+ * Copies all frontend resources from JAR files into a given folder.
+ * <p>
+ * The task considers "frontend resources" all files placed in
+ * {@literal META-INF/frontend}, {@literal META-INF/resources/frontend} and
+ * {@literal META-INF/resources/[**]/themes} folders.
```
knoobie, line 39: Why no usage of `frontendExtraFileExtensions` as filter? Now even development only internal files like `README.md` or `README.txt` which might contain internal informations are also published

mcollovati: A reason for not using `frontendExtraFileExtensions` is that it is a build-plugin-only setting. If you run the application from your IDE in dev mode, then this will be ignored. The way to get around this would be to duplicate the configuration, e.g. plugin + application.properties (or system property for non-Spring applications).

About publishing internal resources, I'm not 100% sure: the files are copied in `frontend/jar-resources` but they should end up in the production bundle only if they are referenced somewhere in the application code.

Also, files in `META-INF/resources/frontend` are most likely also published by the servlet container, so they will be visible even without the Flow task filter.

knoobie: > they should end up in the production bundle only if they are referenced somewhere in the application code

That puts peace to my mind 😀 thanks for clarification

PR #22265, opened 2025-09-16
fix: handle ArrayNode in MapPutChange to fix ReactAdapterComponent.setState() (#22251)

### LINKED ISSUE #22251: Regression with state values on ReactAdapterComponent
### Description of the bug

B"H
upgrading from 24.5 to 24.8.7
.`    public void setPreferredCountries(String preferredCountries) {
        String [] countries = preferredCountries.split(",");
        String[] reversed = new String[countries.length];
        for (int i = 0; i < countries.length; i++) {
            reversed[i] = countries[countries.length - 1 - i];
        }
        super.setState("preferredCountries", reversed);
    }`

code fail:
com.vaadin.flow.internal.JsonCodec.encodeWithoutTypeInfo
throw new IllegalArgumentException("Can't encode " + value.getClass() + " to json");

this seems to be a regression. 
java.lang.IllegalArgumentException: Can't encode class com.fasterxml.jackson.databind.node.ArrayNode to json
	at com.vaadin.flow.internal.JsonCodec.encodeWithoutTypeInfo(JsonCodec.java:208)

### Expected behavior

if we have type restriction should be take in compile time

use any ReactAdapterComponent and setState an String[] or list of strings

### Versions

- Vaadin / Flow version:
- Java version:
- OS version:
- IDE (if applicable):

Artur-:
When setState() is called with a String array in ReactAdapterComponent, the array gets converted to Jackson's ArrayNode. MapPutChange.populateJson() did not handle ArrayNode, causing it to fall through to the else clause which uses the old Elemental JsonCodec that doesn't recognize Jackson types.

This resulted in: IllegalArgumentException: Can't encode class com.fasterxml.jackson.databind.node.ArrayNode to json

The fix adds a specific case to handle ArrayNode values by converting them to Elemental JSON using Json.instance().parse().

Fixes #22251

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/internal/change/MapPutChange.java
+++ b/flow-server/src/main/java/com/vaadin/flow/internal/change/MapPutChange.java
@@ -17,6 +17,7 @@
 package com.vaadin.flow.internal.change;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.BooleanNode;
 import com.fasterxml.jackson.databind.node.NumericNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
@@ -32,6 +33,7 @@ import com.vaadin.flow.shared.JsonConstants;
 
 import elemental.json.Json;
 import elemental.json.JsonObject;
+import elemental.json.JsonValue;
 
 /**
  * Change describing a changed value in a map feature.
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/internal/change/MapPutChange.java
```diff
@@ -109,6 +111,13 @@ protected void populateJson(JsonObject json, ConstantPool constantPool) {
         } else if (value instanceof TextNode node) {
             json.put(JsonConstants.CHANGE_PUT_VALUE, Json.create(JacksonCodec
                     .encodeWithConstantPool(node, constantPool).textValue()));
+        } else if (value instanceof ArrayNode node) {
+            // Convert ArrayNode to Elemental JSON array
+            json.put(JsonConstants.CHANGE_PUT_VALUE,
+                    (JsonValue) Json.instance()
+                            .parse(JacksonCodec
+                                    .encodeWithConstantPool(node, constantPool)
+                                    .toString()));
```
caalador:
```suggestion
                    JacksonUtils.createElementalArray((ArrayNode) JacksonCodec
                            .encodeWithConstantPool(node, constantPool)));
```
We have a arrynode to jsonarray method that would be cleaner to use.

caalador (approved):

PR #22273, opened 2025-09-16
fix: log creating dev bundle differently

### LINKED ISSUE #16038: "Creating a new express mode bundle. This can take a while is" reported at a late stage
### Description of the bug

The first time a bundle is built, most of the time is spent between
```
2023-02-24T16:24:05.252+02:00  INFO 56853 --- [  restartedMain] c.v.f.s.frontend.TaskRunDevBundleBuild   : An express mode bundle build is needed
```
and
```
2023-02-24T16:26:16.636+02:00  INFO 56853 --- [onPool-worker-1] c.v.f.s.frontend.TaskRunDevBundleBuild   : Creating a new express mode bundle. This can take a while but will only run when the project setup is changed, addons are added or frontend files are modified
```

because `npm install` takes place between them. The time after the "this can take a while" message is actually much shorter.

### Expected behavior

```
2023-02-24T16:26:16.636+02:00  INFO 56853 --- [onPool-worker-1] c.v.f.s.frontend.TaskRunDevBundleBuild   : Creating a new express mode bundle. This can take a while but will only run when the project setup is changed, addons are added or frontend files are modified
```

should be output instead of `An express mode bundle build is needed` so it comes before `npm install`

Any new starter + make a change to e.g. theme.json

### Versions

- Vaadin / Flow version: 24 beta2

caalador:
Log the dev bundle creation at
a earlier time.

Closes #16038

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
+++ b/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
@@ -89,13 +89,12 @@ public class TaskRunDevBundleBuild implements FallibleCommand {
         README_NOT_CREATED = "Failed to create a README file in "
                 + options.getBuildDirectoryName() + "/"
                 + Constants.DEV_BUNDLE_LOCATION;
+        getLogger().info(
+                "Creating a new development mode bundle. This can take a while but will only run when the project setup is changed, addons are added or frontend files are modified");
     }
 
     @Override
     public void execute() throws ExecutionFailedException {
-        getLogger().info(
-                "Creating a new development mode bundle. This can take a while but will only run when the project setup is changed, addons are added or frontend files are modified");
-
         runFrontendBuildTool("Vite", "vite", "vite", "build");
 
         copyPackageLockToBundleFolder();


flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskRunDevBundleBuild.java
```diff
@@ -89,13 +89,12 @@ public class TaskRunDevBundleBuild implements FallibleCommand {
...
                 + Constants.DEV_BUNDLE_LOCATION;
+        getLogger().info(
+                "Creating a new development mode bundle. This can take a while but will only run when the project setup is changed, addons are added or frontend files are modified");
     }
 
     @Override
     public void execute() throws ExecutionFailedException {
-        getLogger().info(
-                "Creating a new development mode bundle. This can take a while but will only run when the project setup is changed, addons are added or frontend files are modified");
-
```
tepi, line 98: Looking at the code I'm not 100% if it's possible to get an exception between constructor and execute? In that case the logging would not make sense - although not sure if that matters if the task building failed anyway

caalador: Yes, for instance npm install may fail, but we still have started npm install especially for creating the devBundle.
I would say that anything done before actually executing the  devBundleBuild vite run is mostly part of the devBundle requirements.

tepi: I agree. And anyway it is a very minor detail in that situation anyway.

PR #22277, opened 2025-09-16
fix: copy all frontend resources from JAR files (#22264) (CP: 2.12)

### LINKED ISSUE #22208: Production build are missing .vue files

mcollovati:
Frontend resources in JAR files are filtered to only copy a hard-coded set of file types like JS, TS, CSS, etc.
Since the resources are collected only from specific folders, the filter is most likely unecessary and also makes it impossible to use fronted frameworks that rely on custom extensions (e.g. Vue).
This change removes the file extension filter, allowing all files in META-INF/frontend and META-INF/resources/frontend to be copied in the frontend generated folder.

Fixes #22208


mcollovati:
> 2.12 still uses the `node_modules/@vaadin/flow` as the target output folder so I'm not certain we should add this change here to copy everything there.

Good point!
Let's close this PR. We can think about adding a similar change later if needed.

PR #22283, opened 2025-09-18
chore: Deprecate WebBrowser and BrowserDetails

### LINKED ISSUE #21830: Deprecate our user agent parser and associated classes
### Describe your motivation

We should no longer be using the user agent string for anything specific. In particular, the decision to show the unsupported browser page is based on feature detection rather than data in the user agent string.

We should deprecate our own parser and instead provide explicit access to the raw string from the request header and suggest that developers use e.g. https://github.com/ua-parser/uap-java if they want to extract some specific details from it.

Rather than parsing all received strings and logging a message if encountering something that isn't supported by our parser, we should parse the string only when application code calls one of the deprecated methods that rely on prased data. Furthermore, we shouldn't log messages about unparseable strings but instead just return values representing an unknown result.

caalador:
BrowserDetail only parses required parts.
Log server parsing exceptions as debug only.

Fixes #21830


flow-client/src/main/java/com/vaadin/client/BrowserInfo.java
```diff
@@ -79,26 +69,34 @@ public static BrowserInfo get() {
...
-        return !!navigator.msMaxTouchPoints;
+        if (navigator && "maxTouchPoints" in navigator) {
+            return navigator.maxTouchPoints > 0;
+        } else if (navigator && "msMaxTouchPoints" in navigator) {
+            return navigator.msMaxTouchPoints > 0;
+        } else {
+            var mQ = $wnd.matchMedia && matchMedia("(pointer:coarse)");
+            if (mQ && mQ.media === "(pointer:coarse)") {
+                return !!mQ.matches;
+            }
```
mshabarov, line 82: This adds a new kind of check to just deprecation task, just a thing that you noticed and want to fix in one PR?

flow-client/src/main/java/com/vaadin/client/BrowserInfo.java
```diff
@@ -228,7 +256,10 @@ public boolean isTouchDevice() {
      * Checks if the browser is run on Android.
      *
      * @return true if the browser is run on Android, false otherwise
+     * @deprecated use a parsing library like ua-parser-js to parse the user
+     *             agent from {@link #getBrowserString()}
      */
+    @Deprecated
     public boolean isAndroid() {
```
mshabarov, line 263: Also deprecate `isAndroidWithBrokenScrollTop`.

flow-client/src/main/java/com/vaadin/client/BrowserInfo.java
```diff
@@ -269,7 +303,10 @@ public int getBrowserMajorVersion() {
      * @see #getBrowserMajorVersion()
      *
      * @return The minor version of the browser, or -1 if not known/parsed.
+     * @deprecated use a parsing library like ua-parser-js to parse the user
+     *             agent from {@link #getBrowserString()}
      */
+    @Deprecated
     public int getBrowserMinorVersion() {
```
mshabarov, line 314: Shall we also deprecate `isTouchDevice` ?

flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
```diff
@@ -504,7 +577,10 @@ public boolean isGecko() {
      * @return true if it is WebKit, false otherwise
      */
     public boolean isWebKit() {
```
mshabarov, line 559: Deprecate?

mshabarov: Also deprecate `BrowserInfo::isWebKit`?

flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
```diff
@@ -39,90 +42,128 @@ public enum OperatingSystem {
...
+        // userAgent.contains(" edga/") || userAgent.contains(" edgios/")) {
+        // isEdge = true;
+        // isChrome = false;
+        // isOpera = false;
+        // isIE = false;
+        // isSafari = false;
+        // isFirefox = false;
+        // isWebKit = false;
+        // isGecko = false;
+        // }
```
mshabarov: Remove comments

flow-client/src/main/java/com/vaadin/client/BrowserInfo.java
```diff
@@ -79,26 +69,34 @@ public static BrowserInfo get() {
...
+        } catch(e){
+            return false;
+        }
     }-*/;
 
     /**
      * Checks if the browser is IE.
      *
      * @return true if the browser is IE, false otherwise
+     * @deprecated use a parsing library like ua-parser-js to parse the user
```
mshabarov, line 96: I wonder if we shall deprecate the whole `BrowserInfo` class?

mshabarov: Well, likely not, as we suggest to use raw string.

flow-client/src/main/java/com/vaadin/client/BrowserInfo.java
caalador: It isa basically a fix to not parse the user agent just for knowing if Chrome or IE and checking for the touch feature not depending on browser.

flow-client/src/main/java/com/vaadin/client/BrowserInfo.java
caalador: done

flow-client/src/main/java/com/vaadin/client/BrowserInfo.java
caalador: That's checked against feature and doesn't handle the UA-string

flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
caalador: Done

flow-server/src/main/java/com/vaadin/flow/shared/BrowserDetails.java
caalador: The whole class is deprecated so I don't think we need to deprecate each method. or?

mshabarov: Right, the current deprecation point are fine.

Issue #22294, opened 2025-09-19
Preventing default for events should be applied after the filter


Artur-:
### Description of the bug

```
        something.getElement().addEventListener("keydown", e -> {
            System.out.println("Space or enter pressed");
        }).setFilter("event.key === ' ' || event.key === 'Enter'")
          .preventDefault();
```

### Expected behavior

When you press space, the page should not scroll but pressing tab should still move focus to the next element.

### Minimal reproducible example

Above

### Versions

- Vaadin / Flow version: 25.0-SNAPSHOT

Artur-:
Fixed in #22298

PR #22298, opened 2025-09-19
fix: Apply preventDefault only to filtered events (#22294)

Artur-:
When preventDefault() is used with setFilter(), it now only prevents default behavior for events that match the filter. This allows for more granular control, e.g., preventing default only for space key while allowing tab key to function normally.

The fix modifies ElementListenerMap to check for preventDefault and stopPropagation expressions and make them conditional when a filter is present.


taefi (commented):
I'm not quite sure about this approach of making `preventDefault` and `setFilter` tied together this way. On one hand, it makes sense that preventDefault only matches the filtered events. On the other, it is not improving the granularity of the condition checking for filtering (what's delivered to the listener) and preventing the default behavior in the browser. 

There's other user cases that requires all the events delivered to the listener, while having a conditional prevention execution of `preventDefault()`. With this implementation, depending on the complexity of condition, there should be at list two listeners registered to a single DOM event: one with a filter and preventDefault, and another one with the complement condition and not `preventDefault`. 

I propose to have a conditional `preventDefault(condition)` instead to only filter what should be prevented, and `setFilter` continue doing business as usual. If it is about granularity, this can cover a much wider range of situation, and a simpler implementation, only at the cost of repeating a condition in both `setFilter` and `preventDefault` which is easier to read comparing to having two listeners with complementing conditions.

I have no statistics to prove which approach is more desired by the majority of the users, but so far there hasn't been that much of complaints about `preventDefault` not acting based on the filter (or I may have missed them).

Artur-:
I think we should optimize for the typical case and not the untypical. The typical case would be to prevent default for what is handled and that should not requiring duplicating the filter in a `preventDefault` method (and additionally then also in `stopPropagation`)

Artur-:
If you want to handle multiple (or all) keyevents and prevent default for some, it sounds fine to have two listeners

flow-server/src/main/java/com/vaadin/flow/dom/DomListenerRegistration.java
```diff
@@ -379,6 +379,10 @@ default public DomListenerRegistration stopPropagation() {
      * Tries to prevent the default behavior of the event in the browser, such
      * as shortcut action on key press or context menu on "right click". This
      * might not be possible for some events.
+     * <p>
+     * When used with {@link #setFilter(String)}, preventDefault will only be
+     * called when the filter condition is met. For example, if you set a filter
+     * for specific keys, preventDefault will only apply to those keys.
```
mshabarov, line 389: This should also be placed to the `stopPropagation()` javadoc.

flow-server/src/test/java/com/vaadin/flow/internal/nodefeature/ElementListenersTest.java
```diff
@@ -491,6 +491,89 @@ public void eventDataKeyNotPresentNotFail() {
         Assert.assertEquals(1, eventCount.get());
     }
 
+    @Test
+    public void testPreventDefaultWithFilter() {
```
mshabarov, line 495: Would be also good to have same tests for stopPropagation().

mshabarov (changes requested):
Requesting small additions, but overall looks good to me.

flow-server/src/test/java/com/vaadin/flow/internal/nodefeature/ElementListenersTest.java
Artur-: Done

flow-server/src/main/java/com/vaadin/flow/dom/DomListenerRegistration.java
Artur-: Done

Issue #22279, opened 2025-09-17
Make server-side modality less strict by default


mshabarov:
### Describe your motivation

We consider changing server-side modality enforcement to not be enabled by default while still being available as an opt-in feature. Background and motivation is given in [RFC](https://vaadin.com/forum/t/rfc-make-sever-side-modality-less-strict-by-default/178502).

### Describe the solution you'd like

- [ ] Add a new `void setModalityMode(ModalityMode mode)` method and a corresponding getter to `Dialog`, `ConfirmDialog` and `LoginOverlay`.
    `ModalityMode` is an enum in Flow with the values `MODELESS`, `VISUAL`, and `STRICT`.
    Flow provides an API to make it easier to toggle server-side modality from the components.
    - [ ] `STRICT` is the current default. There’s a visual modality curtain that darkens the view behind the dialog. Focus is trapped inside the dialog and pointer events are not passed through the curtain to background components. Requests to components behind that curtain are blocked unless there’s a low-level opt-in 
     through e.g. `DomListenerRegistration::allowInert` or similar mechanisms for other request triggers.
     - [ ] `VISUAL` behaves like `STRICT` except that request are not blocked.
     - [ ] `MODELESS` doesn’t use modality at all – no modality curtain, no focus trap, no pointer event blocking, and no request blocking.
- [ ] Change `Dialog` to use `VISUAL` by default whereas `LoginOverlay` and `ConfirmDialog` are STRICT by default.
- [ ] Deprecate the `void setModal(boolean)` and `boolean isModal()` methods in `Dialog`. (The other two components don’t have this API but are currently always modal.)
- [ ] Document this as a breaking change in Vaadin 25.0 with potential security implications in release notes. Recommend users to update their code to explicitly use `STRICT` if they rely on the current security guarantees.

### Additional context

Vaadin 25.0 is the target

Legioth:
As discussed in the forum, we should consider keeping `STRICT` as the default also for `ConfirmDialog` with the motivation that it's typically used in a situation where the user must make a choice and it's not a valid option to just dismiss the dialog and continue working with what's behind it. In that way, the only changed default would be for the base `Dialog` component.

mshabarov:
Thanks, I've corrected this in the description above.

PR #22302, opened 2025-09-19
 feat: add setChildComponentModal(ModalityMode) to UI

tltv:
Introduce setChildComponentModal(ModalityMode) helper in UI for components.

Part of: #22279


flow-server/src/main/java/com/vaadin/flow/component/UI.java
```diff
@@ -1542,50 +1542,72 @@ public String getCsrfToken() {
     /**
      * Adds the given component as a modal child to the UI, making the UI and
      * all other (existing) components added to the UI impossible for the user
-     * to interact with. This is useful for modal dialogs which should make the
-     * UI in the background inert. Note that this only prevents user
-     * interaction, but doesn't show a modality curtain or change the visible
-     * state of the components in the UI - that should be handled by the
-     * component separately. Thus this is purely a server side feature.
+     * to interact with. Using {@link ModalityMode#STRICT} mode. This is useful
```
mcollovati, line 1545: side note: Javadoc for `ModalityMode.STRICT`  say "A visual curtain dims the background behind the dialog", but this depends on the component implementation. Maybe it would be good to add a note there; otherwise people might expect that all using strict with `setChildComponentModal(...)` automatically provide the visual curtain.

tltv: Wouldn't it be enough to copy text here telling that showing modality curtain is up to the component into the new setChildComponentModal method? enum can be kept as it is, as a contract for those who use it.

mcollovati: Yeah, makes sense.

tltv: Added note. Also added it in the existing method.

PR #22303, opened 2025-09-19
chore: Deprecate Theme annotation and CssImport parameters

mshabarov:
Related-to https://github.com/vaadin/platform/issues/7453

### DIFF
--- a/flow-server/src/main/java/com/vaadin/flow/component/dependency/CssImport.java
+++ b/flow-server/src/main/java/com/vaadin/flow/component/dependency/CssImport.java
@@ -123,6 +123,7 @@ public @interface CssImport {
      *
      * @return the include value.
      */
+    @Deprecated(since = "25.0")
     String include() default "";
 
     /**
@@ -130,6 +131,7 @@ public @interface CssImport {
      *
      * @return the id.
      */
+    @Deprecated(since = "25.0")
     String id() default "";
 
     /**
@@ -137,7 +139,11 @@ public @interface CssImport {
      * target.
      *
      * @return the themable element.
+     *
+     * @deprecated Component styling can be done entirely with normal "light
+     *             DOM" CSS.
      */
+    @Deprecated(since = "25.0")
     String themeFor() default "";
 
     /**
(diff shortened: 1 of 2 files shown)


flow-server/src/main/java/com/vaadin/flow/component/dependency/CssImport.java
```diff
@@ -123,21 +123,27 @@
      *
      * @return the include value.
      */
+    @Deprecated(since = "25.0")
```
mshabarov, line 126: Don't know what must be written into `deprecated` parameter.

flow-server/src/main/java/com/vaadin/flow/component/dependency/CssImport.java
```diff
@@ -123,21 +123,27 @@
...
     String include() default "";
 
     /**
      * The 'id' of the new 'dom-module' created.
      *
      * @return the id.
      */
+    @Deprecated(since = "25.0")
```
mshabarov, line 134: Don't know what must be written into `deprecated` parameter.

flow-server/src/main/java/com/vaadin/flow/component/dependency/CssImport.java
caalador: Why is CssImport deprecated?
For instance Lumo specifically uses `@CssImport("@vaadin/vaadin-lumo-styles/lumo.css")` and the file is loaded in the global scope.

mshabarov: `CssImport` is not deprecated, but it's parameter are: `themeFor`, `include` and `id`. As stated in https://github.com/vaadin/platform/issues/7453. These are only needed for custom themes with `AtTheme`, but I don't recall exactly why :)

caalador: id and include seem to be for shared style sheets: https://vaadin.com/docs/v14/flow/styling/sharing-style-sheets#creating-shared-style-sheets
But there is no documentation for this in 24

caalador: Going through the code and samples I really can't find where include and id would have been used.
Most that I found was that they were added when the fallback bundle was made and that they were perhaps meant for usage with polymerTemplate and shadow dom?

Seem ok to just deprecate these fields.

caalador (approved):

PR #22304, opened 2025-09-20
chore: Add automated frontend dependency updates

Artur-:
- Add weekly GitHub workflow (Mondays 04:00 UTC) that runs the update script and opens a PR per branch via peter-evans/create-pull-request
- Matrix over main, 25.2, 25.1, 25.0, 24.10 and 24.9
- Use VAADIN_BOT_TOKEN so the created PRs trigger CI
- Update strategy is branch-based: latest versions on main, patch only on maintenance branches
- Let create-pull-request handle the commit/branch/push; drop committing from the update script

### DIFF
--- a/.github/workflows/update-frontend-dependencies.yml
+++ b/.github/workflows/update-frontend-dependencies.yml
@@ -1,0 +1,72 @@
+name: Update Frontend Dependencies
+
+on:
+  schedule:
+    # Runs at 04:00 UTC every Monday
+    - cron: '0 4 * * 1'
+  workflow_dispatch:  # Allow manual trigger
+
+jobs:
+  update-dependencies:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: [main, '25.2', '25.1', '25.0', '24.10', '24.9']
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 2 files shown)


Artur-:
This would need some changes to use vaadin-bot  or whatever should be used @manolo. Not sure how to test it either - the script itself works but the action is unclear

mcollovati:
Running the script on maintenance branches causes the build to succeed with no errors if there are deprecated packages that do not have a patch available but require a major version bump.

Could the script run `npx npm-check-updates@18.1.1 --no-deprecated` right after the upgrade attempt and make the build fail if deprecation warnings were found?

This way we could manually review each package and make a PR manually.

Artur-:
> Could the script run npx npm-check-updates@18.1.1 --no-deprecated right after the upgrade attempt and make the build fail if deprecation warnings were found?

Updated the script with that

Artur-:
Instead of running once a week, maybe it would be better if it ran once a day, used a dedicated branch so if a PR is left open, it can push additional updates in there

caalador:
Should this then also update the check to not update to <1day  old packages to have the PR always succeed on the first build?

Artur-:
Yes, updated

caalador (approved):
Seems ok apart from the open question if it should run every day instead of only on mondays.

Artur-:
Let's see how it works this way first

